PR #52, opened 2017-02-10
SAMZA-1086. New Grouper for ZK based standalone.

sborya:
SAMZA-1086. 
Create new grouper with support for arbitrary container ids.
Add support for this list of container IDs in the JobModelManager.


samza-core/src/main/java/org/apache/samza/container/grouper/task/BalancingTaskNameGrouper.java
```diff
@@ -54,5 +55,11 @@
    * @param localityManager provides a persisted task to container map to use as a baseline
    * @return                the grouped tasks in the form of ContainerModels
    */
-  Set<ContainerModel> balance(Set<TaskModel> tasks, LocalityManager localityManager);
+  default Set<ContainerModel> balance(Set<TaskModel> tasks, LocalityManager localityManager) {
```
fredji97: Nice to see Java 8 feature!

samza-core/src/main/java/org/apache/samza/container/grouper/task/SimpleGroupByContainerCount.java
```diff
@@ -0,0 +1,102 @@
...
+    for (int i = 0; i < containerCount; i++) {
+      taskGroups[i] = new HashMap<>();
+    }
+    for (int i = 0; i < sortedTasks.size(); i++) {
+      TaskModel tm = sortedTasks.get(i);
+      taskGroups[i % containerCount].put(tm.getTaskName(), tm);
+    }
+
+    // Convert to a Set of ContainerModel
+    Set<ContainerModel> containerModels = new HashSet<>();
```
fredji97: What is the reason of using a set instead of an array for containerModels? If we are using the array, we just keep the taskGroups[i] for each element of the array, and it seems better to me for the quick access on index.

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskNameGrouper.java
```diff
@@ -50,4 +51,8 @@
    * @return Set of containers, which contain the tasks that were passed in.
    */
   Set<ContainerModel> group(Set<TaskModel> tasks);
+
+  default Set<ContainerModel> group(Set<TaskModel> tasks, List<Integer> containersIds) {
+    return null;
```
fredji97: If an implemented class uses this default group method, does it work with returned null or it has to override with some implementation?

samza-core/src/main/java/org/apache/samza/container/grouper/task/SimpleGroupByContainerCount.java
sborya: Mainly, that the output of this method is defined as Set. So in any case we will need a conversion.

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskNameGrouper.java
sborya: It just returns null. No override is necessary.

navina: Returning null seems problematic to be used as the default behavior. What does it mean when caller invokes the grouper with a set of task models and a list of containerIds. Returning null implies that grouping cannot be done? Or that there was an error ??

samza-core/src/main/java/org/apache/samza/container/grouper/task/BalancingTaskNameGrouper.java
```diff
@@ -54,5 +55,11 @@
...
+    return Collections.<ContainerModel>emptySet();
+  }
+  default Set<ContainerModel> balance(Set<Integer> containerIds, Set<TaskModel> tasks, LocalityManager localityManager) {
```
navina: why do we need this balance method override? I don't see any usage of this

samza-core/src/main/java/org/apache/samza/container/grouper/task/BalancingTaskNameGrouper.java
navina: Why is this now a default method ?

sborya: Otherwise each one of existing classes that extends this class would need to override the method. This way only the new class will override it.

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskNameGrouper.java
sborya: I see your point. I changed it to call the other method (basically ignoring the containers list).

samza-core/src/main/java/org/apache/samza/container/grouper/task/BalancingTaskNameGrouper.java
sborya: When we implement balancing for standalone we will need this method, correct?

navina: Ok. Like we discussed yesterday, I want to clarify a couple of things before we merge this interface change:
1. Why is "balance" method any different than "group" ? The only difference is the use of Locality information in grouping/balancing tasks. Was the reason to introduce BalancingTaskNameGrouper to avoid breaking existing interface or to define a functionality that is different than group? My understanding is the former and hence, we should move towards unifying the 2 interface methods.
2. There is no BalancingTaskNameGrouper implementation "yet" in standalone or this patch. So, if it is not being used currently, I would recommend not adding it now. My two cents. 

@nickpan47 @jmakes  any thoughts?

sborya: I agree we should review if we need both of them. But for now I am just going to remove the unused balance() api. We will added it or merge with group() when we add Locality feature to standalone.

samza-core/src/main/java/org/apache/samza/container/grouper/task/BalancingTaskNameGrouper.java
xinyuiscool: Wait, this seems not making much sense to me if this interface has only one api method and it's a default. If existing classes doesn't want to implement the single api method, why bother implementing this interface? Seems pretty confusing to me.

samza-core/src/main/java/org/apache/samza/container/grouper/task/SimpleGroupByContainerCount.java
```diff
@@ -0,0 +1,102 @@
...
+  public Set<ContainerModel> group(Set<TaskModel> tasks, List<Integer> containersIds) {
+    if(tasks.isEmpty())
+      throw new IllegalArgumentException("cannot group an empty set. containersIds=" + Arrays
+          .toString(containersIds.toArray()));
+
+    if(containersIds.size() > tasks.size())
+      throw new IllegalArgumentException("number of containers "  + containersIds.size() + " is bigger than number of tasks " + tasks.size());
+
+    if (containersIds == null)
+      return this.group(tasks);
```
xinyuiscool: what about the containerids being empty?

samza-core/src/main/java/org/apache/samza/container/grouper/task/SimpleGroupByContainerCount.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  public SimpleGroupByContainerCount(int containerCount) {
+    if (containerCount <= 0) throw new IllegalArgumentException("Must have at least one container");
+    this.startContainerCount = containerCount;
+  }
+
+  @Override
+  public Set<ContainerModel> group(Set<TaskModel> tasks) {
+    if(tasks.isEmpty())
```
xinyuiscool: nit pick: Please add {} after "if". it's less error-prone. There are a few "if" without {}

samza-core/src/main/java/org/apache/samza/container/grouper/task/SimpleGroupByContainerCount.java
```diff
@@ -0,0 +1,102 @@
...
+ * It exposes two group methods - one that assumes sequential container numbers and one that gets a set of container
+ * IDs as an argument. Please note - this first implementation ignores locality information.
+ */
+public class SimpleGroupByContainerCount implements TaskNameGrouper {
+  private final int startContainerCount;
+  public SimpleGroupByContainerCount() {
+    this.startContainerCount = 1;
+  }
+
+  public SimpleGroupByContainerCount(int containerCount) {
```
xinyuiscool: Is this constructor being used? Seems pretty confusing we can pass in a containerCount here and later in group(tasks, containerIds) this count is not being used. If this is not needed then we just need a a private method of group(tasks, n) and used by both group(tasks) and group(tasks, containerIds).

samza-core/src/main/java/org/apache/samza/container/grouper/task/SimpleGroupByContainerCount.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Set;
+
+
+/**
+ * Simple grouper.
+ * It exposes two group methods - one that assumes sequential container numbers and one that gets a set of container
+ * IDs as an argument. Please note - this first implementation ignores locality information.
+ */
+public class SimpleGroupByContainerCount implements TaskNameGrouper {
+  private final int startContainerCount;
```
xinyuiscool: this is not used in group(Set<TaskModel> tasks, List<Integer> containersIds). why we need this?

samza-core/src/main/java/org/apache/samza/container/grouper/task/BalancingTaskNameGrouper.java
sborya: right, I was cleaning this file on Navina's suggestion, but didn't clean completely. Fixing.

samza-core/src/main/java/org/apache/samza/container/grouper/task/SimpleGroupByContainerCount.java
sborya: but it is used in group(Set tasks).

samza-core/src/main/java/org/apache/samza/container/grouper/task/SimpleGroupByContainerCount.java
sborya: we do needed it for group(tasks). 
I agree it may be simplified but group(tasks) is an existing interface, while group(tasks, containerIds) is the new one.

sborya:
changed the name to GroupByContainerIds

PR #53, opened 2017-02-15
Samza 1075 - Refactor SystemAdmin Interface to expose a common method to create streams.

jmakes:


samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -0,0 +1,159 @@
...
+   * @param physicalName  The physical identifier for the stream. This is the identifier that will be used in remote
+   *                      systems to identify the stream. In Kafka this would be the topic name whereas in HDFS it
+   *                      might be a file URN.
+   *
+   * @param systemName    The System name on which this stream will exist. Corresponds to a named implementation of the
+ *                        Samza System abstraction. See {@link SystemFactory}
+   *
+   * @param properties    A set of properties for the stream. These may be System-specfic.
+   */
+  public StreamSpec(String id, String physicalName, String systemName, Properties properties) {
```
nickpan47: Minor: it would be nice to add documentation showing the corresponding config variables accordingly. e.g.

StreamSpec("mystream", "my-topic", "kafka1", Properties...)
is equivalent to config:
streams.mystream.system=kafka1
streams.mystream.name=my-topic
streams.mystream.prop1=x1
...

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -89,4 +91,25 @@
    * @return -1 if offset1 &lt; offset2; 0 if offset1 == offset2; 1 if offset1 &gt; offset2. Null if not comparable
    */
   Integer offsetComparator(String offset1, String offset2);
+
+  /**
+   * Create a stream described by the spec. If the stream already exists, this
+   * call should simply return.
```
nickpan47: return true or false?

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -89,4 +91,25 @@
...
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Validates the stream described by the streamSpec on the system.
+   * An exception should be thrown for any validation error.
+   *
+   * @param streamSpec  The spec, or blueprint for the physical stream on the system.
+   */
+  default void validateStream(StreamSpec streamSpec) {
```
nickpan47: shouldn't we declare the method to throw exceptions, since the javadoc says so?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaStreamSpec.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Properties;
+import org.apache.samza.config.KafkaConfig;
+import org.apache.samza.system.StreamSpec;
+
+
+/**
+ * Extends StreamSpec with the ability to easily get the topic replication factor.
+ */
+public class KafkaStreamSpec extends StreamSpec {
+  private static final int DEFAULT_PARTITION_COUNT = 1;
```
nickpan47: QQ: why only partition count and replication factor? Shouldn't cleanup policy and segment size also be part of the Kafka specific specs?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaStreamSpec.java
```diff
@@ -0,0 +1,102 @@
...
+   *
+   * @param id                The application-unique logical identifier for the stream. It is used to distinguish between
+   *                          streams in a Samza application so it must be unique in the context of one deployable unit.
+   *                          It does not need to be globally unique or unique with respect to a host.
+   *
+   * @param topicName         The physical identifier for the stream. This is the identifier that will be used in remote
+   *                          systems to identify the stream. In Kafka this would be the topic name whereas in HDFS it
+   *                          might be a file URN.
+   *
+   * @param systemName        The System name on which this stream will exist. Corresponds to a named implementation of the
```
nickpan47: It maybe just me. I felt it more natural to have systemName before topicName, since the scope of system is bigger. The same to the method signature in the parent class StreamSpec.

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -0,0 +1,159 @@
...
+   * Samza System abstraction.
+   */
+  private final String systemName;
+
+  /**
+   * The physical identifier for the stream. This is the identifier that will be used in remote
+   * systems to identify the stream. In Kafka this would be the topic name whereas in HDFS it
+   * might be a file URN.
+   */
+  private String physicalName;
```
cpettitt-linkedin: final?

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -0,0 +1,159 @@
...
+
+  /**
+   * The number of partitions for the stream.
+   */
+  private int partitionCount = 1;
```
cpettitt-linkedin: final?

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -147,16 +161,22 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
     kafkaChangeLogProperties.setProperty("cleanup.policy", "compact")
     kafkaChangeLogProperties.setProperty("segment.bytes", KafkaConfig.CHANGELOG_DEFAULT_SEGMENT_SIZE)
     kafkaChangeLogProperties.setProperty("delete.retention.ms", String.valueOf(new StorageConfig(config).getChangeLogDeleteRetentionInMs(name)))
-    filteredConfigs.foreach{kv => kafkaChangeLogProperties.setProperty(kv._1, kv._2)}
+    filteredConfigs.foreach { kv => kafkaChangeLogProperties.setProperty(kv._1, kv._2) }
     kafkaChangeLogProperties
   }
 
+  def getTopicKafkaProperties(systemName: String, streamName: String) = {
```
nickpan47: So, I guess that the config variable name changes are not part of this PR's scope? It would be nice to open another JIRA to track the config name structure change, if planned,

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -0,0 +1,159 @@
...
+
+
+/**
+ * StreamSpec is a blueprint for creating, validating, or simply describing a stream in the runtime environment.
+ *
+ * It has specific attributes for common behaviors that samza uses.
+ *
+ * It also includes a set of properties which may be systemName-specific.
+ */
+public class StreamSpec {
```
cpettitt-linkedin: I see how this fits nicely with the current supported streams, but what would the system name and physical name be for an array based stream (for example)? Also, in a world where we support different types of streams wouldn't we need to know how to construct that stream (maybe that is what is meant by system name now?). Maybe we're not using StreamSpecs for arrays, etc.? If not, what is the alternative?

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -0,0 +1,159 @@
...
+
+  /**
+   * The number of partitions for the stream.
+   */
+  private int partitionCount = 1;
+
+  /**
+   * A set of all systemName-specific properties for the stream.
+   */
+  private final Properties properties = new Properties();
```
cpettitt-linkedin: Map<Object, Object> is nicer (does not require synchronization).

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -0,0 +1,159 @@
...
+
+    if (systemName == null) {
+      throw new NullPointerException("Parameter 'systemName' must not be null");
+    }
+
+    this.id = id;
+    this.systemName = systemName;
+    this.physicalName = physicalName;
+
+    if (properties != null) {
```
cpettitt-linkedin: If we go with a map, we can make this spec immutable (highly desirable) by copying the properties and then wrapping them with an immutable map wrapper.

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -0,0 +1,159 @@
...
+  public int getPartitionCount() {
+    return partitionCount;
+  }
+
+  public void setPartitionCount(int partitionCount) {
+    this.partitionCount = partitionCount;
+  }
+
+  public Properties getProperties() {
+    return properties;
```
cpettitt-linkedin: Just to reinforce some of the points above: this actually allows the backing collection to be manipulated post construction. If we use the techniques above we are protected against this.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -502,24 +500,32 @@ class KafkaSystemAdmin(
...
   override def createChangelogStream(topicName: String, numKafkaChangelogPartitions: Int) = {
-    createTopicInKafka(topicName, numKafkaChangelogPartitions)
-    validateChangelogStream(topicName, numKafkaChangelogPartitions)
+    val topicMeta = topicMetaInformation.getOrElse(topicName, throw new KafkaChangelogException("Unable to find topic information for topic " + topicName))
+    val spec = new KafkaStreamSpec(topicName, topicName, systemName, numKafkaChangelogPartitions, topicMeta.replicationFactor, topicMeta.kafkaProps)
+
+    if (createStream(spec)) {
+      info("Created stream %s." format topicName)
+    } else {
+      info("Changelog stream %s already exists." format topicName)
```
nickpan47: The javadoc for createStream() needs to include a bit more info regarding to the failure cases. Under which condition should the user expect a "false"? Is there any exception thrown instead of true/false result?

The log line here seems to indicate that "false" means the topic already exists. Is it always the case?

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
jmakes: The systemName would still apply for Collection streams; it would correspond to a new CollectionSystemFactory. In other words, we'd continue to leverage the System abstraction, even for trivial cases like Collections.

The physical name would be unnecessary for any local-only system. But as the existence or absence of that field would be the only difference between local and remote systems, I didn't feel that a more complicated class hierarchy was justified.

I think all the questions were answered by the above two statements. If not, let me know.

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
jmakes: Could be for this patch, but I didn't want to rule out something dynamic. e.g. would DALI need to update the physicalName based on a view query?

I don't feel strongly about this. Happy to change it if you think it's better.

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
jmakes: This one can't be unless we do copy on write. The planner in the ExecutionEnvironment will set the partition count on intermediate StreamSpecs which originate in the logical layer.

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
jmakes: If the synchronization is the only reason, I'd argue that the benefits of Map<Object, Object> are not worth the pain. 

StreamSpecs are not a frequently-used object and I don't want to prematurely optimize them. By using Properties, we get easier integration/translation with config and Kafka, our primary system. 

Are there other benefits of Map<Object, Object> that I'm not considering?

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
cpettitt-linkedin: Sounds good. We can optionally play around with the constructor when we get to collection based streams.

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
cpettitt-linkedin: It's easier to relax restrictions later than the other way around.

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
cpettitt-linkedin: Given how close we are to an immutable data structure, I would be OK with that tradeoff - especially since planning is not going to be an ongoing cost we pay while actually running the job. I'm pretty heavily biased in that direction though.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaStreamSpec.java
jmakes: Partition count and replication factor are required explicitly by the Kafka AdminUtils, so it's nice to not lump them in with the "bag of properties"

We could add the other ones, but:
1. Where should the line be drawn? Kafka has a lot of topic configs.
2. They would need to be added to the "bag of properties" that gets passed to the AdminUtils.createTopic(), so it just complicates the code.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
jmakes: Yes. Thanks for catching that.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
jmakes: Yes. Thanks for catching that too.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaStreamSpec.java
jmakes: We can order hierarchically or by salience. I'll change it if you feel strongly about it.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
jmakes: Correct. That'll be handled by the ExecutionEnvironment's static constructors for StreamSpec

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
jmakes: Yes, that is always the case. Will update the doc

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
jmakes: Good point. I'll make the change.

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
cpettitt-linkedin: I agree with your motivation to preserve simple interaction with Kafka and Config.

One additional benefit to the Map approach - btw, it should have been Map<String, String> not Map<Object, Object> - would be immutability. We could get the best of both by continuing to take a Properties via the constructor and providing a method to convert to Properties. Both directions should be trivial.

Your call.

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
jmakes: I'm a big fan of immutability too. I'll make the change, and we can revisit if it doesn't jive well with the API, planner, etc.

Thanks!

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
jmakes: I like this idea. Feels more flexible. 
Stay tuned for a new patch!

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
jmakes: I agree that this could be useful. But it probably doesn't belong here. StreamSpec should not be directly concerned with config structure.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
nickpan47: Great. Then, please ignore my comments earlier regarding to the config variable examples in javadoc.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaStreamSpec.java
nickpan47: No strong view on this one. Your call.

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
jmakes: I think the latest patch addresses everything. 

Note: in order to avoid confusion RE: Properties vs. Map, I renamed this field to "config". Not sure I like it, so I welcome suggestions.

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -0,0 +1,197 @@
...
+      throw new NullPointerException("Parameter 'systemName' must not be null");
+    }
+
+    this.id = id;
+    this.systemName = systemName;
+    this.physicalName = physicalName;
+    this.partitionCount = partitionCount;
+
+    if (config != null) {
+      this.config = Collections.unmodifiableMap(config);
```
cpettitt-linkedin: You also need to copy the map here. Otherwise it would be possible to modify the map if you had a reference to the backing map.

Something like this would work:

```
this.config = Collections.unmodifiableMap(new HashMap<String, String>(config));
```

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    if (systemName == null) {
+      throw new NullPointerException("Parameter 'systemName' must not be null");
+    }
+
+    this.id = id;
+    this.systemName = systemName;
+    this.physicalName = physicalName;
+    this.partitionCount = partitionCount;
```
nickpan47: Shouldn't we also assert that partitionCount > 0? I saw some assertion of other parameters in the implementation of KafkaStreamSpec, couldn't help wondering assertion on partitionCount would also be desirable?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -502,24 +493,32 @@ class KafkaSystemAdmin(
...
   }
 
   override def createChangelogStream(topicName: String, numKafkaChangelogPartitions: Int) = {
-    createTopicInKafka(topicName, numKafkaChangelogPartitions)
-    validateChangelogStream(topicName, numKafkaChangelogPartitions)
+    val topicMeta = topicMetaInformation.getOrElse(topicName, throw new KafkaChangelogException("Unable to find topic information for topic " + topicName))
+    val spec = new KafkaStreamSpec(topicName, topicName, systemName, numKafkaChangelogPartitions, topicMeta.replicationFactor, topicMeta.kafkaProps)
+
+    if (createStream(spec)) {
+      info("Created stream %s." format topicName)
```
nickpan47: nit: since this is within the method of createChangelogStream, it would make sense to log more specific message as:
"Created changelog stream %s"...

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -143,13 +145,17 @@ public StreamSpec(String id, String physicalName, String systemName, int partiti
       throw new NullPointerException("Parameter 'systemName' must not be null");
     }
 
+    if (partitionCount < 1) {
+      throw new NullPointerException("Parameter 'partitionCount' must not be greater than 0");
```
nickpan47: Two comments on this one:
- Shouldn't this be IllegalArgumentException?
- "Parameter 'partitionCount' must be greater than 0"

jmakes: Really screwed that one up, didn't I?
I will fix it in a separate PR, since this one is already merged

PR #50, opened 2017-02-09
SAMZA-1102: Zk controller

sborya:
SAMZA-1102: Added ZKController and ZkControllerImpl

PR #54, opened 2017-02-15
SAMZA-1084 - User thread does not see errors from the processor thread

navina:
Adding a ProcessorLifecycleCallback for specifying user-defined error handling

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/processor/ProcessorLifecycleCallback.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/ProcessorLifecycleCallback.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.processor;
+
+import org.apache.samza.annotation.InterfaceStability;
+
+@InterfaceStability.Evolving
+public interface ProcessorLifecycleCallback {
+  /**
+   * Callback method that is invoked when the processing loop exits with an exception/error.
+   *
+   * Note: This callback is invoked in the context of the StreamProcessor's container thread.
+   *
+   * @param error Any kind of {@link Throwable} thrown by the container's run loop
+   */
+  void onProcessorError(Throwable error);
+}
--- a/samza-core/src/main/java/org/apache/samza/processor/ProcessorLifecycleCallback.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/ProcessorLifecycleCallback.java
@@ -29,5 +29,5 @@ public interface ProcessorLifecycleCallback {
    *
    * @param error Any kind of {@link Throwable} thrown by the container's run loop
    */
-  void onProcessorError(Throwable error);
+  void onError(Throwable error);
 }
--- a/samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
@@ -22,10 +22,8 @@ package org.apache.samza.processor;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.apache.samza.config.ClusterManagerConfig;
 import org.apache.samza.config.Config;
-import org.apache.samza.config.TaskConfigJava;
 import org.apache.samza.container.LocalityManager;
 import org.apache.samza.container.SamzaContainer;
-import org.apache.samza.container.SamzaContainer$;
 import org.apache.samza.job.model.ContainerModel;
 import org.apache.samza.metrics.JmxServer;
 import org.apache.samza.metrics.MetricsReporter;
@@ -34,25 +32,21 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Map;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 
 public class SamzaContainerController {
   private static final Logger log = LoggerFactory.getLogger(SamzaContainerController.class);
 
-  private final ExecutorService executorService;
   private volatile SamzaContainer container;
   private final Map<String, MetricsReporter> metricsReporterMap;
   private final Object taskFactory;
-  private final long containerShutdownMs;
   private final ProcessorLifecycleCallback lifecycleCallback;
 
   // Internal Member Variables
   private Future containerFuture;
+  private ExecutorService executorService;
 
   /**
    * Creates an instance of a controller for instantiating, starting and/or stopping {@link SamzaContainer}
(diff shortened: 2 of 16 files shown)


samza-core/src/main/java/org/apache/samza/processor/ProcessorLifecycleCallback.java
```diff
@@ -0,0 +1,33 @@
...
+@InterfaceStability.Evolving
+public interface ProcessorLifecycleCallback {
+  /**
+   * Callback method that is invoked when the processing loop exits with an exception/error.
+   *
+   * Note: This callback is invoked in the context of the StreamProcessor's container thread.
+   *
+   * @param error Any kind of {@link Throwable} thrown by the container's run loop
+   */
+  void onProcessorError(Throwable error);
```
vjagadish1989: nit: ProcessorError is probably redundant. Can have it name as `onError()`.

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
```diff
@@ -106,7 +109,17 @@ public void startContainer(ContainerModel containerModel, Config config, int max
...
     log.info("About to start container: " + containerModel.getContainerId());
-    containerFuture = executorService.submit(() -> container.run());
+    containerFuture = executorService.submit(() -> {
+        try {
+          container.run();
+        } catch (Throwable t) {
+          if (errorHandler != null) {
+            errorHandler.onProcessorError(t);
+          } else {
+            throw t;
```
vjagadish1989: Question: If this is running in a different thread, does it still make sense to throw the error? 
Will it make sense to log this error? (Or is logging handled higher or lower up the stack)?

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
```diff
@@ -132,7 +132,7 @@ public void testStreamProcessorWithNoTask() {
     configMap.remove("task.class");
     final Config configs = new MapConfig(configMap);
 
-    StreamProcessor processor = new StreamProcessor(1, configs, new HashMap<>());
```
vjagadish1989: Optionally, we can add an unit test on processor error.

vjagadish1989 (approved):
Looks good! approved

final round of feedbackk.
Thanks

samza-core/src/main/java/org/apache/samza/processor/ProcessorLifecycleCallback.java
navina: Sure

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
navina: We can log error. I actually didn't know what to do in the else block, because even if you throw here, the user doesn't have the handle to the container's future. So, its pretty useless. I think logging and then, throwing is probably better. Let me add that.

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
navina: I couldn't add unit test because the callback in invoke inside a Runnable in SamzaContainerController. And, that class uses a static method to create the container instance. So, at this point, I don't know how to add the unit test without extensive surgery to the existing code :) if you have any suggestions, please let me know :)

vjagadish1989: Not sure if there's extensive surgery needed in existing code.  But, it seems that there's probably work involved in writing the test. So, this is not a blocker.
- I think there is a constructor in `StreamProcessor` already takes in a taskFactory. 
- You can make the passed-in taskFactory to return an instance of `StreamTask` that throws an exception in its methods.

navina:
@vjagadish1989 Yeah. I don't have problem creating a task that throws exception. The problem is ensuring that the correct execution path is triggered through my unit test. It can be done only when I bring up a fully operational processor, like I do in TestStreamProcessor. I felt it was a tad-bit unnecessary for adding this test at this point. Thanks for your quick review, btw! 

@xinyuiscool Check [here](https://issues.apache.org/jira/browse/SAMZA-1084?focusedCommentId=15866556&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15866556) for details. It basically didn't seem like a big value add for the user. I remember this is a different conclusion from what we had discussed previously.
> I am not sure whether you plan to add more samza specific stuff to this ProcessorLifecyleCallback interface. 

As of today, we don't intend to include anything else on this callback. 

> If we just need to know when it's complete and failure, you can use java8 CompletableFuture. You can wire up your single thread executor with that easily (something like this: http://www.nurkiewicz.com/2015/11/which-thread-executes.html)

That sounds awesome! I didn't know about this. Let me give it a read and see if it works for us. Thanks!

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
```diff
@@ -67,6 +68,7 @@ public SamzaContainerController(
       Object taskFactory,
       long containerShutdownMs,
       String processorId,
+      ProcessorLifecycleCallback errorHandler,
```
cpettitt-linkedin: We should document errorHandler and also note that it can be null.

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
```diff
@@ -107,7 +110,18 @@ public void startContainer(ContainerModel containerModel, Config config, int max
         Util.<String, MetricsReporter>javaMapAsScalaMap(metricsReporterMap),
         taskFactory);
     log.info("About to start container: " + containerModel.getContainerId());
-    containerFuture = executorService.submit(() -> container.run());
+    containerFuture = executorService.submit(() -> {
```
cpettitt-linkedin: Not related to this commit, but it appears that there is nothing to stop two consecutive calls to startContainer. It appears that the intention of this class is to start and stop a container once (since we do not null out container in stop and we check in shutdown if the container is created). Given that, I would suggest checking to see if startContainer has already been called here. I would probably also consolidate stopContainer and shutdown into one method - shutdown really does mean shutdown with a single use controller.

---

Also, again not strictly related to this commit: there seems to be some mismatch between startContainer and stopContainer. The former runs asynchronously while the latter runs synchronously. I would either make them consistent (and e.g. have an awaitShutdown) or distinguish the behaviors through the method name (e.g. shutdownSync).

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
cpettitt-linkedin: +1 to adding a test. I disagree that it is optional though.

---

Again not necessarily related to this specific commit, but I would also test start / start, start / stop / stop, and stop without start.

navina: Ok. Let me take another look at adding tests. 

>  Again not necessarily related to this specific commit, but I would also test start / start, start / stop / stop, and stop without start.

Initial implementation goals were to allow start and stop on a processor only once. Hence, I remember adding the tests in the LinkedIn patch. But I guess code has evolved quite a bit and I missed those tests. I will address that case as well.

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
navina: Got it. Will do

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
navina: >  Given that, I would suggest checking to see if startContainer has already been called here.

Yeah. You are right. Let me try to add in that check. 

> I would probably also consolidate stopContainer and shutdown into one method - shutdown really does mean shutdown with a single use controller.

Agreed. 

> I would either make them consistent (and e.g. have an awaitShutdown) or distinguish the behaviors through the method name (e.g. shutdownSync).

Let me consolidate stop and shutdown into one. Then, I can evaluate which option works better - shutdownSync or awaitShutdown. I don't have an opinion on this one way or the other. 

Thanks for the suggestions, @cpettitt  ! Very useful.

navina:
@cpettitt-linkedin I tried to make the changes you recommended in terms of keeping either synchronous start & stop or asynchronous start & stop.  Here are the roadblocks I have hit :)

* On a fully synchronous API: ZkJobCoordinator is fully asynchronous ( See PR #70 & PR #50 ) - meaning, streamProcessor.start() simply starts the Zk JobCoordinator - which will participate in the leader election and then, WAIT on notification for the jobmodel to be available. Waiting for the notification is not time-bound and can take very long time , if the processors group is constantly changing. Hence, the creating a blocking call with a timeout may not be a good idea for StreamProcessor user. 
(Additionally, ZkJobCoordinator implementation doesn't bind to the contract defined in JobCoordinator interface :( I have created a JIRA to fix this [SAMZA-1147](https://issues.apache.org/jira/browse/SAMZA-1147) )

* On the other hand, if we make start/stop asynchronous, we should provide callbacks for the user. I took a stab at adding callbacks for the processor lifecycle - onStart (invoked just before container starts the runloop), onStop (invoked when container has fully stopped) and onError (invoked when container fails/throws an exception). This looks good, however, it does not capture any error that can happen in the Job Coordinator. I am little confused about whether it sufficiently provides the user control over the behavior of the processor. 

* Another point of difference (between pre-Zk and post-Zk world), StreamProcessor managed both jobcoordinator and container. But now, we have refactored it in a way that StreamProcessor manages JobCoordinator, which in turn, controls the container life-cycle. I feel like the latter is not the best approach as it seems to pollute the JobCoordinator interface with StreamProcessor requirements. 

These are my 2 cents. We can sync-up in-person if it is hard to evaluate the options in the PR.

cpettitt-linkedin:
Probably easiest to discuss. My calendar is pretty open today and tomorrow if you'd like to set something up.

navina:
Couple of notes on recent changes that I pushed. 
1. We are missing "state" definitions for each of the components interacting in StreamProcessor - namely, JobCoordinator and SamzaContainer. This makes it difficult for us to define the state of the StreamProcessor. For now, it merely keeps track of whether JobCoordinator was started or not (JobCoordinator does not provide a way to check status). Making these changes is much larger scope and I hope to tackle it in SAMZA-1147 and SAMZA-1150. 
2. Similarly, there is no clear error handling semantics for JobCoordinator. It was straightforward with StandaloneJobCoordinator because JobCoordinator was simply a pass-through. So, we were interested only in failures in SamzaContainer. With the introduction of other JobCoordinator implementations, we need to handle the interaction between JobCoordinator and Container to propagate errors up the stack. With this patch, we propagate only container errors using the callback. With SAMZA-1150, errors from jobcoordinator should also propagate to the user.

I have kept the above changes in separate JIRAs as it has much larger scope.

navina:
SAMZA-1084 is subsumed by SAMZA-1212 that introduces a listener based model for SamzaContainer (PR #148 ). Hence, closing this PR.

PR #58, opened 2017-02-18
SAMZA-1092: replace stream spec in fluent API

nickpan47:
Replaced the StreamSpec class w/ the new one from SAMZA-1075.


samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -200,4 +200,8 @@ public String get(String propertyName) {
   public String getOrDefault(String propertyName, String defaultValue) {
     return config.getOrDefault(propertyName, defaultValue);
   }
+
+  public SystemStream getSystemStream() {
```
jmakes: This should be avoided, if possible. Since the systemStream is using the physical name:
1. It won't work properly for local systems like CollectionSystemAdmin because there is no physicalName for that one.
2. The physical name is more of a runtime thing, so if the logical layer is depending on it, even indirectly, it's a violation of the separation of concerns.

If the graph needs an identifier, it should use getId()

nickpan47: Make sense. So, that probably means that all usage as "unique id" for a stream should be changed to use StreamSpec.getId() in the programming layer, and SystemStream should only be visible by the ExecutionEnvironment. I will make the change.

PR #60, opened 2017-02-23
SAMZA-1091: Implement key-based inner join operator with no time constraints

prateekm:


samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -116,38 +124,49 @@ public K getKey(M message) {
...
       }
 
       @Override
-      public K getKey(OM message) {
+      public K getKey(JM message) {
         return joinFn.getSecondKey(message);
       }
 
       @Override
-      public K getOtherKey(M message) {
```
vjagadish1989: Nice! This method is no longer required. The newer API for `PartialJoinFn` is also more general.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -113,12 +113,13 @@
...
    * @param otherStream the other {@link MessageStream} to be joined with
    * @param joinFn the function to join messages from this and the other {@link MessageStream}
+   * @param ttlMs the ttl in ms for retaining messages in each stream
    * @param <K> the type of join key
    * @param <OM> the type of messages in the other stream
    * @param <RM> the type of messages resulting from the {@code joinFn}
    * @return the joined {@link MessageStream}
    */
-  <K, OM, RM> MessageStream<RM> join(MessageStream<OM> otherStream, JoinFunction<K, M, OM, RM> joinFn);
+  <K, OM, RM> MessageStream<RM> join(MessageStream<OM> otherStream, JoinFunction<K, M, OM, RM> joinFn, long ttlMs);
```
vjagadish1989: I wonder if this API can take a `Duration` instead. 
* It's slightly consistent with the rest of our APIs for windows.
* `Duration.ofHours(2)` is arguably more readable(?) than 2 * 60 * 60 * 1000 milliseconds

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
```diff
@@ -20,28 +20,70 @@
...
+  public void onTimer(MessageCollector collector, TaskCoordinator coordinator) {
+    KeyValueStore<K, PartialJoinMessage<M>> thisState = thisPartialJoinFn.getState();
+    KeyValueIterator<K, PartialJoinMessage<M>> iterator = thisState.all();
+    List<K> keysToRemove = new ArrayList<>();
+    long now = System.currentTimeMillis();
+
+    while (iterator.hasNext()) {
+      Entry<K, PartialJoinMessage<M>> entry = iterator.next();
+      if (entry.getValue().getTimestamp() < now - ttlMs) {
+        keysToRemove.add(entry.getKey());
```
vjagadish1989: Instead of adding the keys to a separate list and iterating over the list again (using `deleteAll`), I wonder if it will be efficient to invoke remove() on the iterator right away? 

AFAIK, it should be safe modify the map through the same iterator?

vjagadish1989 (commented):
first round of feedback. 

Mostly minor questions/ comments. looking quite good. 

thanks for the contributions

renato2099:
This is really cool @prateekm ! I have some thoughts on it :)
If this is an unwindowed join, and we are setting the TTL for tuples in the joinState, shouldn't we  have a default TTL value so we don't get OOM exception too often? wdyt? Also the  "purging procedure" of removing tuples from the join state should also include removing old tuples from each relation internal state, so they don't grow unboundedly either. Maybe this is on the patch and I missed it ;)

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
prateekm: Makes sense, will update for consistency.

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
prateekm: This is a leftover from when I was using InMemoryKVStore whose Iterator does not support remove. Wondering if we should a pattern that we know that one of the default KVStore implementations don't support.

prateekm:
@renato2099 Thanks for the feedback!

For some context, this is an early version of the Join API and implementation. We'll add support for more join types and a persistent state store soon.

1. A default TTL: We decided against adding one in this implementation since it's hard for the framework to make a reasonable guess about the value - it would depend on the memory allocated to the container and the incoming message throughput and size. The plan is to add support for a persistent state store soon, so in the mean time it's probably better to let the users set a value based on their expected usage.

2. Expiring from relation internal state: In this implementation there's no separate join state, only the internal state for each relation (contained in its "PartialJoinFunction"). Join results are emitted as matches are found, and entries are purged from the relation's internal state when they're older than the TTL. There's another possible version where a match causes the entries corresponding to the emitted results to be removed immediately, but we decided to implement this more general version instead.

Does this make sense?

renato2099:
Thanks for the reply @prateekm ! :)
For (1) fair enough, it's better to make the user define a proper TTL.
For (2) I see what you mean. I thought the two in-memory hashmaps were used to keep both streams states, and that the join result was also kept in another hashmap which could be consumed from the next operator or handed to the user who would see it as an evolving data set.
But yeah makes sense to pass it and let the next operator figure out what it needs (i.e. stateless vs state-full). Cool! thanks for the explanation @prateekm !

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
prateekm: I'll leave this as-is to avoid future surprises. Let me know if you prefer using remove.

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -101,13 +108,16 @@ public void sink(SinkFunction<M> sinkFn) {
     return wndOp.getNextStream();
   }
 
-  @Override public <K, OM, RM> MessageStream<RM> join(MessageStream<OM> otherStream, JoinFunction<K, M, OM, RM> joinFn) {
+  @Override
+  public <K, JM, RM> MessageStream<RM> join(MessageStream<JM> otherStream, JoinFunction<K, M, JM, RM> joinFn, Duration ttl) {
     MessageStreamImpl<RM> outputStream = new MessageStreamImpl<>(this.graph);
 
-    PartialJoinFunction<K, M, OM, RM> parJoin1 = new PartialJoinFunction<K, M, OM, RM>() {
+    PartialJoinFunction<K, M, JM, RM> thisPartialJoinFn = new PartialJoinFunction<K, M, JM, RM>() {
```
vjagadish1989: nit: `PartialJoinState` instead maybe? This seems to be a wrapper around the `JoinFunction` + a store

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
vjagadish1989: I guess you could directly do:

`store.remove(entry.getKey());`

And get rid of the `removeAll`. However, It could be argued that some stores may choose to optimize bulk deletes differently. So, no preference on this one.

vjagadish1989 (approved):
looks good, thanks! approved.

samza-core/src/main/java/org/apache/samza/operators/functions/PartialJoinFunction.java
```diff
@@ -18,39 +18,52 @@
...
-   * @param message  the input message from the other stream
-   * @return  the join key in the {@code message}
+   * @return the key value store containing the state for this stream
    */
-  K getOtherKey(OM message);
+  KeyValueStore<K, PartialJoinMessage<M>> getState();
+
+  class PartialJoinMessage<M> {
+    private final M message;
+    private final long timestamp;
```
nickpan47: nit: insertion time or expiration time? It would be nice to document it.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorGraph.java
```diff
@@ -68,14 +68,22 @@ public void init(Map<SystemStream, MessageStreamImpl> inputStreams, Config confi
...
     return this.operatorGraph.get(ss);
   }
 
   /**
+   * Method to get the all {@link RootOperatorImpl}s
+   *
+   * @return  all {@link RootOperatorImpl}s for the graph
+   */
+  public Collection<RootOperatorImpl> getAll() {
+    return this.operatorGraph.values();
```
nickpan47: nit: would be nice to make this return set immutable.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -64,4 +77,16 @@ void propagateResult(RM outputMessage, MessageCollector collector, TaskCoordinat
...
 
+  /**
+   * Helper method to propagate the timer tick to all registered downstream operators.
+   *
+   * This method <b>must</b> be called from {@link #onTimer} to propagate the timer tick correctly.
+   *
+   * @param collector  the {@link MessageCollector} in the context
+   * @param coordinator  the {@link TaskCoordinator} in the context
+   */
+  void propagateTimer(MessageCollector collector, TaskCoordinator coordinator) {
```
nickpan47: Same suggestion as to InternalInMemoryStore, Duration, etc.

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
```diff
@@ -20,28 +20,70 @@
...
+    if (otherMessage != null) {
+      RM joinResult = thisPartialJoinFn.apply(message, otherMessage.getMessage());
+      this.propagateResult(joinResult, collector, coordinator);
+    }
+  }
+
+  @Override
+  public void onTimer(MessageCollector collector, TaskCoordinator coordinator) {
+    KeyValueStore<K, PartialJoinMessage<M>> thisState = thisPartialJoinFn.getState();
+    KeyValueIterator<K, PartialJoinMessage<M>> iterator = thisState.all();
```
nickpan47: It would also be a nice practice to:
1) bound the time to spend on this GC work to avoid long pause
2) close the iterator after the batch delete (since RocksDB has poor perf when iterators are left open)

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
```diff
@@ -20,28 +20,70 @@
...
   }
 
   @Override
   public void onNext(M message, MessageCollector collector, TaskCoordinator coordinator) {
-    // TODO: implement PartialJoinOperatorImpl processing logic
+    K key = thisPartialJoinFn.getKey(message);
+    thisPartialJoinFn.getState().put(key, new PartialJoinMessage<>(message, System.currentTimeMillis()));
+    PartialJoinMessage<JM> otherMessage = otherPartialJoinFn.getState().get(key);
+    if (otherMessage != null) {
+      RM joinResult = thisPartialJoinFn.apply(message, otherMessage.getMessage());
```
nickpan47: Since we are doing a lazy removal of buffered messages, we should skip old expired messages in the join result here as well.

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
prateekm: Yup. Btw, RocksDbIterator doesn't support remove either.

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -33,11 +28,20 @@
 import org.apache.samza.operators.functions.SinkFunction;
 import org.apache.samza.operators.spec.OperatorSpec;
 import org.apache.samza.operators.spec.OperatorSpecs;
+import org.apache.samza.operators.util.InternalInMemoryStore;
```
prateekm: It's ok, I can wait for Jagadish's PR - his is more up to date. There are a couple of other changes (like onTimer) that we'll need to merge.

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
prateekm: Minor preference for PartialJoinFunction as well. If we change, we should probably call it PartialJoinManager instead of PartialJoinState.

samza-core/src/main/java/org/apache/samza/operators/functions/PartialJoinFunction.java
prateekm: Good point, will update.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorGraph.java
prateekm: Yup, will do.

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
prateekm: Isn't ttl-based expiry expected to be best-effort by users anyway?

prateekm: On second thought it doesn't hurt to make it more precise, so let me do that.

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
prateekm: 1. Can we postpone this until we support persistent state stores?
2. Good catch, will do.

nickpan47: @prateekm sure, we can postpone this change. But I would strongly prefer to have some metrics/indicator to show the time spent in onTimer(). Main concern here is that since the state is fully controlled by us, user won't see it and when the container gets really slow, we need a fast way to detect the issue if the latency is too long in onTimer(). However, per operator metrics is also a big scope of work. I think that I am OK for now to just add two log lines at the beginning and ending of the onTimer() to report the time latency. We definitely need to design the per operator metrics soon.

prateekm: Makes sense, will add a debug log with time spent in onTimer.

samza-core/src/main/java/org/apache/samza/operators/functions/PartialJoinFunction.java
```diff
@@ -51,19 +51,19 @@
 
   class PartialJoinMessage<M> {
     private final M message;
-    private final long timestamp;
+    private final long receivedAt;
```
nickpan47: minor: I slightly prefer the name as noun. e.g. receiveTimeNano.

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
```diff
@@ -59,30 +63,34 @@ public void onNext(M message, MessageCollector collector, TaskCoordinator coordi
...
+
     KeyValueStore<K, PartialJoinMessage<M>> thisState = thisPartialJoinFn.getState();
     KeyValueIterator<K, PartialJoinMessage<M>> iterator = thisState.all();
     List<K> keysToRemove = new ArrayList<>();
-    long now = System.currentTimeMillis();
 
     while (iterator.hasNext()) {
       Entry<K, PartialJoinMessage<M>> entry = iterator.next();
-      if (entry.getValue().getTimestamp() < now - ttlMs) {
+      if (entry.getValue().getReceivedAt() < now - ttlMs) {
```
nickpan47: Also, noticed that getReceivedAt() does not carry the unit of timestamp in the variable/method name. I would prefer to add Ms or Nano to the variable/method name s.t. it becomes explicit.

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
```diff
@@ -59,30 +63,34 @@ public void onNext(M message, MessageCollector collector, TaskCoordinator coordi
...
         keysToRemove.add(entry.getKey());
       } else {
         break;
       }
     }
 
+    iterator.close();
     thisState.deleteAll(keysToRemove);
 
+    LOGGER.debug("onTimer self time: {} ms", System.currentTimeMillis() - now);
```
nickpan47: How often do we trigger onTimer()? I don't think that it would be often, right? I would suggest keeping this at info level until we have metrics reporting. One more suggestion: let's print the operator name and ID here as well.

samza-core/src/main/java/org/apache/samza/operators/functions/PartialJoinFunction.java
prateekm: Will updated to receivedTimeMs

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
prateekm: I'm guessing every second or so, since window controls the minimum granularity for onTimer across all operators (e.g. firing window triggers). Will change to info, we can turn it down later. Name should be present in the Logger class, will add operator ID.

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
prateekm: Usually `timestamp` implies linux epoch time ms (as opposed to duration where it's more confusing), but doesn't hurt to make it explicit, will update.

vjagadish1989:
@prateekm : There was a minor import issue in one of the tests in the patch that I fixed in a subsequent patch. (Since it was minor, I submitted it myself without your review). Hopefully that's okay with you.

prateekm:
@vjagadish1989 My bad, should have verified the merge. Thanks for fixing!

PR #61, opened 2017-02-24
SAMZA-1103: ZkBarrier

sborya:
SAMZA-1103: Barrier for JobModel upgrades. When all the processors got notification about the new JobModel, only after that they can start using the new model.


samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,28 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.zk;
+
+import java.util.List;
+
+
+public interface BarrierForVersionUpgrade {
```
sborya: done.

samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,46 @@
...
+ * Interface for a barrier - to allow synchronization between different processors to switch to a newly published
+ * JobModel.
+ */
+public interface BarrierForVersionUpgrade {
+  /**
+   * Barrier is started by the leader.
+   * @param version - for which the barrier is started.
+   * @param processorsNames - list of processors available at the time of the JobModel generation.
+   */
+  void leaderStartBarrier(String version,  List<String> processorsNames);
```
fredji97: The name of this method is a little weird for me. I would call it startBarrier. If it can only be run by leader, maybe we should put this function in a class Leader belongs to and trigger it with leader.startBarrier().

sborya: fixed.

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,148 @@
...
+    processors.add("p1");
+    processors.add("p2");
+
+    class Status {
+      boolean p1 = false;
+      boolean p2 = false;
+    }
+    final Status s = new Status();
+
+    barrier.leaderStartBarrier(ver, processors);
```
fredji97: this name is not changed yet.

samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,46 @@
...
+ * Interface for a barrier - to allow synchronization between different processors to switch to a newly published
+ * JobModel.
+ */
+public interface BarrierForVersionUpgrade {
+  /**
+   * Barrier is usually started by the leader.
+   * @param version - for which the barrier is started.
+   * @param processorsNames - list of processors available at the time of the JobModel generation.
+   */
+  void startBarrier(String version,  List<String> processorsNames);
```
fredji97: since it seems it is called as barrier.startBarrier, would it be cleaner to just use start() as function name and call it like barrier.start()?

samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Called by the processor.
+   * Updates the processor readiness to use the new version and wait on the barrier, until all other processors
+   * joined.
+   * @param version of the jobModel this barrier is protecting.
+   * @param processorsName as it appears in the list of processors.
+   * @param callback  will be invoked, when barrier is reached.
+   */
+  void waitForBarrier(String version, String processorsName, Runnable callback);
```
fredji97: similar here, would it be simpler to do barrier.wait() instead of barrier.waitForBarrier()?

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,148 @@
...
+
+    barrier.waitForBarrier(ver, "p1", new Runnable() {
+      @Override
+      public void run() {
+        s.p1 = true;
+      }
+    });
+
+    barrier.waitForBarrier(ver, "p2", new Runnable() {
```
fredji97: Need update here.

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,148 @@
...
+    class Status {
+      boolean p1 = false;
+      boolean p2 = false;
+      boolean p3 = false;
+    }
+    final Status s = new Status();
+
+    barrier.leaderStartBarrier(ver, processors);
+
+    barrier.waitForBarrier(ver, "p1", new Runnable() {
```
fredji97: need update here.

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,148 @@
...
+    processors.add("p3");
+
+    class Status {
+      boolean p1 = false;
+      boolean p2 = false;
+      boolean p3 = false;
+    }
+    final Status s = new Status();
+
+    barrier.leaderStartBarrier(ver, processors);
```
fredji97: need update here.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,168 @@
...
+
+public class ZkBarrierForVersionUpgrade implements BarrierForVersionUpgrade {
+  private final ZkUtils zkUtils;
+  private final ZkKeyBuilder keyBuilder;
+  private final static String BARRIER_DONE = "done";
+  private final static Logger LOG = LoggerFactory.getLogger(ZkBarrierForVersionUpgrade.class);
+
+  private final ScheduleAfterDebounceTime debounceTimer;
+
+  final private String barrierPrefix;
```
navina: nit: can you change this to "private final" in order to maintain coding style?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,168 @@
...
+    @Override
+    public void handleDataChange(String dataPath, Object data)
+        throws Exception {
+      String done = (String) data;
+      LOG.info("got notification about barrier path=" + barrierPathDone + "; done=" + done);
+      if (done.equals(BARRIER_DONE)) {
+        zkUtils.unsubscribeDataChanges(barrierPathDone, this);
+        debounceTimer.scheduleAfterDebounceTime(ScheduleAfterDebounceTime.JOB_MODEL_VERSION_CHANGE, 0, callback);
+      } else {
+        // TODO do we need to resubscribe?
```
navina: Let's remove this TODO and add the ZKClient behavior here.

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,148 @@
...
+
+    class Status {
+      boolean p1 = false;
+      boolean p2 = false;
+    }
+    final Status s = new Status();
+
+    barrier.startBarrier(ver, processors);
+
+    barrier.waitForBarrier(ver, "p1", new Runnable() {
```
navina: Tests look pretty simple.. Nice! 👍

samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
sborya: I think it is clearer if we use startBarrier (or createBarrier). It is more reflective of the barrier semantics.

samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
sborya: same as above.

samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
navina: Ideally speaking, what we have implemented is some kind of computation latch that only synchronizes the beginning of computation. I would actually expect the functions in barrier to be "create" (done by the leader, which simply creates the znode), "enter" (every processor joins the barrier znode) and "leave" (every processor leaves the barrier znode after computation is complete. See [example](https://zookeeper.apache.org/doc/r3.3.6/zookeeperTutorial.html#sc_barriers)

Having said that, I agree with Fred's suggestion about removing the word "barrier" from the function name. It seems redundant as the interface already reads BarrierForVersionUpgrade. It will also be nice to use "start" instead of "create".

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,166 @@
...
+    // if participant makes this call it means it has already stopped the old container and got the new job model.
+    String barrierPath = String.format("%s/barrier_%s", barrierPrefix, version);
+    String barrierDonePath = String.format("%s/barrier_done", barrierPath);
+    String barrierProcessors = String.format("%s/barrier_processors", barrierPath);
+    String barrierProcessorThis = String.format("%s/%s", barrierProcessors, processorsName);
+
+
+    // update the barrier for this processor
+    LOG.info("Creating a child for barrier at " + barrierProcessorThis);
+    zkUtils.getZkClient().createPersistent(barrierProcessorThis);
```
navina: Why should we create a persistent node to enter a barrier? If we create an ephemeral one, the node will automatically clean-up when it times-out/fails. This doesn't affect the latch since the barrier is deemed to fail when a processor leaves the "processors" group itself. Otherwise, we are at the risk of ending up with too many zknodes with each version upgrade.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,166 @@
...
+      LOG.info("list of children in the barrier = " + parentPath + ":" + sb.toString());
+      sb = new StringBuilder();
+      for (String child : names) {
+        sb.append(child).append(",");
+      }
+      LOG.info("list of children to compare against = " + parentPath + ":" + sb.toString());
+
+
+      // check if all the names are in
+      for (String n : names) {
```
navina: I think we definitely discussed this before. Please remind me and perhaps, add some documentation. Do we have to compare the names of the registered nodes necessarily? Isn't the size sufficient?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,166 @@
...
+    @Override
+    public void handleChildChange(String parentPath, List<String> currentChildren) throws Exception {
+      // Find out the event & Log
+      boolean allIn = true;
+
+      if (currentChildren == null) {
+        LOG.info("Got handleChildChange with null currentChildren");
+        return;
+      }
+      // debug
```
navina: nit : comment is vague. Remove it

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,166 @@
...
+      StringBuilder sb = new StringBuilder();
+      for (String child : currentChildren) {
+        sb.append(child).append(",");
+      }
+      LOG.info("list of children in the barrier = " + parentPath + ":" + sb.toString());
```
navina: nit: should be LOG.debug ?? Same for Line 118

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,166 @@
...
+      for (String n : names) {
+        if (!currentChildren.contains(n)) {
+          LOG.info("node " + n + " is still not in the list ");
+          allIn = false;
+          break;
+        }
+      }
+      if (allIn) {
+        LOG.info("ALl nodes reached the barrier");
+        callback.run(); // all the names have registered
```
navina: Why not simply call zkClient.writeData() here instead of using a runnable? It's really not a callback either. is it for the sake of testing ?

xinyuiscool: Do we need a timeout/retry for starting the barrier? I mean, is it possible that the processors never reach the initial set?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,166 @@
...
+    // now subscribe for the barrier
+    zkUtils.getZkClient().subscribeDataChanges(barrierDonePath, new ZkBarrierReachedHandler(barrierDonePath, debounceTimer, callback));
+  }
+
+  /**
+   * listener for the subscription.
+   */
+  class ZkBarrierChangeHandler implements IZkChildListener {
+    Runnable callback;
+    List<String> names;
```
xinyuiscool: final

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -0,0 +1,166 @@
...
+
+    // now subscribe for the barrier
+    zkUtils.getZkClient().subscribeDataChanges(barrierDonePath, new ZkBarrierReachedHandler(barrierDonePath, debounceTimer, callback));
+  }
+
+  /**
+   * listener for the subscription.
+   */
+  class ZkBarrierChangeHandler implements IZkChildListener {
+    Runnable callback;
```
xinyuiscool: final

samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
sborya: For me 'startBarrier' sounds more specific and clear, but if both of you feel it should be just 'start' I can change it.
But I cannot change the interface, because the waitForBarrier must be async (all the actions are invoked FROM the ZK), so we cannot tie up the thread.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: done.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: done. Also made it private.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: for now we should do it for history and debugability, we can change it later if needed.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: fixed.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: for now it is info. May be changed to debug later.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: Extra precautions, this is more precise. There might be some runaway/rogue processor joining in theory.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: I think you are right! I will change it in the next patch.

sborya: I will mull it over and create a separate patch for it if needed.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
navina: Is it actually possible for a runaway / rogue processor to enter the barrier without registering in the processor group? I don't see a possibility there.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
navina: Even the short-term, I feel it is too much data on ZK to store for the purpose of debugging. The leader is watching the barrier. So, the leader log should have sufficient data to troubleshoot.

navina: If we want to keep it as a temporary code, we should at least add a TODO to describe how the behavior can be changed and when that change should be implemented.

navina:
@sborya are you making these newer changes in the same branch ? or in the follow-up patch? 
This PR says "Closed with unmerged commits"

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: I will added it in a separate PR.

PR #64, opened 2017-02-24
SAMZA-1106:Zk test utils

sborya:
Some test utils needed for StandAlone zk based testing.


navina:
@sborya I am not clear on what this JIRA / PR is for? You are adding JavaJobConfig. But where is ZkTestUtils being used? Am I reviewing the PRs in the wrong order?

samza-test/src/main/java/org/apache/samza/system/standalone/ZkTestUtils.java
```diff
@@ -0,0 +1,88 @@
+package org.apache.samza.system.standalone;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.samza.config.ZkConfig;
+
+
+public class ZkTestUtils {
+  private static final String ZK_JOB_COORDINATOR_FACTORY = "org.apache.samza.zk.ZkJobCoordinatorFactory";
+  private static final String KAFKA_SYSTEM_FACTORY = "org.apache.samza.system.kafka.KafkaSystemFactory";
```
navina: Might as well use KafkaSystemFactory.class.getName();

samza-test/src/main/java/org/apache/samza/system/standalone/ZkTestUtils.java
```diff
@@ -0,0 +1,88 @@
+package org.apache.samza.system.standalone;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.samza.config.ZkConfig;
+
+
+public class ZkTestUtils {
+  private static final String ZK_JOB_COORDINATOR_FACTORY = "org.apache.samza.zk.ZkJobCoordinatorFactory";
```
navina: May be there is someother PR that needs to be reviewed before this one because there is no ZkJobCoordinatorFactory defined in samza master

PR #65, opened 2017-02-26
SAMZA-1093: instantiating StreamOperatorTask in SamzaContainer (WIP)

nickpan47:
SAMZA-1093: Instantiating StreamOperatorTask in SamzaContainer

- minor refactor to get the TaskFactory based on task class configuration
- add instantiation of StreamOperatorTask in SamzaContainer


samza-core/src/main/java/org/apache/samza/task/TaskFactories.java
```diff
@@ -0,0 +1,95 @@
...
+    boolean isAsyncTaskClass = AsyncStreamTask.class.isAssignableFrom(Class.forName(taskClassName));
+    if (isAsyncTaskClass) {
+      return new AsyncStreamTaskFactory() {
+        @Override
+        public AsyncStreamTask createInstance() {
+          try {
+            return (AsyncStreamTask) Class.forName(taskClassName).newInstance();
+          } catch (Exception e) {
+            log.error("Error loading AsyncStreamTask class: {}. error: {}", taskClassName, e);
+            throw new RuntimeException(e);
```
vjagadish1989: nits: 

throw `SamzaException` for consistency with the rest of samza-core? Also, SamzaException extends RuntimeException anyways.

Should we catch `Throwable` here instead of `Exception`? For instance, `newInstance` can throw an `ExceptionInInitializerError` when static initialization fails?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -25,6 +25,8 @@ import java.util
 import java.util.concurrent.{CountDownLatch, ExecutorService, Executors, TimeUnit}
 import java.lang.Thread.UncaughtExceptionHandler
 import java.net.{URL, UnknownHostException}
+import javax.servlet.SingleThreadModel
```
vjagadish1989: nit: probably not used?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -64,13 +66,7 @@ import org.apache.samza.system.SystemStreamPartition
...
 import org.apache.samza.system.chooser.MessageChooserFactory
 import org.apache.samza.system.chooser.RoundRobinChooserFactory
-import org.apache.samza.task.AsyncRunLoop
-import org.apache.samza.task.AsyncStreamTask
-import org.apache.samza.task.AsyncStreamTaskAdapter
-import org.apache.samza.task.AsyncStreamTaskFactory
-import org.apache.samza.task.StreamTask
-import org.apache.samza.task.StreamTaskFactory
-import org.apache.samza.task.TaskInstanceCollector
+import org.apache.samza.task._
```
vjagadish1989: nit: blanket imports probably not needed?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -462,6 +436,39 @@ object SamzaContainer extends Logging {
...
+      if (isAsyncTaskClass) {
+        info("Got an AsyncStreamTask implementation.")
+      }
+
+      if (singleThreadMode && isAsyncTaskClass) {
+        throw new SamzaException("AsyncStreamTask cannot run on single thread mode.")
+      }
+
+      if (!singleThreadMode && !isAsyncTaskClass) {
+        info("Converting StreamTask to AsyncStreamTaskAdapter when running StreamTask w/ multiple threads")
```
vjagadish1989: nit: s/w//with

samza-core/src/main/java/org/apache/samza/task/TaskFactories.java
```diff
@@ -0,0 +1,95 @@
...
+
+public class TaskFactories {
+  private static final Logger log = LoggerFactory.getLogger(TaskFactories.class);
+
+  public static Object fromTaskClassConfig(Config config) throws ClassNotFoundException {
+
+    String taskClassName;
+
+    // if there is configuration to set the job w/ a specific type of task, instantiate the corresponding task factory
+    if (isStreamOperatorTask(config)) {
```
vjagadish1989: Should we throw an exception if a user implements `GraphBuilder` along with a  low level task API like StreamTask ?

vjagadish1989 (approved):
looks good to me. approved. thank you for the contributions

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -483,24 +490,14 @@ object SamzaContainer extends Logging {
 
       val taskName = taskModel.getTaskName
 
-      val taskObj = if (taskFactory != null) {
-        debug("Using task factory to create task instance")
-        taskFactory match {
+      debug("Using task factory to create task instance")
+      val task =
+        finalTaskFactory match {
```
navina: Do we still need this match statement since we are already validating taskFactory instance in TaskFactories?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -462,6 +436,39 @@ object SamzaContainer extends Logging {
...
+        info("Got an AsyncStreamTask implementation.")
+      }
+
+      if (singleThreadMode && isAsyncTaskClass) {
+        throw new SamzaException("AsyncStreamTask cannot run on single thread mode.")
+      }
+
+      if (!singleThreadMode && !isAsyncTaskClass) {
+        info("Converting StreamTask to AsyncStreamTaskAdapter when running StreamTask w/ multiple threads")
+        new AsyncStreamTaskFactory {
```
navina: Would it be better to encapsulate this logic as well in TaskFactories?

samza-api/src/main/java/org/apache/samza/operators/StreamGraphBuilder.java
```diff
@@ -27,6 +27,8 @@
  */
 @InterfaceStability.Unstable
 public interface StreamGraphBuilder {
+  static final String BUILDER_CLASS_CONFIG = "job.graph.builder.class";
```
navina: Can you please add some javadoc for what this config is for? We can later add the same to the config table.

I am also curious why this is called "job.graph.builder.class". Ideally, I would expect a graph to be composed of one or more jobs. But making it as a sub-namespace of job implies that a job itself contains a graph. This is based on my understanding of what we mean by a "job" today, in Samza. A job, today, indicates a single stage of processing defined by the user. Are we going to change the meaning of what a "job" means? 

I think applying all the stream graph related configurations "above" the job abstraction will simplify our configuration without impacting the notion of a job. My 2 cents :)

samza-core/src/main/java/org/apache/samza/task/TaskFactories.java
```diff
@@ -0,0 +1,95 @@
...
+          }
+        }
+      };
+    }
+
+    return new StreamTaskFactory() {
+      @Override
+      public StreamTask createInstance() {
+        try {
+          return taskClassName == StreamOperatorTask.class.getName() ? createStreamOperatorTask(config) :
```
cpettitt-linkedin: Use .equals here.

samza-core/src/main/java/org/apache/samza/task/TaskFactories.java
```diff
@@ -0,0 +1,95 @@
...
+        } catch (Exception e) {
+          log.error("Error loading StreamTask class: {}. error: {}", taskClassName, e);
+          throw new RuntimeException(e);
+        }
+      }
+    };
+  }
+
+  private static StreamTask createStreamOperatorTask(Config config) throws Exception {
+    StreamGraphBuilder graphBuilder = (StreamGraphBuilder) Class.forName(config.get(StreamGraphBuilder.BUILDER_CLASS_CONFIG)).newInstance();
```
cpettitt-linkedin: It will not always be obvious what part of this statement failed when looking at a stack trace because you're doing a couple of reflection calls in the same statement. I'd suggest splitting these up for better debuggability.

samza-core/src/main/java/org/apache/samza/task/TaskFactories.java
```diff
@@ -0,0 +1,95 @@
...
+    return new StreamOperatorTask(graphBuilder);
+  }
+
+  private static boolean isStreamOperatorTask(Config config) {
+    try {
+      if (config.get(StreamGraphBuilder.BUILDER_CLASS_CONFIG) != null && config.get(StreamGraphBuilder.BUILDER_CLASS_CONFIG) != "") {
```
cpettitt-linkedin: Use .equals.

Also, the try-catch can be moved inside the if block.

samza-core/src/main/java/org/apache/samza/task/TaskFactories.java
nickpan47: Good point. Will add more validation on this one.

samza-core/src/main/java/org/apache/samza/task/TaskFactories.java
nickpan47: 👍 will fix both comments.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
nickpan47: IDE auto-import... Good catch.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
nickpan47: Yeah, IDE acting up on this one again. Will fix.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
nickpan47: Thanks! Will do!

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
nickpan47: I tried. The difficulty is mainly in the taskThreadPool: it is instantiated within SamzaContainer context and optionally used in this case. The same taskThreadPool is needed outside the TaskFactories as well, which makes it really messy. There is an option to passing all these parameters to TaskFactories.fromConfig(). Let me make one more try to see whether the code looks better.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
nickpan47: TaskFactories only returns an Object, we will still need to do if {} else {} based on whether it is an instance of AsyncStreamTaskFactory or StreamTaskFactory. The code would end up the same.

samza-api/src/main/java/org/apache/samza/operators/StreamGraphBuilder.java
nickpan47: Good points. I am not sure what's the new scope that we are creating for pipeline. I can probably use that. @jmakes and Xinyu, any suggestions?

samza-core/src/main/java/org/apache/samza/task/TaskFactories.java
nickpan47: Good catch! I will fix this. Had already hit one of the bug some other places. Thanks!

samza-core/src/main/java/org/apache/samza/task/TaskFactories.java
nickpan47: 👍

samza-core/src/main/java/org/apache/samza/task/TaskFactories.java
nickpan47: 👍

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
navina: Got it. I wonder why didn't use a marker interface or typed class for this scenario. It is making all interactions with taskfactory very messy.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
navina: Awesome.. Thanks for trying it out ! 👍

nickpan47: Forgot another factor: in standalone, when the taskFactory is an object that passed in, "fromConfig" actually does not make sense. Yet, the "maybeAsyncStreamTaskFactory" still applies to the passed in taskFactory as well. Hence, creating weird case that TaskFactories method would take an already instantiated taskFactory as the input parameter to generate the final task factory. Will try one more time to see whether that would clean up better.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
nickpan47: Yeah, maybe a generic type TaskFactory with the task type parameter is better? Something like: 
{code}
interface TaskFactory<T> {
  T createInstance();
}
{code}

That sounds good to me. Thanks for the suggestion!

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -462,6 +436,39 @@ object SamzaContainer extends Logging {
     else
       null
 
+    def maybeAsyncStreamTaskFactory : Object = {
```
nickpan47: 👍

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -462,6 +436,39 @@ object SamzaContainer extends Logging {
...
+
+      val taskClassFactory = taskFactory match {
```
nickpan47: I don't see much of the difference here. Anyways, it is scala coding style here and I don't see one way is better than the other. 
Also I have refactored the code a little bit more and seems match works better. Feel free to take a look at the revision and let me know if you still feel strong about this.

samza-core/src/main/java/org/apache/samza/task/TaskFactories.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    return new StreamTaskFactory() {
+      @Override
+      public StreamTask createInstance() {
+        try {
+          return taskClassName == StreamOperatorTask.class.getName() ? createStreamOperatorTask(config) :
+              (StreamTask) Class.forName(taskClassName).newInstance();
+        } catch (Exception e) {
+          log.error("Error loading StreamTask class: {}. error: {}", taskClassName, e);
```
nickpan47: Not quite follow with this comment. Are you referring to exception messaging or logging? In logging, shouldn't we just form a string?

samza-core/src/main/java/org/apache/samza/task/TaskFactories.java
```diff
@@ -0,0 +1,95 @@
...
+
+    // if there is configuration to set the job w/ a specific type of task, instantiate the corresponding task factory
+    if (isStreamOperatorTask(config)) {
+      taskClassName = StreamOperatorTask.class.getName();
+    } else {
+      taskClassName = new TaskConfig(config).getTaskClass().getOrElse(defaultValue(null));
+    }
+
+    if (taskClassName == null) {
+      throw new ConfigException("There is no task class defined in the configuration. Failed to create a valid TaskFactory");
```
nickpan47: Sure. It ends up w/ a bit more lines due to convertion from Scala Function0, but works.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -483,24 +470,8 @@ object SamzaContainer extends Logging {
...
-        debug("Using task class name: %s to create instance" format taskClassName)
-        Class.forName(taskClassName).newInstance
-      }
-
-      val task = if (!singleThreadMode && !isAsyncTask)
-        // Wrap the StreamTask into a AsyncStreamTask with the build-in thread pool
-        new AsyncStreamTaskAdapter(taskObj.asInstanceOf[StreamTask], taskThreadPool)
-      else
-        taskObj
+      debug("Using task factory to create task instance")
```
prateekm: Can remove this message since it's always true now.

samza-api/src/main/java/org/apache/samza/application/StreamApplication.java
```diff
@@ -16,17 +16,20 @@
...
+import org.apache.samza.operators.StreamGraph;
 
 
 /**
  * This interface defines a factory class that user will implement to create user-defined operator DAG in a {@link StreamGraph} object.
  */
 @InterfaceStability.Unstable
-public interface StreamGraphBuilder {
+public interface StreamApplication {
+  static final String APP_CLASS_CONFIG = "app.class";
```
prateekm: Minor: This should probably not be here. I think so far we've kept it in config classes. If there isn't one, maybe keep in TaskFactoryUtils?

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
```diff
@@ -105,7 +106,9 @@ public void startContainer(ContainerModel containerModel, Config config, int max
         localityManager,
         new JmxServer(),
         Util.<String, MetricsReporter>javaMapAsScalaMap(metricsReporterMap),
-        taskFactory);
+        taskFactory,
+        // TODO: need to use the correct local ApplicationRunner here
+        ApplicationRunner.getLocalRunner(config));
```
prateekm: Shouldn't this be getRemoteRunner? If I understand correctly, local runner is meant for standalone.

samza-api/src/main/java/org/apache/samza/application/StreamApplication.java
nickpan47: I have another patch working on adding AppConfig to wrap those configure variables as in other configs. Neither here nor the TaskFactoryUtils is the right place. I simply placed it here to be closer to its own usage, for the short term. Planning to move it to AppConfig.

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
```diff
@@ -0,0 +1,156 @@
...
+
+    if (!(factory instanceof StreamTaskFactory) && !(factory instanceof AsyncStreamTaskFactory)) {
+      throw new SamzaException(String.format("TaskFactory must be either StreamTaskFactory or AsyncStreamTaskFactory. %s is not supported",
+          factory.getClass()));
+    }
+  }
+
+  private static StreamTask createStreamOperatorTask(Config config, ApplicationRunner runner) throws Exception {
+    Class<?> builderClass = Class.forName(config.get(StreamApplication.APP_CLASS_CONFIG));
+    StreamApplication graphBuilder = (StreamApplication) builderClass.newInstance();
```
nickpan47: This would be the same for a lot of classes we load. We would either need to change them all, or just follow the current pattern. Besides, the constructor w/o input parameter seems to be OK to me.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -462,6 +441,14 @@ object SamzaContainer extends Logging {
     else
       null
 
+    val finalTaskFactory = TaskFactoryUtil.finalizeTaskFactory(
+      taskFactory match {
+        case null => TaskFactoryUtil.fromTaskClassConfig(config, appRunner)
+        case _ => taskFactory.asInstanceOf[TaskFactory[_]]
```
nickpan47: Discussed offline. Will remove since it does not completely solve the type issue when we classload this factory.

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
nickpan47: Actually, this should be neither the local runner nor remote runner. It needs to be ContainerLocalRunner, which is just play a role of stream provider as we discussed today. Let me set it to null for now.

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
```diff
@@ -105,7 +105,9 @@ public void startContainer(ContainerModel containerModel, Config config, int max
         localityManager,
         new JmxServer(),
         Util.<String, MetricsReporter>javaMapAsScalaMap(metricsReporterMap),
-        taskFactory);
+        taskFactory,
+        // TODO: need to use the correct local ApplicationRunner here
+        null);
```
prateekm: Is this going to cause NPEs later right now? Maybe create a ticket to finish this wiring and add to the TODO?

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
```diff
@@ -0,0 +1,155 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.ExecutorService;
+
+import scala.runtime.AbstractFunction0;
+
+public class TaskFactoryUtil {
+  private static final Logger log = LoggerFactory.getLogger(TaskFactoryUtil.class);
+
+  public static Object fromTaskClassConfig(Config config, ApplicationRunner runner) {
```
xinyuiscool: Since this is public, please add java doc to it.

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
```diff
@@ -0,0 +1,155 @@
...
+          return (StreamTask) Class.forName(taskClassName).newInstance();
+        } catch (Throwable t) {
+          log.error("Error loading StreamTask class: {}. error: {}", taskClassName, t);
+          throw new SamzaException(String.format("Error loading StreamTask class: %s", taskClassName), t);
+        }
+      }
+    };
+  }
+
+  public static Object finalizeTaskFactory(Object factory, boolean singleThreadMode, ExecutorService taskThreadPool) {
```
xinyuiscool: Please add javadoc to this.

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
nickpan47: Ticket was already opened: SAMZA-1137

nickpan47:
Jenkins failure was due to fail to checkout master. Local build has been successful

PR #66, opened 2017-02-26
SAMZA-1108: Implementation of Windows and various kinds of Triggers (WIP)

vjagadish1989:
* Implemented various triggers and the orchestration logic of the window operator.
* Implemented wire-up of window and the flow of messages through various trigger implementations.
* Implementations for count, time, timeSinceFirst, timeSinceLast, Any, Repeating triggers.

Pending:
- More tests.


samza-core/src/main/java/org/apache/samza/operators/util/InternalInMemoryStore.java
```diff
@@ -0,0 +1,124 @@
...
+ * Implements a {@link KeyValueStore} using an in-memory Java Map.
+ * @param <K>
+ * @param <V>
+ *
+ * TODO: Provide a persistent implementation of the KeyValueStore API. Maybe, we don't need this class.
+ *
+ */
+public class InternalInMemoryStore<K, V> implements KeyValueStore<K, V> {
+
+  final Map<K, V> map = new LinkedHashMap<>();
```
cpettitt-linkedin: private?

samza-core/src/main/java/org/apache/samza/operators/triggers/TriggerImpls.java
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.samza.operators.data.MessageEnvelope;
+
+/**
+ * Factory methods for instantiating {@link TriggerImpl}s from individual {@link Trigger}s.
+ */
+public class TriggerImpls {
+
+  public static <M extends MessageEnvelope> TriggerImpl<M> createTriggerImpl(Trigger<M> trigger, TriggerContext context, TriggerImpl.TriggerCallbackHandler handler) {
+
+    if (trigger instanceof CountTrigger) {
```
cpettitt-linkedin: Maybe Trigger should know how to create its Impl? Certainly it looks like this method would be easy to miss when adding a new trigger type.

samza-core/src/main/java/org/apache/samza/operators/triggers/TriggerImpls.java
```diff
@@ -0,0 +1,46 @@
...
+    } else if (trigger instanceof AnyTrigger) {
+      return new AnyTriggerImpl<>((AnyTrigger) trigger, context, handler);
+    } else if (trigger instanceof TimeSinceLastMessageTrigger) {
+      return new TimeSinceLastMessageTriggerImpl<>((TimeSinceLastMessageTrigger) trigger, context, handler);
+    } else if (trigger instanceof TimeTrigger) {
+      return new TimeTriggerImpl((TimeTrigger) trigger, context, handler);
+    } else if (trigger instanceof TimeSinceFirstMessageTrigger) {
+      return new TimeSinceFirstMessageTriggerImpl<>((TimeSinceFirstMessageTrigger) trigger, context, handler);
+    }
+    return null;
```
cpettitt-linkedin: Should this throw? I thought I saw places earlier in the code that assumed that this would not return null.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,231 @@
...
+    }
+
+    WindowKey<K> windowKey = new WindowKey(key, windowId);
+    return windowKey;
+  }
+
+  private TriggerImpl.TriggerCallbackHandler createTriggerHandler(WindowKey<K> key, TriggerType type) {
+    TriggerImpl.TriggerCallbackHandler handler = new TriggerImpl.TriggerCallbackHandler() {
+      @Override
+      public void onTrigger(TriggerImpl impl, Object key) {
```
cpettitt-linkedin: If we need the most recent collector and coordinator can we make that explicit by taking them as arguments? That would make this much clearer and remove the possibility of not setting the recentCollector / recentCoordinator on a path where it is required.

samza-core/src/main/java/org/apache/samza/operators/triggers/RepeatingTriggerImpl.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private TriggerImpl underlyingTriggerImpl;
+
+  private boolean cancelled = false;
+
+  public RepeatingTriggerImpl(RepeatingTrigger<M> repeatingTrigger, TriggerContext tContext, TriggerCallbackHandler handler) {
+    super(tContext, handler);
+    this.underlyingTrigger = repeatingTrigger.getTrigger();
+
+    if (!cancelled) {
```
cpettitt-linkedin: Would this ever be false?

samza-core/src/main/java/org/apache/samza/operators/triggers/RepeatingTriggerImpl.java
```diff
@@ -0,0 +1,64 @@
...
+      this.underlyingTriggerImpl = TriggerImpls.createTriggerImpl(underlyingTrigger, context, createNewCallback());
+    }
+  }
+
+  private TriggerCallbackHandler createNewCallback() {
+    return new TriggerCallbackHandler() {
+      @Override
+      public void onTrigger(TriggerImpl impl, Object storeKey) {
+        underlyingTriggerImpl = TriggerImpls.createTriggerImpl(underlyingTrigger, context, createNewCallback());
```
cpettitt-linkedin: Do you need to check for cancellation here?

samza-core/src/main/java/org/apache/samza/operators/triggers/TriggerImpl.java
```diff
@@ -0,0 +1,64 @@
...
+ * {@link TriggerImpl}s can use the {@link TriggerContext} to schedule and cancel callbacks (for example, implementations
+ * of time-based triggers).
+ *
+ * <p> State management: The state maintained by {@link TriggerImpl}s is not durable across re-starts and is transient.
+ * New instances of {@link TriggerImpl} are created on a re-start.
+ *
+ */
+public abstract class TriggerImpl<M> {
+
+  protected final TriggerCallbackHandler handler;
```
cpettitt-linkedin: TriggerImpl (name probably needs some tweaking) could be an interface if it weren't for stashing handler and context here. However, neither are used in this context, so I would dump them.

If subclasses need the handler they should store it themselves. It would get rid of the super boilerplate and it would be much clearer from a code review perspective where the handler field comes from.

samza-core/src/main/java/org/apache/samza/operators/triggers/CountTriggerImpl.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Implementation class for a {@link CountTrigger}
+ */
+public class CountTriggerImpl<M extends MessageEnvelope> extends TriggerImpl<M> {
+
+  private final long triggerCount;
+  private long currentCount;
+  private boolean cancelled = false;
+
+  public CountTriggerImpl(CountTrigger<M> triggerCount, TriggerContext context, TriggerCallbackHandler handler) {
```
cpettitt-linkedin: I see a fair amount of boilerplate around propagating the context and handler. I made some comments on this below. I wonder if we couldn't just save these in one place and pass them into message instead? That would simplify each one of these TriggerImpl subclasses.

samza-core/src/main/java/org/apache/samza/operators/triggers/TimeSinceLastMessageTriggerImpl.java
```diff
@@ -0,0 +1,62 @@
...
+    Runnable runnable = () -> {
+      handler.onTrigger(this, context.getWindowKey());
+    };
+
+    latestFuture = context.scheduleCallback(runnable, callbackTime);
+  }
+
+  @Override
+  public void onCancel() {
+    latestFuture.cancel();
```
cpettitt-linkedin: Can onCancel be called before before onMessage? If so, then it seems we need a guard in several of these classes. If not, we should document this in TriggerImpl so it is explicit.

---

One other thought on cancellation: it seems that in all cases I've seen so far that cancellation involves not doing something in onMessage or cancelling a timer. I wonder if we couldn't simplify the duplication happening for these two cases by handling cancellation a level above this: if cancel is invoked then we no longer call onMessage and any scheduled callbacks (assuming there is one given my understanding of the design) would also be cancelled immediately.

This restricts what can be done in cancellation. We could still provide the onCancel hook if needed for other cleanup. Alternatively if this actually complicates the implementation we should certainly drop the idea.

navina:
@vjagadish1989 Can you please refer the JIRA for this PR?

samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
```diff
@@ -24,8 +24,11 @@
...
 import org.apache.samza.operators.triggers.Triggers;
 import org.apache.samza.operators.windows.internal.WindowInternal;
+import org.apache.samza.operators.windows.internal.WindowType;
 
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.function.BiFunction;
 import java.util.function.Function;
```
navina: This comment pertains to Line 82 (Github is not allowing me to add a comment there :( )

QQ (perhaps orthogonal to current patch): The javadoc in Line 82 - Does it "require" an early trigger to be defined by the user? Looks like there is no default behavior. If that is the intended case, how are we enforcing this on the user? Shouldn't Trigger be a parameter in globalWindow() ?

I could be wrong as I just started looking into the api code. Please correct if I am wrong. Thanks!

vjagadish1989: Good point. Here's the other side of your question -  If we take the "early trigger" as a parameter to the `globalwindow`, and if a user chooses to `setEarlyTrigger` on it, what the behavior should be? (we can choose to obviously clobber the configured earlyTrigger)

For the MVP, we agreed that `globalwindow` is not supported anyways. I'll remove it from the API. We can consider adding it later if there's a strong use-case.

samza-api/src/main/java/org/apache/samza/operators/windows/WindowPane.java
```diff
@@ -46,8 +46,8 @@ public V getMessage() {
     return this.key;
   }
 
-  static public <K, M> WindowPane<K, M> of(WindowKey<K> key, M result) {
-    return new WindowPane<>(key, result, AccumulationMode.DISCARDING);
+  static public <K, M> WindowPane<K, M> of(WindowKey<K> key, M result, AccumulationMode mode) {
```
vjagadish1989: Sure, will fix.

samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
navina: Exactly. That would have been my next question :) So let's remove it if there isn't sufficient clarity. Thanks!

samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
```diff
@@ -143,6 +146,9 @@ private Windows() { }
    */
   public static <M, K> Window<M, K, Collection<M>> keyedTumblingWindow(Function<M, K> keyFn, Duration interval) {
     BiFunction<M, Collection<M>, Collection<M>> aggregator = (m, c) -> {
+      if (c == null) {
+        return new ArrayList<>(Arrays.asList(m));
```
vjagadish1989: 1. +1
2. Don't we still need the branch anyways? (depending on whether `c` is null)
```
      if (c == null) {
        c = new ArrayList<>();
      }
      c.add(m);
      return c;

```
Do let me know if it can be re-written dropping the branch altogether.

cpettitt-linkedin:
@vjagadish1989: I'm working through the pull request locally as discussed. One observation: we're losing type information in WindowOperatorImpl and when I add the missing type information (e.g. to spec.getWindow()) the types don't match up (e.g. there is an assumption that K = WK that is not explicitly worked out in the type system). I would suggest taking a look.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -52,6 +52,26 @@ void registerNextOperator(OperatorImpl<RM, ?> nextOperator) {
...
+   * @param collector
+   * @param coordinator
+   */
+  public final void onTick(MessageCollector collector, TaskCoordinator coordinator) {
+    onTimer(collector, coordinator);
+    propagateTimer(collector, coordinator);
+  }
+
+  /**
+   * Invoked at every tick. Implementations may call {@link #propagateResult} if they wish to propagate the output to registered
```
cpettitt-linkedin: If `propagateResult` is correct here then we should update the documentation at line 77 which states that it should only be called from `onNext`.

samza-core/src/main/java/org/apache/samza/operators/triggers/TriggerImpls.java
```diff
@@ -0,0 +1,48 @@
...
+      return new AnyTriggerImpl<>((AnyTrigger<M>) trigger);
+    } else if (trigger instanceof TimeSinceLastMessageTrigger) {
+      return new TimeSinceLastMessageTriggerImpl<>((TimeSinceLastMessageTrigger<M>) trigger);
+    } else if (trigger instanceof TimeTrigger) {
+      return new TimeTriggerImpl((TimeTrigger<M>) trigger);
+    } else if (trigger instanceof TimeSinceFirstMessageTrigger) {
+      return new TimeSinceFirstMessageTriggerImpl<>((TimeSinceFirstMessageTrigger<M>) trigger);
+    }
+
+    throw new SamzaException("No implementation class defined. " + trigger);
```
cpettitt-linkedin: It might be helpful to actually log the trigger class here as well as that is how we're looking up the associated implementation.

samza-core/src/main/java/org/apache/samza/operators/util/InternalInMemoryStore.java
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Implements a {@link KeyValueStore} using an in-memory Java Map.
+ * @param <K>
+ * @param <V>
+ *
+ * TODO: Provide a persistent implementation of the KeyValueStore API. Maybe, we don't need this class.
+ *
+ */
+public class InternalInMemoryStore<K, V> implements KeyValueStore<K, V> {
```
cpettitt-linkedin: We need to be throwing NPE's for several of these functions per the documentation in KeyValueStore.

samza-core/src/test/java/org/apache/samza/operators/triggers/TestWindowOperator.java
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  @Test
+  public void testTumblingWindowsDiscardingMode() throws Exception {
+    StreamGraphBuilder sgb = new KeyedTumblingWindowStreamGraphBuilder(AccumulationMode.DISCARDING);
+    StreamOperatorTask task = new StreamOperatorTask(sgb);
+    task.init(config, taskContext);
+
+    integers.forEach(n -> task.process(new IntegerMessageEnvelope(n, n), messageCollector, taskCoordinator));
+    Thread.sleep(1000);
```
cpettitt-linkedin: It would be preferable to use a clock abstraction for testing (I believe Samza core already has one). This will keep the tests fast as we add more cases.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
vjagadish1989: Fixed. (with updated docs).

samza-core/src/main/java/org/apache/samza/operators/triggers/TriggerImpls.java
vjagadish1989: Fixed.

samza-core/src/test/java/org/apache/samza/operators/triggers/TestWindowOperator.java
vjagadish1989: Good suggestion, Makes sense! However, It does make the `Clock` bleed to in all of the `OperatorImpl` and the `TriggerImpl` classes. (all the way till `StreamOperatorTask`). But, that's totally acceptable given how fast / deterministic our tests run now :-). 

I'll update the PR with more tests, and with the change.

samza-core/src/main/java/org/apache/samza/operators/util/InternalInMemoryStore.java
vjagadish1989: Yup, will fix. Good find! FWIW, This store class is temporary (and will go away after we integrate persistence and `SerDe` implementations into operators)

samza-core/src/main/java/org/apache/samza/operators/triggers/AnyTriggerImpl.java
```diff
@@ -0,0 +1,60 @@
...
+    this.triggerList = anyTrigger.getTriggers();
+
+    for (Trigger<M> trigger : triggerList) {
+      triggerImpls.put(TriggerImpls.createTriggerImpl(trigger), false);
+    }
+  }
+
+  @Override
+  public void onMessage(M message, TriggerContext context, TriggerCallbackHandler handler) {
+    for (TriggerImpl<M> triggerImpl : triggerImpls.keySet()) {
```
nickpan47: I thought that AnyTriggerImpl means if any of the triggerImpl fired up, then we don't need to continue evaluating other triggers? So, if handler.isTriggered() after onMessage(), we should break out from the loop?

samza-core/src/main/java/org/apache/samza/operators/triggers/Cancellable.java
```diff
@@ -0,0 +1,34 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.operators.triggers;
+
+/**
+ * Represents a task or an operation whose execution can be cancelled.
+ */
+public interface Cancellable {
```
nickpan47: Why don't we use the standard Future interface?

samza-core/src/main/java/org/apache/samza/operators/triggers/RepeatingTriggerImpl.java
```diff
@@ -0,0 +1,58 @@
...
+          //re-schedule the underlying trigger for execution again.
+          underlyingTriggerImpl = TriggerImpls.createTriggerImpl(underlyingTrigger);
+          handler.onTrigger();
+      }
+    };
+  }
+
+  @Override
+  public void onMessage(M message, TriggerContext context, TriggerCallbackHandler handler) {
+    underlyingTriggerImpl.onMessage(message, context, createNewHandler(handler));
```
nickpan47: Question: why are we creating new handler for each incoming message? Wouldn't it create many TriggerImpls, one for each message? I think that it would make more sense to avoid creating new TriggerImpl until the current TriggerImpl is fired.

samza-core/src/main/java/org/apache/samza/operators/triggers/RepeatingTriggerImpl.java
```diff
@@ -0,0 +1,58 @@
...
+    this.underlyingTrigger = repeatingTrigger.getTrigger();
+    this.underlyingTriggerImpl = TriggerImpls.createTriggerImpl(underlyingTrigger);
+  }
+
+  private TriggerCallbackHandler createNewHandler(TriggerCallbackHandler handler) {
+    return new TriggerCallbackHandler() {
+      @Override
+      public void onTrigger() {
+          //re-schedule the underlying trigger for execution again.
+          underlyingTriggerImpl = TriggerImpls.createTriggerImpl(underlyingTrigger);
```
nickpan47: I am also a bit confused about the handler.onTrigger()'s invoking sequence. Based on the code in line 51, the current handler is passed into this createNewHandler() as the parameter. Hence, the newer handler for the future repeating TriggerImpl is returned by this method. So, handler.onTrigger() will be invoked in backward order, comparing to the invocation sequence of onMessage(). How are we keep the ordering of repeating triggers' firing sequence?

samza-core/src/main/java/org/apache/samza/operators/triggers/RepeatingTriggerImpl.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.samza.operators.data.MessageEnvelope;
+
+/**
+ * Implementation class for a {@link RepeatingTrigger}
+ */
+public class RepeatingTriggerImpl<M extends MessageEnvelope> implements TriggerImpl<M> {
+
+  private final Trigger<M> underlyingTrigger;
+
+  private TriggerImpl<M> underlyingTriggerImpl;
```
nickpan47: It seems to me that it would be easier to read if we keep private final RepeatingTrigger<M> instead of underlyingTrigger here. And I believe underlyingTriggerImpl is just referring to the current TriggerImpl object for the repeating trigger, right? If yes, can we rename it to currentTriggerImpl?

samza-core/src/main/java/org/apache/samza/operators/triggers/TimeTriggerImpl.java
```diff
@@ -0,0 +1,53 @@
...
+    if (latestFuture == null) {
+      latestFuture =  context.scheduleCallback(() -> {
+          handler.onTrigger();
+        }, callbackTime);
+    }
+  }
+
+  @Override
+  public void cancel() {
+    latestFuture.cancel();
```
nickpan47: nit: better to check latestFuture != null;

samza-core/src/main/java/org/apache/samza/operators/triggers/TimeSinceLastMessageTriggerImpl.java
```diff
@@ -0,0 +1,61 @@
...
+    callbackTime = currTime + durationMs;
+    Runnable runnable = () -> {
+      handler.onTrigger();
+    };
+
+    latestFuture = context.scheduleCallback(runnable, callbackTime);
+  }
+
+  @Override
+  public void cancel() {
```
nickpan47: nit: check null here.

samza-core/src/main/java/org/apache/samza/operators/triggers/TimeSinceFirstMessageTriggerImpl.java
```diff
@@ -0,0 +1,52 @@
...
+      Long callbackTime = now + triggerDurationMs;
+
+      latestFuture =  context.scheduleCallback(() -> {
+          handler.onTrigger();
+        }, callbackTime);
+    }
+  }
+
+  @Override
+  public void cancel() {
```
nickpan47: nit: check null

samza-core/src/main/java/org/apache/samza/operators/util/InternalInMemoryStore.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Implements a {@link KeyValueStore} using an in-memory Java Map.
+ * @param <K>
+ * @param <V>
+ *
+ * TODO: Provide a persistent implementation of the KeyValueStore API. Maybe, we don't need this class.
```
nickpan47: This is also because that we don't have a way to instantiate the KeyValueStore w/o the configuration, right? The TaskContext does not support creating a pluggable KeyValueStore from within the program. Hence, we have to create this class as a close-gap solution in short-term. We should make this point clear as well.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -100,8 +103,11 @@ public final void process(IncomingMessageEnvelope ime, MessageCollector collecto
   }
 
   @Override
-  public final void window(MessageCollector collector, TaskCoordinator coordinator) throws Exception {
-    // TODO: invoke timer based triggers
+  public final void window(MessageCollector collector, TaskCoordinator coordinator)  {
+    context.getSystemStreamPartitions().forEach(ssp -> {
+        OperatorImpl impl = this.operatorGraph.get(ssp.getSystemStream());
+        impl.onTick(collector, coordinator);
```
nickpan47: I wonder that it would help if we allow passing in a clock function in onTick that always returns the nanoSecond when window() is invoked, to avoid the potential perf issue in calling System.nanoTime() multiple times in each operator.onTimer().

samza-core/src/test/java/org/apache/samza/operators/triggers/TestWindowOperator.java
nickpan47: 👍 I think that would also help to address the potential perf issue on calling System.nanoTime() multiple times when invoking the onTimer() multiple times.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -75,14 +76,16 @@ public StreamOperatorTask(StreamGraphBuilder graphBuilder) {
     this.graphBuilder = graphBuilder;
   }
 
+  private TaskContext context;
```
vjagadish1989: Good point, will fix in the upcoming PR.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
vjagadish1989: The Clock interface is in place now, but it uses the SystemClock.instance. We can certainly define `SettableClock` that has a `setCurrentTime` method and returns the cached timestamp instead of querying a `getCurrentTimeMillis()`. 

Are you fine if optimizations such as these are done as a follow-up to this PR?

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -100,8 +103,11 @@ public final void process(IncomingMessageEnvelope ime, MessageCollector collecto
   }
 
   @Override
-  public final void window(MessageCollector collector, TaskCoordinator coordinator) throws Exception {
-    // TODO: invoke timer based triggers
+  public final void window(MessageCollector collector, TaskCoordinator coordinator)  {
+    context.getSystemStreamPartitions().forEach(ssp -> {
+        OperatorImpl impl = this.operatorGraph.get(ssp.getSystemStream());
```
vjagadish1989: Was mostly to prevent a long chained method invocation. (It makes it easier on my eyes to read two lines instead of one ;) ). 
Thanks for the suggestion. I've fixed it in the draft

samza-api/src/main/java/org/apache/samza/operators/windows/WindowKey.java
```diff
@@ -52,4 +52,29 @@ public String toString() {
     }
     return String.format("%s%s", wndKey, paneId);
   }
+
+  @Override
+  public boolean equals(Object o) {
```
vjagadish1989: Good point. I modified the docs to have more details (on what `paneId` means for session and tumbling windows).

samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
```diff
@@ -114,10 +117,10 @@ private Windows() { }
    * @return the created {@link Window} function.
    */
   public static <M, K, WV> Window<M, K, WV>
-    keyedTumblingWindow(Function<M, K> keyFn, Duration interval, BiFunction<M, WV, WV> foldFn) {
+    keyedTumblingWindow(Function<M, K> keyFn, Duration interval, Supplier<WV> initialValue, BiFunction<M, WV, WV> foldFn) {
```
vjagadish1989: IIUC, aggregator, key Function, initialValue supplier etc. are purely stateless JDK style `Function` classes. If there's state needed, can't we always access it from the context manager? A nice property is that users can use JDK style functional interfaces in code.

samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
```diff
@@ -264,110 +275,12 @@ private Windows() { }
...
-   * </pre>
-   *
-   * @param <M> the type of message
-   * @return the created {@link Window} function.
-   */
-  public static <M> Window<M, Void, Collection<M>> globalWindow() {
-    BiFunction<M, Collection<M>, Collection<M>> aggregator = (m, c) -> {
-      c.add(m);
-      return c;
+    Supplier<Collection<M>> initialValue = () -> {
```
vjagadish1989: I've refactored the lambda to be a one-liner as
`() -> new ArrayList<>()` . 

Let me know if you still prefer the inner function class approach.

samza-api/src/main/java/org/apache/samza/operators/windows/internal/WindowInternal.java
```diff
@@ -55,17 +56,23 @@
    */
   private final Function<M, Long> eventTimeExtractor;
 
-  private Trigger earlyTrigger;
+  private final WindowType windowType;
```
vjagadish1989: For ones that probably need explanation (for instance, `initialValueSupplier` and `windowType`, I ended up adding javadocs. I refrained from adding javadocs for obvious ones like `earlyTrigger`, `lateTrigger` and `defaultTrigger`, since they are rather superfluous.

samza-api/src/main/java/org/apache/samza/operators/windows/internal/WindowInternal.java
```diff
@@ -86,18 +93,22 @@ public WindowInternal(Trigger defaultTrigger, BiFunction<M, WV, WV> foldFunction
...
+  public Trigger<M> getDefaultTrigger() {
     return defaultTrigger;
   }
 
-  public Trigger getEarlyTrigger() {
+  public Trigger<M> getEarlyTrigger() {
     return earlyTrigger;
   }
 
   public Trigger getLateTrigger() {
```
vjagadish1989: Good catch, will fix in the upcoming draft

samza-api/src/main/java/org/apache/samza/operators/windows/internal/WindowType.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.operators.windows.internal;
+
+public enum WindowType {
```
vjagadish1989: Fixed.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorGraph.java
```diff
@@ -154,7 +154,7 @@ public void init(Map<SystemStream, MessageStreamImpl> inputStreams, Config confi
     } else if (operatorSpec instanceof SinkOperatorSpec) {
       return new SinkOperatorImpl<>((SinkOperatorSpec<M>) operatorSpec, config, context);
     } else if (operatorSpec instanceof WindowOperatorSpec) {
-      return new WindowOperatorImpl<>((WindowOperatorSpec<M, ?, ?>) operatorSpec, source, config, context);
+      return new WindowOperatorImpl((WindowOperatorSpec<M, ?, ?>) operatorSpec);
```
vjagadish1989: Good point, fixed!

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -52,6 +52,26 @@ void registerNextOperator(OperatorImpl<RM, ?> nextOperator) {
...
+   * downstream operators.
+   *
+   * @param collector
+   * @param coordinator
+   */
+  public abstract void onTimer(MessageCollector collector, TaskCoordinator coordinator);
```
vjagadish1989: A nice property is that it forces the implementor to think if there's anything to be done periodically in the `timer` callback. (and explicitly define a no-op). We seem to have equal # of stateless and stateful operators, so, it's no big deal either way.

I've fixed as per your suggestion.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -64,4 +84,7 @@ void propagateResult(RM outputMessage, MessageCollector collector, TaskCoordinat
     nextOperators.forEach(sub -> sub.onNext(outputMessage, collector, coordinator));
   }
 
+  private void propagateTimer(MessageCollector collector, TaskCoordinator coordinator) {
```
vjagadish1989: There are 2 possible approaches here:
1. Have a single timer queue for the entire operatorGraph.
2. Have per-instance queues for each of the individual `operatorImpl`s in the operatorGraph. (as is being done currently by this patch)

The net number of callbacks is exactly equivalent. Assuming `k` operators, Having one queue with size `N` is equivalent to `k` queues with size `N/k`. There maybe trivial gains in iterating over the graph (but, that's just a no-op traversal or two)

Approach [2] also has a nice property that there's no shared state across operator impls. (queues are local to the operator).

(Had a discussion and explained this with Yi offline)

samza-core/src/main/java/org/apache/samza/operators/impl/RootOperatorImpl.java
```diff
@@ -32,4 +32,8 @@
   public void onNext(M message, MessageCollector collector, TaskCoordinator coordinator) {
     this.propagateResult(message, collector, coordinator);
   }
+
+  @Override
+  public void onTimer(MessageCollector collector, TaskCoordinator coordinator) {
```
vjagadish1989: Thanks for the suggestion, Fixed.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,323 @@
...
+ *
+ * @param <M>  the type of the incoming {@link MessageEnvelope}
+ * @param <K>  the type of the key in this {@link org.apache.samza.operators.MessageStream}
+ * @param <WK> the type of the key in the emitted window pane
+ * @param <WV> the type of the value in the emitted window pane
+ * @param <WM> the type of the emitted window pane
+ *
+ * TODO: Implement expiration of entries and triggers
+ */
+public class WindowOperatorImpl<M extends MessageEnvelope, K, WK, WV, WM extends WindowPane<K, WV>> extends OperatorImpl<M, WM> {
```
vjagadish1989: Good suggestion, a nice fallout is that there's one less type parameter in the `WindowOperatorImpl` (which I <3).

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,323 @@
...
+    WV newVal = foldFunction.apply(message, wv);
+    store.put(storeKey, newVal);
+
+    if (window.getEarlyTrigger() != null) {
+      TriggerKey<K> triggerKey = new TriggerKey<>(TriggerType.EARLY, storeKey);
+      getOrCreateTriggerWrapper(triggerKey, window.getEarlyTrigger())
+          .onMessage(message, collector, coordinator);
+    }
+
+    if (window.getDefaultTrigger() != null) {
```
vjagadish1989: Good point, processing time windows do not have late arrivals. All emissions are either early or on time.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,323 @@
...
+import java.util.function.Function;
 
-  private final WindowInternal<M, WK, WV> window;
+/**
+ * Implementation of a window operator that groups messages into finite windows for processing.
+ *
+ * <p>Notes:
+ * This class implements the processing logic for various types of windows and triggers. It tracks and manages state for
+ * all open windows, the active triggers that correspond to each of the windows and the pending callbacks. It provides
+ * an implementation of {@link TriggerContext} that {@link TriggerImpl}s can use to schedule and cancel callbacks. It
```
vjagadish1989: Added more docs for the relationship between these classes. (in addition to additional comments).

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,323 @@
...
+      key = keyExtractor.apply(message);
+    }
+
+    String windowId = null;
+    if (window.getWindowType() == WindowType.TUMBLING) {
+      long triggerDurationMs = ((TimeTrigger<M>) window.getDefaultTrigger()).getDuration().toMillis();
+      final long now = System.currentTimeMillis();
+      Long windowBoundary = now - now % triggerDurationMs;
+      windowId = windowBoundary.toString();
+    }
```
vjagadish1989: For a processing time session window, there will be atmost one entry in the store for every key `K`. However, when we emit the output, we will lift the `K` and add a timestamp of the first message as the `paneId`. 


>> How about count-based window?
I don't think view "count-based window" as a top-level concept. We can still have a tumbling time window with a `count-based` trigger. The current implementation allows for that.

>> I think that at least it makes sense to move this to a getWindowId(M message) function, which can be made abstract later 

For now, there are only two window types (and we have generalized windows abstractly using triggers as much as we can) and a simple `if` is probably sufficient. We can (and probably will) revisit this when we have more window types and event time. It will also be worth to think "Should different window types emit outputs differently?".

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,323 @@
...
+    TriggerHandlerImpl triggerHandler = new TriggerHandlerImpl();
+    TriggerImpl<M> triggerImpl = TriggerImpls.createTriggerImpl(trigger);
+    TriggerContextImpl triggerContext = new TriggerContextImpl(triggerKey);
+    wrapper = new TriggerImplWrapper(triggerKey, triggerImpl, triggerContext, triggerHandler);
+    triggers.put(triggerKey, wrapper);
+
+    return wrapper;
+  }
+
+  private void handleTrigger(TriggerKey<K> triggerKey, MessageCollector collector, TaskCoordinator coordinator) {
```
vjagadish1989: `handleTrigger()` is called when the trigger when the window determines that the trigger has fired. So, I can rename it as `onTriggerFired`.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,323 @@
...
+    }
+
+    WindowKey<K> windowKey = triggerKey.key;
+    WV wv = store.get(windowKey);
+
+    if (wv == null) {
+      return;
+    }
+
+    WindowPane<K, WV> paneOutput = new WindowPane<>(windowKey, wv, window.getAccumulationMode());
```
vjagadish1989: Neat :-) thanks for the suggestion. Fixed.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,323 @@
...
+    if (window.getAccumulationMode() == AccumulationMode.DISCARDING) {
+      store.put(windowKey, null);
+    }
+
+    if (paneOutput.getMessage() instanceof Collection) {
+      WV valCopy = (WV) new ArrayList<>((Collection<WV>) paneOutput.getMessage());
+      paneOutput = new WindowPane(windowKey, valCopy, window.getAccumulationMode());
+    }
+
+    WindowOperatorImpl.super.propagateResult((WM) paneOutput, collector, coordinator);
```
vjagadish1989: Good point. It was previously done directly in the handler's `onTrigger` function. So, it required the `super` object to be accessed by closure (instead of directly). Now that we don't propagate results in the callback, we can directly use `super`.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,323 @@
...
+
+    @Override
+    public Cancellable scheduleCallback(Runnable runnable, long callbackTimeMs) {
+      TriggerTimerState timerState = new TriggerTimerState(triggerKey, runnable, callbackTimeMs);
+      pendingCallbacks.add(timerState);
+      return timerState;
+    }
+  }
+
+  private static class TriggerHandlerImpl implements TriggerImpl.TriggerCallbackHandler {
```
vjagadish1989: Good point, But,
1. Having this async API is more general, and it simplifies `TriggerImpl` classes.  They can merely invoke `handler.onTrigger`
2. It's still possible to implement this as a sync API, by having a handler implementation that merely declares an intent to trigger. And the window polls the condition later.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,323 @@
...
+    public boolean isTriggered() {
+      return triggered;
+    }
+
+    public void clearTrigger() {
+      triggered = false;
+    }
+  }
+
+  private class TriggerTimerState implements Comparable<TriggerTimerState>, Cancellable {
```
vjagadish1989: Having a `Future` is not needed because we will not be using any other method in the `Future` API other than `cancel`. For instance, `get`s does not make much sense, as we would simply return `Void`.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,323 @@
...
+      this.impl = impl;
+      this.context = context;
+      this.handler = handler;
+    }
+
+    public void onMessage(M message, MessageCollector collector, TaskCoordinator coordinator) {
+      impl.onMessage(message, context, handler);
+      if (handler.isTriggered()) {
+        //repeating trigger can trigger multiple times, So, clear the handler to allow future triggerings.
+        handler.clearTrigger();
```
vjagadish1989: Yup, you could totally do that and it would be equivalent. However,

1. It would push the complexity of managing the flags to the `TriggerImpl` classes.
2. It would add quite a bit of boiler plate to each of the `TriggerImpl` implementation classes.

This approach makes the `TriggerImpl`s much simpler (maybe useful when we allow for custom trigger impls later.)

samza-core/src/main/java/org/apache/samza/operators/triggers/AnyTriggerImpl.java
vjagadish1989: Good catch. In this case, the up-stream trigger called cancel on the `AnyTriggerImpl`. So, the rest of the triggers got invalidated. I have added the `cancel` in the callback handler to make it obvious.

samza-core/src/main/java/org/apache/samza/operators/triggers/Cancellable.java
vjagadish1989: Having a `Future<T>` is not needed because we will not be using any other method in the `Future<T>` API other than cancel. For instance, `T get()`s don't not make much sense, as we would simply return Void.

samza-core/src/main/java/org/apache/samza/operators/triggers/RepeatingTriggerImpl.java
vjagadish1989: Renamed it to `currentTriggerImpl` so that it's more obvious. Let me know if you had other suggestions.

samza-core/src/main/java/org/apache/samza/operators/triggers/RepeatingTriggerImpl.java
vjagadish1989: - We construct a wrapped handler object that proxies the invocation to the parent (after cancelation).
- Consider a case `Repeating(Triggers.count(5))`. 

In this case, when the handler for the `CountTrigger` fires, the `RepeatingTrigger` will `cancel` the `CountTrigger`, create a new instance of the `CountTriggerImpl` for a potential future firing, and propagate the firing by calling `onTrigger` on it's own handler (passed to it upstream).

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
vjagadish1989: Good suggestion, Let me prototype this, and see if it makes things simpler.

samza-core/src/main/java/org/apache/samza/operators/triggers/RepeatingTriggerImpl.java
vjagadish1989: With the revised draft, the `handler` is removed in favor of `triggerImpl.isTrigger()`.

samza-core/src/main/java/org/apache/samza/operators/util/InternalInMemoryStore.java
vjagadish1989: Yes, that's correct. This class is a stop-gap until we create a persistent store from the `TaskContext` API.

samza-core/src/main/java/org/apache/samza/operators/triggers/TriggerImpls.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Factory methods for instantiating {@link TriggerImpl}s from individual {@link Trigger}s.
+ */
+public class TriggerImpls {
+
+  public static <M extends MessageEnvelope> TriggerImpl<M> createTriggerImpl(Trigger<M> trigger, Clock clock) {
+
+    if (trigger == null) {
+      throw new SamzaException("Null trigger passed in");
```
prateekm: "Trigger must not be null". Should also probably use IllegalArgumentException.

samza-api/src/main/java/org/apache/samza/operators/functions/FoldFunction.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.apache.samza.operators.functions;
+
+/**
+ * A fold function that incrementally combines and aggregates values for a window.
+ */
+public interface FoldFunction<M, WV> extends InitableFunction {
+
+  /**
+   * Incrementally combine and aggregate values for the window. Guaranteed to be invoked for every
```
prateekm: As an aside, when we say fold I think we mean a foldLeft. Not sure if it' worth renaming or documenting this.

samza-api/src/main/java/org/apache/samza/operators/windows/internal/WindowInternal.java
```diff
@@ -55,27 +61,34 @@
    */
   private final Function<M, Long> eventTimeExtractor;
 
-  private Trigger earlyTrigger;
+  /**
+   * The type of this window. Tumbling and Session windows are supported for now.
```
prateekm: Minor: Second sentence is redundant, is documented in WindowType.

samza-core/src/main/java/org/apache/samza/operators/triggers/AnyTriggerImpl.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.samza.operators.data.MessageEnvelope;
+import org.apache.samza.util.Clock;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Implementation of an {@link AnyTrigger}
```
prateekm: Minor: It seems like an AllTrigger should be very similar to this impl. If it doesn't involve much work beyond copy-pasta + minor tweaks and you have the extra cycles might be worth adding it.

samza-core/src/main/java/org/apache/samza/operators/triggers/TriggerContext.java
```diff
@@ -0,0 +1,34 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.operators.triggers;
+
+/**
+ * Provides convenience methods for use by {@link TriggerImpl}s. This includes scheduling and cancelling
+ * callbacks, accessing window keys etc.
```
prateekm: Doesn't allow accessing window keys anymore. Also see the comment about removing this interface.

samza-core/src/test/java/org/apache/samza/operators/triggers/TestClock.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * An implementation of {@link Clock} that allows to advance the time by an arbitrary duration.
+ * Used for testing.
+ */
+public class TestClock implements Clock {
+
+  long currentTime = 1;
+
+  public void advanceTime(Duration duration) {
```
prateekm: Minor: Prefer `long ms`. Maybe add an overloaded version?

samza-core/src/main/java/org/apache/samza/operators/triggers/AnyTriggerImpl.java
```diff
@@ -0,0 +1,80 @@
...
+ *
+ */
+public class AnyTriggerImpl<M extends MessageEnvelope> implements TriggerImpl<M> {
+
+  private final List<Trigger<M>> triggerList;
+
+  private final Map<TriggerImpl<M>, Boolean> triggerImpls = new HashMap<>();
```
vjagadish1989: Initially there was a need to use a `CHM` because of a prior implementation. Now that, it's not needed, we can use a `List`. Good catch

samza-core/src/main/java/org/apache/samza/operators/triggers/AnyTriggerImpl.java
```diff
@@ -0,0 +1,80 @@
...
+      impl.cancel();
+      it.remove();
+    }
+  }
+
+  @Override
+  public boolean shouldFire() {
+    for (TriggerImpl<M> impl : triggerImpls.keySet()) {
+      if (impl.shouldFire()) {
+        shouldFire = true;
```
vjagadish1989: Added a `break` as an optimization.

samza-core/src/main/java/org/apache/samza/operators/triggers/RepeatingTriggerImpl.java
```diff
@@ -0,0 +1,63 @@
...
+    currentTriggerImpl.onMessage(message, context);
+  }
+
+  @Override
+  public void cancel() {
+    currentTriggerImpl.cancel();
+  }
+
+  @Override
+  public void clear() {
```
vjagadish1989: No strong preference on this one. Re-wrote as-per  your suggestion.

>>It feels slightly weird that this clear method exists on the Trigger interface solely for RepeatingTrigger.

- For instance, other APIs (like beam) have a `onTrigger` method that `RepeatingTriggerImpls` can use to potentially reset themselves. 

- It was not clear if other trigger implementations will potentially need to clear their state.

Removing it now, and let's plan to add it later (if there are use-cases). The `instanceOf` checks are not pretty though.

samza-core/src/main/java/org/apache/samza/operators/triggers/TriggerContext.java
vjagadish1989: Removed the doc from this interface.

samza-core/src/test/java/org/apache/samza/operators/triggers/TestClock.java
vjagadish1989: added an overloaded method.

samza-api/src/main/java/org/apache/samza/operators/functions/FoldFunction.java
vjagadish1989: @nickpan47 any preference here given the trade-offs?

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,379 @@
...
+      wv = existingState.getWindowValue();
+    }
+
+    WV newVal = window.getFoldFunction().apply(message, wv);
+    WindowState<WV> newState = new WindowState(newVal, earliestTimeStamp);
+
+    return newState;
+  }
+
+  private TriggerImplState getOrCreateTriggerWrapper(TriggerKey<K> triggerKey, Trigger<M> trigger) {
```
vjagadish1989: Renamed them to be `TriggerImplWrapper`.

samza-core/src/main/java/org/apache/samza/operators/triggers/TriggerImpl.java
```diff
@@ -0,0 +1,70 @@
...
+ * of time-based triggers).
+ *
+ * <p> State management: The state maintained by {@link TriggerImpl}s is not durable across re-starts and is transient.
+ * New instances of {@link TriggerImpl} are created on a re-start.
+ *
+ */
+public interface TriggerImpl<M> {
+
+  /**
+   * Invoked when a MessageEnvelope added to the window corresponding to this {@link TriggerImpl}.
```
vjagadish1989: Removed all references to `MessageEnvelope` in docs (here and elsewhere)

vjagadish1989:
Address @prateekm 's feedback.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,379 @@
...
+      TriggerCallbackState timerState = new TriggerCallbackState(triggerKey, runnable, callbackTimeMs);
+      pendingCallbacks.add(timerState);
+      return timerState;
+    }
+  }
+
+  /**
+   * State corresponding to pending timer callbacks scheduled by various {@link TriggerImpl}s.
+   */
+  private class TriggerCallbackState implements Comparable<TriggerCallbackState>, Cancellable {
```
vjagadish1989: Actually, one other way to do this could be:

`TriggerCallbackState(TriggerKey<WK> triggerKey, Runnable callback, long scheduledTimeMs)` to pass in the `WindowOperatorImpl` into the `TriggerCallbackState`'s constructor.

That would make `TriggerCallbackState<M, WK, WV>` be typed as well. I'd be nice to not have this verbosity for storing something simple as a combination of a key, and a Runnable :-)

samza-api/src/main/java/org/apache/samza/operators/functions/FoldFunction.java
prateekm: 1. Agree. A `getInitialValue` method is one way, others are to inform users that they should either handle the initial null value with an if-null check in apply (as earlier) or create the initial value int #init.

2. FoldFunction should be `Initable`, we don't know what other context users will need for the fold.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,340 @@
...
+ * also orchestrates the flow of messages through the various {@link TriggerImpl}s.
+ *
+ * <p> An instance of a {@link TriggerImplWrapper} is created corresponding to each {@link Trigger} configured for a
+ * particular window. For every message added to the window, this class looks up the corresponding {@link TriggerImplWrapper}
+ * for the trigger and invokes {@link TriggerImplWrapper#onNext(Object, MessageCollector, TaskCoordinator)}. The
+ * {@link TriggerImplWrapper} forwards the callback to its {@link TriggerImpl} instance. A {@link TriggerImpl} instance is scoped
+ * to a window and its firing determines when results for its window are emitted. The {@link WindowOperatorImpl} checks if
+ * the trigger fired, and propagates the result of the firing to its downstream operators.
+ *
+ * @param <M>  the type of the incoming message
```
prateekm: Codestyle: Single space after `<M>` etc.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,340 @@
...
+  private WindowState<WV> applyFoldFunction(WindowState<WV> existingState, M message) {
+    WV wv;
+    long earliesttimestamp;
+
+    if (existingState == null) {
+      wv = window.getInitializer().get();
+      earliesttimestamp = clock.currentTimeMillis();
+    } else {
+      wv = existingState.getWindowValue();
+      earliesttimestamp = existingState.getEarliestRecvTime();
```
prateekm: Naming: `existingState.getEarliestTimestamp()` for consistency.

Naming: `earliestTimestamp` (camelcase)

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,379 @@
...
+    if (windowVal instanceof Collection) {
+      windowVal = (WV) new ArrayList<>((Collection<WV>) windowVal);
+    }
+
+    WindowPane<K, WV> paneOutput = new WindowPane<>(windowKey, windowVal, window.getAccumulationMode(), triggerKey.getType());
+    return paneOutput;
+  }
+
+  /**
+   * Cancels the firing of the {@link TriggerImpl} identified by this {@link TriggerKey} and optionally removes it.
```
prateekm: Thanks. Let's document what the implications of removing vs not removing a trigger are, and how the state for the window and trigger removal is related.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,379 @@
...
+    }
+    if (shouldRemove && triggerKey != null) {
+      triggers.remove(triggerKey);
+    }
+  }
+
+  /**
+   * Implementation of the {@link TriggerContext} that allows {@link TriggerImpl}s to schedule and cancel callbacks.
+   */
+  public class TriggerContextImpl implements TriggerContext {
```
prateekm: Looks like we're only ever going to have 1 implementation for this at a time, so not sure if the interface really helps. Don't feel strongly about it, your call.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
prateekm: To clarify, I'm not proposing extracting for reusability, only for making it clearer which classes access/mutate which fields and call which methods on the parent class. Passing pendingCallbacks etc. explicitly as a shared data structures/classes is what I want :)

Agree that it makes things more verbose, but IMHO it makes code easier to understand by making it clearer how things are related. Don't think the extra type parameters would be that bad, but -1 to passing all of WindowOperatorImpl to the constructor - it's too opaque.

Maybe @cpettitt-linkedin has some thoughts on this? Otherwise your call - feel free to keep as-is if you feel it's clearer that way.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,379 @@
...
+        onTriggerFired(triggerKey, collector, coordinator);
+      }
+    }
+
+    public void cancel() {
+      impl.cancel();
+      isCancelled = true;
+    }
+
+    public boolean isRepeating() {
```
prateekm: Good name, thanks. `TriggerImplWrapper#isRepeating` makes sense, let's keep it.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,340 @@
...
+      return this.impl instanceof RepeatingTriggerImpl;
+    }
+  }
+
```
prateekm: Nitpick: Delete extra newline.

samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
```diff
@@ -84,6 +87,8 @@
  * and triggers are fired and window panes are emitted per-key. It is possible to construct "keyed" variants of all the above window
  * types.
  *
+ * <p> Time granularity for windows: Currently, time durations are always measured in milliseconds. Time units of
```
prateekm: Thanks for moving. This still sounds like a vague disclaimer - not sure if it's helpful for users. If we know what we specifically don't support, would be good to document that (maybe as an example).

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,379 @@
...
+    //For session windows, we will create a new window key by using the time of the first message in the window as
+    //the paneId.
+    if (window.getWindowType() == WindowType.SESSION) {
+      windowKey = new WindowKey<>(windowKey.getKey(), Long.toString(state.getEarliestTime()));
+    }
+
+    //Make a defensive copy so that we are immune to further mutations on the collection
+    if (windowVal instanceof Collection) {
+      windowVal = (WV) new ArrayList<>((Collection<WV>) windowVal);
```
prateekm: The safety question was for the outermost cast back to WV. 

For example, let's say users provide the initial value as a Set<String> for accumulation. IIUC, in that case the WV would be Set<String>, the inner cast would be (Collection<Set<String>>) and outer cast would be casting the ArrayList to Set<String>, right? 

Am I misunderstanding what the type WV is?

vjagadish1989: Good point, Would a check for an `ArrayList` specifically be more meaningful?

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
nickpan47: I do prefer to make the whole queue of callbacks, the context to schedule and cancel callbacks in a single class (maybe TriggerCallbackScheduler??). The functions that refer to a lot of WindowOperatorImpl internals can stay in the inner classes that mainly implementing callback functions to be invoked by the scheduler.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,363 @@
...
+    public boolean cancel() {
+      LOG.trace("Cancelled a callback: {} at {} for triggerKey {}", new Object[] {callback, scheduledTimeMs, triggerKey});
+      return pendingCallbacks.remove(this);
+    }
+  }
+
+  /**
+   * State corresponding to a created {@link TriggerImpl} instance.
+   */
+  private class TriggerImplWrapper {
```
nickpan47: So, why not TriggerImplState?

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
nickpan47: I don't quite like the name wrapper.... It does not tell me what this class is supposed to do. In fact, TriggerImplState sounds better to me because this class does keep all relevant runtime objects and states related to a trigger. If the reason for renaming is simply TriggerImplState.isRepeating() sounds weird, we can change the method to TriggerImplState.canRepeat().

samza-api/src/main/java/org/apache/samza/operators/windows/WindowKey.java
```diff
@@ -26,14 +26,19 @@
  *
  */
 public class WindowKey<K> {
-
+  /**
+   * A (key,paneId) tuple uniquely identifies an emission from a window. For instance, in case of keyed-tumbling time windows,
+   * the key is provided by the keyExtractor function, and the paneId is the start of the time window boundary. In case
+   * of session windows, the key is provided by the keyExtractor function, and the paneId is the time at which the earliest
```
vjagadish1989: Yes that's right. I can also document it. The two are essentially equivalent in case of session windows.

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
```diff
@@ -44,4 +44,9 @@
   public void onNext(M message, MessageCollector collector, TaskCoordinator coordinator) {
     // TODO: implement PartialJoinOperatorImpl processing logic
   }
+
+  @Override
+  public void onTimer(MessageCollector collector, TaskCoordinator coordinator) {
+
```
vjagadish1989: This class has changed quite a bit after @prateekm 's RB. Will rebase against the latest.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -18,26 +18,340 @@
...
+ *
+ * This class implements the processing logic for various types of windows and triggers. It tracks and manages state for
+ * all open windows, the active triggers that correspond to each of the windows and the pending callbacks. It provides
+ * an implementation of {@link TriggerContext} that {@link TriggerImpl}s can use to schedule and cancel callbacks. It
+ * also orchestrates the flow of messages through the various {@link TriggerImpl}s.
+ *
+ * <p> An instance of a {@link TriggerImplWrapper} is created corresponding to each {@link Trigger} configured for a
+ * particular window. For every message added to the window, this class looks up the corresponding {@link TriggerImplWrapper}
+ * for the trigger and invokes {@link TriggerImplWrapper#onNext(Object, MessageCollector, TaskCoordinator)}. The
+ * {@link TriggerImplWrapper} forwards the callback to its {@link TriggerImpl} instance. A {@link TriggerImpl} instance is scoped
```
vjagadish1989: I thought more about this after our discussion. `triggerImplState.onTimer()` and `triggerImplState.onMessage()` felt awkward. All that is doing, is wrapping the `TriggerImpl` and forwarding the callback to its underlying `TriggerImpl`.

prateekm: @nickpan47 I'd strongly prefer to not call it TriggerImplState:
1. It makes the word state ambiguous in this context, since it also refers to the state  corresponding to the window key (and hence the trigger).
2. Calling it state seems to imply that it's a POJO container for the runtime state for the trigger, but the more interesting part of this class's functionality is the behavior (methods) associated with that state. Hence TriggerImplWrapper, TriggerImplManager sounds more appropriate to me.

vjagadish1989:
@nickpan47 :

>> I do prefer to make the whole queue of callbacks, the context to schedule and cancel callbacks in a single class (maybe TriggerCallbackScheduler??). 

Wouldn't this will move complexity (and simply a bunch of code) from `WindowOperatorImpl` to a different`TriggerCallbackScheduler` class? The context to schedule and cancel callbacks must be created *per* trigger key instance. 

>>The functions that refer to a lot of WindowOperatorImpl internals can stay in the inner classes that mainly implementing callback functions to be invoked by the scheduler.

I extracted out inner classes for greater readability into their separate files.

checkstyle/checkstyle.xml
```diff
@@ -60,10 +60,9 @@
     <module name="TypeName"/>
     
     <!-- whitespace -->
-    <module name="GenericWhitespace"/>
```
prateekm: Why remove this?

vjagadish1989: I believe this check is buggy for handling collections with templatized, non-static inner classes. 

`PriorityQueue<TriggerContext<WK>.TriggerCallbackState<WK>>`

prateekm: Seems like a useful check to have. Probably better to locally suppress it in the few locations where it's broken.

vjagadish1989: Was able to disable checkStyle for `WindowOperatorImpl.java`. 

For future reference, we can now use `// CHECKSTYLE:OFF` to disable checkstyle checks for specific files.

samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
vjagadish1989: I have renamed it to `FoldLeftFunction` since you and @prateekm preferred it.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -268,100 +267,40 @@ private void cancelTrigger(TriggerKey<WK> triggerKey, boolean shouldRemove) {
...
-
-    @Override
-    public boolean cancel() {
-      LOG.trace("Cancelled a callback: {} at {} for triggerKey {}", new Object[] {callback, scheduledTimeMs, triggerKey});
-      return pendingCallbacks.remove(this);
-    }
-  }
-
-  /**
    * State corresponding to a created {@link TriggerImpl} instance.
```
nickpan47: Now, with the cleanup of code, this class seems to be implementing "handler" functions of a TriggerImpl. Maybe better to name as TriggerImplHandler?

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -101,29 +103,27 @@ public void onNext(M message, MessageCollector collector, TaskCoordinator coordi
...
   }
 
   @Override
   public void onTimer(MessageCollector collector, TaskCoordinator coordinator) {
-    long now = clock.currentTimeMillis();
-    TriggerCallbackState state;
-    while ((state = pendingCallbacks.peek()) != null && state.getScheduledTimeMs() <= now) {
-      pendingCallbacks.remove();
-      state.getCallback().run();
+    List<TriggerKey<WK>> keys = triggerScheduler.runPendingCallbacks();
```
nickpan47: nit: I thought that the following pattern will allow window operator specific code lives here and let scheduler to invoke callbacks accordingly, something like:
triggerScheduler.runPendingCallbacks((key) -> {
  TriggerImplWrapper wrapper = triggers.get(key);
  if (wrapper != null) {
    wrapper.onTimer(key, collector, coordinator);
  }
})

Not too strong on this point. If you have specific reasons not to, I am fine with the current code.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
vjagadish1989: Sure, renamed it to `TriggerImplHandler` as suggested.

PR #67, opened 2017-02-27
SAMZA-1107:Job model publish

sborya:
add utils for publishing job model and job model version to ZK.


samza-core/src/main/java/org/apache/samza/zk/ZkKeyBuilder.java
```diff
@@ -88,5 +88,7 @@ public String getJobModelPath(String jobModelVersion) {
     return String.format("%s/%s", getJobModelPathPrefix(), jobModelVersion);
   }
 
-
+  public String getJobModelVersionBarrierPrefix() {
+    return String.format("/%s/versionBarriers", pathPrefix);
```
fredji97: when you use pathPrefix, does it already have a "/" at the beginning? If so, you may not want the additional "/"

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -156,16 +161,70 @@ public void subscribeToJobModelVersionChange(IZkDataListener dataListener) {
...
+
+  /**
    * read the jobmodel version from ZK
    * @return jobmodel version as a string
    */
   public String getJobModelVersion() {
     return zkClient.<String>readData(keyBuilder.getJobModelVersionPath());
   }
 
+  public void publishNewJobModelVersion(String oldVersion, String newVersion) {
```
fredji97: it would be helpful to add javadoc as well for this function.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -156,16 +161,70 @@ public void subscribeToJobModelVersionChange(IZkDataListener dataListener) {
...
+    Stat stat = new Stat();
+    String currentVersion = zkClient.<String>readData(keyBuilder.getJobModelVersionPath(), stat);
+    LOG.info("pid=" + processorId + " publishing new version: " + newVersion + "; oldVersion = " + oldVersion + "(" + stat.getVersion() + ")");
+    if (currentVersion != null && !currentVersion.equals(oldVersion)) {
```
fredji97: Is it valid if currentVersion == null?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -156,16 +161,70 @@ public void subscribeToJobModelVersionChange(IZkDataListener dataListener) {
...
+      throw new SamzaException("Someone change JMVersion while Leader was generating: expected" + oldVersion + ", got " + currentVersion);
+    }
+    int dataVersion = stat.getVersion();
+    stat = zkClient.writeDataReturnStat(keyBuilder.getJobModelVersionPath(), newVersion, dataVersion);
+    if (stat.getVersion() != dataVersion + 1)
+      throw new SamzaException("Someone changed data version of the JMVersion while Leader was generating a new one. current= " + dataVersion + ", old version = " + stat.getVersion());
```
fredji97: which one is new data version? Is it opposite?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -156,16 +161,70 @@ public void subscribeToJobModelVersionChange(IZkDataListener dataListener) {
...
+
+    LOG.info("pid=" + processorId +
+             " published new version: " + newVersion + "; expected data version = " + dataVersion + "(" + stat.getVersion()
```
fredji97: what do the log mean for the expected data version = <version_1>(<version_2>)? It is really hard for engineers who are not into the code to interpret the log.

samza-core/src/main/java/org/apache/samza/zk/ZkKeyBuilder.java
sborya: good point.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: good point. It is valid - means first update ever, but we need a separate check for this.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: data version, is ZK version of the data written. So after update it needs to be previous version + 1. (to make sure no one update in the meanwhile)

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: fixed log.

samza-core/src/main/java/org/apache/samza/zk/ZkKeyBuilder.java
fredji97: Alright.  I think I was misled by the function getJobModelPath(), that one has a "/" at the beginning. I think we should keep them in consistency for the convention about path prefix.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
fredji97: Seems unlikely, but I am wondering when someone changed the version, is it possible that he/she deleted the data version?

sborya: I've verified. CurrentVersion will be null for the first time, and stat.getVersion() will return 0 in this case.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -156,6 +161,42 @@ public void subscribeToJobModelVersionChange(IZkDataListener dataListener) {
   }
 
   /**
+   * publishes new job model into ZK
+   * @param jobModelVersion  version of the jobModeL to publish
+   * @param jobModel jobModel to publish
+   */
+  public void publishNewJobModel(String jobModelVersion, JobModel jobModel) {
+    try {
+      // This call should FAIL if the node already exists.
```
navina: nit: move this to javadoc for the function as it defines the behavior of the function

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -156,6 +161,42 @@ public void subscribeToJobModelVersionChange(IZkDataListener dataListener) {
   }
 
   /**
+   * publishes new job model into ZK
+   * @param jobModelVersion  version of the jobModeL to publish
+   * @param jobModel jobModel to publish
+   */
+  public void publishNewJobModel(String jobModelVersion, JobModel jobModel) {
```
navina: Can we rename the function to publishJobModel, instead of publishNewJobModel?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -156,6 +161,42 @@ public void subscribeToJobModelVersionChange(IZkDataListener dataListener) {
...
+    try {
+      // This call should FAIL if the node already exists.
+      ObjectMapper mmapper = SamzaObjectMapper.getObjectMapper();
+      String jobModelStr = mmapper.writerWithDefaultPrettyPrinter().writeValueAsString(jobModel);
+      LOG.info("pid=" + processorId + " jobModelAsString=" + jobModelStr);
+      zkClient.createPersistent(keyBuilder.getJobModelPath(jobModelVersion), jobModelStr);
+      LOG.info("wrote jobModel path =" + keyBuilder.getJobModelPath(jobModelVersion));
+    } catch (Exception e) {
+      throw new SamzaException(e);
```
navina: nit: Please add some logs

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -156,6 +161,42 @@ public void subscribeToJobModelVersionChange(IZkDataListener dataListener) {
...
+    try {
+      jm = mmapper.readValue((String) data, JobModel.class);
+    } catch (IOException e) {
+      throw new SamzaException("failed to read JobModel from ZK", e);
+    }
+    return jm;
+  }
+
+  /**
    * read the jobmodel version from ZK
```
navina: Looks like getJobModelVersion() is used only in tests. Can you add @VisibleForTesting annotation?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -164,8 +205,33 @@ public String getJobModelVersion() {
...
+   * @param oldVersion - used to validate, that no one has changed the version in the meanwhile.
+   * @param newVersion - new version.
+   */
+  public void publishNewJobModelVersion(String oldVersion, String newVersion) {
+    Stat stat = new Stat();
+    String currentVersion = zkClient.<String>readData(keyBuilder.getJobModelVersionPath(), stat);
+    LOG.info("pid=" + processorId + " publishing new version: " + newVersion + "; oldVersion = " + oldVersion + "(" + stat.getVersion() + ")");
+
+    if (currentVersion != null && !currentVersion.equals(oldVersion)) {
+      throw new SamzaException("Someone change JMVersion while Leader was generating: expected" + oldVersion + ", got " + currentVersion);
```
navina: nit: JMVersion -> Job Model Version 
I read it as JVMVersion the first time :P

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -164,8 +205,33 @@ public String getJobModelVersion() {
   }
 
   /**
+   * publish the version number of the next JobModel
+   * @param oldVersion - used to validate, that no one has changed the version in the meanwhile.
+   * @param newVersion - new version.
+   */
+  public void publishNewJobModelVersion(String oldVersion, String newVersion) {
```
navina: Same as above: can we rename publishNewJobModelVersion to  updateJobModelVersion??

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -164,8 +205,33 @@ public String getJobModelVersion() {
...
+    Stat stat = new Stat();
+    String currentVersion = zkClient.<String>readData(keyBuilder.getJobModelVersionPath(), stat);
+    LOG.info("pid=" + processorId + " publishing new version: " + newVersion + "; oldVersion = " + oldVersion + "(" + stat.getVersion() + ")");
+
+    if (currentVersion != null && !currentVersion.equals(oldVersion)) {
+      throw new SamzaException("Someone change JMVersion while Leader was generating: expected" + oldVersion + ", got " + currentVersion);
+    }
+    // data version is the ZK version of the data from the ZK.
+    int dataVersion = stat.getVersion();
+    stat = zkClient.writeDataReturnStat(keyBuilder.getJobModelVersionPath(), newVersion, dataVersion);
```
navina: Iirc, this method can throw KeeperException with Bad Version if the expected version is not available. Can you please verify and add exception handling, if needed?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -164,8 +205,33 @@ public String getJobModelVersion() {
...
+    if (stat.getVersion() != dataVersion + 1)
```
navina: I don't think you need this check if you handle the KeeperException from the conditional write. Please update according to the previous line's fix.

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
```diff
@@ -31,6 +37,7 @@
 import org.junit.BeforeClass;
 import org.junit.Test;
 
+
```
navina: nit: extra whiteline

samza-core/src/main/java/org/apache/samza/zk/ZkKeyBuilder.java
sborya: Most of the get*path*() methods use pathPrefix at the start, that's why they use '/' as the first character.
getJobModelVersion() uses getJobModelPathPrefix(), which already includes the '/' at the beginning.

I added a test anyway.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: good point.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: well, this will be used by JobCoordinator for publishing.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: Verified. You are right. Added try/catch.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: I guess so.

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
sborya: done.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -209,20 +211,23 @@ public String getJobModelVersion() {
...
     // data version is the ZK version of the data from the ZK.
     int dataVersion = stat.getVersion();
-    stat = zkClient.writeDataReturnStat(keyBuilder.getJobModelVersionPath(), newVersion, dataVersion);
-    if (stat.getVersion() != dataVersion + 1)
-      throw new SamzaException("Someone changed data version of the JMVersion while Leader was generating a new one. current= " + dataVersion + ", old version = " + stat.getVersion());
-
+    try {
+      stat = zkClient.writeDataReturnStat(keyBuilder.getJobModelVersionPath(), newVersion, dataVersion);
+    } catch (Exception e) {
+      LOG.error("publish job model version failed for new version = " + newVersion + "; old version = " + oldVersion, e);
```
navina: Shouldn't we re-throw the exception here?

PR #68, opened 2017-02-27
SAMZA-1100: Exception when using a stream as both bootstrap and broadcast.

shanthoosh:
When a task input stream is used as both broadcast and bootstrap stream in a samza job, Bootstrappingchooser marks the stream as bootstrapped when a single task finishes consuming all the SystemStreamPartitions(This happens when all the starting offset for each partition in the input stream is of type upcoming). This patch fixes this, by marking a stream as bootstrapped, only when all the systemStreamPartitions in a input stream is consumed by all the expected tasks.

More details here : https://issues.apache.org/jira/browse/SAMZA-1100

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
+++ b/samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
@@ -354,7 +354,7 @@ object SamzaContainer extends Logging {
 
     val chooserFactory = Util.getObj[MessageChooserFactory](chooserFactoryClassName)
 
-    val chooser = DefaultChooser(inputStreamMetadata, chooserFactory, config, samzaContainerMetrics.registry)
+    val chooser = DefaultChooser(inputStreamMetadata, chooserFactory, config, samzaContainerMetrics.registry, systemAdmins)
 
     info("Setting up metrics reporters.")
 
--- a/samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
+++ b/samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
@@ -19,7 +19,6 @@
 
 package org.apache.samza.system.chooser
 
-import java.util.concurrent.atomic.AtomicInteger
 import org.apache.samza.system.SystemStream
 import org.apache.samza.system.SystemStreamPartition
 import org.apache.samza.system.IncomingMessageEnvelope
@@ -28,9 +27,11 @@ import org.apache.samza.metrics.MetricsHelper
 import org.apache.samza.metrics.MetricsRegistryMap
 import org.apache.samza.metrics.MetricsRegistry
 import org.apache.samza.system.SystemStreamMetadata
+
 import scala.collection.JavaConversions._
-import org.apache.samza.SamzaException
 import org.apache.samza.system.SystemStreamMetadata.OffsetType
+
+import scala.collection.mutable
 
 /**
  * BootstrappingChooser is a composable MessageChooser that only chooses
@@ -69,7 +70,13 @@ class BootstrappingChooser(
   /**
    * An object that holds all of the metrics related to bootstrapping.
    */
-  metrics: BootstrappingChooserMetrics = new BootstrappingChooserMetrics) extends MessageChooser with Logging {
+  metrics: BootstrappingChooserMetrics = new BootstrappingChooserMetrics,
+
+  /**
+   * A map from system stream name to SystemAdmin that is used for
+   * offset comparisons.
+   */
+  systemAdmins: mutable.Map[String, SystemAdmin] = mutable.Map()) extends MessageChooser with Logging {
 
   /**
    * The number of lagging partitions for each SystemStream that's behind.
@@ -93,7 +100,7 @@ class BootstrappingChooser(
   /**
    * Store all the systemStreamPartitions registered
    */
-  var registeredSystemStreamPartitions = Set[SystemStreamPartition]()
+  var registeredSystemStreamPartitions: mutable.Map[SystemStreamPartition, String] = mutable.Map[SystemStreamPartition, String]()
 
   /**
    * The number of lagging partitions that the underlying wrapped chooser has
@@ -102,8 +109,16 @@ class BootstrappingChooser(
   var updatedSystemStreams = Map[SystemStream, Int]()
 
   def start = {
+    for ((systemStreamPartition, offset) <- registeredSystemStreamPartitions) {
+      // If the offset we're starting to consume from is the same as the upcoming
+      // offset for this system stream partition, then we've already read all
+      // messages in the stream, and we're at head for this system stream
+      // partition.
+      checkOffset(systemStreamPartition, offset, OffsetType.UPCOMING)
+    }
+
     // remove the systemStreamPartitions not registered.
-    laggingSystemStreamPartitions = laggingSystemStreamPartitions.filter(registeredSystemStreamPartitions.contains(_))
+    laggingSystemStreamPartitions = laggingSystemStreamPartitions.filter(registeredSystemStreamPartitions.keys.contains(_))
     systemStreamLagCounts = laggingSystemStreamPartitions.groupBy(_.getSystemStream).map {case (systemStream, ssps) => systemStream -> ssps.size}
 
     debug("Starting bootstrapping chooser with bootstrap metadata: %s" format bootstrapStreamMetadata)
@@ -120,15 +135,25 @@ class BootstrappingChooser(
   override def register(systemStreamPartition: SystemStreamPartition, offset: String) {
     debug("Registering stream partition with offset: %s, %s" format (systemStreamPartition, offset))
 
-    // If the offset we're starting to consume from is the same as the upcoming 
-    // offset for this system stream partition, then we've already read all
-    // messages in the stream, and we're at head for this system stream 
-    // partition.
-    checkOffset(systemStreamPartition, offset, OffsetType.UPCOMING)
-
     wrapped.register(systemStreamPartition, offset)
 
-    registeredSystemStreamPartitions += systemStreamPartition
+    val systemStream = systemStreamPartition.getSystem
+    val systemAdmin: SystemAdmin = systemAdmins.getOrElse(systemStream,
+                                                          throw new SamzaException("SystemAdmin is undefined for SystemStream: %s" format systemStream))
+    /**
+     * SAMZA-1100: When a input SystemStream is consumed as both bootstrap and broadcast
+     * BootstrappingChooser should record the lowest offset for each registered SystemStreamPartition.
+     * When multiple tasks processing a SystemStreamPartition runs within a container
+     * and share the same chooser, then the lowest offset should be chosen as starting offset.
+     */
+    if (!registeredSystemStreamPartitions.contains(systemStreamPartition)) {
+      registeredSystemStreamPartitions += systemStreamPartition -> offset
+    } else if (offset != null) {
+      val comparatorResult: Integer = systemAdmin.offsetComparator(registeredSystemStreamPartitions(systemStreamPartition), offset)
+      if (comparatorResult != null && comparatorResult > 0) {
+        registeredSystemStreamPartitions += systemStreamPartition -> offset
+      }
+    }
   }
 
   def update(envelope: IncomingMessageEnvelope) {
--- a/samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
+++ b/samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
@@ -19,18 +19,19 @@
 
 package org.apache.samza.system.chooser
 
+import org.apache.samza.system.SystemAdmin
 import org.apache.samza.system.SystemStream
+import org.apache.samza.system.SystemStreamMetadata
 import org.apache.samza.system.SystemStreamPartition
 import org.apache.samza.system.IncomingMessageEnvelope
+import org.apache.samza.SamzaException
 import org.apache.samza.util.Logging
 import org.apache.samza.metrics.MetricsHelper
 import org.apache.samza.metrics.MetricsRegistryMap
 import org.apache.samza.metrics.MetricsRegistry
-import org.apache.samza.system.SystemStreamMetadata
-
-import scala.collection.JavaConversions._
 import org.apache.samza.system.SystemStreamMetadata.OffsetType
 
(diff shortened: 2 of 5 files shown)


samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -354,7 +355,14 @@ object SamzaContainer extends Logging {
 
     val chooserFactory = Util.getObj[MessageChooserFactory](chooserFactoryClassName)
 
-    val chooser = DefaultChooser(inputStreamMetadata, chooserFactory, config, samzaContainerMetrics.registry)
+    val sspToTaskCount: mutable.Map[SystemStreamPartition, Int] = mutable.Map[SystemStreamPartition, Int]()
```
sborya: Please add a comment here explaining why we are doing it.

shanthoosh: The explanation was done at multiple places, including BootstrappingChooser, DefaultChooser constructors. Hence didn't add it here. Done. Added a comment, describing the intention.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -78,17 +85,19 @@ class BootstrappingChooser(
...
-  var laggingSystemStreamPartitions = bootstrapStreamMetadata
-    .flatMap {
+  var laggingSystemStreamPartitions = mutable.Map[SystemStreamPartition, Int]()
+  bootstrapStreamMetadata.foreach {
       case (systemStream, metadata) =>
-        metadata
-          .getSystemStreamPartitionMetadata
-          .keys
-          .map(new SystemStreamPartition(systemStream, _))
+        metadata.getSystemStreamPartitionMetadata.map {
```
sborya: See if it can be rewritten more scala style :)

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -103,8 +112,8 @@ class BootstrappingChooser(
 
   def start = {
     // remove the systemStreamPartitions not registered.
-    laggingSystemStreamPartitions = laggingSystemStreamPartitions.filter(registeredSystemStreamPartitions.contains(_))
-    systemStreamLagCounts = laggingSystemStreamPartitions.groupBy(_.getSystemStream).map {case (systemStream, ssps) => systemStream -> ssps.size}
+    laggingSystemStreamPartitions = laggingSystemStreamPartitions.filter((sspToTaskCount: (SystemStreamPartition, Int)) => registeredSystemStreamPartitions.contains(sspToTaskCount._1))
+    systemStreamLagCounts = laggingSystemStreamPartitions.keys.groupBy(_.getSystemStream).map {case (systemStream, ssps) => systemStream -> ssps.size}
```
sborya: please add comment why do we have 2 counters and what is the relationship between them.

samza-core/src/main/scala/org/apache/samza/system/chooser/DefaultChooser.scala
```diff
@@ -244,7 +245,13 @@ class DefaultChooser(
...
    * Metrics registry to be used when wiring up wrapped choosers.
    */
-  registry: MetricsRegistry = new MetricsRegistryMap) extends MessageChooser with Logging {
+  registry: MetricsRegistry = new MetricsRegistryMap,
+
+  /**
+   * Stores the mapping from SystemStreamPartition to number of tasks consuming from it.
+   * This will be used by the MessageChooser to customize the envelope selection.
+   */
+  sspToTasksCount: Map[SystemStreamPartition, Int] = Map()) extends MessageChooser with Logging {
```
sborya: I think it is not a very good style to use default values in the arugments. If you don't really need them - please remove them.

samza-core/src/test/scala/org/apache/samza/system/chooser/TestBootstrappingChooser.scala
```diff
@@ -193,11 +193,35 @@ class TestBootstrappingChooser(getChooser: (MessageChooser, Map[SystemStream, Sy
...
+
+    // Consume a partition from SystemStream1
+    chooser.register(envelope1.getSystemStreamPartition, "123")
+
+    assertEquals(Map(envelope2.getSystemStreamPartition -> 1), chooser.laggingSystemStreamPartitions)
+
+    // Consume a partition from SystemStream2
+    chooser.register(envelope2.getSystemStreamPartition, "321")
+
+    assertEquals(Map(), chooser.laggingSystemStreamPartitions)
```
sborya: please also test chooser.systemStreamLagCounts.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -354,7 +355,19 @@ object SamzaContainer extends Logging {
...
+     * by all the tasks that are assigned to it. This will help in enforcing the condition.
+     */
+    val sspToTaskCount: mutable.Map[SystemStreamPartition, Int] = mutable.Map[SystemStreamPartition, Int]()
+    for(taskModel <- containerModel.getTasks.values())  {
+      for (systemStreamPartition <- taskModel.getSystemStreamPartitions) {
+        sspToTaskCount.put(systemStreamPartition, sspToTaskCount.getOrElse(systemStreamPartition, 0) + 1)
+      }
+    }
+
+    val chooser = DefaultChooser(inputStreamMetadata, chooserFactory, config, samzaContainerMetrics.registry, sspToTaskCount.toMap)
```
jmakes: Something feels wrong here. Why should a chooser be concerned with which or how many tasks consume it?

I'll need to dig into the code to understand more, but I suspect that all the code for broadcast streams is on the task side, not the chooser side. If so, then we shouldn't be updating the chooser with broadcast logic.

shanthoosh: Had an offline discussion. The consensus is passing in TaskCount map into the chooser will work, but it breaks the abstraction barrier and making the chooser aware of the number of downstream tasks that are consuming from it in unnecessary(violates the separation of concerns). 

The solution we'd agreed upon is retaining the lowest offset for each of the registered SystemStreamPartition and calling checkOffset for each SSP with lowest offset in BootstrappingChooser.start .

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -120,15 +129,12 @@ class BootstrappingChooser(
...
-    // If the offset we're starting to consume from is the same as the upcoming 
-    // offset for this system stream partition, then we've already read all
-    // messages in the stream, and we're at head for this system stream 
-    // partition.
-    checkOffset(systemStreamPartition, offset, OffsetType.UPCOMING)
-
     wrapped.register(systemStreamPartition, offset)
 
-    registeredSystemStreamPartitions += systemStreamPartition
+    if (!registeredSystemStreamPartitions.contains(systemStreamPartition)
```
jmakes: This logic should be explained with comments.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -120,15 +129,12 @@ class BootstrappingChooser(
...
+        || (offset != null && registeredSystemStreamPartitions(systemStreamPartition).toLong > offset.toLong)) {
```
jmakes: I don't think it's safe to assume Long offsets. SystemAdmin.offsetComparator should be used.

shanthoosh: Switched to use SystemAdmin.offsetComparator.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
shanthoosh: Explained the reasoning in a comment and added a JIRA to comment as well. Please check if it would suffice.

jmakes:
Looks like there were build errors with the latest commit. Please resolve.

shanthoosh:
@jmakes : 

Fixed the broken tests. 

I've added Map(SystemName -> SystemAdmins) in the classes DefaultChooser and BootstrappingChooser. In tests, those values were not set correctly. Hence the failures.

shanthoosh:
The failure is due to a Flaky Integration Test : TestStatefulTask. It passes in my local box. It does a kill -9 on a job in the test. Lack of permissions leads to failure.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -120,15 +133,25 @@ class BootstrappingChooser(
...
-    // messages in the stream, and we're at head for this system stream 
-    // partition.
-    checkOffset(systemStreamPartition, offset, OffsetType.UPCOMING)
-
     wrapped.register(systemStreamPartition, offset)
 
-    registeredSystemStreamPartitions += systemStreamPartition
+    val systemStream = systemStreamPartition.getSystem
+    val systemAdmin: SystemAdmin = systemAdmins.getOrElse(systemStream,
+                                                          throw new SamzaException("SystemAdmin is undefined for SystemStream: %s" format systemStream))
```
shanthoosh: Fixed.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -93,7 +98,7 @@ class BootstrappingChooser(
   /**
    * Store all the systemStreamPartitions registered
    */
-  var registeredSystemStreamPartitions = Set[SystemStreamPartition]()
+  var registeredSystemStreamPartitions: mutable.Map[SystemStreamPartition, String] = mutable.Map[SystemStreamPartition, String]()
```
shanthoosh: Removed the type annotation.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -120,15 +133,25 @@ class BootstrappingChooser(
...
-    // If the offset we're starting to consume from is the same as the upcoming 
-    // offset for this system stream partition, then we've already read all
-    // messages in the stream, and we're at head for this system stream 
-    // partition.
-    checkOffset(systemStreamPartition, offset, OffsetType.UPCOMING)
-
     wrapped.register(systemStreamPartition, offset)
 
-    registeredSystemStreamPartitions += systemStreamPartition
+    val systemStream = systemStreamPartition.getSystem
```
shanthoosh: It's `system`. Fixed.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -28,10 +27,12 @@ import org.apache.samza.metrics.MetricsHelper
 import org.apache.samza.metrics.MetricsRegistryMap
 import org.apache.samza.metrics.MetricsRegistry
 import org.apache.samza.system.SystemStreamMetadata
+
 import scala.collection.JavaConversions._
-import org.apache.samza.SamzaException
 import org.apache.samza.system.SystemStreamMetadata.OffsetType
 
+import scala.collection.mutable
```
shanthoosh: Fixed

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -120,15 +133,25 @@ class BootstrappingChooser(
...
+                                                          throw new SamzaException("SystemAdmin is undefined for SystemStream: %s" format systemStream))
+    /**
+     * SAMZA-1100: When a input SystemStream is consumed as both bootstrap and broadcast
+     * BootstrappingChooser should record the lowest offset for each registered SystemStreamPartition.
+     * When multiple tasks processing a SystemStreamPartition runs within a container
+     * and share the same chooser, then the lowest offset should be chosen as starting offset.
+     */
+    if (!registeredSystemStreamPartitions.contains(systemStreamPartition)) {
+      registeredSystemStreamPartitions += systemStreamPartition -> offset
+    } else if (offset != null) {
```
shanthoosh: Last checkpoint for a task could be null(when it's added for the first time)  or if there is tasks addition.

shanthoosh:
Added a unit test that contains multiple bootstrappingChooser.register calls for same SSP with different offset(mimicking the same scenario that is the source of this issue)

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -120,15 +136,24 @@ class BootstrappingChooser(
...
+      val comparatorResult: Integer = systemAdmin.offsetComparator(registeredSystemStreamPartitions(systemStreamPartition), offset)
```
nickpan47: So, what happens to the systems w/ non-comparable offsets? I would recommend to capture the UnsupportedOperationException here and wrap it w/ more meaningful logging info under this context.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -93,7 +101,7 @@ class BootstrappingChooser(
   /**
    * Store all the systemStreamPartitions registered
    */
-  var registeredSystemStreamPartitions = Set[SystemStreamPartition]()
+  var registeredSystemStreamPartitions = mutable.Map[SystemStreamPartition, String]()
```
shanthoosh: Fixed. Updated the javadoc description.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
shanthoosh: SystemAdmin.offsetComparator returns null(as mentioned in javadoc of the interface), when the offsets are incomparable. In that case, this patch already chooses the offset provided by the first task for a ssp(lineNo: 153). 

Adding a log message at warn level. I don't want the chooser to fail & throw up the exception.

vjagadish1989:
Merged, and submitted thanks!

PR #69, opened 2017-03-01
Samza 1114

renato2099:
When running the release candidate I noticed that the bin/check-all.sh silently failed.
Yeah my env was not properly set up but I got no clue until I double checked. This should make it easier for users to figure it out.


bin/check-all.sh
```diff
@@ -60,9 +60,9 @@ fi
 if [ -e "${base_dir}/gradlew" ]; then
   gradle_file="${base_dir}/gradlew"
 else
-  gradle_base_file=$(which gradle)
+  gradle_base_file=$(which gradle) && chk_gradle=$? || chk_gradle=$?
```
vjagadish1989: `chk_gradle=$?`

Is this is just a way of checking `chk_gradle` when `set -e` is set? Otherwise it would cause the shell to exit if any subcommand or pipeline returns a non-zero status?

Wondering what's the need for this additional check? It seems that `-f`  should already do the trick? (if `gradle_base_file` does not exist)

renato2099: hey @vjagadish1989  thanks for checking this out. The problem is that in my case the $(which gradle) command failed, and then the script exits without prompting anything, not even getting to the <-f> check.
So this line puts the result of executing $(which gradle) into the variable chk_gradle so we can check if the $(which gradle) didn't fail.

vjagadish1989: Firstly, Thanks for the patch @renato2099 👍 
>> The `which` gradle failed

QQ: Did the command fail because `gradle` was not installed or `which` was not installed? For the former, wouldn't the `-f` in the current implementation would catch it? 

I wrote a small test to verify this:

Check with valid gradle path:
```
sh-3.2$ gradle_base_file=$(which gradle)
sh-3.2$ if [ -f "$gradle_base_file" ]; then    echo "gradle_base_file $gradle_base_file exists."; else    echo "gradle_base_file $gradle_base_file does not exist."; fi
gradle_base_file /usr/local/bin/gradle exists.
```

Check with invalid gradle path:
```
sh-3.2$ gradle_base_file=$(which gradle-notfound)
sh-3.2$ if [ -f "$gradle_base_file" ]; then    echo "gradle_base_file $gradle_base_file exists."; else    echo "gradle_base_file $gradle_base_file does not exist."; fi
gradle_base_file  does not exist.
```

renato2099: Hey @vjagadish1989 thanks again for checking this out with me.
So on my env (Debian Lenny) the script exited when executing `which gradle` and it didn't make it to the `if` statement. `which` was installed because I didn't get a `command not found` error. 
I have also tried to reproduce the issue on different servers now and everything works as expected i.e. `which` doesn't crash. I guess it was just something wrong with my local env.

PR #70, opened 2017-03-01
SAMZA-1115:Job coordinator

sborya:
Job coordinator to work with the ZooKeeper.


samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -105,30 +97,24 @@ public void handleChildChange(String parentPath, List<String> currentChildren) t
...
-      LOG.info("list of children to compare against = " + parentPath + ":" + sb.toString());
 
+      LOG.info("list of children in the barrier = " + parentPath + ":" + Arrays.toString(currentChildren.toArray()));
+      LOG.info("list of children to compare against = " + parentPath + ":" + Arrays.toString(names.toArray()));
 
       // check if all the names are in
-      for (String n : names) {
-        if (!currentChildren.contains(n)) {
-          LOG.info("node " + n + " is still not in the list ");
+      for (String name : names) {
```
fredji97: We may consider using Apache CollectionUtil.containsAll(currentChildren, names) for allIn

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -0,0 +1,219 @@
...
+   */
+  private void generateNewJobModel() {
+    // get the current list of processors
+    List<String> currentProcessors = zkUtils.getSortedActiveProcessors();
+
+    // get the current version
+    String currentJMVersion  = zkUtils.getJobModelVersion();
+    String nextJMVersion;
+    if (currentJMVersion == null)
+      nextJMVersion = "1";
```
fredji97: I think we may want to add log.info in this case.

samza-core/src/main/java/org/apache/samza/config/JavaJobConfig.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  public String getJobName() {
+    if (!containsKey(JOB_NAME)) {
+      throw new ConfigException("Missing " + JOB_NAME + " config!");
+    }
+    return get(JOB_NAME);
+  }
+
+  public String getJobName(String defaultValue) {
```
sborya: removed it.

samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
```diff
@@ -32,12 +32,13 @@
    * @param version - for which the barrier is started.
    * @param processorsNames - list of processors available at the time of the JobModel generation.
    */
-  void startBarrier(String version,  List<String> processorsNames);
+  void start(String version, List<String> processorsNames);
```
sborya: we can't change it for wait(). It will override Object.wait().

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: done.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java

samza-core/src/main/java/org/apache/samza/config/JavaJobConfig.java
```diff
@@ -0,0 +1,41 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.config;
+
+public class JavaJobConfig extends MapConfig {
```
xinyuiscool: Just curious, seems all the configs are already defined in scala. Any reason to introduce this?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -0,0 +1,221 @@
...
+    log.info("ZkJobCoordinator::onProcessorChange - Processors changed! List: " + Arrays.toString(processorIds.toArray()));
+    generateNewJobModel();
+  }
+
+  @Override
+  public void onNewJobModelAvailable(final String version) {
+    newJobModelVersion = version;
+    log.info("pid=" + processorId + "new JobModel available");
+    // stop current work
+    containerController.stopContainer();
```
xinyuiscool: I looked at the stopContainer code again, seems it only does stop the RunLoop. Seems to me all the dependencies used by the container are not properly shutdown. E.g. the kafka system consumer/producer will not be shut down. The next time controller.startContainer() will create the container with all these dependencies again. The Kafka producers/consumers will be created again without closing the previous ones, right?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -0,0 +1,221 @@
...
+    log.info("pid=" + processorId + "new JobModel available.Container stopped.");
+    // get the new job model
+    newJobModel = zkUtils.getJobModel(version);
+    log.info("pid=" + processorId + "new JobModel available. ver=" + version + "; jm = " + newJobModel);
+
+    String currentPath = zkUtils.getEphemeralPath();
+    String zkProcessorId = keyBuilder.parseIdFromPath(currentPath);
+
+    // update ZK and wait for all the processors to get this new version
+    barrier.waitForBarrier(version, String.valueOf(zkProcessorId), new Runnable() {
```
xinyuiscool: Do we need some timeout mechanism here? I commented in the previous barrier PR too. If a processor failed during previous stopContainer, then the Barrier will not be reached in the future. I think this can happen in real jobs.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: may be, it is not a part of this patch. 
I've opened a ticket SAMZA-1127 to verify/fix this.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: this is addressed in SAMZA-1124.

samza-core/src/main/java/org/apache/samza/config/JavaJobConfig.java
sborya: Hmm, interesting. I didn't add this file myself, so I assumed it is an existing one, and I just modified it. I will remove it.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -0,0 +1,221 @@
...
+    this.containerController = containerController;
+    this.zkController = new ZkControllerImpl(String.valueOf(processorId), zkUtils, debounceTimer, this);
+    this.config = config;
+
+
+    barrier = new ZkBarrierForVersionUpgrade(zkUtils, debounceTimer);
+    streamMetadataCache = initJobModelGeneration();
+  }
+
+  private StreamMetadataCache initJobModelGeneration() {
```
navina: method name is misleading as it says "initJobModelGeneration" and all it does is to populate the StreamMetadataCache.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -0,0 +1,221 @@
...
+    // model generation - NEEDS TO BE REVIEWED
+    JavaSystemConfig systemConfig = new JavaSystemConfig(this.config);
```
navina: Why should the StreamMetadataCache be initialized when the jobCoordinator is created? Isn't it sufficient to do so only in the leader ??

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -0,0 +1,221 @@
...
+      }
+      SystemFactory systemFactory = Util.getObj(systemFactoryClassName);
+      systemAdmins.put(systemName, systemFactory.getAdmin(systemName, this.config));
+    }
+
+    return new StreamMetadataCache(Util.<String, SystemAdmin>javaMapAsScalaMap(systemAdmins), 5000, SystemClock.instance());
+  }
+
+  @Override
+  public void start() {
```
navina: ZkJobCoordinator's lifecycle interfaces are directly tied to the StreamProcessor lifecycle. In such a world, isn't it possible for the user to invoke start() multiple times? For that matter, how are we guaranteeing that the order of invocation of start/stop/awaitStart does not impact the controller behavior.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -0,0 +1,221 @@
...
+
+  @Override
+  public void stop() {
+    zkController.stop();
+  }
+
+  @Override
+  public boolean awaitStart(long timeoutMs)
+      throws InterruptedException {
+    return containerController.awaitStart(timeoutMs);
```
navina: The awaitStart behavior is also confusing. At this point, if containerController.startContainer has not been called, then awaitStart will fail since the container has not yet been initialized. We need to fix this by de-coupling the Streamprocessor API from the JobCoordinator API. I created [SAMZA-1147](https://issues.apache.org/jira/browse/SAMZA-1147) to fix this. But even without it, this statement will fail if container has not yet been initialized.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -0,0 +1,221 @@
...
+    StringBuilder sb = new StringBuilder();
+    List<Integer> containerIds = new ArrayList<>();
+    for (String processor : currentProcessors) {
+      String zkProcessorId = keyBuilder.parseIdFromPath(processor);
+      sb.append(zkProcessorId).append(",");
+      containerIds.add(Integer.valueOf(zkProcessorId));
+    }
+    log.info("generate new job model: processorsIds: " + sb.toString());
+
+    jobModelManager = JobModelManager$.MODULE$.getJobCoordinator(this.config, null, null, streamMetadataCache, null,
```
navina: Based on this statement, it seems like we are not supporting localitymanager and changelog manager. So, are we explicitly saying we are not going to support stateful jobs in this version? 
Thanks to @santhoshvenkat1988 for pointing it out.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -64,12 +64,13 @@
   private volatile String ephemeralPath = null;
   private final ZkKeyBuilder keyBuilder;
   private final int connectionTimeoutMs;
-  private final String processorId = "TO BE PASSED IN THE CONSTRUCTOR"; //TODO
+  private final String processorId;
 
-  public ZkUtils(ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs) {
+  public ZkUtils(String processorId, ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs) {
```
navina: I prefer to keep this class as a util class for interacting with Zk and hence, avoid any state associated with it. If processorId is only used in log lines, I would rather keep it out.

sborya:
submitted as a part of another PR.

PR #73, opened 2017-03-02
SAMZA-1117 set up certificate signing request in the context for resource localization

fredji97:
Tested manually with the following samza configuration:
"csr.ssl.request.path" => "certfs://mytestserver.com:10180/request_service_certificate"

and the localization for the certificate signing request happens as expected by checking the log.


fredji97:
closing this one as we are introducing a new approach to for universal resource localization.
https://github.com/apache/samza/pull/90

PR #74, opened 2017-03-03
SAMZA-1096: StreamSpec constructors in the ExecutionEnvironments

jmakes:


docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -254,7 +261,7 @@
                 </tr>
 
                 <tr>
-                    <td class="property" id="job-systemstreampartition-matcher-class">job.systemstreampartition.matcher.class</td>
+                    <td class="property" id="job-systemstreampartition-matcher-class">job.systemstreampartition.<br>matcher.class</td>
```
vjagadish1989: do we need a `br` here? Does it look better on the eyes on rendering the html?

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -822,6 +828,40 @@
...
+                    <td class="description">
+                        The <span class="system">system-name</span> of the system on which this stream will be accessed.
+                        This property binds the stream to one of the systems defined with the property
+                        systems.<span class="system">system-name</span>.samza.factory. <br>
+                        If this property isn't specified, it is inherited from job.default.system.
+                    </td>
+                </tr>
+
+                <tr>
+                    <td class="property" id="streams-physical-name">streams.<span class="stream">stream-id</span>.samza.physical.name</td>
```
vjagadish1989: - Don't we already have properties of the following format - `systems.system-name.
streams.stream-name.samza.key.serde` ? (also, offset reset default settings are another example). Here, the property name seems to be `streams.stream-name.*`.?  ie, the specification of system and stream is interchanged. I suspect that's intentional? Wondering if we can unify the two? (or are the two meant to co-exist?)

- Also, if I want to set `system` level properties, should I set all of those properties for each of the `streams` in the system?

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1789,37 +1829,37 @@
...
                 </tr>
                 <tr>
-                  <td class="property" id="hdfs-write-batch-size-bytes">systems.*.producer.hdfs.write.batch.size.bytes</td>
+                  <td class="property" id="hdfs-write-batch-size-bytes">systems.<span class="system">system-name</span>.<br>.producer.hdfs.write.batch.size.bytes</td>
                   <td class="default">268435456</td>
                   <td class="description">The number of bytes of outgoing messages to write to each HDFS output file before cutting a new file. Defaults to 256MB if not set.</td>
                 </tr>
                 <tr>
-                    <td class="property" id="hdfs-write-batch-size-records">systems.*.producer.hdfs.write.batch.size.records</td>
+                    <td class="property" id="hdfs-write-batch-size-records">systems.<span class="system">system-name</span>.<br>.producer.hdfs.write.batch.size.records</td>
```
vjagadish1989: thanks :)

samza-api/src/main/java/org/apache/samza/system/StreamProvider.java
```diff
@@ -0,0 +1,78 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.system;
+
+/**
+ * Describes the common interface for classes that construct instances of {@link StreamSpec}.
+ */
+public interface StreamProvider {
```
xinyuiscool: From the previous discussion, this API seems not needed for now. We will use ExecutionEnv directly to create the streams. Add seems ExecutionEnvironment already has these methods now. Maybe a cleanup?

samza-core/src/main/java/org/apache/samza/system/AbstractExecutionEnvironment.java
```diff
@@ -0,0 +1,63 @@
...
+    if (config == null) {
+      throw new NullPointerException();
+    }
+
+    this.config = config;
+  }
+
+  @Override
+  public StreamSpec streamFromConfig(String streamId) {
+    StreamConfig streamConfig = new StreamConfig(config);
```
xinyuiscool: seems the code can be simplified to call the next streamFromConfig(streamId, physicalName), e.g.;
StreamSpec streamFromConfig(string streamId) {
  SteamConfig streamConfig = ...
  String physicalName = ...
  return streamFromConfig(streamId, physicalName);
}

samza-api/src/main/java/org/apache/samza/system/ExecutionEnvironment.java
```diff
@@ -70,4 +73,57 @@ static ExecutionEnvironment fromConfig(Config config) {
...
+   * <ul>
+   *   <li>samza.system - The name of the System on which this stream will be used. If this property isn't defined
+   *                      the stream will be associated with the System defined in {@code job.default.system}</li>
+   * </ul>
+   *
+   * @param streamId      The logical identifier for the stream in Samza.
+   * @param physicalName  The system-specific name for this stream. It could be a file URN, topic name, or other identifer.
+   * @return              The {@link StreamSpec} instance.
+   */
+  StreamSpec streamFromConfig(String streamId, String physicalName);
```
vjagadish1989: Can't we unify both the `streamId` and `physicalName`? The API can instead be:

`StreamSpec streamFromConfig(String streamId)`

That way, `streamId` is the means to look up the config, and we don't treat `physicalName` as a special property?

samza-api/src/main/java/org/apache/samza/system/StreamProvider.java
```diff
@@ -0,0 +1,78 @@
...
+   * <ul>
+   *   <li>samza.system - The name of the System on which this stream will be used. If this property isn't defined
+   *                      the stream will be associated with the System defined in {@code job.default.system}</li>
+   * </ul>
+   *
+   * @param streamId      The logical identifier for the stream in Samza.
+   * @param physicalName  The system-specific name for this stream. It could be a file URN, topic name, or other identifer.
+   * @return              The {@link StreamSpec} instance.
+   */
+  StreamSpec streamFromConfig(String streamId, String physicalName);
```
vjagadish1989: Can't we simply use `streamId`?

vjagadish1989 (commented):
First round of review and feedback. thank you for the contribution.

docs/learn/documentation/versioned/jobs/configuration-table.html
jmakes: It was a convention for readability that was abandoned. Since the property and default columns are set to no-wrap, large values increase the table size and push the description off the page. When you see the result, you'll like it. :-)

docs/learn/documentation/versioned/jobs/configuration-table.html
jmakes: Yes it's intentional and in a separate task we'll go deprecate the SystemStream approach to configurations. That will be a large effort, so I didn't pursue it here. For now, the new streams.streamName.* approach takes precedence. When we remove the legacy stream configs, the documentation for those properties will be moved here with the other streams properties.

We're doing this because the "stream" in SystemStream could be any physical name like an HDFS file URN. Obviously that will be hard to parse in a config file that has systems.sysName.streams.streamName.* 
So that path is already broken for streams that have special characters. StreamId fixes that by decoupling the logical identifier used by samza from the physical/remote identifier used by the system. That way we can enforce no special characters for the streamId so we can use it as a key in the configs.

System level properties will still be set at the system level for now. i.e. systems.systemName.*. In the future, I'd like to define template streams which better serve this purpose. Something like 
templates.streams.myAvroTemplate.samza.msg.serde=avro
streams.myStream.samza.template=myAvroTemplate

Then you could define a bunch of properties on the template and apply the template as a set of defaults to each stream. Just a rough idea for now.

samza-api/src/main/java/org/apache/samza/system/ExecutionEnvironment.java
jmakes: StreamId will not work as the physicalName when the physicalName has special characters (see the hdfs example above). 

But we could hide this signature for now and if users need to set the physicalName, they can do it in config and if it's not in the config, we'll assume physicalName==streamId.

This will work for kafka, but HDFS users will likely end up setting the config. 

The only risk in hiding this signature is that users will associate streamID and physicalName when they serve different purposes (samza identifier vs external identifier)

samza-api/src/main/java/org/apache/samza/system/StreamProvider.java
jmakes: Yes, thanks for catching that!
I moved the methods but apparently didn't remove this class from the git index.

samza-core/src/main/java/org/apache/samza/system/AbstractExecutionEnvironment.java
jmakes: Totally. I thought I already did that, but apparently not. Great catch! Thanks

samza-api/src/main/java/org/apache/samza/system/StreamProvider.java
jmakes: I think this is addressed above, except I should note that this interface should have been deleted before I submitted the PR.

samza-api/src/main/java/org/apache/samza/system/ExecutionEnvironment.java
vjagadish1989: Thanks @jmakes for explaining.
1. Do we see a case where there is not a one-one correspondence between a `streamID` and the `physicalName`? (they may not be the same value, isn't there's always a 1-1 mapping between the two) ?
2. If the answer to the above is `No`, would it still make sense to ask for both in the API?
`StreamSpec streamFromConfig(String streamId, String physicalName)`.  ( I believe we can get away with just `streamID` in the API then)

jmakes: Well, an application with 1 streamId could be deployed in multiple environments with different configs pointing to a different physicalName for each environment.
 
That's a 1:N relationship in terms of the values. But for any given config each stream has 1 streamId and 1 physicalName. So 1:1 in terms of the properties themselves.

A file URN is not a valid streamId because it would prevent us from properly parsing the config. But that same URN is totally valid as a physicalName. So I would argue that physicalName is a necessary attribute.

The real question here is where the physicalName should come from:
1. Exclusively from config. (I felt this would be tedious for users)
2. Same as streamId but overridable via config (this is the what the first signature of streamFromConfig() does)
3. Specified from code (this is what the other signatures do)

I think Prateek convinced me that we can start with signature 1 and expand to the others if necessary. Stay tuned for the next patch, which will do this.

docs/learn/documentation/versioned/jobs/configuration-table.html
navina: Thanks for explaining @jmakes ! I was going to ask the same question. 
I understand that it is a significant overhaul of configs to change to the new format. But having the intermediate period where we support both systems.<streamname> amd "streams.<streamname>" is going to be confusing for out users.  I am still not clear how we can cleanly transition over from the legacy configs to the new configs. Can we please document the proposed change and how the transition is going to happen? 

Additionally, the config template approach seems to increase the burden on the user to understand the config redirection. I think it arises due to this flattened config format of simply string based key-value pair. Wonder if there is a better way to do it. 

Quick question orthogonal to what we are discussing here: If the separation of logical and physical identifier for stream is aiding the HDFS use-case, how are they doing it now? Right now, does HDFS not support stream level properties?

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -156,13 +156,20 @@
...
                 </tr>
-                <tr>
                     <td class="property" id="job-coordinator-system">job.coordinator.system</td>
                     <td class="default"></td>
                     <td class="description">
                         <strong>Required:</strong> The <span class="system">system-name</span> to use for creating and maintaining the <a href="../container/coordinator-stream.html">Coordinator Stream</a>.
                     </td>
                 </tr>
+                <tr>
+                    <td class="property" id="job-default-system">job.default.system</td>
```
jmakes: So, I have this longer term dream that most samza users will not have to specify a bunch of .system properties. I see two main use cases:

A) Job creates all streams, IO and samza metadata on the same system. This job could just specify the job.default.system and ignore the others

B) Job creates all IO streams on 1 system, metadata streams on another system. This job would configure
job.metadata.system for all the metadata streams, like we do at LI
job.default.system for all other systems
Override on a case-by-case basis by setting 
streams.x.samza.system

For case A, I think the current naming would make more sense, because it's the default for all streams.

My intent was to update the doc to reflect that when I make that change. But if it's misleading now, I could do it now. Thoughts?

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -822,6 +828,40 @@
                 </tr>
 
                 <tr>
+                    <th colspan="3" class="section" id="streams"><a href="../container/streams.html">Streams</a></th>
+                </tr>
+
+                <tr>
+                    <td class="property" id="streams-system">streams.<span class="stream">stream-id</span>.samza.system</td>
```
jmakes: In the past, streamName was used as both:
1. a logical identifier in samza. e.g. to look up stream properties in the config
2. a physical identifier in the System e.g. as the topic name for Kafka

StreamId and physicalName represent the separation of those two roles and provide the following benefits:
1. StreamId is now a single unique identifier for a stream. A streams ID is no longer coupled to a System, so it's more portable and less verbose.
2. A streamID does not allow special characters so we can guarantee config parsing, etc. Whereas physicalName does, and that enables physical identifiers that have special characters like an HDFS file URN

jmakes:
@navina github isn't letting me address your last comment inline so quoting here:

> Quick question orthogonal to what we are discussing here: If the separation of logical and physical identifier for stream is aiding the HDFS use-case, how are they doing it now? Right now, does HDFS not support stream level properties?
A combination of luck and simple use cases :-)
The luck part is that we happen to parse task.inputs such that we get the system name first and then assume everything after the "." is the stream name. But this wouldn't work if there were commas or spaces in the stream name.
The other part is that I haven't seen an HDFS case that sets any stream specific properties. Thus they avoid the config pattern 
systems.file-system.streams.file://my.dir/my.file.avro.samza.msg.serde
which is harder to parse. And it will get even worse when we start supporting any properties s.t. they don't all have a samza.* suffix.

docs/learn/documentation/versioned/jobs/configuration-table.html
nickpan47: Let's use job.default.system for yet another reason: job.io.system also implicitly forces input and output systems into one, which is different than saying: if you don't specify, input/output system would be in the default.

I have another question related to scoping of the configuration: is this at job-level or at the whole application-level? Right now, I saw three levels of configuration scopes: application ==> job ==> task. Curious whether this "default IO system" should be default for the whole application or a specific job?

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -683,7 +689,7 @@
                 </tr>
 
                 <tr>
-                    <th colspan="3" class="section" id="streams"><a href="../container/streams.html">Systems (input and output streams)</a></th>
+                    <th colspan="3" class="section" id="systems"><a href="../container/streams.html">Systems (input and output streams)</a></th>
```
nickpan47: Seems to me it is better to change the titles to Streams instead of the id to "systems". Anyways, we are talking about "Streams" here and the link is also to streams.html.

docs/learn/documentation/versioned/jobs/configuration-table.html
navina: I don't particularly agree with case A because we should not ignore configs. It makes transitioning out of the old way harder. It also makes the config model more restrictive (compared to what we have today) because now I can't have system-scoped properties defined. Is my interpretation correct about not supporting system-scoped configs ?

With Case B, I like the idea of unifying configs for metadata streams under the same umbrella. However, we should still allow overrides at system-scope. Let's say Kafka introduces exactly-once producer and we want to leverage that for all out metadata systems. That implies, I have to set 2 or more properties at stream-scope (streams.<coordinator-stream-name>.producer.<property>, streams.<checkpoint-stream-name>.producer.<property>, instead of one system-scoped property for kafka. Also, metadata stream names are generated by samza during runtime. So letting expecting the user to statically define it complicates our model, doesn't it? 

Please correct me if I have understood your proposed solutions incorrectly :)

I saw you that created JIRAs for follow-up on changes. However, I think we need to be on the same page before executing the jiras :)

docs/learn/documentation/versioned/jobs/configuration-table.html
nickpan47: Hi, @navina and @vjagadish1989 the purpose of the logic stream-id has its extended meaning even in code. Currently, the stream name and system name are physical concepts that changes when the runtime environment changes (we have seen this when one user sending output to local.PageViewEventByMemberId fail to deploy to production since local cluster is not there and the input has to be queuing.PageViewEventByMemberId, even in the code). Using logic stream-id and make the physical system name and stream name a config value instead of part of the config variable name helps to make the code portable: in user code, they only need to refer to "PageViewEventByMemberId", which can be a Kafka topic in queuing by config, or a HDFS file by config. User won't need to re-compile there code to port over their output between different systems.

Co-existence of systems.&lt;xyz&gt;.streams.&lt;abc&gt;. and streams.&lt;abcId&gt;.system... is confusing. We might want to mark the whole section of systems.&lt;xyz&gt;.streams.&lt;abc&gt; as deprecated, and re-write w/ the new set of configuration to avoid this confusion.

samza-api/src/main/java/org/apache/samza/system/ExecutionEnvironment.java
```diff
@@ -51,8 +52,10 @@ static ExecutionEnvironment getLocalEnvironment(Config config) {
    */
   static ExecutionEnvironment fromConfig(Config config) {
     try {
-      if (ExecutionEnvironment.class.isAssignableFrom(Class.forName(config.get(ENVIRONMENT_CONFIG, DEFAULT_ENVIRONMENT_CLASS)))) {
-        return (ExecutionEnvironment) Class.forName(config.get(ENVIRONMENT_CONFIG, DEFAULT_ENVIRONMENT_CLASS)).newInstance();
+      Class<?> environmentClass = Class.forName(config.get(ENVIRONMENT_CONFIG, DEFAULT_ENVIRONMENT_CLASS));
+      if (ExecutionEnvironment.class.isAssignableFrom(environmentClass)) {
+        Constructor<?> constructor = environmentClass.getConstructor(Config.class); // *sigh*
```
nickpan47: Do we have some document stating that all ExecutionEnvironment implementation should have a constructor take Config as parameter? We probably should.

navina:
@jmakes Yeah. Something is wrong with github. 

>  But this wouldn't work if there were commas or spaces in the stream name.

Commas are treated as list separators in some of configs. i would guess its the same way it is treated in most "properties" based config definitions. I don't think spaces make sense in any config definition :)

>Thus they avoid the config pattern systems.file-system.streams.file://my.dir/my.file.avro.samza.msg.serde

Yeah. It is one thing for us to say that they can avoid. A whole other thing to say we are blocking it actively for their sake. Right now, they don't actually call out that they cannot configure stream-scope properties (may be they should :D)  We should be facing the same problem with databus and the only reason why it works is because of how we parse it (anything after a .) . 

While rolling out this new way of configuration, is it going to be backward compatible with the current config definition? That is why I asked the question.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -43,6 +43,7 @@ object JobConfig {
   val SAMZA_FWK_PATH = "samza.fwk.path"
   val SAMZA_FWK_VERSION = "samza.fwk.version"
   val JOB_COORDINATOR_SYSTEM = "job.coordinator.system"
+  val JOB_DEFAULT_SYSTEM = "job.default.system"
```
nickpan47: The same scoping question: app or job?

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
```diff
@@ -19,64 +19,187 @@
...
+  val CONSUMER_RESET_OFFSET =   SAMZA_PROPERTY + "reset.offset"
+  val CONSUMER_OFFSET_DEFAULT = SAMZA_PROPERTY + "offset.default"
+
+  val STREAMS_PREFIX = "streams."
+  val STREAM_ID_PREFIX = STREAMS_PREFIX + "%s."
+  val SYSTEM_FOR_STREAM_ID = STREAM_ID_PREFIX + SYSTEM
+  val PHYSICAL_NAME_FOR_STREAM_ID = STREAM_ID_PREFIX + PHYSICAL_NAME
+  val SAMZA_STREAM_PROPERTY_PREFIX = STREAM_ID_PREFIX + SAMZA_PROPERTY
+
   val STREAM_PREFIX = "systems.%s.streams.%s."
```
nickpan47: nit: I prefer to move this to some inner class of OldStreamConfig s.t. we can mark it deprecated and delete those later.

docs/learn/documentation/versioned/jobs/configuration-table.html
navina: @nickpan47 
>  job.io.system also implicitly forces input and output systems into one

I agree that it should force the input and output into one. But calling it "default" doesn't say that it will not be used for metadata system. From the documentation here, I infer that this configuration will be used for "_any input or output streams for which the system is not explicitly configured_". We are defining a separate "default" system for metadata topics, are we? 

Also, are we going to use "job.coordinator.system" for all samza metadata system or just coordinator and checkpoint?

jmakes: @nickpan47 
I couldn't find an example of an application-scoped config. Is there one already existing, or in the works?
I don't mind making it application scope but prefer to keep parity with job.coordinator.system, etc. Feels like they should all be the same scope?

Perhaps there's a separate task to rescope configs?

docs/learn/documentation/versioned/jobs/configuration-table.html
jmakes: Good catch. That was a mistake. 

I feel like we need a different doc to talk specifically about the system abstraction and this config section should link to that. I could file a JIRA for that if you agree.

docs/learn/documentation/versioned/jobs/configuration-table.html
navina: @nickpan47 
> we have seen this when one user sending output to local.PageViewEventByMemberId fail to deploy to production since local cluster is not there and the input has to be queuing.PageViewEventByMemberId, even in the code)

I don't agree to this use-case to support your point on code portability. The user can always define app-level configs to indicate inputsystem = "queuing" and outputsystem = "local". They can refer to this in their code without requiring a re-compilation of code. It is just a choice by the user to hard-code the system and/or stream name in code. 

> Co-existence of systems.<xyz>.streams.<abc>. and streams.<abcId>.system... is confusing

+1 on this. I still have question unanswered about whether we are restricting system-scoped properties in the new model. in case @jmakes is OOO, you can address my questions above?

samza-api/src/main/java/org/apache/samza/system/ExecutionEnvironment.java
jmakes: Sure I will add it

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
jmakes: Same answer :-)

docs/learn/documentation/versioned/jobs/configuration-table.html
jmakes: > Co-existence of systems.<xyz>.streams.<abc>. and streams.<abcId>.system... is confusing. We might want to mark the whole section of systems.<xyz>.streams.<abc> as deprecated, and re-write w/ the new set of configuration to avoid this confusion.

That's not a bad idea, but I likely won't be able to get that done and reviewed before EOD Monday. So some options:
1. I checkin this patch to the integration branch as is and file a JIRA to do the rewriter before merging into master. This could be done by someone else if I'm not back in time, or I could take it up when I'm back.
2. Start working on it and see how far I get by EOD Monday. Update this PR and have someone else take over if need be.

docs/learn/documentation/versioned/jobs/configuration-table.html
jmakes: @navina For A, I misspoke. I didn't mean the job would ignore the old configs but that the user would. 

I don't understand your example for system scope. To use the exact once producer for metadata and the normal producer for others, one would do this. 

systems.kafka-exact.samza.factory=org.apache.pkg.KafkaExactOnceSystemFactory
systems.kafka-at-least.samza.factory=org.apache.pkg.KafkaAtLeastOnceSystemFactory
systems.hdfs.samza.factory=org.apache.pkg.HdfsSystemFactory

job.default.system=kafka-at-least
job.metadata.system=kafka-exact
streams.my-only-hdfs-stream.samza.system=hdfs

This demonstrates:
1. How to default all metadata streams to use one system
2. How to default all other streams to use another system
3. How to override the system for a particular stream

docs/learn/documentation/versioned/jobs/configuration-table.html
jmakes: @navina We're not restricting anything with this patch. Just introducing the new way of configuring streams which is motivated by our desire to create intermediate streams (and eventually output streams) with appropriate configs, easily, and without potential parsing issues.

jmakes:
@navina 

> >But this wouldn't work if there were commas or spaces in the stream name.

> Commas are treated as list separators in some of configs. i would guess its the same way it is treated in most "properties" based config definitions. I don't think spaces make sense in any config definition :)

Lets try another example then. When the user wants to define some property called "my.prop" on a stream called "my.stream.name"

How should samza parse the config:
systems.mySystem.streams.my.stream.name.my.prop=value

Where does the stream name end and the property name begin?

By introducing rules for logical identifiers (system and streamId) we can ensure that the identifier is always parseable. 

Would you not agree we're at risk with our current wild-west anything-goes approach?

And yes. This patch is fully backward compatible. The new config style takes precedence, but the old configs are still read. As @nickpan47 suggested though. Maybe a rewriter is cleaner.

docs/learn/documentation/versioned/jobs/configuration-table.html
nickpan47: @navina regarding to the code portability issue I referred to, I agree that my example is not the best one to illustrate it. Let me try another round: Samza today supports both HDFS and Kafka inputs. Now, let's say the user wants to process "PageViewEvent" in a Samza application, which may take the same "PageViewEvent" from HDFS or Kafka. W/o a unified logic id for the stream "PageViewEvent", the user's code has to deal w/ input is either "hdfs.file://my.dir/PageViewEvent" or "kafka.PageViewEvent". How is that portable between systems? The idea of introducing the logic stream id helps to:
1. make the stream ID as the top-level configure variable and allow user program directly refer to it
2. avoid fixation of physical stream name to the configure variable name
3. avoid fixation of a stream's configuration to a system's configuration variable name

Meanwhile, I don't think @jmakes is planning to remove the support for the old configuration here. I am proposing to deprecate the old configuration for streams to avoid the confusion, especially in the documentation for configuration table here.

We got a chance to move forward w/ our API in this round. I view configuration also as part of the API and I would rather to clean it now than keeping both s.t. the configuration for systems and streams are muddy. We can still support the old deprecated configuration for now, but need to mark those deprecated configurations clear in the web doc.

@jmakes the "re-write" I referred to in my comment was not referring to a config re-writer. I was merely referring to re-write the configuration table in the web doc w/ new config only and move the old ones to deprecated section.

As for the code check-in, I would suggest that we check this code in w/ support for both configurations, and open a new JIRA to discuss the deprecation of the old configuration for streams. We need to aware that we *have to* resolve the new JIRA before release to avoid the confusion as @navina pointed out.

My few cents.

docs/learn/documentation/versioned/jobs/configuration-table.html
nickpan47: @jmakes as for the scope of configuration, I thought that you were creating one more level above job when you were working on the pipeline feature. Anyways, w/ the new API that potentially create a multi-stage Samza application (as in SAMZA-1041), I believe that we need to have the following scopes of configuration variables now: application ==> job ==> task, w/ the definition that application configuration include all common configures for the multi-stage Samza application that may have multiple jobs, job configuration include all common configuration for Samza containers in a job, and task configuration include configuration at task level. If that sounds good to you, let me open a new JIRA regarding to creating the new scope in Samza configuration.

samza-log4j/src/main/java/org/apache/samza/config/Log4jSystemConfig.java
```diff
@@ -78,7 +81,8 @@ public String getSerdeClass(String name) {
   }
 
   public String getStreamSerdeName(String systemName, String streamName) {
-    String streamSerdeNameConfig = String.format(StreamConfig.MSG_SERDE(), systemName, streamName);
-    return get(streamSerdeNameConfig, null);
+    StreamConfig streamConfig =  new StreamConfig(this);
+    scala.Option<String> option = streamConfig.getStreamMsgSerde(new SystemStream(systemName, streamName));
```
nickpan47: qq: so it seems that we do not plan to create logic streamId for the logging topic? Can we make it clear and document how the logging topic is created s.t. it does not require  a logic streamId?

docs/learn/documentation/versioned/jobs/configuration-table.html
jmakes: @nickpan47 ahh, now I remember what you were talking about. Yes, the pipeline feature was going to do something like what you described. We were going to introduce a new "processors" scope that uses the same syntax as this "streams" scope. So to configure a particular processor, configs could be added to 
processors.processorId.*

Note 2 things about this approach:
1. It probably isn't needed until we actually support separate processors. So I wasn't planning on adding it for MVP.
2. Instead of "adding a level above the job" the job now becomes the application and we're adding a level below it. The net result is the same, other than naming (job/application)

I'm not committed to this, but it felt like the least invasive approach from a migration standpoint.

I'll create the JIRA, but will rely on your judgment for priority (MVP or not)

docs/learn/documentation/versioned/jobs/configuration-table.html
jmakes: @nickpan47 excellent summary and suggestions. I will do exactly that:
1. Mark the old configs deprecated and document them in the new format
2. File a JIRA to discuss full deprecation of the old configs

Thanks for the thorough response!

navina: > The idea of introducing the logic stream id helps to:
> make the stream ID as the top-level configure variable and allow user program directly refer > to it 
> avoid fixation of physical stream name to the configure variable name
> avoid fixation of a stream's configuration to a system's configuration variable name

+1 for all the above. I am still not convinced with your user example because it is not the user code that should be dealing with the input parsing. But the ideas behind logical stream-id sound good to me. 
I am all for decoupling systems and streams configurations. We just need to make sure we roll it out correctly instead of removing the configs in one shot. 

> We got a chance to move forward w/ our API in this round. I view configuration also as part of the API and I would rather to clean it now than keeping both s.t. the configuration for systems and streams are muddy. We can still support the old deprecated configuration for now, but need to mark those deprecated configurations clear in the web doc.

Is there a way to do multiple +1 on this comment 👍  

> @jmakes the "re-write" I referred to in my comment was not referring to a config re-writer. 

Thanks for mentioning this Yi. I was going to oppose Jake's suggestion on a "rewriter" in the other comment :) I am fully opposed to any form of "config rewriter". It was a mistake we made in the first place and I would like to undo it at an opportune moment 👍

navina: > Mark the old configs deprecated and document them in the new format
> File a JIRA to discuss full deprecation of the old configs

@jmakes Sounds good to me!

docs/learn/documentation/versioned/jobs/configuration-table.html
navina: > systems.kafka-exact.samza.factory=org.apache.pkg.KafkaExactOnceSystemFactory
> systems.kafka-at-least.samza.factory=org.apache.pkg.KafkaAtLeastOnceSystemFactory
> systems.hdfs.samza.factory=org.apache.pkg.HdfsSystemFactory

Well, imo, exactly-once is a "property" of the producer. So, I would expect a configuration like this:
systems.kafka-exactly.samza.factory = org.apache.pkg.KafkaSystemFactory
systems.kafka-exactly.samza.producer.exactly-once.enabled = true

Factory includes a producer and consumer. I don't think we will introduce a separate factory for exactly-once producer. What does that mean for the consumer then? 
Also, I think we are all projecting what the config "might" look like. But my point was to distinguish system-scoped properties Vs stream-scoped properties. 

Another question : Based on your example, if I specify a producer property for a stream A and not stream B (which belongs to the same system), will we create 2 producer instances or 1 ?

navina: >  I believe that we need to have the following scopes of configuration variables now: application ==> job ==> task

+1 to Yi's suggestion here. Let's distinguish between "application" and "job" level properties. 

> We were going to introduce a new "processors" scope that uses the same syntax as this "streams" scope. So to configure a particular processor, configs could be added to
processors.processorId.*

I would like to discuss the "processors" scope further. I am planning on a proposal that doesn't require the user to specify a processorId. It is a hassle for the user to specify a unique ID for each processor within the job. I am saying this with the assumption that the relationship between StreamApplication and processors is 1:N --> 1 Streamapplication consists of 1 or more processors. Ideally, N is bounded by the number of partitions in the input streams and grouper criteria. 
I am guessing for multi-stage, you want to associate properties at every stage. If that is case, I would recommend a different approach of associating a group of processors to a stage identifier. This doesn't have to exposed to the user. It can be internally generated and managed. I would like to understand the requirements for this in the multi-stage scenario before I suggest any more solutions :D

docs/learn/documentation/versioned/jobs/configuration-table.html
navina: Thanks for explaining, @jmakes ! 
This explanation is pretty clear. Consider saving this for your feature documentation :)

docs/learn/documentation/versioned/jobs/configuration-table.html
jmakes: > Well, imo, exactly-once is a "property" of the producer. So, I would expect a configuration like this:
> systems.kafka-exactly.samza.factory = org.apache.pkg.KafkaSystemFactory
> systems.kafka-exactly.samza.producer.exactly-once.enabled = true

That's an even more simple scenario. It will be EXACTLY the same in the new build. We're only separating out stream configs, not system configs. 

Anything in the old config that wasn't prefixed by "systems.systemName.streams.streamsName." will have 0 changes after this patch.

> Another question : Based on your example, if I specify a producer property for a stream A and not stream B (which belongs to the same system), will we create 2 producer instances or 1 ?
Invalid scenario. We don't define producers on a per stream basis. We do it on a per system basis.

navina:
> Would you not agree we're at risk with our current wild-west anything-goes approach?

Yes, I agree to that. I was again concerned only about compatibility. It seems like we all want to have more stricter rules on our config hierarchy and nomenclature. May be, this is a good time to discuss it, even if we don't fully implement it. 

> And yes. This patch is fully backward compatible. The new config style takes precedence, but the old configs are still read. As @nickpan47 suggested though. Maybe a rewriter is cleaner.
I think you and Yi have already clarified this. Thanks!

samza-api/src/main/java/org/apache/samza/system/ExecutionEnvironment.java
```diff
@@ -51,8 +52,10 @@ static ExecutionEnvironment getLocalEnvironment(Config config) {
    */
   static ExecutionEnvironment fromConfig(Config config) {
```
navina: QQ: Is this the same ExecutionEnvironment that we were planning to rename it as ApplicationRunner?

samza-core/src/main/java/org/apache/samza/system/StandaloneExecutionEnvironment.java
```diff
@@ -28,7 +28,11 @@
 /**
  * This class implements the {@link ExecutionEnvironment} that runs the applications in standalone environment
  */
-public class StandaloneExecutionEnvironment implements ExecutionEnvironment {
+public class StandaloneExecutionEnvironment extends AbstractExecutionEnvironment {
```
navina: Can we rename this to "LocalExecutionEnvironment" ? Standalone is pretty overused and seems like everyone has their own interpretation of it.

samza-core/src/main/java/org/apache/samza/system/RemoteExecutionEnvironment.java
```diff
@@ -24,7 +24,11 @@
 /**
  * This class implements the {@link ExecutionEnvironment} that runs the applications in YARN environment
```
navina: nit: I am guessing this applies to any execution environment where job needs to be submitted to a central location. Perhaps you can call out that we only support Yarn environment today

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
navina: +1 to Yi's suggestion. 

However, I prefer to slowly move out of scala for newer configs :)

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -165,13 +166,6 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
     kafkaChangeLogProperties
   }
 
-  def getTopicKafkaProperties(systemName: String, streamName: String) = {
```
navina: Thanks for removing ununsed methods!

samza-api/src/main/java/org/apache/samza/system/ExecutionEnvironment.java
jmakes: Yes. @xinyuiscool is working on that part, if you have any questions.

samza-log4j/src/main/java/org/apache/samza/config/Log4jSystemConfig.java
jmakes: I wouldn't say that we don't plan on it. It's just not part of this patch. There are a number of places in the code in which SystemStream is still used (e.g. task.inputs). SAMZA-1118 was filed to go through all the configurations and migrate them to streamId.
When I get back, I'd be happy to start on that one if you think it's MVP.

jmakes: BTW, I just realized that the logging page doesn't explicitly describe the default naming convention, so I added it. 

With this info, it should be easy for users to set properties on that stream using either the old or new stream config pattern

samza-core/src/main/java/org/apache/samza/system/StandaloneExecutionEnvironment.java
jmakes: cc @xinyuiscool since this should probably be done while renaming to ApplicationRunner

samza-core/src/main/java/org/apache/samza/system/RemoteExecutionEnvironment.java
jmakes: cc @xinyuiscool since this should probably be done while renaming to ApplicationRunner

samza-core/src/main/java/org/apache/samza/system/AbstractExecutionEnvironment.java
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.samza.config.StreamConfig;
+
+
+public abstract class AbstractExecutionEnvironment implements ExecutionEnvironment {
+
+  private final Config config;
+
+  public AbstractExecutionEnvironment(Config config) {
+    if (config == null) {
+      throw new NullPointerException();
```
jmakes: I disagree. According to the documentation, "Thrown when an application attempts to use null in a case where an object is required." https://docs.oracle.com/javase/7/docs/api/java/lang/NullPointerException.html
I think this is a totally valid use of NPE and can provide examples of the Java API using it similarly. 

I don't feel as strongly about the message part. NPEs are pretty self explanatory, but if you think a message will help, I'm ok with adding one.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
```diff
@@ -19,64 +19,187 @@
...
   def getSerdeStreams(systemName: String) = {
     val subConf = config.subset("systems.%s.streams." format systemName, true)
-    subConf
+    val legacySystemStreams = subConf
       .keys
-      .filter(k => k.endsWith(".samza.msg.serde") || k.endsWith(".samza.key.serde"))
+      .filter(k => k.endsWith(StreamConfig.MSG_SERDE) || k.endsWith(StreamConfig.KEY_SERDE))
       .map(k => {
-        val streamName = k.substring(0, k.length - 16 /* .samza.XXX.serde length */ )
+        val streamName = k.substring(0, k.length - 15 /* samza.XXX.serde length */ )
```
jmakes: Yes. Thank you for catching that!

samza-core/src/main/java/org/apache/samza/system/AbstractExecutionEnvironment.java
prateekm: The use cases listed there are all related to dereferencing/using a null pointer, which is my general experience as well. Don't feel strongly either way, just felt odd. +1 for including a message.

samza-log4j/src/main/java/org/apache/samza/config/Log4jSystemConfig.java
jmakes: Scope: SAMZA-1120
Deprecate: SAMZA-1118

PR #76, opened 2017-03-07
SAMZA-1122: Rename ExecutionEnvironment to ApplicationRunner

xinyuiscool:
Some refactoring/cleanup:

- rename ExecutionEnvironment to ApplicationRunner, including all the subclasses.
- rename the package to be org.apache.samza.runtime
- rename the StandalondApplicationRunner to be LocalApplicationRunner

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
+++ b/samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
@@ -16,58 +16,60 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.samza.system;
+package org.apache.samza.runtime;
 
 import java.lang.reflect.Constructor;
 import org.apache.samza.annotation.InterfaceStability;
 import org.apache.samza.config.ConfigException;
 import org.apache.samza.operators.StreamGraphBuilder;
 import org.apache.samza.config.Config;
+import org.apache.samza.system.StreamSpec;
 
 
 /**
  * Interface to be implemented by physical execution engine to deploy the config and jobs to run the {@link org.apache.samza.operators.StreamGraph}
  *
  * Implementations of this interface must define a constructor with a single {@link Config} as the argument in order
- * to support the {@link ExecutionEnvironment#fromConfig(Config)} static constructor.
+ * to support the {@link ApplicationRunner#fromConfig(Config)} static constructor.
  */
 @InterfaceStability.Unstable
-public interface ExecutionEnvironment {
+public interface ApplicationRunner {
 
-  String ENVIRONMENT_CONFIG = "job.execution.environment.class";
-  String DEFAULT_ENVIRONMENT_CLASS = "org.apache.samza.system.StandaloneExecutionEnvironment";
+  String RUNNER_CONFIG = "job.runner.class";
+  String DEFAULT_RUNNER_CLASS = "org.apache.samza.runtime.RemoteApplicationRunner";
 
   /**
-   * Static method to load the local standalone environment
+   * Static method to create the local {@link ApplicationRunner}.
    *
-   * @param config  configuration passed in to initialize the Samza standalone process
-   * @return  the standalone {@link ExecutionEnvironment} to run the user-defined stream applications
+   * @param config  configuration passed in to initialize the Samza local process
+   * @return  the local {@link ApplicationRunner} to run the user-defined stream applications
    */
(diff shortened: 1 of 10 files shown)


samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
```diff
@@ -16,34 +16,35 @@
...
+ * to support the {@link ApplicationRunner#fromConfig(Config)} static constructor.
  */
 @InterfaceStability.Unstable
-public interface ExecutionEnvironment {
+public interface ApplicationRunner {
 
-  String ENVIRONMENT_CONFIG = "job.execution.environment.class";
-  String DEFAULT_ENVIRONMENT_CLASS = "org.apache.samza.system.StandaloneExecutionEnvironment";
+  String RUNNER_CONFIG = "job.runner.class";
+  String DEFAULT_RUNNER_CLASS = "org.apache.samza.runtime.AbstractApplicationRunner";
```
prateekm: Default should be a concrete implementation. Maybe default to Yarn or Standalone?

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -16,17 +16,18 @@
...
-package org.apache.samza.system;
+package org.apache.samza.runtime;
 
 import org.apache.samza.operators.StreamGraphBuilder;
 import org.apache.samza.config.Config;
 
+
 /**
- * This class implements the {@link ExecutionEnvironment} that runs the applications in YARN environment
+ * This class implements the {@link ApplicationRunner} that runs the applications in YARN environment
```
prateekm: If this is YARN specific, should probably be called YarnApplicationRunner. If not, should update documentation to say something generic like "runs the application in a remote cluster."

samza-core/src/test/java/org/apache/samza/example/KeyValueStoreExample.java
```diff
@@ -71,7 +71,7 @@
   public static void main(String[] args) throws Exception {
     CommandLine cmdLine = new CommandLine();
     Config config = cmdLine.loadConfig(cmdLine.parser().parse(args));
-    ExecutionEnvironment standaloneEnv = ExecutionEnvironment.getLocalEnvironment(config);
+    ApplicationRunner standaloneEnv = ApplicationRunner.getLocalRunner(config);
```
prateekm: Nitpick: update variable names too, here and elsewhere.

samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
xinyuiscool: Defaulted as Remote (since legacy jobs are using remote execution).

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
xinyuiscool: It's not Yarn specific. I corrected the javadoc. Thanks for the catch.

samza-core/src/test/java/org/apache/samza/example/KeyValueStoreExample.java
xinyuiscool: Good catch. I updated the javadoc and var names in all these examples.

samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
```diff
@@ -16,58 +16,60 @@
...
- * to support the {@link ExecutionEnvironment#fromConfig(Config)} static constructor.
+ * to support the {@link ApplicationRunner#fromConfig(Config)} static constructor.
  */
 @InterfaceStability.Unstable
-public interface ExecutionEnvironment {
+public interface ApplicationRunner {
 
-  String ENVIRONMENT_CONFIG = "job.execution.environment.class";
-  String DEFAULT_ENVIRONMENT_CLASS = "org.apache.samza.system.StandaloneExecutionEnvironment";
+  String RUNNER_CONFIG = "job.runner.class";
```
nickpan47: As I discussed w/ @jmakes and @navina in another PR, this configuration should be at application level now: app.runner.class

xinyuiscool: sure. I will open another pr for this change.

PR #77, opened 2017-03-07
SAMZA-1124:Job coordinator with time out

sborya:
If a processor doesn't join the barrier for the TimeOut time - the barrier is cancelled. All the processor should unsubscribe from it.


xinyuiscool:
Jenkins has style check failures for the patch. Could you please take a look?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -0,0 +1,221 @@
...
+  //////////////////////////////////////////////// LEADER stuff ///////////////////////////
+  @Override
+  public void onBecomeLeader() {
+    log.info("ZkJobCoordinator::onBecomeLeader - I became the leader!");
+
+    List<String> emptyList = new ArrayList<>();
+
+    // actual actions to do are the same as onProcessorChange()
+    debounceTimer.scheduleAfterDebounceTime(ScheduleAfterDebounceTime.ON_PROCESSOR_CHANGE,
+        ScheduleAfterDebounceTime.DEBOUNCE_TIME_MS, () -> onProcessorChange(emptyList));
```
xinyuiscool: is it better to pass in the function () -> generateNewJobModel() directly here? Why adding this extra level of indirection?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -31,6 +33,8 @@
   private final ZkUtils zkUtils;
```
xinyuiscool: Can we make this ZkBarrier general like ZkLeaderElector? LocalApplicationRunner also needs this to be able to have the barrier for environment initialization (e.g creating streams). By looking at the code it seems if we replace "version" with a general "name" then this can be a general Barrier. Can we make this change so this class can be reused by LocalApplicationRunner?

samza-core/src/main/java/org/apache/samza/config/JavaJobConfig.java
```diff
@@ -0,0 +1,41 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.config;
+
+public class JavaJobConfig extends MapConfig {
```
sborya: removed in the previous patch.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: That was my original code, but then I thought that if we ever needed something else to be done when processors change, we would want it to be executed too.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -0,0 +1,221 @@
...
+    this.containerController = containerController;
+    this.zkController = new ZkControllerImpl(String.valueOf(processorId), zkUtils, debounceTimer, this);
+    this.config = config;
+
+
+    barrier = new ZkBarrierForVersionUpgrade(zkUtils, debounceTimer);
+    streamMetadataCache = initJobModelGeneration();
+  }
+
+  private StreamMetadataCache initJobModelGeneration() {
```
sborya: sure.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: I will see if it can be done. I've opened SAMZA-1128 for it.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -45,36 +49,66 @@ public ZkBarrierForVersionUpgrade(ZkUtils zkUtils, ScheduleAfterDebounceTime deb
...
+    }
+  }
+
   @Override
-  public void startBarrier(String version, List<String> processorsNames) {
-    String barrierPath = String.format("%s/barrier_%s", barrierPrefix, version);
-    String barrierDonePath = String.format("%s/barrier_done", barrierPath);
-    String barrierProcessors = String.format("%s/barrier_processors", barrierPath);
+  public void start(String version, List<String> processorsNames) {
+    final String barrierPath = String.format("%s/barrier_%s", barrierPrefix, version);
```
navina: Why are these Strings declared final here?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -31,6 +33,8 @@
   private final ZkUtils zkUtils;
   private final ZkKeyBuilder keyBuilder;
   private final static String BARRIER_DONE = "done";
+  private final static String BARRIER_TIMED_OUT = "TIMED_OUT";
+  private final static long BARRIER_TIMED_OUT_MS = 60 * 1000;
```
navina: Shouldn't this be configurable?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -152,6 +171,11 @@ public void handleDataChange(String dataPath, Object data)
       if (done.equals(BARRIER_DONE)) {
         zkUtils.unsubscribeDataChanges(barrierPathDone, this);
         debounceTimer.scheduleAfterDebounceTime(ScheduleAfterDebounceTime.JOB_MODEL_VERSION_CHANGE, 0, callback);
+      } else if (done.equals(BARRIER_TIMED_OUT)) {
+        // timed out
+        LOG.error("Barrier for " + dataPath + " timed out");
+        System.out.println("Barrier for " + dataPath + " timed out");
+        zkUtils.unsubscribeDataChanges(barrierPathDone, this);
```
navina: What does the comment on Line 180 mean?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -152,6 +171,11 @@ public void handleDataChange(String dataPath, Object data)
       if (done.equals(BARRIER_DONE)) {
```
navina: Is there any documentation on the allowed values for this ZkNode? What do each of these values mean?

PR #78, opened 2017-03-07
SAMZA-1121: StreamAppender should not propagate exceptions to the caller

prateekm:
StreamAppender#append currently propagates any exceptions while sending messages to the underlying logging system to the calling code. Since users don't expect log statements to throw exceptions, this can cause unexpected failures scenarios. We should catch exceptions and log to stderr instead.


vjagadish1989 (approved):
approved, looks good to me. thank you for the contributions.

PR #79, opened 2017-03-08
Samza 1123: Create intermediate stream in partitionBy() operator

xinyuiscool:
For partitionBy() operator, Samza generates an intermediate stream with id based on operator name and id, and system based on config. The intermediate streams will be materialized later by different execution environments. For example, if the intermediate stream is a Kafka stream, the topic will be created before the application starts.

Also renamed the config from "job.runner.class" to "app.runner.class".

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
+++ b/samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
@@ -35,7 +35,7 @@ import org.apache.samza.system.StreamSpec;
 @InterfaceStability.Unstable
 public interface ApplicationRunner {
 
-  String RUNNER_CONFIG = "job.runner.class";
+  String RUNNER_CONFIG = "app.runner.class";
   String DEFAULT_RUNNER_CLASS = "org.apache.samza.runtime.RemoteApplicationRunner";
 
   /**
--- a/samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
+++ b/samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
@@ -16,58 +16,60 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.samza.system;
+package org.apache.samza.runtime;
 
 import java.lang.reflect.Constructor;
 import org.apache.samza.annotation.InterfaceStability;
 import org.apache.samza.config.ConfigException;
 import org.apache.samza.operators.StreamGraphBuilder;
 import org.apache.samza.config.Config;
+import org.apache.samza.system.StreamSpec;
 
 
 /**
  * Interface to be implemented by physical execution engine to deploy the config and jobs to run the {@link org.apache.samza.operators.StreamGraph}
  *
  * Implementations of this interface must define a constructor with a single {@link Config} as the argument in order
- * to support the {@link ExecutionEnvironment#fromConfig(Config)} static constructor.
+ * to support the {@link ApplicationRunner#fromConfig(Config)} static constructor.
  */
 @InterfaceStability.Unstable
-public interface ExecutionEnvironment {
+public interface ApplicationRunner {
 
-  String ENVIRONMENT_CONFIG = "job.execution.environment.class";
-  String DEFAULT_ENVIRONMENT_CLASS = "org.apache.samza.system.StandaloneExecutionEnvironment";
+  String RUNNER_CONFIG = "job.runner.class";
+  String DEFAULT_RUNNER_CLASS = "org.apache.samza.runtime.RemoteApplicationRunner";
 
   /**
-   * Static method to load the local standalone environment
+   * Static method to create the local {@link ApplicationRunner}.
    *
-   * @param config  configuration passed in to initialize the Samza standalone process
-   * @return  the standalone {@link ExecutionEnvironment} to run the user-defined stream applications
+   * @param config  configuration passed in to initialize the Samza local process
+   * @return  the local {@link ApplicationRunner} to run the user-defined stream applications
    */
-  static ExecutionEnvironment getLocalEnvironment(Config config) {
+  static ApplicationRunner getLocalRunner(Config config) {
     return null;
   }
 
   /**
-   * Static method to load the non-standalone environment.
+   * Static method to load the {@link ApplicationRunner}
    *
    * Requires the implementation class to define a constructor with a single {@link Config} as the argument.
    *
    * @param config  configuration passed in to initialize the Samza processes
-   * @return  the configure-driven {@link ExecutionEnvironment} to run the user-defined stream applications
+   * @return  the configure-driven {@link ApplicationRunner} to run the user-defined stream applications
    */
-  static ExecutionEnvironment fromConfig(Config config) {
+  static ApplicationRunner fromConfig(Config config) {
     try {
-      Class<?> environmentClass = Class.forName(config.get(ENVIRONMENT_CONFIG, DEFAULT_ENVIRONMENT_CLASS));
-      if (ExecutionEnvironment.class.isAssignableFrom(environmentClass)) {
-        Constructor<?> constructor = environmentClass.getConstructor(Config.class); // *sigh*
-        return (ExecutionEnvironment) constructor.newInstance(config);
+      Class<?> runnerClass = Class.forName(config.get(RUNNER_CONFIG, DEFAULT_RUNNER_CLASS));
+      if (ApplicationRunner.class.isAssignableFrom(runnerClass)) {
+        Constructor<?> constructor = runnerClass.getConstructor(Config.class); // *sigh*
+        return (ApplicationRunner) constructor.newInstance(config);
       }
     } catch (Exception e) {
-      throw new ConfigException(String.format("Problem in loading ExecutionEnvironment class %s", config.get(ENVIRONMENT_CONFIG)), e);
+      throw new ConfigException(String.format("Problem in loading ApplicationRunner class %s", config.get(
+          RUNNER_CONFIG)), e);
     }
     throw new ConfigException(String.format(
-        "Class %s does not implement interface ExecutionEnvironment properly",
-        config.get(ENVIRONMENT_CONFIG)));
+        "Class %s does not implement interface ApplicationRunner properly",
+        config.get(RUNNER_CONFIG)));
   }
 
   /**
--- a/samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
+++ b/samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
@@ -163,10 +163,12 @@ public class MessageStreamImpl<M> implements MessageStream<M> {
 
   @Override
   public <K> MessageStream<M> partitionBy(Function<M, K> parKeyExtractor) {
-    MessageStreamImpl<M> intStream = this.graph.createIntStream(parKeyExtractor);
+    int opId = graph.getNextOpId();
+    String streamId = String.format("%s-%s", OperatorSpec.OpCode.PARTITION_BY.name().toLowerCase(), opId);
+    MessageStreamImpl<M> intStream = this.graph.createIntStream(streamId, parKeyExtractor);
     OutputStream<M> outputStream = this.graph.getOutputStream(intStream);
     this.registeredOperatorSpecs.add(OperatorSpecs.createPartitionOperatorSpec(outputStream.getSinkFunction(),
-        this.graph, outputStream));
+        this.graph, outputStream, opId));
     return intStream;
   }
   /**
--- a/samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
+++ b/samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
@@ -164,8 +164,7 @@ public class MessageStreamImpl<M> implements MessageStream<M> {
   @Override
   public <K> MessageStream<M> partitionBy(Function<M, K> parKeyExtractor) {
     int opId = graph.getNextOpId();
-    String streamId = String.format("%s-%s", OperatorSpec.OpCode.PARTITION_BY.name().toLowerCase(), opId);
-    MessageStreamImpl<M> intStream = this.graph.createIntStream(streamId, parKeyExtractor);
+    MessageStreamImpl<M> intStream = this.graph.generateIntStreamFromOpId(opId, parKeyExtractor);
     OutputStream<M> outputStream = this.graph.getOutputStream(intStream);
     this.registeredOperatorSpecs.add(OperatorSpecs.createPartitionOperatorSpec(outputStream.getSinkFunction(),
         this.graph, outputStream, opId));
--- a/samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
+++ b/samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
@@ -166,8 +166,7 @@ public class MessageStreamImpl<M> implements MessageStream<M> {
     int opId = graph.getNextOpId();
     MessageStreamImpl<M> intStream = this.graph.generateIntStreamFromOpId(opId, parKeyExtractor);
     OutputStream<M> outputStream = this.graph.getOutputStream(intStream);
-    this.registeredOperatorSpecs.add(OperatorSpecs.createPartitionOperatorSpec(outputStream.getSinkFunction(),
-        this.graph, outputStream, opId));
+    this.registeredOperatorSpecs.add(OperatorSpecs.createPartitionOperatorSpec(outputStream.getSinkFunction(), outputStream, opId));
     return intStream;
   }
   /**
--- a/samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
+++ b/samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
@@ -24,8 +24,9 @@ import java.util.Map;
 import java.util.function.Function;
 import org.apache.samza.operators.data.MessageEnvelope;
 import org.apache.samza.operators.functions.SinkFunction;
+import org.apache.samza.operators.spec.OperatorSpec;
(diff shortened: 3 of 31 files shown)


samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
```diff
@@ -35,7 +35,7 @@
 @InterfaceStability.Unstable
 public interface ApplicationRunner {
 
-  String RUNNER_CONFIG = "job.runner.class";
+  String RUNNER_CONFIG = "app.runner.class";
```
prateekm: Issue: Did we decide to rename this? If not yet, let's get feedback from @nickpan47 @jmakes @navina etc.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -231,9 +241,8 @@ public MessageStreamImpl getInputStream(SystemStream sstream) {
    * @param <M>  the type of input message
    * @return  the {@link OutputStream} object for the re-partitioned stream
    */
-  <PK, M> MessageStreamImpl<M> createIntStream(Function<M, PK> parKeyFn) {
-    // TODO: placeholder to auto-generate intermediate streams via {@link StreamSpec}
-    StreamSpec streamSpec = this.createIntStreamSpec();
+  <PK, M> MessageStreamImpl<M> createIntStream(String streamId, Function<M, PK> parKeyFn) {
+    StreamSpec streamSpec = runner.streamFromConfig(streamId);
```
prateekm: Issue: How does this work? The streamId isn't going to exist in user configs, right?

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -39,7 +40,8 @@ public AbstractApplicationRunner(Config config) {
   @Override
   public StreamSpec streamFromConfig(String streamId) {
     StreamConfig streamConfig = new StreamConfig(config);
-    String physicalName = streamConfig.getPhysicalName(streamId, streamId);
+    String defaultPhysicalName = String.format("%s-%s-%s", config.get(JobConfig.JOB_NAME()), config.get(JobConfig.JOB_ID(), "1"), streamId);
```
prateekm: Will be good to make this reversible - e.g. by making sure `-` is only used as a separator and is not present in JOB_NAME.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -39,7 +40,8 @@ public AbstractApplicationRunner(Config config) {
...
+    String physicalName = streamConfig.getPhysicalName(streamId, defaultPhysicalName);
```
prateekm: Issue: This method signature feels a little weird. Why pass the default name through via the method call? Can't the default be either in the config class itself, or used only by the caller?

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -77,8 +77,9 @@ public StreamOperatorTask(StreamGraphBuilder graphBuilder) {
 
   @Override
   public final void init(Config config, TaskContext context) throws Exception {
+    ApplicationRunner runner = ApplicationRunner.fromConfig(config);
```
prateekm: Issue: This should probably be passed in from SamzaContainer via the constructor.

samza-core/src/test/java/org/apache/samza/operators/TestMessageStreamImpl.java
```diff
@@ -18,6 +18,15 @@
  */
 package org.apache.samza.operators;
 
+import java.util.ArrayList;
```
prateekm: Codestyle: Here and in other RB too, I think the current order b/w org.* and java.* is as it was earlier. Let's make sure it's consistence w/ the existing code.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -208,8 +218,8 @@ public ContextManager getContextManager() {
    */
   public MessageStreamImpl getInputStream(SystemStream sstream) {
     for (MessageStream entry: this.inStreams.values()) {
-      if (((InputStreamImpl) entry).getSpec().getSystemName() == sstream.getSystem() &&
-          ((InputStreamImpl) entry).getSpec().getPhysicalName() == sstream.getStream()) {
+      if (((InputStreamImpl) entry).getSpec().getSystemName().equals(sstream.getSystem()) &&
```
prateekm: Thanks!

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -163,10 +163,12 @@ public K getOtherKey(M message) {
 
   @Override
   public <K> MessageStream<M> partitionBy(Function<M, K> parKeyExtractor) {
-    MessageStreamImpl<M> intStream = this.graph.createIntStream(parKeyExtractor);
+    int opId = graph.getNextOpId();
+    String streamId = String.format("%s-%s", OperatorSpec.OpCode.PARTITION_BY.name().toLowerCase(), opId);
```
prateekm: Can't tell how this works from this RB alone. Is this going to become the logical id for the stream when you create a spec for it in the Planner? And then the physical name by default appends the JOB_NAME and JOB_ID?

samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
nickpan47: @prateekm that was my suggestion from another PR. Currently, w/ the introduction of the concept of application, we have three different scopes now: application -> job -> task. I chat w/ @navina and @jmakes already. I will let them to confirm.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
nickpan47: I have the same question. I think we will need a util method to create the intermediate StreamSpec from the StreamGraphImpl, partitionBy operator, and application ID (I would argue this would have the same meaning as the Pipeline ID that Jacob came up w/ earlier).

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
nickpan47: Question again on the configuration scope: are we "borrowing" job.name and job.id as app level configuration here? I think that the unique intermediate StreamSpec should be at app level?

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
nickpan47: Agree w/ @prateekm we can wrap this in StreamConfig.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
nickpan47: @xinyuiscool thanks for fixing my fault!

samza-core/src/test/java/org/apache/samza/operators/TestMessageStreamImpl.java
xinyuiscool: After intellij optimize the order of imports, it's now alphabetically sorted.

prateekm: Yeah, but a lot of existing files have the previous order. Not a big deal, it's already inconsistent across files.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -224,16 +236,17 @@ public MessageStreamImpl getInputStream(SystemStream sstream) {
...
+   * @param opId  operator ID
    * @param parKeyFn  the function to extract the partition key from the input message
    * @param <PK>  the type of partition key
    * @param <M>  the type of input message
    * @return  the {@link OutputStream} object for the re-partitioned stream
    */
-  <PK, M> MessageStreamImpl<M> createIntStream(Function<M, PK> parKeyFn) {
-    // TODO: placeholder to auto-generate intermediate streams via {@link StreamSpec}
-    StreamSpec streamSpec = this.createIntStreamSpec();
+  <PK, M> MessageStreamImpl<M> generateIntStreamFromOpId(int opId, Function<M, PK> parKeyFn) {
```
prateekm: Naming: Prefer `createRepartionStream` or something similar since this is specific to a repartitioner.

Could also just inline this method.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -224,16 +236,17 @@ public MessageStreamImpl getInputStream(SystemStream sstream) {
...
+    String streamId = String.format("%s-%s", OperatorSpec.OpCode.PARTITION_BY.name().toLowerCase(), opId);
+    StreamSpec streamSpec = runner.streamFromConfig(streamId);
 
     if (!this.inStreams.containsKey(streamSpec.getId())) {
       this.inStreams.putIfAbsent(streamSpec.getId(), new IntermediateStreamImpl(this, streamSpec, null, null, parKeyFn));
```
prateekm: Issue: Don't think we need both the `containsKey` check and `putIfAbsent`.

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -163,10 +163,11 @@ public K getOtherKey(M message) {
 
   @Override
   public <K> MessageStream<M> partitionBy(Function<M, K> parKeyExtractor) {
-    MessageStreamImpl<M> intStream = this.graph.createIntStream(parKeyExtractor);
+    int opId = graph.getNextOpId();
+    MessageStreamImpl<M> intStream = this.graph.generateIntStreamFromOpId(opId, parKeyExtractor);
     OutputStream<M> outputStream = this.graph.getOutputStream(intStream);
     this.registeredOperatorSpecs.add(OperatorSpecs.createPartitionOperatorSpec(outputStream.getSinkFunction(),
-        this.graph, outputStream));
+        this.graph, outputStream, opId));
```
prateekm: Issue: I think the `graph` parameter is only used to generate the operatorId later. If we're already getting the ID here here, should not pass both to avoid confusion.

We can also do a consistent pass over all operators separately, I'd prefer to get and pass the ID (like here) instead of the whole graph, unless @nickpan47 had some other usage of `graph` in mind.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -69,16 +69,19 @@
...
+  public StreamOperatorTask(StreamGraphBuilder graphBuilder, ApplicationRunner runner) {
     this.graphBuilder = graphBuilder;
+    this.runner = runner;
   }
 
   @Override
   public final void init(Config config, TaskContext context) throws Exception {
     // create the MessageStreamsImpl object and initialize app-specific logic DAG within the task
-    StreamGraphImpl streams = new StreamGraphImpl();
+    StreamGraphImpl streams = new StreamGraphImpl(this.runner);
```
prateekm: Unrelated to this RB, but:
1. Don't think this needs to wait for init, should assign in constructor.
2. Prefer renaming to `streamGraph` for clarity.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
```diff
@@ -145,10 +145,14 @@ class StreamConfig(config: Config) extends ScalaMapConfig(config) with Logging {
     * Gets the physical name for the specified streamId.
     *
     * @param streamId             the identifier for the stream in the config.
-    * @param defaultPhysicalName  the default to use if the physical name is missing.
     * @return                     the physical identifier for the stream or the default if it is undefined.
     */
-  def getPhysicalName(streamId: String, defaultPhysicalName: String) = {
+  def getPhysicalName(streamId: String) = {
+    // Add prefix to default so the physical name is unique per job
+    val defaultPhysicalName = String.format("%s-%s-%s",
```
prateekm: One of the considerations behind picking a default physical name was to make it the same as the logical name, so that for systems like Kafka etc. you could make your logical ID == default physical name == PageViewEvent. From that perspective, does it make sense to put the job name and job id in the logical stream ID instead?

Also, let's check if this default naming convention is part of public documentation for streams.* configs.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
```diff
@@ -194,7 +198,7 @@ class StreamConfig(config: Config) extends ScalaMapConfig(config) with Logging {
     * will use the streamId as the stream name if the physicalName doesn't exist.
```
prateekm: See comment above. If we keep the new impl, update this + other public documentation.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
xinyuiscool: It's actually more general. Since planner might generate intermediate streams for other operators for optimization, then this will not be limited to partitionBy.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
xinyuiscool: Good catch! I didn't look at this part of the code in detail.

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
xinyuiscool: Makes sense. I removed the graph in this function. Let's sync up with @nickpan47 about your second suggestion.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
xinyuiscool: To address your comment below, I need to pass in the config now for the job name and id. So I still need to keep streamgraph created here.

Renamed it to streamGraph.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
xinyuiscool: This is a great point. I think by adding the prefix as default physical name does violate the expectation that streamId is physicalId in normal cases. I refactored it so the prefix is only appended in the logical ID. Thanks a lot for catching this!

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
xinyuiscool: I reverted back the changes for the prefix.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
prateekm: Minor, but in that case prefer `createIntStream(int opId, Function<M, PK> parKeyFn)`.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -69,25 +69,28 @@
...
   private final StreamGraphBuilder graphBuilder;
 
+  private final ApplicationRunner runner;
+
   private ContextManager contextManager;
 
-  public StreamOperatorTask(StreamGraphBuilder graphBuilder) {
+  public StreamOperatorTask(StreamGraphBuilder graphBuilder, ApplicationRunner runner) {
     this.graphBuilder = graphBuilder;
+    this.runner = runner;
```
vjagadish1989: It's slightly strange to have `ApplicationRunner` in the constructor. All you need is to be able to get a `StreamSpec` for a given `streamId`.  We probably need a separate interface for that. 

Long term we want to be able to serialize this, and make the resolution more obvious.

vjagadish1989 (approved):
looks good to me.

approved.

added feedback from final round of reviews:

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
navina: Are we introducing default physical name to allow for backward compatibility ?? Or some other motivation?
Looking at jobname-jobid prefix, I don't think compatibility is the motivation. 

I also didn't realize based on this patch that we are "generating" physicalStreamName! How is this "generated" physical name used? Isn't physical name the actual name of the stream defined in the system??

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -224,30 +240,25 @@ public MessageStreamImpl getInputStream(SystemStream sstream) {
...
-    }
+  <PK, M> MessageStreamImpl<M> generateIntStreamFromOpId(int opId, Function<M, PK> parKeyFn) {
+    String opNameWithId = String.format("%s-%s", OperatorSpec.OpCode.PARTITION_BY.name().toLowerCase(), opId);
+    String streamId = String.format("%s-%s-%s",
+        config.get(JobConfig.JOB_NAME()),
+        config.get(JobConfig.JOB_ID(), "1"),
+        opNameWithId);
+    StreamSpec streamSpec = runner.streamFromConfig(streamId);
+
+    this.inStreams.putIfAbsent(streamSpec.getId(), new IntermediateStreamImpl(this, streamSpec, null, null, parKeyFn));
```
prateekm: @xinyuiscool It looks like this method is meant as a replacement for the existing `createIntStream` (which still exists in this class), but doesn't accept and set the key and msg serdes on the IntermediateStream. What's the expected usage for these two methods?

PR #81, opened 2017-03-10
SAMZA-1134:Simplify barrier for zk version upgrade.

sborya:
see SAMZA-1134 for details.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
@@ -19,9 +19,6 @@
 
 package org.apache.samza.zk;
 
-import java.util.List;
-
-
 /**
  * Interface for a barrier - to allow synchronization between different processors to switch to a newly published
  * JobModel.
--- a/samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
@@ -29,19 +29,16 @@ import java.util.List;
 public interface BarrierForVersionUpgrade {
   /**
    * Barrier is usually started by the leader.
-   * @param version - for which the barrier is started.
-   * @param processorsNames - list of processors available at the time of the JobModel generation.
    */
-  void start(String version, List<String> processorsNames);
+  void start();
 
   /**
    * Called by the processor.
    * Updates the processor readiness to use the new version and wait on the barrier, until all other processors
    * joined.
    * The call is async. The callback will be invoked when the barrier is reached.
-   * @param version of the jobModel this barrier is protecting.
-   * @param processorsName as it appears in the list of processors.
+   * @param thisProcessorsName as it appears in the list of processors.
    * @param callback  will be invoked, when barrier is reached.
    */
-  void waitForBarrier(String version, String processorsName, Runnable callback);
+  void waitForBarrier(String thisProcessorsName, Runnable callback);
 }
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
@@ -23,6 +23,7 @@ import java.util.Arrays;
 import java.util.List;
 import org.I0Itec.zkclient.IZkChildListener;
 import org.I0Itec.zkclient.IZkDataListener;
+import org.I0Itec.zkclient.exception.ZkBadVersionException;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.samza.SamzaException;
 import org.apache.zookeeper.data.Stat;
@@ -95,12 +96,12 @@ public class ZkBarrierForVersionUpgrade implements BarrierForVersionUpgrade {
     try {
       // write a new value "TIMED_OUT", if the value was changed since previous value, make sure it was changed to "DONE"
       zkUtils.getZkClient().writeData(barrierDonePath, BARRIER_TIMED_OUT, currentStatOfBarrierDone.getVersion());
-    } catch (Exception e) {
+    } catch (ZkBadVersionException badVerE) {
       // failed to write, make sure the value is "DONE"
-      LOG.info("Barrier timeout write failed");
+      LOG.warn("Barrier timeout write failed");
       String done = zkUtils.getZkClient().<String>readData(barrierDonePath);
       if (!done.equals(BARRIER_DONE)) {
-        throw new SamzaException("Failed to write to the barrier_done, version=" + version, e);
+        throw new SamzaException("Failed to write to the barrier_done, version=" + version, badVerE);
       }
     }
   }
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
@@ -29,6 +29,20 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 
+/**
+ * This class creates a barrier for version upgrade.
+ * Leader should call barrier.start() - this will create barrier_${VERSION}, barrier_processors and barrier_done nodes in ZK,
+ * and will create a listener on any changes to barrier_processors. start() will also store the list of the processors expected
+ * to join the barrier.
+ * Each participant will register its name under barrier_processor and subscribe to the changes of barrier_done.
+ * The callback (subscribed by start()) will evaluate the current list under barrier_processors and compare it to the list
+ * of expected processors. When the lists match it means that the barrier has been reached. It will change the value of
+ * the barrier_done node to "DONE". This will inform all the subscribed processors that they may continue with the new version.
+ *
+ * start() also starts a timer for a time-out. If the timer fires before the barrier is reached, it will put "TIME_OUT" value
+ * into the barrier_done, which, in turn, will cause all the processors to be notified and unsubscribe from the barrier.
+ * This barrier becomes invalid.
+ */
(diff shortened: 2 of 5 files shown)


samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
```diff
@@ -19,29 +19,23 @@
...
    * Updates the processor readiness to use the new version and wait on the barrier, until all other processors
    * joined.
    * The call is async. The callback will be invoked when the barrier is reached.
-   * @param version of the jobModel this barrier is protecting.
-   * @param processorsName as it appears in the list of processors.
+   * @param thisProcessorsName as it appears in the list of processors.
    * @param callback  will be invoked, when barrier is reached.
    */
-  void waitForBarrier(String version, String processorsName, Runnable callback);
+  void waitForBarrier(String thisProcessorsName, Runnable callback);
```
sborya: As we discussed before, this is not a super generic barrier. It is designed specifically for the version upgrade and that's why besides the amount of processors that reached the barrier, we also validate that the names of these processors match the list of the names provided by the leader, when the barrier was created.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -40,74 +55,71 @@
...
-        // failed to write, try read/write again
-        LOG.info("Barrier timeout write failed");
-        done = zkUtils.getZkClient().<String>readData(barrierDonePath, stat);
-        if (done.equals(BARRIER_DONE))
-          return; //nothing to do
+  private void timerOff(final String version, final Stat currentStatOfBarrierDone) {
+    try {
+      // write a new value "TIMED_OUT", if the value was changed since previous value, make sure it was changed to "DONE"
+      zkUtils.getZkClient().writeData(barrierDonePath, BARRIER_TIMED_OUT, currentStatOfBarrierDone.getVersion());
+    } catch (Exception e) {
```
sborya: You are right, but I couldn't find any documentation on what was the specific exception for this call. I will look little bit more and update if I find it.

navina: I think it will be KeeperVersion. May have to walk-through the [client library code](https://github.com/sgroschupf/zkclient/blob/6ae1ca7312aa3cf19a09551f65c6753370a02aa4/src/main/java/org/I0Itec/zkclient/ZkConnection.java) to find out if there are specific errors that indicate the version mismatch.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -40,74 +55,71 @@
...
+      // failed to write, make sure the value is "DONE"
+      LOG.info("Barrier timeout write failed");
+      String done = zkUtils.getZkClient().<String>readData(barrierDonePath);
```
sborya: The only valid case for the write to fail is that the barrier was reached, and the value is DONE. In this case we do nothing. If we successfully write "TIMED_OUT", this will cause all the participants to abandon the barrier.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -40,74 +55,71 @@
...
+      if (!done.equals(BARRIER_DONE)) {
+        throw new SamzaException("Failed to write to the barrier_done, version=" + version, e);
```
sborya: I agree we should review it. There is a separate JIRA to do so.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -69,7 +69,7 @@ public ZkJobCoordinator(int processorId, Config config, ScheduleAfterDebounceTim
     this.config = config;
 
 
-    barrier = new ZkBarrierForVersionUpgrade(zkUtils, debounceTimer);
+    //barrier = new ZkBarrierForVersionUpgrade(zkUtils, debounceTimer);
```

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
navina: Sure. My point is that if the leader is the only one to update the flag to either DONE or TIMED_OUT, is there even any need for reading the data again? Is the state transition for the barrier_done flag not straightforward : 
1. <init_value> -> Done (if leader marks it as done)
2. <init_value> -> Timed Out (again, this timer is triggered by the leader??) 

I don't see any other writer possibility. So, why should we read the data again and validate?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: I looked into some examples on the Inet, and found people using ZkBadVersionException. So I changed to that one.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: I agreed with you and removed it already.

samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
```diff
@@ -19,29 +19,23 @@
...
  * JobModel.
  */
 public interface BarrierForVersionUpgrade {
   /**
    * Barrier is usually started by the leader.
-   * @param version - for which the barrier is started.
-   * @param processorsNames - list of processors available at the time of the JobModel generation.
    */
-  void start(String version, List<String> processorsNames);
+  void start();
```
navina: start() seems to be doing a lot more than starting a service. It is creating the barrier related subtree(s), setting a timer etc. It seems more apt to call it "prepareBarrier" with version and list of processors (as you had before). 
Otherwise, this information (about version and list of processors) is passed through the constructor of the implementation. Since it is not meant to be generic, why keep the interface vague?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -23,12 +23,28 @@
...
+import org.I0Itec.zkclient.exception.ZkBadVersionException;
 import org.apache.commons.collections4.CollectionUtils;
+import org.apache.samza.SamzaException;
 import org.apache.zookeeper.data.Stat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 
+/**
+ * This class creates a barrier for version upgrade.
```
navina: A general comment on documentation:
Explaining the code verbatim makes the documentation hard to read. Instead, it will be cleaner to point out what the different roles are (leader and non-leader), what is sequence of events (leader prepares barrier and members create node under barrier), what is the criteria for completion (processor Id count and values match) and what kind of failures are possible (time out).

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -40,74 +56,71 @@
...
 
   /**
    * set the barrier for the timer. If the timer is not achieved by the timeout - it will fail
    * @param version for which the barrier is created
    * @param timeout - time in ms to wait
    */
-  public void setTimer(String version, long timeout) {
-    debounceTimer.scheduleAfterDebounceTime("VersionUpgradeTimeout", timeout, ()->timerOff(version));
+  public void setTimer(final String version, final long timeout, final Stat currentStatOfBarrierDone) {
+    debounceTimer.scheduleAfterDebounceTime("VersionUpgradeTimeout", timeout, ()->timerOff(version, currentStatOfBarrierDone));
```
navina: Is there a list of allowed strings for this debounce timer? Can we define the allowed constants in a single place?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: modified.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -40,74 +56,71 @@
...
   }
 
   /**
    * set the barrier for the timer. If the timer is not achieved by the timeout - it will fail
    * @param version for which the barrier is created
    * @param timeout - time in ms to wait
    */
-  public void setTimer(String version, long timeout) {
-    debounceTimer.scheduleAfterDebounceTime("VersionUpgradeTimeout", timeout, ()->timerOff(version));
+  public void setTimer(final String version, final long timeout, final Stat currentStatOfBarrierDone) {
```
sborya: fixed.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: There is no single list. These values are used to identify the debounced calls, so the previous ones can be cancelled when a new one is registered. These values come from the callers/users of the debounce class, and should be defined in their scope. I can add a constant here.

samza-core/src/main/java/org/apache/samza/zk/BarrierForVersionUpgrade.java
sborya: Well, the original name was startBarrier() - but you guys, asked to change it.

navina: Yeah. it can still be prepare() if you don't like prepareBarrier. Anyway, I will let you make the final call :)

navina:
+1 lgtm!

PR #82, opened 2017-03-11
SAMZA-1135: Support Scala 2.12

logarithm:
Now days more and more projects are using Scala 2.12 and Samza as a cutting edge project should support the latest version of language. Samza codebase is almost ready to update to scala 2.12. The main thing which has to be done is removing deprecated JavaConversions. As to dependencies, only kafka and scalatest should be updated, because current versions don't have build for Scala 2.12.


vjagadish1989:
Thank you for the patch!

@logarithm Can we please open a [DISCUSS] thread in the open source mailing list? IMHO, it'd be a good idea to seek comments from more stakeholders. 

In addition to this, it has other ramifications like for example, the name of the binaries that a Samza release publishes to the maven artifactory.

logarithm:
@vjagadish1989 yeah, sure! Is it enough just write an email to dev@samza.apache.org for opening discussion?

As to artifact names: as far as I understand scala artifacts for samza with scala 2.12 will have another postfix, but we always can build artifacts for other versions.

vjagadish1989:
Yes, please open an [DISCUSS] thread by sending email to dev@samza.apache.org . Thanks alot!

logarithm:
@vjagadish1989 I've sent email to dev@samza.apache.org but I don't see it in Mailing list archive. Should it be approved somehow before it appears in the list?

vjagadish1989 (approved):
looks good to me.

approved.

logarithm:
@vjagadish1989 Should I rebase it agains master, since it has conflict now?

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -324,21 +324,21 @@ object Util extends Logging {
...
   def javaMapAsScalaMap[T, K](javaMap: java.util.Map[T, K]): Map[T, K] = {
-    javaMap.toMap
+    javaMap.asScala.toMap
   }
 
   /**
    * Returns the the first host address which is not the loopback address, or {@link java.net.InetAddress#getLocalHost InetAddress.getLocalhost()} as a fallback
    *
-   * @return the {@link java.net.InetAddress InetAddress} which represents the localhost
+   * @return the {java.net.InetAddress InetAddress} which represents the localhost
```
prateekm: Accidental change?

samza-core/src/main/java/org/apache/samza/container/RunLoopFactory.java
```diff
@@ -98,7 +98,7 @@ public Boolean apply(TaskInstance t) {
       log.info("Run loop in asynchronous mode.");
 
       return new AsyncRunLoop(
-        JavaConversions.mapAsJavaMap(taskInstances),
+        JavaConverters.mapAsJavaMapConverter(taskInstances).asJava(),
```
prateekm: I think `JavaConverters#mapAsJavaMap` etc. should work for conversions like these in Java code. If true, let's change all other occurrences too.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -19,19 +19,12 @@
...
-import org.apache.samza.system.SystemStream
-import org.apache.samza.system.SystemStreamMetadata
-import org.apache.samza.system.SystemStreamPartition
-import org.apache.samza.system.IncomingMessageEnvelope
 import org.apache.samza.SamzaException
-import org.apache.samza.util.Logging
-import org.apache.samza.metrics.MetricsHelper
-import org.apache.samza.metrics.MetricsRegistryMap
-import org.apache.samza.metrics.MetricsRegistry
+import org.apache.samza.metrics.{MetricsHelper, MetricsRegistry, MetricsRegistryMap}
```
prateekm: Minor: Prefer keeping these as explicit imports (not  `_`) and on separate lines.

samza-kafka/src/test/scala/org/apache/samza/config/TestKafkaConfig.scala
```diff
@@ -52,7 +48,7 @@ class TestKafkaConfig {
     val factory = new PropertiesConfigFactory()
     props.setProperty("systems." + SYSTEM_NAME + ".samza.factory", "org.apache.samza.system.kafka.KafkaSystemFactory")
 
-    val mapConfig = new MapConfig(props.toMap[String, String])
+    val mapConfig = new MapConfig(props.asScala.toMap[String, String].asJava)
```
prateekm: Can do `props.asScala.asJava` (here and rest of the class).

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -172,7 +172,8 @@ class OffsetManager(
   val systemStreamPartitions = mutable.Map[TaskName, mutable.Set[SystemStreamPartition]]()
 
   def register(taskName: TaskName, systemStreamPartitionsToRegister: Set[SystemStreamPartition]) {
-    systemStreamPartitions.getOrElseUpdate(taskName, mutable.Set[SystemStreamPartition]()).addAll(systemStreamPartitionsToRegister)
+    systemStreamPartitions.getOrElseUpdate(taskName, mutable.Set[SystemStreamPartition]()).asJava
```
logarithm: Yeah, good catch. But in this PR what I exactly done - just replaced implicit conversions by calling to JavaConverters. Any weird conversions means that it was weird implicit conversion before. Do you think it's better to update all these places in this PR or by the new one?

prateekm: Agree, let's leave them as-is in general. This and a couple of others I mentioned just seem awkward enough that I don't mind updating them in this PR, esp. since the fix is simple.

We can also do a follow up cleanup PR for these, up to you.

samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
```diff
@@ -92,8 +92,8 @@ class JobRunner(config: Config) extends Logging {
 
     val oldConfig = coordinatorSystemConsumer.getConfig()
     if (resetJobConfig) {
-      info("Deleting old configs that are no longer defined: %s".format(oldConfig.keySet -- config.keySet))
-      (oldConfig.keySet -- config.keySet).foreach(key => {
+      info("Deleting old configs that are no longer defined: %s".format(oldConfig.keySet.asScala -- config.keySet.asScala))
```
logarithm: I've checked `MapConfig` and found that sometimes `MapConfig` class might be backed by immutable Map. So, now I convert it to immutable map and find a diff.

samza-core/src/main/scala/org/apache/samza/util/Util.scala
logarithm: Idea highlighted it as `Unknown scaladoc tag @link`. Reverted back.

samza-core/src/main/java/org/apache/samza/container/RunLoopFactory.java
logarithm: `mapAsJavaMap` won't work in scala 2.10 because this method was introduced in scala 2.11. If we want to keep backward compatibility with scala 2.10 we can't use it.

logarithm:
@prateekm Thanks for the review. I addressed all your comments and rebased branch agains master. Could you please take a look?

samza-core/src/main/java/org/apache/samza/container/RunLoopFactory.java
prateekm: Ah ok, makes sense then.

PR #84, opened 2017-03-13
SAMZA-1138: Yarn capability check is broken

logarithm:
After migration from `yarn.container.*` properties to `cluster-manager.container.*` properties we have to use either of them and `ClusterManagerConfig` provides backward compatibility for these properties. But in `YarnClusterResourceManage`r only old properties are used (from `YarnConfig`), hence if job config migrated to new `cluster-manager.*` properties names then check will be evaluated against default values, not against actual values.

### DIFF
--- a/samza-yarn/src/main/java/org/apache/samza/config/YarnConfig.java
+++ b/samza-yarn/src/main/java/org/apache/samza/config/YarnConfig.java
@@ -27,40 +27,15 @@ public class YarnConfig extends MapConfig {
    */
   public static final String PACKAGE_PATH = "yarn.package.path";
 
-  // Configs related to each yarn container
-  /**
-   * Memory, in megabytes, to request from YARN per container
-   */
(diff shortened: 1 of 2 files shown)


vjagadish1989 (approved):
approved. looks good to me. 

Thanks for the contribution!  We'll include this in the upcoming release. 

Final f/b from my review:
For others who don't have much context on Yarn internals:
- The capabilities check validates whether there are enough resources in the cluster to run your job and this patch fixes it by parsing the new `cluster-manager.*` style configs. 
- The resource request mechanism to Yarn is not changing. We will continue to honor `cluster-manager.*` configs when making resource requests. (as is being done currently)

logarithm:
@vjagadish1989 Thanks for the review! Can we merge and close this PR?

vjagadish1989:
@logarithm Thank you for the patch, committed! Looking forward to more contributions :-)

logarithm:
@vjagadish1989 Thanks!

PR #85, opened 2017-03-14
SAMZA-1140 : Non blocking commit in Async Runloop

shanthoosh:
Adds non blocking commit into AsyncRunLoop. Clients can enable it by setting an opt-in config task.async.commit (which is disabled by default). Contains the following list of changes for AsyncStreamTask type of tasks.

a) Enable commit for a task,  when there're uncompleted callbacks from a task. Progressive commit of finished callbacks, when there are messages in flight for a task.
b) Enable processing of messages for a task, when there're commits in progress from a task.

Documentation for this change will be done in a separate PR.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
+++ b/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
@@ -74,6 +74,7 @@ public class AsyncRunLoop implements Runnable, Throttleable {
   private volatile boolean shutdownNow = false;
   private volatile Throwable throwable = null;
   private final HighResolutionClock clock;
+  private final boolean isAsyncCommitEnabled;
 
   public AsyncRunLoop(Map<TaskName, TaskInstance> taskInstances,
       ExecutorService threadPool,
@@ -84,7 +85,8 @@ public class AsyncRunLoop implements Runnable, Throttleable {
       long callbackTimeoutMs,
       long maxThrottlingDelayMs,
       SamzaContainerMetrics containerMetrics,
-      HighResolutionClock clock) {
+      HighResolutionClock clock,
+      boolean isAsyncCommitEnabled) {
 
     this.threadPool = threadPool;
     this.consumerMultiplexer = consumerMultiplexer;
@@ -106,6 +108,7 @@ public class AsyncRunLoop implements Runnable, Throttleable {
     // Partions and tasks assigned to the container will not change during the run loop life time
     this.sspToTaskWorkerMapping = Collections.unmodifiableMap(getSspToAsyncTaskWorkerMap(taskInstances, workers));
     this.taskWorkers = Collections.unmodifiableList(new ArrayList<>(workers.values()));
+    this.isAsyncCommitEnabled = isAsyncCommitEnabled;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


shanthoosh:
Last test in TestAsyncRunLoop is broken here.  After finishing the test/assertions successfully, the runLoop doesn't terminate. I'm working on fixing it.

shanthoosh:
@nickpan47 : Can you please take a look when you get a chance. Build failure is due to hudson.plugins.git.GitException(not due to tests failing, build's green in my devbox). Build machine is not able to pull this patch for some reason.

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -611,14 +616,30 @@ private boolean isReady() {
...
+       * b) When task.async.commit is true and window, commit are not in progress.
+       */
+      if (needCommit) {
+        return (messagesInFlight.get() == 0 || isAsyncCommitEnabled) && !windowOrCommitInFlight;
+      } else if (needWindow || endOfStream) {
+        /*
+         * A task is ready for window operation, when task.window(needWindow) is requested by either user or window thread
+         * and window, commit are not in progress.
+         */
         return messagesInFlight.get() == 0 && !windowOrCommitInFlight;
```
nickpan47: Question: if we need messagesInFlight.get() == 0, that would mean that the commit will have to wait till all in-flight messages to be completed? Wouldn't it still cause bursts in message process rate? I thought that we need to allow commit and process can always be scheduled simultaneously in order to perform async commit? The current logic seems to allow process() to be scheduled when async commit is in-flight, but not the other way. Wouldn't that cause stopage in processing when a commit is waiting (i.e. needCommit == true)? I don't think that this would have significant reduction of burstiness, since the difference before and after this change is simply more process() can be scheduled when commit is in flight. However, the number of commits are low and the latency of a single commit is not large. The burstiness caused by stopping the commit and process while waiting the inflight process() to be completed is not reduced, which is the main cause of the burstiness.

I thought that the right approach is to:
1) allow process and commit to always be scheduled simultaneously
2) bookkeeping the low watermark of each input SSP's offsets while callbacks are triggered and only checkpoint if the low watermark proceeds in commit

Happy to discuss in person.

shanthoosh: @nickpan47 : 

If I understand the comment correctly, it's pointed out that the only process calls are allowed to go through when commit is in progress and commit is not allowed to go through when messagesInFlight.get() > 0(process is in progress) which is the main cause for burst in message process rate. 

I think the check on line 628 does that and answers the question. If needCommit is true, if either asyncCommit is enabled and neither window  nor commit is in flight, commit is chosen as next step for the task. By that this will allow the commit to scheduled simultaneously when process calls are in progress. This will reduce the burstiness caused by stopping the commit when messagesInFlight > 0(Point 1. will be accomplished by this.)

-- Point 2: bookkeeping the low watermark of each input SSP's offsets 
TaskCallBackManager(existing implementation) which handles the offsets from the completed callbacks, updates the OffsetManager with the highest sequence number message that is processed & committed by the task. This logic is agnostic of input ssp, amongst all input SSP's highest sequence number  of message that is committed & processed by task is chosen for commit. If I understand this point correctly(picking lowest offset for each input ssp for commit), what you're asking for is done in PR 87.

Sure, let's discuss this in person. Please point out if i'm misunderstanding something.

nickpan47: Had a offline discussion and consider this comment resolved. Thanks!

PR #86, opened 2017-03-15
SAMZA-1145: Provide Ability To Confgure The Default Number Of Changel…

jwlent55:
…og Replicas

### DIFF
--- a/docs/learn/documentation/versioned/jobs/configuration-table.html
+++ b/docs/learn/documentation/versioned/jobs/configuration-table.html
@@ -1272,9 +1272,17 @@
 
                 <tr>
                     <td class="property" id="store-changelog-replication-factor">stores.<span class="store">store-name</span>.changelog.<br>replication.factor</td>
-                    <td class="default">2</td>
+                    <td class="default">stores.default.changelog.replication.factor</td>
                     <td class="description">
                         The property defines the number of replicas to use for the change log stream.
+                    </td>
+                </tr>
+
+                <tr>
+                    <td class="property" id="store-default-changelog-replication-factor">stores.default.changelog.replication.factor</td>
+                    <td class="default">2</td>
+                    <td class="description">
+                        This property defines the default number of replicas to use for the change log stream.
                     </td>
                 </tr>
 
@@ -1343,8 +1351,9 @@
                     <td class="description">
                         This property defines a store, Samza's mechanism for efficient
                         <a href="../container/state-management.html">stateful stream processing</a>. You can give a
-                        store any <span class="store">store-name</span>, and use that name to get a reference to the
-                        store in your stream task (call
+                        store any <span class="store">store-name</span> except <em>default</em> (the <span class="store">store-name</span>
+                        <em>default</em> is reserved for defining default store parameters), and use that name to get a 
+                        reference to the store in your stream task (call
                         <a href="../api/javadocs/org/apache/samza/task/TaskContext.html#getStore(java.lang.String)">TaskContext.getStore()</a>
                         in your task's
                         <a href="../api/javadocs/org/apache/samza/task/InitableTask.html#init(org.apache.samza.config.Config, org.apache.samza.task.TaskContext)">init()</a>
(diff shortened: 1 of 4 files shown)


jwlent55:
Not sure how to debug the "samza-pr-checks" failure.  The key lines from the Jenkins log appear to be:

Caused by: hudson.plugins.git.GitException: Command "git checkout -b master origin/master" returned status code 1:
stdout: samza-core/src/main/java/org/apache/samza/operators/util/InternalInMemoryStore.java: needs merge

That file was added by a commit that I think precedes my commit:

jwlent55:
Thanks for your quick response.  Our use case is that we have many stores / changelog topics and two typical configurations: simple test deployments with a single Kafka node (and therefore want a single changelog replica) and production (and more complex test) )deployments where we currently configure three replicas.  The system default is two.  Therefore we currently need to explicitly list each of the changelog topic stores in our configuration files or they inadvertently default to two.  Easier for us would be to configure a single changelog replica default and only if necessary configure a store specific one.  In other words our desired number of replicas does not currently vary by store, but, instead by deployment topology.

I put some of this information in the JIRA, but, failed to copy it into the pull request.

vjagadish1989 (approved):
looks great, approved. thank you for the contribution

Would be nice to add an unit test to quickly verify correct defaults - across default configs, overriden configs.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1272,13 +1272,21 @@
 
                 <tr>
                     <td class="property" id="store-changelog-replication-factor">stores.<span class="store">store-name</span>.changelog.<br>replication.factor</td>
-                    <td class="default">2</td>
+                    <td class="default">stores.changelog.replication.factor.default</td>
```
nickpan47: prefer to change it to:
stores.changelog.default.replication.factor
It would be easier to add more default configurations for the stores.changelog configuration later.

jwlent55: I originally considered that naming convention, but, I was paranoid.  What if someone named a store "default"?  I did not want to suggest a change that could adversely impact an existing deployment.  That is why I placed the "default" at the end.  A compromise might be something like:
    stores.__default__.changelog.replication.factor
Still possible for a name collision, but, a lot more unlikely (and potentially deserved).

Paranoia aside, I am wide open to suggestions on the naming convention to use.  I just want to get a name established so that we can patch that name back to one of our private branches based on 0.11 or 0.12.

jwlent55: I have amended the commit to include two unit tests to test the stores.changelog.replication.factor.default configuration parameter.

jwlent55:
Anything left for me to do?  Happy to make more changes.  I realize this is a low priority issue - just a configuration convenience for our specific use case.

docs/learn/documentation/versioned/jobs/configuration-table.html
nickpan47: @jwlent55 sorry for the late reply. I would still prefer to have "default" right after the stores. It would be a documentation effort to declare that "default" is a reserved word in Samza configuration (i.e. this would be a pattern we will use for other plural nouns like systems/streams for default as well). I don't think that putting the default at the end will avoid that problem (e.g. what if someone have a store named "changelog"?). Besides, by putting the default as a "scope", we avoided the adhoc addition of default configuration variables and group them into one scope. Thanks!

jwlent55: No problem on the late reply.  This is a pretty low priority request.

You suggestion works for me if there is a broad consensus that the risk that someone currently uses a a store name of "default" is acceptably low.  We certainly have no stores with that name.

Not sure how to document "default" as a reserved word.  I will look around, but, any pointers would be appreciated.

nickpan47: Hi, @jwlent55 thanks! The document I referred to is the web document: http://samza.apache.org/learn/documentation/0.12/jobs/configuration-table.html

The doc is located and updated via ${SAMZA_ROOT}/docs/learn/documentation/versioned/jobs/configuration-table.html 

Hope that helps.

jwlent55: I have amended the commit to use the naming convention you prefer.  I hope that amending was the correct approach (as opposed to a second commit).  I also tried to add some documentation to indicate the the store name default was reserved for defining default store parameters.

jwlent55:
Looks like I need to rebase my change.  Builds fine in my dev branch, but, when I apply the changes to the latest version of master I can reproduce the problem above.  Investigating.

PR #87, opened 2017-03-15
SAMZA-1146: TaskCallbackManager commit fix.

shanthoosh:
Each task callback in samza belongs to different SystemStreamPartition. When multiple callbacks in contagious order are available for commit, callback with highest sequence number is chosen for commit. This will prevent checkpointing of completed callbacks that has commit request and doesn't have highest sequence number. Upon task restart this will lead to duplicate reprocessing of already processed messages (since completed callbacks for some SystemStreamPartition's aren't committed earlier).

This PR fixes it and commits all completed callbacks that has commit request defined. Added a test to verify the behavior.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/task/TaskCallbackManager.java
+++ b/samza-core/src/main/java/org/apache/samza/task/TaskCallbackManager.java
@@ -19,6 +19,9 @@
 
 package org.apache.samza.task;
 
+import com.google.common.collect.ImmutableList;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.PriorityQueue;
 import java.util.Queue;
 import java.util.concurrent.ScheduledExecutorService;
@@ -46,40 +49,22 @@ class TaskCallbackManager {
      * Adding the newly complete callback to the callback queue
      * Move the queue to the last contiguous callback to commit offset
      * @param cb new callback completed
-     * @return callback of highest watermark needed to be committed
+     * @return list of callbacks to be committed
      */
-    TaskCallbackImpl update(TaskCallbackImpl cb) {
+    List<TaskCallbackImpl> update(TaskCallbackImpl cb) {
       synchronized (lock) {
         callbacks.add(cb);
-
-        TaskCallbackImpl callback = null;
-        TaskCallbackImpl callbackToCommit = null;
-        TaskCoordinator.RequestScope shutdownRequest = null;
+        List<TaskCallbackImpl> callbacksToUpdate = new ArrayList<>();
         // look for the last contiguous callback
         while (!callbacks.isEmpty() && callbacks.peek().matchSeqNum(nextSeqNum)) {
           ++nextSeqNum;
-          callback = callbacks.poll();
-
+          TaskCallbackImpl callback = callbacks.poll();
+          callbacksToUpdate.add(callback);
           if (callback.coordinator.commitRequest().isDefined()) {
-            callbackToCommit = callback;
-          }
-
-          if (callback.coordinator.shutdownRequest().isDefined()) {
-            shutdownRequest = callback.coordinator.shutdownRequest().get();
+            break;
           }
         }
-
-        // if there is no manual commit, use the highest contiguous callback message offset
-        if (callbackToCommit == null) {
-          callbackToCommit = callback;
-        }
-
-        // if there is a shutdown request, merge it into the coordinator to commit
-        if (shutdownRequest != null) {
-          callbackToCommit.coordinator.shutdown(shutdownRequest);
-        }
-
-        return callbackToCommit;
+        return callbacksToUpdate;
       }
     }
   }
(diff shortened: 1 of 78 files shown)


shanthoosh:
@nickpan47 : Can you please take a look, when you get chance.

samza-core/src/main/java/org/apache/samza/task/TaskCallbackManager.java
```diff
@@ -46,22 +49,22 @@
...
      */
-    TaskCallbackImpl update(TaskCallbackImpl cb) {
+    List<TaskCallbackImpl> update(TaskCallbackImpl cb) {
+
       synchronized (lock) {
         callbacks.add(cb);
-
         TaskCallbackImpl callback = null;
-        TaskCallbackImpl callbackToCommit = null;
+        List<TaskCallbackImpl> committableCallbacks = new ArrayList<>();
```
xinyuiscool: Actually we don't need to have a separate commitableCallbacks. We need to update for all the callbacks below the low watermark. So this should just be callbacksToUpdate or something

samza-core/src/main/java/org/apache/samza/task/TaskCallbackManager.java
```diff
@@ -46,22 +49,22 @@
...
         TaskCoordinator.RequestScope shutdownRequest = null;
         // look for the last contiguous callback
         while (!callbacks.isEmpty() && callbacks.peek().matchSeqNum(nextSeqNum)) {
           ++nextSeqNum;
           callback = callbacks.poll();
 
           if (callback.coordinator.commitRequest().isDefined()) {
```
xinyuiscool: no need to check. Add the callback to the list.

samza-core/src/main/java/org/apache/samza/task/TaskCallbackManager.java
```diff
@@ -70,16 +73,18 @@ TaskCallbackImpl update(TaskCallbackImpl cb) {
         }
 
         // if there is no manual commit, use the highest contiguous callback message offset
-        if (callbackToCommit == null) {
-          callbackToCommit = callback;
+        if (committableCallbacks.isEmpty() && callback != null) {
```
xinyuiscool: no need to do this.

samza-core/src/main/java/org/apache/samza/task/TaskCallbackManager.java
```diff
@@ -46,22 +49,22 @@
...
         while (!callbacks.isEmpty() && callbacks.peek().matchSeqNum(nextSeqNum)) {
           ++nextSeqNum;
           callback = callbacks.poll();
 
           if (callback.coordinator.commitRequest().isDefined()) {
-            callbackToCommit = callback;
+            committableCallbacks.add(callback);
           }
 
           if (callback.coordinator.shutdownRequest().isDefined()) {
```
xinyuiscool: no need to do this anymore.

samza-core/src/main/java/org/apache/samza/task/TaskCallbackManager.java
```diff
@@ -70,16 +73,18 @@ TaskCallbackImpl update(TaskCallbackImpl cb) {
...
 
         // if there is no manual commit, use the highest contiguous callback message offset
-        if (callbackToCommit == null) {
-          callbackToCommit = callback;
+        if (committableCallbacks.isEmpty() && callback != null) {
+          committableCallbacks.add(callback);
         }
 
         // if there is a shutdown request, merge it into the coordinator to commit
         if (shutdownRequest != null) {
```
xinyuiscool: no need to do this.

samza-core/src/main/java/org/apache/samza/task/TaskCallbackManager.java
```diff
@@ -46,22 +49,22 @@
...
         callbacks.add(cb);
-
         TaskCallbackImpl callback = null;
-        TaskCallbackImpl callbackToCommit = null;
+        List<TaskCallbackImpl> committableCallbacks = new ArrayList<>();
         TaskCoordinator.RequestScope shutdownRequest = null;
         // look for the last contiguous callback
         while (!callbacks.isEmpty() && callbacks.peek().matchSeqNum(nextSeqNum)) {
           ++nextSeqNum;
           callback = callbacks.poll();
```
xinyuiscool: Seems to me the logic in this while loop should be extremely simple now. Just do 

callbacksToUpdate.add(callback)

and that should be it. Please think about it.

samza-core/src/main/java/org/apache/samza/task/TaskCallbackManager.java
shanthoosh: Ranemed to callbacksToUpdate.

samza-core/src/main/java/org/apache/samza/task/TaskCallbackManager.java
shanthoosh: Yes, this is fixed. 
I'd misunderstanding that task.commit is for commit in only the callback  SSP that is processed & committed by the task. Thanks for pointing it out that it's for all the input SSP's that are processed until the callback SSP that's processed & committed.

samza-core/src/main/java/org/apache/samza/task/TaskCallbackManager.java
shanthoosh: Removed.

samza-core/src/main/java/org/apache/samza/task/TaskCallbackManager.java
shanthoosh: Removed.

samza-core/src/main/java/org/apache/samza/task/TaskCallbackManager.java
shanthoosh: Removed.

prateekm:
@shanthoosh Just FYI, you should rebase from upstream master before updating the PR so that upstream commits don't show up as part of your changes.

shanthoosh:
@prateekm : 

I did a bad rebase because of which it showed up as a part of my PR.  I will be careful from now on.

PR #88, opened 2017-03-16
SAMZA-1131: RemoteApplicationRunner for cluster-based Samza applications

xinyuiscool:
RemoteApplicationRunner starts the Samza StreamApplication on the remote cluster, e.g. Yarn. It uses ExecutionPlanner for planning physical execution, and JobRunner to start each stage of the application.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
@@ -26,7 +26,7 @@ import org.apache.samza.system.StreamSpec;
 
 public abstract class AbstractApplicationRunner implements ApplicationRunner {
 
-  private final Config config;
+  protected final Config config;
 
   public AbstractApplicationRunner(Config config) {
     if (config == null) {
--- a/samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.runtime;
+
+import joptsimple.OptionSet;
+import org.apache.samza.config.Config;
+import org.apache.samza.job.JobRunner;
+import org.apache.samza.operators.StreamGraphBuilder;
+import org.apache.samza.util.CommandLine;
+import org.apache.samza.util.Util;
+
+
+/**
+ * This is the main() entry point to run a Samza application.
+ * For a StreamApplication, it creates the {@link ApplicationRunner} based on the config, and then run the application.
+ * For a Samza job using low level task API, it will create the JobRunner to run it.
+ */
+public class ApplicationRunnerMain {
+  // TODO: have the app configs consolidated in one place
+  private static final String STREAM_APPLICATION_CLASS_CONFIG = "'app.class";
+
+  public static void main(String[] args) throws Exception {
+    CommandLine cmdLine = new CommandLine();
+    OptionSet options = cmdLine.parser().parse(args);
+    Config config = cmdLine.loadConfig(options);
+
+    if (config.containsKey(STREAM_APPLICATION_CLASS_CONFIG)) {
+      ApplicationRunner runner = ApplicationRunner.fromConfig(config);
+      StreamGraphBuilder app = (StreamGraphBuilder) Class.forName(config.get(STREAM_APPLICATION_CLASS_CONFIG)).newInstance();
+      runner.run(app);
+    } else {
+      new JobRunner(Util.rewriteConfig(config)).run(true);
+    }
+  }
+}
--- a/samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
@@ -28,7 +28,7 @@ import org.apache.samza.util.Util;
 
 
 /**
- * This is the main() entry point to run a Samza application.
+ * This class contains the main() method used by run-app.sh.
  * For a StreamApplication, it creates the {@link ApplicationRunner} based on the config, and then run the application.
  * For a Samza job using low level task API, it will create the JobRunner to run it.
  */
--- a/samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
@@ -16,10 +16,19 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
 package org.apache.samza.runtime;
 
+import org.apache.samza.SamzaException;
+import org.apache.samza.execution.ExecutionPlanner;
+import org.apache.samza.execution.ProcessorGraph;
+import org.apache.samza.job.JobRunner;
+import org.apache.samza.operators.StreamGraph;
 import org.apache.samza.operators.StreamGraphBuilder;
 import org.apache.samza.config.Config;
+import org.apache.samza.operators.StreamGraphImpl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 /**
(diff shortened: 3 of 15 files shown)


xinyuiscool:
@navina @jmakes : please take a look. Thanks!

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.samza.util.Util;
+
+
+/**
+ * This is the main() entry point to run a Samza application.
+ * For a StreamApplication, it creates the {@link ApplicationRunner} based on the config, and then run the application.
+ * For a Samza job using low level task API, it will create the JobRunner to run it.
+ */
+public class ApplicationRunnerMain {
+  // TODO: have the app configs consolidated in one place
```
jmakes: There's already a JIRA for this?

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
```diff
@@ -0,0 +1,52 @@
...
+  private static final String STREAM_APPLICATION_CLASS_CONFIG = "'app.class";
```
jmakes: Would "app.builder.class" be more accurate?

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -16,27 +16,56 @@
...
+  public void run(StreamGraphBuilder app) {
+    try {
+      // 1. build stream graph
+      StreamGraph streamGraph = new StreamGraphImpl(this, config);
+      app.init(streamGraph, config);
 
+      // 2. create the physical execution plan
+      ExecutionPlanner planner = new ExecutionPlanner(config);
+      ProcessorGraph processorGraph = planner.plan(streamGraph);
+
```
jmakes: I expected stream creation here. Where is that happening?

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -26,7 +26,7 @@
 
 public abstract class AbstractApplicationRunner implements ApplicationRunner {
```
navina: I have a general question, not necessarily related to this PR. JDK8 allows interface methods to have default implementations. What is the standard criteria to decide when the default implementation belongs to the interface and when to put it in an abstract class?

The reason I ask is that AbstractApplicationRunner provides an implementation of streamFromConfig. Does that mean we allow classes derived from AbstractApplicationRunner to override this method? Is there a use-case for this? 
Nothing stops the user from directly implementing the ApplicationRunner interface or extending AbstractApplicationRunner. So, I am trying to understand the motivation behind this choice. 

_Clearly, I didn't review the PR that committed this changed. So sorry about asking questions this late_

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
```diff
@@ -0,0 +1,52 @@
...
+
+  public static void main(String[] args) throws Exception {
```
navina: I thought user will directly use ApplicationRunner.get{Local|Remote}Runner().run() ?? 
What is the need for ApplicationRunnerMain? I thought ApplicationRunner will have the logic to figure out if it should use JobRunner or not. Also, as far as JobRunner goes, it is simply used to launch jobs in a given environment. 

Seems like we are doing this simply because ApplicationRunner interface's run method takes a StreamGraphBuilder. If that is the interface we want, shouldn't we adapt the low-level taskAPI to a StreamGraph?

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -16,27 +16,56 @@
...
+      // 3. submit jobs for remote execution
+      processorGraph.getProcessors().forEach(processor -> {
```
navina: ProcessorGraph consists of ProcessorNode(s). Can we change this method name to getProcessorsNodes()? It will be more clear.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -16,27 +16,56 @@
...
+          Config processorConfig = processor.generateConfig();
+          log.info("Starting processor {} with config {}", processor.getId(), config);
+          JobRunner runner = new JobRunner(processorConfig);
```
navina: This remote runner can start acting weird or most likely, fail if the user has configured "job.factory.class". This seems totally possible considering the number of configs we require from the user :)

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
jmakes: > Clearly, I didn't review the PR that committed this changed. So sorry about asking questions this late
Yes you did: https://github.com/apache/samza/pull/74

> So, I am trying to understand the motivation behind this choice.
There were a few reasons, but the primary one is that if the streamFromConfig() method were implemented as a default in the interface, it would require the Config as a parameter and since all the Runners take config as a constructor parameter, this looks clunky to the user. It's better to have them provide the config only once.

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
xinyuiscool: 1. this is going to be the main class running in Yarn RM box for remote case, and can also be the main class running locally with Zk. The runner will be created by config, like ApplicationRunner runner = ApplicationRunner.fromConfig(config). So by configuring different application runner, we can run the same job in either yarn or locally. There is no change in code required.

2. I think we discussed about this. So the plan will be that ApplicationRunner will be able to run StreamTask directly, so no adapter needed. But that work is done yet.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
xinyuiscool: It's supposed to happen inside that app.init(). It's not finalized yet. So right now the app cannot use the stream creation inside the runner.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
xinyuiscool: Sure. Fixed it.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
navina: > Yes you did: #74

Umm..My bad.. I didn't realize ExecutionEnvironment PR is same as the one ApplicationRunner introduced. Anyway, I have some questions now.

> since all the Runners take config as a constructor parameter, 

Is that assumption right? Because if we expect all Runner implementations to contain state (as in Config), then why not just have ApplicationRunner as an abstract class? We can fully get rid of the interface and use a final method for streamFromConfig. 

If we do decide to go with an interface for ApplicationRunner, it is pretty misleading for the user to call streamFromConfig when the expected parameter is streamId.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
xinyuiscool: I invoke my fifth amendment right again (SAMZA-1120)! This ticket is going to be my passport for any config issues :).

Yes, right now the remote runner still allows those two factories so we can run it in local process mode. I add this to the config ticket so we will take a look at this issue when we work on it.

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
navina: Ok. Got it. 
Can you please change the comment on Line 31 to say this is the main method used by run-app.sh, instead of main() entry point to Samza application? That way, you make it clear that it will not be used in the embedded use-case.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
navina: lol.. Ok I don't know how big the scope of SAMZA-1120. That only talks about the scope of configs. My concern here is regarding validation. 
Please think about how we can avoid this issue now. I think it is better to add some basic validations here.

This is why I was asking you about the responsibilities of JobRunner and ApplicationRunner. :) 
 Based on my understanding, ApplicationRunner is the way to launch one or more jobs depending on the Stream Application (even though today, it is all in a single job. And JobRunner is the old generation way of launching a single Samza job. Whether it is remote or local is controlled by job.factory.class, even though it only loosely addresses the semantics of remote and local.

navina: On an unrelated note, I noticed accidentally that job.factory.class is used in JobModelManager is some weird way (See Line 159 in JobModelManager.scala). Please make sure your changes don't impact it or the other way around :)

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
jmakes: Nah, I forgot about the rename. That works. Thanks!

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
jmakes: No need to hold up this PR, but we should discuss this. I didn't realize init would cause any external side effects, and I'm not sure how that will work with the stream/store reset functionality for batch.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
jmakes: Yes the assumption is right. 

Abstract classes and interfaces are not mutually exclusive. For example see:
http://docs.oracle.com/javase/8/docs/api/java/util/AbstractMap.html

However, I will take a second look at this case tomorrow, to ensure that the interface is still useful. If not, I'll open a separate PR to remove it. 

Excellent point about the method name. streamFromConfig was the original name when the method took config as a parameter. I failed to rename it after the refactor. Will fix. Thanks!

navina: > Abstract classes and interfaces are not mutually exclusive. For example see:

I didn't imply they are mutually exclusive. I am just saying it is a pattern that should be used with care. Whenever possible, we should simplify it. 

>However, I will take a second look at this case tomorrow, to ensure that the interface is still useful

Thank you!

samza-shell/src/main/bash/run-app.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
vjagadish1989: We already have a `run-job.sh` script right? How does this script relate to that? Is the intention to subsume functionality in `run-job.sh` into this, can't we remove `run-job.sh`? Or is that the end-goal?

vjagadish1989 (approved):
approved. looks good to me. thanks for contributions. 

final round of feedback/questions from my review.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
xinyuiscool: Oh, @jmakes: I misunderstood your question, I thought you meant stream creation as use apprunner to create StreamSpec. The stream creation is temporarily hosted inside the planner.plan() right now. I haven't separated out yet. @navina also pointed out. I will do a refactoring and separate out stream creation when we have the LocalApplicationRunner pr.

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
xinyuiscool: Good catch! Fixed it.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
xinyuiscool: Oh, I think I misunderstood your question. I thought you were asking about StreamSpec creation. Currently stream creation is temporarily hosted inside planner.plan(). I plan to make a refactoring and separate it out when working on the RemoteApplicationRunner PR. @navina also pointed this out.

samza-shell/src/main/bash/run-app.sh
xinyuiscool: run-app.sh will replace run-job.sh going forward. But for legacy jobs, we can either keep run-job.sh or force them to upgrade run script. I would prefer keeping it for now so it won't cause panic for our users :)

PR #90, opened 2017-03-16
SAMZA-1143 Universal config support for localized resource

fredji97:
More details in https://issues.apache.org/jira/browse/SAMZA-1143

Tests: ./gradlew clean check successful and all unit tests passed


samza-yarn/src/main/java/org/apache/samza/job/yarn/LocalizerResourceMapper.java
```diff
@@ -0,0 +1,114 @@
...
+  private YarnConfiguration yarnConfiguration; //yarn configurations
+  private Map<String, LocalResource> localResourceMap;
+
+  public LocalizerResourceMapper(Config config, YarnConfiguration yarnConfiguration) {
+    this.config = config;
+    this.yarnConfiguration = yarnConfiguration;
+    this.localResourceMap = new HashMap<>();
+  }
+
+  public void map() throws LocalizerResourceException {
```
vjagadish1989: Don't need an explicit `map` here. I wonder if we can use invoke `map` internally inside the constructor. 

- Not sure I see a use-case to invoke `map` multiple times in this class.
- We can make `localResourceMap` immutable

samza-yarn/src/main/java/org/apache/samza/job/yarn/LocalizerResourceMapper.java
```diff
@@ -0,0 +1,114 @@
...
+      localResource.setVisibility(resourceVisibility);
+      return localResource;
+    } catch (IOException ioe) {
+      log.error("IO Exception when accessing the resource file status from the filesystem", ioe);
+      throw new LocalizerResourceException("IO Exception when accessing the resource file status from the filesystem");
+    }
+
+  }
+
+  public Map<String, LocalResource> getResourceMap() {
```
vjagadish1989: 1. I wonder if this class can instead be made an util / or a function. 
2. Should   we return an `Immutable` collection or a copy here? That way, external callers don't mutate the map, and this class is immune to those changes.

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnClusterResourceManager.java
```diff
@@ -120,6 +119,11 @@ public YarnClusterResourceManager(Config config, JobModelManager jobModelManager
     hConfig = new YarnConfiguration();
     hConfig.set("fs.http.impl", HttpFileSystem.class.getName());
 
+    if (config.containsKey("fs.certfs.impl.override")) { //TODO: change to use constants from CertFSConstants once that CertFSConstants.java is in
```
vjagadish1989: Perhaps, I'm missing something. Why do we need 2 configs - one for `fs.certfs.impl` and the `override` config?

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnContainerRunner.java
```diff
@@ -47,6 +49,8 @@
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.*;
```
vjagadish1989: remove blanket imports

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnContainerRunner.java
```diff
@@ -190,13 +197,25 @@ private void startContainer(Path packagePath,
       throw new SamzaContainerLaunchException("IO Exception when writing credentials to output buffer");
     }
 
+    Map<String, LocalResource> localResourceMap = new HashMap<>();
+    localResourceMap.put("__package", packageResource);
+
+    // include the resources from the universal resource configurations
+    try {
+      LocalizerResourceMapper resourceMapper = new LocalizerResourceMapper(config, yarnConfiguration);
```
vjagadish1989: Would n't it be nicer to standardize all localizer related configs in one-place? Currently, they appear to be in here, as well as in `LocalizerResourceMapper`. I wonder if we can consolidate all of them into a single function that takes in a config , and returns a map of local resources we can directly pass in to `SamzaContainerContext`.?

Something along the lines of :
```
public Map<String, LocalResource> getLocalResources(Config config) {

}
```
You can use the existing `LocalizerResourceMapper` class, or provide a `static` function (in one of the util classes)

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnContainerRunner.java
```diff
@@ -190,13 +197,25 @@ private void startContainer(Path packagePath,
...
+
     ContainerLaunchContext context = Records.newRecord(ContainerLaunchContext.class);
     context.setEnvironment(env);
     context.setTokens(allTokens.duplicate());
     context.setCommands(new ArrayList<String>() {{add(cmd);}});
-    context.setLocalResources(Collections.singletonMap("__package", packageResource));
+    context.setLocalResources(localResourceMap);
 
-    log.debug("setting package to {}", packageResource);
+    log.debug("setting localResourceMap to {}", localResourceMap);
```
vjagadish1989: thanks for carefully changing in all the logging statements 👍

samza-yarn/src/main/scala/org/apache/samza/job/yarn/YarnJobFactory.scala
```diff
@@ -37,6 +38,12 @@ class YarnJobFactory extends StreamJobFactory {
     if (config.containsKey(YarnConfiguration.RM_ADDRESS)) {
       hConfig.set(YarnConfiguration.RM_ADDRESS, config.get(YarnConfiguration.RM_ADDRESS, "0.0.0.0:8032"))
     }
+
+    if (config.containsKey("fs.certfs.impl.override")) { //TODO: change to use constants from CertFSConstants once that CertFSConstants.java is in
+      hConfig.set("fs.certfs.impl", config.get("fs.certfs.impl.override"))
+      logger.info("samza job config fs.certfs.impl.override is used for yarn.")
```
vjagadish1989: I'm probably missing some context here.

samza-yarn/src/main/java/org/apache/samza/config/YarnConfig.java
```diff
@@ -27,6 +27,13 @@
    */
   public static final String PACKAGE_PATH = "yarn.package.path";
 
+  /**
+   * The prefix for defining any optional yarn localizer resource
+   * The key will be localizer.resource.&lt;resourceVisibility&gt;.&lt;resourceType&gt;.&lt;resourceName&gt;
+   * The value corresponding to the key will be the URI.
+   */
+  public static final String LOCALIZER_RESOURCE_PREFIX = "yarn.localizer.resource.";
```
fredji97: Thanks for the suggestion. I am changing to this pattern of config.

samza-yarn/src/main/java/org/apache/samza/job/yarn/LocalizerResourceConfigParser.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+package org.apache.samza.job.yarn;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.apache.commons.lang.StringUtils;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.yarn.api.records.LocalResourceType;
+import org.apache.hadoop.yarn.api.records.LocalResourceVisibility;
+import org.apache.samza.clustermanager.*;
```
fredji97: done.

samza-yarn/src/main/java/org/apache/samza/job/yarn/LocalizerResourceMapper.java
fredji97: made map() private now and made localResourceMap immutable

samza-yarn/src/main/java/org/apache/samza/job/yarn/LocalizerResourceMapper.java
fredji97: I have put the map() function into the constructor, and now the caller just need to initialize the  object and then call getResourceMap(). I think it is simple enough. Although we can make it one static function, but that is not very testing friendly. 
Regarding the immutable map, Thanks for the suggestion, I changed it to ImmutableMap.

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnClusterResourceManager.java
fredji97: updated to use the same config name and made it general for any scheme.

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnContainerRunner.java
fredji97: removed

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnContainerRunner.java
fredji97: Currently, it calls the one from LocalizerResourceMapper and returns the localResourceMap for the context. 
There might be some legacy for package due to historical reason, and we plan to remove it later (TODO: SAMZA-1144 in the code), so I chose not to touch them now.

samza-yarn/src/main/scala/org/apache/samza/job/yarn/YarnJobFactory.scala
fredji97: changed to use fs.<scheme>.impl and made the same config name for both samza job config and yarn config.

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnContainerRunner.java
```diff
@@ -190,13 +196,24 @@ private void startContainer(Path packagePath,
       throw new SamzaContainerLaunchException("IO Exception when writing credentials to output buffer");
     }
 
+    Map<String, LocalResource> localResourceMap = new HashMap<>();
```
vjagadish1989: Not sure I understand completely. I was proposing to remove this duplication by refactoring the logic that populates the package resource inside `LocalizerResourceMapper`.  
- The logic for parsing configs is now spread across 2 places (one to construct the mapping for `yarn.package.path` and the other to construct this resources map). I'm wondering if the two can be in the same class.

samza-yarn/src/main/java/org/apache/samza/job/yarn/FsImplConfigManager.java
```diff
@@ -0,0 +1,114 @@
...
+
+
+/**
+ * FsImplConfigManager is intented to manage the Samza config for fs.&lt;scheme&gt;impl.
+ * e.g. fs.http.impl
+ *
+ * Usually this config exist in yarn config,
+ * but it is also allowed the clients define this for Samza Job and pick specific implementation for the job.
+ */
+public class FsImplConfigManager {
```
fredji97: I looked at several cases like this in our code base, like JavaStorageConfig, etc. But it used inherence instead of composition. From my perspective, I think it is better to use composition over inherence for this case, as it is easier to focus only the business logic for the specific config. If we use inheritance, we also expose the get() function which can return any config values although a lot of them are not related to this specific config type, for example, JavaStorageConfig can even return config not related to storage, which does not make too much sense to me. 

I propose to use composition for this case and do migration for other cases. But if there is a strong preferences for doing inherence for this case, I am OK to change it back to inherence as well.

samza-yarn/src/main/java/org/apache/samza/job/yarn/FsImplConfigManager.java
```diff
@@ -0,0 +1,114 @@
...
+      throw new LocalizerResourceException(fsImplKey + " does not have configured class implementation");
+    }
+    return fsImplClassName;
+  }
+
+  /**
+   * Take the fs.&lt;scheme&gt;impl from the samza job configuration and override the values in yarn configuration
+   * @param yarnConfiguration
+   */
+  public void overrideYarnConfiguration(YarnConfiguration yarnConfiguration) {
```
fredji97: This is to avoid the duplicate code in YarnJobFactory and YarnClusterResourceManager.

samza-yarn/src/main/java/org/apache/samza/job/yarn/LocalizerResourceException.java
```diff
@@ -0,0 +1,72 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.job.yarn;
+
+public class LocalizerResourceException extends Exception {
```
fredji97: Nice catch. Will change. Thanks.

samza-yarn/src/main/java/org/apache/samza/job/yarn/FsImplConfigManager.java
vjagadish1989: @fredji97 : 

0. I did not bring up the inheritance versus composition question in my comment. I'm totally okay with composition (in fact, I'd argue we should favor composition over inheritance). 

1. I 'm slightly opposed to the name - `Manager` in this class when it's just parsing configs and returning them. Usually `Manager`s in Samza do much more than that. For instance, the `ClusterResourceManager` handles life cycle management of all interactions with Yarn, `CheckpointManager` handles checkpoint persistence.

samza-yarn/src/main/java/org/apache/samza/job/yarn/FsImplConfigManager.java
vjagadish1989: If the responsibility of this class is parsing and returning a subset of configs, I'm not sure `overrideYarnConfiguration` belongs here. In fact, this should not have to do anything with Yarn.

samza-yarn/src/main/java/org/apache/samza/job/yarn/LocalizerResourceMapper.java
```diff
@@ -0,0 +1,103 @@
...
+  private ImmutableMap<String, LocalResource> localResourceMap;
+
+  public LocalizerResourceMapper(Config config, YarnConfiguration yarnConfiguration) throws LocalizerResourceException {
+    this.config = config;
+    this.yarnConfiguration = yarnConfiguration;
+    this.resourceConfigManager = new LocalizerResourceConfigManager(this.config);
+    map(); // set up localResourceMap
+  }
+
+  private void map() throws LocalizerResourceException {
```
fredji97: updated based on the comment.

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnContainerRunner.java
fredji97: the old yarn.package.path does not fit in the general LocalizerResourceConfig and LocalizerResourceMapper. As we would like to see if we can deprecate this one in favor of using yarn.resources.package.path which follows the general localizer resource config, I prefer not to hack  LocalizerResourceConfig and LocalizerResourceMapper to support this old config.

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnContainerRunner.java
```diff
@@ -144,13 +147,15 @@ public void runContainer(int samzaContainerId, Container container, CommandBuild
    *    Runs a command as a process on the container. All binaries needed by the physical process are packaged in the URL
    *    specified by packagePath.
    */
-  private void startContainer(Path packagePath,
+  private void startContainer(Path packagePath, //keep packagePath out of resourceNamePathMap for backward compatibility purpose
```
fredji97: Remove this comment for avoiding confusion.

samza-yarn/src/main/java/org/apache/samza/job/yarn/LocalizerResourceMapper.java
```diff
@@ -0,0 +1,103 @@
...
+      localResource.setVisibility(resourceVisibility);
+      return localResource;
+    } catch (IOException ioe) {
+      log.error("IO Exception when accessing the resource file status from the filesystem: " + resourcePath, ioe);
+      throw new LocalizerResourceException("IO Exception when accessing the resource file status from the filesystem: " + resourcePath);
+    }
+
+  }
+
+  public ImmutableMap<String, LocalResource> getResourceMap() {
```
fredji97: updated.

samza-yarn/src/main/java/org/apache/samza/job/yarn/LocalizerResourceMapper.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+public class LocalizerResourceMapper {
+  private static final Logger log = LoggerFactory.getLogger(LocalizerResourceMapper.class);
+
+  private Config config; //job configurations
+  private YarnConfiguration yarnConfiguration; //yarn configurations
+  private LocalizerResourceConfigManager resourceConfigManager;
+  private ImmutableMap<String, LocalResource> localResourceMap;
+
+  public LocalizerResourceMapper(Config config, YarnConfiguration yarnConfiguration) throws LocalizerResourceException {
```
fredji97: yeah, I changed to take LocalizerResourceConfig

samza-yarn/src/main/java/org/apache/samza/job/yarn/LocalizerResourceMapper.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.samza.config.Config;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * A universal approach to generate local resource map which can be put in ContainerLaunchContext directly
+ */
+public class LocalizerResourceMapper {
+  private static final Logger log = LoggerFactory.getLogger(LocalizerResourceMapper.class);
```
fredji97: updated

samza-yarn/src/main/java/org/apache/samza/job/yarn/FsImplConfigManager.java
fredji97: removed overrideYarnConfiguration.

samza-yarn/src/main/java/org/apache/samza/job/yarn/FsImplConfigManager.java
```diff
@@ -0,0 +1,114 @@
...
+    if (null == yarnConfiguration) {
+      log.error("yarnConfiguration is null, skipping overrideYarnConfiguration");
+      return;
+    }
+    try {
```
fredji97: removed overrideYarnConfiguration.

samza-yarn/src/main/java/org/apache/samza/job/yarn/FsImplConfigManager.java
```diff
@@ -0,0 +1,114 @@
...
+    } else {
+      this.config = config;
+    }
+  }
+
+  /**
+   * Get all schemes
+   * @return List of schemes in strings
+   */
+  public List<String> getSchemes() {
```
fredji97: after removing overrideYarnConfiguration, this should be public API now.

samza-yarn/src/main/java/org/apache/samza/job/yarn/FsImplConfigManager.java
```diff
@@ -0,0 +1,114 @@
...
+  private static final Logger log = LoggerFactory.getLogger(FsImplConfigManager.class);
+  private static final String FS_IMPL_PREFIX = "fs.";
+  private static final String FS_IMPL_SUFFIX = ".impl";
+  private static final String FS_IMPL = "fs.%s.impl";
+
+  private final Config config;
+
+  public FsImplConfigManager(final Config config) {
+    if (null == config) {
+      this.config = new MapConfig();
```
fredji97: changed to throw IllegalArgumentException instead.

samza-yarn/src/main/java/org/apache/samza/job/yarn/FsImplConfigManager.java
fredji97: Thanks Jagadish. I kept the same name convention, so it is called LocalizerResourceConfig, but I used composition over inheritance to only exposing the necessary APIs. I am going to open a discussion in open source to migrate all other similar configs to the composition pattern.

samza-yarn/src/test/java/org/apache/samza/job/yarn/TestYarnJobFactory.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.samza.config.MapConfig;
+import org.apache.samza.util.hadoop.HttpFileSystem;
+import org.junit.Test;
+import static org.junit.Assert.assertEquals;
+
+public class TestYarnJobFactory {
+
+  @Test
+  public void testGetJobWithDefaultFsImpl() {
+
```
vjagadish1989: nit: Get rid off unnecessary new lines.

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnContainerRunner.java
vjagadish1989: I'm fine if this is not done right away. We agree that once we deprecate `yarn.package.path`, the code path will be unified as per  my previous comment.

samza-yarn/src/main/java/org/apache/samza/job/yarn/FileSystemImplConfig.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.commons.lang.StringUtils;
+import org.apache.samza.config.Config;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * FsImplConfigManager is intented to manage the Samza config for fs.&lt;scheme&gt;impl.
```
vjagadish1989: nit: java docs seem to be still referring to `FsImplConfigManager`

samza-yarn/src/main/java/org/apache/samza/job/yarn/FileSystemImplConfig.java
```diff
@@ -0,0 +1,90 @@
...
+ * e.g. fs.http.impl
+ *
+ * Usually this config exist in yarn config,
```
vjagadish1989: nit: Java docs can be clearer and concise.

vjagadish1989 (approved):
looks good. approved.

feedback from final review:

Thanks for addressing all issues raised in previous drafts. 

Mostly nits (in this final pass). Also, some classes are missing docs. In some cases, docs could be more concise. Feel free to take a call, and re-word them as appropriate.

Fix It, and Ship It!

PR #51, opened 2017-02-09
SAMZA-1073: top-level fluent API

nickpan47:
`Initial draft of top-level fluent API for operator DAGs

PR #92, opened 2017-03-17
SAMZA-1094: Remove MessageEnvelope from public operator APIs. SAMZA-1101: Delay the creation of SinkFunction for output streams. SAMZA-1159: Move StreamSpec from a public API to an internal class.

prateekm:
Removed the MessageEnvelope and OutputStream interfaces from public operator APIs.
Moved the creation of SinkFunction for output streams to SinkOperatorSpec.
Moved StreamSpec from a public API to an internal class.

Additionally,
1. Removed references to StreamGraph in OperatorSpecs. It was being used to getNextOpId(). MessageStreamsImpl now gets the ID and gives it to OperatorSpecs itself.
2. Updated and cleaned up the StreamGraphBuilder examples.
3. Renamed SinkOperatorSpec to OutputOperatorSpec since its used by sink, sendTo and partitionBy.

@nickpan47 and @xinyuiscool, please take a look.


samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -144,122 +58,108 @@ public StreamGraphImpl(ApplicationRunner runner, Config config) {
...
-    }
+    IntermediateStreamImpl<K, V, M> intStream = (IntermediateStreamImpl<K, V, M>) inStreams.computeIfAbsent(
+        streamSpec.getId(), streamId -> new IntermediateStreamImpl<>(this, streamSpec,
+            keyExtractor, msgExtractor, msgBuilder, keySerde, msgSerde));
+    outStreams.putIfAbsent(streamSpec.getId(), intStream);
     return intStream;
   }
 
-  @Override public Map<StreamSpec, MessageStream> getInStreams() {
+  public Map<StreamSpec, MessageStream> getInStreams() {
```
xinyuiscool: shall we return Map<StreamSpec, InputStream>?

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -144,122 +58,108 @@ public StreamGraphImpl(ApplicationRunner runner, Config config) {
...
-  @Override public Map<StreamSpec, OutputStream> getOutStreams() {
-    Map<StreamSpec, OutputStream> outStreamMap = new HashMap<>();
-    this.outStreams.forEach((ss, entry) -> {
-        StreamSpec streamSpec = (entry instanceof IntermediateStreamImpl) ?
-          ((IntermediateStreamImpl) entry).getSpec() :
-          ((OutputStreamImpl) entry).getSpec();
-        outStreamMap.put(streamSpec, entry);
-      });
-    return Collections.unmodifiableMap(outStreamMap);
+  public Map<StreamSpec, MessageStream> getOutStreams() {
```
xinyuiscool: similar above, Map<StreamSpec, OutputStream>?

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -144,122 +58,108 @@ public StreamGraphImpl(ApplicationRunner runner, Config config) {
...
-   * @return  a {@link MessageStreamImpl} object corresponding to the {@code systemStream}
+   * @param sStream the {@link SystemStream}
+   * @return the {@link MessageStreamImpl} corresponding to the {@code systemStream}, or null
    */
-  public MessageStreamImpl getInputStream(SystemStream sstream) {
-    for (MessageStream entry: this.inStreams.values()) {
-      if (((InputStreamImpl) entry).getSpec().getSystemName().equals(sstream.getSystem()) &&
-          ((InputStreamImpl) entry).getSpec().getPhysicalName().equals(sstream.getStream())) {
-        return (MessageStreamImpl) entry;
+  public MessageStreamImpl getInputStream(SystemStream sStream) {
```
xinyuiscool: return InputStream/InputStreamImpl?

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
prateekm: Changed in the PR for SAMZA-1159.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
prateekm: Changed in the PR for SAMZA-1159.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
prateekm: Changed in the PR for SAMZA-1159.

samza-core/src/main/java/org/apache/samza/operators/spec/SinkOperatorSpec.java
```diff
@@ -44,45 +50,36 @@
...
    *
    * @param sinkFn  a user defined {@link SinkFunction} that will be called with the output message,
    *                the output {@link org.apache.samza.task.MessageCollector} and the
    *                {@link org.apache.samza.task.TaskCoordinator}.
-   * @param opCode  the specific {@link OpCode} for this {@link SinkOperatorSpec}. It could be {@link OpCode#SINK}, {@link OpCode#SEND_TO},
-   *                or {@link OpCode#PARTITION_BY}
-   * @param opId  the unique id of this {@link SinkOperatorSpec} in the {@link org.apache.samza.operators.StreamGraph}
+   * @param opCode  the specific {@link OpCode} for this {@link SinkOperatorSpec}.
+   *                It could be {@link OpCode#SINK}, {@link OpCode#SEND_TO}, or {@link OpCode#PARTITION_BY}
+   * @param opId  the unique ID of this {@link SinkOperatorSpec} in the {@link org.apache.samza.operators.StreamGraph}
```
nickpan47: Question: is SAMZA-1101 included in this patch or later? If included, the sink function to actually send to an output stream should not be directly coded and instantiated in SinkOperatorSpec(). The OutputOperatorSpec should not directly include SinkFunction() { collector.send(...); }, if we want to allow dynamic instantiation of different send functions to different Stream implementation in various different runtime environment. Is it something that you plan to do by creating different collector instance from the environment?

samza-core/src/main/java/org/apache/samza/operators/spec/SinkOperatorSpec.java
```diff
@@ -106,11 +103,30 @@ public int getOpId() {
...
+   * @param <M>  the type of input message
+   * @return  a {@link SinkFunction} that sends messages to the provided {@code outStream}
+   */
+  private static <M> SinkFunction<M> createSinkFn(MessageStream<M> outStream) {
+    if (!(outStream instanceof OutputStream)) {
+      throw new SamzaException("MessageStream used for sending messages must be an output stream.");
+    }
+
+    OutputStream<?, ?, M> outputStream = (OutputStream<?, ?, M>) outStream;
+    return (M message, MessageCollector mc, TaskCoordinator tc) -> {
```
nickpan47: Similar question here as in the above one.

samza-api/src/main/java/org/apache/samza/operators/StreamGraph.java
```diff
@@ -19,75 +19,56 @@
...
+   * @param msgSerde the serde used to deserialize the incoming message body
+   * @param <K> the type of key in the incoming message
+   * @param <V> the type of message in the incoming message
+   * @param <M> the type of message in the input {@link MessageStream}
    * @return the input {@link MessageStream}
    */
-  Map<StreamSpec, MessageStream> getInStreams();
+  <K, V, M> MessageStream<M> createInStream(StreamSpec streamSpec,
+      BiFunction<K, V, M> msgBuilder,
+      Serde<K> keySerde, Serde<V> msgSerde);
```
prateekm: No, this just predates that discussion. I'll remove them in this PR.

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -65,51 +65,55 @@
    *
    * @param graph the {@link StreamGraphImpl} object that this stream belongs to
    */
-  MessageStreamImpl(StreamGraphImpl graph) {
+  public MessageStreamImpl(StreamGraphImpl graph) {
```
prateekm: It's currently used in the "streams" sub-package that contains Input/Intermediate/OutputStreamImpl classes (didn't want to leave them all in the top-level package). I'll change it back if I end up switching to Input/OutputOperatorSpecs.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -144,122 +58,108 @@ public StreamGraphImpl(ApplicationRunner runner, Config config) {
...
    */
-  @Override
-  public <K, V, M extends MessageEnvelope<K, V>> OutputStream<M> createOutStream(StreamSpec streamSpec, Serde<K> keySerde, Serde<V> msgSerde) {
-    if (!this.outStreams.containsKey(streamSpec.getId())) {
-      this.outStreams.putIfAbsent(streamSpec.getId(), new OutputStreamImpl<K, V, M>(this, streamSpec, keySerde, msgSerde));
-    }
-    return this.outStreams.get(streamSpec.getId());
+  <K, V, M> MessageStreamImpl<M> createIntStream(String streamName, Function<M, K> keyExtractor, Function<M, V> msgExtractor,
+      BiFunction<K, V, M> msgBuilder) {
+    String streamId = String.format("%s-%s-%s",
```
prateekm: As an aside, when we move application runner out of SamzaContainer (SAMZA-1159 phase 2, depends on SAMZA-1118), ExecutionPlanner should be the one adding the job/app name and id to the operator name instead of MessageStreamImpl.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
prateekm: Yeah I was thinking about this too. This OutputOperatorSpec would be the same as the SinkOperatorSpec. The InputOperatorSpec is the (currently missing) spec for the RootOperatorImpl. Both are better names for the operators. Let me try this out and see if the ExecutionPlanner changes required for this make sense too.

samza-core/src/main/java/org/apache/samza/operators/spec/SinkOperatorSpec.java
prateekm: SAMZA-1101 is included in this patch. You're right, the idea is to create a different MessageCollector (or an even lower abstraction like the SystemConsumer/Producer) instead of creating a different SinkFunction for test vs prod environments. The user-provided or generated SinkFunction only describes the transforms for the messages and identity of the stream before sending it out via the collector. The environment abstracts how the collector functions (e.g. sending to a Kafka stream vs adding to an in-memory collection).

prateekm:
@nickpan47 @xinyuiscool

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpecs.java
```diff
@@ -100,77 +100,75 @@ public void init(Config config, TaskContext context) {
...
-   * Creates a {@link SinkOperatorSpec}.
+   * Creates an {@link OutputOperatorSpec} for the sink operator.
    *
-   * @param sinkFn  the sink function
+   * @param sinkFn  the sink function provided by the user
+   * @param opId  the unique ID of the operator
    * @param <M>  type of input message
-   * @param graph  the {@link StreamGraphImpl} object
-   * @return  the {@link SinkOperatorSpec}
+   * @return  the {@link OutputOperatorSpec} for the sink operator
```
nickpan47: It seems that you have renamed the SinkOperatorSpec to OutputOperatorSpec w/o changing the opCode correspondingly. Any particular reason? I thought that InputOperatorSpec/OutputOperatorSpec is not in the scope of this change?

Following the line of thought on sink() or sendTo() methods, here are the few things that we had talked about in another PR #51 
- if sink() is only send to non-stream terminals, maybe we should define SinkFunction as SinkFunction(M message, TaskCoordinator coordinator), removing the MessageCollector parameter there?
- For sendTo(), if we implemented SAMZA-1101, it would be translated into an OutputOperatorSpec, which only contains StreamSpec, keyExtractor, msgExtractor, instead of an actual SinkFunction?

If the answers to the above questions are not intended in this PR, I would prefer to move the changes to address SAMZA-1101 to a separate one

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -18,113 +18,114 @@
...
+    operatorImplGraph.init(streamGraph, config, initializedTaskContext);
+    this.operatorImplGraph = operatorImplGraph;
+
+    // TODO: SAMZA-1118 - Remove mapping after SystemConsumer starts returning logical streamId with incoming messages
+    inputSystemStreamToInputStream = new HashMap<>();
+    streamGraph.getInputStreams().forEach((streamSpec, inputStream)-> {
+        SystemStream systemStream = new SystemStream(streamSpec.getSystemName(), streamSpec.getPhysicalName());
+        inputSystemStreamToInputStream.put(systemStream, inputStream);
       });
-    operatorGraph.init(inputBySystemStream, config, this.contextManager.initTaskContext(config, context));
```
nickpan47: minor: The reason I only initialize the operatorGraph w/ the input streams from the task context is to avoid instantiating unnecessary processing logic. I think for now it does not make any difference. But it would change after we have multi-job applications implemented.

samza-core/src/test/java/org/apache/samza/example/BroadcastExample.java
```diff
@@ -0,0 +1,62 @@
...
+    inputStream.filter(m -> m.key.equals("key1")).sendTo("outputStream1", m -> m.key);
+    inputStream.filter(m -> m.key.equals("key2")).sendTo("outputStream2", m -> m.key);
+    inputStream.filter(m -> m.key.equals("key3")).sendTo("outputStream3", m -> m.key);
+  }
+
+  // local execution mode
+  public static void main(String[] args) throws Exception {
+    CommandLine cmdLine = new CommandLine();
+    Config config = cmdLine.loadConfig(cmdLine.parser().parse(args));
+    // for remote execution: ApplicationRunner runner = ApplicationRunner.getRemoteRunner(config);
```
nickpan47: Please remove this line. For remote runner, the entrance point is not through user program's main method here.

samza-core/src/test/java/org/apache/samza/example/KeyValueStoreExample.java
```diff
@@ -18,58 +18,43 @@
...
-
+        sendTo("pageViewEventPerMemberStream", statsOutput -> statsOutput.memberId);
   }
 
-  // local program model
+  // local execution mode
   public static void main(String[] args) throws Exception {
     CommandLine cmdLine = new CommandLine();
     Config config = cmdLine.loadConfig(cmdLine.parser().parse(args));
+    // for remote execution: ApplicationRunner runner = ApplicationRunner.getRemoteRunner(config);
```
nickpan47: Remove.

samza-core/src/test/java/org/apache/samza/example/NoContextStreamExample.java
```diff
@@ -42,87 +36,52 @@
...
-    return new JsonMessageEnvelope(
-        ((MessageType) ism.getMessage()).joinKey,
-        (MessageType) ism.getMessage(),
-        ism.getOffset(),
-        ism.getSystemStreamPartition());
+  // local execution mode
+  public static void main(String[] args) throws Exception {
+    CommandLine cmdLine = new CommandLine();
+    Config config = cmdLine.loadConfig(cmdLine.parser().parse(args));
+    // for remote execution: ApplicationRunner runner = ApplicationRunner.getRemoteRunner(config);
```
nickpan47: Same here.

samza-core/src/test/java/org/apache/samza/example/OrderShipmentJoinExample.java
```diff
@@ -38,83 +33,47 @@
...
+        .join(shipments, new MyJoinFunction(), Duration.ofMinutes(1))
+        .sendTo("joinedOrderShipmentStream", m -> m.orderId);
   }
 
-  // standalone local program model
+  // local execution mode
   public static void main(String[] args) throws Exception {
     CommandLine cmdLine = new CommandLine();
     Config config = cmdLine.loadConfig(cmdLine.parser().parse(args));
+    // for remote execution: ApplicationRunner runner = ApplicationRunner.getRemoteRunner(config);
```
nickpan47: Remove here as well.

samza-core/src/test/java/org/apache/samza/example/PageViewCounterExample.java
```diff
@@ -42,32 +39,28 @@
...
+            .setAccumulationMode(AccumulationMode.DISCARDING))
+        .map(PageViewCount::new)
+        .sendTo("pageViewEventPerMemberStream", m -> m.memberId);
   }
 
+  // local execution mode
   public static void main(String[] args) {
     CommandLine cmdLine = new CommandLine();
     Config config = cmdLine.loadConfig(cmdLine.parser().parse(args));
+    // for remote execution: ApplicationRunner runner = ApplicationRunner.getRemoteRunner(config);
```
nickpan47: remove.

samza-core/src/test/java/org/apache/samza/example/RepartitionExample.java
```diff
@@ -39,59 +36,28 @@
...
+        .map(MyStreamOutput::new)
+        .sendTo("pageViewEventPerMemberStream", m -> m.memberId);
   }
 
-  // standalone local program model
+  // local execution mode
   public static void main(String[] args) throws Exception {
     CommandLine cmdLine = new CommandLine();
     Config config = cmdLine.loadConfig(cmdLine.parser().parse(args));
+    // for remote execution: ApplicationRunner runner = ApplicationRunner.getRemoteRunner(config);
```
nickpan47: remove.

samza-core/src/test/java/org/apache/samza/example/WindowExample.java
```diff
@@ -0,0 +1,73 @@
...
+    // for 1 minute.
+    inputStream.window(Windows.tumblingWindow(Duration.ofMinutes(10), initialValue, counter)
+            .setLateTrigger(Triggers.any(Triggers.count(30000), Triggers.timeSinceLastMessage(Duration.ofMinutes(1)))));
+  }
+
+  // local execution mode
+  public static void main(String[] args) throws Exception {
+    CommandLine cmdLine = new CommandLine();
+    Config config = cmdLine.loadConfig(cmdLine.parser().parse(args));
+    // for remote execution: ApplicationRunner runner = ApplicationRunner.getRemoteRunner(config);
```
nickpan47: Won't comment on each line now. Please scrub through and remove this comment.

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpecs.java
prateekm: > It seems that you have renamed the SinkOperatorSpec to OutputOperatorSpec w/o changing the opCode correspondingly. Any particular reason? 

As earlier, the opCodes are passed by OperatorSpecs when it creates the OutputOperatorSpec. So an OutputOperatorSpec opCode can be a SEND_TO, SINK or PARTITION_BY. This change just renames the class to be more general (sink vs output).

> I thought that InputOperatorSpec/OutputOperatorSpec is not in the scope of this change?

Yup, not related to the Input/OutputOperatorSpec in the previous discussion. I couldn't find a consistent way to introduce both, this is as close as I could get.

> Following the line of thought on sink() or sendTo() methods, here are the few things that we had talked about in another PR #51 
> 1) if sink() is only send to non-stream terminals, maybe we should define SinkFunction as SinkFunction(M message, TaskCoordinator coordinator), removing the MessageCollector parameter there?
> 2) For sendTo(), if we implemented SAMZA-1101, it would be translated into an OutputOperatorSpec, which only contains StreamSpec, keyExtractor, msgExtractor, instead of an actual SinkFunction?

Agree with both. I think this PR still achieves the goal of SAMZA-1101 which was to not have the SinkFn for intermediate streams be created in the StreamGraphImpl, and move it down to the OperatorSpec/OperatorImpl layer instead. When removing the MessageCollector from the SinkFn, it's quite likely that we'll end up with 2 different specs - one for sink and one for sendTo and partitionBy. In that version, as you suggest, we'll probably end up not reusing SinkFn for sendTo and partitionBy. 

I think we're mostly on the same page in terms of what we want, I'm just leaving the sink/sendTo split for a later PR. The SinkOperatorSpec -> OutputOperatorSpec renaming is probably causing some confusion, so I can revert the rename and do it in the subsequent PR if that'll be clearer.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
prateekm: What do you have in mind? Would you prefer passing in only inputStreams directly? I like passing in StreamGraphImpl since it makes sense as the interface - pass in the operatorSpec DAG and get back the operatorImpl DAG.

samza-core/src/test/java/org/apache/samza/example/PageViewCounterExample.java
```diff
@@ -42,32 +40,29 @@
 public class PageViewCounterExample implements StreamApplication {
 
   @Override public void init(StreamGraph graph, Config config) {
+    MessageStream<PageViewEvent> pageViewEvents =
+        graph.getInputStream("pageViewEventStream", (k, m) -> (PageViewEvent) m);
+    OutputStream<String, PageViewCount, PageViewCount> pageViewEventPerMemberStream = graph
+        .getOutputStream("pageViewEventPerMemberStream", m -> m.memberId, m -> m);
```
vjagadish1989: minor: Not sure if there is like a single pattern we want to advocate. inlining versus having `OutputStream`s being pre-declared. If so, would be nice to use them consistently.

samza-core/src/test/java/org/apache/samza/example/PageViewCounterExample.java
```diff
@@ -42,32 +40,29 @@
...
-
-    pageViewEvents.
-        window(Windows.<PageViewEvent, String, Integer>keyedTumblingWindow(m -> m.getMessage().memberId, Duration.ofSeconds(10), initialValue, (m, c) -> c + 1).
-            setEarlyTrigger(Triggers.repeat(Triggers.count(5))).
-            setAccumulationMode(AccumulationMode.DISCARDING)).
-        map(MyStreamOutput::new).
-        sendTo(pageViewPerMemberCounters);
-
+    pageViewEvents
+        .window(Windows.<PageViewEvent, String, Integer>keyedTumblingWindow(m -> m.memberId, Duration.ofSeconds(10), initialValue, (m, c) -> c + 1)
```
vjagadish1989: Minor: Wondering if we still expect people to provide types in calls to `Windows.` methods? Or if the inference does n't work? The other example appears to work without it.

samza-core/src/test/java/org/apache/samza/example/WindowExample.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.samza.operators.windows.Windows;
+import org.apache.samza.runtime.ApplicationRunner;
+import org.apache.samza.util.CommandLine;
+
+import java.time.Duration;
+import java.util.function.Supplier;
+
+
+/**
+ * Example implementation of a simple user-defined tasks w/ window operators
```
vjagadish1989: Very minor: s/tasks/task

vjagadish1989 (approved):
approved. looks good to me. 

final round of f/b from my review: 
Mostly minor comments on docs. Feel free to defer some/ all of them.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -57,7 +57,7 @@ public ExecutionPlanner(Config config, StreamManager streamManager) {
     this.streamManager = streamManager;
   }
 
-  public JobGraph plan(StreamGraph streamGraph) throws Exception {
```
prateekm: Since StreamGraph is an user-facing API, IMHO it shouldn't expose getInputStreams/getOutputStreams (because users don't need them). Feels cleaner to move internal methods to the Impl class and use it directly. This is similar to how StreamOperatorTask uses MessageStreamImpl directly, how MessageStreamImpl uses StreamGraphImpl directly, etc. @

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -65,52 +67,59 @@
...
     return op.getNextStream();
   }
 
   @Override
   public <TM> MessageStream<TM> flatMap(FlatMapFunction<M, TM> flatMapFn) {
-    OperatorSpec<TM> op = OperatorSpecs.createStreamOperatorSpec(flatMapFn, this.graph, new MessageStreamImpl<>(this.graph));
+    OperatorSpec<TM> op = OperatorSpecs.createStreamOperatorSpec(
+        flatMapFn, new MessageStreamImpl<>(this.graph), this.graph.getNextOpId());
     this.registeredOperatorSpecs.add(op);
     return op.getNextStream();
```
prateekm: There's a (too?) subtle distinction b/w `nextStream` and `outputStream` in this PR - `nextStream` is used for the next MessageStream to propagate results to, while `outputStream` means an actual/physical output stream. If the usage is not consistent please LMK. For example, the `outputStream` to sendTo et. al is _not_ the nextStream for the SinkOperator since it's terminal.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -18,248 +18,120 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.BiFunction;
+import java.util.function.Function;
 
 /**
- * The implementation of {@link StreamGraph} interface. This class provides implementation of methods to allow users to
- * create system input/output/intermediate streams.
+ * The {@link StreamGraph} implementation that provides APIs for accessing {@link MessageStream}s to be used to
```
prateekm: Thanks.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -18,248 +18,120 @@
...
-        config.get(JobConfig.JOB_NAME()),
-        config.get(JobConfig.JOB_ID(), "1"),
-        opNameWithId);
-    StreamSpec streamSpec = runner.getStreamSpec(streamId);
-
-    this.inStreams.putIfAbsent(streamSpec.getId(), new IntermediateStreamImpl(this, streamSpec, null, null, parKeyFn));
-    IntermediateStreamImpl intStream = (IntermediateStreamImpl) this.inStreams.get(streamSpec.getId());
-    this.outStreams.putIfAbsent(streamSpec.getId(), intStream);
-    return intStream;
+  int getNextOpId() {
```
prateekm: Feels a little boiler-platey in general, but makes sense to call it out explicitly for this method. Thanks.

samza-core/src/main/java/org/apache/samza/operators/stream/OutputStreamInternal.java
```diff
@@ -17,38 +17,28 @@
...
-
-  /**
-   * Whether this {@link MessageEnvelope} indicates deletion of a previous message with this key.
-   *
-   * @return  true if the current {@link MessageEnvelope} indicates deletion of a previous message with this key
-   */
-  default boolean isDelete() {
-    return false;
-  }
+public interface OutputStreamInternal<K, V, M> extends MessageStream<M>, OutputStream<K, V, M> {
```
prateekm: Because `IntermediateStreamInternal implements OutputStreamInternal, InputStreamInternal`, but can't extend the two concrete classes directly.

samza-core/src/test/java/org/apache/samza/example/PageViewCounterExample.java
prateekm: Fixed, was meant to be pre-declaration everywhere. Do you prefer the inlined version?

samza-core/src/test/java/org/apache/samza/example/PageViewCounterExample.java
prateekm: Explicit annotation is required in this version, but is not if I extract the FoldLeftFunction as a variable. Changed to the latter. It makes sense, but is a weird user-experience. Anything we can do about it?

samza-core/src/test/java/org/apache/samza/example/RepartitionExample.java
```diff
@@ -39,59 +36,27 @@
...
-        sendTo(pageViewPerMemberCounters);
-
+    MessageStream<PageViewEvent> pageViewEvents =
+        graph.getInputStream("pageViewEventStream", (k, m) -> (PageViewEvent) m);
+
+    pageViewEvents
+        .partitionBy(m -> m.memberId)
+        .window(Windows.keyedTumblingWindow(m -> m.memberId, Duration.ofMinutes(5), initialValue, (m, c) -> c + 1))
+        .map(MyStreamOutput::new)
+        .sendTo(graph.getOutputStream("steramId", m -> m.memberId, m -> m.memberId));
```
prateekm: Changed to a more meaningful name. Accidentally committed the demo code I was showing you and Yi earlier :(

samza-core/src/test/java/org/apache/samza/operators/TestMessageStreamImpl.java
```diff
@@ -210,7 +212,7 @@ private void validateMergeOperator(MessageStream<TestMessageEnvelope> mergeSourc
     assertEquals(outputs.iterator().next(), mockMsg);
   }
 
-  @Test
```
prateekm: No idea what happened, thanks for catching!

samza-api/src/main/java/org/apache/samza/operators/StreamGraph.java
```diff
@@ -19,76 +19,52 @@
...
+  <K, V, M> MessageStream<M> getInputStream(String streamId, BiFunction<K, V, M> msgBuilder);
 
   /**
-   * Method to get the {@link OutputStream}s
+   * Gets the {@link OutputStream} corresponding to the logical {@code streamId}.
    *
-   * @return  the map of all {@link OutputStream}s
+   * @param streamId the unique logical ID for the stream
+   * @param keyExtractor the {@link Function} to extract the outgoing key from the output message
+   * @param msgExtractor the {@link Function} to extract the outgoing message from the output message
```
vjagadish1989: Sounds good, 👍 for a clean-up PR later.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
vjagadish1989: Makes sense, thanks!

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
vjagadish1989: As long as we update the docs for `getNextStream` to consistently reflect that, It's reasonable.

samza-core/src/test/java/org/apache/samza/example/PageViewCounterExample.java
vjagadish1989: Thanks. No strong preference on this one!  I suspect the inlined version is slightly easier for type inference. But, that's a very minor/trivial convenience.

samza-core/src/test/java/org/apache/samza/example/PageViewCounterExample.java
vjagadish1989: Yeah, there appear to be trade-offs either ways - pre-declaration versus providing types in-line  (depending on which route we go). For now, I think we can live with this.

PR #93, opened 2017-03-20
samza-1156: Improve ApplicationRunner method naming and class structure

jmakes:
@navina @xinyuiscool @nickpan47 take a look

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
+++ b/samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
@@ -27,16 +27,15 @@ import org.apache.samza.system.StreamSpec;
 
 
 /**
- * Interface to be implemented by physical execution engine to deploy the config and jobs to run the {@link org.apache.samza.operators.StreamGraph}
- *
- * Implementations of this interface must define a constructor with a single {@link Config} as the argument in order
- * to support the {@link ApplicationRunner#fromConfig(Config)} static constructor.
+ * A physical execution engine to deploy the config and jobs to run the {@link org.apache.samza.operators.StreamGraph}
  */
 @InterfaceStability.Unstable
-public interface ApplicationRunner {
+public abstract class ApplicationRunner {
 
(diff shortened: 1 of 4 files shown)


samza-core/src/test/java/org/apache/samza/runtime/TestAbstractApplicationRunner.java
```diff
@@ -52,85 +52,85 @@ public void testConfigValidation() {
 
   // The physical name should be pulled from the StreamConfig.PHYSICAL_NAME property value.
   @Test
-  public void testStreamFromConfigWithPhysicalNameInConfig() {
+  public void testgetStreamWithPhysicalNameInConfig() {
     Config config = buildStreamConfig(STREAM_ID,
                                       StreamConfig.PHYSICAL_NAME(), TEST_PHYSICAL_NAME,
                                       StreamConfig.SYSTEM(), TEST_SYSTEM);
 
     AbstractApplicationRunner env = new TestAbstractApplicationRunnerImpl(config);
```
nickpan47: nit: rename to runner?

samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
```diff
@@ -98,5 +104,5 @@ static ApplicationRunner fromConfig(Config config) {
    * @param streamId  The logical identifier for the stream in Samza.
    * @return          The {@link StreamSpec} instance.
    */
-  StreamSpec streamFromConfig(String streamId);
+  public abstract StreamSpec getStream(String streamId);
```
prateekm: @jmakes I think it'll be clearer to call this getStreamSpec to disambiguate b/w all the uses of Stream in fluent API (I'm thinking logical stream creation in StreamGraph and physical stream creation in ExecutionPlanner). What do you think?

jmakes: That totally makes sense. I'll put it on my list. We should really revisit all the names before the release, esp. with the architecture diagram so we can fully see the interactions/roles.

PR #94, opened 2017-03-20
SAMZA-1137: Instantiate ApplicationRunner in SamzaContainer

xinyuiscool:
Create an ApplicationRunner in SamzaContainer to provide StreamSpecs for fluent API.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -108,17 +108,6 @@ public class StreamProcessor {
     this(processorId, config, customMetricsReporters, (Object) streamTaskFactory);
   }
 
-  /**
-   * Same as {@link #StreamProcessor(int, Config, Map, AsyncStreamTaskFactory)}, except task instances are created
-   * using the "task.class" configuration instead of a task factory.
-   * @param processorId - this processor Id
-   * @param config - config
-   * @param customMetricsReporters metrics
-   */
-  public StreamProcessor(int processorId, Config config, Map<String, MetricsReporter> customMetricsReporters) {
-    this(processorId, config, customMetricsReporters, (Object) null);
-  }
-
   private StreamProcessor(int processorId, Config config, Map<String, MetricsReporter> customMetricsReporters,
                           Object taskFactory) {
     this.processorId = processorId;
--- a/samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
@@ -32,7 +32,7 @@ import org.apache.samza.util.Util;
  * For a StreamApplication, it creates the {@link ApplicationRunner} based on the config, and then run the application.
  * For a Samza job using low level task API, it will create the JobRunner to run it.
  */
-public class ApplicationRunnerMain {
+public class  ApplicationRunnerMain {
   // TODO: have the app configs consolidated in one place
   private static final String STREAM_APPLICATION_CLASS_CONFIG = "app.class";
 
--- a/samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
@@ -32,7 +32,7 @@ import org.apache.samza.util.Util;
  * For a StreamApplication, it creates the {@link ApplicationRunner} based on the config, and then run the application.
  * For a Samza job using low level task API, it will create the JobRunner to run it.
  */
-public class  ApplicationRunnerMain {
+public class ApplicationRunnerMain {
   // TODO: have the app configs consolidated in one place
   private static final String STREAM_APPLICATION_CLASS_CONFIG = "app.class";
 
--- a/samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
@@ -33,6 +33,7 @@ import org.apache.samza.job.model.ContainerModel;
 import org.apache.samza.job.model.JobModel;
 import org.apache.samza.metrics.JmxServer;
 import org.apache.samza.metrics.MetricsReporter;
+import org.apache.samza.task.TaskFactoryUtil;
 import org.apache.samza.util.ScalaToJavaUtils;
 import org.apache.samza.util.Util;
 import org.slf4j.Logger;
@@ -41,11 +42,11 @@ import org.slf4j.LoggerFactory;
 /**
  * LocalContainerRunner is the local runner for Yarn {@link SamzaContainer}s. It is an intermediate step to
  * have a local runner for yarn before we consolidate the Yarn container and coordination into a
- * StreamProcessor. This class will be replaced by the {@link org.apache.samza.processor.StreamProcessor} local runner
- * once that's done.
+ * {@link org.apache.samza.processor.StreamProcessor}. This class will be replaced by the {@link org.apache.samza.processor.StreamProcessor}
+ * local runner once that's done.
  *
- * Since we don't have the jobCoordinator implementation in Yarn, the components (jobModel and containerId) are directly
- * inside the runner.
+ * Since we don't have the {@link org.apache.samza.coordinator.JobCoordinator} implementation in Yarn, the components (jobModel and containerId)
+ * are directly inside the runner.
  */
 public class LocalContainerRunner extends AbstractApplicationRunner {
   private static final Logger log = LoggerFactory.getLogger(LocalContainerRunner.class);
@@ -64,6 +65,7 @@ public class LocalContainerRunner extends AbstractApplicationRunner {
     try {
       jmxServer = new JmxServer();
       ContainerModel containerModel = jobModel.getContainers().get(containerId);
+      Object taskFactory = TaskFactoryUtil.createTaskFactory(config, streamApp, this);
 
       SamzaContainer container = SamzaContainer$.MODULE$.apply(
           containerModel.getContainerId(),
@@ -73,8 +75,9 @@ public class LocalContainerRunner extends AbstractApplicationRunner {
           SamzaContainer.getLocalityManager(containerId, config),
           jmxServer,
           Util.javaMapAsScalaMap(new HashMap<String, MetricsReporter>()),
-          null,
+          taskFactory,
           this);
+
       container.run();
     } finally {
       if (jmxServer != null) {
@@ -83,7 +86,8 @@ public class LocalContainerRunner extends AbstractApplicationRunner {
     }
   }
 
-  public static void main(String[] args) {
+
+  public static void main(String[] args) throws Exception {
     setExceptionHandler(() -> System.exit(1));
 
     Integer containerId = Integer.valueOf(System.getenv(ShellCommandConfig.ENV_CONTAINER_ID()));
@@ -103,7 +107,8 @@ public class LocalContainerRunner extends AbstractApplicationRunner {
     MDC.put("jobName", jobName);
     MDC.put("jobId", jobId);
 
-    new LocalContainerRunner(jobModel, containerId).run(null);
+    StreamApplication streamApp = TaskFactoryUtil.createStreamApplication(config);
+    new LocalContainerRunner(jobModel, containerId).run(streamApp);
   }
 
   /* package private */ static void setExceptionHandler(Runnable runnable) {
--- a/samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
@@ -51,9 +51,9 @@ import org.slf4j.LoggerFactory;
 public class LocalContainerRunner extends AbstractApplicationRunner {
   private static final Logger log = LoggerFactory.getLogger(LocalContainerRunner.class);
   private final JobModel jobModel;
-  private final Integer containerId;
+  private final int containerId;
 
-  public LocalContainerRunner(JobModel jobModel, Integer containerId) {
+  public LocalContainerRunner(JobModel jobModel, int containerId) {
     super(jobModel.getConfig());
     this.jobModel = jobModel;
     this.containerId = containerId;
@@ -88,7 +88,10 @@ public class LocalContainerRunner extends AbstractApplicationRunner {
 
 
   public static void main(String[] args) throws Exception {
-    setExceptionHandler(() -> System.exit(1));
+    setExceptionHandler(() -> {
+      log.info("Exiting process now.");
+      System.exit(1);
+    });
 
     Integer containerId = Integer.valueOf(System.getenv(ShellCommandConfig.ENV_CONTAINER_ID()));
     log.info(String.format("Got container ID: %d", containerId));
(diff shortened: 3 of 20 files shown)


samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -106,6 +107,11 @@ object SamzaContainer extends Logging {
     putMDC("jobName", config.getName.getOrElse(throw new SamzaException("can not find the job name")))
     putMDC("jobId", config.getJobId.getOrElse("1"))
     var jmxServer: JmxServer = null
+    val appRunner = if (config.containsKey(StreamApplication.APP_CLASS_CONFIG)) {
+      ApplicationRunner.fromConfig(config)
+    } else {
+      null
```
jmakes: Will this work for standalone? Will they have the app class specified in the config?

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -0,0 +1,117 @@
...
+    if (jobConfig.getName().isEmpty()) {
+      throw new SamzaException("can not find the job name");
+    }
+    String jobName = jobConfig.getName().get();
+    String jobId = jobConfig.getJobId().getOrElse(ScalaToJavaUtils.defaultValue("1"));
+    MDC.put("containerName", "samza-container-" + containerId);
+    MDC.put("jobName", jobName);
+    MDC.put("jobId", jobId);
+
+    new LocalContainerRunner(jobModel, containerId).run(null);
```
nickpan47: It looks weird to have a null parameter here. Can we use runFromConfig(config)? Or simply run()?

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
```diff
@@ -521,6 +521,7 @@ public void testEndOfStreamOffsetManagement() throws Exception {
   }
 
   @Test
+  @Ignore
```
nickpan47: Why ignore here? I thought that the test was working?

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
```diff
@@ -147,29 +150,31 @@ private static void validateFactory(Object factory) {
...
+        return (StreamApplication) builderClass.newInstance();
       } catch (Throwable t) {
         log.error("Failed to validate StreamApplication class from the config. {}={}",
-            StreamApplication.APP_CLASS_CONFIG, config.get(StreamApplication.APP_CLASS_CONFIG));
-        return false;
+            StreamApplication.APP_CLASS_CONFIG, appClassName);
+        throw new ConfigException(String.format("%s is not a StreamApplication.", appClassName));
       }
+    } else {
+      return null;
```
prateekm: It's weird that this both throws an exception and returns null, should probably throw in the else case too.
Also, don't need to both log and throw the exception - we can throw the exception with an informative message here and make sure it gets caught and logged upstream.

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -0,0 +1,122 @@
...
+ * {@link org.apache.samza.processor.StreamProcessor}. This class will be replaced by the {@link org.apache.samza.processor.StreamProcessor}
+ * local runner once that's done.
+ *
+ * Since we don't have the {@link org.apache.samza.coordinator.JobCoordinator} implementation in Yarn, the components (jobModel and containerId)
+ * are directly inside the runner.
+ */
+public class LocalContainerRunner extends AbstractApplicationRunner {
+  private static final Logger log = LoggerFactory.getLogger(LocalContainerRunner.class);
+  private final JobModel jobModel;
+  private final Integer containerId;
```
prateekm: `int containerId`

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -0,0 +1,122 @@
...
+    MDC.put("jobName", jobName);
+    MDC.put("jobId", jobId);
+
+    StreamApplication streamApp = TaskFactoryUtil.createStreamApplication(config);
+    new LocalContainerRunner(jobModel, containerId).run(streamApp);
+  }
+
+  /* package private */ static void setExceptionHandler(Runnable runnable) {
+    Thread.UncaughtExceptionHandler exceptionHandler = (t, e) -> {
+      log.error(String.format("Uncaught exception in thread (name=%s). Exiting process now.", t.getName(), e));
```
prateekm: Remove "Exiting process now." from message here, should be in the runnable.

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -0,0 +1,122 @@
...
+      e.printStackTrace(System.err);
```
prateekm: Why both log error and print stacktrace? Also, I think e.printStackTrace will print to System.err by default.

Maybe use a `Callable` instead of Runnable that's called with the actual exception so that it can act on it?

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
```diff
@@ -39,28 +39,34 @@
...
+  private static StreamTaskFactory createStreamOperatorTaskFactory(StreamApplication streamApp, ApplicationRunner runner) {
+    return () -> new StreamOperatorTask(streamApp, runner);
+  }
 
+  /**
+   * Create {@link StreamTaskFactory} or {@link AsyncStreamTaskFactory} based on the configured task.class.
+   * @param config
+   * @return task factory instance
+   */
+  public static Object fromTaskClassConfig(Config config) {
```
prateekm: Wondering why this needs to be a public method. Is it not possible to only use createTaskFactory consistently?

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -0,0 +1,122 @@
...
+    MDC.put("containerName", "samza-container-" + containerId);
+    MDC.put("jobName", jobName);
+    MDC.put("jobId", jobId);
+
+    StreamApplication streamApp = TaskFactoryUtil.createStreamApplication(config);
+    new LocalContainerRunner(jobModel, containerId).run(streamApp);
+  }
+
+  /* package private */ static void setExceptionHandler(Runnable runnable) {
+    Thread.UncaughtExceptionHandler exceptionHandler = (t, e) -> {
```
prateekm: Minor: would be a good candidate for a Util method. I think something similar is also used for the BrokerProxy thread?

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
xinyuiscool: Currently it's only used in the main thread, so I will keep it here for now. We can refactor it out if there is another usage.

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
xinyuiscool: This part of the code is copied from SamzaContainer.main(). It seems pretty redundant to me too. Just for safe, I would keep it here :)

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
xinyuiscool: This is still used in SamzaContainer to keep the existing unit tests happy ...

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
xinyuiscool: It returns null when the application is not a StreamApplication (task.class is configured), and throws exception when the StreamAppliction is misconfigured. I updated the java doc for it, also fixed the logging issue.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -439,7 +394,7 @@ object SamzaContainer extends Logging {
       null
 
     val taskFactoryInstance = Option(taskFactory)
```
prateekm: IIUC, taskFactory should never be null at this point. If true, let's remove this variable.

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
```diff
@@ -39,28 +39,34 @@
...
+  }
 
-    String taskClassName;
+  private static StreamTaskFactory createStreamOperatorTaskFactory(StreamApplication streamApp, ApplicationRunner runner) {
+    return () -> new StreamOperatorTask(streamApp, runner);
+  }
 
+  /**
+   * Create {@link StreamTaskFactory} or {@link AsyncStreamTaskFactory} based on the configured task.class.
+   * @param config
```
prateekm: Empy description for @param. I'd prefer we remove that warning in checkstyle, but until then let's add some boilerplate.

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
prateekm: Since the goal of adding this class was to clean up task factory creation, I'd prefer to consolidate all the logic and have #createTaskFactory be the only public method. Maybe you can just pass it null for streamApp to get the same behavior in tests?

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
```diff
@@ -147,29 +150,31 @@ private static void validateFactory(Object factory) {
...
-  private static StreamTask createStreamOperatorTask(Config config, ApplicationRunner runner) throws Exception {
-    Class<?> builderClass = Class.forName(config.get(StreamApplication.APP_CLASS_CONFIG));
-    StreamApplication graphBuilder = (StreamApplication) builderClass.newInstance();
-    return new StreamOperatorTask(graphBuilder, runner);
-  }
-
-  private static boolean isStreamOperatorTask(Config config) {
+  /**
+   * Create a {@link StreamApplication} from config.
+   * Throw {@link ConfigException} if the configured class name is not a StreamApplication type, or
```
prateekm: Let's make this a @throw after the @param

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
xinyuiscool: I did some cleanup and now it can be finally removed.

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
xinyuiscool: After some code refactoring and cleanup, I am finally able to do this! Now the code is pretty clean :)

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
xinyuiscool: Seems checkstyle doesn't like it since the signature doesn't have a throws there (ConfigException is runtime exception). I add throws without @ below param which is pretty close.

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
```diff
@@ -32,7 +32,7 @@
  * For a StreamApplication, it creates the {@link ApplicationRunner} based on the config, and then run the application.
  * For a Samza job using low level task API, it will create the JobRunner to run it.
  */
-public class ApplicationRunnerMain {
+public class  ApplicationRunnerMain {
```
prateekm: Minor: Remove extra space.

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.samza.util.ScalaToJavaUtils;
+import org.apache.samza.util.Util;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * LocalContainerRunner is the local runner for Yarn {@link SamzaContainer}s. It is an intermediate step to
+ * have a local runner for yarn before we consolidate the Yarn container and coordination into a
+ * {@link org.apache.samza.processor.StreamProcessor}. This class will be replaced by the {@link org.apache.samza.processor.StreamProcessor}
+ * local runner once that's done.
```
prateekm: Minor: If we have a Samza JIRA ticket to track this would be good to add here.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -108,17 +108,6 @@ public StreamProcessor(int processorId, Config config, Map<String, MetricsReport
     this(processorId, config, customMetricsReporters, (Object) streamTaskFactory);
   }
 
-  /**
-   * Same as {@link #StreamProcessor(int, Config, Map, AsyncStreamTaskFactory)}, except task instances are created
-   * using the "task.class" configuration instead of a task factory.
-   * @param processorId - this processor Id
-   * @param config - config
-   * @param customMetricsReporters metrics
-   */
```
prateekm: Let's check with @navina if this is OK to remove.

navina: @xinyuiscool Why are we removing this? I am not against breaking changes to the StreamProcessor API because I will be introducing more breaking changes in my next PR. But it will be good to understand why we are removing this particular constructor. Thanks!

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
xinyuiscool: Fixed by creating a StreamApplication here.

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
xinyuiscool: It was failing about 1/5 for me so I commented it out.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
navina: I see. That sounds good. So, I am guessing the taskfactory creation will be removed from SamzaContainer as well??

navina: Nvm. I checked your code changes.. Looks like you are removing it. I like it. It makes the stack more component-friendly and clean. Thanks!

PR #95, opened 2017-03-23
SAMZA-1161: Adding metrics into LocalStoreMonitor.

shanthoosh:
Rocksdb LocalStoreMonitor is responsible for clearing up unused local task partition stores. Metrics are required to understand the behavior of this monitor in production(especially when it's clearing up unused rocksdb local stores).

The following two metrics will be emitted from this monitor:
a) Total disk space cleared in bytes.
b) Total number of rocksdb task partition stores cleared.

### DIFF
--- a/samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitor.java
+++ b/samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitor.java
@@ -28,7 +28,6 @@ import java.util.List;
 import org.apache.commons.io.filefilter.DirectoryFileFilter;
 import org.apache.commons.io.FileUtils;
 import org.apache.samza.container.TaskName;
-import org.apache.samza.metrics.MetricsRegistry;
 import org.apache.samza.rest.model.JobStatus;
 import org.apache.samza.rest.model.Task;
 import org.apache.samza.rest.proxy.job.JobInstance;
@@ -56,15 +55,16 @@ public class LocalStoreMonitor implements Monitor {
 
   private final LocalStoreMonitorConfig config;
 
-  // MetricsRegistry should be used in the future to send metrics from this monitor.
-  // Metrics from the monitor is a way to know if the monitor is alive.
+  private final LocalStoreMonitorMetrics localStoreMonitorMetrics;
+
   public LocalStoreMonitor(LocalStoreMonitorConfig config,
-                           MetricsRegistry metricsRegistry,
+                           LocalStoreMonitorMetrics localStoreMonitorMetrics,
                            JobsClient jobsClient) {
     Preconditions.checkState(!Strings.isNullOrEmpty(config.getLocalStoreBaseDir()),
                              String.format("%s is not set in config.", LocalStoreMonitorConfig.CONFIG_LOCAL_STORE_DIR));
     this.config = config;
     this.jobsClient = jobsClient;
+    this.localStoreMonitorMetrics = localStoreMonitorMetrics;
   }
 
   /**
@@ -81,8 +81,6 @@ public class LocalStoreMonitor implements Monitor {
     for (JobInstance jobInstance : getHostAffinityEnabledJobs(localStoreDir)) {
       File jobDir = new File(localStoreDir,
                              String.format("%s-%s", jobInstance.getJobName(), jobInstance.getJobId()));
-      Preconditions.checkState(jobDir.exists(), "JobDir is null");
-      Preconditions.checkNotNull(jobDir , "JobDir is null");
       JobStatus jobStatus = jobsClient.getJobStatus(jobInstance);
       for (Task task : jobsClient.getTasks(jobInstance)) {
         for (String storeName : jobDir.list(DirectoryFileFilter.DIRECTORY)) {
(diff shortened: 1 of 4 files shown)


shanthoosh:
@jmakes : Please take a look when you get a chance.

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitorConfig.java
```diff
@@ -46,7 +46,7 @@
    * Defines the comma separated list of job status servers of the form
    * "Host1:Port1,Host2:Port2".
    */
-  private static final String CONFIG_JOB_STATUS_SERVERS = "job.status.servers";
+  private static final String CONFIG_JOB_STATUS_SERVERS = "jobs.status.servers";
```
jmakes: This makes it inconsistent with the properties in JobsResourceConfig and therefore more confusing for users. Why is this change desirable?

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitor.java
```diff
@@ -147,6 +145,8 @@ private void markSweepTaskStore(File taskStoreDir) throws IOException {
     if (!offsetFile.exists()) {
       LOG.info("Deleting the task store : {}, since it has no offset file.", taskStorePath);
       FileUtils.deleteDirectory(taskStoreDir);
+      localStoreMonitorMetrics.diskSpaceFreedInBytes.inc(taskStoreDir.getTotalSpace());
```
jmakes: Does getTotalSpace() return an appropriate value AFTER the dir has already been deleted (line 147)?

How was this tested?

samza-rest/src/test/java/org/apache/samza/monitor/TestLocalStoreMonitor.java
```diff
@@ -65,7 +65,7 @@ public void setUp() throws Exception {
 
     // Initialize the local store monitor with mock and config
     localStoreMonitor = new LocalStoreMonitor(new LocalStoreMonitorConfig(new MapConfig(config)),
-                                              new NoOpMetricsRegistry(),
+                                              new LocalStoreMonitorMetrics("TestMonitorName", new NoOpMetricsRegistry()),
```
jmakes: Maybe NoOpMetricsRegistry is no longer appropriate and the new metrics should be tested.

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitor.java
shanthoosh: This was tested in my devbox(execution environment is on open-jdk). 

`Does getTotalSpace() return appropriate value?`

It depends upon the JDK/JVM implementation of File interface. In some implementations, the size of the File handle in bytes is read(along with other file attributes) when the contents are read/written(which is inline with the documentation for this interface). My sample test totally separate from this work fine with that.  

In some implementations, this total bytes value is fetched during each call of this API(no pre-fetch & storage during read/write call). 

 I'm moving this getTotalSpace() call before the delete operation.

samza-rest/src/test/java/org/apache/samza/monitor/TestLocalStoreMonitor.java
shanthoosh: Added more assertions based upon the metrics generated.

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitorConfig.java
shanthoosh: The recent for renaming is for consistency. 

All other config keys that are in this Config class have prefix jobs(and this has prefix job). 
* How this collides with JobsResourceConfig? 
* Why this is confusing? (Reason for extracting this out separately is to group all LocalStore Monitor config together. May be we could consider renaming all the config within this class to keep it different from JobsResourceConfig).

jmakes: Ahh didn't see those other properties in LocalStoreMonitorConfig, so from that perspective it makes sense. But... 

> How this collides with JobsResourceConfig?
> Why this is confusing?
 
Not collide, inconsistent. If a user configures samza-rest with both the jobs resource and this monitor they will have a mix of properties using "job." and "jobs." to refer to the same thing, right? 
`
job.proxy.factory.class=org.apache.samza.myClass
job.installations.path=./
jobs.status.servers=128.0.0.1
`
That will seem odd, and users will be unsure if there's some subtle difference they're missing.

shanthoosh: Yes, that wouldn't make less sense to the user (when using both the resources together). 

Is it reasonable to change the prefix of all the config keys in `LocalStoreMonitorConfig` from 'jobs' to 'job' (we'll have consistency across all the config keys in `samza-rest`)

jmakes: That's probably the best option. LocalStoreMonitor is newer than JobsResource, so it has had less exposure and will have less/no impact on users.

shanthoosh: Incorporated this change in the upcoming commit.

shanthoosh:
@jmakes 

The build was successful (https://builds.apache.org/job/samza-pr-checks/111/), however the github-jenkins plugin doesn't update the result here.

PR #96, opened 2017-03-26
SAMZA-1170: Integration testing harness for StreamApplications

vjagadish1989:
- Added an integration testing harness for testing `StreamApplication`s. This is convenient for running and interacting with Kafka brokers, Zk servers and Samza components locally.
- Added the following integration tests that use this harness to test actual `StreamApplication`s:
1. Keyed Tumbling Window
2. Keyed Session Window
3. Repartition with Keyed Session Window

Bonus: A couple of additional bug-fixes that were blockers for this test.


samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
```diff
@@ -54,7 +54,6 @@
   /**
    * Testing a basic identity stream task - reads data from a topic and writes it to another topic
    * (without any modifications)
```
jmakes: Personal preference: I like the whitespace for readability.

jmakes: Above comment is referring to the deleted line 57

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -0,0 +1,245 @@
...
+    ApplicationRunner runner = ApplicationRunner.fromConfig(new MapConfig(configs));
+    runner.run(app);
+  }
+
+  /**
+   * Shutdown and clear Zookeeper and Kafka broker state.
+   */
+  @Override
+  public void tearDown() {
+    super.tearDown();
```
prateekm: I think this also needs to runner.stop() (or equivalent) to stop the application. If that's not implemented yet,
 (check with @xinyuiscool or @jmakes for the schedule) maybe add a stub or comment?

samza-test/src/test/java/org/apache/samza/test/operator/TestRepartitionWindowApp.java
```diff
@@ -0,0 +1,69 @@
...
+    produceMessage(INPUT_TOPIC, 0, "userId1", "userId1,india,hotmail.com");
+    produceMessage(INPUT_TOPIC, 2, "userId1", "userId1,india,hotmail.com");
+    produceMessage(INPUT_TOPIC, 0, "userId1", "userId1,india,hotmail.com");
+
+    // run the application
+    RepartitionWindowApp app = new RepartitionWindowApp();
+    runApplication(app, APP_NAME, null);
+
+    // consume and validate result
+    List<ConsumerRecord<String, String>> messages = getMessages(Collections.singletonList(OUTPUT_TOPIC), 2);
```
prateekm: Do we know if the producer buffers have been flushed at this point? If not, is there a way to enforce that (maybe set Kafka Producer to flush every message)? Otherwise might result in flaky tests.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -0,0 +1,245 @@
...
+   * {@link #NUM_EMPTY_POLLS} polls return no messages.
+   *
+   * The default poll time out is determined by {@link #POLL_TIMEOUT_MS} and the number of empty polls are
+   * determined by {@link #NUM_EMPTY_POLLS}
+   *
+   * @param topics the list of topics to consume from
+   * @param threshold the number of messages to consume
+   * @return the list of {@link ConsumerRecord}s whose size can be atmost {@param threshold}
+   */
+  public List<ConsumerRecord<String, String>> getMessages(Collection<String> topics, int threshold) {
```
prateekm: Would it be cleaner for the users to get back a `Map<SS or SSP, List<ConsumerRecord<String, String>>`?

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
prateekm: +1. Maybe add a `<p>` too.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -55,7 +55,6 @@ public RemoteApplicationRunner(Config config) {
       // 2. create the physical execution plan
       ExecutionPlanner planner = new ExecutionPlanner(config);
       ProcessorGraph processorGraph = planner.plan(streamGraph);
-
```
prateekm: Prefer restoring newline.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
jmakes: The runner.kill() command is coming in samza-1089. ETA Wednesday.

samza-core/src/main/java/org/apache/samza/execution/ProcessorGraph.java
```diff
@@ -307,13 +307,13 @@ private void validateReachability() {
       while (!q.isEmpty()) {
         ProcessorNode node = q.poll();
         sortedNodes.add(node);
+        visited.add(node);
```
vjagadish1989: This handles a bug that this integration test harness found - In single node `ProcessorGraph`s, topological sort returns duplicate nodes. Please refer SAMZA-1172 for more context, and PR:#100 for the actual fix.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -303,6 +303,9 @@ private static void createStreams(ProcessorGraph graph, Map<String, SystemAdmin>
   }
 
   private static int maxPartition(Collection<StreamEdge> edges) {
+    if (edges.size() == 0) {
```
vjagadish1989: >>  Is this a bug or for a test case?

This was a bug that occurs for single node graphs. SAMZA-1172 (PR:#100) already fixed this. I'll just re-base with it.

samza-test/src/test/java/org/apache/samza/test/operator/RepartitionWindowApp.java
```diff
@@ -0,0 +1,61 @@
...
+  public void init(StreamGraph graph, Config config) {
+    StreamSpec pageViewStreamSpec = new StreamSpec("page-views", TestRepartitionWindowApp.INPUT_TOPIC, "kafka");
+    Function<MessageEnvelope<String, String>, String> keyFn = pageView -> (String) pageView.getKey();
+
+    MessageStream<MessageEnvelope<String, String>> pageViews = graph.createInStream(pageViewStreamSpec, null, null);
+    pageViews
+        .partitionBy(keyFn)
+        .window(Windows.keyedSessionWindow(keyFn, Duration.ofSeconds(3)))
+        // emit output
+        .sink((windowOutput, collector, coordinator) -> {
```
vjagadish1989: Thanks for clarifying the definition in SAMZA-1194 (PR:#92). I was under the impression that the two were interchangeable.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * Execution model: {@link StreamApplication}s are run as their own {@link org.apache.samza.job.local.ThreadJob}s.
+ * Similarly, embedded Kafka servers and Zookeeper servers are run as their own threads.
+ * {@link #produceMessage(String, int, String, String)} and {@link #getMessages(Collection, int)} are blocking calls.
+ *
+ * <h3>Usage Example</h3>
+ * Here is an actual test that publishes a message into Kafka, runs an application, and verifies consumption
+ * from the output topic.
+ *
+ *  <pre> {@code
```
vjagadish1989: Slightly prefer code examples in Java docs - as opposed to looking for a separate link, opening the file through an IDE and viewing actual code. Also, Makes documentation clearer when viewed from web-pages and documentation tools. 

Agreed, that keeping documentation consistent with refactors is a minor tax to pay.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -0,0 +1,245 @@
...
+  public void setUp() {
+    super.setUp();
+
+    Properties consumerDeserializerProperties = new Properties();
+    consumerDeserializerProperties.setProperty("key.deserializer", DEFAULT_DESERIALIZER);
+    consumerDeserializerProperties.setProperty("value.deserializer", DEFAULT_DESERIALIZER);
+
+    producer = TestUtils.createNewProducer(
+        bootstrapServers(),
+        1,
```
vjagadish1989: Neat, will fix.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -0,0 +1,245 @@
...
+   * @param key the key in the message
+   * @param val the value in the message
+   */
+  public void produceMessage(String topicName, int partitionId, String key, String val) {
+    producer.send(new ProducerRecord(topicName, partitionId, key, val));
+    producer.flush();
+  }
+
+  @Override
+  public int clusterSize() {
```
vjagadish1989: Should be `public` since it overrides a method from the `KafkaIntegrationTestHarness`. 

All methods of harness classes should be `protected` IMHO. Not sure what the scope for that is. Prefer to keep this `public` for now, and address it in a separate clean-up?

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -0,0 +1,245 @@
...
+    return Integer.parseInt(KafkaConfig.TOPIC_DEFAULT_REPLICATION_FACTOR());
+  }
+
+
+  /**
+   * Read messages from the provided list of topics until {@param threshold} messages have been read or until
+   * {@link #NUM_EMPTY_POLLS} polls return no messages.
```
vjagadish1989: Can we make it a package private field instead?

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
vjagadish1989: Will fix. Thanks.

vjagadish1989: The implementation is modeled after the `KafkaConsumer<K,V>` apis which return a `List<ConsumerRecord>` and let the higher level handle de-multiplexing them . 

Not sure if one pattern is strictly better than the other. In this usage, there's only one output topic, so it's a trivial convenience anyways.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
vjagadish1989: I do not think we have `runner.kill()` for the `TestApplicationRunner` yet. Will add it once the design of the `TestApplicationRunner` is finalized. The existing `RemoteApplicationRunner` is state-less, and cannot be used to start/stop jobs within the same JVM.

samza-test/src/test/java/org/apache/samza/test/operator/TestRepartitionWindowApp.java
vjagadish1989: We flush the `producer` buffers at every send. So, every `produceRequest` is synchronous. 

>> set Kafka Producer to flush every message?

We do that already. Here's the implementation of `produceMessage`. 
```
 producer.send(new ProducerRecord(topicName, partitionId, key, val));
 producer.flush();
```

Is this what you had in mind?

samza-test/src/test/java/org/apache/samza/test/operator/RepartitionWindowApp.java
```diff
@@ -0,0 +1,61 @@
...
+            String key = windowOutput.getKey().getKey();
+            String count = new Integer(windowOutput.getMessage().size()).toString();
+            LOG.info("Count is " + count  + " for key " + key);
+            collector.send(new OutgoingMessageEnvelope(new SystemStream("kafka", TestRepartitionWindowApp.OUTPUT_TOPIC), key, count));
```
vjagadish1989: Good point, Fixed.

samza-test/src/test/java/org/apache/samza/test/operator/RepartitionWindowApp.java
```diff
@@ -0,0 +1,61 @@
...
+
+    MessageStream<MessageEnvelope<String, String>> pageViews = graph.createInStream(pageViewStreamSpec, null, null);
+    pageViews
+        .partitionBy(keyFn)
+        .window(Windows.keyedSessionWindow(keyFn, Duration.ofSeconds(3)))
+        // emit output
+        .sink((windowOutput, collector, coordinator) -> {
+            String key = windowOutput.getKey().getKey();
+            String count = new Integer(windowOutput.getMessage().size()).toString();
+            LOG.info("Count is " + count  + " for key " + key);
```
vjagadish1989: This was for debugging, revised this in the updated draft.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+
+/**
+ * Harness for writing integration tests for {@link StreamApplication}s.
+ *
+ * <p> This provides the following features for its sub-classes:
```
vjagadish1989: There is nothing specific about state stores in this harness. This does not preclude usage / instantiation of state stores in any way. 

Do you still think it's a good idea to call it out in the doc? I can add it.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -0,0 +1,245 @@
...
+    consumer.subscribe(topics);
+
+    while (emptyPollCount < NUM_EMPTY_POLLS && recordList.size() < threshold) {
+      ConsumerRecords<String, String> records = consumer.poll(POLL_TIMEOUT_MS.toMillis());
+      if (!records.isEmpty()) {
+        Iterator<ConsumerRecord<String, String>> iterator = records.iterator();
+        while (iterator.hasNext() && recordList.size() < threshold) {
+          ConsumerRecord record = iterator.next();
+          recordList.add(record);
+        }
```
vjagadish1989: The current implementation exits the loop when there are 3 empty polls. Having 3 empty *successive* polls, is probably a stronger guarantee. Will fix.

vjagadish1989:
@jmakes @nickpan47 @prateekm Please can you guys take a pass at this?

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
```diff
@@ -169,7 +169,10 @@ public static StreamApplication createStreamApplication(Config config) {
         Class<?> builderClass = Class.forName(appClassName);
         return (StreamApplication) builderClass.newInstance();
       } catch (Throwable t) {
-        throw new ConfigException(String.format("%s is not a StreamApplication.", appClassName));
+        String errorMsg = String.format("Failed to create StreamApplication class from the config. %s = %s",
+            StreamApplication.APP_CLASS_CONFIG, config.get(StreamApplication.APP_CLASS_CONFIG));
+        log.error(errorMsg, t);
```
prateekm: Minor: Should just throw and let the caller log exceptions.

samza-test/src/test/java/org/apache/samza/test/operator/RepartitionWindowApp.java
```diff
@@ -0,0 +1,61 @@
...
+    Function<String, String> keyFn = pageView -> new PageView(pageView).getUserId();
+
+    OutputStream<String, String, WindowPane<String, Collection<String>>> outputStream = graph
+        .getOutputStream(TestRepartitionWindowApp.OUTPUT_TOPIC, m -> m.getKey().getKey(), m -> new Integer(m.getMessage().size()).toString());
+
+    pageViews
+        .map(m -> m)
+        .partitionBy(keyFn)
+        .window(Windows.keyedSessionWindow(keyFn, Duration.ofSeconds(3)))
+        // emit output
```
prateekm: Minor: remove here and in other classes?

samza-test/src/test/java/org/apache/samza/test/operator/RepartitionWindowApp.java
```diff
@@ -0,0 +1,61 @@
...
+ * A {@link StreamApplication} that demonstrates a repartition followed by a windowed count.
+ */
+public class RepartitionWindowApp implements StreamApplication {
+
+  private static final Logger LOG = LoggerFactory.getLogger(RepartitionWindowApp.class);
+
+  @Override
+  public void init(StreamGraph graph, Config config) {
+
+    BiFunction<String, String, String> msgBuilder = (k, v) -> v;
```
prateekm: Minor: Since these serve as samples, prefer in-lining these unless actually reused. I expect that it'll be the most common use case.

samza-test/src/test/java/org/apache/samza/test/operator/SessionWindowApp.java
```diff
@@ -0,0 +1,60 @@
...
+    BiFunction<String, String, PageView> msgBuilder = (k, v) -> new PageView(v);
+    MessageStream<PageView> pageViews = graph.getInputStream("page-views", msgBuilder);
+    OutputStream<String, String, WindowPane<String, Collection<PageView>>> outputStream = graph
+        .getOutputStream(TestSessionWindowApp.OUTPUT_TOPIC, m -> m.getKey().getKey(), m -> new Integer(m.getMessage().size()).toString());
+
+    Function<PageView, String> keyFn = pageView -> pageView.getUserId();
+
+    pageViews
+        .filter(m -> !FILTER_KEY.equals(m.getUserId()))
+        // emit output
```
prateekm: Minor: Remove.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
prateekm: The idea (at least with the RuntimeEnvironment) was that we'll allow mocking of state stores in the harness as well. Maybe not necessary here?

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -0,0 +1,253 @@
...
+ * Similarly, embedded Kafka servers and Zookeeper servers are run as their own threads.
+ * {@link #produceMessage(String, int, String, String)} and {@link #consumeMessages(Collection, int)} are blocking calls.
+ *
+ * <h3>Usage Example</h3>
+ * Here is an actual test that publishes a message into Kafka, runs an application, and verifies consumption
+ * from the output topic.
+ *
+ *  <pre> {@code
+ * class MyTest extends StreamApplicationIntegrationTestHarness {
+ *   private final StreamApplication myApp = new StreamApplication();
```
prateekm: Minor: new MyStreamApplication();

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -0,0 +1,253 @@
...
+        bootstrapServers(), // bootstrap-server url
+        1, // acks
+        60 * 1000L, // maxBlockMs
+        1024L * 1024L, // buffer size
+        0, // numRetries
+        0L, // lingerMs
+        5 * 1000L, // requestTimeout
+        SecurityProtocol.PLAINTEXT,
+        null,
+        Option$.MODULE$.<Properties>apply(new Properties()),
```
prateekm: Can probably use Option.apply directly, here and elsewhere.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -0,0 +1,253 @@
...
+    configs.put("job.default.system", "kafka");
+    configs.put("job.coordinator.replication.factor", "1");
+    configs.put("task.window.ms", "1000");
+
+    if (overriddenConfigs != null) {
+      configs.putAll(overriddenConfigs);
+    }
+
+    app = streamApplication;
+    runner = ApplicationRunner.fromConfig(new MapConfig(configs));
```
prateekm: Where are we specifying the runner to use? Is this using the RemoteApplicationRunner by default?

samza-test/src/test/java/org/apache/samza/test/operator/TestRepartitionWindowApp.java
```diff
@@ -0,0 +1,70 @@
...
+    produceMessage(INPUT_TOPIC, 2, "userId1", "userId1,india,1.com");
+    produceMessage(INPUT_TOPIC, 0, "userId1", "userId1,india,2.com");
+    produceMessage(INPUT_TOPIC, 1, "userId1", "userId1,india,3.com");
+
+    // run the application
+    RepartitionWindowApp app = new RepartitionWindowApp();
+    runApplication(app, APP_NAME, null);
+
+    // consume and validate result
+    List<ConsumerRecord<String, String>> messages = consumeMessages(Collections.singletonList(OUTPUT_TOPIC), 2);
```
prateekm: Minor: Would it be simpler to group by topic and return a map from consumeMessages instead?

samza-test/src/test/java/org/apache/samza/test/operator/TestSessionWindowApp.java
```diff
@@ -0,0 +1,81 @@
...
+  @Test
+  public void testSessionWindowCounter() throws Exception {
+    // create topics
+    createTopic(INPUT_TOPIC, 1);
+    createTopic(OUTPUT_TOPIC, 1);
+
+    long startTime = System.currentTimeMillis();
+    // Produce messages to the same partition. (intersperse them with messages with badKeys to be
+    // filtered out.)
+    produceMessage(INPUT_TOPIC, 0, "badKey", "badKey,india,google.com");
```
prateekm: Minor: Would it be useful to add `produceMessages(String topic, int partition, Collection<M> messages)` that produces them in order?

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
```diff
@@ -55,7 +55,7 @@
...
    *
-   * The standalone version in this test uses KafkaSystemFactory and it uses a SingleContainerGrouperFactory. Hence,
+   * <p> The standalone version in this test uses KafkaSystemFactory and it uses a SingleContainerGrouperFactory. Hence,
```
prateekm: Very Minor: <p> on previous line.

samza-test/src/test/java/org/apache/samza/test/operator/TumblingWindowApp.java
```diff
@@ -0,0 +1,60 @@
...
+  public void init(StreamGraph graph, Config config) {
+    BiFunction<String, String, PageView> msgBuilder = (k, v) -> new PageView(v);
+    MessageStream<PageView> pageViews = graph.getInputStream("page-views", msgBuilder);
+    Function<PageView, String> keyFn = pageView -> pageView.getUserId();
+    OutputStream<String, String, WindowPane<String, Collection<PageView>>> outputStream = graph
+        .getOutputStream(TestTumblingWindowApp.OUTPUT_TOPIC, m -> m.getKey().getKey(), m -> new Integer(m.getMessage().size()).toString());
+
+    pageViews
+        .filter(m -> !FILTER_KEY.equals(m.getUserId()))
+        // identity map
```
prateekm: Why? Also, remove comments.

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
vjagadish1989: Slightly err on the side of double logs instead of missed logs. Run time exceptions like these are not handled in any caller. Hence, would be nice to log this here anyways.

samza-test/src/test/java/org/apache/samza/test/operator/RepartitionWindowApp.java
vjagadish1989: The current sample:
```
    BiFunction<String, String, String> msgBuilder = (k, v) -> v;
    MessageStream<String> pageViews = graph.getInputStream("page-views", msgBuilder);

```
The new sample: 

```graph.<String, String, String>getInputStream("page-views", (k, v) -> v)```

No strong preferences on one versus the other.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
vjagadish1989: Reads better, since `StreamApplication` is an interface.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
vjagadish1989: Yeah, that's right. Once the test/local application runner is available, we'll use that.

samza-test/src/test/java/org/apache/samza/test/operator/TestRepartitionWindowApp.java
vjagadish1989: The implementation is modeled after the `KafkaConsumer<K,V>` apis which return a `List<ConsumerRecord>` and let callers handle de-multiplexing them (if need be) .

Not sure if one pattern is strictly better than the other. In this usage, there's only one output topic, so it's a trivial convenience anyways. I'll leave it as is, and change it if need be later.

samza-test/src/test/java/org/apache/samza/test/operator/TestSessionWindowApp.java
vjagadish1989: >> Would it be useful to add produceMessages(String topic, int partition, Collection<M> messages) that produces them in order?

In your example, Did you have in mind that `M` should wrap the key, and the message?
 
A more reasonable API IMHO could be something along the lines of:
`produceMessages(Collection<ProducerRecord> messages)`

That way, it's also more inline with `consumeMessages` which returns a list of `ConsumerRecord`s.

Most of the `TestRepartitionWindow` tests produce across multiple partitions. 
Will leave it as is, and add this as a future utility should a use-case arise.

samza-test/src/test/java/org/apache/samza/test/operator/TumblingWindowApp.java
vjagadish1989: This was to test wire-up for a `DAG` with multiple operators. The actual `map` function itself is probably less interesting.

PR #97, opened 2017-03-27
SAMZA-1143 Include fs.<scheme>.impl.* subkeys to YarnConfiguration used in YarnJobFactory and YarnClusterResourceManager

fredji97:
SAMZA-1143 Include fs.&lt;scheme&gt;.impl.* subkeys, in addition tofs.&lt;scheme&gt;.impl, to YarnConfiguration used in YarnJobFactory and YarnClusterResourceManager. 

When there are additional subconfigurations under fs.myScheme.impl, such as fs.myScheme.impl.client, we need to keep the set of configuration completed in  YarnJobFactory and YarnClusterResourceManager. When the context is set for localizing the resource in ClientHelper and YarnContainerRunner, it may rely on this configuration to get the FileStatus information, which may or may not depends on fs.&lt;scheme&gt;.impl and all possible fs.&lt;scheme&gt;.impl.* sub-configuration.  

This is an enhanced feature to the PR#90.

### DIFF
--- a/samza-yarn/src/main/java/org/apache/samza/job/yarn/FileSystemImplConfig.java
+++ b/samza-yarn/src/main/java/org/apache/samza/job/yarn/FileSystemImplConfig.java
@@ -72,4 +72,16 @@ public class FileSystemImplConfig {
     Config schemeConfig = config.subset(fsSchemeImpl, false); // do not strip off the prefix
     return schemeConfig;
   }
+
+  /**
+   * Get the set of subKeys for fs.&lt;scheme&gt;.impl if .&lt;scheme&gt;. has additional subKeys in the configuration
+   * e.g. fs.myScheme.impl.client and fs.myScheme.impl.server are the subKeys for myScheme
+   * @param scheme scheme name, such as http, hdfs, myscheme
+   * @return a set of subKeys from the configuration without stripping off prefix
+   */
+  public Set<String> getFsImplSubKeys(final String scheme) {
+    String fsImplSubKeyPrefix = String.format(FS_IMPL_SUBKEY_PREFIX, scheme);
+    Config subConfig = config.subset(fsImplSubKeyPrefix, false); // do not strip off the prefix
+    return subConfig.keySet();
+  }
 }
--- a/samza-yarn/src/main/java/org/apache/samza/job/yarn/FileSystemImplConfig.java
+++ b/samza-yarn/src/main/java/org/apache/samza/job/yarn/FileSystemImplConfig.java
@@ -20,6 +20,7 @@ package org.apache.samza.job.yarn;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Set;
 import org.apache.commons.lang.StringUtils;
 import org.apache.samza.config.Config;
 import org.slf4j.Logger;
@@ -35,6 +36,7 @@ public class FileSystemImplConfig {
   private static final String FS_IMPL_PREFIX = "fs.";
   private static final String FS_IMPL_SUFFIX = ".impl";
   private static final String FS_IMPL = "fs.%s.impl";
+  private static final String FS_IMPL_SUBKEY_PREFIX = "fs.%s.impl.";
 
   private final Config config;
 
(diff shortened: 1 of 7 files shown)


samza-yarn/src/main/java/org/apache/samza/job/yarn/FileSystemImplConfig.java
```diff
@@ -83,4 +85,16 @@ public String getFsImplClassName(final String scheme) {
...
     return fsImplClassName;
   }
+
+  /**
+   * Get the set of subKeys for fs.&lt;scheme&gt;.impl if .&lt;scheme&gt;. has additional subKeys in the configuration
+   * e.g. fs.myScheme.impl.client and fs.myScheme.impl.server are the subKeys for myScheme
+   * @param scheme scheme name, such as http, hdfs, myscheme
+   * @return a set of subKeys from the configuration without stripping off prefix
+   */
+  public Set<String> getFsImplSubKeys(final String scheme) {
```
vjagadish1989: Prefer an API like 

`` `
public Map<String, String> getSchemeConfig(final String scheme) {

} ``` 

1. It makes it more obvious from the API.
2. Usually, returning just `keys` is of limited value. Almost always, you will want to lookup the value configured for that key. Atleast in all interactions with `FsImplConfig` in the code base this is true.

fredji97:
Thanks @vjagadish1989 ! I have updated based on your comments.

samza-yarn/src/main/java/org/apache/samza/job/yarn/FileSystemImplConfig.java
```diff
@@ -61,26 +61,15 @@ public FileSystemImplConfig(final Config config) {
   }
 
   /**
-   * Get the fs.&lt;scheme&gt;impl as the config key from scheme
+   * Get the config subset for fs.&lt;scheme&gt;.impl
+   * It can include config for fs.&lt;scheme&gt;.impl and additional config for the subKeys fs.&lt;scheme&gt;.impl.* from the configuration
+   * e.g. for scheme "myScheme", there could be config for fs.myScheme.impl, fs.myScheme.impl.client and fs.myScheme.impl.server
    * @param scheme scheme name, such as http, hdfs, myscheme
-   * @return fs.&lt;scheme&gt;impl
+   * @return a set of sub configurations without stripping off prefix
```
vjagadish1989: nit:
`@return config for the particular scheme`

`set` of sub configurations is slightly mis-leading. Probably not worth calling out that the pre-fix is not stripped.

vjagadish1989 (approved):
approved.

samza-yarn/src/test/java/org/apache/samza/job/yarn/TestFileSystemImplConfig.java
```diff
@@ -45,12 +47,6 @@ public void testFileSystemImplConfigSuccess() {
     assertEquals(2, manager.getSchemes().size());
     assertEquals("http", manager.getSchemes().get(0));
     assertEquals("myscheme", manager.getSchemes().get(1));
-
-    assertEquals("fs.http.impl", manager.getFsImplKey("http"));
-    assertEquals("fs.myscheme.impl", manager.getFsImplKey("myscheme"));
-
-    assertEquals("org.apache.samza.HttpFileSystem", manager.getFsImplClassName("http"));
```
nickpan47: Why do we remove these two assertions?

samza-yarn/src/test/java/org/apache/samza/job/yarn/TestFileSystemImplConfig.java
```diff
@@ -61,15 +57,28 @@ public void testNullConfig() {
   }
 
   @Test
-  public void testEmptyImpl() {
```
nickpan47: Why are we removing this test for empty implementation?

samza-yarn/src/test/java/org/apache/samza/job/yarn/TestFileSystemImplConfig.java
fredji97: Because we removed these two functions, and introduced another function to get the key-value together as a sub-config.

samza-yarn/src/test/java/org/apache/samza/job/yarn/TestFileSystemImplConfig.java
fredji97: It is because we removed getFsImplClassName(), and this test was originally for testing against getFsImplClassName().

vjagadish1989:
@fredji97 Looks like there are some conflicts. Can we please resolve them?

fredji97:
@vjagadish1989 thanks I resolved the conflicts. Please let me know if you have any question.

vjagadish1989:
@fredji97 : Thanks for the patch, Merged, and submitted to master. Can you please close the associated JIRA with the fix version?

PR #100, opened 2017-03-28
SAMZA-1172: Fix for the topological sort to handle single-node loop

xinyuiscool:
In the processor graph, the topological sort missed adding to the visited set during graph traversal. This caused wrong graph being generated for single-node loop. This is fixed in the patch.

Also fixed the maxPartition method not handling empty collection correctly.

Added a few new unit tests for these. Also adjust the timing of previous async commit unit tests so it can run more reliably. Long term wise we need to fix the timer inside the AsyncRunLoop tests.

### DIFF
--- a/samza-core/src/test/java/org/apache/samza/execution/TestProcessorGraph.java
+++ b/samza-core/src/test/java/org/apache/samza/execution/TestProcessorGraph.java
@@ -34,6 +34,7 @@ public class TestProcessorGraph {
 
   ProcessorGraph graph1;
   ProcessorGraph graph2;
+  ProcessorGraph graph3;
   int streamSeq = 0;
 
   private StreamSpec genStream() {
@@ -88,6 +89,14 @@ public class TestProcessorGraph {
     graph2.addIntermediateStream(genStream(), "5", "5");
     graph2.addIntermediateStream(genStream(), "5", "7");
     graph2.addSink(genStream(), "7");
+
+    /**
+     * graph3 is a single node graph with a loop
+     * 1<->1
+     */
+    graph3 = new ProcessorGraph(null);
+    graph3.addSource(genStream(), "1");
+    graph3.addIntermediateStream(genStream(), "1", "1");
   }
 
   @Test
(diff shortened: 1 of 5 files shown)


xinyuiscool:
@vjagadish1989 : could you please take a look?

samza-core/src/test/java/org/apache/samza/execution/TestProcessorGraph.java
```diff
@@ -194,5 +203,9 @@ public void testTopologicalSort() {
     assertTrue(idxMap2.get("6") > idxMap2.get("1"));
     assertTrue(idxMap2.get("5") > idxMap2.get("4"));
     assertTrue(idxMap2.get("7") > idxMap2.get("5"));
+
+    //test graph3
+    List<ProcessorNode> sortedNodes3 = graph3.topologicalSort();
```
jmakes: This test proves the new conditional "if (pNodes.size() == 1)" works, but is there another test to validate that the visited node tracking is now correct?

xinyuiscool: I ran the unit test without this "if" condition and it worked as expected. The "if" was added later for optimization.

jmakes: It's a good optimization. But my point is that if someone decided those two "visited.add()" statements were unnecessary and deleted them. No unit test would catch that.

xinyuiscool: Good point. let me see whether I can come up with a unit test that will catch this.

xinyuiscool: I add another test graph with double loop (1<>1) -> (2<>2). This will test the visited.add() change directly. Previous code will not be able to generate sort results correctly.

vjagadish1989 (approved):
approved. looks good to me. thank you xinyu for the contribution!

PR #101, opened 2017-03-29
SAMZA-1167: New streamId-specific configs do not override equivalent system-scoped configs

jmakes:


jmakes:
@vjagadish1989 This should fix the integration test!

samza-core/src/main/java/org/apache/samza/config/DefaultChooserConfig.java
```diff
@@ -30,15 +30,17 @@
  * A convenience class for fetching configs related to the {@link org.apache.samza.system.chooser.DefaultChooser}
  */
 public class DefaultChooserConfig extends MapConfig {
-  public static final String BOOTSTRAP = StreamConfig.STREAM_PREFIX() + "samza.bootstrap";
-  public static final String PRIORITY = StreamConfig.STREAM_PREFIX() + "samza.priority";
-  public static final String BATCH_SIZE = "task.consumer.batch.size";
+  private static final String BOOTSTRAP = StreamConfig.SAMZA_PROPERTY() + "bootstrap";
```
xinyuiscool: Shall we define these directly inside StreamConfig, and then expose methods like getStreamBootstrapEnabled() and getStreamPriority() in StreamConfig? That way it might be easier to use elsewhere too.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
```diff
@@ -153,23 +170,67 @@ class StreamConfig(config: Config) extends ScalaMapConfig(config) with Logging {
...
     *
     * Note, because the translation is not perfect between SystemStream and streamId,
     * this method is not identical to getProperty(streamId, property)
+    *
+    * @param systemStream the SystemStream for which the property value will be retrieved.
+    * @param property the samza property name excluding the leading delimiter. e.g. "samza.x.y"
     */
-  private def getProperty(systemStream: SystemStream, property: String): String = {
-    val streamVal = getStreamProperties(systemStreamToStreamId(systemStream)).get(property)
+  def getSamzaProperty(systemStream: SystemStream, property: String): String = {
```
xinyuiscool: shall we make this private/package private so we will only expose the getters for specific configs?

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
```diff
@@ -153,23 +170,67 @@ class StreamConfig(config: Config) extends ScalaMapConfig(config) with Logging {
...
+    *
+    * Note, because the translation is not perfect between SystemStream and streamId,
+    * this method is not identical to getProperty(streamId, property)
+    *
+    * @param systemStream the SystemStream for which the property value will be retrieved.
+    * @param property the samza property name excluding the leading delimiter. e.g. "samza.x.y"
+    * @param defaultValue the default value to use if the property value is not found
+    *
+    */
+  def getSamzaProperty(systemStream: SystemStream, property: String, defaultValue: String): String = {
```
xinyuiscool: same above

samza-core/src/main/java/org/apache/samza/config/DefaultChooserConfig.java
jmakes: I thought about this and have no strong feelings about it. 

This would basically mean deleting this class, as it provides no other value.

Preference?

xinyuiscool: I prefer at least moving the reading config part to StreamConfig. This class does provides a few util methods there (the class name is pretty misleading), so either leaves them here or inline them directly in the default chooser(). Probably deleting the class is cleaner.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
jmakes: Done, Thanks!

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
jmakes: Done

samza-core/src/main/java/org/apache/samza/config/DefaultChooserConfig.java
```diff
@@ -55,7 +55,7 @@ public int getChooserBatchSize() {
     Set<SystemStream> bootstrapInputs = new HashSet<>();
     Set<SystemStream> allInputs = taskConfigJava.getAllInputStreams();
     for (SystemStream systemStream : allInputs) {
-      if (getBoolean(String.format(BOOTSTRAP, systemStream.getSystem(), systemStream.getStream()), false)) {
+      if (streamConfig.getBootstrapEnabled(systemStream)) {
```
vjagadish1989: neat :)

PR #103, opened 2017-03-30
SAMZA-1126 - Semantics of processorId in Samza

navina:
Implementation based on [SEP-1](https://cwiki.apache.org/confluence/display/SAMZA/SEP-1%3A+Semantics+of+ProcessorId+in+Samza)

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
+++ b/samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
@@ -18,6 +18,10 @@
  */
 package org.apache.samza.runtime;
 
+import org.apache.samza.annotation.InterfaceStability;
+import org.apache.samza.config.Config;
+
+@InterfaceStability.Evolving
 public interface ProcessorIdGenerator {
   /**
    * Generates a String representation to identify the processor instance
@@ -25,8 +29,9 @@ public interface ProcessorIdGenerator {
    *
    * <b>Note</b>: processorId has to be unique among the processors within a job
    *
+   * @param config Config instance
    * @return String Identifier for the processor
    */
-  String generateProcessorId();
+  String generateProcessorId(Config config);
 
 }
--- a/samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
+++ b/samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
@@ -33,5 +33,4 @@ public interface ProcessorIdGenerator {
    * @return String Identifier for the processor
    */
   String generateProcessorId(Config config);
-
 }
--- a/samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
+++ b/samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
@@ -33,9 +33,10 @@ public interface ProcessorIdGenerator {
    * <b>Specification of processor identifier</b>:
    * <ul>
    *  <li>Processor identifier has to be unique among the processors within a job</li>
-   *  <li>When more than one processor co-exist within the same JVM, the processor identifier should be of the format:
+   *  <li>When more than one processor co-exist within the same JVM, the processor identifier can be of the format:
    *  $x_$y, where 'x' is a unique identifier for the executing JVM and 'y' is a unique identifier for the
-   *  processor instance within the JVM.</li>
+   *  processor instance within the JVM. When there is only one processor within a JVM, 'x' should be sufficient to
+   *  uniquely identify the processor instance.</li>
    * </ul>
    *
    * <b>Note</b>:
--- a/samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
+++ b/samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
@@ -24,10 +24,11 @@ import org.apache.samza.config.Config;
 @InterfaceStability.Evolving
 public interface ProcessorIdGenerator {
   /**
-   * Generates a String representation to identify the processor instance
+   * Generates a String representation to identify a single instance of StreamProcessor
    * This value can be representative of its current executing environment. It can also be custom-managed by the user.
    *
-   * <b>Note</b>: processorId has to be unique among the processors within a job
+   * <b>Note</b>: processorId has to be unique among the processors within a job. Processors can co-exists within the
+   * same JVM, as long as the identifiers are guaranteed to be unique.
    *
    * @param config Config instance
    * @return String Identifier for the processor
--- a/samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
+++ b/samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
@@ -24,11 +24,24 @@ import org.apache.samza.config.Config;
 @InterfaceStability.Evolving
 public interface ProcessorIdGenerator {
   /**
-   * Generates a String representation to identify a single instance of StreamProcessor
-   * This value can be representative of its current executing environment. It can also be custom-managed by the user.
+   * Generates a String representation to identify a single instance of StreamProcessor.
    *
-   * <b>Note</b>: processorId has to be unique among the processors within a job. Processors can co-exists within the
-   * same JVM, as long as the identifiers are guaranteed to be unique.
+   * This value can be representative of its current executing environment. It can also be custom-managed by the user,
+   * as long as it adheres to the specification below. More than one processor can co-exist within the same JVM,
+   * as long as their identifiers are guaranteed to be unique.
+   *
+   * <b>Specification of processor identifier</b>:
+   * <ul>
+   *  <li>Processor identifier has to be unique among the processors within a job</li>
+   *  <li>When more than one processor co-exist within the same JVM, the processor identifier should be of the format:
+   *  $x_$y, where 'x' is a unique identifier for the executing JVM and 'y' is a unique identifier for the
+   *  processor instance within the JVM.</li>
+   * </ul>
+   *
+   * <b>Note</b>:
+   * In case of more than one processors within the same JVM, the custom implementation of ProcessorIdGenerator can
+   * contain a static counter, which is incremented on each call to generateProcessorId. The counter value can
+   * be treated as the identifier for the processor instance within the JVM.
    *
    * @param config Config instance
    * @return String Identifier for the processor
--- a/samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
+++ b/samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
@@ -16,20 +16,17 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.samza.coordinator;
+package org.apache.samza.runtime;
 
-import org.apache.samza.annotation.InterfaceStability;
-import org.apache.samza.config.Config;
-import org.apache.samza.processor.SamzaContainerController;
-
-
-@InterfaceStability.Evolving
-public interface JobCoordinatorFactory {
+public interface ProcessorIdGenerator {
   /**
-   * @param config Configs relevant for the JobCoordinator TODO: Separate JC related configs into a "JobCoordinatorConfig"
-   * @param containerController Controller interface for starting and stopping container. In future, it may simply
-   *                            pause the container and add/remove tasks
-   * @return An instance of IJobCoordinator
+   * Generates a String representation to identify the processor instance
+   * This value can be representative of its current executing environment. It can also be custom-managed by the user.
+   *
+   * <b>Note</b>: processorId has to be unique among the processors within a job
+   *
+   * @return String Identifier for the processor
    */
-  JobCoordinator getJobCoordinator(Config config, SamzaContainerController containerController);
+  String generateProcessorId();
+
 }
--- a/samza-core/src/main/java/org/apache/samza/config/ApplicationConfig.java
+++ b/samza-core/src/main/java/org/apache/samza/config/ApplicationConfig.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.config;
+
+/**
+ * Accessors for configs associated with Application scope
+ */
+public class ApplicationConfig extends MapConfig {
+  /**
+   * processor.id is equivalent to containerId in samza. It is a logical identifier used by Samza for a processor.
+   * In a distributed environment, this logical identifier is mapped to a physical identifier of the resource. For
+   * example, Yarn provides a "Yarn containerId" for every resource it allocates.
+   * In an embedded environment, this identifier is provided by the user by directly using the StreamProcessor API.
+   * <p>
+   * <b>Note:</b>This identifier has to be unique across the instances of StreamProcessors.
+   * </p>
+   * TODO: Deprecated in 0.13. After 0.13+, this id is generated using {@link org.apache.samza.runtime.ProcessorIdGenerator}
+   */
+  @Deprecated
+  public static final String PROCESSOR_ID = "processor.id";
+
+  /**
+   * Class implementing the {@link org.apache.samza.runtime.ProcessorIdGenerator} interface
+   * Used to generate a unique identifier for a {@link org.apache.samza.processor.StreamProcessor} based on the runtime
+   * environment. Hence, durability of the identifier is same as the guarantees provided by the runtime environment
+   */
+  public static final String APP_PROCESSOR_ID_GENERATOR_CLASS = "app.processor-id-generator.class";
+
+  public ApplicationConfig(Config config) {
+    super(config);
+  }
+
+  public String getAppProcessorIdGeneratorClass() {
+    return get(APP_PROCESSOR_ID_GENERATOR_CLASS, null);
+  }
+
+  @Deprecated
+  public String getProcessorId() {
+    return get(PROCESSOR_ID, null);
+  }
+
+}
--- a/samza-core/src/main/java/org/apache/samza/config/ApplicationConfig.java
+++ b/samza-core/src/main/java/org/apache/samza/config/ApplicationConfig.java
@@ -23,13 +23,15 @@ package org.apache.samza.config;
  */
 public class ApplicationConfig extends MapConfig {
   /**
-   * processor.id is equivalent to containerId in samza. It is a logical identifier used by Samza for a processor.
-   * In a distributed environment, this logical identifier is mapped to a physical identifier of the resource. For
-   * example, Yarn provides a "Yarn containerId" for every resource it allocates.
-   * In an embedded environment, this identifier is provided by the user by directly using the StreamProcessor API.
+   * <p>processor.id is similar to the logical containerId generated in Samza. However, in addition to identifying the JVM
+   * of the processor, it also contains a segment to identify the instance of the
+   * {@link org.apache.samza.processor.StreamProcessor} within the JVM. More detail can be found in
+   * {@link org.apache.samza.runtime.ProcessorIdGenerator}. </p>
    * <p>
-   * <b>Note:</b>This identifier has to be unique across the instances of StreamProcessors.
+   * This is an important distinction because Samza 0.13.0 in Yarn has a 1:1 mapping between the processor and the Yarn
+   * container (JVM). However, Samza in an embedded execution can contain more than one processor within the same JVM.
    * </p>
+   * <b>Note:</b>This identifier has to be unique across the instances of StreamProcessors.
    * TODO: Deprecated in 0.13. After 0.13+, this id is generated using {@link org.apache.samza.runtime.ProcessorIdGenerator}
    */
   @Deprecated
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
@@ -74,7 +74,7 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
     // Convert to a Set of ContainerModel
     Set<ContainerModel> containerModels = new HashSet<>();
     for (int i = 0; i < containerCount; i++) {
-      containerModels.add(new ContainerModel(i, taskGroups[i]));
+      containerModels.add(new ContainerModel(String.valueOf(i), taskGroups[i]));
     }
 
     return Collections.unmodifiableSet(containerModels);
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
@@ -44,6 +44,8 @@ import org.slf4j.LoggerFactory;
  * happens to be). No consideration is given towards locality, even distribution
  * of aggregate SSPs within a container, even distribution of the number of
  * taskNames between containers, etc.
+ *
+ * TODO: SAMZA-1197 - need to modify balance to work with processorId strings
  */
 public class GroupByContainerCount implements BalancingTaskNameGrouper {
   private static final Logger log = LoggerFactory.getLogger(GroupByContainerCount.class);
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
@@ -27,6 +27,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
+import org.apache.samza.SamzaException;
 import org.apache.samza.container.LocalityManager;
 import org.apache.samza.container.TaskName;
 import org.apache.samza.job.model.ContainerModel;
@@ -74,7 +76,7 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
     // Convert to a Set of ContainerModel
     Set<ContainerModel> containerModels = new HashSet<>();
     for (int i = 0; i < containerCount; i++) {
-      containerModels.add(new ContainerModel(String.valueOf(i), taskGroups[i]));
+      containerModels.add(new ContainerModel(String.valueOf(i), i, taskGroups[i]));
     }
 
     return Collections.unmodifiableSet(containerModels);
@@ -143,6 +145,13 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
    */
   private List<TaskGroup> getPreviousContainers(TaskAssignmentManager taskAssignmentManager, int taskCount) {
     Map<String, String> taskToContainerId = taskAssignmentManager.readTaskAssignment();
+    taskToContainerId.values().forEach(id -> {
+      try {
+        int intId = Integer.parseInt(id);
+      } catch (NumberFormatException nfe) {
+        throw new SamzaException("GroupByContainerCount cannot handle non-integer processorIds!", nfe);
+      }
+    });
     if (taskToContainerId.isEmpty()) {
       log.info("No task assignment map was saved.");
       return null;
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
@@ -142,7 +142,7 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
    *                              if the previous mapping doesn't exist or isn't usable.
    */
   private List<TaskGroup> getPreviousContainers(TaskAssignmentManager taskAssignmentManager, int taskCount) {
-    Map<String, Integer> taskToContainerId = taskAssignmentManager.readTaskAssignment();
+    Map<String, String> taskToContainerId = taskAssignmentManager.readTaskAssignment();
     if (taskToContainerId.isEmpty()) {
       log.info("No task assignment map was saved.");
       return null;
@@ -178,7 +178,7 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
   private void saveTaskAssignments(Set<ContainerModel> containers, TaskAssignmentManager taskAssignmentManager) {
     for (ContainerModel container : containers) {
       for (TaskName taskName : container.getTasks().keySet()) {
-        taskAssignmentManager.writeTaskContainerMapping(taskName.getTaskName(), container.getContainerId());
+        taskAssignmentManager.writeTaskContainerMapping(taskName.getTaskName(), container.getProcessorId());
       }
     }
   }
@@ -294,14 +294,14 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
    * @param taskToContainerId a map from each task name to the containerId to which it is assigned.
    * @return                  a list of TaskGroups ordered ascending by containerId.
    */
-  private List<TaskGroup> getOrderedContainers(Map<String, Integer> taskToContainerId) {
+  private List<TaskGroup> getOrderedContainers(Map<String, String> taskToContainerId) {
     log.debug("Got task to container map: {}", taskToContainerId);
 
     // Group tasks by container Id
-    HashMap<Integer, List<String>> containerIdToTaskNames = new HashMap<>();
-    for (Map.Entry<String, Integer> entry : taskToContainerId.entrySet()) {
+    HashMap<String, List<String>> containerIdToTaskNames = new HashMap<>();
+    for (Map.Entry<String, String> entry : taskToContainerId.entrySet()) {
       String taskName = entry.getKey();
-      Integer containerId = entry.getValue();
+      String containerId = entry.getValue();
       List<String> taskNames = containerIdToTaskNames.get(containerId);
       if (taskNames == null) {
         taskNames = new ArrayList<>();
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
@@ -146,12 +146,12 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
   private List<TaskGroup> getPreviousContainers(TaskAssignmentManager taskAssignmentManager, int taskCount) {
     Map<String, String> taskToContainerId = taskAssignmentManager.readTaskAssignment();
     taskToContainerId.values().forEach(id -> {
-      try {
-        int intId = Integer.parseInt(id);
-      } catch (NumberFormatException nfe) {
-        throw new SamzaException("GroupByContainerCount cannot handle non-integer processorIds!", nfe);
-      }
-    });
+        try {
+          int intId = Integer.parseInt(id);
+        } catch (NumberFormatException nfe) {
+          throw new SamzaException("GroupByContainerCount cannot handle non-integer processorIds!", nfe);
+        }
+      });
     if (taskToContainerId.isEmpty()) {
       log.info("No task assignment map was saved.");
       return null;
@@ -293,7 +293,8 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
         TaskModel model = taskNameToModel.get(taskName);
         containerTaskModels.put(model.getTaskName(), model);
       }
-      containerModels.add(new ContainerModel(container.containerId, containerTaskModels));
+      containerModels.add(
+          new ContainerModel(container.containerId, Integer.valueOf(container.containerId), containerTaskModels));
     }
     return Collections.unmodifiableSet(containerModels);
   }
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
@@ -314,8 +314,8 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
     // Build container tasks
     List<TaskGroup> containerTasks = new ArrayList<>(containerIdToTaskNames.size());
     for (int i = 0; i < containerIdToTaskNames.size(); i++) {
-      if (containerIdToTaskNames.get(i) == null) throw new IllegalStateException("Task mapping is missing container: " + i);
-      containerTasks.add(new TaskGroup(String.valueOf(i), containerIdToTaskNames.get(i)));
+      if (containerIdToTaskNames.get(String.valueOf(i)) == null) throw new IllegalStateException("Task mapping is missing container: " + i);
+      containerTasks.add(new TaskGroup(String.valueOf(i), containerIdToTaskNames.get(String.valueOf(i))));
     }
 
     return containerTasks;
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
@@ -234,7 +234,7 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
    * @param taskNamesToAssign     the list of tasks to assign to the containers.
    * @param containers            the containers (as {@link TaskGroup}) to which the tasks will be assigned.
    */
-  // TODO: Change logic from using int arrays to ordered array list
+  // TODO: Change logic from using int arrays to a Map<String, Integer> (id -> taskCount)
   private void assignTasksToContainers(int[] taskCountPerContainer, List<String> taskNamesToAssign,
       List<TaskGroup> containers) {
     for (TaskGroup taskGroup : containers) {
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
@@ -211,7 +211,7 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
   private List<TaskGroup> createContainers(int startContainerId, int endContainerId) {
     List<TaskGroup> containers = new ArrayList<>(endContainerId - startContainerId);
     for (int i = startContainerId; i < endContainerId; i++) {
-      TaskGroup taskGroup = new TaskGroup(i, new ArrayList<String>());
+      TaskGroup taskGroup = new TaskGroup(String.valueOf(i), new ArrayList<String>());
       containers.add(taskGroup);
     }
     return containers;
@@ -225,10 +225,11 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
    * @param taskNamesToAssign     the list of tasks to assign to the containers.
    * @param containers            the containers (as {@link TaskGroup}) to which the tasks will be assigned.
    */
+  // TODO: Change logic from using int arrays to ordered array list
   private void assignTasksToContainers(int[] taskCountPerContainer, List<String> taskNamesToAssign,
       List<TaskGroup> containers) {
     for (TaskGroup taskGroup : containers) {
-      for (int j = taskGroup.size(); j < taskCountPerContainer[taskGroup.getContainerId()]; j++) {
+      for (int j = taskGroup.size(); j < taskCountPerContainer[Integer.valueOf(taskGroup.getContainerId())]; j++) {
         String taskName = taskNamesToAssign.remove(0);
         taskGroup.addTaskName(taskName);
         log.info("Assigned task {} to container {}", taskName, taskGroup.getContainerId());
@@ -314,7 +315,7 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
     List<TaskGroup> containerTasks = new ArrayList<>(containerIdToTaskNames.size());
     for (int i = 0; i < containerIdToTaskNames.size(); i++) {
       if (containerIdToTaskNames.get(i) == null) throw new IllegalStateException("Task mapping is missing container: " + i);
-      containerTasks.add(new TaskGroup(i, containerIdToTaskNames.get(i)));
+      containerTasks.add(new TaskGroup(String.valueOf(i), containerIdToTaskNames.get(i)));
     }
 
     return containerTasks;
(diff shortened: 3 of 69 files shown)


navina:
Fyi, primary implementation changes are in StreamProcessor.java, ProcessorIdGenerator.java, SamzaObjectMapper.java, JsonContainerModelMixin.java and JsonJobModelMixin.java. 
Everything else is a side-effect of changing processorId from int to String. 

Due to the large-scope of this change, I would prefer to get explicit approvals from folks owning certain components. I have listed them below. Please review the main classes listed above and the components listed under your name. Appreciate the extra pair of eyes to make sure I don't break existing behavior. 

@vjagadish1989 All classes related to YarnClusterResourceManager - AppMasterService, ApplicationMaster{Rest,Web}Servlet, ContainerAllocator, ContainerRunner etc. 

@jmakes All class related to TaskAssignmentManager, BalancingTaskNameGrouper implementation and tests.

@nickpan47 All classes :) 

Thanks a lot everyone!

samza-core/src/test/scala/org/apache/samza/container/TestTaskInstance.scala
```diff
@@ -48,7 +48,7 @@ import org.apache.samza.system.SystemAdmin
 import scala.collection.mutable.ListBuffer
 
 class TestTaskInstance {
-  @Test
+ /* @Test
```
nickpan47: Same here. Please add comments and would prefer to move the changes to disable and fix the tests in a separate JIRA and PR.

samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.apache.samza.runtime;
+
+import org.apache.samza.annotation.InterfaceStability;
+import org.apache.samza.config.Config;
+
+@InterfaceStability.Evolving
+public interface ProcessorIdGenerator {
+  /**
+   * Generates a String representation to identify the processor instance
```
navina: ProcessorId is a unique identifier for each StreamProcessor instance. not the jvm. 
May be I should rephrase document as : 
"Generates a String representation to identify an instance of {@link StreamProcessor}"

I can further add a note about the case where a single JVM has multiple StreamProcessor instance. 
Do you think that will clarify it better?

samza-core/src/main/java/org/apache/samza/config/ApplicationConfig.java
```diff
@@ -0,0 +1,58 @@
...
+ * under the License.
+ */
+package org.apache.samza.config;
+
+/**
+ * Accessors for configs associated with Application scope
+ */
+public class ApplicationConfig extends MapConfig {
+  /**
+   * processor.id is equivalent to containerId in samza. It is a logical identifier used by Samza for a processor.
```
navina: I think we are supporting it. 
>  I don't think the processor.id is equivalent to containerId.
I hope you are also referring to containerId as the logical containerId used by Samza, which is different from the physical containerId provided by Yarn. Why do you think they are not equivalent? If we modify our Yarn-based implementation today to bring up more than one instance of SamzaContainer within the same JVM, it would be equivalent to the use-case we are offering now. Are you saying they are not equivalent because we have a 1:1 mapping between JVM and SamzaContainer today?

samza-core/src/main/java/org/apache/samza/coordinator/JobCoordinator.java
```diff
@@ -57,10 +57,11 @@
...
   /**
    * Returns the logical ID assigned to the processor
-   * It is up to the user to ensure that different instances of StreamProcessor within a job have unique processor ID.
-   * @return integer representing the logical processor ID
+   * It is the responsibility of the JobCoordinator to assign a unique identifier based on the underlying environment.
+   *
+   * @return String representing a unique logical processor ID
    */
-  int getProcessorId();
+  String getLocalProcessorId();
```
navina: > JobCoordinator's APIs are only called within local StreamProcessor, and there is no use case to call JobCoordinator to get processIds other than the StreamProcessor wrapping the JobCoordinator

I actually never even thought about the method for getting all processorIds through the JobCoordinator interface. That will be internal to the JC implementation and will not be used by any other interacting component, such as the StreamProcessor. Hence, I didn't find the documentation be confusing. 
I will try to clarify in the documentation that this interface container methods that are required for interacting with StreamProcessor.

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
```diff
@@ -60,16 +60,14 @@
...
    * @param metricsReporterMap  Map of metric reporter name and {@link MetricsReporter} instance
    */
   public SamzaContainerController(
       Object taskFactory,
       long containerShutdownMs,
-      String processorId,
       Map<String, MetricsReporter> metricsReporterMap) {
     this.executorService = Executors.newSingleThreadExecutor(new ThreadFactoryBuilder()
-        .setNameFormat("p" + processorId + "-container-thread-%d").build());
+        .setNameFormat("container-thread-%d").build());
```
navina: Mm.. Good point. I think I was simply re-factoring and didn't see the need for processorId except debug logging. But multiple processors within the same JVM will make it hard to differentiate since the threadpool is scoped within each StreamProcessor. 

Let me fix this. Thanks for pointing it out.

samza-core/src/main/java/org/apache/samza/serializers/model/JsonContainerModelMixIn.java
```diff
@@ -19,23 +19,28 @@
...
 
+import java.util.Map;
+
 /**
  * A mix-in Jackson class to convert Samza's ContainerModel to/from JSON.
  */
 public abstract class JsonContainerModelMixIn {
   @JsonCreator
-  public JsonContainerModelMixIn(@JsonProperty("container-id") int containerId, @JsonProperty("tasks") Map<TaskName, TaskModel> tasks) {
+  public JsonContainerModelMixIn(@JsonProperty("processor-id") String processorId, @JsonProperty("tasks") Map<TaskName, TaskModel> tasks) {
```
navina: When we read the old JSON container model, we won't be using the mixin anymore. I changed it to use the de-serializer. If the containerModel contains processor-id, all is well. If it  doesn't contain processor-id, it will return null. We, then, check to see if container-id is defined. If it is, it is converted to a processor-id String. Otherwise, we throw an exception as we expect the Container Model to either contain container-id or processor-id. 
I was able to verify the deserialization with a unit test in TestSamzaObjectMapper. 
Let me know if you are talking about something else here.

samza-core/src/main/java/org/apache/samza/serializers/model/JsonJobModelMixIn.java
```diff
@@ -30,12 +30,12 @@
  */
 public abstract class JsonJobModelMixIn {
   @JsonCreator
-  public JsonJobModelMixIn(@JsonProperty("config") Config config, @JsonProperty("containers") Map<Integer, ContainerModel> containers) {
+  public JsonJobModelMixIn(@JsonProperty("config") Config config, @JsonProperty("containers") Map<String, ContainerModel> containers) {
```
navina: I think jackson decoder handles integer/string conversions correctly. I remember trying out a simple test before making this change. 
Let me try to add a test so that it uses this Mixin. That way, we can all sleep peacefully :)

samza-core/src/main/java/org/apache/samza/serializers/model/SamzaObjectMapper.java
```diff
@@ -89,10 +91,30 @@ public static ObjectMapper getObjectMapper() {
     mapper.getSerializationConfig().addMixInAnnotations(TaskModel.class, JsonTaskModelMixIn.class);
     mapper.getDeserializationConfig().addMixInAnnotations(TaskModel.class, JsonTaskModelMixIn.class);
     mapper.getSerializationConfig().addMixInAnnotations(ContainerModel.class, JsonContainerModelMixIn.class);
-    mapper.getDeserializationConfig().addMixInAnnotations(ContainerModel.class, JsonContainerModelMixIn.class);
     mapper.getSerializationConfig().addMixInAnnotations(JobModel.class, JsonJobModelMixIn.class);
     mapper.getDeserializationConfig().addMixInAnnotations(JobModel.class, JsonJobModelMixIn.class);
 
+    module.addDeserializer(ContainerModel.class, new JsonDeserializer<ContainerModel>() {
+      @Override
+      public ContainerModel deserialize(JsonParser jp, DeserializationContext ctxt) throws IOException, JsonProcessingException {
```
navina: Rolling back shouldn't be a problem, right? Please note, that when we are serializing , we are writing both container-id and processor-id. Skipping one of them will make it hard to handle rollbacks. Does that address your concern?

samza-core/src/main/java/org/apache/samza/util/ClassLoaderHelper.java
```diff
@@ -19,11 +19,30 @@
...
+
 public class ClassLoaderHelper {
 
   public static <T> T fromClassName(String className) throws ClassNotFoundException, InstantiationException, IllegalAccessException {
     Class<T> clazz = (Class<T>) Class.forName(className);
     T instance = clazz.newInstance();
     return instance;
   }
+
+  public static <T> T fromClassName(String className, Class<T> classType) {
```
navina: I thought this was the new util class for Class Loading. I am aware that ApplicationRunner also has similar usage. I wasn't sure if I should change it as a part of this patch. Let me know if you think otherwise.

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
```diff
@@ -26,12 +26,10 @@
 
 
 public class ZkCoordinationServiceFactory implements CoordinationServiceFactory {
-
-
   synchronized public CoordinationUtils getCoordinationService(String groupId, String participantId, Config config) {
     ZkConfig zkConfig = new ZkConfig(config);
     ZkClient zkClient = new ZkClient(zkConfig.getZkConnect(), zkConfig.getZkSessionTimeoutMs(), zkConfig.getZkConnectionTimeoutMs());
-    ZkUtils zkUtils = new ZkUtils(participantId, new ZkKeyBuilder(groupId), zkClient, zkConfig.getZkConnectionTimeoutMs());
```
navina: Good question. When I created ZkUtils, I wanted it to be a util class for interacting with Zookeeper. Hence, I wanted it to be stateless. 

It has somehow evolved into a stateful class (with processorId, which is what is being named as participantId). I have urged @sborya to put up a SEP for discussing the design of the CoordinationService because I don't fully agree with how it has been implemented. Due to lack of documentation, I am unable to answer all your questions with clarity.

So to answer your questions:

> What's this participantId?

Not sure why Boris called it that. But it seems to be the same as processorID.
> Are we changing the znode path after removing the participantId from the parameter? 
No. This "participantId" was only used in log statements.  It doesn't affect the coordination implementation in anyway.

All I wanted to make sure was that processorId is not being treated as a state in ZkUtils. Iirc, processorId was being used in the log statements only. Hence, I felt it is ok to remove it.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -61,17 +65,24 @@
   private final CoordinationUtils coordinationUtils;
 
   private JobModel newJobModel;
-  private String newJobModelVersion;  // version published in ZK (by the leader)
```
navina: I refactored it because it didn't need to be a member variable. However, since I don't have a good idea about this code, I think I will change this back. 

> Is it relevant to the processorId change?
No. Not related at all.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -160,7 +158,7 @@ public void close() throws ZkInterruptedException {
     * @param dataListener describe this
     */
   public void subscribeToJobModelVersionChange(IZkDataListener dataListener) {
-    LOG.info("pid=" + processorId + " subscribing for jm version change at:" + keyBuilder.getJobModelVersionPath());
+    LOG.info(" subscribing for jm version change at:" + keyBuilder.getJobModelVersionPath());
```
navina: getJobModelVersion Path returns the path to the znode where the version is stored. The paths are managed by ZkKeyBuilder. 

newJobModelVersion field was not used anywhere in the ZkJobCoordinator. As I explained above, I only refactored as it was not actually used by ZkJobCoordinator.

Let's sync-up in-person if you have further questions.

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
```diff
@@ -163,31 +162,31 @@ public void testGroupManyTasks() {
    *  because the scenario is infrequent, the benefits are not guaranteed, and the code complexity
    *  wasn't worth it. It certainly could be implemented in the future.
    */
-  @Test
+//  @Test - Not working - TODO: Fix it
```
navina: Oh I fixed this test. Let me check and fix it. Thanks for pointing out.

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
```diff
@@ -244,7 +243,7 @@ public void testBalancerAfterContainerIncrease() {
    *  from C2 and C3 to containers that were on the same respective hosts, it wasn't implemented
    *  because the scenario is infrequent, the benefits are not guaranteed, and the code complexity
    *  wasn't worth it. It certainly could be implemented in the future.
-   */
+   *//*
```
navina: Ah.. these bunch of tests. I think I forgot to get to them after fixing the other million test failures. Let me work on this. Thanks for pointing out Yi! :)

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
```diff
@@ -833,5 +832,5 @@ public void testBalancerResultImmutable() {
 
     Set<ContainerModel> containers = new GroupByContainerCount(2).balance(taskModels, localityManager);
     containers.remove(containers.iterator().next());
-  }
+  }*/
```
navina: As stated above, this was not intentional. i will fix it.

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
```diff
@@ -519,7 +519,7 @@ public void testEndOfStreamOffsetManagement() throws Exception {
     callbackExecutor.awaitTermination(100, TimeUnit.MILLISECONDS);
   }
 
-  @Test
+//  @Test
```
navina: This is part of the flaky test. This is being fixed by @santhoshvenkat1988  in [SAMZA-1183](issues.apache.org/jira/browse/SAMZA-1183)

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
```diff
@@ -161,7 +155,7 @@ public void handleDataDeleted(String dataPath)
     Assert.assertTrue(testWithDelayBackOff(() -> "newProcessor".equals(res.getRes()), 2, 1000));
   }
 
-  @Test
+/*  @Test
```
navina: No. I just missed fixing this. Will take care of it now.

samza-core/src/test/scala/org/apache/samza/container/TestTaskInstance.scala
navina: Ah.. Yet another file I missed . Fixing it :)

samza-core/src/main/java/org/apache/samza/config/ApplicationConfig.java
navina: > we should move other app scope config variables to this class as well, just as app.class, app.runner.class, etc. And if some of those variables are used in API classes (e.g. ApplicationRunner.fromConfig() or RuntimeEnvironment.fromConfig()), we may need to move this to samza-api package as well

I spoke with Xinyu about this. He said he would follow-up with moving configs to this class because I didn't have full context on the ApplicationRunner implementation.

samza-core/src/main/java/org/apache/samza/serializers/model/JsonJobModelMixIn.java
navina: I verified the behavior of json parser in jackson library. Unless we configure json parser to "ALLOW_UNQUOTED_FIELD_NAMES", it will always serialize the field name as String and it will correctly deserialize it back to Integer/String as specified in the MixinAnnotation. 

For existing jobs, we can safely assume that containerId/processorId is always an integer value. So, it shouldn't be a problem to rollback or roll-forward.

navina:
@nickpan47 I have addressed all your comments. Please let me know if anything pending.

@jmakes  @vjagadish1989 : Can you guys take a look at those classes I have assigned to you guys?

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
```diff
@@ -74,7 +74,7 @@ public GroupByContainerCount(int containerCount) {
     // Convert to a Set of ContainerModel
     Set<ContainerModel> containerModels = new HashSet<>();
     for (int i = 0; i < containerCount; i++) {
-      containerModels.add(new ContainerModel(i, taskGroups[i]));
+      containerModels.add(new ContainerModel(String.valueOf(i), taskGroups[i]));
```
jmakes: This line, and I believe a few others, assume the container/processor ids are still a sequence of integers (even though the type is String). 

I see how this patch ensures that this code is only invoked in the legacy case, where that would be true. But this precondition/assumption is now subtle; it's not obvious by looking at the types. 

I suggest that we at least add a validation with a clear exception that ensures that the external strings (e.g. those that come from the TaskAssignmentManager) are all parseable integers.

jmakes: A better long term approach is to completely overhaul these groupers to have no assumptions about the strings and treat them as arbitrary tokens.

navina: > I suggest that we at least add a validation with a clear exception that ensures that the external strings (e.g. those that come from the TaskAssignmentManager) are all parseable integers.

What kind of validation would you suggest? In the Grouper itself? 

> A better long term approach is to completely overhaul these groupers to have no assumptions about the strings and treat them as arbitrary tokens

Yes. That was the intention with the standalone work. The task to replace with the new grouper has not yet been completed. I will add a TODO with a JIRA here so that we don't forget about it. For existing users running Yarn, this should continue to work as long as the AppMaster logic is not changed.

jmakes: What kind of validation would you suggest? In the Grouper itself?
Yes, if the class expects all container/processor ids to be parsable as integers, it should have an explicit validation of all external processor ids with a message explaining that the grouper will not work with arbitrary strings like UUIDs. Then if someone mistakenly changes the code, they get a clear exception rather than strange grouping behavior.

samza-core/src/main/java/org/apache/samza/coordinator/JobCoordinator.java
```diff
@@ -55,12 +57,14 @@
    * @throws InterruptedException if the current thread is interrupted while waiting for the JobCoordinator to start-up
    */
   boolean awaitStart(long timeoutMs) throws InterruptedException;
+
   /**
-   * Returns the logical ID assigned to the processor
-   * It is up to the user to ensure that different instances of StreamProcessor within a job have unique processor ID.
-   * @return integer representing the logical processor ID
+   * Returns the logical ID assigned to the processor that is local to the instance of StreamProcessor.
+   * It is the responsibility of the JobCoordinator to assign a unique identifier based on the underlying environment.
```
vjagadish1989: >> It is the responsibility of the JobCoordinator to assign a unique identifier based on the underlying environment

- Move it to the docs of the class instead of documenting it here? It we have already called out, this is probably redundant

- Will also be helpful to state the requirements of the ID here.

samza-core/src/main/java/org/apache/samza/job/model/ContainerModel.java
```diff
@@ -37,32 +37,48 @@
...
   private final int containerId;
+  private final String processorId;
   private final Map<TaskName, TaskModel> tasks;
 
-  public ContainerModel(int containerId, Map<TaskName, TaskModel> tasks) {
+  public ContainerModel(String containerId, Map<TaskName, TaskModel> tasks) {
+    this(containerId, Integer.parseInt(containerId), tasks);
+  }
+
+  public ContainerModel(String processorId, int containerId, Map<TaskName, TaskModel> tasks) {
```
vjagadish1989: Nit: Is it fair to say that this constructor will be deprecated?

samza-core/src/main/java/org/apache/samza/job/model/ContainerModel.java
```diff
@@ -87,6 +103,6 @@ public boolean equals(Object obj) {
   }
 
   public int compareTo(ContainerModel other) {
-    return containerId - other.getContainerId();
+    return processorId.compareTo(other.getProcessorId());
```
vjagadish1989: Wondering if we should check for null processorId here and handle comparisons accordingly? That way, we don't crash with an NPE. The `hashCode` does seem to handle null values.

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
```diff
@@ -44,7 +44,7 @@
 public class SamzaContainerController {
   private static final Logger log = LoggerFactory.getLogger(SamzaContainerController.class);
 
-  private final ExecutorService executorService;
```
vjagadish1989: Can a single `SamzaContainerController` manage lifecycle of multiple containers? Otherwise, we can take the processorId, and make the executorService immutable?

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
```diff
@@ -94,19 +90,21 @@ public SamzaContainerController(
   public void startContainer(ContainerModel containerModel, Config config, int maxChangelogStreamPartitions) {
```
vjagadish1989: Future clean-up: (pun intended ;) )

>> This method does not actually wait for the container to fully start-up
		
Would be nice to return a `Future` here. That way, the API looks more reasonable. Not sure what the scope of this will be. So, can defer it later.

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
```diff
@@ -60,16 +60,12 @@
...
-   * @param processorId         Id of the processor
    * @param metricsReporterMap  Map of metric reporter name and {@link MetricsReporter} instance
    */
   public SamzaContainerController(
       Object taskFactory,
       long containerShutdownMs,
-      String processorId,
       Map<String, MetricsReporter> metricsReporterMap) {
-    this.executorService = Executors.newSingleThreadExecutor(new ThreadFactoryBuilder()
-        .setNameFormat("p" + processorId + "-container-thread-%d").build());
```
vjagadish1989: - What is the cardinality of the relationship between `SamzaContainerController` and `SamzaContainer`? 

Some potential opportunities for clean-up:
- It seems that invoking `startContainer()` multiple times can potentially clobber the saved state in this class. I suspect we can add a guard to prevent this.
- Very minor: Are n't we creating multiple instances of `executorService` but only stopping the last created instance in `shutdown()` if `startContainer` is invoked multiple times? 

These are probably not blockers, so feel free to defer these for a future clean-up.

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
```diff
@@ -44,7 +44,7 @@
 public class SamzaContainerController {
   private static final Logger log = LoggerFactory.getLogger(SamzaContainerController.class);
 
-  private final ExecutorService executorService;
+  private ExecutorService executorService;
   private volatile SamzaContainer container;
```
vjagadish1989: Is this class thread-safe? If not, the `volatile` modifier is probably redundant here? If it is, there are other variables like `containerFuture` that are not thread-safe yet. Would be nice to fix one or both.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -83,51 +70,49 @@
...
+  private StreamProcessor(Config config, Map<String, MetricsReporter> customMetricsReporters,
                           Object taskFactory) {
-    this.processorId = processorId;
-
-    Map<String, String> updatedConfigMap = new HashMap<>();
-    updatedConfigMap.putAll(config);
-    updatedConfigMap.put(PROCESSOR_ID, String.valueOf(this.processorId));
-    Config updatedConfig = new MapConfig(updatedConfigMap);
+    // TODO: This check to be removed after 0.13+
+    ApplicationConfig applicationConfig = new ApplicationConfig(config);
```
vjagadish1989: Minor: Wondering if this check should be moved down to the `JobCoordinator` implementation instead of here?

samza-core/src/main/java/org/apache/samza/serializers/model/SamzaObjectMapper.java
```diff
@@ -89,10 +91,30 @@ public static ObjectMapper getObjectMapper() {
     mapper.getSerializationConfig().addMixInAnnotations(TaskModel.class, JsonTaskModelMixIn.class);
     mapper.getDeserializationConfig().addMixInAnnotations(TaskModel.class, JsonTaskModelMixIn.class);
     mapper.getSerializationConfig().addMixInAnnotations(ContainerModel.class, JsonContainerModelMixIn.class);
-    mapper.getDeserializationConfig().addMixInAnnotations(ContainerModel.class, JsonContainerModelMixIn.class);
     mapper.getSerializationConfig().addMixInAnnotations(JobModel.class, JsonJobModelMixIn.class);
     mapper.getDeserializationConfig().addMixInAnnotations(JobModel.class, JsonJobModelMixIn.class);
 
+    module.addDeserializer(ContainerModel.class, new JsonDeserializer<ContainerModel>() {
```
vjagadish1989: Minor: Slightly prefer being consistent with all other `DeSerializer`s. All others appear to define static inner classes as opposed to inlining them.

`  public static class ConfigDeserializer extends JsonDeserializer<Config> `

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
vjagadish1989: @nickpan47 @navina 
I remember I preferred (and still prefer) calling this `participantId` instead of `processorId` when I reviewed Boris's PR.  

It keeps the scope of this interface self-contained as a general purpose coordination API which is independent of `StreamProcessor` (or even `Samza` for that matter).

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -64,13 +64,11 @@
   private volatile String ephemeralPath = null;
   private final ZkKeyBuilder keyBuilder;
   private final int connectionTimeoutMs;
-  private final String processorId;
 
-  public ZkUtils(String processorId, ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs) {
```
vjagadish1989: Very minor: Slightly prefer `utils` classes as a collection of `static` methods instead of classes that contain state. Can rename this to `ZkClientWrapper` or something similar.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -190,7 +188,7 @@ public void publishJobModel(String jobModelVersion, JobModel jobModel) {
    * @return job model for this version
    */
   public JobModel getJobModel(String jobModelVersion) {
-    LOG.info("pid=" + processorId + "read the model ver=" + jobModelVersion + " from " + keyBuilder.getJobModelPath(jobModelVersion));
+    LOG.info("read the model ver=" + jobModelVersion + " from " + keyBuilder.getJobModelPath(jobModelVersion));
```
vjagadish1989: Is it usually helpful to log the `processorId` in state transitions / reads like these? Could potentially be useful if there are multiple `StreamProcessor`s running?

samza-core/src/main/java/org/apache/samza/config/ApplicationConfig.java
nickpan47: @navina yes, what I meant is exactly the current 1:1 mapping between a JVM process and SamzaContainer instance. Right now, the SamzaContainer.main() is invoked when starting the JVM process, which mandates a 1:1 mapping here. Also, the integer container_id is set in the current container process's environment and YarnClusterResourceManager.java is referring to that as a single integer to identify the physical JVM process locally.

samza-core/src/main/java/org/apache/samza/serializers/model/JsonContainerModelMixIn.java
nickpan47: 👍 thanks for the explanation. Makes sense.

samza-core/src/main/java/org/apache/samza/serializers/model/JsonJobModelMixIn.java
nickpan47: 👍 thanks for verifying! Super helpful.

samza-core/src/main/java/org/apache/samza/serializers/model/SamzaObjectMapper.java
nickpan47: Yep. That should be fine.

samza-core/src/main/java/org/apache/samza/util/ClassLoaderHelper.java
nickpan47: I agree that we can put it in another PR, as we see more use cases like this.

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
nickpan47: a) I would prefer to use processorId in the constructor parameter here, with more precise meaning of what's the participant in this particular context.
b) as long as the ZNode path is not affected by removing the input parameter here, it should be fine.

P.S. a side note to the ZkUtils class: usually, the list of entity names in the ZooKeeper methods are used to construct a znode path, which made me concerned about removing the participant parameter. Non-path related parameters usually occurs at the end.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
nickpan47: 👍 to remove if newJobModelVersion is not used by ZkJobCoordinator. Wondering how it ends up in the code base if not used?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
nickpan47: Prefer what Jagadish said as well.

samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.samza.config.Config;
+
+@InterfaceStability.Evolving
+public interface ProcessorIdGenerator {
+  /**
+   * Generates a String representation to identify a single instance of StreamProcessor
+   * This value can be representative of its current executing environment. It can also be custom-managed by the user.
+   *
+   * <b>Note</b>: processorId has to be unique among the processors within a job. Processors can co-exists within the
+   * same JVM, as long as the identifiers are guaranteed to be unique.
```
nickpan47: Question: is the part to differentiate multiple processors in the same physical JVM process defined? i.e. did we define that a processorId include:
1) a cluster-specific way to identify the physical JVM process in a job
2) a defined way to differentiate multiple processors in the same physical JVM process

I think that we should make it explicit in the javadoc for this class.

samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
```diff
@@ -0,0 +1,37 @@
...
+package org.apache.samza.runtime;
+
+import org.apache.samza.annotation.InterfaceStability;
+import org.apache.samza.config.Config;
+
+@InterfaceStability.Evolving
+public interface ProcessorIdGenerator {
+  /**
+   * Generates a String representation to identify a single instance of StreamProcessor
+   * This value can be representative of its current executing environment. It can also be custom-managed by the user.
```
navina: Will fix the typos

> Not sure if this is helpful: This value can be representative of its current executing environment. It can also be custom-managed by the user. I suspect it can be removed. You are already stating out the general contract for the ID. (in the following lines).

Why is not helpful though? If you think it is misleading, I can remove it. I will be updating this doc again based on some of @nickpan47 's feedback.

samza-core/src/main/java/org/apache/samza/coordinator/JobCoordinator.java
navina: > Move it to the docs of the class instead of documenting it here? It we have already called out, this is probably redundant

Sure. Makes sense.

> Will also be helpful to state the requirements of the ID here.

Shouldn't requirements of the ID already stated in ProcessorIdGenerator?

samza-core/src/main/java/org/apache/samza/job/model/ContainerModel.java
navina: Yeah.. I wasn't sure whether to annotate with deprecated because I introduced only now :) 
We should be deprecating this after 0.13, though.

samza-core/src/main/java/org/apache/samza/job/model/ContainerModel.java
navina: Hmm.. Yeah. Just read the compareTo documentation. We should be throwing NPE if either of them our null. Let me fix that and the documentation of the class, as recommended in Comparable interface

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
navina: We may not manage multiple containers (at least, not as of today). However, based on the current ZkJobCoordinator implementation, we should be able to start and stop the container multiple times within the same lifecycle of the StreamProcessor. That said, this class can use some refactoring. I would like keep this outside the scope of this change. 
Ideally, I shouldn't have fixed this here either. So, I am going to revert back any unrelated changes in this class and fix them in the StreamProcessor/JobCoordinator refactoring.

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
navina: Good question. I would like to keep changes not associated with SEP-1 out of this PR.

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
navina: As mentioned before, we have a separate task for refactoring.  Let's keep the clean-up in a separate patch. Thanks for the suggestions, though. Keep them coming :)

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
navina: Same answer as before. Let's focus on SEP-1 related changes alone here.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
navina: Why do you think so? Since this check is only for existing use-cases of StreamProcessors, it seems acceptable to keep it here and removing it after 0.13+. If we move this to JobCoordinator abstraction layer, are you suggesting we move it to each and every implementation? 
If we decide to allow this check only for StandaloneJobCoordinator (because ZkJobCoordinator didn't exist before and hence, there is no question about `processor.id`), then the behavior is different across jobcoordinator implementation. We allow `processor.id` in Standalone versions in 0.13 and not in Zk versions in 0.13. That seems weird too. That's the reason why I have kept it at StreamProcessor level. 

This may be pushed to ApplicationRunner level though. However, ApplicationRunner is not yet fully fleshed out when I made these changes. We can move this check to other abstraction layers later on, if the need calls for it.

samza-core/src/main/java/org/apache/samza/serializers/model/SamzaObjectMapper.java
navina: Sure

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
navina: I prefer addressing non SEP-1 related suggestions in a separate PR. This is an awfully large change and should not be bloated with other important changes.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
navina: Yeah. I am going to change this. I think I already responded to @nickpan47 's comments before.

samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
navina: @nickpan47 As I showed you in-person before, I added this to the design doc. I will include the specific details in the javadoc as well. Please wait for me to push the changes before reviewing it again. Thanks!

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
navina: @jmakes  This grouper essentially relies on job.container.count. It "generates" containerId according to the count and creates ContainerModels. There is no question of not being able to parse UUIDs. So, I am slightly lost of where the "external" processor ids are even passed in and where to add the validation. Let's discuss this in person and sort it out.

samza-core/src/main/java/org/apache/samza/config/ApplicationConfig.java
navina: Yep. Got it. I will rephrase the documentation based on our discussion from morning. Thanks for clarifying! 👍

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
navina: > a) I would prefer to use processorId in the constructor parameter here, with more precise meaning of what's the participant in this particular context.

I would defer to @sborya  on whether to call it processorId or participantId. You can add this question here - [SAMZA-1173](https://issues.apache.org/jira/browse/SAMZA-1173)

> as long as the ZNode path is not affected by removing the input parameter here, it should be fine.

It shouldn't be affected.

> Non-path related parameters usually occurs at the end.
That's a good suggestion. Seems like a good pattern to follow :)

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
navina: > Wondering how it ends up in the code base if not used?
It was probably missed in the review of the jobcoordinator PR?? Not sure Yi.

samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
navina: > s/StreamProcessor/{@link StreamProcessor}

ProcessorIdGenerator is in samza-api and StreamProcessor is in samza-core. I won't be able to make this change unless we move StreamProcessor to samza-api.

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
jmakes: It "generates" some containerIds according to the count. It receives others from external sources like the task assignment manager. I was only suggesting that the latter should be validated s.t. we throw an exception rather than letting unexpected/undefined behavior occur.

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
```diff
@@ -225,10 +225,11 @@ private void validateTasks(Set<TaskModel> tasks) {
    * @param taskNamesToAssign     the list of tasks to assign to the containers.
    * @param containers            the containers (as {@link TaskGroup}) to which the tasks will be assigned.
    */
+  // TODO: Change logic from using int arrays to ordered array list
```
navina: Oh.. That is a really old comment when I had not fully read through the implementation of the class. Let me fix the comment. Thanks!

navina:
> It "generates" some containerIds according to the count. It receives others from external sources like the task assignment manager. I was only suggesting that the latter should be validated s.t. we throw an exception rather than letting unexpected/undefined behavior occur.

Yeah. I agree we can test the latter. The place where you had added the comment threw me off. It should probably be in getPreviousContainers, where we readTaskAssignment.

> Also, I believe this is not equivalent to integer comparison.
In a sort scenario, for the same id's:
Integer comparison would yield [3, 8, 10]
String comparison would yield [10, 3, 8]
We should be certain that this will not affect any other code that uses ContainerModel as a comparator. e.g. for sort. If that code expects them to be in integer order for example, it could break. Or it could produce inconsistent behavior, different mappings of anything that is assigned to containers in a round-robin fashion.

I totally agree that comparison between Integer and String will result in different order. I am actually not even sure if "comparable" even makes sense anymore. I am trying to figure out where in our code we rely on "ordering" or "sorting" of ContainerModels. I couldn't find any use-case for this. 
In fact, I removed it and ran a build and it seems to work fine. However, that may not be sufficient as our test coverage is pretty poor :(

navina:
@jmakes @nickpan47 I have updated based on the feedback. Hopefully,I didn't miss any related to this SEP. Most recent changes:
* updated the documentation in ProcessorIdGenerator and ApplicationConfig
* Added validation in GroupByContainerCount
* Removed compareTo in ContainerModel

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
jmakes: The comment hasn't been removed yet.

navina: Missed it before. Comment fixed!

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
```diff
@@ -211,7 +220,7 @@ private void validateTasks(Set<TaskModel> tasks) {
   private List<TaskGroup> createContainers(int startContainerId, int endContainerId) {
     List<TaskGroup> containers = new ArrayList<>(endContainerId - startContainerId);
     for (int i = startContainerId; i < endContainerId; i++) {
```
nickpan47: nit: seems like this is the reason why this Grouper does not work w/ non-integer processorIds? We probably should open a JIRA to fix it.

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerIds.java
```diff
@@ -89,7 +89,7 @@ public GroupByContainerIds(int count) {
     // Convert to a Set of ContainerModel
     Set<ContainerModel> containerModels = new HashSet<>();
     for (int i = 0; i < containerCount; i++) {
-      containerModels.add(new ContainerModel(containersIds.get(i), taskGroups[i]));
+      containerModels.add(new ContainerModel(containersIds.get(i), -1, taskGroups[i]));
```
nickpan47: Not sure why here we are using -1 and in line 79 we are using 'i'? Better to be consistent. If there is a reason to use either '-1' or 'i', please add a comment to explain.

samza-core/src/main/java/org/apache/samza/coordinator/JobCoordinator.java
```diff
@@ -55,12 +59,16 @@
    * @throws InterruptedException if the current thread is interrupted while waiting for the JobCoordinator to start-up
    */
   boolean awaitStart(long timeoutMs) throws InterruptedException;
+
   /**
-   * Returns the logical ID assigned to the processor
-   * It is up to the user to ensure that different instances of StreamProcessor within a job have unique processor ID.
-   * @return integer representing the logical processor ID
+   * Returns the logical ID assigned to the processor that is local to the instance of StreamProcessor.
```
nickpan47: nit: why is this called "logic"? I would think this is a "physical" ID since it is specific to the underlying environment already.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
nickpan47: IMO, this would better be encapsulated in a loader method to get the ProcessorIdGenerator implementation class. No matter what implementation a JobCoordinator class takes, as long as it is loading ProcessorIdGenerator implementation, it should use the loader method and will perform this validation.

samza-core/src/main/java/org/apache/samza/runtime/UUIDGenerator.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.samza.config.Config;
+
+import java.util.UUID;
+
+public class UUIDGenerator implements ProcessorIdGenerator {
+  /**
+   * Generates a String representation to identify the processor instance
+   * This value can be representative of its current executing environment. It can also be custom-managed by the user.
+   * <p>
+   * <b>Note</b>: processorId has to be unique among the processors within a job
```
nickpan47: This does not follow the template we stated in the Javadoc (i.e. $x_$y format). It would better to be consistent with the documentation.

samza-api/src/main/java/org/apache/samza/runtime/ProcessorIdGenerator.java
```diff
@@ -0,0 +1,50 @@
...
+   *
+   * <b>Specification of processor identifier</b>:
+   * <ul>
+   *  <li>Processor identifier has to be unique among the processors within a job</li>
+   *  <li>When more than one processor co-exist within the same JVM, the processor identifier should be of the format:
+   *  $x_$y, where 'x' is a unique identifier for the executing JVM and 'y' is a unique identifier for the
+   *  processor instance within the JVM.</li>
+   * </ul>
+   *
+   * <b>Note</b>:
```
navina: I clarified in the second point that when one processor exists per JVM, it should be fine to simply use "x" part to identify the processor. "Note" itself gives an example of how it can be achieved.

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
navina: I think there is more than one place in this Grouper where the integer assumption is used. The intention was to re-write this grouper with the new standalone. I will make sure there is a JIRA for this. Thanks :)

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerIds.java
navina: We are setting it to -1 because there is no integer equivalent of the processorId anymore. 
I believe we are using "i" before just as index to the array of processorIds passed in the argument. 
I added a comment next to the -1 to clarify. Thanks!

samza-core/src/main/java/org/apache/samza/coordinator/JobCoordinator.java
navina: Yeah. This is a legacy comment. Let me fix it. Thanks for pointing it out.

samza-core/src/main/java/org/apache/samza/runtime/UUIDGenerator.java
navina: I have stated why the format exception is made in this use-case.

navina:
> No matter what implementation a JobCoordinator class takes, as long as it is loading ProcessorIdGenerator implementation, it should use the loader method and will perform this validation.

Our config validations are usually encapsulated in the corresponding "Config" class. Here, we are simply testing this for compatibility with existing streamprocessor users. Since it is temporary, I think it should be fine.

PR #104, opened 2017-03-30
SAMZA-1181: Fix AppMaster hang after submitting jobs to Yarn

shanthoosh:
Currently when a job is submitted to Yarn, it is going to hang after AppMaster is created. The log shows that it hangs during bootstrapping from Coordinator stream. Further debugging shows that the jobs hang in the second time of bootstrap while reading locality data from LocalityManager. The sequence is the following:
1. JobModelManager creates CoordinatorStreamConsumer, and bootstrap it,
2. LocalityManager writes locality info into coordinator stream
3. JobModelManager closes CoordinatorStreamConsumer 
4. Later localityManager bootstraps CoordinatorStreamConsumer again
Step 3 is the problem here. Since CoordinatorStreamConsumer is still held by LocalityManager, it cannot be closed prematurely. Step 3 is introduced in SAMZA-1154, as a refactoring of JobModelManager for task rest end point. To fix this issue, we will revert this change of step 3.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
+++ b/samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
@@ -81,76 +81,65 @@ object JobModelManager extends Logging {
    *                                from the coordinator stream, and instantiate a JobModelManager.
    */
   def apply(coordinatorSystemConfig: Config, metricsRegistryMap: MetricsRegistryMap): JobModelManager = {
-    var coordinatorSystemConsumer: CoordinatorStreamSystemConsumer = null
-    var coordinatorSystemProducer: CoordinatorStreamSystemProducer = null
-    try {
-      val coordinatorStreamSystemFactory: CoordinatorStreamSystemFactory = new CoordinatorStreamSystemFactory()
-      coordinatorSystemConsumer = coordinatorStreamSystemFactory.getCoordinatorStreamSystemConsumer(coordinatorSystemConfig, metricsRegistryMap)
-      coordinatorSystemProducer = coordinatorStreamSystemFactory.getCoordinatorStreamSystemProducer(coordinatorSystemConfig, metricsRegistryMap)
-      info("Registering coordinator system stream consumer.")
-      coordinatorSystemConsumer.register
-      debug("Starting coordinator system stream consumer.")
-      coordinatorSystemConsumer.start
-      debug("Bootstrapping coordinator system stream consumer.")
-      coordinatorSystemConsumer.bootstrap
-      info("Registering coordinator system stream producer.")
-      coordinatorSystemProducer.register(SOURCE)
+    val coordinatorStreamSystemFactory: CoordinatorStreamSystemFactory = new CoordinatorStreamSystemFactory()
+    val coordinatorSystemConsumer: CoordinatorStreamSystemConsumer = coordinatorStreamSystemFactory.getCoordinatorStreamSystemConsumer(coordinatorSystemConfig, metricsRegistryMap)
+    val coordinatorSystemProducer: CoordinatorStreamSystemProducer = coordinatorStreamSystemFactory.getCoordinatorStreamSystemProducer(coordinatorSystemConfig, metricsRegistryMap)
+    info("Registering coordinator system stream consumer.")
+    coordinatorSystemConsumer.register
+    debug("Starting coordinator system stream consumer.")
+    coordinatorSystemConsumer.start
+    debug("Bootstrapping coordinator system stream consumer.")
+    coordinatorSystemConsumer.bootstrap
+    info("Registering coordinator system stream producer.")
+    coordinatorSystemProducer.register(SOURCE)
 
-      val config = coordinatorSystemConsumer.getConfig
-      info("Got config: %s" format config)
-      val changelogManager = new ChangelogPartitionManager(coordinatorSystemProducer, coordinatorSystemConsumer, SOURCE)
-      changelogManager.start()
-      val localityManager = new LocalityManager(coordinatorSystemProducer, coordinatorSystemConsumer)
-      // We don't need to start() localityManager as they share the same instances with checkpoint and changelog managers.
-      // TODO: This code will go away with refactoring - SAMZA-678
+    val config = coordinatorSystemConsumer.getConfig
+    info("Got config: %s" format config)
+    val changelogManager = new ChangelogPartitionManager(coordinatorSystemProducer, coordinatorSystemConsumer, SOURCE)
+    changelogManager.start()
+    val localityManager = new LocalityManager(coordinatorSystemProducer, coordinatorSystemConsumer)
+    // We don't need to start() localityManager as they share the same instances with checkpoint and changelog managers.
+    // TODO: This code will go away with refactoring - SAMZA-678
 
-      localityManager.start()
+    localityManager.start()
 
-      // Map the name of each system to the corresponding SystemAdmin
-      val systemAdmins = getSystemAdmins(config)
+    // Map the name of each system to the corresponding SystemAdmin
+    val systemAdmins = getSystemAdmins(config)
 
-      val streamMetadataCache = new StreamMetadataCache(systemAdmins = systemAdmins, cacheTTLms = 0)
-      var streamPartitionCountMonitor: StreamPartitionCountMonitor = null
-      if (config.getMonitorPartitionChange) {
-        val extendedSystemAdmins = systemAdmins.filter{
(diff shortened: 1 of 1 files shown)


shanthoosh:
@xinyuiscool Please take a look when you get a chance.

navina:
Is there a JIRA associated with this PR? 

> There were issues with bootstrapping from CoordinatorStream

What were the issues with bootstrapping? 

Whatever was the problem with bootstrapping, did it exist from before or was it introduced recently due to some change?

vjagadish1989:
Hi @shanthoosh,

Thanks for the patch. 

Some questions / observations here:

1.
>> One BrokerProxy is created for a kafkaCluster (Host, Port) combination and shared by all KafkaSystemConsumer instances.

It's my understanding that Broker proxy threads are not shared across `KafkaSystemConsumer` instances. @jmakes / @navina can correct me.

2.
>> BrokerProxy for each (Host, Port) is stopped in KafkaSystemConsumer.stop and not recreated during successive KafkaSystemConsumer.create calls.

This is by design and should be always true. 

3.
>> Altering the KafkaSystemConsumer.create logic to recreate BrokerProxy
for each (Host, Port) impacts other portions of KafkaSystemConsumer

Wondering why this option is even worth considering? What problem are we solving?

4. 
>>  hence for now this patch removes the coordinatorSystemConsumer.stop call from finally block
in classes JobModelManager and SamzaTaskProxy.

I don't like stopping the producer and not the consumer asymmetrically. So, I'm -1 for this patch.

5.

>> There were issues with bootstrapping from CoordinatorStream because of the problem described above. This change resolves them.

What were the issues and is there a test to reproduce this? How does not stopping the consumer fix them? 

6. Is this a regression? If so, wondering which commit/feature introduced it?

shanthoosh:
@vjagadish1989  @navina 

Here is the detailed explanation of why I’m proposing to remove coordinatorStreamConsumer.stop, coordinatorStreamProducer.stop in JobModelManager.

> Problem: 

During AM startup, me & @xinyuiscool noticed that when we stop coordinatorStreamConsumer after 
jobModel is generated in JobModelManager, future reads from the coordinatorStream in AM doesn’t read until the highest watermark offset. 
Subsequents coordinatorStream consumption was until the last offset held in the cache(BlockingEnvelopeMap) of KafkaConsumer. 
We also noticed from the logs that the BrokerProxy was getting started only once and after call to coordinatorStreamConsumer.stop doesn’t start again before coordinator stream read.

@vjagadish1989 : 

For points 1,2,3 of your comment.

I misunderstood that BrokerProxy was shared by all KafkaSystemConsumer and felt that this is the reason why subsequent coordinator stream reads was not returning messages till highest offset.

For point 6 : 
> Why this change was made? Where was it introduced?  

As a part of JobModelManager refactoring, I added stop in the finally block of JobModelManager.apply method. 

Reason for doing it : These resources(coordinatorStreamConsumer, coordinatorStreamProducer) were created in the apply method and most of them goes out of scope after the method execution. 

With the intent to handle the lifecycle and free up resources, this change was done. 

A BrokerProxy thread is spawned for each KafkaSystemConsumer creation. 

And a KafkaSystemConsumer is created for each call to JobModelManager.apply. After the completion,  kind of felt this is unnecessary to leave this thread open, when it will be unused.

> What is the actual cause of this? Why was messages are not read till highest watermark offset? 

In jobModel creation, localityManager is passed in as a parameter and is used for reading ContainerLocality mappings.

Passing in a live resource(that holds open socket connections, threads) to a pojo object that should’ve been immutable kinds of now creates a global state dependency. 

Since the reference to both coordinatorStreamConsumer, coordinatorStreamProducer is held indirectly through jobModel(a client which has jobModel reference can get stale data if getContainerLocality is invoked after coordinatorStreamConsumer.close), it doesn’t make sense to stop them just after generating the jobModel.

In SamzaContainerProcessManager.start, we read containerToHostMappings from the jobModel, which in turn uses localityManager to read from it. 

Since the systemConsumer is closed after the job model creation , this read containerLocality just returns the messages to be returned only from the KafkaConsumer cache.

 
> Ideal fix would entail doing the following. 

* To handle lifecycle properly, should be just passing the container locality map(map<String, map<String, String>) in jobModel generation, which is beyond the scope of this patch. 
* Fixing the BalancingTaskNameGrouper interface which takes in localityManager. Since samza users can indirectly reference(coordinatorStreamConsumer, coordinatorStreamProducer)   in their taskNameGrouper implementation and can have logic based upon that. This interface needs to change before we can conclusively say lifecycle of this resource is now done.  


> I don't like stopping the producer and not the consumer asymmetrically.

 I’m not stopping both coordinatorStreamConsumer, coordinatorStreamProducer in it. 

> -1 for complicating lifecycle management. 

This was how it was before my refactoring. This  complication was introduced because of how jobModel was constructed which existed much prior to JobModelManager refactoring.  

Moreover, this is not a permanent fix to this problem. Just a interim solution to unblock @xinyuiscool testing, before we fix it properly. 

@vjagadish1989 if you’re still not convinced, we can discuss this offline.

shanthoosh:
@xinyuiscool : Can you please take a look when you're free.

navina:
@santhoshvenkat1988 I thought we discussed this in the morning. Before I respond to your comments, can you please move the description of the problem and its proposed solution to a JIRA? PR is not a place to be doing root-cause analysis. Please read [the docs](http://samza.apache.org/contribute/contributors-corner.html#contributing-to-apache-samza)

>  when we stop coordinatorStreamConsumer after jobModel is generated in JobModelManager, future reads from the coordinatorStream in AM doesn’t read until the highest watermark offset.

Which component is doing the "future reads" ? Afaik, the leader creates the JobModel and exposes it using the HTTP end-point. even though the goal is for the leader to continuously catch-up with coordinator stream, I don't believe we are doing that yet. Correct me if I am wrong.

>In jobModel creation, localityManager is passed in as a parameter and is used for reading ContainerLocality mappings.
> 
> Passing in a live resource(that holds open socket connections, threads) to a pojo object that should’ve been immutable kinds of now creates a global state dependency.

Agreed. 

>  it doesn’t make sense to stop them just after generating the jobModel.

That makes sense. But how are you proposing to fix this? Not handling life-cycle within JobModelManager and pushing it up to the callers is a fine model too. It doesn't look like that is what is in the patch though.

> To handle lifecycle properly, should be just passing the container locality map(map<String, map<String, String>) in jobModel constructor during creation, which is beyond the scope of this patch.

It seems like you have identified a real issue here. However, I disagree on the fix being outside the scope of this patch. In fact, this PR seems to have identified a good deal of issues we have. But it doesn't address any of them, leaving behind loopholes. -1 on the scoping of the issue/patch

> This interface needs to change before we can conclusively say lifecycle of this resource is now done.

It is not straightforward to change a public interface. Please create a SEP or at the very least a JIRA for everyone to understand why you are proposing this change. 

> This was how it was before my refactoring. This complication was introduced because of how jobModel was constructed which existed much prior to JobModelManager refactoring.

It is awesome that you have identified a crucial bug/issue. Rather than taking it back to the broken state that it was, I prefer to fix it forward. 

> Just a interim solution to unblock @xinyuiscool testing, before we fix it properly.

What part of this fix is @xinyuiscool blocked on?

Overall, I don't understand the basis behind the solution in this PR. -1 for this change. Please create a JIRA for further discussions. Thanks!

xinyuiscool:
This is the fix for the coordinator stream bootstrap hanging issue in appmaster. The bug is introduced by the JobCoordinator refactoring patch, which closes the coordinator consumer/producer prematurely (they are still being used later by locality manager). Since a clean approach to solve this problem is a much larger scope and requires more testing, I think it's okey to revert to old logic. With this fix, 0.13 testing can be unblocked.

I completely agree we need a JIRA to clearly say what's the problem we have right now and why we are making this change for now. And we should have another JIRA tracking the future cleanup. @shanthoosh : could you please create the JIRAs?

shanthoosh:
@navina 

Due to losing access to my dev machine & losing stored passwords for all website accounts because of that,  was unable to create the JIRA(wasn't able to login to JIRA). Now recovering access, will update it here once done.

navina:
> I think it's okey to revert to old logic

From what I understand, reverting this logic doesn't solve the core issues identified through this PR discussion. What testing in 0.13 are we talking about if we don't fix known issues? May be it is related to the upcoming MVP release date. But I am not comfortable releasing a version with known bugs. 

>Now recovering access, will update it here once done.

Thank you.

xinyuiscool:
@navina: the bug introduced previously is actually a major blocker issue. Samza jobs hang in yarn because of this (no testing can be done based on the master branch). I think we need a fix asap. After that @shanthoosh  can work on a proper fix, which can still be in the 0.13 release.

shanthoosh:
@navina 
Updated PR with JIRA created. Can you take a look when you're free.

PR #106, opened 2017-03-31
SAMZA-1089: Runner should support kill and status commands

jmakes:


samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
```diff
@@ -54,8 +54,12 @@ void addTargetNode(ProcessorNode targetNode) {
     targetNodes.add(targetNode);
   }
 
-  StreamSpec getStreamSpec() {
-    return streamSpec;
+  public StreamSpec getStreamSpec() {
+    if (partitions == PARTITIONS_UNKNOWN) {
+      return streamSpec;
+    } else {
+      return streamSpec.copyWithPartitionCount(partitions);
```
prateekm: @xinyuiscool to verify this won't break the traversals.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -46,25 +53,91 @@ public RemoteApplicationRunner(Config config) {
...
+      ProcessorGraph processorGraph = getExecutionPlan(app);
 
-      // 2. create the physical execution plan
-      ExecutionPlanner planner = new ExecutionPlanner(config);
-      ProcessorGraph processorGraph = planner.plan(streamGraph);
+      // 2. create the necessary streams
+      List<StreamSpec> streams = processorGraph.getIntermediateStreams().stream()
+          .map(streamEdge -> streamEdge.getStreamSpec())
+          .collect(Collectors.toList());
+      StreamManager streamManager = new StreamManager(new JavaSystemConfig(config).getSystemAdmins());
```
xinyuiscool: Can we create StreamManager in the beginning of run() and pass it into the planner so the system admins won't be created twice?

samza-core/src/main/java/org/apache/samza/config/JavaSystemConfig.java
```diff
@@ -60,4 +66,33 @@ public String getSystemFactory(String name) {
     }
     return systemNames;
   }
+
+  /**
+   * @return map of system name to {@link SystemAdmin}
+   */
+  public Map<String, SystemAdmin> getSystemAdmins() {
```
prateekm: Should we update Scala SystemConfig for parity as well? Not sure what the difference b/w the two is.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
jmakes: Great idea. It won't over-complicate the plan() method, will it?

If not, I'll make the change.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -46,25 +53,91 @@ public RemoteApplicationRunner(Config config) {
...
+        Config processorConfig = processor.generateConfig();
+        JobRunner runner = new JobRunner(processorConfig);
+        ApplicationStatus status = runner.status();
+        log.debug("Status is {} for processor {} with config {}", new Object[]{status, processor.getId(), processorConfig});
+
+        switch (status) {
+          case Running:
+            return ApplicationStatus.Running;
+          case UnsuccessfulFinish:
+            unsuccessfulFinish = true;
```
prateekm: IIUC, probably cleaner to return ApplicationStatus.UnsuccessfulFinish immediately here. Don't need the `unsuccessfulFinish` boolean that way.

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -35,7 +35,6 @@ import org.apache.samza.config.MapConfig
 import org.apache.samza.config.SystemConfig
 import org.apache.samza.config.JobConfig.Config2Job
 import org.apache.samza.config.SystemConfig.Config2System
-
```
prateekm: Revert?

samza-core/src/main/java/org/apache/samza/config/JavaSystemConfig.java
jmakes: What do you suggest? There's currently no method in the scala SystemConfig to get the factories or the admins, only the names.

samza-core/src/main/java/org/apache/samza/execution/StreamManager.java
```diff
@@ -0,0 +1,73 @@
...
+
+public class StreamManager {
+  private static final Logger LOGGER = LoggerFactory.getLogger(StreamManager.class);
+
+  private final Map<String, SystemAdmin> sysAdmins;
+
+  public StreamManager(Map<String, SystemAdmin> sysAdmins) {
+    this.sysAdmins = sysAdmins;
+  }
+  public void createStreams(List<StreamSpec> streams) {
```
jmakes: I blame code folding in IntelliJ.  
Thanks 👍

samza-core/src/main/java/org/apache/samza/execution/StreamManager.java
```diff
@@ -0,0 +1,73 @@
...
+    Multimap<String, StreamSpec> streamsGroupedBySystem = HashMultimap.create();
+    streams.forEach(streamSpec ->
+      streamsGroupedBySystem.put(streamSpec.getSystemName(), streamSpec));
+
+
```
jmakes: No excuse for this one. :)

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
```diff
@@ -34,19 +34,48 @@
...
-  private static final String STREAM_APPLICATION_CLASS_CONFIG = "app.class";
+  public static final String STREAM_APPLICATION_CLASS_CONFIG = "app.class";
+
+  public static class ApplicationRunnerCommandLine extends CommandLine {
+    public OptionSpec operationOpt =
+        parser().accepts("operation", "The operation to perform; run, status, kill.")
+            .withRequiredArg()
+            .ofType(String.class)
+            .describedAs("operation=run")
+            .defaultsTo("run");
```
jmakes: Legacy support in JobRunner. It never required this parameter before and would always execute run(). 

Defaulting to run just results in cleaner code than making the argument optional, checking if it's missing, and then executing run()

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
```diff
@@ -34,19 +34,48 @@
...
+          (StreamApplication) Class.forName(config.get(STREAM_APPLICATION_CLASS_CONFIG)).newInstance();
+      switch (op) {
+        case RUN:
+          runner.run(app);
+          break;
+        case KILL:
+          runner.kill(app);
+          break;
+        case STATUS:
+          System.out.println(runner.status(app));
```
jmakes: The invoking method. 

At LI, the control script will see this output and interpret it. 

For the record, I don't love this approach, but the only other idea I had was to encode meanings into a set of exit codes.

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerOperation.java
```diff
@@ -0,0 +1,41 @@
...
+
+  public static ApplicationRunnerOperation fromString(String string) {
+    return ApplicationRunnerOperation.valueOf(string.toUpperCase());
+  }
+
+  ApplicationRunnerOperation(String str) {
+    this.str = str;
+  }
+
+  public String toString() {
```
jmakes: 👍

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -41,4 +42,13 @@ public LocalApplicationRunner(Config config) {
     // 5. create the configuration for StreamProcessor
     // 6. start the StreamProcessor w/ optimized instance of StreamApplication
   }
+
+  @Override
+  public void kill(StreamApplication streamApp) {
+  }
```
jmakes: This whole class is unimplemented thus far. I figured if I added the exception here, whomever eventually implements the class may assume the exceptions were the appropriate behavior. 

Ultimately, all 3 methods need implementations and I think @xinyuiscool is planning to do it.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -46,25 +53,91 @@ public RemoteApplicationRunner(Config config) {
...
   }
+
+  @Override
+  public void kill(StreamApplication app) {
+    try {
+      ProcessorGraph processorGraph = getExecutionPlan(app);
+
+      processorGraph.getProcessorNodes().forEach(processor -> {
+          Config processorConfig = processor.generateConfig();
+          log.info("Killing processor {} with config {}", processor.getId(), processorConfig);
```
jmakes: Fair enough. Less is more. I'm going to omit the config for now. I think we use the console appender for the JobRunner/ApplicationRunner, so most of this is just for debugging anyway.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
jmakes: Based on a discussion with @nickpan47 and @xinyuiscool, the agreement was to return Running if any of the Jobs was running, similar to YARN's application status when not all the Containers are running.

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
jmakes: There is only one caller to this method and it's on the code path I modified/verified. :-)

samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
```diff
@@ -79,13 +80,28 @@ public ApplicationRunner(Config config) {
...
   /**
+   * Kill the Samza jobs represented by {@link StreamApplication}
+   *
+   * @param streamApp  the user-defined {@link StreamApplication} object
+   */
+  public abstract void kill(StreamApplication streamApp);
+
+  /**
+   * Get the collective status of the Samza jobs represented by {@link StreamApplication}.
+   * Returns {@link ApplicationStatus#Running} if any of the jobs are running.
```
prateekm: Discussed offline. We can add it to ApplicationStatus when we have multi-stage jobs.

PR #108, opened 2017-04-03
SAMZA-1183: Fix flaky tests in TestAsyncRunLoop

shanthoosh:
Changes:

* Moved all states locally within the test methods(no global state, since tests can be run parallely).
* Increased the timeout waiting for all task threads to finish.
* Changed message chooser mock to return null at the end(When EOS message is used a final msg, during some runs task.process gets invoked more than once for it. This is due to test setup, other tests within this test class also do the same.)

How was it verified: 

Ran the test class repeatedly to check for failures(there were none) using following script.

`
while [ $i -lt 30 ] do     sleep 3;     i=`expr $i + 1`;     ./gradlew :samza-core:clean :samza-core:test -Dtest.single=TestAsyncRunLoop  done
`

### DIFF
--- a/samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
+++ b/samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
@@ -29,7 +29,6 @@ import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.samza.Partition;
 import org.apache.samza.checkpoint.OffsetManager;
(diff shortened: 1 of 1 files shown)


shanthoosh:
@xinyuiscool @navina 

navina:
> (When EOS message is used a final msg, during some runs task.process gets invoked more than once for it. This is due to test setup, other tests within this test class also do the same.

Why does the test setup invoke task.process more than once when EOS message is used?

Isn't SAMZA-1183 supposed to address testProcessInOrder test as well that was commented out?

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
```diff
@@ -567,15 +566,15 @@ public void testCommitBehaviourWhenAsyncCommitIsEnabled() throws InterruptedExce
         try {
           firstMsgCompletionLatch.await();
           secondMsgCompletionLatch.await();
-          Thread.sleep(100);
+          Thread.sleep(500);
```
prateekm: Is there a way to do these tests using a settable Clock vs Thread sleeps?

shanthoosh:
@navina 
> Isn't SAMZA-1183 supposed to address testProcessInOrder test as well that was commented out?

Ideally, sync commit tests(testProcessInOrder) shouldn't be affected. Code associated with them aren't modified for ages. @xinyuiscool can correct me here, if i'm wrong. Despite that we're seeing failures. Success/Failure of individual tests depends upon proper initialization of shared global state in before method.  In parallel execution of tests, assignments can get interleaved leading to race conditions. 

As a fix I've moved object hierarchy construction to individual test methods & local build's green.

> Why does the test setup invoke task.process more than once when EOS message is used?

I should have phrased that better. 

There's a lag between processing last message & shutting the executor down in tests(This was allowed since in some runs executor is shutdown before task threads complete). Due to which, the last value returned by the mock will be the value it returns throughout the lag period(this is how mocks work). So it keeps on hitting task.process method with that EOS message. When chooser returns null as last value, AsyncRunLoop doesn’t hit task instances(hence the change).

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
shanthoosh: From what I understand, clock instance we pass in through AsyncRunloop constructor can't be used to control/achieve that.  Can you please throw more light on this comment?

shanthoosh:
Build is successful here : https://builds.apache.org/job/samza-pr-checks/171/

When we restart it, git-jenkins plugin doesn't update here properly.

navina:
> As a fix I've moved object hierarchy construction to individual test methods & local build's green.

Whether the test has been modified recently or not, a failure is something we need to fix. By moving the object hierarchy construction, are you also addressing the testProcessInOrder failures? That is my question. 

> There's a lag between processing last message & shutting the executor down in tests(This was allowed since in some runs executor is shutdown before task threads complete). Due to which, the last value returned by the mock will be the value it returns throughout the lag period(this is how mocks work). So it keeps on hitting task.process method with that EOS message. When chooser returns null as last value(instead of EOS), AsyncRunLoop doesn’t hit task instances(hence the change).

I have tried to condense your verbose comment into what I understand - Basically, the mock consumer used was causing the EOS to be delivered more than once, when the executor was shutdown before task threads complete.  
Hence, my follow-up question is - Why is the executor shutdown before the tasks complete? Is that the expected behavior?

shanthoosh:
@navina 

Discussed with you offline.
> By moving the object hierarchy construction, are you also addressing the testProcessInOrder failures? That is my question.

Code & tests associated with it are unchanged for very long time, hence surprising to see failures. Yes, it will fix it(no common global state shared by test methods which is potential cause).

> Why is the executor shutdown before the tasks complete? Is that the expected behavior?

Yes, task threads might take more time to finish. Shutting down executor before it, will cause assertion failure(depending upon the timing of events). The change I want to test is AsyncTaskState state transition, however everything within AsyncRunLoop(AsyncTaskState) is private(Changing that would be harder) and all other tests follow the same pattern. Modifying them would be a harder & bigger change(may be we can create a JIRA for it to work in future)

shanthoosh:
@xinyuiscool  Removed physical wait(Thread.sleep) from TestAsyncRunLoop. Build is red due to  TestJmxReporter failures(unrelated to this change)

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
shanthoosh: Removed Thread sleeps from tests. Accomplishing same thing by triggering coordinator.shutdown on a particular message in task impl.

prateekm:
@shanthoosh This doesn't remove the wait, only the Thread.sleep(). I still see `callbackExecutor.awaitTermination(600, TimeUnit.MILLISECONDS);`.

shanthoosh:
@prateekm  

You're right. 

Removing callbackExecutor.await from all tests & testing just the behavior, if i understand right would involve the following changes.

* Changing access specifier for testable internal class/methods to protected.
* Changing all tests(which do this) in TestAsyncRunLoop to achieve uniformity. 

Is it okay to create JIRA & solve it later.

navina:
> Is it okay to create JIRA & solve it later.
The goal of the JIRA is to fix the tests (flaky or not) in TestAsyncRunloop. I don't see a lot of value in postponing this task in yet another "fix test" JIRA. I would rather implement the right fix now rather than later, while we have context on what the correct implementation is.

shanthoosh:
@navina 

> The goal of the JIRA is to fix the tests (flaky or not) in TestAsyncRunloop. I don't see a lot of value in postponing this task in yet another "fix test" JIRA. 

The goal was to actually fix the flaky test as described in the JIRA. 

However, it makes sense to remove awaitTermination(physical wait) here(since it will wait 'x' seconds every time, even though all the task threads finishes before that within the test). 

I've an idea(which I think will work), will discuss it with xinyu(since it will touch all tests in this class) and update it here.

navina:
> I've an idea(which I think will work), will discuss it with xinyu(since it will touch all tests in this class) and update it here.

Sounds good. Let's not carry over tech-debts. Thanks 👍

shanthoosh:
@navina 

PR #110, opened 2017-04-04
SAMZA-1178: Generate JSON from StreamPlan

xinyuiscool:
As the first step to visualize the StreamGraph/Plan, this patch generates a json representation of it. For the example StreamGraph in TestPlanJsonGenerator, here is the json that is generated (https://github.com/xinyuiscool/examples/blob/master/example_plan.json). This patch also introduces the immutable StreamPlan and did some code cleanup.


xinyuiscool:
@jmakes and @prateekm : please take a look at this patch. Thanks.

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
```diff
@@ -45,7 +45,7 @@
  * Note that intermediate streams are both the input and output of a JobNode in JobGraph.
  * So the graph may have cycles and it's not a DAG.
  */
-public class JobGraph {
+/* package private */ class JobGraph {
```
prateekm: Minor: Can we just leave out the `/*package private */` everywhere?

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -102,12 +94,11 @@ public ApplicationStatus status(StreamApplication app) {
...
 
-      JobGraph jobGraph = getExecutionPlan(app);
-      for (JobNode job : jobGraph.getJobNodes()) {
-        Config jobConfig = job.generateConfig();
+      StreamPlan plan = getExecutionPlan(app);
+      for (JobConfig jobConfig : plan.getJobConfigs()) {
         JobRunner runner = new JobRunner(jobConfig);
         ApplicationStatus status = runner.status();
-        log.debug("Status is {} for jopb {}", new Object[]{status, job.getId()});
+        log.debug("Status is {} for jopb {}", new Object[]{status, jobConfig.getName()});
```
prateekm: Typo in `jobp`.

samza-core/src/test/java/org/apache/samza/execution/TestUtil.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.samza.config.Config;
+import org.apache.samza.job.ApplicationStatus;
+import org.apache.samza.operators.functions.JoinFunction;
+import org.apache.samza.runtime.AbstractApplicationRunner;
+import org.apache.samza.runtime.ApplicationRunner;
+import org.apache.samza.system.SystemAdmin;
+import org.apache.samza.system.SystemStreamMetadata;
+import org.apache.samza.system.SystemStreamPartition;
+
+public class TestUtil {
```
prateekm: Prefer more descriptive/specific name.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -57,7 +59,7 @@ public ExecutionPlanner(Config config, StreamManager streamManager) {
     this.streamManager = streamManager;
   }
 
-  public JobGraph plan(StreamGraph streamGraph) throws Exception {
+  public StreamPlan plan(StreamGraph streamGraph) throws Exception {
```
prateekm: `StreamPlan` doesn't seem very meaningful. Maybe `ExecutionPlan`?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -66,7 +68,12 @@ public JobGraph plan(StreamGraph streamGraph) throws Exception {
       calculatePartitions(streamGraph, jobGraph);
     }
 
-    return jobGraph;
+    List<JobConfig> jobConfigs = jobGraph.getJobNodes().stream().map(JobNode::generateConfig).collect(Collectors.toList());
+    String planJson = new PlanJsonGenerator().toJson(jobGraph);
+    List<StreamSpec> intermediateStreams = jobGraph.getIntermediateStreams().stream()
+        .map(streamEdge -> streamEdge.getStreamSpec())
+        .collect(Collectors.toList());
+    return new StreamPlan(streamGraph, jobConfigs, intermediateStreams, planJson);
```
prateekm: It's a little weird that the JSON representation of the plan is required to construct it, and that this class needs to know about the PlanJsonGenerator. Maybe add a helper to StreamPlan that returns the JSON representation instead? Maybe even override its toString?

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
```diff
@@ -0,0 +1,266 @@
...
+    public int getPartition() {
+      return partition;
+    }
+
+    public void setPartition(int partition) {
+      this.partition = partition;
+    }
+  }
+
+  public static final class GraphNodes {
```
prateekm: It's a little strange that there are public classes for all these nodes that are only there to generate the Plan representation.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
xinyuiscool: I name it StreamPlan to be consistent with StreamApplication, StreamGraph, StreakTask .. :). In Flink it's called StreamingPlan so I guess I will keep it for now.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
xinyuiscool: Fixed. Now the plan has both StreamGraph and JobGraph, and the objects/methods are internal to the plan.

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
xinyuiscool: Good catch! fixed.

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
xinyuiscool: Oh, I think the public is not needed here. I removed it to be package private (for testing).

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
xinyuiscool: Fixed.

samza-core/src/test/java/org/apache/samza/execution/TestUtil.java
xinyuiscool: Fixed.

xinyuiscool:
@prateekm : I think I also fixed the stuff you mentioned in your early feedback. As we discussed, NodeId is the ID for either operator node or stream node, and it is used to construct the topology of the graph. For rendering, stream is also a node in this json graph (we will start from the source stream). For join, now two operator nodes are created to keep it in sync with StreamGraph. Each node has a reference of the other in the pair. I updated the link to the example json. Please take a look again.

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
```diff
@@ -0,0 +1,295 @@
...
+import org.codehaus.jackson.map.ObjectMapper;
+
+/**
+ * This class generates the json representation of the plan.
+ */
+public class PlanJsonGenerator {
+
+  private static abstract class GraphNode {
+    private Integer nodeId;
+    private Set<Integer> successors = new HashSet<>();
```
prateekm: Maybe `nextNodes` to indicate that the int is the nodeId (as opposed to streamId/opId)?

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
```diff
@@ -0,0 +1,295 @@
...
+
+    public Set<Integer> getSuccessors() {
+      return successors;
+    }
+  }
+
+  static final class OpNode extends GraphNode {
+    private String opCode;
+    private int opId;
+    private int pairNodeId = -1;  //for join operator, we will have a pair nodes for two partial joins
```
prateekm: Naming: `pairedNodeId`

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
```diff
@@ -0,0 +1,295 @@
...
+      node.setStreamId(streamId);
+      node.setSystem(edge.getStreamSpec().getSystemName());
+      node.setPhysicalName(edge.getStreamSpec().getPhysicalName());
+      node.setPartition(edge.getPartitionCount());
+      streamNodes.put(streamId, node);
+    }
+    return node;
+  }
+
+  private static <T extends GraphNode> List<T> createListAndSort(Collection<T> nodes) {
```
prateekm: Minor: Name makes it sound like a more general helper than it is. Maybe `sortGraphNodesById`?

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  private void buildJsonGraph(MessageStream inputStream, GraphNode node, Map<Integer, OpNode> opNodes, Map<OutputStream, StreamEdge> outputToStreamSpecs) {
+    Collection<OperatorSpec> specs = ((MessageStreamImpl) inputStream).getRegisteredOperatorSpecs();
+    for (OperatorSpec spec : specs) {
+      GraphNode next = getOrCreateOpNode(opNodes, spec);
+      node.getSuccessors().add(next.getNodeId());
+
+      if (spec instanceof SinkOperatorSpec) {
+        StreamEdge streamEdge = outputToStreamSpecs.get(((SinkOperatorSpec) spec).getOutStream());
```
prateekm: How does this work? After my changes in https://github.com/apache/samza/pull/92, SinkOperatorSpec#getNextStream returns null since it's a terminal operator. Did I get something wrong?

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
```diff
@@ -0,0 +1,295 @@
...
+
+  private int nextID = 0;
+  // streamId to StreamNode mapping
+  private final Map<String, StreamNode> streamNodes = new HashMap<>();
+  // Mapping from the output stream to the join spec. Since StreamGraph creates two partial join operators for a join and they
+  // will have the same output stream, this mapping is used to choose one of them as the unique join spec representing this join
+  // (who register first in the map wins).
+  Map<MessageStream, OperatorSpec> outputStreamToJoinSpec = new HashMap<>();
+
+  /* package private */ String toJson(JobGraph jobGraph) throws Exception {
```
prateekm: Looks like this class is stateful and is meant to be created once and saved for reuse. Is this method idempotent?

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
```diff
@@ -0,0 +1,295 @@
...
+        .collect(Collectors.toList());
+    nodes.setJobs(jobs);
+    nodes.setStreams(createListAndSort(streamNodes.values()));
+    ByteArrayOutputStream out = new ByteArrayOutputStream();
+    ObjectMapper mapper = new ObjectMapper();
+    mapper.writeValue(out, nodes);
+    return new String(out.toByteArray());
+  }
+
+  private Job buildJob(JobNode jobNode, StreamGraphImpl streamGraph, JobGraph jobGraph) {
```
prateekm: Minor: Maybe `toJob`?

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
```diff
@@ -0,0 +1,295 @@
...
+      });
+
+    Job job = new Job();
+    job.setJobName(jobNode.getJobName());
+    job.setJobId(jobNode.getJobId());
+    job.setOpNodes(createListAndSort(opNodes.values()));
+    return job;
+  }
+
+  private void buildJsonGraph(MessageStream inputStream, GraphNode node, Map<Integer, OpNode> opNodes, Map<OutputStream, StreamEdge> outputToStreamSpecs) {
```
prateekm: Needs documentation.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
prateekm: StreamingPlan seems fine, but StreamPlan doesn't make sense by itself. @jmakes, any thoughts? We can also rename later, not a blocker.

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
```diff
@@ -0,0 +1,295 @@
...
+import org.apache.samza.operators.spec.PartialJoinOperatorSpec;
+import org.apache.samza.operators.spec.SinkOperatorSpec;
+import org.codehaus.jackson.map.ObjectMapper;
+
+/**
+ * This class generates the json representation of the plan.
+ */
+public class PlanJsonGenerator {
+
+  private static abstract class GraphNode {
```
prateekm: Should add documentation for these classes, and generally what the Graph structure & representation is.

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
jmakes: fwiw, I like it. It assures future readers that the qualifier wasn't forgotten and that it was package private by design.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
jmakes: Putting it another way, I think it's more accurate to say this object is "a plan for execution" than "a plan for streaming". Hence ExecutionPlan.

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
jmakes: @prateekm getNextStream() returns null but getOutStream() doesn't. It seems a Sink has an OutputStream but not an output MessageStream

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
jmakes: There's definitely something odd about the naming or modularity. 

I'd expect a method called "buildJob"  to have an implementation that ONLY has the last 5 lines here.  So like:
`

  private Job buildJob(JobNode jobNode, Map<Integer, OpNode> opNodes) {

     Job job = new Job();
     job.setJobName(jobNode.getJobName());
     job.setJobId(jobNode.getJobId());
     job.setOpNodes(createListAndSort(opNodes.values()));
     return job;
   }

`
It does what its name implies. It builds a job and does NOTHING else.

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
jmakes: +1 Particularly because it seems to mutate some args and not others. It's not evident from the method signature what it does and how the output is captured.

samza-core/src/main/java/org/apache/samza/execution/StreamPlan.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.samza.system.StreamSpec;
+
+
+/**
+ * This class represents Samza {@link org.apache.samza.application.StreamApplication}
+ * plans for physical execution.
+ */
+public class StreamPlan {
+  private final StreamGraph streamGraph;
+  private final JobGraph jobGraph;
```
jmakes: Commentary (no action in this PR):
This is the most compelling evidence yet that StreamGraph needs to be renamed. When I first read this, I thought the StreamPlan was keeping separate graphs to track the streams and the jobs. 

Now if they were named more meaningfully, like 
ApplicationLogic
JobGraph

I'd have a much better idea of what this plan was tracking.

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
```diff
@@ -0,0 +1,295 @@
...
+import org.apache.samza.operators.StreamGraphImpl;
+import org.apache.samza.operators.spec.OperatorSpec;
+import org.apache.samza.operators.spec.PartialJoinOperatorSpec;
+import org.apache.samza.operators.spec.SinkOperatorSpec;
+import org.codehaus.jackson.map.ObjectMapper;
+
+/**
+ * This class generates the json representation of the plan.
+ */
+public class PlanJsonGenerator {
```
jmakes: This feels like an unnecessarily home-grown solution. 

Why not add an entry to SamzaObjectMapper implementing the Jackson JsonSerializer and let Jackson do the rest?

samza-core/src/test/java/org/apache/samza/execution/TestExecutionUtils.java
```diff
@@ -0,0 +1,117 @@
...
+      }
+
+      @Override
+      public ApplicationStatus status(StreamApplication streamApp) {
+        return null;
+      }
+    };
+  }
+
+  static JoinFunction createJoin() {
```
jmakes: nit: this really has nothing to do with "execution". Perhaps belongs in a separate TestOperatorUtils

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
prateekm: @jmakes This seems pretty reasonable. It was on every method in this class earlier.

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
prateekm: getOutStream() was renamed to getNextStream() in #92 to distinguish b/w actual/physical output streams and the next MessageStream in the operator DAG. The OutputStream in SinkOperatorSpec is the one that actual output is sent to. It has no next stream since it's a terminal operator. partitionBy returns a new InputStream to chain further operations on. Does that make sense?

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
prateekm: Agree that this feels odd, but prefer not adding to SamzaObjectMapper. It seems like that class is named more generally than what it is - it's actually the JSON serdes for coordinator stream contents. Is that right?

samza-core/src/test/java/org/apache/samza/execution/TestExecutionUtils.java
prateekm: Also nothing to do with Operators, it's just some common code b/w ExecutionPlanner layer (and below) test classes. One of these methods (createRunner) will probably be irrelevant after the refactor in #92. Also, @xinyuiscool why not use mocks for these instead of creating actual instances?

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
jmakes: Is it committed? https://github.com/apache/samza/blob/8515448a2023ae6c78b9b0bb8e297cf346775e13/samza-core/src/main/java/org/apache/samza/operators/spec/SinkOperatorSpec.java#L109

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
jmakes: @prateekm I don't see the stigma with SamzaObjectMapper. It could be renamed or split up if necessary.  As a compromise any other standard Jackson implementation would be fine. But we should not reinvent the wheel.

prateekm: @jmakes To add context to what I meant: This class is also using the regular Jackson ObjectMapper to serialize its data (in lines 203-206). Rest/most of the code + logic is for creating the data structures/POJOs to be serialized. What' I'm proposing is that instead of creating new data structures for serialization it'd be better if we could serialize the existing classes (JobGraph, JobNode, StreamEdge). That would get rid of this `PlanJsonGenerator`. Whether the actual serialization is done using a global registry like SamzaObjectMapper or using an ObjectMapper directly would then be a separate discussion.

jmakes: Makes sense and it sounds like that would help. My focus was on the apparent inversion of the logic to have this homegrown class as the encapsulation to convert the plan to json and using Jackson under the covers. That's opposed to exposing a standard Jackson JsonSerializer that has some custom logic under the covers.

samza-core/src/main/java/org/apache/samza/execution/PlanJsonGenerator.java
xinyuiscool: Removed. Did a lot of clean up and now it should be much better to read now.

samza-core/src/main/java/org/apache/samza/execution/StreamPlan.java
xinyuiscool: :). Based on the feedback, I refactored the code, and now ExecutionPlan is an interface, and JobModel implements it. I think now it looks cleaner.

samza-core/src/test/java/org/apache/samza/execution/TestExecutionUtils.java
xinyuiscool: Good point! I used mock for JoinFunction. ApplicationRunner also uses mock from @prateekm 's patch. The class is removed now.

xinyuiscool:
I merged @prateekm 's recent patch and addressed all the comments. Please take a look again. Thanks!

xinyuiscool:
btw, also updated the example json here (https://github.com/xinyuiscool/examples/blob/master/example_plan.json).

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -59,18 +56,14 @@ public RemoteApplicationRunner(Config config) {
   public void run(StreamApplication app) {
     try {
       // 1. initialize and plan
-      JobGraph jobGraph = getExecutionPlan(app);
+      ExecutionPlan plan = getExecutionPlan(app);
```
jmakes: This whole class reads much better now. Nice work!

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
```diff
@@ -54,23 +55,49 @@
...
-    this.streamGraph = streamGraph;
+  JobGraph(Config config) {
     this.config = config;
   }
 
   /**
+   * Returns the configs for single stage job, in the order of topologically sort.
+   * @return list of job configs
+   */
+  public List<JobConfig> getJobConfigs() {
```
prateekm: Minor: @Override for public interface methods and remove javadoc (it'll be inherited).

samza-core/src/test/java/org/apache/samza/execution/TestExecutionPlanner.java
```diff
@@ -65,26 +65,7 @@
...
-      public Object getSecondKey(Object message) {
-        return null;
-      }
-    };
-  }
-
-  private SystemAdmin createSystemAdmin(Map<String, Integer> streamToPartitions) {
+  static SystemAdmin createSystemAdmin(Map<String, Integer> streamToPartitions) {
 
     return new SystemAdmin() {
```
prateekm: Would still prefer using mocks. Makes the expectations in tests much more clearer.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -0,0 +1,233 @@
...
+import org.apache.samza.operators.spec.OperatorSpec;
+import org.apache.samza.operators.spec.PartialJoinOperatorSpec;
+import org.apache.samza.operators.spec.SinkOperatorSpec;
+import org.codehaus.jackson.annotate.JsonProperty;
+import org.codehaus.jackson.map.ObjectMapper;
+
+/**
+ * This class generates the JSON representation of the {@link JobGraph}.
+ */
+public class JobGraphJsonGenerator {
```
prateekm: Minor: package private?

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -0,0 +1,233 @@
...
+
+  /**
+   * This class provides the necessary connection of operators for traversal.
+   */
+  static abstract class Traversable {
+    @JsonProperty("NextOperatorIds")
```
prateekm: Let's use camelCase for field names in JSON.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -0,0 +1,233 @@
...
+    int opId;
+    @JsonProperty("OutputStreamId")
+    String outputStreamId;
+    @JsonProperty("PairedOpId")
+    int pairedOpId = -1;  //for join operator, we will have a pair nodes for two partial joins
+  }
+
+  static final class StreamSpecJson {
+    @JsonProperty("Id")
+    String id;
```
prateekm: `streamId`

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -0,0 +1,244 @@
...
+    int partitionCount;
+  }
+
+  static final class StreamEdgeJson {
+    @JsonProperty("StreamSpec")
+    StreamSpecJson streamSpec;
+  }
+
+  static final class OperatorGraphJson {
+    @JsonProperty("InputStreams")
```
prateekm: OutputStreams should be here as well for consistency.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -0,0 +1,244 @@
...
+    @JsonProperty("SystemName")
+    String systemName;
+    @JsonProperty("PhysicalName")
+    String physicalName;
+    @JsonProperty("PartitionCount")
+    int partitionCount;
+  }
+
+  static final class StreamEdgeJson {
+    @JsonProperty("StreamSpec")
```
prateekm: Let's add source and destination jobIds as well to represent the connectivity information.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -0,0 +1,244 @@
...
+        InputStreamJson inputJson = new InputStreamJson();
+        inputJson.streamId = streamSpec.getId();
+        opGraph.inputStreams.add(inputJson);
+        updateOperatorGraphJson((MessageStreamImpl) stream, inputJson, opGraph);
+      });
+    return opGraph;
+  }
+
+  /**
+   * Traverse the {@StreamGraph} recursively and update the operator graph JSON POJO.
```
prateekm: Minor: Will be good to add more descriptive documentation (and maybe use a more specific name) about what this is doing.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -0,0 +1,244 @@
...
+      });
+  }
+
+  /**
+   * Get or create the JSON POJO for an operator.
+   * @param opSpec {@link OperatorSpec}
+   * @param opGraph {@link org.apache.samza.execution.JobGraphJsonGenerator.OperatorGraphJson}
+   * @return {@link org.apache.samza.execution.JobGraphJsonGenerator.OperatorJson}
+   */
+  private OperatorJson getOrCreateOperatorJson(OperatorSpec opSpec, OperatorGraphJson opGraph) {
```
prateekm: Why getOrCreate? Should probably be just buildOperatorJson?

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -0,0 +1,244 @@
...
+    return opJson;
+  }
+
+  /**
+   * Get or create the JSON POJO for a {@link StreamEdge}
+   * @param edge {@link StreamEdge}
+   * @param streamEdges map of streamId to {@link org.apache.samza.execution.JobGraphJsonGenerator.StreamEdgeJson}
+   * @return {@link org.apache.samza.execution.JobGraphJsonGenerator.StreamEdgeJson}
+   */
+  private StreamEdgeJson getOrCreateStreamEdgeJson(StreamEdge edge, Map<String, StreamEdgeJson> streamEdges) {
```
prateekm: Same as previous, should probably be just buildStreamEdgeJson?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlan.java
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Returns the intermediate streams that need to be created.
+   * @return intermediate {@link StreamSpec}s
+   */
+  List<StreamSpec> getIntermediateStreams();
+
+  /**
+   * Returns the JSON representation of the plan for visualization
+   * @return json string
+   * @throws Exception exception
```
prateekm: Can remove "for visualization". E.g., could be used for logging.
"@return json string representation of the plan"
Document what exception.

PR #112, opened 2017-04-05
Samza-1187 : TestZkProcessorLatch tests share state causing transient failures in CI builds

navina:
* Removed testSingleCountdown - didn't understand what was being tested
* The timeout behavior for ZkProcessorLatch has not been documented. Hence, I have fixed testLatchExpires as best I could understand. 
* Removed redundant/unused member variables from ZkProcessorLatch

### DIFF
--- a/samza-core/src/test/java/org/apache/samza/zk/TestZkProcessorLatch.java
+++ b/samza-core/src/test/java/org/apache/samza/zk/TestZkProcessorLatch.java
@@ -71,28 +71,34 @@ public class TestZkProcessorLatch {
     zkServer.teardown();
   }
 
-  @Test
-  public void testLatchSizeOne() {
-    final int latchSize = 1;
-    final String latchId = "latchSizeOne";
-    final String participant1 = "participant1";
-
-    ExecutorService pool = Executors.newFixedThreadPool(3);
-    Future processor = pool.submit(
-      () -> {
-        ZkUtils zkUtils = getZkUtilsWithNewClient(participant1);
+  private Runnable getParticipantRunnable(int latchSize, String latchId, String participantId) {
+    return new Runnable() {
+      @Override
+      public void run() {
+        ZkUtils zkUtils = getZkUtilsWithNewClient(participantId);
         zkUtils.connect();
         ZkProcessorLatch latch = new ZkProcessorLatch(
-            latchSize, latchId, participant1, zkUtils);
+            latchSize, latchId, participantId, zkUtils);
         latch.countDown();
         try {
           latch.await(30, TimeUnit.SECONDS);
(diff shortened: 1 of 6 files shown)


navina:
@xinyuiscool & @nickpan47 : Can you guys please take a look at this PR? Thanks!

samza-core/src/test/java/org/apache/samza/zk/TestZkProcessorLatch.java
```diff
@@ -39,256 +28,193 @@
...
-
     try {
-      f1.get(300, TimeUnit.MILLISECONDS);
-      f2.get(300, TimeUnit.MILLISECONDS);
-      f3.get(300, TimeUnit.MILLISECONDS);
+      List<String> latchParticipants =
+          testZkUtils.getZkClient().getChildren(
+              String.format("%s/%s_%s", KEY_BUILDER.getRootPath(), ZkProcessorLatch.LATCH_PATH, latchId));
+      Assert.assertNotNull(latchParticipants);
+      Assert.assertEquals(1, latchParticipants.size());
```
nickpan47: I thought that the test is to test 1 latch with 2 participants? Why are we asserting 1 participant here?

samza-core/src/test/java/org/apache/samza/zk/TestZkProcessorLatch.java
```diff
@@ -39,256 +28,193 @@
...
     }
   }
 
   @Test
-  public void testLatchExpires() {
-    System.out.println("Started expiring");
-    String latchId = "l4";
-
-    int latchSize = 3;
+  public void testLatchSizeN() {
```
nickpan47: I am confused of the test names now. :/ testLatchSizeN vs testLatchSizeOneWithNParticipants, which one is more accurate here?

samza-core/src/test/java/org/apache/samza/zk/TestZkProcessorLatch.java
navina: @nickpan47 :) I didn't even know that a latch size can be different from the number of participating members. 
"testLatchSizeOneWithTwoParticipants" creates a latch of size 1. When there are 2 participants, one of them has to countdown. That is what is being tested. 
This is kind of weird because latch size should be indicative of how many participants we expect to join. Not sure how useful this behavior is.

samza-core/src/test/java/org/apache/samza/zk/TestZkProcessorLatch.java
navina: testLatchSizeNWithNParticipants may be more accurate here?

PR #113, opened 2017-04-06
SAMZA-1191: Fixed flaky test: TestExponentialSleepStrategy testThreadInterruptInRetryLoop

prateekm:
It's possible that the interruptee thread (see `#interruptedThread`) gets pre-empted before it has a chance to run the operation (increment `iterations`) and then gets interrupted, causing these assertions to fail.

I think these assertions also aren't critical for the tests which I presume want to test interrupt propagation behavior, so removing them in this change.

@vjagadish1989 & @xinyuiscool, please take a look.


vjagadish1989 (approved):
approved. thank you for the contribution

+1.

PR #114, opened 2017-04-06
SAMZA-1089: Enable YarnJob and ClientHelper to kill a job by name rather than YARN ApplicationID

jmakes:
Missed a couple files in the previous commit to enable YarnJob to kill and get status of a Job based on the job name rather than the YARN ApplicationName. 

These changes have been manually verified in a Yarn cluster at LI.

### DIFF
--- a/samza-yarn/src/main/scala/org/apache/samza/job/yarn/ClientHelper.scala
+++ b/samza-yarn/src/main/scala/org/apache/samza/job/yarn/ClientHelper.scala
@@ -220,6 +220,20 @@ class ClientHelper(conf: Configuration) extends Logging {
     appId
   }
 
+  def getActiveApplicationIds(appName: String): List[ApplicationId] = {
+    val getAppsRsp = yarnClient.getApplications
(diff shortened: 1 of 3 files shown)


jmakes:
@xinyuiscool you should have some familiarity with these changes. They're a direct copy from the prototype.

samza-yarn/src/main/scala/org/apache/samza/job/yarn/ClientHelper.scala
```diff
@@ -220,6 +220,20 @@ class ClientHelper(conf: Configuration) extends Logging {
     appId
   }
 
+  def getActiveApplicationIds(appName: String): List[ApplicationId] = {
```
xinyuiscool: Can we add javadocs for this?

jmakes: No, but I can add scaladoc :-)
👍

PR #116, opened 2017-04-06
SAMZA-1198: disable the flaky test TestZkBarrierForVersionUpgrade.testZkBarrierForVersionUpgrade

fredji97:
We are seeing the fails sometimes from this test, disabling it for build success first. See details in SAMZA-1198, and the JIRA for fix is in SAMZA-1193.

### DIFF
--- a/samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
+++ b/samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
@@ -34,6 +34,7 @@ import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.Before;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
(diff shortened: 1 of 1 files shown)


samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
```diff
@@ -74,6 +75,7 @@ public static void teardown() {
     zkServer.teardown();
   }
 
+  @Ignore("The test is flaky, see SAMZA-1193")
```
navina: Can you please comment out the @Test annotation instead of using Ignore annotation?
I believe last time @santhoshvenkat1988 tried with Ignore, it caused some failures in hudson. We decided that commenting out test annotation and adding a TODO with JIRA number was cleaner. You can talk to @santhoshvenkat1988 or @jmakes on why it failed on hudson.

PR #117, opened 2017-04-07
SAMZA-1132: LocalApplicationRunner for StreamApplication

xinyuiscool:
LocalApplicationRunner runs the StreamApplication locally on every node that the application is deployed to. LocalRunner.start() is blocking until the process is killed or user invoke LocalRunner.stop() from another thread. It can be configured to use Zk for coordination.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
@@ -113,14 +113,14 @@ public class SamzaContainerController {
     log.info("About to start container: " + containerModel.getContainerId());
     containerFuture = executorService.submit(() -> {
       try {
-        lifeCycleAwares.forEach(l -> l.onContainerStart());
+        lifeCycleAwares.forEach(l -> l.onStart());
 
         container.run();
 
-        lifeCycleAwares.forEach(l -> l.onContainerShutdown());
+        lifeCycleAwares.forEach(l -> l.onShutdown());
 
       } catch (Throwable t) {
-        lifeCycleAwares.forEach(l -> l.onContainerFailure(t));
+        lifeCycleAwares.forEach(l -> l.onFailure(t));
       }
     });
   }
--- a/samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
@@ -20,6 +20,7 @@
 package org.apache.samza.processor;
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import java.util.List;
 import org.apache.samza.config.ClusterManagerConfig;
 import org.apache.samza.config.Config;
 import org.apache.samza.config.TaskConfigJava;
@@ -49,6 +50,7 @@ public class SamzaContainerController {
   private final Map<String, MetricsReporter> metricsReporterMap;
   private final Object taskFactory;
   private final long containerShutdownMs;
+  private final List<StreamProcessorLifeCycleAware> lifeCycleAwares;
 
   // Internal Member Variables
   private Future containerFuture;
@@ -67,7 +69,8 @@ public class SamzaContainerController {
       Object taskFactory,
       long containerShutdownMs,
       String processorId,
-      Map<String, MetricsReporter> metricsReporterMap) {
+      Map<String, MetricsReporter> metricsReporterMap,
+      List<StreamProcessorLifeCycleAware> lifeCycleAwares) {
     this.executorService = Executors.newSingleThreadExecutor(new ThreadFactoryBuilder()
         .setNameFormat("p" + processorId + "-container-thread-%d").build());
     this.taskFactory = taskFactory;
@@ -77,6 +80,7 @@ public class SamzaContainerController {
     } else {
       this.containerShutdownMs = containerShutdownMs;
     }
+    this.lifeCycleAwares = lifeCycleAwares;
   }
 
   /**
@@ -107,7 +111,18 @@ public class SamzaContainerController {
         Util.<String, MetricsReporter>javaMapAsScalaMap(metricsReporterMap),
         taskFactory);
     log.info("About to start container: " + containerModel.getContainerId());
-    containerFuture = executorService.submit(() -> container.run());
+    containerFuture = executorService.submit(() -> {
+      try {
+        lifeCycleAwares.forEach(l -> l.onContainerStart());
+
+        container.run();
+
+        lifeCycleAwares.forEach(l -> l.onContainerShutdown());
+
+      } catch (Throwable t) {
+        lifeCycleAwares.forEach(l -> l.onContainerFailure(t));
+      }
+    });
   }
 
   /**
--- a/samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
@@ -20,7 +20,13 @@
 package org.apache.samza.processor;
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
-import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import org.apache.samza.config.ClusterManagerConfig;
 import org.apache.samza.config.Config;
 import org.apache.samza.config.TaskConfigJava;
@@ -34,14 +40,6 @@ import org.apache.samza.util.Util;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Map;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
-
 public class SamzaContainerController {
   private static final Logger log = LoggerFactory.getLogger(SamzaContainerController.class);
 
@@ -50,7 +48,8 @@ public class SamzaContainerController {
   private final Map<String, MetricsReporter> metricsReporterMap;
   private final Object taskFactory;
   private final long containerShutdownMs;
-  private final List<StreamProcessorLifeCycleAware> lifeCycleAwares;
+  private final String processorId;
+  private final StreamProcessorLifeCycleAware lifeCycleAware;
 
   // Internal Member Variables
   private Future containerFuture;
@@ -59,17 +58,20 @@ public class SamzaContainerController {
    * Creates an instance of a controller for instantiating, starting and/or stopping {@link SamzaContainer}
    * Requests to execute a container are submitted to the {@link ExecutorService}
    *
+   * @param processorId         {@link StreamProcessor} ID
    * @param taskFactory         Factory that be used create instances of {@link org.apache.samza.task.StreamTask} or
    *                            {@link org.apache.samza.task.AsyncStreamTask}
    * @param containerShutdownMs How long the Samza container should wait for an orderly shutdown of task instances
    * @param metricsReporterMap  Map of metric reporter name and {@link MetricsReporter} instance
-   * @param lifeCycleAwares list of {@link StreamProcessorLifeCycleAware}s
+   * @param lifeCycleAware {@link StreamProcessorLifeCycleAware}
    */
   public SamzaContainerController(
+      String processorId,
       Object taskFactory,
       long containerShutdownMs,
       Map<String, MetricsReporter> metricsReporterMap,
-      List<StreamProcessorLifeCycleAware> lifeCycleAwares) {
+      StreamProcessorLifeCycleAware lifeCycleAware) {
+    this.processorId = processorId;
     this.taskFactory = taskFactory;
     this.metricsReporterMap = metricsReporterMap;
     if (containerShutdownMs == -1) {
@@ -78,7 +80,7 @@ public class SamzaContainerController {
       this.containerShutdownMs = containerShutdownMs;
     }
     // life cycle callbacks when shutdown and failure happens
-    this.lifeCycleAwares = lifeCycleAwares;
+    this.lifeCycleAware = lifeCycleAware;
   }
 
   /**
@@ -114,9 +116,9 @@ public class SamzaContainerController {
     containerFuture = executorService.submit(() -> {
         try {
           container.run();
-          lifeCycleAwares.forEach(l -> l.onShutdown());
+          lifeCycleAware.onShutdown(processorId);
         } catch (Throwable t) {
-          lifeCycleAwares.forEach(l -> l.onFailure(t));
+          lifeCycleAware.onFailure(processorId, t);
         }
       });
   }
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -18,8 +18,7 @@
  */
 package org.apache.samza.processor;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.Map;
 import org.apache.samza.annotation.InterfaceStability;
 import org.apache.samza.config.ApplicationConfig;
 import org.apache.samza.config.Config;
@@ -29,11 +28,11 @@ import org.apache.samza.config.TaskConfigJava;
 import org.apache.samza.coordinator.JobCoordinator;
 import org.apache.samza.coordinator.JobCoordinatorFactory;
 import org.apache.samza.metrics.MetricsReporter;
+import org.apache.samza.runtime.ProcessorIdGenerator;
 import org.apache.samza.task.AsyncStreamTaskFactory;
 import org.apache.samza.task.StreamTaskFactory;
+import org.apache.samza.util.ClassLoaderHelper;
 import org.apache.samza.util.Util;
-
-import java.util.Map;
 
 /**
  * StreamProcessor can be embedded in any application or executed in a distributed environment (aka cluster) as an
@@ -61,7 +60,8 @@ import java.util.Map;
 @InterfaceStability.Evolving
 public class StreamProcessor {
   private final JobCoordinator jobCoordinator;
-  private final List<StreamProcessorLifeCycleAware> lifeCycleAwares = new ArrayList<>();
+  private final StreamProcessorLifeCycleAware lifeCycleAware;
+  private final String processorId;
 
   /**
    * Create an instance of StreamProcessor that encapsulates a JobCoordinator and Samza Container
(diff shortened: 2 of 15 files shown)


xinyuiscool:
I have tested with an integration tests (https://github.com/xinyuiscool/examples/blob/master/TestZkLocalApplicationRunner.java). Tried partionBy() and map() and then sendTo(). All works now.

I will add unit tests to this soon.

samza-core/src/main/java/org/apache/samza/zk/ZkLeaderElector.java
```diff
@@ -84,6 +84,7 @@ private String getHostName() {
 
   @Override
   public void tryBecomeLeader(LeaderElectorListener leaderElectorListener) {
+    zkUtils.makeSurePersistentPathsExists(new String[]{keyBuilder.getProcessorsPath()});
```
navina: Why this line added here?

This was previously done in the constructor. Line 71-73 are commented out. What is the reason for this change?

xinyuiscool: @sborya moved the commented-out logic to ZkControllerImpl:54. I guess he wanted to create all the zk nodes needed for ZkCoordinator together there. However, this doesn't work for coordinator service which has ZkLeaderElector but not ZkController. So I added this logic back here. Maybe we can clean up the creation of these paths by:
1) ZkLeaderElector ensures the processor paths exist
2) ZkBarrierForVersionUpdate ensures the job model version paths exist

This way they can be used independently. What do you think?

navina: >  I guess he wanted to create all the zk nodes needed for ZkCoordinator together there. However, this doesn't work for coordinator service which has ZkLeaderElector but not ZkController.

Yes. CoordinationService usage should be independent of components in the JobCoordinator abstraction. 

> . So I added this logic back here. 

I don't think that is the right solution. We should not pollute the leader election process with Zk node setup. 

> This way they can be used independently. What do you think?

Yep. That's sounds right.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -19,36 +19,201 @@
...
+  }
+
+  /**
+   * Create {@link StreamProcessor} based on {@link StreamApplication} and the config
+   * @param config config
+   * @param app {@link StreamApplication}
+   * @return {@link StreamProcessor]}
+   */
+  /* package private */ StreamProcessor createStreamProcessor(Config config, StreamApplication app) {
+    Object taskFactory = TaskFactoryUtil.createTaskFactory(config, app, this);
```
nickpan47: side note again. This makes me feel that it is still worth to use a common generic type for TaskFactory, at least for the TaskFactory object itself to be bound to a defined interface instead of an Object. Granted that we are going to load the TaskFactory class and lose the type info of the task object returned from createInstance(). But we at least know the type of the TaskFactory object and have a defined method createInstance() to be called without the need to type cast the factory object to StreamTaskFactory/AsyncStreamTaskFactory before we can even invoke the createInstance() method.
@navina and @prateekm, thoughts?

samza-core/src/main/java/org/apache/samza/zk/ZkLeaderElector.java
nickpan47: Agree with @xinyuiscool when we make the Zk coordination as a generic service, leader election and other job-coordinator specific functions should be separated. Let's open a JIRA to cleanup in another PR.

samza-core/src/test/java/org/apache/samza/runtime/TestLocalApplicationRunner.java
```diff
@@ -0,0 +1,212 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+
+public class TestLocalApplicationRunner {
+
+  @Test
+  public void testStreamCreation() throws Exception {
```
nickpan47: So, this test the case that there is no coordination service right? It would be good to add test cases including a mock coordination service.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -19,36 +19,168 @@
...
+      return null;
+    }
+  }
+
+  private StreamProcessor createStreamProcessor(Config config, Object taskFactory) {
+    if (taskFactory instanceof StreamTaskFactory) {
+      return new StreamProcessor(processorId, config, new HashMap<>(), (StreamTaskFactory) taskFactory);
+    } else if (taskFactory instanceof AsyncStreamTaskFactory) {
+      return new StreamProcessor(processorId, config, new HashMap<>(), (AsyncStreamTaskFactory) taskFactory);
+    } else {
```
jmakes: What's the point of the factory pattern if StreamProcessor only supports 2 implementations? Maybe this is a question for @navina 

Ideally this createStreamProcessor method wouldn't even exist. We'd just pass any task factory to the StreamProcessor.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -19,36 +19,201 @@
...
+   */
+  /* package private */ void createStreams(List<StreamSpec> intStreams) throws Exception {
+    if (!intStreams.isEmpty()) {
+      if (coordination != null) {
+        Latch initLatch = coordination.getLatch(1, LATCH_INIT);
+        coordination.getLeaderElector().tryBecomeLeader(() -> {
+          getStreamManager().createStreams(intStreams);
+          initLatch.countDown();
+        });
+        initLatch.await(LATCH_TIMEOUT, TimeUnit.MINUTES);
```
nickpan47: Just clarified with Navina that the latch from the Coordinator is shared among the leader and followers. Hence, discard the previous comment.

xinyuiscool: That's the beauty of this coordination service :), everything is hidden from the interface so the runner is much simpler to write. Thanks @sborya for providing it.

Seems @navina already  committed the processorId generation part. I am going to integrate with that and address all the comments here. Thanks.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -19,36 +19,201 @@
...
+      }
+      plan.getJobConfigs().forEach(jobConfig -> {
+        log.info("Starting job {} StreamProcessor with config {}", jobConfig.getName(), jobConfig);
+
+        StreamProcessor processor = createStreamProcessor(jobConfig, app);
+        processor.addLifeCycleAware(new LocalProcessorListener(processor));
+        processors.add(processor);
+        processor.start();
+      });
+      appStatus = ApplicationStatus.Running;
```
xinyuiscool: Makes sense. The error code will be there, since we pass in the life cycle hookups and the StreamProcessor should trigger it whenever there is a failure.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -19,36 +19,201 @@
...
+   * Create the Coordination needed by the application runner.
+   * @return {@link CoordinationUtils}
+   * @throws Exception exception
+   */
+  private CoordinationUtils getCoordinationUtils() throws Exception {
+    ApplicationConfig appConfig = new ApplicationConfig(config);
+    String clazz = appConfig.getCoordinationServiceFactoryClass();
+    if (clazz != null) {
+      CoordinationServiceFactory factory = ClassLoaderHelper.fromClassName(clazz);
+      String groupId = String.format("app-%s-%s", config.get(JobConfig.JOB_NAME()), config.get(JobConfig.JOB_ID(), "1"));
```
xinyuiscool: Sure, added those to the ApplicationConfig.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: I agree we should do something to improve this TaskFactory mess. This part of the code looks pretty ugly.

samza-core/src/main/java/org/apache/samza/zk/ZkLeaderElector.java
xinyuiscool: Created https://issues.apache.org/jira/browse/SAMZA-1201 to track this.

samza-core/src/test/java/org/apache/samza/runtime/TestLocalApplicationRunner.java
xinyuiscool: Good point! Added another test for it.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -59,6 +61,7 @@
 @InterfaceStability.Evolving
 public class StreamProcessor {
   private final JobCoordinator jobCoordinator;
+  private final List<StreamProcessorLifeCycleAware> lifeCycleAwares = new ArrayList<>();
```
navina: Why is this a list of StreamProcessorLifeCycleAware? 
Iirc, there is a 1:N mapping between an ApplicationRunner and StreamProcessor. Shouldn't this relationship be maintained on the ApplicationRunner layer, as opposed to StreamProcessor layer? 
It is extremely weird to see StreamProcessor taking in multiple listeners. If the relationship were inverted where there are N ApplicationRunners and we need to notify each of them from a single StreamProcessor, this makes sense. But that is not the case. 
Can you please explain the motivation behind this choice?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessorLifeCycleAware.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.samza.annotation.InterfaceStability;
+
+
+/**
+ * This class listens to the life cycle events in a {@link StreamProcessor},
+ * and triggers the corresponding callbacks.
+ */
+
+@InterfaceStability.Evolving
+public interface StreamProcessorLifeCycleAware {
```
navina: Is this a user-facing interface for lifecycle ? Or is it simply internal? 

Even though this class listens to the lifecycle events in StreamProcessor, the interaction seems to leak into SamzaContainerController class. This breaks the logical abstractions where StreamProcessor is responsible for the lifecycle of container and coordinator.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -19,36 +19,199 @@
...
+
+  final class LocalProcessorListener implements StreamProcessorLifeCycleAware {
+    StreamProcessor processor;
+
+    LocalProcessorListener(StreamProcessor processor) {
+      this.processor = processor;
+    }
+
+    @Override
+    public void onStart() {
```
navina: Shouldn't a processor be added to the set in onStart as opposed to when you are creating the StreamProcessors?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -19,36 +19,199 @@
...
+
+      // 3. start the StreamProcessors
+      if (plan.getJobConfigs().isEmpty()) {
+        throw new SamzaException("No jobs to run.");
+      }
+      plan.getJobConfigs().forEach(jobConfig -> {
+          log.info("Starting job {} StreamProcessor with config {}", jobConfig.getName(), jobConfig);
+
+          StreamProcessor processor = createStreamProcessor(jobConfig, app);
+          processor.addLifeCycleAware(new LocalProcessorListener(processor));
```
navina: Is the lifecycle aware listener an optional listener for the StreamProcessor? Assuming that ApplicationRunner is the user-facing api, it seems like a cleaner choice to make this a constructor parameter to the StreamProcessor, as opposed to adding it explicitly. 
There is nothing in the StreamProcessor Api that imposes the requirement that lifecycleaware listener must be added before processor.start. 
This, again , goes back to the question of why the StreamProcessor requires a "list" of lifecycleawarelisteners? 
Also, I don't think it is a good idea to pass in a reference to the StreamProcessor in the LocalProcessorListener. you may want to leverage the uniqueness of processorIds to countdown your latch.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessorLifeCycleAware.java
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Callback when the Samza processor is started
```
navina: For consistency, please use "StreamProcessor, as opposed to "Samza processor".

samza-core/src/main/java/org/apache/samza/processor/StreamProcessorLifeCycleAware.java
```diff
@@ -0,0 +1,48 @@
...
+   */
+  void onStart();
+
+  /**
+   * Callback when the Samza processor is shut down.
```
navina: For consistency, please use "StreamProcessor", as opposed to "Samza processor".

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -19,36 +19,199 @@
...
 
 /**
  * This class implements the {@link ApplicationRunner} that runs the applications in standalone environment
  */
 public class LocalApplicationRunner extends AbstractApplicationRunner {
 
-  public LocalApplicationRunner(Config config) {
+  private static final Logger log = LoggerFactory.getLogger(LocalApplicationRunner.class);
+  private static final String LATCH_INIT = "init";
+  private static final long LATCH_TIMEOUT = 10; // 10 min timeout
```
navina: Does this timeout need to configurable? Why 10 min?

Also, can you rename the variable to LATCH_TIMEOUT_MIN? It is not very obvious that you are using "Minutes" as the time unit.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -19,36 +19,199 @@
...
+      }
+      plan.getJobConfigs().forEach(jobConfig -> {
+          log.info("Starting job {} StreamProcessor with config {}", jobConfig.getName(), jobConfig);
+
+          StreamProcessor processor = createStreamProcessor(jobConfig, app);
+          processor.addLifeCycleAware(new LocalProcessorListener(processor));
+          processors.add(processor);
+          processor.start();
+        });
+      appStatus = ApplicationStatus.Running;
```
navina: Shouldn't application status be set to running only when all StreamProcessors have "started" ?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -19,36 +19,199 @@
...
+      throw new SamzaException("Failed to run application", t);
+    } finally {
+      if (coordination != null) {
+        coordination.reset();
+      }
+    }
   }
 
   @Override
   public void kill(StreamApplication streamApp) {
```
navina: QQ: Is kill a synchronous or asynchronous in LocalApplication Runner?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
navina: @nickpan47  I think we discussed offline whether it makes sense to change to use a generic type or a common marker interface. I don't recollect where we landed. Can we please document the conclusion here? Or in a separate JIRA, if there is follow-up expected in the short-term or long-term.

Fwiw, I am all for getting rid of a very vague "Object" in the interface.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
xinyuiscool: Makes sense. I added the life cycle listeners as a collection to make it more general, in case some other class other than ApplicationRunner wants to listen to it. Since it's causing confusion, I changed it to be a single aware object.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: This is to be consistent with Yarn case, where any job is running will mark it running.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: I assumed StreamProcessor.stop() is synchronous here. Is that the case?

xinyuiscool:
I consolidated the processorId generation part into StreamProcessor constructor, and then pass it into Coordinator and ContainerController. Now the logic should be much cleaner.

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
```diff
@@ -77,6 +80,7 @@ public SamzaContainerController(
     } else {
       this.containerShutdownMs = containerShutdownMs;
     }
+    this.lifeCycleAwares = lifeCycleAwares;
```
navina: > Shutdown means either user triggered shutdown or finite input streams. Failure happens when exception thrown from the container thread.
@xinyuiscool Can you please capture this in the documentation? 

A more general question: If this lifecycle callback is not user-facing, what kind of interactions does it expects from StreamProcessor? The fact that you are directly injecting the lifecycle handler by-passing the StreamProcessor interface is extremely confusing and blurs the contract between the StreamProcessor and ApplicationRunner. Can you please clarify the contract here?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
navina: StreamProcessor.stop is synchronous. Is kill also a synchronous interface then?

samza-core/src/main/java/org/apache/samza/processor/SamzaContainerController.java
xinyuiscool: For the application runner, it needs to know about the life cycle of StreamProcessors, i.e. start, shutdown and failure. So it expects the stream processor to trigger the life cycle callbacks. For example, in case of stream processor shutdown (either end of stream or use request shutdown), StreamProcessor needs to notify the application runner. So after all the stream processors finish or there is a failure, ApplicationRunner will shut down.

Btw, all these are internal life cycle methods. StreamProcessor is not exposed to the user, nor the life cycle management.

navina:
> I consolidated the processorId generation part into StreamProcessor constructor, and then pass it into Coordinator and ContainerController. Now the logic should be much cleaner.

I am currently working on refactoring the interaction between StreamProcessor/JobCoordinator/Container. I think the goal is for the StreamProcessor to listen to JobCoordinator events and accordingly, manage the container lifecycle. This way, we de-couple jobcoordinator from container lifecycle. While it makes sense for processorId to come from StreamProcessor to make it for LocalApplicationRunner, I think the change to JobCoordinatorFactory to include processorId is incorrect and should have been called out explicitly in the PR. 

In future, please document motivations for changing samza-core apis as well. It impacts other developers relying on the samza-core interface. 

In general, please seek for explicit approvals from committers who have explicitly given a -1 in the PR before committing the code. While it defies the general [commit policy](https://www.apache.org/foundation/voting.html#votes-on-code-modification) followed in Apache projects, more importantly, it impacts the productivity of other project developers, and ultimately, the progress of the project itself. Thanks!

xinyuiscool:
Apologies for committing it before +1 form everyone in the reviews. Chatted with @navina offline and this patch is the first attempt to define the contract between ApplicationRunner and StreamProcessor. @navina will have a much better implementation of the life cycle management of StreamProcessor and its components.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -19,36 +19,195 @@
...
+   */
+  /* package private */ void createStreams(List<StreamSpec> intStreams) throws Exception {
+    if (!intStreams.isEmpty()) {
+      if (coordination != null) {
+        Latch initLatch = coordination.getLatch(1, LATCH_INIT);
+        coordination.getLeaderElector().tryBecomeLeader(() -> {
+            getStreamManager().createStreams(intStreams);
+            initLatch.countDown();
+          });
+        initLatch.await(LATCH_TIMEOUT_MINUTES, TimeUnit.MINUTES);
```
prateekm: Seems like all non-leader LocalRunners are going to wait for 10 minutes on startup here. Am I misunderstanding something?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
prateekm: @navina Wouldn't StreamProcessor#stop be async since it gives back a onShutdown lifecycle callback?

samza-core/src/test/java/org/apache/samza/runtime/TestLocalApplicationRunner.java
```diff
@@ -0,0 +1,302 @@
...
+    plannerField.setAccessible(true);
+    plannerField.set(runner, planner);
+
+    StreamManager streamManager = mock(StreamManager.class);
+    Field streamManagerField = runner.getClass().getSuperclass().getDeclaredField("streamManager");
+    streamManagerField.setAccessible(true);
+    streamManagerField.set(runner, streamManager);
+    ArgumentCaptor<List> captor = ArgumentCaptor.forClass(List.class);
+
+    ExecutionPlan plan = new ExecutionPlan() {
```
prateekm: Prefer using mocks instead of lambdas for these. Makes tests more readable.

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
```diff
@@ -51,6 +52,20 @@
 import static org.apache.samza.test.processor.IdentityStreamTask.endLatch;
 
 public class TestStreamProcessor extends StandaloneIntegrationTestHarness {
+  private final StreamProcessorLifeCycleAware listener = new StreamProcessorLifeCycleAware() {
```
prateekm: Please use mocks instead of these.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -19,36 +19,195 @@
...
+  }
+
+  /**
+   * Create {@link StreamProcessor} based on {@link StreamApplication} and the config
+   * @param config config
+   * @param app {@link StreamApplication}
+   * @return {@link StreamProcessor]}
+   */
+  /* package private */ StreamProcessor createStreamProcessor(Config config, StreamApplication app, StreamProcessorLifeCycleAware listener) {
+    Object taskFactory = TaskFactoryUtil.createTaskFactory(config, app, this);
```
prateekm: Shouldn't this be a constructor in StreamProcessor instead of a helper here?
Btw, @navina, Is StreamProcessor going to be a framework internal class in future?

prateekm:
@xinyuiscool @nickpan47 Also, looks like LocalApplicationRunner is managing StreamProcessor directly. Is the plan to add the (no-op) LocalJobRunner layer later? If not, can we update SEP-2?

PR #122, opened 2017-04-13
SAMZA-1210: Fixing merge issue and container id generations.

sborya:
This PR is just for fixing issues introduced by a merge and changing container id generation.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
@@ -25,6 +25,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import org.apache.samza.SamzaException;
+import org.apache.samza.config.ApplicationConfig;
 import org.apache.samza.config.Config;
 import org.apache.samza.config.JavaSystemConfig;
 import org.apache.samza.config.JobConfig;
@@ -73,13 +74,12 @@ public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
 
     // TODO - this groupId may be passed from the caller as an application ID.
     JobConfig jobConfig = new JobConfig(config);
-    String groupId = jobConfig.getName() + "-" + jobConfig.getJobId();
 
     this.coordinationUtils = Util.
         <CoordinationServiceFactory>getObj(
             new JobCoordinatorConfig(config)
                 .getJobCoordinationServiceFactoryClassName())
-        .getCoordinationService(groupId, String.valueOf(processorId), config);
+        .getCoordinationService(new ApplicationConfig(config).getGlobalAppId(), String.valueOf(processorId), config);
 
     this.zkUtils = ((ZkCoordinationUtils) coordinationUtils).getZkUtils();
     this.keyBuilder = zkUtils.getKeyBuilder();
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
@@ -145,8 +145,8 @@ public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
 
   @Override
   public void onProcessorChange(List<String> processors) {
-    log.info("ZkJobCoordinator::onProcessorChange - Processors changed! List size=" + processors.size());
-    // if processorIds is empty - it means we are called from 'onBecomeLeader'
+    log.info("ZkJobCoordinator::onProcessorChange - list of processors changed! List size=" + processors.size());
+    // if list of processors is empty - it means we are called from 'onBecomeLeader'
     generateNewJobModel(processors);
   }
 
@@ -158,15 +158,19 @@ public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
     log.info("pid=" + processorId + "new JobModel available.Container stopped.");
     // get the new job model
     newJobModel = zkUtils.getJobModel(version);
-    log.info("pid=" + processorId + "new JobModel available. ver=" + version + "; jm = " + newJobModel);
 
     String currentPath = zkUtils.getEphemeralPath();
-    String zkProcessorPid = zkUtils.getProcessorsPID(currentPath);
+    String data =  zkUtils.getProcessorsData(currentPath);
+    ZkUtils.ProcessorData processorData = new ZkUtils.ProcessorData(data);
+    String processorId = processorData.getProcessorId();
+    String processorHost = processorData.getHost();
+
+    log.info("pid=" + processorId + "; host=" + processorHost + ": new JobModel available. ver=" + version + "; jm = " + newJobModel);
 
     // update ZK and wait for all the processors to get this new version
     ZkBarrierForVersionUpgrade barrier = (ZkBarrierForVersionUpgrade) coordinationUtils.getBarrier(
         JOB_MODEL_UPGRADE_BARRIER);
-    barrier.waitForBarrier(version, String.valueOf(zkProcessorPid), new Runnable() {
+    barrier.waitForBarrier(version, String.valueOf(processorId), new Runnable() {
       @Override
       public void run() {
         onNewJobModelConfirmed(version);
@@ -190,14 +194,14 @@ public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
    * Generate new JobModel when becoming a leader or the list of processor changed.
    */
   private void generateNewJobModel(List<String> processors) {
-    List<String> currentProcessorsPids;
+    List<String> currentProcessorsIds;
     if (processors.size() > 0) {
       // we should use this list
       // but it needs to be converted into PIDs, which is part of the data
-      currentProcessorsPids = zkUtils.getSortedActiveProcessorsPIDs(processors);
+      currentProcessorsIds = zkUtils.getSortedActiveProcessorsIDs(processors);
     } else {
       // get the current list of processors
-      currentProcessorsPids = zkUtils.getSortedActiveProcessorsPIDs();
+      currentProcessorsIds = zkUtils.getSortedActiveProcessorsIDs();
     }
 
     // get the current version
@@ -211,8 +215,8 @@ public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
     }
     log.info("pid=" + processorId + "generating new model. Version = " + nextJMVersion);
 
-    List<String> containerIds = new ArrayList<>(currentProcessorsPids.size());
-    for (String processorPid : currentProcessorsPids) {
+    List<String> containerIds = new ArrayList<>(currentProcessorsIds.size());
+    for (String processorPid : currentProcessorsIds) {
       containerIds.add(processorPid);
     }
     log.info("generate new job model: processorsIds: " + Arrays.toString(containerIds.toArray()));
@@ -229,7 +233,7 @@ public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
     // start the barrier for the job model update
     BarrierForVersionUpgrade barrier = coordinationUtils.getBarrier(
         JOB_MODEL_UPGRADE_BARRIER);
-    barrier.start(nextJMVersion, currentProcessorsPids);
+    barrier.start(nextJMVersion, currentProcessorsIds);
 
     // publish new JobModel version
     zkUtils.publishJobModelVersion(currentJMVersion, nextJMVersion);
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
@@ -159,18 +159,12 @@ public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
     // get the new job model
     newJobModel = zkUtils.getJobModel(version);
 
-    String currentPath = zkUtils.getEphemeralPath();
-    String data =  zkUtils.getProcessorsData(currentPath);
-    ProcessorData processorData = new ProcessorData(data);
-    String processorId = processorData.getProcessorId();
-    String processorHost = processorData.getHost();
-
-    log.info("pid=" + processorId + "; host=" + processorHost + ": new JobModel available. ver=" + version + "; jm = " + newJobModel);
+    log.info("pid=" + processorId + ": new JobModel available. ver=" + version + "; jm = " + newJobModel);
 
     // update ZK and wait for all the processors to get this new version
     ZkBarrierForVersionUpgrade barrier = (ZkBarrierForVersionUpgrade) coordinationUtils.getBarrier(
         JOB_MODEL_UPGRADE_BARRIER);
-    barrier.waitForBarrier(version, String.valueOf(processorId), new Runnable() {
+    barrier.waitForBarrier(version, processorId, new Runnable() {
       @Override
       public void run() {
         onNewJobModelConfirmed(version);
@@ -198,7 +192,7 @@ public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
     if (processors.size() > 0) {
       // we should use this list
       // but it needs to be converted into PIDs, which is part of the data
-      currentProcessorsIds = zkUtils.getSortedActiveProcessorsIDs(processors);
+      currentProcessorsIds = zkUtils.getActiveProcessorsIDs(processors);
     } else {
       // get the current list of processors
       currentProcessorsIds = zkUtils.getSortedActiveProcessorsIDs();
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
@@ -51,7 +51,7 @@ import java.util.Map;
  */
 public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
   private static final Logger log = LoggerFactory.getLogger(ZkJobCoordinator.class);
-  private static final String JOB_MODEL_VERSION_BARRIER = "JobModelVersion";
+  private static final String JOB_MODEL_VERSION_BARRIER = "jobModelUpgradeBarrier";
 
   private final ZkUtils zkUtils;
   private final String processorId;
@@ -68,10 +68,8 @@ public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
   private String newJobModelVersion;  // version published in ZK (by the leader)
   private JobModel jobModel;
 
-  public ZkJobCoordinator(String groupId, Config config, ScheduleAfterDebounceTime debounceTimer, ZkUtils zkUtils,
+  public ZkJobCoordinator(String groupId, Config config, ScheduleAfterDebounceTime debounceTimer,
                           SamzaContainerController containerController) {
-    this.zkUtils = zkUtils;
-    this.keyBuilder = zkUtils.getKeyBuilder();
     this.debounceTimer = debounceTimer;
     ApplicationConfig appConfig = new ApplicationConfig(config);
     if (appConfig.getProcessorId() != null) {    // TODO: This check to be removed after 0.13+
@@ -83,13 +81,16 @@ public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
       this.processorId = idGenerator.generateProcessorId(config);
     }
     this.containerController = containerController;
-    this.zkController = new ZkControllerImpl(processorId, zkUtils, debounceTimer, this);
     this.config = config;
     this.coordinationUtils = Util.
         <CoordinationServiceFactory>getObj(
             new JobCoordinatorConfig(config)
                 .getJobCoordinationServiceFactoryClassName())
         .getCoordinationService(groupId, String.valueOf(processorId), config);
+
+    this.zkUtils = ((ZkCoordinationUtils)coordinationUtils).getZkUtils();
+    this.keyBuilder = zkUtils.getKeyBuilder();
+    this.zkController = new ZkControllerImpl(processorId, zkUtils, debounceTimer, this);
 
     streamMetadataCache = getStreamMetadataCache();
   }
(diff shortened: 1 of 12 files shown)


samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -47,7 +48,7 @@
  */
 public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
   private static final Logger log = LoggerFactory.getLogger(ZkJobCoordinator.class);
-  private static final String JOB_MODEL_VERSION_BARRIER = "JobModelVersion";
+  private static final String JOB_MODEL_VERSION_BARRIER = "jobModelUpgradeBarrier";
```
navina: Rename the variable name so that it matches the barrier name

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -64,21 +65,25 @@
   private String newJobModelVersion;  // version published in ZK (by the leader)
   private JobModel jobModel;
 
-  public ZkJobCoordinator(String processorId, String groupId, Config config, ScheduleAfterDebounceTime debounceTimer, ZkUtils zkUtils,
+  public ZkJobCoordinator(String groupId, Config config, ScheduleAfterDebounceTime debounceTimer,
```
navina: Why is processorId removed from here?

The ZkJobCoordinator is passing the first constructor argument as processorId and you are mapping it to groupId here. Please correct it. 

Also, please clarify what groupId means through documentation

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -64,21 +65,25 @@
...
                           SamzaContainerController containerController) {
-    this.processorId = processorId;
-    this.zkUtils = zkUtils;
-    this.keyBuilder = zkUtils.getKeyBuilder();
     this.debounceTimer = debounceTimer;
     this.containerController = containerController;
-    this.zkController = new ZkControllerImpl(processorId, zkUtils, debounceTimer, this);
     this.config = config;
+    this.processorId = new ApplicationConfig(config).getProcessorId();
```
navina: You shouldn't need this if you are using ZkJobCoordinatorFactory as it is defined.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -64,21 +65,25 @@
   private String newJobModelVersion;  // version published in ZK (by the leader)
```
navina: What is the purpose of this variable? Last time I checked, it was not being used anywhere.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -136,6 +137,35 @@ public synchronized String getEphemeralPath() {
     return children;
   }
 
+  /**
+   * Method is used to get the <i>sorted</i> list of currently active/registered processors PIDs
+   *
+   * @return List of absolute ZK node paths
+   */
+  public List<String> getSortedActiveProcessorsPIDs() {
```
navina: can you please add some test for this new method in ZkUtils?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: deleted.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: Well, because the processorId was not there when I merged it previously. Now it is back. So I will change it back to use processorId.
GroupId - is probably the future applicationId, so for now I will generate it right here and later will pass it from the ApplicationRunner when it is available.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: see above.
Will remove it from here.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
navina: > GroupId - is probably the future applicationId,

We should figure out what the semantics of groupId are and how it should be wired in from the user-facing API. 

> I will generate it right here and later will pass it from the ApplicationRunner when it is available.

Please don't generate it here if it is expected to be provided by some other component. ApplicationRunner is available now.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -188,6 +193,7 @@ public void onNewJobModelConfirmed(String version) {
   private void generateNewJobModel() {
     // get the current list of processors
     List<String> currentProcessors = zkUtils.getSortedActiveProcessors();
```
sborya: removing it will requires some more changes, but I guess it will be cleaner. I will add it.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -188,6 +193,7 @@ public void onNewJobModelConfirmed(String version) {
...
+    List<String> currentProcessorsPids = zkUtils.getSortedActiveProcessorsPIDs();
```
sborya: yeah, same as getSortedActiveProcessors() today. 
I am not sure it can be handled here. It requires handling at 'debounce' scheduler level.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
navina: Hmm..ok.. Yeah I think we can handle it in SAMZA-1150. But would it more clear to explicitly handle exception here (say, log it) and then, throw it? I am not sure what is the right approach here.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -61,23 +63,27 @@
...
-    this.processorId = processorId;
-    this.zkUtils = zkUtils;
-    this.keyBuilder = zkUtils.getKeyBuilder();
     this.debounceTimer = debounceTimer;
     this.containerController = containerController;
-    this.zkController = new ZkControllerImpl(processorId, zkUtils, debounceTimer, this);
     this.config = config;
+    this.processorId = processorId;
+
+    // TODO - this groupId may be passed from the caller as an application ID.
```
navina: Remove comment

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -61,23 +63,27 @@
...
+    JobConfig jobConfig = new JobConfig(config);
```
navina: Unused variable

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -142,13 +148,15 @@ public void onBecomeLeader() {
 
   @Override
   public void onProcessorChange(List<String> processorIds) {
-    log.info("ZkJobCoordinator::onProcessorChange - Processors changed! List: " + Arrays.toString(processorIds.toArray()));
+    log.info("ZkJobCoordinator::onProcessorChange - Processors changed! List size=" + processorIds.size());
+    // if processorIds is empty - it means we are called from 'onBecomeLeader'
+    //We currently ignore the list: " + Arrays.toString(processorIds.toArray()
```
navina: Disregarding the list returned from ZK event notification and fetching it again can cause inconsistency of the state in the job coordinator's view. The correct approach would be to pass this list into generateNewJobModel and fetch data only for this list. That way, the view provided by Zk and acted upon by the JC (the event handler) will be the same.

For the initial case where generateNewJobModel is called by the newly elected leader, this list can be empty and the leader can poll for the current list. This scenario can be differentiated using a flag argument in generateNewJobModel method that will indicate whether to ignore the provided list or not.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -217,8 +225,9 @@ private void generateNewJobModel() {
     log.info("pid=" + processorId + "published new JobModel ver=" + nextJMVersion + ";jm=" + jobModel);
 
     // start the barrier for the job model update
-    ZkBarrierForVersionUpgrade barrier = (ZkBarrierForVersionUpgrade) coordinationUtils.getBarrier(JOB_MODEL_VERSION_BARRIER);
-    barrier.start(nextJMVersion, currentProcessors);
+    ZkBarrierForVersionUpgrade barrier = (ZkBarrierForVersionUpgrade) coordinationUtils.getBarrier(
```
navina: QQ: Why do we need an explicit casting here ? Is the interface for barrier not the same as that returned by  CoordinationUtils?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -136,6 +137,44 @@ public synchronized String getEphemeralPath() {
     return children;
   }
 
+  /**
+   * Method is used to read processors PIDs
```
navina: It is not very obvious if you are referring to the OS's Process ID as PID or the Processor's ID as PID. Please change "PID" to explicitly read "processorId" and the id assigned by zk can be referred as znode id. It will be useful to use the same terminology across the codebase.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -136,6 +137,44 @@ public synchronized String getEphemeralPath() {
...
+    if (data == null) {
+      throw new SamzaException(String.format("Cannot read ZK node:", fullPath));
+    }
+    // data format is "host pid"
+    String [] pidHost = data.split(" ");
+    return pidHost[1];
+  }
+
+    /**
+     * Method is used to get the <i>sorted</i> list of currently active/registered processors PIDs
```
navina: Same comment as above regarding documentation

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -136,6 +137,44 @@ public synchronized String getEphemeralPath() {
...
+     *
+     * @return List of absolute ZK node paths
+     */
+  public List<String> getSortedActiveProcessorsPIDs() {
+    String processorPath = keyBuilder.getProcessorsPath();
+    List<String> children = zkClient.getChildren(processorPath);
+    List<String> childrenPids = new ArrayList<>(children.size());
```
navina: same comment as above. Keeping the variable name to be the same as the terms used in documentation makes the code more readable. ChildrenPids should simply be processorIds, right ?

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
```diff
@@ -50,6 +51,7 @@
   public static void setup() throws InterruptedException {
     zkServer = new EmbeddedZookeeper();
     zkServer.setup();
+    System.out.println("zk port=" + zkServer.getPort());
```
navina: nit: Remove sop from tests

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
```diff
@@ -89,11 +91,11 @@ public static void teardown() {
 
   @Test
   public void testRegisterProcessorId() {
-    String assignedPath = zkUtils.registerProcessorAndGetId("0.0.0.0");
+    String assignedPath = zkUtils.registerProcessorAndGetId("0.0.0.0 1");
```
navina: Looks like there is some specific data format expected by this method - registerProcessorAndGetId. maybe it will be cleaner to change the API to take two arguments, as in registerProcessorAndGetId(String hostname, String processorId). Serialization format can be completely hidden from the caller. Same thing applies when reading the data from the ZK. Deserialization should be hidden from the caller if ZkUtils. 

As a side question: Is this "hostname" used anywhere?

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
```diff
@@ -103,6 +105,20 @@ public void testGetActiveProcessors() {
     zkUtils.registerProcessorAndGetId("processorData");
     Assert.assertEquals(1, zkUtils.getSortedActiveProcessors().size());
   }
+
+  @Test
+  public void testGetProcessorsPIDs() {
```
navina: nit: same comment as above. Don't use PIDs unless it is the OS's process Id

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: done

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: done

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: fixed.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: we don't needed. Removed.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: ok.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: yes.

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
sborya: done.

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
sborya: i guess we can do that.

Not sure who is using hostname, it was added by someone else.

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
sborya: done.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -94,6 +95,36 @@ public ZkKeyBuilder getKeyBuilder() {
     return keyBuilder;
   }
 
+  public static class ProcessorData {
```
navina: I don't think this static class is necessarily. All you are trying to do here is parse the string provided by the caller to registerProcessorAndGetId. Based on the current code usage, you are using the constructor more as a util method. I don't see a need for a static class here.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -123,17 +154,60 @@ public synchronized String getEphemeralPath() {
...
+    List<String> znodeIds = zkClient.getChildren(keyBuilder.getProcessorsPath());
+    if (znodeIds.size() > 0) {
+      Collections.sort(znodeIds);
+      LOG.info("Found these children - " + znodeIds);
+    }
+    return znodeIds;
+  }
+
+  /**
+   * Method is used to read processors PIDs
```
navina: this still reads "PIDs" . Please change it to "processor's data in Znode".  

Method name reads "getProcessorsPID". It is plural. But it returns only one processorId. The method is returning data for a znode path. It will be more clear to call the method name getProcessorData.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -123,17 +154,60 @@ public synchronized String getEphemeralPath() {
...
+      throw new SamzaException(String.format("Cannot read ZK node:", fullPath));
+    }
+    return new ProcessorData(data).getProcessorId();
+  }
+
+  /**
+   * Method is used to get the <i>sorted</i> list of currently active/registered processor ids
+   * @return List of processorIds
+   */
+  public List<String> getSortedActiveProcessorsPIDs() {
```
navina: Please fix the "PIDs" in the method name.

Why is this overload method needed?? When should the caller use one Vs the other? Please clarify it in the docs or change this pattern.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -141,14 +144,14 @@ public void onBecomeLeader() {
   }
 
   @Override
-  public void onProcessorChange(List<String> processorIds) {
-    log.info("ZkJobCoordinator::onProcessorChange - Processors changed! List: " + Arrays.toString(processorIds.toArray()));
-    generateNewJobModel();
+  public void onProcessorChange(List<String> processors) {
+    log.info("ZkJobCoordinator::onProcessorChange - Processors changed! List size=" + processors.size());
+    // if processorIds is empty - it means we are called from 'onBecomeLeader'
```
navina: Comment doesn't match the argument name.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -185,9 +189,16 @@ public void onNewJobModelConfirmed(String version) {
...
    * Generate new JobModel when becoming a leader or the list of processor changed.
    */
-  private void generateNewJobModel() {
-    // get the current list of processors
-    List<String> currentProcessors = zkUtils.getSortedActiveProcessors();
+  private void generateNewJobModel(List<String> processors) {
+    List<String> currentProcessorsPids;
+    if (processors.size() > 0) {
+      // we should use this list
+      // but it needs to be converted into PIDs, which is part of the data
```
navina: Please refrain from sing PIDs. It affects the readability of your code.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -123,17 +154,60 @@ public synchronized String getEphemeralPath() {
...
+    return getSortedActiveProcessorsPIDs(getSortedActiveProcessors());
+  }
+
+  /**
+   * Method is used to get the <i>sorted</i> list of currently active/registered processors ids
+   * @param znodeIds - list of relative paths of the children's znodes
+   * @return List of processor ids
+   */
+  public List<String> getSortedActiveProcessorsPIDs(List<String> znodeIds) {
```
navina: Please fix PIDs in the method name

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorFactory.java
```diff
@@ -31,27 +28,19 @@
   /**
    * Method to instantiate an implementation of JobCoordinator
    *
+   * @param processorId -
```
navina: complete documentation here

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: Well, that what it was before. But now, I actually think your previous comment about abstracting the data format is a good idea event if at this point it may be just a simple string.

And now, come to think of it, hostname will be very usefully when debugging a really distributed app. So I am going to leave it there, and add it to some logging.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: yeah, I thought I hunted all of them. Will change the name and the comment.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: Will clarify in the comment.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: ah, one more down.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: fixed.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorFactory.java
sborya: done.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -123,17 +123,60 @@ public synchronized String getEphemeralPath() {
...
+    String processorPath = keyBuilder.getProcessorsPath();
+    List<String> processorIds = new ArrayList<>(znodeIds.size());
+    if (znodeIds.size() > 0) {
+
+      for (String child : znodeIds) {
+        String fullPath = String.format("%s/%s", processorPath, child);
+        processorIds.add(getProcessorsData(fullPath));
+      }
+
+      Collections.sort(processorIds);
```
navina: I don't think you should sort this list again. The input is a list of sorted znodeIds. "processorIds" is a list of data stored in the znode. Sorting here will re-order everything. 

Please change the name of the method appropriately.

PR #124, opened 2017-04-14
SAMZA-1209: Improve error handling in LocalStoreMonitor

shanthoosh:
Changes:

1. Add opt-in configuration to continue garbage collection of local stores
   when there’s a failure in garbage collecting one local store.
2. Handle failures gracefully. In getJobModel, JobModel is expected as return response.
   Incase of failures, returned error-message Map is deserialized to JobModel
   resulting in ClassCastException. Handle 2xx, failures separately. Log error
   messages returned properly in httpGet call.
3. Fix getTasks url format.
4. Minor code cleanup(Switch to using ',' as seperator for job.status.servers configuration
   instead of '.' for ease of use).
5. Fix disabled tests in TestLocalStoreMonitor.

### DIFF
--- a/samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
+++ b/samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
@@ -20,10 +20,11 @@ package org.apache.samza.monitor;
 
 import com.google.common.base.Preconditions;
 import java.io.IOException;
-import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Function;
+
+import org.apache.commons.codec.binary.StringUtils;
 import org.apache.commons.httpclient.HttpClient;
 import org.apache.commons.httpclient.methods.GetMethod;
 import org.apache.samza.SamzaException;
@@ -68,7 +69,7 @@ public class JobsClient {
    */
   public List<Task> getTasks(JobInstance jobInstance) {
     return retriableHttpGet(baseUrl -> String.format(ResourceConstants.GET_TASKS_URL, baseUrl,
-        jobInstance.getJobName(), jobInstance.getJobId()));
+        jobInstance.getJobName(), jobInstance.getJobId()), new TypeReference<List<Task>>(){});
   }
 
   /**
@@ -79,7 +80,7 @@ public class JobsClient {
    */
   public JobStatus getJobStatus(JobInstance jobInstance) {
     Job job = retriableHttpGet(baseUrl -> String.format(ResourceConstants.GET_JOBS_URL, baseUrl,
-        jobInstance.getJobName(), jobInstance.getJobId()));
+        jobInstance.getJobName(), jobInstance.getJobId()), new TypeReference<Job>(){});
     return job.getStatus();
   }
 
@@ -90,41 +91,45 @@ public class JobsClient {
    * When a job status server is down or returns a error response, it tries to reach out to
    * the next job status server in the sequence, to complete the http get request.
    *
-   * @param urlMapFunction to build the request url, given job status server base url.
+   * @param requestUrlBuilder to build the request url, given job status server base url.
    * @param <T> return type of the http get response.
    * @return the response from any one of the job status server.
    * @throws Exception when all the job status servers are unavailable.
    *
    */
-  private <T> T retriableHttpGet(Function<String, String> urlMapFunction) {
+  private <T> T retriableHttpGet(Function<String, String> requestUrlBuilder, TypeReference<T> typeReference) {
     Exception fetchException = null;
     for (String jobStatusServer : jobStatusServers) {
-      String requestUrl = urlMapFunction.apply(jobStatusServer);
+      String requestUrl = requestUrlBuilder.apply(jobStatusServer);
       try {
         ObjectMapper objectMapper = new ObjectMapper();
-        return objectMapper.readValue(httpGet(requestUrl), new TypeReference<T>() {});
+        return objectMapper.readValue(httpGet(requestUrl), typeReference);
       } catch (Exception e) {
         LOG.error(String.format("Exception when fetching tasks from the url : %s", requestUrl), e);
         fetchException = e;
       }
     }
-    throw new SamzaException(String.format("Exception during http get from urls : %s", jobStatusServers),
-        fetchException);
+    throw new SamzaException(String.format("Exception during http get from urls : %s", jobStatusServers), fetchException);
   }
 
   /**
    * This method initiates http get request on the request url and returns the
    * response returned from the http get.
    * @param requestUrl url on which the http get request has to be performed.
-   * @return the input stream of the http get response.
+   * @return the http get response.
    * @throws IOException if there are problems with the http get request.
    */
-  private InputStream httpGet(String requestUrl) throws IOException {
+  private byte[] httpGet(String requestUrl) throws IOException {
     GetMethod getMethod = new GetMethod(requestUrl);
     try {
       int responseCode = httpClient.executeMethod(getMethod);
       LOG.debug("Received response code {} for the get request on the url : {}", responseCode, requestUrl);
-      return getMethod.getResponseBodyAsStream();
+      byte[] response = getMethod.getResponseBody();
+      if (responseCode != 200) {
+        throw new SamzaException(String.format("Received response code: %s for get request on: %s, with message: %s.",
+                                               responseCode, requestUrl, StringUtils.newStringUtf8(response)));
+      }
+      return response;
     } finally {
       getMethod.releaseConnection();
     }
--- a/samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
+++ b/samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
@@ -26,6 +26,7 @@ import java.util.function.Function;
 
 import org.apache.commons.codec.binary.StringUtils;
 import org.apache.commons.httpclient.HttpClient;
+import org.apache.commons.httpclient.HttpStatus;
 import org.apache.commons.httpclient.methods.GetMethod;
 import org.apache.samza.SamzaException;
 import org.apache.samza.rest.model.Job;
@@ -68,7 +69,7 @@ public class JobsClient {
    * @throws SamzaException if there were any problems with the http request.
    */
   public List<Task> getTasks(JobInstance jobInstance) {
-    return retriableHttpGet(baseUrl -> String.format(ResourceConstants.GET_TASKS_URL, baseUrl,
+    return queryJobStatusServers(baseUrl -> String.format(ResourceConstants.GET_TASKS_URL, baseUrl,
         jobInstance.getJobName(), jobInstance.getJobId()), new TypeReference<List<Task>>(){});
   }
 
@@ -79,7 +80,7 @@ public class JobsClient {
    * @throws SamzaException if there are any problems with the http request.
    */
   public JobStatus getJobStatus(JobInstance jobInstance) {
-    Job job = retriableHttpGet(baseUrl -> String.format(ResourceConstants.GET_JOBS_URL, baseUrl,
+    Job job = queryJobStatusServers(baseUrl -> String.format(ResourceConstants.GET_JOBS_URL, baseUrl,
         jobInstance.getJobName(), jobInstance.getJobId()), new TypeReference<Job>(){});
     return job.getStatus();
   }
@@ -97,19 +98,21 @@ public class JobsClient {
    * @throws Exception when all the job status servers are unavailable.
    *
    */
-  private <T> T retriableHttpGet(Function<String, String> requestUrlBuilder, TypeReference<T> typeReference) {
-    Exception fetchException = null;
+  private <T> T queryJobStatusServers(Function<String, String> requestUrlBuilder, TypeReference<T> typeReference) {
+    SamzaException fetchException = null;
     for (String jobStatusServer : jobStatusServers) {
       String requestUrl = requestUrlBuilder.apply(jobStatusServer);
       try {
         ObjectMapper objectMapper = new ObjectMapper();
-        return objectMapper.readValue(httpGet(requestUrl), typeReference);
+        byte[] response = httpGet(requestUrl);
+        return objectMapper.readValue(response, typeReference);
       } catch (Exception e) {
-        LOG.error(String.format("Exception when fetching tasks from the url : %s", requestUrl), e);
-        fetchException = e;
+        String exceptionMessage = String.format("Exception in http get request from url: %s.", requestUrl);
+        LOG.error(exceptionMessage, e);
+        fetchException = new SamzaException(exceptionMessage, e);
       }
     }
-    throw new SamzaException(String.format("Exception during http get from urls : %s", jobStatusServers), fetchException);
+    throw fetchException;
   }
 
   /**
@@ -123,9 +126,9 @@ public class JobsClient {
     GetMethod getMethod = new GetMethod(requestUrl);
     try {
       int responseCode = httpClient.executeMethod(getMethod);
-      LOG.debug("Received response code {} for the get request on the url : {}", responseCode, requestUrl);
+      LOG.debug("Received response code: {} for the get request on the url: {}", responseCode, requestUrl);
       byte[] response = getMethod.getResponseBody();
-      if (responseCode != 200) {
+      if (responseCode != HttpStatus.SC_OK) {
         throw new SamzaException(String.format("Received response code: %s for get request on: %s, with message: %s.",
                                                responseCode, requestUrl, StringUtils.newStringUtf8(response)));
       }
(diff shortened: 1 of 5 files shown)


samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
```diff
@@ -79,7 +80,7 @@ public JobsClient(List<String> jobStatusServers) {
    */
   public JobStatus getJobStatus(JobInstance jobInstance) {
     Job job = retriableHttpGet(baseUrl -> String.format(ResourceConstants.GET_JOBS_URL, baseUrl,
-        jobInstance.getJobName(), jobInstance.getJobId()));
+        jobInstance.getJobName(), jobInstance.getJobId()), new TypeReference<Job>(){});
```
prateekm: Why do you need a TypeReference? Why does retriableHttpGet need to be generic? It's a private method, used once, and its logic and logging makes it specific to fetching tasks.

samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
```diff
@@ -90,41 +91,45 @@ public JobStatus getJobStatus(JobInstance jobInstance) {
...
    *
-   * @param urlMapFunction to build the request url, given job status server base url.
+   * @param requestUrlBuilder to build the request url, given job status server base url.
    * @param <T> return type of the http get response.
    * @return the response from any one of the job status server.
    * @throws Exception when all the job status servers are unavailable.
    *
    */
-  private <T> T retriableHttpGet(Function<String, String> urlMapFunction) {
+  private <T> T retriableHttpGet(Function<String, String> requestUrlBuilder, TypeReference<T> typeReference) {
```
prateekm: Not retriable either. I'd actually prefer moving the logic for iterating over the jobStatusServers and error handling to the caller method.

samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
```diff
@@ -90,41 +91,45 @@ public JobStatus getJobStatus(JobInstance jobInstance) {
...
     Exception fetchException = null;
     for (String jobStatusServer : jobStatusServers) {
-      String requestUrl = urlMapFunction.apply(jobStatusServer);
+      String requestUrl = requestUrlBuilder.apply(jobStatusServer);
       try {
         ObjectMapper objectMapper = new ObjectMapper();
-        return objectMapper.readValue(httpGet(requestUrl), new TypeReference<T>() {});
+        return objectMapper.readValue(httpGet(requestUrl), typeReference);
```
prateekm: Once again, please don't inline method calls with side effects as parameters.

samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
```diff
@@ -90,41 +91,45 @@ public JobStatus getJobStatus(JobInstance jobInstance) {
...
       } catch (Exception e) {
         LOG.error(String.format("Exception when fetching tasks from the url : %s", requestUrl), e);
         fetchException = e;
       }
     }
-    throw new SamzaException(String.format("Exception during http get from urls : %s", jobStatusServers),
-        fetchException);
+    throw new SamzaException(String.format("Exception during http get from urls : %s", jobStatusServers), fetchException);
```
prateekm: Message is misleading. Implies the same exception from each server. Either keep and log a Map<url, exception> or make message clearer.

samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
```diff
@@ -90,41 +91,45 @@ public JobStatus getJobStatus(JobInstance jobInstance) {
...
    */
-  private InputStream httpGet(String requestUrl) throws IOException {
+  private byte[] httpGet(String requestUrl) throws IOException {
     GetMethod getMethod = new GetMethod(requestUrl);
     try {
       int responseCode = httpClient.executeMethod(getMethod);
       LOG.debug("Received response code {} for the get request on the url : {}", responseCode, requestUrl);
-      return getMethod.getResponseBodyAsStream();
+      byte[] response = getMethod.getResponseBody();
+      if (responseCode != 200) {
```
prateekm: Use constant for HTTP OK.

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitor.java
```diff
@@ -81,25 +81,33 @@ public void monitor() throws Exception {
...
-              && task.getPreferredHost().equals(localHostName)) {
-            LOG.info(String.format("Store %s is actively used by the task: %s.", storeName, task.getTaskName()));
-          } else {
-            LOG.info(String.format("Store %s not used by the task: %s.", storeName, task.getTaskName()));
-            markSweepTaskStore(TaskStorageManager.getStorePartitionDir(jobDir, storeName, new TaskName(task.getTaskName())));
+      try {
+        JobStatus jobStatus = jobsClient.getJobStatus(jobInstance);
+        for (Task task : jobsClient.getTasks(jobInstance)) {
+          for (String storeName : jobDir.list(DirectoryFileFilter.DIRECTORY)) {
+            LOG.info("Job: {} has the running status: {} with preferred host:  {}", jobInstance, jobStatus, task.getPreferredHost());
```
prateekm: Nitpitck: Single space before last `{}`

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitor.java
```diff
@@ -81,25 +81,33 @@ public void monitor() throws Exception {
...
+                && task.getPreferredHost().equals(localHostName)) {
+              LOG.info(String.format("Store %s is actively used by the task: %s.", storeName, task.getTaskName()));
+            } else {
+              LOG.info(String.format("Store %s not used by the task: %s.", storeName, task.getTaskName()));
+              markSweepTaskStore(TaskStorageManager.getStorePartitionDir(jobDir, storeName, new TaskName(task.getTaskName())));
+            }
           }
         }
+      } catch (Exception ex) {
+        if (!config.getIgnoreFailures()) {
```
prateekm: Don't use 'get' prefix for boolean config getters. Use isXYZ or just ignoreFailures.

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitor.java
```diff
@@ -81,25 +81,33 @@ public void monitor() throws Exception {
...
+          throw ex;
+        }
+        LOG.warn("Local store cleanup for job: {} resulted in exception: {}. Config: {} turned on, failures are ignored.",
```
prateekm: What's the scenario when this would and would not be turned on?

Rephrase so that the exception is the last part of the message. Otherwise the stacktrace will split the relevant message across multiple lines.

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitorConfig.java
```diff
@@ -53,6 +53,12 @@
    */
   private static final long DEFAULT_OFFSET_FILE_TTL_MS = 1000 * 60 * 60 * 24 * 7;
 
+  /**
+   * Boolean variable if set to true will make LocalStoreMonitor to ignore failures
```
prateekm: Don't need 'Boolean variable'. Esp. since the actual type is string.

'will make LocalStoreMonitor ignore'

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitorConfig.java
```diff
@@ -80,6 +86,15 @@ public long getOffsetFileTTL() {
    *         on the job status server.
    */
   public List<String> getJobStatusServers() {
-     return Arrays.asList(StringUtils.split(get(CONFIG_JOB_STATUS_SERVERS), '.'));
+     return Arrays.asList(StringUtils.split(get(CONFIG_JOB_STATUS_SERVERS), ','));
+  }
+
+  /**
+   *
```
prateekm: Once again, please add description for javadocs.

samza-rest/src/main/java/org/apache/samza/rest/resources/ResourceConstants.java
```diff
@@ -20,7 +20,7 @@
 
 public class ResourceConstants {
 
-  public static final String GET_TASKS_URL = "http://%s/%s/%s/tasks/";
+  public static final String GET_TASKS_URL = "http://%s/v1/jobs/%s/%s/tasks/";
```
prateekm: How was this working earlier?

samza-rest/src/test/java/org/apache/samza/monitor/TestLocalStoreMonitor.java
```diff
@@ -78,18 +82,22 @@ public void setUp() throws Exception {
...
   @After
-  public void cleanUp() throws Exception {
+  public void cleanUp()  {
     // clean up the temp files created
-    FileUtils.deleteDirectory(taskStoreDir);
+    try {
+      FileUtils.deleteDirectory(taskStoreDir);
+    } catch (IOException e) {
+      // Happens when task store can't be deleted after test finishes.
+      Assert.fail();
```
prateekm: Why are you failing an assert in cleanUp? What do you want to do here?

samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
```diff
@@ -90,41 +91,45 @@ public JobStatus getJobStatus(JobInstance jobInstance) {
...
+        throw new SamzaException(String.format("Received response code: %s for get request on: %s, with message: %s.",
+                                               responseCode, requestUrl, StringUtils.newStringUtf8(response)));
```
prateekm: Why not just use the String constructor?

samza-rest/src/test/java/org/apache/samza/monitor/TestLocalStoreMonitor.java
shanthoosh: Exception in cleaning up the directory here can cause failures in actual test cases(since the directory will be counted in tests). 

Previously this was swallowed and hence we might not know if this delete failure has actually caused any assertion failure. 

If tests fails due to lack of permissions in cleaning up scaffold test directories, we will know that.

samza-rest/src/main/java/org/apache/samza/rest/resources/ResourceConstants.java
shanthoosh: Behavior was tested and verified initially within a LI cluster. 

However, there were problems with merging the independent patches  where in the  url format of endpoint got changed and the actual client wasn't updated properly.

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitorConfig.java
shanthoosh: Added. 

> please add description for javadocs.

Just a curious question. Wondering does it makes sense to have java doc for these config methods. They're just wrapper on top of get methods and has default value if undefined(doesn't have complicated logic that need explanation, code is plain obvious) and the name is self explanatory. Just thinking, what's the value addition by these docs/explanation? Please share your thoughts.

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitorConfig.java
shanthoosh: Fixed.

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitor.java
shanthoosh: > What's the scenario when this would and would not be turned on?

When failure in either getting status/task details of a job or  store cleanup of one job should not affect the other job store cleanup in NM host. Swallowing exceptions within code isn't usual modus operandi, hence felt the need that the client should control to say either stop at first failure or continue even when there're failures to next job(Making it explicit. This was problem within linkedin, since some jobs were installed in primary RM and not in standby RM, there by causing problems. In some cases, jobs were not installed in both the RM's). 

> Rephrase so that the exception is the last part of the message. 

Makes sense, Fixed it. Please check and see if it's alright.

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitor.java
shanthoosh: I agree with you (Proper name for this is `canIgnoreFailures` which was the name I'd for it initially). 

However, everywhere within samza codebase config method for boolean variables are used with prefix `get`(JobConfig.scala (getMonitorPartitionChange, getSingleThreadMode), ClusterBasedConfig(getHostAffinityEnabled) etc). To maintain consistency throughout the codebase, I'd to name like that.

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitor.java
shanthoosh: Fixed.

samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
shanthoosh: StringUtils `newStringUtf8` static method provides different kinds of error handling (guards against NPE's (I've verified this), graceful handling of encoding errors) that default string constructor doesn't have. If we switch we need to do all those things ourselves.

samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
shanthoosh: Makes sense. Fixed.

samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
shanthoosh: Makes sense. Changed.

samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
shanthoosh: Makes sense. Fixed.

samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
shanthoosh: Discussed with you offline. 

Renamed it to queryJobStatusServers. Since the iterating logic is common, hence it's separate.

samza-rest/src/main/java/org/apache/samza/monitor/JobsClient.java
shanthoosh: This(`retriableHttpGet`) was shared by both `getJobStatus`, `getTasks` method within this class. 

Since the same logic is shared by both of these methods, it's kept separate to avoid duplication.

Reason for keeping it generic is both of these methods expect different type as result, passing in type for that.

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitorConfig.java
prateekm: I think our current checkstyle settings require javadoc for public methods. I'd personally prefer skipping them when they're redundant/obvious as well. In any case, if we add them let's make them consistent.

samza-rest/src/test/java/org/apache/samza/monitor/TestLocalStoreMonitor.java
prateekm: Let's log the exception in that case. Does assert.fail take an exception?

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitorConfig.java
```diff
@@ -53,6 +53,12 @@
    */
   private static final long DEFAULT_OFFSET_FILE_TTL_MS = 1000 * 60 * 60 * 24 * 7;
 
+  /**
+   * Will make LocalStoreMonitor to ignore failures during store clean ups.
```
prateekm: "make LocalStoreMonitor ignore"

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitor.java
prateekm: Sorry, I meant to say that you should clarify in documentation (btw, let's also update configuration documentation as a part of this PR) what kind of failures this is talking about. Also if I understand correctly, wouldn't the issue you mention not result in an exception since at least one of the RMs has a response in queryJobStatusServers?

samza-rest/src/main/java/org/apache/samza/monitor/LocalStoreMonitorConfig.java
shanthoosh: Fixed.

samza-rest/src/test/java/org/apache/samza/monitor/TestLocalStoreMonitor.java
shanthoosh: Assert.fail doesn't take exception. 

It takes in message as argument. Added e.getMessage() to it. 

`e.printStackTrace()` might not be useful, it depends if gradle setting redirects stdout/LOG messages in test to build output.

prateekm: You can just use a Logger to log in that case. e.getMessage isn't usually sufficient.

shanthoosh: Logged exception at error level before Assert.fail. Done.

PR #125, opened 2017-04-18
SAMZA-1213 - StreamProcessorLifeCycleAware interface should not use processorId

navina:
Refactoring LocalApplicationRunner s.t. each processor has its own listener instance, instead of a single listener keeping track of all processors.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessorLifeCycleAware.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessorLifeCycleAware.java
@@ -31,21 +31,18 @@ import org.apache.samza.annotation.InterfaceStability;
 public interface StreamProcessorLifeCycleAware {
   /**
    * Callback when the {@link StreamProcessor} is started
-   * @param processorId id of the StreamProcessor
    */
-  void onStart(String processorId);
+  void onStart();
 
   /**
    * Callback when the {@link StreamProcessor} is shut down.
-   * @param processorId id of the StreamProcessor
    */
-  void onShutdown(String processorId);
+  void onShutdown();
 
   /**
    * Callback when the {@link StreamProcessor} fails
-   * @param processorId id of the StreamProcessor
    * @param t exception of the failure
    */
-  void onFailure(String processorId, Throwable t);
+  void onFailure(Throwable t);
 
 }
--- a/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
@@ -30,6 +30,7 @@ import org.apache.samza.SamzaException;
 import org.apache.samza.application.StreamApplication;
 import org.apache.samza.config.ApplicationConfig;
 import org.apache.samza.config.Config;
+import org.apache.samza.config.ConfigException;
 import org.apache.samza.coordinator.CoordinationServiceFactory;
 import org.apache.samza.coordinator.CoordinationUtils;
 import org.apache.samza.coordinator.Latch;
@@ -61,19 +62,27 @@ public class LocalApplicationRunner extends AbstractApplicationRunner {
   private final List<StreamProcessor> processors = new ArrayList<>();
   private final CountDownLatch latch = new CountDownLatch(1);
   private final AtomicReference<Throwable> throwable = new AtomicReference<>();
+  private final ConcurrentHashSet<String> processorIds = new ConcurrentHashSet<>();
 
   private ApplicationStatus appStatus = ApplicationStatus.New;
 
-  final class LocalProcessorListener implements StreamProcessorLifeCycleAware {
-    private final ConcurrentHashSet<String> processorIds = new ConcurrentHashSet<>();
+  final class LocalStreamProcessorListener implements StreamProcessorLifeCycleAware {
+    public final String processorId;
+
+    public LocalStreamProcessorListener(String processorId) {
+      if (processorId == null) {
+        throw new NullPointerException("processorId cannot be null in LocalStreamProcessorListener!!");
+      }
+      this.processorId = processorId;
+    }
 
     @Override
-    public void onStart(String processorId) {
+    public void onStart() {
       processorIds.add(processorId);
     }
 
     @Override
-    public void onShutdown(String processorId) {
+    public void onShutdown() {
       processorIds.remove(processorId);
       if (processorIds.isEmpty()) {
         latch.countDown();
@@ -81,7 +90,7 @@ public class LocalApplicationRunner extends AbstractApplicationRunner {
     }
 
     @Override
-    public void onFailure(String processorId, Throwable t) {
+    public void onFailure(Throwable t) {
       processorIds.remove(processorId);
       throwable.compareAndSet(null, t);
       latch.countDown();
@@ -107,10 +116,11 @@ public class LocalApplicationRunner extends AbstractApplicationRunner {
       if (plan.getJobConfigs().isEmpty()) {
         throw new SamzaException("No jobs to run.");
       }
-      LocalProcessorListener listener = new LocalProcessorListener();
       plan.getJobConfigs().forEach(jobConfig -> {
           log.info("Starting job {} StreamProcessor with config {}", jobConfig.getName(), jobConfig);
-          StreamProcessor processor = createStreamProcessor(jobConfig, app, listener);
+          String processorId = getProcessorId(config);
+          StreamProcessor processor =
+              createStreamProcessor(processorId, jobConfig, app, new LocalStreamProcessorListener(processorId));
           processor.start();
           processors.add(processor);
         });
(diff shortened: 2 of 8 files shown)


navina:
@xinyuiscool @nickpan47 Changes based on what I proposed in slack. Review please!

navina:
@prateekm

samza-core/src/main/java/org/apache/samza/processor/StreamProcessorLifeCycleAware.java
```diff
@@ -31,21 +31,18 @@
...
    * Callback when the {@link StreamProcessor} is shut down.
-   * @param processorId id of the StreamProcessor
    */
-  void onShutdown(String processorId);
+  void onShutdown();
 
   /**
    * Callback when the {@link StreamProcessor} fails
-   * @param processorId id of the StreamProcessor
    * @param t exception of the failure
```
prateekm: Minor: "cause of the failure"

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -61,27 +62,35 @@
...
 
   private ApplicationStatus appStatus = ApplicationStatus.New;
 
-  final class LocalProcessorListener implements StreamProcessorLifeCycleAware {
-    private final ConcurrentHashSet<String> processorIds = new ConcurrentHashSet<>();
+  final class LocalStreamProcessorListener implements StreamProcessorLifeCycleAware {
+    public final String processorId;
+
+    public LocalStreamProcessorListener(String processorId) {
+      if (processorId == null) {
```
prateekm: Minor: StringUtils.isEmpty?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -61,27 +62,35 @@
...
+        throw new NullPointerException("processorId cannot be null in LocalStreamProcessorListener!!");
```
prateekm: Minor: s/!!/.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -179,20 +189,51 @@ public ApplicationStatus status(StreamApplication streamApp) {
   }
 
   /**
+   * Generates processorId for each {@link StreamProcessor} using the configured {@link ProcessorIdGenerator}
+   *
+   * @param config Application config
+   * @return String that uniquely represents an instance of {@link StreamProcessor} in the current JVM
+   */
+  /* package private */
+  String getProcessorId(Config config) {
```
prateekm: Same question as in PR 117. Shouldn't JobCoordinator be generating the processorId according to SEP-1? Is this temporary? Same with passing processorId as a constructor param to StreamProcessor.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -179,20 +189,51 @@ public ApplicationStatus status(StreamApplication streamApp) {
...
+
+  /**
    * Create {@link StreamProcessor} based on {@link StreamApplication} and the config
    * @param config config
    * @param app {@link StreamApplication}
    * @return {@link StreamProcessor]}
    */
-  /* package private */ StreamProcessor createStreamProcessor(Config config, StreamApplication app, StreamProcessorLifeCycleAware listener) {
+  /* package private */
+  StreamProcessor createStreamProcessor(
```
prateekm: Same as PR 117: I think will be cleaner to make it a StreamProcessor constructor.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessorLifeCycleAware.java
```diff
@@ -31,21 +31,18 @@
 public interface StreamProcessorLifeCycleAware {
```
prateekm: Same as PR 117: StreamProcessorLifecycleListener (and update field names too).

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -179,20 +189,51 @@ public ApplicationStatus status(StreamApplication streamApp) {
...
+      return new StreamProcessor(
+          processorId, config, new HashMap<>(), (StreamTaskFactory) taskFactory, listener);
     } else if (taskFactory instanceof AsyncStreamTaskFactory) {
-      return new StreamProcessor(config, new HashMap<>(), (AsyncStreamTaskFactory) taskFactory, listener);
+      return new StreamProcessor(
+          processorId, config, new HashMap<>(), (AsyncStreamTaskFactory) taskFactory, listener);
     } else {
       throw new SamzaException(String.format("%s is not a valid task factory",
-          taskFactory.getClass().getCanonicalName().toString()));
+          taskFactory.getClass().getCanonicalName()));
```
prateekm: Thanks.

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
```diff
@@ -85,7 +71,27 @@ public void testStreamProcessor() {
     // Note: createTopics needs to be called before creating a StreamProcessor. Otherwise it fails with a
     // TopicExistsException since StreamProcessor auto-creates them.
     createTopics(inputTopic, outputTopic);
-    final StreamProcessor processor = new StreamProcessor(new MapConfig(configs), new HashMap<>(), IdentityStreamTask::new, listener);
+    final StreamProcessor processor = new StreamProcessor(
+        "1",
+        new MapConfig(configs),
+        new HashMap<>(),
+        IdentityStreamTask::new,
+        new StreamProcessorLifeCycleAware() {
```
prateekm: Same as PR 117: Prefer using mocks for these instead of anonymous inner classes.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessorLifeCycleAware.java
navina: Ok. I will update it.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessorLifeCycleAware.java
navina: Ok

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
navina: Sure

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
navina: ok

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
navina: > Shouldn't JobCoordinator be generating the processorId according to SEP-1? 

Yeah. We established that processorId is to be generated by the jobCoordinator. However, if the interface for generating processorID is configurable, then it should ideally come from the user-facing layer. 

> Is this temporary? 

Not sure if it is temporary. There is some blurring of the lines in terms of which component generates the processorId and how it relates to the executing environment. When JobCoordinator was initially designed, Jobcoordinator was taking on the role of coordinator AND as a specification of the executing environment. With ApplicationRunners and the newer interfaces, this assumption seems to have shifted. 

I don't believe there was any deliberate discussion when it was decided that processorId can be moved out of the JobCoordinator layer. Xinyu and I just decided agreed over slack because it was useful for ApplicationRunner to track the lifecycle of the StreamProcessor. However, whether it has to be done via the processorId is a reasonable question. It stems from the fact that there is no clear contract between the applicationrunners and StreamProcessor. My two cents.

navina: This also brings up another question - what is the purpose of getProcessorId in the JobCoordinator interface?? :)

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
navina: Are you suggesting not to use a helper method? I would love to clean it up. However, it means I have to update all the test code which is heavily pivoting on the mocks. 
I will give it a try. But if it is significant changes to the test code, I would rather have the test owner to fix it :)

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
navina: np!

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
navina: I am not a big fan of mocks at this time. I think mocks are useful when the APIs are relatively stable/mature. Given how frequently we are shuffling classes around, I prefer not to rely on mocks so much.

One of the reasons why I had to refactor LocalApplicationRunner this way is that it was very difficult to change the mocked behavior (spy,mock, capture etc) in the test class. I will refrain from using further mocks until we can stabilize the components.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
navina: @prateekm  I am not addressing this here. It requires changes to the test class as well and I don't feel comfortable doing it. I suggest opening a separate JIRA to clarify (in written form please) your questions on PR117 and fix forward the changes you recommend.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: I am wondering about that too. Is there a future plan to use Zk to provide processorId? Feels that can still be done by configuring a Zk processorId generator.

prateekm: Summarizing our offline discussion with @vjagadish1989:

> However, if the interface for generating processorID is configurable, then it should ideally come from the user-facing layer.
> What is the purpose of getProcessorId in the JobCoordinator interface?

From the StreamProcessor's perspective, processorId is a way for it to know what part of the JobModel it needs to work on. Some JobCoordinators (e.g. ZK based JobCoordinator) may want to handle work assignment (and hence processorId assignment) dynamically as processors join and leave the group. Therefore, it makes sense to me to leave the option of  assigning the processorId dynamically at runtime open and let JC be the source of truth for these IDs. 

Regarding the ProcessorIdGenerator interface, it's pluggable in the sense that it lets JCs avoid reimplementing the same getProcessorId() logic (e.g. get id from config) by choosing a concrete implementation, but not in the sense that users can pick any one they like. A particular JobCoordinator implementation will usually require a specific ProcessorIdGenerator. Since we don't have that many JC implementations I don't mind removing it.

Does this make sense?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
prateekm: No problem, we can address this in @xinyuiscool's follow up for #117. I already commented there.

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
prateekm: Makes sense, let's change to mocks when things settle down.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
navina: > let JC be the source of truth for these IDs.

Just to be clear, the JC within each StreamProcessor will be responsible for creating this processorId and registering. This is the current behavior and I hope everyone is in agreement with that. 
In case of yarn, the leader (aka AM) will assign processorIds through the environment. When we switch to use StreamProcessor with YarnJobCoordinator, the processorId generator will simply parse the processorId from the environment variable set by the AM.

PR #127, opened 2017-04-18
SAMZA-1204: Visualize StreamGraph and ExecutionPlan

xinyuiscool:
Once a Samza application (using fluent API) is deployed, an execution plan will be generated by the ExecutionPlanner. The plan JSON will be written to a file (plan.json) under the ./plan directory, which also contains the plan.html and javscripts (js folder).


xinyuiscool:
btw, also addressed all the comments Prateek had in the previous [PR 110](https://github.com/apache/samza/pull/110).

xinyuiscool:
@cpettitt : thanks for providing all the demo examples for Dagre. It works out of the box. Very cool library!

navina:
@xinyuiscool  Can you move the detailed description from the PR description to the JIRA itself? You should be aware of the merge-script issue ([SAMZA-1184](https://issues.apache.org/jira/browse/SAMZA-1184)) that causes the entire description to be used in the commit message.

xinyuiscool:
@navina : sure, already copied the description into the JIRA. I will leave the description here for now so reviewers can see the visual graph directly, and will remove the description when committing. Thanks for pointing it out.

xinyuiscool:
@prateekm and @jmakes : please take a look when you get a chance. Thanks!

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -106,4 +112,25 @@ final ExecutionPlan getExecutionPlan(StreamApplication app) throws Exception {
...
   }
+
+  /**
+   * Write the execution plan JSON to a file
+   * @param planJson JSON representation of the plan
+   */
+  final void writePlanJsonFile(String planJson) {
+    try {
+      String content = "plan='" + planJson + "'";
+      String binPath = System.getenv(ShellCommandConfig.JOB_BIN_DIR());
```
jmakes: Why bin dir and not log?

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -106,4 +112,25 @@ final ExecutionPlan getExecutionPlan(StreamApplication app) throws Exception {
...
+      if (binPath != null && !binPath.isEmpty()) {
+        // Write the plan json to bin path which also contains the html and js files
+        File file = new File(binPath + "/plan.json");
+        file.setReadable(true, false);
+        PrintWriter writer = new PrintWriter(file, "UTF-8");
+        writer.println(content);
+        writer.close();
+      }
+    } catch (Throwable t) {
+      log.warn("fail to write execution plan json to file", t);
```
jmakes: Nit: case and tense -> "Failed"

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
xinyuiscool: hmm, you're right. log is a better place to put it. I was thinking maybe put them along with plan.html is easier since they can be copied over.  But I think I can let the html to cover both cases. Let me update the patch with this change.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
xinyuiscool: Fixed!

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
xinyuiscool: I looked at the code again and have to say it's actually more practical to have the json under the bin folder. Since the plan.html needs a static reference to the path of the json file, and the log folder can be configured to anywhere by the user, it's less error-prone to put the json along with the html under bin folder. Since neither bin or log folder in RM box is exposed in yarn web proxy, the user has to log on to the RM box to get the files in either case. So I guess it should be Okey to have it in bin.

jmakes: I just feel that it's unconventional to write to anything to bin. I wouldn't even look there for output of an application.

If it must be a static location, I suggest creating a dir called "execution_plan" and put the files there.

xinyuiscool: Sounds like a good idea. Let me give that try.

xinyuiscool: I updated the patch with this change. Now the plan.json will be under /plan folder. I tested with the samza-api-test job and it works well. Thanks a lot for the suggestion!

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlan.java
```diff
@@ -43,9 +45,10 @@
   List<StreamSpec> getIntermediateStreams();
 
   /**
-   * Returns the JSON representation of the plan for visualization
+   * Returns the JSON representation of the plan.
    * @return json string
```
prateekm: "JSON representation of the plan"

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlan.java
```diff
@@ -43,9 +45,10 @@
...
-   * @throws Exception exception
+   * @throws Exception exception during JSON serialization, including {@link java.io.IOException}
+   * and {@link org.codehaus.jackson.JsonGenerationException}
```
prateekm: Indent to align with "exception during"

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
```diff
@@ -65,40 +66,46 @@
...
-   * Returns the configs for single stage job, in the order of topologically sort.
-   * @return list of job configs
-   */
+  @Override
   public List<JobConfig> getJobConfigs() {
-    return getJobNodes().stream().map(JobNode::generateConfig).collect(Collectors.toList());
+    String json = "";
+    try {
+      json = getPlanAsJson();
+    } catch (Throwable t) {
```
prateekm: Should catch Exception, not Throwable.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -39,69 +40,85 @@
...
-    @JsonProperty("PairedOpId")
+    @JsonProperty("pairedOpId")
     int pairedOpId = -1;  //for join operator, we will have a pair nodes for two partial joins
+    @JsonProperty("caller")
+    String caller;
   }
 
   static final class StreamSpecJson {
-    @JsonProperty("Id")
+    @JsonProperty("id")
```
prateekm: "streamId"

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -119,13 +136,18 @@
     JobGraphJson jobGraphJson = new JobGraphJson();
 
     // build StreamEdge JSON
-    jobGraphJson.streams = new HashMap<>();
-    jobGraph.getSources().forEach(e -> getOrCreateStreamEdgeJson(e, jobGraphJson.streams));
-    jobGraph.getSinks().forEach(e -> getOrCreateStreamEdgeJson(e, jobGraphJson.streams));
-    jobGraph.getIntermediateStreamEdges().forEach(e -> getOrCreateStreamEdgeJson(e, jobGraphJson.streams));
+    ApplicationConfig appConfig = jobGraph.getApplicationConfig();
+    jobGraphJson.applicationName = appConfig.getAppName();
```
prateekm: Same comment as for other places where we're mutating DAG nodes. It's usually safer to pre-create these and pass as constructor param to avoid accidentally forgetting to set a field. It's also easier to understand since the object is immutable once created. Would it be difficult to do that here?

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -137,10 +159,9 @@
   /**
    * Create JSON POJO for a {@link JobNode}, including the {@link org.apache.samza.operators.StreamGraph} for this job
    * @param jobNode job node in the {@link JobGraph}
-   * @param streamEdges map of {@link org.apache.samza.execution.JobGraphJsonGenerator.StreamEdgeJson}
    * @return {@link org.apache.samza.execution.JobGraphJsonGenerator.JobNodeJson}
```
prateekm: Should add description for the returned object, not just type. Also, probably don't need the fully qualified name.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -191,13 +216,18 @@ private void updateOperatorGraphJson(MessageStreamImpl messageStream, Traversabl
...
-  private OperatorJson getOrCreateOperatorJson(OperatorSpec opSpec, OperatorGraphJson opGraph) {
+  private OperatorJson buildOperatorJson(OperatorSpec opSpec, OperatorGraphJson opGraph,
+      MessageStreamImpl messageStream) {
     Map<Integer, OperatorJson> operators = opGraph.operators;
     OperatorJson opJson = operators.get(opSpec.getOpId());
     if (opJson == null) {
       opJson = new OperatorJson();
       opJson.opCode = opSpec.getOpCode().name();
       opJson.opId = opSpec.getOpId();
+      StackTraceElement stackTraceElement = messageStream.getCallerStackTrace(opSpec.getOpId());
```
prateekm: Instead of the ExecutionPlanner needing to know the stacktrace, it'll be better if this is a field in the OperatorSpec directly. I.e., this class should just call opSpec.getSourceLine() or whatever this is called (not caller).

Even more generally, it'll be better to make this a part of the toString or toJson of each OperatorSpec. If we do that each OperatorSpec can provide additional information about itself, e.g. trigger information in windows or TTL for join.

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
```diff
@@ -92,14 +93,21 @@ void addOutEdge(StreamEdge out) {
     return outEdges;
   }
 
-  public JobConfig generateConfig() {
+  /**
+   * Generate the configs for a job
+   * @param executionPlanJson JSON representation of the execution plan
+   * @return config of the job
+   */
+  public JobConfig generateConfig(String executionPlanJson) {
```
prateekm: I don't understand this method signature. Why should generateJobConfig take executionPlanJson? Also, why is it being added to the config?

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -63,6 +65,11 @@
   private final Set<OperatorSpec> registeredOperatorSpecs = new HashSet<>();
 
   /**
+   * Track the caller stack trace
+   */
+  private final Map<Integer, StackTraceElement> callerStackTraces = new HashMap<>();
```
prateekm: Let's move to each OperatorSpec. See comment above.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -106,4 +112,25 @@ final ExecutionPlan getExecutionPlan(StreamApplication app) throws Exception {
   final StreamManager getStreamManager() {
     return streamManager;
   }
+
+  /**
+   * Write the execution plan JSON to a file
+   * @param planJson JSON representation of the plan
+   */
+  final void writePlanJsonFile(String planJson) {
```
prateekm: I'd prefer logging the plan and allowing user to copy paste it into the viewer instead of creating files on disk ourselves. This environment variable might not exist in standalone for example. If users want a separate logger/file for this they can do that in their log4j.xml.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -106,4 +112,25 @@ final ExecutionPlan getExecutionPlan(StreamApplication app) throws Exception {
...
+      String planPath = System.getenv(ShellCommandConfig.EXECUTION_PLAN_DIR());
+      if (planPath != null && !planPath.isEmpty()) {
+        // Write the plan json to bin path which also contains the html and js files
+        File file = new File(planPath + "/plan.json");
+        file.setReadable(true, false);
+        PrintWriter writer = new PrintWriter(file, "UTF-8");
+        writer.println(content);
+        writer.close();
+      }
+    } catch (Throwable t) {
```
prateekm: Shouldn't catch and ignore Throwable.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -39,69 +40,85 @@
...
-    @JsonProperty("Operators")
+    @JsonProperty("operators")
     Map<Integer, OperatorJson> operators = new HashMap<>();
+    @JsonProperty("outputStreams")
+    List<String> outputStreams;
   }
 
   static final class InputStreamJson extends Traversable {
-    @JsonProperty("StreamId")
+    @JsonProperty("streamId")
```
prateekm: Same comment as previous PR: Input and output streams should be symmetric. Currently InputStreamJson has nextOperatorIds but outputStreamIds are part of the OperatorJson.

build.gradle
```diff
@@ -391,10 +393,12 @@ project(":samza-shell") {
   }
 
   task shellTarGz(type: Tar) {
+    println 'shellTarGz'
```
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlan.java
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlan.java
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -39,69 +40,85 @@
...
-    @JsonProperty("OpId")
+    @JsonProperty("opId")
     int opId;
-    @JsonProperty("OutputStreamId")
+    @JsonProperty("outputStreamId")
     String outputStreamId;
-    @JsonProperty("PairedOpId")
+    @JsonProperty("pairedOpId")
     int pairedOpId = -1;  //for join operator, we will have a pair nodes for two partial joins
+    @JsonProperty("caller")
```
xinyuiscool: Yup, it does sound like method name. Changed it to sourceLocation.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
xinyuiscool: I will keep this since the id is the same var name as StreamSpec.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
xinyuiscool: Makes sense. I unified both to use the StreamJson type.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
xinyuiscool: If we are using constructor, do we need to have the setter for deserialization? Btw, it is easier to directly operates on the objects during traversal :)

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
xinyuiscool: Ok, I added the sourceLocation to the operator constructor and have the getSourceLocation() in the API now. Also add the toJson() to the operatorSpec. Now we can have customized JSON for different operators. Thanks a lot for this suggestion!

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
xinyuiscool: Right now if we have the file, then user can simplify put the file into the same directory with plan.html and then they can view it. The plan is also serialized in the config so the user can also paste that in a file and view it locally (put the file under the same folder as plan.html). In the future we can also provide a web page that let user enters the json. But to me, the future dashboard will be a better place to put this. Right now generating the file helps local debugging by opening the html directly.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
xinyuiscool: Right now json is only for viewing so it should be ok to ignore. But if we add the serialization of the operators, then I will change here to throw.

prateekm: Shouldn't ignore Throwables since they can be unrelated to the code here (OOM etc.).

samza-core/src/main/java/org/apache/samza/operators/util/OperatorJsonUtils.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.apache.samza.operators.spec.OperatorSpec;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class OperatorJsonUtils {
```
prateekm: There's nothing JSON specific about this class. Also, an OperatorSpec should either provide access to its properties via getters, or return its string/json/map/object representation itself that other code can consume. Maybe just add `OperatorSpec#getSourceLocation()` to go along with `getOpCode` and `getOpId` for now?

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
prateekm: Agree, its easier to write when it's mutable, but it's more difficult to understand/refactor later. Up to you, whichever feels cleaner to you.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -106,4 +112,25 @@ final ExecutionPlan getExecutionPlan(StreamApplication app) throws Exception {
...
     return streamManager;
   }
+
+  /**
+   * Write the execution plan JSON to a file
+   * @param planJson JSON representation of the plan
+   */
+  final void writePlanJsonFile(String planJson) {
+    try {
+      String content = "plan='" + planJson + "'";
```
prateekm: Maybe leave contents as pure JSON? Will make it simpler to reformat this file in an editor to look at it directly.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -106,4 +112,25 @@ final ExecutionPlan getExecutionPlan(StreamApplication app) throws Exception {
...
+      String planPath = System.getenv(ShellCommandConfig.EXECUTION_PLAN_DIR());
+      if (planPath != null && !planPath.isEmpty()) {
+        // Write the plan json to bin path which also contains the html and js files
```
prateekm: Update comment.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
prateekm: I'd still prefer not adding a new public configuration/environment variable for this. Is there one we can reuse? E.g., can we put it in the logs directory and make the path in the bin/plan.html relative?

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpec.java
```diff
@@ -64,6 +65,18 @@
   int getOpId();
 
   /**
+   * Return a map object for JSON representation of the operator
+   * @return a map of JSON POJO objects
+   */
+  Map<String, Object> toJsonMap();
```
prateekm: See comment on OperatorJsonUtils.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
xinyuiscool: I expect refactoring this piece once we start to think about serialize and deserialize the operator graph. I will change it to do something like the serde in SamzaObjectMapper.  It was not done that way earlier is that I didn't expect the deserialization will happen. For now I will keep as is.

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
xinyuiscool: sure. Changed the config name to be samza.internal.execution.plan.

For jobGraph it will be ok. But it might cause issue for operator lamdas. I will send a discussion email about it once the feature is onboard.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
xinyuiscool: Discussed offline. For remote execution plan var is defined with the folder ./plan, so it doesn't expose to the user. For local runner, it requires setting up this var as if the user want to write the plan to a file. We will see how this works out and there will be future improvement for the plan folder.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
xinyuiscool: Done.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
xinyuiscool: That's what I wanted too, but javascript is not able to load it :(. I spent quite a while looking for a good solution and this is the safest way to getting the file loaded in the html. Other solutions such as ajax to load local file might run into security and browser issues.

Good news is that the content of the json is also in the config now, so the user can copy/paste that into the json editor :). This file is primarily just for visualization.

PR #131, opened 2017-04-19
SAMZA-1176: Intermittent TestJoinOperator unit test failure

prateekm:
Join TTL was set too low (10 ms). `joinRetainsMatchedMessagesReverse` will fail if the execution time between line 176 and 186 is longer than that. Increased TTL to 1 min (will not affect test execution time).


navina:
@prateekm I thought `joinRetainsMatchedMessagesReverse` was failing intermittently. How is advancing clock time on other tests going to fix this? Are the tests related?

prateekm:
@navina The clock time in those tests needs to advanced by an amount larger than TTL (100 ms == 10 times TTL of 10 ms). Since I increased the TTL to 1 minute, that amount now needs to be higher. I just rewrote that amount explicitly in terms of the TTL.

PR #132, opened 2017-04-19
Profiling State Performance

s-noghabi:
This is s commit for [SAMZA-1174](https://issues.apache.org/jira/browse/SAMZA-1174). This commit involves gathering a log of operations (read, write, delete, etc.) happening on the state and publishing them into a kafka topic. It is names as "Access Log" behaving similar to changelog, but gathering log information.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/storage/StorageEngineFactory.java
+++ b/samza-api/src/main/java/org/apache/samza/storage/StorageEngineFactory.java
@@ -43,6 +43,7 @@ public interface StorageEngineFactory<K, V> {
    * @param collector MessageCollector the storage engine uses to persist changes.
    * @param registry MetricsRegistry to which to publish storage-engine specific metrics.
    * @param changeLogSystemStreamPartition Samza stream partition from which to receive the changelog.
+   * @param accessLogSystemsStreamPartition Information about the accesslog stream
    * @param containerContext Information about the container in which the task is executing.
    * @return The storage engine instance.
    */
@@ -54,5 +55,6 @@ public interface StorageEngineFactory<K, V> {
     MessageCollector collector,
     MetricsRegistry registry,
     SystemStreamPartition changeLogSystemStreamPartition,
+    SystemStreamPartition accessLogSystemsStreamPartition,
     SamzaContainerContext containerContext);
 }
--- a/samza-api/src/main/java/org/apache/samza/storage/StorageEngineFactory.java
+++ b/samza-api/src/main/java/org/apache/samza/storage/StorageEngineFactory.java
@@ -43,7 +43,6 @@ public interface StorageEngineFactory<K, V> {
    * @param collector MessageCollector the storage engine uses to persist changes.
    * @param registry MetricsRegistry to which to publish storage-engine specific metrics.
    * @param changeLogSystemStreamPartition Samza stream partition from which to receive the changelog.
-   * @param accessLogSystemsStreamPartition Information about the accesslog stream
    * @param containerContext Information about the container in which the task is executing.
    * @return The storage engine instance.
    */
@@ -55,6 +54,5 @@ public interface StorageEngineFactory<K, V> {
     MessageCollector collector,
     MetricsRegistry registry,
     SystemStreamPartition changeLogSystemStreamPartition,
-    SystemStreamPartition accessLogSystemsStreamPartition,
     SamzaContainerContext containerContext);
 }
--- a/samza-core/src/main/scala/org/apache/samza/config/StorageConfig.scala
+++ b/samza-core/src/main/scala/org/apache/samza/config/StorageConfig.scala
@@ -35,9 +35,9 @@ object StorageConfig {
   val CHANGELOG_SYSTEM = "job.changelog.system"
   val CHANGELOG_DELETE_RETENTION_MS = "stores.%s.changelog.delete.retention.ms"
   val DEFAULT_CHANGELOG_DELETE_RETENTION_MS = TimeUnit.DAYS.toMillis(1)
-  val ACCESSLOG_STREAM = "access-log"
-  val ACCESSLOG_SAMPLE = "stores.%s.accesslog.sample"
-  val ACCESSLOG_STATUS = "stores.%s.accesslog"
+  val ACCESSLOG_STREAM_SUFFIX = "access-log"
+  val ACCESSLOG_SAMPLING_RATIO = "stores.%s.accesslog.sampling.ratio"
+  val ACCESSLOG_STATUS = "stores.%s.accesslog.status"
   val DEFAULT_ACCESSLOG_SAMPLE = 50
 
 
@@ -50,7 +50,7 @@ class StorageConfig(config: Config) extends ScalaMapConfig(config) with Logging 
   def getStorageKeySerde(name: String) = getOption(StorageConfig.KEY_SERDE format name)
   def getStorageMsgSerde(name: String) = getOption(StorageConfig.MSG_SERDE format name)
 
-  def getAccessLogSetting(storeName: String) = {
+  def getAccessLogEnabled(storeName: String) = {
     if (containsKey(ACCESSLOG_STATUS format storeName))
       getBoolean(ACCESSLOG_STATUS format storeName)
     else
@@ -78,14 +78,14 @@ class StorageConfig(config: Config) extends ScalaMapConfig(config) with Logging 
     systemStreamRes
   }
 
-  //Given a system, this method returns the accesslog system stream
+  //Returns the accesslog stream name given a changelog stream name
   def getAccessLogStream(changeLogStream: String) = {
-    changeLogStream + "-" + ACCESSLOG_STREAM
+    changeLogStream + "-" + ACCESSLOG_STREAM_SUFFIX
   }
 
-  def getSamplingSetting(storeName: String) = {
-    if (containsKey(ACCESSLOG_SAMPLE format storeName))
-      getInt(ACCESSLOG_SAMPLE format storeName)
+  def getAccessLogSamplingRatio(storeName: String) = {
+    if (containsKey(ACCESSLOG_SAMPLING_RATIO format storeName))
+      getInt(ACCESSLOG_SAMPLING_RATIO format storeName)
     else
       DEFAULT_ACCESSLOG_SAMPLE
   }
--- a/samza-core/src/main/scala/org/apache/samza/config/StorageConfig.scala
+++ b/samza-core/src/main/scala/org/apache/samza/config/StorageConfig.scala
@@ -35,6 +35,7 @@ object StorageConfig {
   val CHANGELOG_SYSTEM = "job.changelog.system"
   val CHANGELOG_DELETE_RETENTION_MS = "stores.%s.changelog.delete.retention.ms"
   val DEFAULT_CHANGELOG_DELETE_RETENTION_MS = TimeUnit.DAYS.toMillis(1)
+  val ACCESSLOG_STREAM = "stores.%s.accesslog"
 
   implicit def Config2Storage(config: Config) = new StorageConfig(config)
 }
@@ -60,6 +61,18 @@ class StorageConfig(config: Config) extends ScalaMapConfig(config) with Logging 
         else {
           throw new SamzaException("changelog system is not defined:" + systemStream.get)
         }
+      } else {
+        systemStream
+      }
+    systemStreamRes
+  }
+
+  def getAccessLogStream(name: String) = {
+    //Looking for the access log stream similar to the change log stream
+    val systemStream = getOption(ACCESSLOG_STREAM format name)
+    val systemStreamRes =
+      if (systemStream.isDefined && ! systemStream.getOrElse("").contains('.')) {
+          throw new SamzaException("accesslog system is not defined " + systemStream.get)
       } else {
         systemStream
       }
--- a/samza-core/src/main/scala/org/apache/samza/config/StorageConfig.scala
+++ b/samza-core/src/main/scala/org/apache/samza/config/StorageConfig.scala
@@ -37,8 +37,8 @@ object StorageConfig {
   val DEFAULT_CHANGELOG_DELETE_RETENTION_MS = TimeUnit.DAYS.toMillis(1)
   val ACCESSLOG_STREAM_SUFFIX = "access-log"
   val ACCESSLOG_SAMPLING_RATIO = "stores.%s.accesslog.sampling.ratio"
-  val ACCESSLOG_STATUS = "stores.%s.accesslog.status"
-  val DEFAULT_ACCESSLOG_SAMPLE = 50
+  val ACCESSLOG_ENABLED = "stores.%s.accesslog.enabled"
+  val DEFAULT_ACCESSLOG_SAMPLING_RATIO = 50
 
 
   implicit def Config2Storage(config: Config) = new StorageConfig(config)
@@ -51,10 +51,7 @@ class StorageConfig(config: Config) extends ScalaMapConfig(config) with Logging 
   def getStorageMsgSerde(name: String) = getOption(StorageConfig.MSG_SERDE format name)
 
   def getAccessLogEnabled(storeName: String) = {
-    if (containsKey(ACCESSLOG_STATUS format storeName))
-      getBoolean(ACCESSLOG_STATUS format storeName)
-    else
-      false
+    getBoolean(ACCESSLOG_ENABLED format storeName, false)
   }
 
   def getChangelogStream(name: String) = {
@@ -84,10 +81,7 @@ class StorageConfig(config: Config) extends ScalaMapConfig(config) with Logging 
   }
 
   def getAccessLogSamplingRatio(storeName: String) = {
-    if (containsKey(ACCESSLOG_SAMPLING_RATIO format storeName))
-      getInt(ACCESSLOG_SAMPLING_RATIO format storeName)
-    else
-      DEFAULT_ACCESSLOG_SAMPLE
+    getInt(ACCESSLOG_SAMPLING_RATIO format storeName, DEFAULT_ACCESSLOG_SAMPLING_RATIO)
   }
 
   def getChangeLogDeleteRetentionInMs(storeName: String) = {
--- a/samza-core/src/main/scala/org/apache/samza/config/StorageConfig.scala
+++ b/samza-core/src/main/scala/org/apache/samza/config/StorageConfig.scala
@@ -35,7 +35,11 @@ object StorageConfig {
   val CHANGELOG_SYSTEM = "job.changelog.system"
   val CHANGELOG_DELETE_RETENTION_MS = "stores.%s.changelog.delete.retention.ms"
   val DEFAULT_CHANGELOG_DELETE_RETENTION_MS = TimeUnit.DAYS.toMillis(1)
-  val ACCESSLOG_STREAM = "stores.%s.accesslog"
+  val ACCESSLOG_STREAM = "access-log"
+  val ACCESSLOG_SAMPLE = "stores.%s.accesslog.sample"
+  val ACCESSLOG_STATUS = "stores.%s.accesslog"
+  val DEFAULT_ACCESSLOG_SAMPLE = 50
+
 
   implicit def Config2Storage(config: Config) = new StorageConfig(config)
 }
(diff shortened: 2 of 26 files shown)


samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,129 @@
...
+
+  var interval = 0
+  val systemStream = accessLogSystemStreamPartition.getSystemStream
+  val partitionId = accessLogSystemStreamPartition.getPartition.getPartitionId
+  val serializer = new StringSerde("UTF-8")
+
+
+  def get(key: K): V = {
+    var message = DBOperations.READ + ", " + key
+    measureLatencyAndWriteToStream(message, store.get(key))
```
vjagadish1989: Would be cleaner to refactor `measureLatencyAndWriteToStream ` into:

```
 // returns the duration for executing a runnable.
      Duration getElapsedTime(Runnable runnable)
```
and 
```
 // serialize and log the operation on key k to the access log
      void accessLog(DBOperation operation, Key k, Duration millis)

```

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  def getAll(keys: util.List[K]): util.Map[K, V] = {
```
vjagadish1989: Often performing a bulk operation maybe efficient than multiple individual operations. Might be useful to log the latencies here as well?

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,129 @@
...
+  def size[T](obj: T, serde: Serde[T]): Integer = {
+    if (obj == null) {
+      return 0
+    }
+    val bytes = serde.toBytes(obj)
+    bytes.size
+  }
+
+  private def addHeader() = {
+    val header = "operation, key, latency, time"
```
vjagadish1989: Assuming we evolve schemas in a compatible manner, Not sure we need this `header` (unless it's a part of each message). The parser should implicitly infer the schema anyways.

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,129 @@
...
+    var msg = message
+    val timeStamp = System.nanoTime().toString
+
+    if (interval == 0) {
+      addHeader()
+    }
+
+    interval = 1
+    msg += ", " + latency + ", " + timeStamp
+    collector.send(new OutgoingMessageEnvelope(systemStream, partitionId, serializer.toBytes(timeStamp), serializer.toBytes(msg)))
```
vjagadish1989: - Would not `millis` suffice for the key in the `OutgoingMessage`? (The duration can be in `nanos` still) 

- As an aside, `System.nanoTime()` should be used only to measure elapsed durations as a timer, and cannot be used as a wall clock / timestamp (It's unrelated to any `time` since it represents the nanos since a totally arbitrary epoch)

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,129 @@
...
+  private def measureLatencyAndWriteToStream[R](message: String, block: => R):R = {
+    val time1 = System.nanoTime()
+    val result = block
+    val time2 = System.nanoTime()
+    val latency = time2 - time1
+    var msg = message
+    val timeStamp = System.nanoTime().toString
+
+    if (interval == 0) {
+      addHeader()
```
vjagadish1989: Can remove the `interval` flag entirely.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -289,13 +289,22 @@ object SamzaContainer extends Logging {
...
+
     val serdeManager = new SerdeManager(
       serdes = serdes,
       systemKeySerdes = systemKeySerdes,
       systemMessageSerdes = systemMessageSerdes,
       systemStreamKeySerdes = systemStreamKeySerdes,
       systemStreamMessageSerdes = systemStreamMessageSerdes,
-      changeLogSystemStreams = changeLogSystemStreams.values.toSet)
+      changeLogSystemStreams = changeLogSystemStreams.values.toSet,
+      accessLogSystemStreams = accessLogSystemStreams.values.toSet)
```
vjagadish1989: Is the `StorageEngineFactory` interface API changing? If this is for a specific temporary profiling, It would be nice to not change it?

samza-api/src/main/java/org/apache/samza/storage/StorageEngineFactory.java
```diff
@@ -43,6 +43,7 @@
    * @param collector MessageCollector the storage engine uses to persist changes.
    * @param registry MetricsRegistry to which to publish storage-engine specific metrics.
    * @param changeLogSystemStreamPartition Samza stream partition from which to receive the changelog.
+   * @param accessLogSystemsStreamPartition Information about the accesslog stream
```
jayasi: Removed the extra parameter now.

samza-core/src/main/scala/org/apache/samza/config/StorageConfig.scala
```diff
@@ -66,6 +67,18 @@ class StorageConfig(config: Config) extends ScalaMapConfig(config) with Logging
     systemStreamRes
   }
 
+  def getAccessLogStream(name: String) = {
```
jayasi: Changed to storeName

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
jayasi: As discussed with @prateekm, I have now removed the extra parameter from the `StorageEngineFactory` interface

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
jayasi: I thought of keeping it this way so as to avoid running the block twice. The current implementation returns the output of the block rather than just the time so that it can be forwarded it to the other wrapper stores

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
jayasi: I've added latency measurements to all methods now

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
jayasi: Removed the header now

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
jayasi: Sorry, I wasn't aware that nanoTime() couldn't be used as a timestamp. Changed to millis now

samza-core/src/main/scala/org/apache/samza/config/StorageConfig.scala
```diff
@@ -66,6 +78,18 @@ class StorageConfig(config: Config) extends ScalaMapConfig(config) with Logging
     systemStreamRes
   }
 
+  //Given a system, this method returns the accesslog system stream
```
prateekm: Documentation: Update comment ("Returns the accesslog stream name given a changelog stream name").

samza-core/src/main/scala/org/apache/samza/config/StorageConfig.scala
```diff
@@ -66,6 +78,18 @@ class StorageConfig(config: Config) extends ScalaMapConfig(config) with Logging
...
+  def getAccessLogStream(changeLogStream: String) = {
+    changeLogStream + "-" + ACCESSLOG_STREAM
+  }
+
+  def getSamplingSetting(storeName: String) = {
```
prateekm: Naming: getAccessLogSamplingRatio

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -298,6 +300,28 @@ object JobModelManager extends Logging {
...
+
+    for ((storeName, systemStream) <- changeLogSystemStreams) {
+      val accessLog = config.getAccessLogSetting(storeName)
+      if (accessLog) {
+        val systemAdmin = Util.getObj[SystemFactory](config
+          .getSystemFactory(systemStream.getSystem)
+          .getOrElse(throw new SamzaException("A stream uses system %s, which is missing from the configuration." format systemStream.getSystem))
+        ).getAdmin(systemStream.getSystem, config)
+
+        val accessLogSpec = new StreamSpec(config.getAccessLogStream(systemStream.getStream) + "_id",
```
prateekm: Minor: Don't need to append _id, can use as is.

samza-core/src/main/scala/org/apache/samza/serializers/SerdeManager.scala
```diff
@@ -33,12 +33,15 @@ class SerdeManager(
...
+  val accessLog = "access-log"
+
   def toBytes(obj: Object, serializerName: String) = serdes
     .getOrElse(serializerName, throw new SamzaException("No serde defined for %s" format serializerName))
     .toBytes(obj)
 
   def toBytes(envelope: OutgoingMessageEnvelope): OutgoingMessageEnvelope = {
-    val key = if (changeLogSystemStreams.contains(envelope.getSystemStream)) {
+    val key = if (changeLogSystemStreams.contains(envelope.getSystemStream)
+      || envelope.getSystemStream.getStream.endsWith(accessLog)) {
```
prateekm: Use ACCESSLOG_STREAM_SUFFIX constant from StorageConfig instead.

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.samza.config.StorageConfig
+import org.apache.samza.task.MessageCollector
+import org.apache.samza.util.Logging
+import org.apache.samza.system.{OutgoingMessageEnvelope, SystemStream, SystemStreamPartition}
+import org.apache.samza.serializers._
+
+class AccessLoggedStore[K, V](
+    val store: KeyValueStore[K, V],
+    val collector: MessageCollector,
+    val systemStreamPartition: SystemStreamPartition,
```
prateekm: Naming: changelogSystemStreamPartition

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,149 @@
...
+    val WRITE_ALL = Value("write_all")
+    val DELETE_ALL = Value("delete_all")
+  }
+
+  val streamName = storageConfig.getAccessLogStream(systemStreamPartition.getSystemStream.getStream)
+  val systemStream = new SystemStream(systemStreamPartition.getSystemStream.getSystem, streamName)
+  val partitionId: Int = systemStreamPartition.getPartition.getPartitionId
+  val serializer = new StringSerde("UTF-8")
+  val sample = storageConfig.getSamplingSetting(storeName)
+  val generator = scala.util.Random
```
prateekm: Naming: val rng (for random number generator)

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,149 @@
...
+  val DELIMITER = " || "
+
+  def get(key: K): V = {
+    var message = DBOperations.READ + DELIMITER + key
```
prateekm: Issue: The "+ key" is going to call key.toString, which is likely not what you want. Same for the toStringKey method.
I'd recommend using binary format for the message and putting the serialized key as a binary blob in it.

jayasi: I've created a separate class which stores the data. An object of this class is serialized and sent as a message now. I'm able to deserialize these messages from another job later on.

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
jayasi: Followed all suggestions about variable names in the latest commit

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -83,6 +83,12 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
     val storeFactory = storageConfig.get("factory")
     var storePropertiesBuilder = new StoreProperties.StorePropertiesBuilder()
 
+    val accessLog = if (storageConfig.containsKey("accesslog.status")) {
```
jayasi: Here, storageConfig is not of type StorageConfig and I get the following error if I try to cast it: 
``` java.lang.ClassCastException: org.apache.samza.config.MapConfig cannot be cast to org.apache.samza.config.StorageConfig ```
I've refactored it to use getBoolean(value, default) though

Comment (prateekm) [inline samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala]: Makes sense, thanks.

File: samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLogMessage.scala [hunk 17]
```diff
@@ -0,0 +1,46 @@
...
+
+package org.apache.samza.storage.kv
+
+import java.io.Serializable
+import java.io.ByteArrayOutputStream
+import java.io.ObjectOutputStream
+import java.util.ArrayList
+
+class AccessLogMessage(val DBOperation: Int,
+    val isRange: Boolean = false,
```
prateekm: @jayasi If you move this parameter with default value to the end of the parameter list it'll become optional. So you won't have to set isRange = false in get etc, only need to set it to true for range().

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,155 @@
...
+      }
+
+    keysInBytes
+  }
+
+  private def measureLatencyAndWriteToStream[R](dBOperations: Int, isRange: Boolean, singleKey: Array[Byte], keys: util.ArrayList[Array[Byte]], block: => R):R = {
+    val time1 = System.nanoTime()
+    val result = block
+    val time2 = System.nanoTime()
+    if (rng.nextInt() > samplingRatio) {
```
vjagadish1989: Would n't it be simpler to get rid off the intermediate `return`? You can invert this logic, optionally write to the access log and return the result anyways.

Something on the the lines of:

```
long startTime =..
retVal = operation();
long endTime = ..

if (shouldLog) {
     //write to Access Log 
  }

return result
```

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,155 @@
...
+
+  def flush(): Unit = {
+    trace("Flushing store.")
+
+    store.flush
+    trace("Flushed store.")
+  }
+
+
+  def toBytesKey(keys: util.Iterator[K]): util.ArrayList[Array[Byte]] = {
```
vjagadish1989: Should be 

```
ArrayList<ByteArray>serializeKeys(List<K> keys)
ArrayList<ByteArray>serializeEntries(List<K,V> entries)
```
unless there is specific semantics related to iterators we need

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,155 @@
...
+    measureLatencyAndWriteToStream(DBOperations.READ, isRange, toBytesOrNull(key), null, store.get(key))
+  }
+
+  def getAll(keys: util.List[K]): util.Map[K, V] = {
+    val isRange = true
+    measureLatencyAndWriteToStream(DBOperations.READ, isRange, null, toBytesKey(keys.iterator()), store.getAll(keys))
+  }
+
+  def put(key: K, value: V): Unit = {
+    val isRange = true
```
vjagadish1989: Why is a `put` on a per-key setting `range = true`? Would this not merely be a per-key operation?

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  def putAll(entries: util.List[Entry[K, V]]): Unit = {
+    val iter = entries.iterator
+    val isRange = true
+    measureLatencyAndWriteToStream(DBOperations.WRITE, isRange, null, toBytesKeyFromEntries(iter), store.putAll(entries))
+  }
+
+  def delete(key: K): Unit = {
+    val isRange = true
```
vjagadish1989: Should n't `isRange` be false here? Otherwise, you would count `delete`s to particular keys as range operations. Maybe, this is expected?

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,155 @@
...
+
+class AccessLoggedStore[K, V](
+    val store: KeyValueStore[K, V],
+    val collector: MessageCollector,
+    val changelogSystemStreamPartition: SystemStreamPartition,
+    val storageConfig: StorageConfig,
+    val storeName: String,
+    val keySerde: Serde[K]) extends KeyValueStore[K, V] with Logging {
+
+  object DBOperations extends Enumeration {
```
vjagadish1989: Very Minor: 
Prefer singular names for enums `s/DBOperations/DBOperation`

jayasi:
Fixed code based on comments. Let me know if there is anything else.

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLogMessage.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import java.io.Serializable
+import java.io.ByteArrayOutputStream
+import java.io.ObjectOutputStream
+import java.util.ArrayList
+
+class AccessLogMessage(val DBOperation: Int,
+    val duration: Long,
+    val keys: ArrayList[Array[Byte]],
+    val timestamp: Long
```
vjagadish1989: Since `timestamp` is always equal to `System.currentTimeMillis`, You can get rid off the `timestamp` parameter as a parameter in the constructor, and mint the `timestamp` inside the constructor.

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,155 @@
...
+    list.add(to)
+    logAccess(DBOperation.RANGE, toBytesKey(list), store.range(from, to))
+  }
+
+  def all(): KeyValueIterator[K, V] = {
+    store.all()
+  }
+
+  def close(): Unit = {
+    trace("Closing.")
```
vjagadish1989: s/Closing./Closing accessLogged store

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,155 @@
...
+
+  def flush(): Unit = {
+    trace("Flushing store.")
+
+    store.flush
+    trace("Flushed store.")
+  }
+
+
+  def toBytesKey(keys: util.List[K]): util.ArrayList[Array[Byte]] = {
```
vjagadish1989: s/toBytesKey/serializeKeys

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -0,0 +1,155 @@
...
+    if (iter != null)
+      while(iter.hasNext()) {
+        val entry = iter.next()
+        keysInBytes.add(toBytesOrNull(entry))
+      }
+
+    keysInBytes
+  }
+
+  def toBytesKeyFromEntries(list: util.List[Entry[K, V]]): util.ArrayList[Array[Byte]] = {
```
vjagadish1989: s/toBytesKeyFromEntries/serializeKeys

Both methods that perform key serialization, and should use the same name (but with different overrides)

vjagadish1989 (approved):
approved. looks good to me. 

f/b from final round of review:

As an aside, it'd be nice to verify actual de-serialization behavior by writing a separate `Samza` job.

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
jayasi: I tried keeping the same name for both methods but they were not able to overload because both of them become of the type ```serializeKeys(java.util.List)```. Hence keeping as separate names

vjagadish1989: Thank you for checking and confirming! LGTM.

vjagadish1989:
@jayasi @s-noghabi 
Looks like there are some merge conflicts? Can you do a quick sanity check and verify?

PR #133, opened 2017-04-20
SAMZA-1226: relax type parameters in MessageStream functions

nickpan47:
relax the type parameter in user supplied functions in fluent API

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/application/StreamApplication.java
+++ b/samza-api/src/main/java/org/apache/samza/application/StreamApplication.java
@@ -24,11 +24,46 @@ import org.apache.samza.operators.StreamGraph;
 
 
 /**
- * This interface defines a template for stream application that user will implement to create operator DAG in {@link StreamGraph}.
+ * This interface defines a template for stream application that user will implement to initialize operator DAG in {@link StreamGraph}.
+ *
+ * <p>
+ * User program implements {@link StreamApplication#init(StreamGraph, Config)} method to initialize the transformation logic
+ * from all input streams to output streams. A simple user code example is shown below:
+ * </p>
+ *
+ * <pre>{@code
+ * public class PageViewCounterExample implements StreamApplication {
+ *   Set<String> blackListMembers = new HashSet<>();
+ *
+ *   public void init(StreamGraph graph, Config config) {
+ *     MessageStream<PageViewEvent> pageViewEvents = graph.getInputStream("pageViewEventStream", (k, m) -> (PageViewEvent) m);
+ *     OutputStream<String, PageViewEvent, PageViewEvent> pageViewEventFilteredStream = graph
+ *       .getOutputStream("pageViewEventFiltered", m -> m.memberId, m -> m);
+ *
+ *     pageViewEvents
+ *       .filter(m -> !this.blackListMembers.contains(m.memberId))
+ *       .sendTo(pageViewEventFilteredStream);
+ *   }
+ *
+ *   // local execution mode
+ *   public static void main(String[] args) {
+ *     CommandLine cmdLine = new CommandLine();
+ *     Config config = cmdLine.loadConfig(cmdLine.parser().parse(args));
+ *     PageViewCounterExample userApp = new PageViewCounterExample();
+ *     userApp.initBlackList(config);
+ *     ApplicationRunner localRunner = ApplicationRunner.getLocalRunner(config);
+ *     localRunner.run(userApp);
+ *   }
+ *
+ *   private void initBlackList(Config config) {
+ *     // Read the blacklist from config
+ *   }
+ * }
+ * }</pre>
+ *
  */
 @InterfaceStability.Unstable
 public interface StreamApplication {
-  static final String APP_CLASS_CONFIG = "app.class";
 
   /**
    * Users are required to implement this abstract method to initialize the processing logic of the application, in terms
@@ -38,4 +73,5 @@ public interface StreamApplication {
    * @param config  the {@link Config} of the application
    */
   void init(StreamGraph graph, Config config);
+
 }
--- a/samza-api/src/main/java/org/apache/samza/application/StreamApplication.java
+++ b/samza-api/src/main/java/org/apache/samza/application/StreamApplication.java
@@ -33,7 +33,8 @@ import org.apache.samza.operators.StreamGraph;
  *
  * <pre>{@code
  * public class PageViewCounterExample implements StreamApplication {
- *   Set<String> blackListMembers = new HashSet<>();
+ *   // max timeout is 60 seconds
+ *   private static final MAX_TIMEOUT = 60000;
  *
  *   public void init(StreamGraph graph, Config config) {
  *     MessageStream<PageViewEvent> pageViewEvents = graph.getInputStream("pageViewEventStream", (k, m) -> (PageViewEvent) m);
@@ -41,7 +42,7 @@ import org.apache.samza.operators.StreamGraph;
  *       .getOutputStream("pageViewEventFiltered", m -> m.memberId, m -> m);
  *
  *     pageViewEvents
- *       .filter(m -> !this.blackListMembers.contains(m.memberId))
+ *       .filter(m -> !(m.getMessage().getEventTime() < System.currentTimeMillis() - MAX_TIMEOUT))
  *       .sendTo(pageViewEventFilteredStream);
  *   }
  *
@@ -55,9 +56,6 @@ import org.apache.samza.operators.StreamGraph;
  *     localRunner.run(userApp);
  *   }
  *
- *   private void initBlackList(Config config) {
- *     // Read the blacklist from config
- *   }
  * }
  * }</pre>
  *
--- a/samza-api/src/main/java/org/apache/samza/application/StreamApplication.java
+++ b/samza-api/src/main/java/org/apache/samza/application/StreamApplication.java
@@ -51,7 +51,6 @@ import org.apache.samza.operators.StreamGraph;
  *     CommandLine cmdLine = new CommandLine();
  *     Config config = cmdLine.loadConfig(cmdLine.parser().parse(args));
  *     PageViewCounterExample userApp = new PageViewCounterExample();
- *     userApp.initBlackList(config);
  *     ApplicationRunner localRunner = ApplicationRunner.getLocalRunner(config);
  *     localRunner.run(userApp);
  *   }
--- a/samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
+++ b/samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
@@ -47,20 +47,20 @@ public interface MessageStream<M> {
    * transformed {@link MessageStream}.
    *
    * @param mapFn the function to transform a message to another message
-   * @param <OM> the type of messages in the output {@link MessageStream}
+   * @param <TM> the type of messages in the transformed {@link MessageStream}
    * @return the transformed {@link MessageStream}
    */
-  <OM> MessageStream<OM> map(MapFunction<? super M, ? extends OM> mapFn);
+  <TM> MessageStream<TM> map(MapFunction<? super M, ? extends TM> mapFn);
 
   /**
    * Applies the provided 1:n function to transform a message in this {@link MessageStream}
    * to n messages in the transformed {@link MessageStream}
    *
    * @param flatMapFn the function to transform a message to zero or more messages
-   * @param <OM> the type of messages in the output {@link MessageStream}
+   * @param <TM> the type of messages in the transformed {@link MessageStream}
    * @return the transformed {@link MessageStream}
    */
-  <OM> MessageStream<OM> flatMap(FlatMapFunction<? super M, ? extends OM> flatMapFn);
+  <TM> MessageStream<TM> flatMap(FlatMapFunction<? super M, ? extends TM> flatMapFn);
 
   /**
    * Applies the provided function to messages in this {@link MessageStream} and returns the
@@ -82,9 +82,8 @@ public interface MessageStream<M> {
    * non-stream systems (e.g., an external database).
    *
    * @param sinkFn the function to send messages in this stream to an external system
-   * @return this {@link MessageStream} object
    */
-  MessageStream<M> sink(SinkFunction<? super M> sinkFn);
+  void sink(SinkFunction<? super M> sinkFn);
 
   /**
    * Allows sending messages in this {@link MessageStream} to an output {@link MessageStream}.
@@ -92,9 +91,8 @@ public interface MessageStream<M> {
    * @param outputStream the output stream to send messages to
    * @param <K> the type of key in the outgoing message
    * @param <V> the type of message in the outgoing message
-   * @return this {@link MessageStream} object
    */
-  <K, V> MessageStream<M> sendTo(OutputStream<K, V, M> outputStream);
+  <K, V> void sendTo(OutputStream<K, V, M> outputStream);
 
   /**
    * Groups the messages in this {@link MessageStream} according to the provided {@link Window} semantics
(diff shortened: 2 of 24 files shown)


samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -82,17 +82,19 @@
    * non-stream systems (e.g., an external database).
    *
    * @param sinkFn the function to send messages in this stream to an external system
+   * @return this {@link MessageStream} object
```
jmakes: This is not as detailed as I would need if I were the user. 

Is "this" stream before or after the external stream? Are there any concerns that should be taken into account when using it for chaining? What if I want to consume the external stream, say, for modularity/testing purposes?

Same comment for the return param of sink() below.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -47,20 +47,20 @@
...
    * Applies the provided 1:n function to transform a message in this {@link MessageStream}
    * to n messages in the transformed {@link MessageStream}
    *
    * @param flatMapFn the function to transform a message to zero or more messages
-   * @param <TM> the type of messages in the transformed {@link MessageStream}
+   * @param <OM> the type of messages in the output {@link MessageStream}
    * @return the transformed {@link MessageStream}
    */
-  <TM> MessageStream<TM> flatMap(FlatMapFunction<M, TM> flatMapFn);
+  <OM> MessageStream<OM> flatMap(FlatMapFunction<? super M, ? extends OM> flatMapFn);
```
prateekm: Issue: Why do these functions need `? super M` as the input type? Seems like this is implying that we can provide any super-class of M (which is effectively Object) as the input. That doesn't seem right. Should this be `? extends M`?

Also, shouldn't the return type also be `? extends OM` for the operators themselves?

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
prateekm: Documentation: "object" is redundant.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -82,17 +82,19 @@
    * non-stream systems (e.g., an external database).
    *
    * @param sinkFn the function to send messages in this stream to an external system
+   * @return this {@link MessageStream} object
    */
-  void sink(SinkFunction<M> sinkFn);
+  MessageStream<M> sink(SinkFunction<? super M> sinkFn);
```
prateekm: Issue: I think we decided to make these terminal because sink and sendTo can repartition the output stream while this returns the un-repartitioned stream.

samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
```diff
@@ -119,11 +121,21 @@ private Windows() { }
...
    * @return the created {@link Window} function.
    */
-  public static <M, K, WV> Window<M, K, WV> keyedTumblingWindow(Function<M, K> keyFn, Duration interval,
-                                                                Supplier<WV> initialValue, FoldLeftFunction<M, WV> foldFn) {
+  public static <M, K, WV> Window<M, K, WV> keyedTumblingWindow(Function<? super M, ? extends K> keyFn, Duration interval,
+                                                                Supplier<? extends WV> initialValue, FoldLeftFunction<? super M, WV> foldFn) {
 
     Trigger<M> defaultTrigger = new TimeTrigger<>(interval);
-    return new WindowInternal<M, K, WV>(defaultTrigger, initialValue, foldFn, keyFn, null, WindowType.TUMBLING);
+    return new WindowInternal<>(defaultTrigger, initialValue::get, new FoldLeftFunction<M, WV>() {
```
prateekm: Issue: Why create wrapper functions here?

samza-api/src/main/java/org/apache/samza/task/TaskContext.java
```diff
@@ -58,10 +58,9 @@
   /**
    * Method to allow user to return customized context
```
prateekm: Documentation: Shouldn't refer to user in third person in public APIs. Also, "Method to" is redundant.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -61,16 +61,22 @@ public StreamGraphImpl(ApplicationRunner runner, Config config) {
...
-        streamSpec -> new InputStreamInternalImpl<>(this, streamSpec, msgBuilder));
+        streamSpec -> new InputStreamInternalImpl<K, V, M>(this, streamSpec, msgBuilder));
   }
 
   @Override
   public <K, V, M> OutputStream<K, V, M> getOutputStream(String streamId,
-      Function<M, K> keyExtractor, Function<M, V> msgExtractor) {
+      Function<? super M, ? extends K> keyExtractor, Function<? super M, ? extends V> msgExtractor) {
+    if (keyExtractor == null || msgExtractor == null) {
+      throw new IllegalArgumentException("Neither keyExtractor nor msgExtractor can be null for an output stream.");
```
prateekm: Thanks for adding these. 
Issue: Should specify which is null.

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpecs.java
```diff
@@ -115,8 +117,20 @@ public void init(Config config, TaskContext context) {
    * @return  the {@link StreamOperatorSpec}
    */
   public static <M, OM> StreamOperatorSpec<M, OM> createStreamOperatorSpec(
-      FlatMapFunction<M, OM> transformFn, MessageStreamImpl<OM> nextStream, int opId) {
-    return new StreamOperatorSpec<>(transformFn, nextStream, OperatorSpec.OpCode.FLAT_MAP, opId);
+      FlatMapFunction<? super M, ? extends OM> transformFn, MessageStreamImpl<OM> nextStream, int opId) {
+    return new StreamOperatorSpec<>(new FlatMapFunction<M, OM>() {
```
prateekm: Issue: Why create wrapper functions?

samza-core/src/main/java/org/apache/samza/operators/stream/IntermediateStreamInternalImpl.java
```diff
@@ -33,13 +33,13 @@
...
-  public IntermediateStreamInternalImpl(StreamGraphImpl graph, StreamSpec streamSpec,
-      Function<M, K> keyExtractor, Function<M, V> msgExtractor, BiFunction<K, V, M> msgBuilder) {
+  public IntermediateStreamInternalImpl(StreamGraphImpl graph, StreamSpec streamSpec, Function<? super M, ? extends K> keyExtractor,
+      Function<? super M, ? extends V> msgExtractor, BiFunction<? super K, ? super V, ? extends M> msgBuilder) {
     super(graph);
     this.streamSpec = streamSpec;
-    this.keyExtractor = keyExtractor;
-    this.msgExtractor = msgExtractor;
-    this.msgBuilder = msgBuilder;
+    this.keyExtractor = keyExtractor::apply;
```
prateekm: Issue: Seems unusual, this is already a functional interface . Why is this required?

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
nickpan47: This means before the external stream. For consuming from the external stream, it should either be partitionBy or sendThrough.

samza-api/src/main/java/org/apache/samza/application/StreamApplication.java
```diff
@@ -24,11 +24,46 @@
 
 
 /**
- * This interface defines a template for stream application that user will implement to create operator DAG in {@link StreamGraph}.
+ * This interface defines a template for stream application that user will implement to initialize operator DAG in {@link StreamGraph}.
```
nickpan47: I always thought that the javadoc needs to refer to the readers as the third person. No strong opinion on this though. Only one thing that if we want to make this call, let's be consistent across the code base.

samza-api/src/main/java/org/apache/samza/application/StreamApplication.java
```diff
@@ -24,11 +24,46 @@
...
+ *   public static void main(String[] args) {
+ *     CommandLine cmdLine = new CommandLine();
+ *     Config config = cmdLine.loadConfig(cmdLine.parser().parse(args));
+ *     PageViewCounterExample userApp = new PageViewCounterExample();
+ *     userApp.initBlackList(config);
+ *     ApplicationRunner localRunner = ApplicationRunner.getLocalRunner(config);
+ *     localRunner.run(userApp);
+ *   }
+ *
+ *   private void initBlackList(Config config) {
```
nickpan47: Let me change the example s.t. it does not include the init phase.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -47,20 +47,20 @@
    * transformed {@link MessageStream}.
    *
    * @param mapFn the function to transform a message to another message
-   * @param <TM> the type of messages in the transformed {@link MessageStream}
+   * @param <OM> the type of messages in the output {@link MessageStream}
```
nickpan47: Sure. I will leave the type parameter name as TM as well then.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
nickpan47: Refer to out discussion earlier on PR #51 , the whole purpose of this PR is to relax the function type parameter, not the MessageStream type parameter. The purpose is to relax the restriction for user to provide their own functions to be applied to the MessageStream&lt;M&gt;. Hence, here are the basic rules I applied in this change:
1. No type change to the types we define (i.e. MessageStream, StreamGraph, OperatorSpec, Window, etc)
2. Relaxed the type parameters of functions user supplies s.t. the user supplied functions can be applied to the message types in MessageStream.

That's why {code}? super M{code} and {code}? extends M{code} are defined as the type parameters for the functions, not to the MessageStream class. Restricting the type to M totally defeat the whole purpose of this PR.

samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
nickpan47: This is to formulate the type of Function that our internal class will depend on. Leaking the captured type parameter to internal functions leads to a whole lot changes over the full implementation stack, which does not seem to be necessary for the purpose (i.e. just allow relaxed type parameter from user-supplied functions).

samza-api/src/main/java/org/apache/samza/task/TaskContext.java
nickpan47: +1 if you want to make a full scale blanket change on this.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
nickpan47: Will add to the error message.

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpecs.java
nickpan47: Same answer as earlier.

samza-core/src/main/java/org/apache/samza/operators/stream/IntermediateStreamInternalImpl.java
nickpan47: Same as the reason for wrapper function.

samza-api/src/main/java/org/apache/samza/task/TaskContext.java
prateekm: Will do. Only mentioned this here since TaskContext isn't part of the operators package so will be easy to miss.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
prateekm: I think it'll be safer and more clearly explainable to leave them as terminal. Users can still chain explicitly on the previous message stream if they wish to. Don't feel strongly about it, up to you.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
prateekm: @nickpan47 explained offline. The way to read this is that the user can provide a function that takes a super-type of M as input since it's always safe to cast the actual provided M to its super-type.

E.g., if M == Integer, this signature allows users to provide a more general FilterFunction<Number> instead of a FilterFunction<Integer> if that's what they want to use.

samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
nickpan47: Talked offline with @prateekm  and we agreed that it should be safe to simply do a type-cast here. Will make the change in several places with the same wrapper.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
nickpan47: Sold on this one. Will keep sink / sendTo terminal. The implementation in PR is more like a side output. Maybe we can add later as sideSink()/sideSend(), if more and more use cases popping up.

samza-core/src/main/java/org/apache/samza/operators/stream/IntermediateStreamInternalImpl.java
```diff
@@ -33,13 +33,13 @@
   private final Function<M, V> msgExtractor;
   private final BiFunction<K, V, M> msgBuilder;
 
-  public IntermediateStreamInternalImpl(StreamGraphImpl graph, StreamSpec streamSpec,
-      Function<M, K> keyExtractor, Function<M, V> msgExtractor, BiFunction<K, V, M> msgBuilder) {
+  public IntermediateStreamInternalImpl(StreamGraphImpl graph, StreamSpec streamSpec, Function<? super M, ? extends K> keyExtractor,
```
prateekm: A little cleaner to do these casts in StreamGraph and leave these types as K, V, M. Both here and InputStreamInternalImpl.

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpecs.java
```diff
@@ -53,7 +53,7 @@ private OperatorSpecs() {}
    * @return  the {@link StreamOperatorSpec}
    */
   public static <M, OM> StreamOperatorSpec<M, OM> createMapOperatorSpec(
-      MapFunction<M, OM> mapFn, MessageStreamImpl<OM> nextStream, int opId) {
+      MapFunction<? super M, ? extends OM> mapFn, MessageStreamImpl<OM> nextStream, int opId) {
```
prateekm: Will be a little cleaner to do these casts in MessageStreamImpl and leave these as K, V, M

samza-api/src/main/java/org/apache/samza/application/StreamApplication.java
```diff
@@ -24,11 +24,44 @@
...
+ *       .filter(m -> !(m.getMessage().getEventTime() < System.currentTimeMillis() - MAX_TIMEOUT))
+ *       .sendTo(pageViewEventFilteredStream);
+ *   }
+ *
+ *   // local execution mode
+ *   public static void main(String[] args) {
+ *     CommandLine cmdLine = new CommandLine();
+ *     Config config = cmdLine.loadConfig(cmdLine.parser().parse(args));
+ *     PageViewCounterExample userApp = new PageViewCounterExample();
+ *     userApp.initBlackList(config);
```
navina: What does userApp.initBlackList(config) mean?

nickpan47: Urh... Good catch!!! That should be gone. I am removing it and committing.

Thanks!

PR #134, opened 2017-04-20
SAMZA-1229: Disk space monitor should only count data in use by the container

prateekm:


prateekm:
@santhoshvenkat1988 @vjagadish1989 Please take a look.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -455,8 +454,6 @@ object SamzaContainer extends Logging {
         loggedStorageBaseDir = defaultStoreBaseDir
       }
 
-      storeWatchPaths.add(loggedStorageBaseDir.toPath)
```
xinyuiscool: Just for my understanding: previous it is counting all the task stores for an application on a node?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -467,25 +464,30 @@ object SamzaContainer extends Logging {
...
+
+            val storeDir = if (changeLogSystemStreamPartition != null) {
               TaskStorageManager.getStorePartitionDir(loggedStorageBaseDir, storeName, taskName)
-            }
-            else {
+            } else {
               TaskStorageManager.getStorePartitionDir(defaultStoreBaseDir, storeName, taskName)
             }
+
+            storeWatchPaths.add(storeDir.toPath)
```
xinyuiscool: My guess this is adding per task store?

vjagadish1989 (approved):
approved. looks good. thanks for the contribution!

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
prateekm: Yeah, it will counting all stores for the job on the host.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
prateekm: Exactly, now it adds only the stores currently used by the tasks in the container.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
prateekm: Here's a sample log message that shows the current paths: 

```Initialized disk space monitor watch paths to: [/export/content/data/samsa-yarn/samza-logged-stores/job_name-job_instance, /mnt/u001/samsa-yarn/usercache/mp_name/appcache/application_1491433610021_0266/container_e01_1491433610021_0266_02_000002/state]```

PR #135, opened 2017-04-20
SAMZA-1222: Clean up LocalApplicationRunner

xinyuiscool:
Clean up the LocalApplicationRunner based on the further feedback. The changes include the following:
1. Remove the processorId from the JobCoordinatorFactory/JobCoordinator interfaces
2. LocalApplicationRunner.run() is non-blocking. Add LocalApplicationRunner.waitForFinish() for blocking for completion
3. Remove the config for CooridnatorServiceFactory, and now the CoordinatorService is created based on the type of JobCoordinator.
4. Clean up the StreamProcessor life cycle listener logic inside LocalApplicationRunner.


xinyuiscool:
@prateekm : update on the LocalApplicationRunner based on your further comments in the previous PR. Please take a look. Thanks!

samza-core/src/main/java/org/apache/samza/coordinator/JobCoordinatorFactory.java
```diff
@@ -26,11 +26,12 @@
...
   /**
+   * Return a new instance of {@link JobCoordinator}
    * @param processorId {@link org.apache.samza.processor.StreamProcessor} id
    * @param config Configs relevant for the JobCoordinator TODO: Separate JC related configs into a "JobCoordinatorConfig"
    * @param containerController Controller interface for starting and stopping container. In future, it may simply
    *                            pause the container and add/remove tasks
-   * @return An instance of IJobCoordinator
+   * @return {@link JobCoordinator} instance
    */
   JobCoordinator getJobCoordinator(String processorId, Config config, SamzaContainerController containerController);
```
prateekm: The processorId parameter shouldn't be here, or in the StreamProcessor constructor. Similarly, LocalApplicationRunner shouldn't be using ProcessorIdGenerator itself (see explanation in previous PR). Let me know if that doesn't make sense.

samza-core/src/main/java/org/apache/samza/coordinator/JobCoordinatorFactory.java
```diff
@@ -26,11 +26,12 @@
 @InterfaceStability.Evolving
 public interface JobCoordinatorFactory {
   /**
+   * Return a new instance of {@link JobCoordinator}
    * @param processorId {@link org.apache.samza.processor.StreamProcessor} id
    * @param config Configs relevant for the JobCoordinator TODO: Separate JC related configs into a "JobCoordinatorConfig"
    * @param containerController Controller interface for starting and stopping container. In future, it may simply
    *                            pause the container and add/remove tasks
-   * @return An instance of IJobCoordinator
+   * @return {@link JobCoordinator} instance
```
prateekm: Thanks!

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -119,24 +122,34 @@ public void run(StreamApplication app) {
...
         });
-      appStatus = ApplicationStatus.Running;
 
       // 4. block until the processors are done or there is a failure
-      awaitComplete();
+      latch.await();
+
+      if (throwable.get() != null) {
+        appStatus = ApplicationStatus.UnsuccessfulFinish;
+        throw throwable.get();
```
prateekm: Wrap in a new exception and throw.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -152,14 +165,15 @@ public ApplicationStatus status(StreamApplication streamApp) {
...
+   * @throws Exception exceptions when creating the instance, such as {@link ClassNotFoundException} and
+   * {@link InstantiationException}
    */
-  /* package private */ CoordinationUtils getCoordinationUtils() throws Exception {
+  /* package private */ CoordinationUtils createCoordinationUtils() throws Exception {
     ApplicationConfig appConfig = new ApplicationConfig(config);
     String clazz = appConfig.getCoordinationServiceFactoryClass();
-    if (clazz != null) {
+    if (StringUtils.isNotEmpty(clazz)) {
       CoordinationServiceFactory factory = ClassLoaderHelper.fromClassName(clazz);
```
prateekm: Not sure why this needs to be a shared/pluggable CoordinatorService. Looks like this is just some utils for working with ZK?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -1,4 +1,4 @@
-  /*
+/*
```
prateekm: Also, we want to move the createStreamProcessor helper to a StreamProcessor constructor.

samza-core/src/main/java/org/apache/samza/coordinator/JobCoordinatorFactory.java
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: As we discussed, let's keep this here for now.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: it will be wrapped in the following catch(throwable).

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: I refactored the code according to our discussion to set coordinatorservice based on JobCooridnator.

xinyuiscool:
@prateekm and @navina : I removed the processorId in the JobCooridnatorFactory/contructor as we discussed. Please take a look again at this clean up patch.

samza-core/src/main/java/org/apache/samza/config/JobCoordinatorConfig.java
```diff
@@ -23,7 +23,6 @@
 
 public class JobCoordinatorConfig extends MapConfig {
   public static final String JOB_COORDINATOR_FACTORY = "job-coordinator.factory";
```
prateekm: @navina Unrelated to this PR, but does this need to be changed to job.coordinator.factory?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
prateekm: Can probably structure like this to consolidate status updates in one place:

```
try {
  startProcessors();
  awaitShutdown();
} catch (Throwable t) {
  throwable.compareAndSet(t, nulll)
} finally {
  if (throwable.get() != null) {
    status = unsuccessful;
    throw new SamzaException(throwable.get())
  } else {
    status = successful;
  }
}
```

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -152,32 +159,37 @@ public ApplicationStatus status(StreamApplication streamApp) {
...
+   * @throws Exception exceptions when creating the instance, such as {@link ClassNotFoundException} and
+   * {@link InstantiationException}
    */
-  /* package private */ CoordinationUtils getCoordinationUtils() throws Exception {
-    ApplicationConfig appConfig = new ApplicationConfig(config);
-    String clazz = appConfig.getCoordinationServiceFactoryClass();
-    if (clazz != null) {
-      CoordinationServiceFactory factory = ClassLoaderHelper.fromClassName(clazz);
+  /* package private */ CoordinationUtils createCoordinationUtils() throws Exception {
+    String jobCoordinatorFactoryClazz = config.get(JobCoordinatorConfig.JOB_COORDINATOR_FACTORY, "");
```
prateekm: Minor: xClazz name is usually used for Class<X> parameters. Maybe jobCoordinatorFactoryClassName instead?

samza-core/src/main/java/org/apache/samza/util/ProcessorIdUtil.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package org.apache.samza.util;
+
+import org.apache.samza.config.ApplicationConfig;
+import org.apache.samza.config.Config;
+import org.apache.samza.config.ConfigException;
+import org.apache.samza.runtime.ProcessorIdGenerator;
+
+public class ProcessorIdUtil {
```
prateekm: Should be in JobCoordinator impl (`getProcessorIdGenerator`).

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -56,44 +55,46 @@
...
-  private static final long LATCH_TIMEOUT_MINUTES = 10; // 10 min timeout
+  // Latch timeout is set to 10 min
+  private static final long LATCH_TIMEOUT_MINUTES = 10;
 
   private final String uid;
-  private final CoordinationUtils coordination;
+  private final CoordinationUtils coordinationUtils;
   private final List<StreamProcessor> processors = new ArrayList<>();
+  private final Set<StreamProcessor> runningProcessors = ConcurrentHashMap.newKeySet();
   private final CountDownLatch latch = new CountDownLatch(1);
```
prateekm: Should rename to describe what this latch is for. SHUTDOWN_LATCH?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -119,24 +120,30 @@ public void run(StreamApplication app) {
...
       appStatus = ApplicationStatus.UnsuccessfulFinish;
       throw new SamzaException("Failed to run application", t);
     } finally {
-      if (coordination != null) {
-        coordination.reset();
+      if (coordinationUtils != null) {
+        coordinationUtils.reset();
       }
     }
   }
```
prateekm: Same question as previous PR: Isn't kill supposed to be blocking in the current version? StreamProcessor stop is non-blocking.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -56,44 +55,46 @@
 public class LocalApplicationRunner extends AbstractApplicationRunner {
 
   private static final Logger log = LoggerFactory.getLogger(LocalApplicationRunner.class);
+  // Latch id that's used for awaiting the init of application before creating the StreamProcessors
   private static final String LATCH_INIT = "init";
```
prateekm: INIT_LATCH_ID?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -56,44 +55,46 @@
...
-  private static final long LATCH_TIMEOUT_MINUTES = 10; // 10 min timeout
+  // Latch timeout is set to 10 min
+  private static final long LATCH_TIMEOUT_MINUTES = 10;
```
prateekm: INIT_LATCH_TIMEOUT_MINUTES? @

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -56,44 +55,46 @@
...
 
   private final String uid;
-  private final CoordinationUtils coordination;
+  private final CoordinationUtils coordinationUtils;
   private final List<StreamProcessor> processors = new ArrayList<>();
```
prateekm: Do we need both processors and runningProcessors?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -56,44 +55,46 @@
...
+      if (runningProcessors.isEmpty()) {
         latch.countDown();
       }
     }
 
     @Override
     public void onFailure(Throwable t) {
-      processorIds.remove(processorId);
+      runningProcessors.remove(processor);
       throwable.compareAndSet(null, t);
```
prateekm: We probably need to stop the remaining runningProcessors somewhere, right?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -119,24 +120,30 @@ public void run(StreamApplication app) {
...
       plan.getJobConfigs().forEach(jobConfig -> {
-          log.info("Starting job {} StreamProcessor with config {}", jobConfig.getName(), jobConfig);
-          String processorId = getProcessorId(config);
-          StreamProcessor processor =
-              createStreamProcessor(processorId, jobConfig, app, new LocalStreamProcessorListener(processorId));
-          processor.start();
+          log.debug("Starting job {} StreamProcessor with config {}", jobConfig.getName(), jobConfig);
+          LocalStreamProcessorLifeCycleListener listener = new LocalStreamProcessorLifeCycleListener();
+          StreamProcessor processor = createStreamProcessor(jobConfig, app, listener);
+          listener.setProcessor(processor);
```
navina: If you know the number of streamprocessors that will be spawned, can you simply maintain a countdown latch with that count, decrement it in the lifecycle callbacks and then, await on it here? 

Setting processor reference inside the listener feels unclean for me. Using startLatch and endLatch may be cleaner. Let me know if you already tried it and didn't work out for some reason.

samza-core/src/main/java/org/apache/samza/config/JobCoordinatorConfig.java
navina: I created a separate JIRA for it - https://issues.apache.org/jira/browse/SAMZA-1224 This is a good reminder :)
I will put a patch for that config today. let's not address it in this PR.

samza-core/src/main/java/org/apache/samza/util/ProcessorIdUtil.java
navina: +1 for @prateekm 's suggestion. 
Creating a public util class is dangerous as it is now open for any component to use it. This will additionally impose a requirement on the implementing class of the interface that multiple calls to generateProcessorId should return the same identifier. This is a much stronger constraint on the interface. Also, if the generator is class-loaded each time, then it cannot uphold the contract. 

Let's make sure that ProcessorIdGenerator is used *only* by the JobCoordinator. Thanks!

navina: @prateekm Does it make sense to make getProcessorId a default method in JobCoordinator interface? Just a thought.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
sborya: +1 on Prateek's comment.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: I refactored it so now I am setting the appstatus in the callback directly. Please take a look again.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: Now they are both non-blocking :)

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/util/ProcessorIdUtil.java
xinyuiscool: I like this idea a lot! Added the default method in JobCoordinator.

samza-core/src/main/java/org/apache/samza/coordinator/JobCoordinator.java
```diff
@@ -86,4 +82,29 @@
    * @return instance of JobModel that describes the partition distribution among the processors (and hence, tasks)
    */
   JobModel getJobModel();
+
+  /**
+   * Returns the identifier assigned to the processor that is local to the instance of StreamProcessor.
```
navina: Can you document a line here about what is expected of other JobCoordinator implementations?  When it should be overriden?

samza-core/src/main/java/org/apache/samza/coordinator/JobCoordinator.java
```diff
@@ -86,4 +82,29 @@
...
+   *
+   * The semantics and format of the identifier returned should adhere to the specification defined in
+   * {@link org.apache.samza.runtime.ProcessorIdGenerator}
+   *
+   * @param config Job {@link Config}
+   * @return String representing a unique logical processor ID
+   */
+  default String getProcessorId(Config config) {
```
navina: Let's document the ConfigException in the doc here

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -152,32 +165,47 @@ public ApplicationStatus status(StreamApplication streamApp) {
...
+   * @param intStreams list of intermediate {@link StreamSpec}s
+   * @throws Exception exception for latch timeout
    */
   /* package private */ void createStreams(List<StreamSpec> intStreams) throws Exception {
     if (!intStreams.isEmpty()) {
-      if (coordination != null) {
-        Latch initLatch = coordination.getLatch(1, LATCH_INIT);
-        coordination.getLeaderElector().tryBecomeLeader(() -> {
+      if (coordinationUtils != null) {
+        Latch initLatch = coordinationUtils.getLatch(1, INIT_LATCH_ID);
```
navina: What does the INIT_LATCH_ID represent here? Shouldn't it be associated within the application's ZK subtree??

navina: Also, can we rename the variable from "latch" to "sharedLatch" or "distributedLatch"? It is easy for the reader to miss that it is not a local countdown latch.

samza-core/src/main/java/org/apache/samza/util/ProcessorIdUtil.java
prateekm: @navina @xinyuiscool I'd strongly prefer not to, for the same reason I mentioned in SEP-1:
"I don't yet see the value of putting the class-loading helper default methods in multiple public interfaces. Seems like they're (usually) going to be used by the framework, are pretty simple to write, and could probably be written as a common Util method if we find them repetitive."

Also, the current implementation has the same issue @navina  pointed out earlier - it's classloading the generator for every request which is not right.

samza-core/src/main/java/org/apache/samza/coordinator/JobCoordinator.java
```diff
@@ -19,7 +19,13 @@
 package org.apache.samza.coordinator;
 
 import org.apache.samza.annotation.InterfaceStability;
+import org.apache.samza.config.ApplicationConfig;
+import org.apache.samza.config.Config;
+import org.apache.samza.config.ConfigException;
 import org.apache.samza.job.model.JobModel;
+import org.apache.samza.runtime.ProcessorIdGenerator;
+import org.apache.samza.util.*;
```
prateekm: Import individually.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -56,53 +54,75 @@
...
+      }
     }
 
     @Override
     public void onShutdown() {
-      processorIds.remove(processorId);
-      if (processorIds.isEmpty()) {
-        latch.countDown();
+      processors.remove(processor);
+      processor = null;
```
prateekm: Is this required (here and in onFailure)? Seems like the reference will go away when the listener reference is removed from the set.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -56,53 +54,75 @@
...
-      processorIds.remove(processorId);
-      throwable.compareAndSet(null, t);
-      latch.countDown();
+      processors.remove(processor);
+      processor = null;
+
+      appStatus = ApplicationStatus.unsuccessfulFinish(t);
+      if (!processors.isEmpty()) {
+        // shut down the other processors
+        processors.forEach(StreamProcessor::stop);
```
prateekm: This will invoke stop on a StreamProcessor multiple times if more than one processor fails. @navina is that ok?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -152,32 +165,47 @@ public ApplicationStatus status(StreamApplication streamApp) {
     return appStatus;
   }
 
+  @Override
+  public void waitForFinish() {
+    try {
+      shutdownLatch.await();
+    } catch (Throwable t) {
```
prateekm: Catch exception.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -129,4 +129,8 @@ public ApplicationStatus status(StreamApplication app) {
     }
   }
 
+  @Override
+  public void waitForFinish() {
```
prateekm: Maybe just move this method to LocalApplicationRunner instead of the AbstractApplicationRunner? No other runners implement this.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -115,30 +135,23 @@ public void run(StreamApplication app) {
...
     } catch (Throwable t) {
-      appStatus = ApplicationStatus.unsuccessfulFinish(t);
-      throw new SamzaException("Failed to run application", t);
-    } finally {
-      if (coordination != null) {
-        coordination.reset();
-      }
+      throw new SamzaException("Failed to start application", t);
     }
   }
```
prateekm: Add javadoc for kill here to explain that this is non-blocking. Same for run.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -115,30 +135,23 @@ public void run(StreamApplication app) {
...
         });
-      appStatus = ApplicationStatus.Running;
-
-      // 4. block until the processors are done or there is a failure
-      awaitComplete();
+      numProcessorsToStart.set(processors.size());
 
+      // 4. start the StreamProcessors
+      processors.forEach(StreamProcessor::start);
     } catch (Throwable t) {
```
prateekm: Catch exception.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
xinyuiscool: makes sense. Let's keep it in LocalApplicationRunner.

samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
```diff
@@ -81,14 +81,16 @@ public ApplicationRunner(Config config) {
...
+   * Deploy and run the Samza jobs to execute {@link StreamApplication}.
+   * It is non-blocking so it doesn't wait for the application running.
    *
    * @param streamApp  the user-defined {@link StreamApplication} object
    */
   public abstract void run(StreamApplication streamApp);
 
   /**
    * Kill the Samza jobs represented by {@link StreamApplication}
+   * It is non-blocking so it doesn't wait for teh application stopping.
```
navina: nit: typo in "teh"

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -153,31 +166,46 @@ public ApplicationStatus status(StreamApplication streamApp) {
...
+   * @return an instance of {@link CoordinationUtils}
    */
-  /* package private */ CoordinationUtils getCoordinationUtils() throws Exception {
-    ApplicationConfig appConfig = new ApplicationConfig(config);
-    String clazz = appConfig.getCoordinationServiceFactoryClass();
-    if (clazz != null) {
-      CoordinationServiceFactory factory = ClassLoaderHelper.fromClassName(clazz);
-      return factory.getCoordinationService(appConfig.getGlobalAppId(), uid, config);
+  /* package private */ CoordinationUtils createCoordinationUtils() {
+    String jobCoordinatorFactoryClassName = config.get(JobCoordinatorConfig.JOB_COORDINATOR_FACTORY, "");
```
navina: Any reason for not using `getJobCoordinatorFactoryClassName()` in `JobCoordinatorConfig.java`?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -153,31 +166,46 @@ public ApplicationStatus status(StreamApplication streamApp) {
...
+      shutdownLatch.await();
+    } catch (Exception e) {
+      log.error("Wait is interrupted by exception", e);
+      throw new SamzaException(e);
+    }
+  }
+
+  /**
+   * Create the {@link CoordinationUtils} needed by the application runner.
+   * @return an instance of {@link CoordinationUtils}
```
navina: nit: update return annotation to include the possibility of returning "null" value

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -153,31 +166,46 @@ public ApplicationStatus status(StreamApplication streamApp) {
...
       return null;
     }
   }
 
   /**
-   * Create intermediate streams.
-   * @param intStreams intermediate {@link StreamSpec}s
-   * @throws Exception exception
+   * Create intermediate streams using {@link org.apache.samza.execution.StreamManager}.
+   * If {@link CoordinationUtils} is provided, this function will first invoke leader election, Pand the leader
```
navina: nit: typo in "Pand the leader"

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
navina: @xinyuiscool I wasn't referring to the method name. I just wanted you to change the local variable name here as distributedLatch or sharedLatch. Can we still do that here and follow-up with refactoring and renaming in SAMZA-1262? It is difficult for me to read the code, otherwise.

navina: > The name of Latch seems too general. 

Does the name of the latch signify anything in the implementation? I assumed it will be within the application hierarchy in ZK. is my assumption correct?

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
```diff
@@ -153,26 +133,10 @@ public void testStreamProcessorWithAsyncStreamTaskFactory() {
...
-          public void onShutdown() {
-
-          }
-
-          @Override
-          public void onFailure(Throwable t) {
-
-          }
-        });
+        listener);
```
navina: Thanks for cleaning this up :)

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
navina: I think it should be fine. It might invoke jc.stop twice. But I think the current implementations of JC don't cause any state change or failure on invoking stop multiple times.. So we should be fine, **I think**

samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
xinyuiscool: Good catch. Fixed it.

navina:
@xinyuiscool since our merge tool has a feature gap, can you update the description in the PR to indicate the changes you made? Like "Removed processorId from JC interface, Adding awaitForFinish method in Application Runner" etc.. The title itself is pretty vague - "Clean up ApplicationRunner". It will be useful to have the change description in the commit log. Thanks!

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: I added the logic to make sure it will call the shutdown only once. Basically first time the listener will set the atomic reference of the throwable, so it won't cascading the shutdown later. Please take a look at the code again.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: I remembered I got that config exception once, couldn't remember what's the case anymore. Seems it should always be there.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: Thanks for the catch!

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
xinyuiscool: yup, exactly. The name is a zk node under the app. Providing the name helps if I need to create other latches in the future so they can be named differently.

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
xinyuiscool: np at all.

xinyuiscool:
Sure. I updated the description in both pr and the JIRA ticket.

PR #138, opened 2017-04-24
SAMZA-1233: Create SystemAdmin only for JobCoordinator System in SamzaTaskProxy

shanthoosh:


shanthoosh:
@jmakes Please take a look when you get a chance.

samza-rest/src/main/java/org/apache/samza/rest/proxy/task/SamzaTaskProxy.java
```diff
@@ -155,8 +156,16 @@ protected JobModel getJobModel(JobInstance jobInstance) {
       changelogManager.start();
       LocalityManager localityManager = new LocalityManager(coordinatorSystemProducer, coordinatorSystemConsumer);
       localityManager.start();
-      return JobModelManager.readJobModel(config, changelogManager.readChangeLogPartitionMapping(), localityManager,
-                                          new StreamMetadataCache(JobModelManager.getSystemAdmins(config), 0, SystemClock.instance()), null);
+
+      String jobCoordinatorSystemName = config.get(JobConfig.JOB_COORDINATOR_SYSTEM());
+
+      /**
+       Select job coordinator system properties from config and instantiate SystemAdmin for it alone.
```
prateekm: Add '*' on each javadoc line.

samza-rest/src/main/java/org/apache/samza/rest/proxy/task/SamzaTaskProxy.java
```diff
@@ -155,8 +156,16 @@ protected JobModel getJobModel(JobInstance jobInstance) {
...
+       Instantiating SystemAdmin's for other input/output systems defined in config is unnecessary.
+      */
+      Config systemAdminConfig = config.subset(String.format("systems.%s", jobCoordinatorSystemName), false);
+      StreamMetadataCache streamMetadataCache = new StreamMetadataCache(JobModelManager.getSystemAdmins(systemAdminConfig), 0, SystemClock.instance());
```
prateekm: Let's not inline side-effecty/meaty method calls in parameters. Same in next line for readChaneLogPartitionMapping.

samza-rest/src/main/java/org/apache/samza/rest/proxy/task/SamzaTaskProxy.java
shanthoosh: Fixed.

samza-rest/src/main/java/org/apache/samza/rest/proxy/task/SamzaTaskProxy.java
shanthoosh: Moved method calls into separate lines. For methods with side-effects this makes sense.

Just curious, reason for chaining is not create unnecessary temporary variables who scope go away after that line(making it more verbose).

Please share your thoughts.

PR #139, opened 2017-04-25
SAMZA-1220 : Add thread name to SamzaContainer shutdown hook and prevent shutdown deadlock

navina:
* SamzaContainerExceptionHandler is written in Java and used by LocalContainerRunner.java


navina:
@prateekm Can you please take a look at the changes? Please comment if this was not what you had in mind. Thanks!

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -803,21 +806,40 @@ class SamzaContainer(
...
         }
-        runLoopThread.join(shutdownMs)
-        if (runLoopThread.isAlive) {
-          warn("Did not shut down within %s ms, exiting" format shutdownMs)
-        } else {
-          info("Shutdown complete")
+        try {
+          runLoopThread.join(shutdownMs)
+        } catch {
+          case e: Exception => // Ignore to avoid deadlock with uncaughtExceptioHandler. See SAMZA-1220
```
vjagadish1989: IMHO it makes sense to catch, log and ignore all `Throwable` here.

- OOM errors happening in shutdown hook are usually in-actionable.
- We will bound the wait time by `shutdownMs` anyways.
- Other throwables happening in the `shutdownHook` will end up calling the `UncaughtExceptionHandler` resulting in a potential deadlock. Catching all `Throwable`s will prevent this.

samza-core/src/main/java/org/apache/samza/container/SamzaContainerExceptionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+
+package org.apache.samza.container;
+
+import java.lang.Thread.UncaughtExceptionHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * An UncaughtExceptionHandler for SamzaContainer that simply shuts down when any thread throws
+ * an uncaught exception.
```
navina: I think this is a result of copying the documentation over from the scala class. I will fix the documentation.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -801,21 +804,40 @@ class SamzaContainer(
...
+  }
+
+  def removeShutdownHook = {
+    try {
+      if (shutdownHookThread != null) {
+        Runtime.getRuntime.removeShutdownHook(shutdownHookThread)
+      }
+    } catch {
+      case e: IllegalStateException => {
+        // When samza is shutdown by external command, IllegalStateException will be thrown.
```
prateekm: Maybe rephrase to:
"Thrown when then JVM is already shutting down, so safe to ignore."?

samza-core/src/test/java/org/apache/samza/container/TestSamzaContainerExceptionHandler.java
```diff
@@ -17,20 +17,23 @@
...
-package org.apache.samza.container
+package org.apache.samza.container;
 
-import org.junit.Test
-import org.junit.Assert._
-import org.junit.Before
-import org.apache.samza.SamzaException
-import org.junit.After
+import org.apache.samza.SamzaException;
+import org.junit.Test;
```
navina: That is how IntelliJ organizes imports. Were you referring some specific ordering?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
navina: Sure. Makes sense.

samza-core/src/main/java/org/apache/samza/container/SamzaContainerExceptionHandler.java
prateekm: +1 if you want to remove the Runnable parameter.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
prateekm: +1 for catching Throwable. Maybe also move the logging to try instead of finally (in case it throws).

samza-core/src/test/java/org/apache/samza/container/TestSamzaContainerExceptionHandler.java
vjagadish1989: Makes sense, I thought the `junit` imports could have been potentially grouped together. I think I may have missed the `static` clause in the `assertTrue`.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
navina: Ok.

vjagadish1989 (approved):
approved. looks great to me. thank you for the contribution navina.

PR #141, opened 2017-04-26
Samza 1214: Allow users to set a default replication.factor for intermediate topics

jmakes:
* Add a new "systems.sysName.default.stream.*" config structure that allows users to set system-wide defaults for streams.
* More thorough testing of system defaults and stream defaults
* Removed the old migration config from the config table since there's no code to support it.
* Moved 2 kafka-specific config accessors out of JobConfig and into KafkaConfig
* Removed duplicate impl of getChangelogStream()


jmakes:
@vjagadish1989 here's the fix for setting replication factor for all intermediate streams. LMK what you think.

samza-kafka/src/test/scala/org/apache/samza/config/TestKafkaConfig.scala
```diff
@@ -195,25 +195,136 @@ class TestKafkaConfig {
...
+    val jobConfig = new JobConfig(mapConfig)
+    assertEquals("kafka-system", jobConfig.getCoordinatorSystemNameOrNull)
+    assertEquals("4", kafkaConfig.getCoordinatorReplicationFactor)
+  }
+
+  @Test
+  def testCoordinatorReplicationFactorWithSystemDefault() {
+    props.setProperty(JobConfig.JOB_DEFAULT_SYSTEM, "kafka-system-39-not38-not40-39")
+    props.setProperty("systems.kafka-system-39-not38-not40-39.default.stream.replication.factor", "8")
+    props.setProperty("systems.kafka-system-39-not38-not40-39.default.stream.segment.bytes", "8675309")
```
vjagadish1989: nit: 
what does the `kafka-system-39-not38-not40-39 ` test? Is there parsing logic that relies on a lot of hyphens? If not, a shorter `system name` might be nicer :-)

vjagadish1989 (commented):
looks good to me. 

Mostly minor feedback:

Nice work adding a thorough, comprehensive suite of unit tests! 👍 

If some suggestions are beyond the scope of this software release, we can defer them to a subsequent release. But, would be nice to evaluate their feasibility regardless.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -82,11 +87,40 @@ object KafkaConfig {
 
 class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
   // checkpoints
-  def getCheckpointSystem = getOption(KafkaConfig.CHECKPOINT_SYSTEM)
+  def getCheckpointSystem = Option(getOrElse(KafkaConfig.CHECKPOINT_SYSTEM, new JobConfig(config).getDefaultSystem.orNull))
```
jmakes: I noticed this too. Here's what I found. 
Checkpointing was originally designed to be either file-based or stream-based, and there was only 1 factory provided for each. 

If you use the file checkpoint factory, you have to also specify the property
```task.checkpoint.path```

If you use the kafka checkpoint factory, you have to also specify the property
```task.checkpoint.system```

So to support any general System for checkpointing we would need to provide a new factory and move this property and I felt that was outside the scope of this fix.

File: samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala [hunk 3]
```diff
@@ -82,11 +87,40 @@ object KafkaConfig {
...
+  }
 
-  def getCheckpointSegmentBytes() = getInt(KafkaConfig.CHECKPOINT_SEGMENT_BYTES, KafkaConfig.DEFAULT_CHECKPOINT_SEGMENT_BYTES)
+  private def getSystemDefaultReplicationFactor(systemName: String, defaultValue: String) = {
+    val defaultReplicationFactor = new JavaSystemConfig(config).getDefaultStreamProperties(systemName).getOrDefault(KafkaConfig.TOPIC_REPLICATION_FACTOR, defaultValue)
+    defaultReplicationFactor
+  }
+
+  def getCheckpointSegmentBytes() = {
+    val defaultsegBytes = new JavaSystemConfig(config).getDefaultStreamProperties(getCheckpointSystem.orNull).getInt(KafkaConfig.SEGMENT_BYTES, KafkaConfig.DEFAULT_CHECKPOINT_SEGMENT_BYTES)
```
jmakes: The current tests verify this, as it was the source of early failures that I had to fix :-)

samza-core/src/main/java/org/apache/samza/config/JavaStorageConfig.java
```diff
@@ -85,4 +85,8 @@ public String getStorageKeySerde(String storeName) {
   public String getStorageMsgSerde(String storeName) {
     return get(String.format(MSG_SERDE, storeName), null);
   }
+
+  public String getChangelogSystem(String defaultValue) {
```
jmakes: 👍

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
```diff
@@ -220,10 +223,14 @@ class StreamConfig(config: Config) extends ScalaMapConfig(config) with Logging {
     * @return           the map of properties for the stream
     */
   private def getSystemStreamProperties(systemName: String, streamName: String) = {
-    if (systemName == null || streamName == null) {
+    if (systemName == null) {
```
jmakes: I don't see the need for the distinction. 

Legacy: "systems.sysName.streams.streamName" accessed by getSystemStreamProperties
New: "streams.streamId" accessed by getAllStreamProperties

They're different enough IMO. And hopefully soon the legacy approach will be completely removed, and this method wont be needed.
If you feel strongly about it, I can change it. Just didn't see the need.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
```diff
@@ -220,10 +223,14 @@ class StreamConfig(config: Config) extends ScalaMapConfig(config) with Logging {
...
       Map()
     }
-    config.subset(StreamConfig.STREAM_PREFIX format(systemName, streamName), true)
+    val jsc = new JavaSystemConfig(config);
```
jmakes: 👍  that's the convention. Sorry, reverted to a coding standard from a past life. I'll fix it.

samza-kafka/src/test/scala/org/apache/samza/config/TestKafkaConfig.scala
```diff
@@ -195,25 +195,136 @@ class TestKafkaConfig {
...
+
+    val mapConfig = new MapConfig(props.asScala.asJava)
+    val kafkaConfig = new KafkaConfig(mapConfig)
+    assertEquals("4", kafkaConfig.getChangelogStreamReplicationFactor("store-with-override"))
+    assertEquals("8", kafkaConfig.getChangelogStreamReplicationFactor("store-without-override"))
+    assertEquals("8", kafkaConfig.getDefaultChangelogStreamReplicationFactor)
+  }
+
+  @Test
+  def testCheckpointReplicationFactor() {
```
jmakes: 👍

samza-kafka/src/test/scala/org/apache/samza/config/TestKafkaConfig.scala
jmakes: It was for fun, but ok, I'll keep it professional. Putting on my suit & tie...

samza-kafka/src/test/scala/org/apache/samza/config/TestKafkaConfig.scala
```diff
@@ -195,25 +195,136 @@ class TestKafkaConfig {
...
+    val emptyConfig = new KafkaConfig(new MapConfig())
+    assertNull(emptyConfig.getCheckpointReplicationFactor.orNull)
```
jmakes: Seems like a misunderstanding. There is no default for checkpoint. 

Defaults in terms of the config class:
coordinator - 3
changelog - 2
checkpoint - null

"But", I hear you say, "the config table clearly says that the default checkpoint replication factor is 3". Yeah, *sigh*, it's another example of how muddled the configs have gotten. Apparently the default for checkpoints is injected in KafkaCheckpointManagerFactory, whereas the defaults for coordinator/changelog are defaulted in the config classes (and sometimes in down stream classes... for good measure? IDK).

Anyway, I agree with the consistency argument. I missed that while doing this change. Will fix it. Thanks!

vjagadish1989 (approved):
approved. 

final feedback from my review:

Might be cleaner to make both `getChangelogReplicationFactor` and `getCoordinatorReplicationFactor` return an int consistently. Currently, one returns an `Option` and the other returns an int. 

You can take the final call 👍  looks good otherwise, thanks!

PR #143, opened 2017-04-26
SAMZA-1200: Scala compile for samza-core fails with ambiguous reference error...

prateekm:
... for some compiler versions.


prateekm:
@vjagadish1989 @navina, please take a look.

navina:
@prateekm How did you figure out that this will fix the error ?

prateekm:
@navina I can reproduce the failure when I build from the LI MP against master branch. I didn't figure out what compiler version is being used in the build, but this change fixes it.

PR #145, opened 2017-04-27
SAMZA-1245: Make stream samza.physical.name config name string public

xinyuiscool:
For certain system such as hdfs, the physical stream name might need to be finalized during the config generation. In order to do that, we will need to expose the stream samza.physical.name config string.


xinyuiscool:
@jmakes : please take a look. Thx!

PR #146, opened 2017-04-27
SAMZA-1224 : Revert job coordinator factory config to the old format

navina:
We didn't release since adding this config. So, it is ok to change the format now.


navina:
@prateekm @xinyuiscool Review please!

PR #149, opened 2017-05-01
SAMZA-1249: Fix equality for WindowKey for Non-keyed tumbling windows

vjagadish1989:
- Fix a `ClassCastException` and an NPE when using Tumbling window without keys
- Fix equality and hashCode for `WindowKey`
- Refactor the `TestWindowOperator` unit tests using simpler types and a mock `MessageCollector`.

More details in `SAMZA-1249`

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
+++ b/samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
@@ -178,7 +178,7 @@ public final class Windows {
    */
   public static <M, WV> Window<M, Void, WV> tumblingWindow(Duration duration, Supplier<? extends WV> initialValue,
                                                            FoldLeftFunction<? super M, WV> foldFn) {
-    Trigger<M> defaultTrigger = Triggers.repeat(new TimeTrigger<>(duration));
+    Trigger<M> defaultTrigger = new TimeTrigger<>(duration);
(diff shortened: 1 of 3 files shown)


samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
```diff
@@ -178,7 +178,7 @@ private Windows() { }
    */
   public static <M, WV> Window<M, Void, WV> tumblingWindow(Duration duration, Supplier<? extends WV> initialValue,
                                                            FoldLeftFunction<? super M, WV> foldFn) {
-    Trigger<M> defaultTrigger = Triggers.repeat(new TimeTrigger<>(duration));
+    Trigger<M> defaultTrigger = new TimeTrigger<>(duration);
```
prateekm: Just FMI, how does a tumbling window repeat?

samza-core/src/test/java/org/apache/samza/operators/TestWindowOperator.java
```diff
@@ -55,37 +57,35 @@
 import static org.mockito.Mockito.when;
 
 public class TestWindowOperator {
-  private final MessageCollector messageCollector = mock(MessageCollector.class);
```
prateekm: Thanks for the cleanup.

PR #150, opened 2017-05-01
SAMZA-1214: Bug - system-scoped stream default configs may not be honored

jmakes:
* Re-introduced deprecated system-stream configs into config table
* Fixed position of task.consumer.batch.size in config table
* Moved system-scoped defaults from StreamConfig to SystemConfig


jmakes:
@vjagadish1989 this one fixes the stream config bug and re-introduces the deprecated configs into the table

PR #151, opened 2017-05-01
fix HdfsSystemAdmin when staging directory is empty

lhaiesp:
getSystemStreamMetadata has the potential side effect to persist metadata to a staging directory on hdfs. This could fail if staging directory is empty. This patch addresses the issue with test to cover the scenario.

### DIFF
--- a/samza-hdfs/src/main/java/org/apache/samza/system/hdfs/HdfsSystemAdmin.java
+++ b/samza-hdfs/src/main/java/org/apache/samza/system/hdfs/HdfsSystemAdmin.java
@@ -21,6 +21,7 @@ package org.apache.samza.system.hdfs;
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


samza-hdfs/src/main/java/org/apache/samza/system/hdfs/HdfsSystemAdmin.java
```diff
@@ -116,6 +116,13 @@ public HdfsSystemAdmin(String systemName, Config config) {
   }
 
   static Map<Partition, List<String>> obtainPartitionDescriptorMap(String stagingDirectory, String streamName) {
+    if (StringUtils.isBlank(stagingDirectory)) {
+      LOG.warn("Empty staging directory");
+      return null;
```
prateekm: Can we return Collections.emptyMap() instead of null?

samza-hdfs/src/main/java/org/apache/samza/system/hdfs/HdfsSystemAdmin.java
```diff
@@ -116,6 +116,13 @@ public HdfsSystemAdmin(String systemName, Config config) {
...
+    }
+    if (StringUtils.isBlank(streamName)) {
+      throw new SamzaException("stream name is empty!");
```
prateekm: "Stream name $streamName is null or empty."

samza-hdfs/src/main/java/org/apache/samza/system/hdfs/HdfsSystemAdmin.java
xinyuiscool: +1. Please also change the logging to info since it's expected when runner only needs to query the meta data.

samza-hdfs/src/main/java/org/apache/samza/system/hdfs/HdfsSystemConsumer.java
```diff
@@ -132,6 +133,12 @@ public HdfsSystemConsumer(String systemName, Config config, HdfsSystemConsumerMe
         public Map<Partition, List<String>> load(String streamName)
           throws Exception {
           Validate.notEmpty(streamName);
+          if (StringUtils.isBlank(stagingDirectory)) {
+            throw new SamzaException("Staging directory can't be empty. "
+                + "Is this not a yarn job (currently hdfs system consumer only works in "
+                + "the same yarn environment on which hdfs is running)? "
+                + "Is STAGING_DIRECTORY not set (see HdfsConfig.scala)?");
```
xinyuiscool: can we put the real config name here?

xinyuiscool:
btw, I also changed the status computation for RemoteApplicationRunner to be consistent with LocalApplicationRunner. So now it follows the following logic:

1. status is new if there is new jobs which are not started
2. status is running if all jobs are started and some are running
3. status is unsuccessfulFinished if all jobs are finished and some fail
4. status is successfulFinished if all jobs are finished successfully

PR #153, opened 2017-05-01
SAMZA-1251 - Remove DebounceTimer dependency from ZkLeaderElector & ZkController

navina:
Addresses the following:
* Makes LeaderElectionListener to be explicitly registered by the caller
* Removes debouncetimer dependency from ZkLeaderElector implementation
* [Bug] onBecomeLeader was scheduling a task in timer under "OnBecomeLeader", when it should actually be the same as "OnProcessorChange". Otherwise, it will not cancel when there is a new OnProcessorChange event. 
* [Transient Test Failure] `TestScheduleAfterDebounceTime` tests were relying on timing controlled by sleep. Fixed it by using latch


navina:
@xinyuiscool @sborya Can you guys please review this PR?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -84,7 +73,7 @@ private void timerOff(final String version, final Stat currentStatOfBarrierDone)
       zkUtils.getZkClient().writeData(barrierDonePath, BARRIER_TIMED_OUT, currentStatOfBarrierDone.getVersion());
     } catch (ZkBadVersionException e) {
       // Expected. failed to write, make sure the value is "DONE"
-      ///LOG.("Barrier timeout write failed");
+      ///LOGGER.("Barrier timeout write failed");
```
xinyuiscool: remove the commented-out logging.

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationUtils.java
```diff
@@ -59,7 +58,7 @@ public Latch getLatch(int size, String latchId) {
   public BarrierForVersionUpgrade getBarrier(String barrierId) {
     return new ZkBarrierForVersionUpgrade(barrierId, zkUtils, debounceTimer, zkConfig.getZkBarrierTimeoutMs());
   }
-  @VisibleForTesting
+
```
sborya: we should remember to get rid of this method completely.

samza-core/src/main/java/org/apache/samza/zk/ZkLeaderElector.java
```diff
@@ -49,28 +49,34 @@
...
     this.zkUtils = zkUtils;
     this.keyBuilder = zkUtils.getKeyBuilder();
     this.hostName = getHostName();
-    this.debounceTimer = (debounceTimer != null) ? debounceTimer : new ScheduleAfterDebounceTime();
+    this.previousProcessorChangeListener = new PreviousProcessorChangeListener();
 
     zkUtils.makeSurePersistentPathsExists(new String[]{keyBuilder.getProcessorsPath()});
   }
 
+  @VisibleForTesting
```
sborya: yeah, I wish we didn't have to do this. Don't have any suggestions at the moment now, but if you do... :)

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
navina: Got it!

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -39,13 +39,9 @@
  * ZK based standalone app.
  */
 public class ScheduleAfterDebounceTime {
-  public static final Logger LOG = LoggerFactory.getLogger(ScheduleAfterDebounceTime.class);
+  public static final Logger LOGGER = LoggerFactory.getLogger(ScheduleAfterDebounceTime.class);
```
navina: I don't think we have settled on what is the convention. I have seen both being used. So, I simply decided to keep it consistent within the zk package in core. If you really want this changed, I can do it.

samza-core/src/main/java/org/apache/samza/coordinator/LeaderElector.java
```diff
@@ -32,11 +32,16 @@
...
 public interface LeaderElector {
   /**
-   * Async method that helps the caller participate in leader election.
+   * Register a LeaderElectorListener
    *
-   * @param leaderElectorListener to be invoked if the caller is chosen as a leader through the leader election process
+   * @param listener {@link LeaderElectorListener} interfaces to be invoked upon completion of leader election participation
    */
-  void tryBecomeLeader(LeaderElectorListener leaderElectorListener);
+  void setLeaderElectorListener(LeaderElectorListener listener);
```
navina: 1. It just makes the code more modularized and testable. 
2. Aside from that, if you are passing some kind of state within a recursive call, you need to manage the lifecycle of that instance in recursive function. It makes the code more vulnerable to bugs. 
3. Speaking of semantics, it is the same callback function that needs to be invoked in every call. So, there is really no need to pass around a listener that doesn't do anything different. Does this make sense?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -36,44 +36,33 @@
  * This class creates a barrier for version upgrade.
  * Barrier is started by the participant responsible for the upgrade. (start())
  * Each participant will mark its readiness and register for a notification when the barrier is reached. (waitFor())
- * If a timer (started in start()) goes off before the barrier is reached, all the participants will unsubscribe
+ * If a timer (started in start()) goes off before the barrier is reached, all the participants will un-subscribe
```
navina: err.. yeah..I think I was blindly following IntelliJ's spelling suggestion. Thanks! I will fix it.

samza-core/src/main/java/org/apache/samza/zk/ZkController.java
```diff
@@ -22,11 +22,18 @@
 
 /**
  * Api to the functionality provided by ZK
+ *
+ * Api for JC -&gt; ZK communication
```
navina: :) ok

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationUtils.java
navina: Yeah. I tried to get rid of this method. But the ZkBarrierVersionUpgrade depends on not only the timers but also the zkutils. I didn't want to start refactoring those classes as well. Perhaps you can address it as a part of [SAMZA-1128](https://issues.apache.org/jira/browse/SAMZA-1128)
If it makes sense to address this in that JIRA, let me know. I will add a TODO in the code as well.

samza-core/src/main/java/org/apache/samza/zk/ZkLeaderElector.java
navina: It is fine to introduce constructors annotated with `@VisibleForTesting`. The goal is make it clear that it is not to be used in any case other than testing. I don't see any pressing reason for us to remove the constructor overload.

PR #154, opened 2017-05-01
SAMZA-1246: ApplicatonRunner.stats() should include exception in case of failure

xinyuiscool:
Current when ApplicationRunner.stats() only returns the enum representing the status. It also need to include the exception if the status is failure.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
@@ -133,7 +133,7 @@ public class LocalApplicationRunner extends AbstractApplicationRunner {
       awaitComplete();
 
     } catch (Throwable t) {
-      appStatus = ApplicationStatus.UnsuccessfulFinish;
+      appStatus = ApplicationStatus.unsuccessfulFinish(t);
       throw new SamzaException("Failed to run application", t);
     } finally {
       if (coordination != null) {
@@ -247,7 +247,6 @@ public class LocalApplicationRunner extends AbstractApplicationRunner {
     latch.await();
 
     if (throwable.get() != null) {
-      appStatus = ApplicationStatus.UnsuccessfulFinish;
       throw throwable.get();
     } else {
       appStatus = ApplicationStatus.SuccessfulFinish;
--- a/samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
@@ -84,8 +84,9 @@ public class RemoteApplicationRunner extends AbstractApplicationRunner {
   @Override
   public ApplicationStatus status(StreamApplication app) {
     try {
-      boolean finished = false;
-      boolean unsuccessfulFinish = false;
+      boolean hasNewJobs = false;
+      boolean hasRunningJobs = false;
+      ApplicationStatus unsuccessfulFinishStatus = null;
 
       ExecutionPlan plan = getExecutionPlan(app);
       for (JobConfig jobConfig : plan.getJobConfigs()) {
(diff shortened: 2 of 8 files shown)


samza-api/src/main/java/org/apache/samza/runtime/ApplicationStatus.java
```diff
@@ -0,0 +1,67 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.runtime;
+
+public class ApplicationStatus {
```
jmakes: So after this PR and @navina change, we're going to have 
JobStatus
ApplicationStatus
ContainerStatus
... others?

It'd be nice to have a clear purpose for each of these. 

For example, I don't see why ApplicationStatus needs to handle exceptions, but JobStatus doesn't. If anything, I'd think the only difference between the two is that ApplicationStatus spans multiple jobs, so it might have a status to indicate PartialFinish for the case that a subset of the jobs are not running. Otherwise, I don't see why both are needed.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -106,11 +106,11 @@ public ApplicationStatus status(StreamApplication app) {
       }
 
       if (unsuccessfulFinish) {
-        return ApplicationStatus.UnsuccessfulFinish;
+        return ApplicationStatus.unsuccessfulFinish(null); // no exception for remote execution.
```
jmakes: Feels like this is a gap. Ideally, we should report an exception for the remote case too. Maybe, for example YarnJob should have the smarts to determine why an application failed using the Diagnostic Info

samza-api/src/main/java/org/apache/samza/runtime/ApplicationStatus.java
xinyuiscool: Makes sense. I reverted back the JobStatus and keep only one ApplicationStatus.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
xinyuiscool: Good suggestion! IN this case, I created a SamzaException with the diagonostic info. Please take a look. Thanks.

xinyuiscool:
btw, I also changed the status computation for RemoteApplicationRunner to be consistent with LocalApplicationRunner. So now it follows the following logic:

status is new if there is new jobs which are not started
status is running if all jobs are started and some are running
status is unsuccessfulFinished if all jobs are finished and some fail
status is successfulFinished if all jobs are finished successfully

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -247,7 +247,6 @@ StreamProcessor createStreamProcessor(
     latch.await();
 
     if (throwable.get() != null) {
-      appStatus = ApplicationStatus.UnsuccessfulFinish;
       throw throwable.get();
```
jmakes: I assume appStatus is no longer set here because the next line throws an exception. But on line 136, appStatus is set under the same circumstances. Why is that? Shouldn't run() and waitForFinish() have the same semantics for exception and status handling?

xinyuiscool: oh, I removed this line is because it's redundant. As you see, the exception thrown here will be caught in line 136 and the status will be set again. The code is pretty messy right now. I am waiting for @navina 's change to finally clean these up in patch (https://github.com/apache/samza/pull/135). I will also add the waitForFinish() in that patch.

jmakes: 👍

PR #155, opened 2017-05-02
SAMZA-1256: Improve trace logging for troubleshooting the fluent API

jmakes:


samza-api/src/main/java/org/apache/samza/operators/windows/WindowKey.java
```diff
@@ -52,7 +52,7 @@ public String getPaneId() {
   @Override
   public String toString() {
     String wndKey = "";
-    if (!(key instanceof Void)) {
```
vjagadish1989: thanks!

vjagadish1989 (approved):
approved. thank you for the contribution. 

LGTM, thanks for adding trace logs!

PR #156, opened 2017-05-02
SAMZA-1257: make sure a dataChange listener in LeaderElection is always initialized.

sborya:
@xinyuiscool, @navina please review.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkLeaderElector.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkLeaderElector.java
@@ -105,12 +105,12 @@ public class ZkLeaderElector implements LeaderElector {
     LOG.info("Index = " + index + " Not eligible to be a leader yet!");
     String predecessor = children.get(index - 1);
     if (!predecessor.equals(currentSubscription)) {
+      
+      // callback in case if the previous node gets deleted (when previous processor dies)
+      if (previousProcessorChangeListener == null)
+        previousProcessorChangeListener =  new PreviousProcessorChangeListener(leaderElectorListener);
+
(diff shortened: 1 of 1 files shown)


samza-core/src/main/java/org/apache/samza/zk/ZkLeaderElector.java
```diff
@@ -105,12 +105,12 @@ public void tryBecomeLeader(LeaderElectorListener leaderElectorListener) {
...
     if (!predecessor.equals(currentSubscription)) {
-      if (currentSubscription != null) {
-
-        // callback in case if the previous node gets deleted (when previous processor dies)
-        if (previousProcessorChangeListener == null)
-          previousProcessorChangeListener =  new PreviousProcessorChangeListener(leaderElectorListener);
+      
+      // callback in case if the previous node gets deleted (when previous processor dies)
+      if (previousProcessorChangeListener == null)
+        previousProcessorChangeListener =  new PreviousProcessorChangeListener(leaderElectorListener);
```
navina: This check is incorrect, imo. If you create a new instance of the DataListener here, when you unsubscribe it is unsubscribing with a different reference.. This logic is very confusing. I have simplified it in SAMZA-1150. Please take a look.

sborya: Good observation. I reviewed the code - and it looks like the original logic is wrong. This was just to fix the NPE.
So I changed it to be more correct for existing logic.
If your change coming in today - sure, let's just wait for it. If it is not yet, let's review and commit this change, because it needed in testing (to avoid NPEs).

PR #160, opened 2017-05-04
SAMZA-1155: Validate users configure window.ms when using the fluent API

vjagadish1989:
We will compute triggering duration as follows:
- If user configures `task.window.ms` we will honor it as the triggering duration
- If not, we will use the `GCD(windowTriggerDurations, joinTTLs)` as the triggering duration.

Changes in this PR:
- Common Interface for all time based triggers
- Additional APIs in `StreamGraphImpl` to recursively traverse all `OperatorSpec`s 
- Recursive computation of `triggerInterval` for each `WindowOperatorSpec`
- Tests for all the above


vjagadish1989:
@xinyuiscool @prateekm can you please look at this?

xinyuiscool:
Is it possible to add logic in the planner to figure out a reasonable window.ms instead of asking the user to configure it? seems it's not too hard to do.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -133,4 +139,12 @@ final void writePlanJsonFile(String planJson) {
       log.warn("Failed to write execution plan json to file", e);
     }
   }
+
+  private void validateConfig() {
+    String appClassName = new ApplicationConfig(config).getAppClass();
+    if (appClassName != null && !config.containsKey(TaskConfig.WINDOW_MS())) {
```
jmakes: Why is task.window.ms always required, even if the application doesn't use any window operator?

prateekm: Currently both window and join require task.window.ms.

We could do traversals to figure out if we have those operators, and we could also override the default value of task.window.ms to something other than -1 based on heuristics, but I'd prefer to keep this simple and explainable for now. WDYT?

jmakes: Yeah, still not every job will be doing join or window, so it seems odd to force those who don't do those operations to configure something unnecessary. 

I think these validations could fit very nicely into the planner, which is the first place we take a collective look at the DAG

vjagadish1989:
@prateekm @jmakes Can you please take a look?

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -156,4 +161,46 @@ public ContextManager getContextManager() {
...
+      }
+    }
+  }
+
+  /**
+   * Returns <tt>true</tt> iff this {@link StreamGraphImpl} contains a join or a window operator
+   *
+   * @return  <tt>true</tt> iff this {@link StreamGraphImpl} contains a join or a window operator
+   */
+  public boolean hasWindowOrJoins() {
```
jmakes: nit: window and join are interesting not because of their identities, but because of their behavior; that they both depend on the "window" phase of the event loop. If there's an obvious name that captures this, it would be better. 

So, something like "hasTimedOperators" or something that captures the general behavior rather than the current known examples. If there is no such name in this case, feel free to ignore.

prateekm: +1 for naming.

----

Not necessary in this PR, but would it be cleaner to consolidate all the logic in one place, either the planner or here? For example, StreamGraph could have a single public method `getTimerInterval` (default = -1) that does the hasTimedOperators + traversal + gcd calculation, and the logic in Planner could be:

```
long timerInterval = graph.getTimerInterval();
if (windowMs == null || windowMs <= 0) {
  LOG.info("...");
  windowMs = timerInterval;
} else if (windowMs > timerInterval) {
  LOG.warn("...");
  windowMs = timerInterval; // maybe 
}
```

samza-core/src/main/java/org/apache/samza/operators/spec/WindowOperatorSpec.java
```diff
@@ -76,4 +88,53 @@ public int getOpId() {
...
+    LOG.info("Got {} timer triggers", timerTriggers.size());
+
+    List<Long> candidateDurations = timerTriggers.stream()
+        .map(timeBasedTrigger -> timeBasedTrigger.getDuration().toMillis())
+        .collect(Collectors.toList());
+
+    return MathUtils.gcd(candidateDurations);
+  }
+
+  private List<TimeBasedTrigger> getTimeTriggers(Trigger rootTrigger) {
```
prateekm: getTimeBasedTriggers

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -156,4 +161,46 @@ public ContextManager getContextManager() {
   /* package private */ int getNextOpId() {
     return this.opId++;
   }
+
+  /**
+   * Get all {@link OperatorSpec}s available in this {@link StreamGraphImpl}
+   *
+   * @return  a set of all available {@link OperatorSpec}s
+   */
+  public Set<OperatorSpec> getAllOperatorSpecs() {
```
prateekm: `public Collection<OperatorSpec>`

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
```diff
@@ -104,6 +109,15 @@ public JobConfig generateConfig(String executionPlanJson) {
 
     List<String> inputs = inEdges.stream().map(edge -> edge.getFormattedSystemStream()).collect(Collectors.toList());
     configs.put(TaskConfig.INPUT_STREAMS(), Joiner.on(',').join(inputs));
+
+    // set triggering interval if a window or join is defined
+    if (streamGraph.hasWindowOrJoins() && !config.containsKey(TaskConfig.WINDOW_MS())) {
```
xinyuiscool: I think we also want to check whether the value is -1, which is the default.

samza-core/src/main/java/org/apache/samza/operators/util/MathUtils.java
```diff
@@ -0,0 +1,50 @@
...
+ * under the License.
+ */
+
+package org.apache.samza.operators.util;
+
+import java.util.List;
+
+public class MathUtils {
+
+  public static long gcd(long a, long b) {
```
xinyuiscool: just curious, Java.math doesn't have gcd?

PR #162, opened 2017-05-04
SAMZA-1228 : StreamProcessor should stop JmxServer

navina:
This is not the solution posted in SAMZA-1228. For now, we are moving jmxserver lifecycle to be within the container. Ideally, it should be within the Streamprocessor so that the job coordinator can also be associated with the same instance.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
+++ b/samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
@@ -678,6 +678,8 @@ class SamzaContainer(
       info("Shutting down.")
       removeShutdownHook
 
+      jmxServer.stop
+
       shutdownConsumers
       shutdownTask
       shutdownStores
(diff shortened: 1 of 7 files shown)


samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -702,6 +703,8 @@ class SamzaContainer(
         status = SamzaContainerStatus.FAILED
     }
 
+    jmxServer.stop
```
prateekm: Should this be in the finally block?

navina: We are not throwing exceptions anymore. So, it doesn't matter. Either ways, it will reach Line 706. I could make it part of the shutdown sequence. But if an exception is thrown, it won't shutdown and, I need a finally. Let me know if I am missing something.

PR #163, opened 2017-05-04
SAMZA-871: Heart-beat mechanism between JobCoordinator and all running containers

abhishekshivanna:


samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatClient.java
```diff
@@ -0,0 +1,106 @@
...
+          throw new IOException("Http Error during heartbeat");
+        } else {
+          BufferedReader br = new BufferedReader(new InputStreamReader(conn.getInputStream()));
+          String body = br.lines().collect(Collectors.joining());
+          br.close();
+          return body;
+        }
+      } catch (Exception e) {
+        try {
+          Thread.sleep((int) Math.round(Math.pow(2, currentTry)) * 1000);
```
navina: Can you please use constants for the backoff time and multipliers? 
Take a look at ExponentialSleepStrategy.scala. It makes the back-off try loop more easier to use. fyi, I am not suggesting you use that class. Just a suggestion on how you can better organize your back-off strategy.

samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatClient.java
```diff
@@ -0,0 +1,106 @@
...
+import org.codehaus.jackson.map.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class ContainerHeartbeatClient {
+  private static Logger log = LoggerFactory.getLogger(ContainerHeartbeatClient.class);
+  private final String coordinatorUrl;
+  private final String executionEnvContainerId;
+  private final int numberOfRetries = 3;
```
navina: It is generally a good practice to define constants with capitalized variable names - for example: NUMBER_OF_RETRIES, TIMEOUT_MS

samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatClient.java
```diff
@@ -0,0 +1,106 @@
...
+  private final int timeout = 5000;
+
+  public ContainerHeartbeatClient(String coordinatorUrl, String executionEnvContainerId) {
+    this.coordinatorUrl = coordinatorUrl;
+    this.executionEnvContainerId = executionEnvContainerId;
+  }
+
+  public boolean isAlive() {
+    String queryUrl = coordinatorUrl + "containerHeartbeat?executionContainerId=" + executionEnvContainerId;
```
navina: Let's not hard-code URL patterns in the code. You can perhaps define the regex pattern as a member variable and simply, format it here.

samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatClient.java
```diff
@@ -0,0 +1,106 @@
...
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.Random;
+import java.util.stream.Collectors;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class ContainerHeartbeatClient {
```
navina: Can you add some javadoc for this class?

navina: Since this class is responsible for sending requests and parsing responses, I think it will be useful to document the protocol used for request and response here.

samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatClient.java
```diff
@@ -0,0 +1,106 @@
...
+  private final String executionEnvContainerId;
+  private final int numberOfRetries = 3;
+  private final int timeout = 5000;
+
+  public ContainerHeartbeatClient(String coordinatorUrl, String executionEnvContainerId) {
+    this.coordinatorUrl = coordinatorUrl;
+    this.executionEnvContainerId = executionEnvContainerId;
+  }
+
+  public boolean isAlive() {
```
navina: Is there any reason why you chose to return a boolean for isAlive, instead of returning the ContainerHeartbeatResponse? 

When I think about a "client" of any kind, it only handles the request/response part. Not the functional details associated with the caller. 

For example, Line 48 in ContainerHeartbeatMonitor -> `containerHeartbeatClient.isAlive()` It is not clear if it is returning the "liveness" of the client or the "liveness" of the container ? 

If you return the response object directly, it will be cleaner. It will also be more re-usable, if we chose to fetch other metadata in the same protocol.

samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatMonitor.java
```diff
@@ -0,0 +1,67 @@
...
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+
+
+public class ContainerHeartbeatMonitor {
+  private static Logger log = LoggerFactory.getLogger(ContainerHeartbeatMonitor.class);
+  private static final ThreadFactory THREAD_FACTORY = new HeartbeatThreadFactoryImpl();
+  private final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor(THREAD_FACTORY);
+  private final Runnable onContainerExpired;
+  private final ContainerHeartbeatClient containerHeartbeatClient;
+  private final int scheduleMS = 60000;
```
navina: Same comment about variable names for constants.

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -98,8 +102,16 @@ public void onContainerFailed(Throwable t) {
             }
           });
 
+      containerHeartbeatMonitor = getContainerHeartbeatMonitor();
```
vjagadish1989: Can re-name all `getXXX()` to `createXXX()` if all they are doing is creating a new object. OTOH, we should retain `getXX()` if we're returning a cached copy

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -147,5 +160,24 @@ public static void main(String[] args) throws Exception {
 
     StreamApplication streamApp = TaskFactoryUtil.createStreamApplication(config);
     new LocalContainerRunner(jobModel, containerId).run(streamApp);
```
vjagadish1989: Document that run blocks until the container is shut-down (forced or otherwise)

samza-core/src/main/scala/org/apache/samza/container/RunLoop.scala
```diff
@@ -22,7 +22,7 @@ package org.apache.samza.container
 import org.apache.samza.task.CoordinatorRequests
 import org.apache.samza.system.{IncomingMessageEnvelope, SystemConsumers, SystemStreamPartition}
 import org.apache.samza.task.ReadableCoordinator
-import org.apache.samza.util.{Logging, Throttleable, ThrottlingExecutor, TimerUtils}
+import org.apache.samza.util._
```
vjagadish1989: no ;)

samza-yarn/src/main/java/org/apache/samza/webapp/YarnContainerHeartbeatServlet.java
```diff
@@ -0,0 +1,87 @@
...
+  protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+      throws ServletException, IOException {
+    PrintWriter printWriter = resp.getWriter();
+    String containerIdParam = req.getParameter(YARN_CONTAINER_ID);
+    resp.setContentType(APPLICATION_JSON);
+    boolean alive = false;
+    ContainerId yarnContainerId;
+    try {
+      yarnContainerId = ContainerId.fromString(containerIdParam);
+      for (Map.Entry pair : yarnAppState.runningYarnContainers.entrySet()) {
```
vjagadish1989: Should use a types for `Map.Entry`. 

minor:
If all we need is a list of values,

Can re-write it as:
```
for (YarnContainer container : yarnAppState.values())
```

samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatClient.java
abhishekshivanna: Agreed. Will add a brief description of the protocol.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -156,7 +156,7 @@ object JobModelManager extends Logging {
     jobModelRef.set(jobModel)
 
     val server = new HttpServer
-    server.addServlet("/*", new JobServlet(jobModelRef))
+    server.addServlet("/", new JobServlet(jobModelRef))
```
abhishekshivanna: the `*` matches all paths from the root to the servlet. since we have a new endpoint this needs to be removed.

samza-core/src/test/java/org/apache/samza/container/TestContainerHeartbeatMonitor.java
```diff
@@ -0,0 +1,66 @@
...
+package org.apache.samza.container;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import junit.framework.Assert;
+import org.junit.Test;
+
+import static org.mockito.Mockito.*;
+
+
```
vjagadish1989: Very minor: extra `\n`

samza-core/src/test/java/org/apache/samza/container/TestContainerHeartbeatMonitor.java
```diff
@@ -0,0 +1,66 @@
...
+public class TestContainerHeartbeatMonitor {
+
+  private ContainerHeartbeatResponse response = new ContainerHeartbeatResponse();
```
vjagadish1989: Can be `final`

samza-yarn/src/test/java/org/apache/samza/webapp/TestYarnContainerHeartbeatServlet.java
```diff
@@ -0,0 +1,99 @@
...
+
+public class TestYarnContainerHeartbeatServlet {
+
+  private YarnContainer container;
+  private YarnAppState yarnAppState;
+  private HttpServer webApp;
+  private ObjectMapper mapper;
+  private ReadableMetricsRegistry registry;
+
+  public ContainerHeartbeatResponse heartbeat;
```
vjagadish1989: Any reason for this to be `public`?

samza-yarn/src/main/java/org/apache/samza/webapp/YarnContainerHeartbeatServlet.java
```diff
@@ -0,0 +1,86 @@
...
+    ContainerId yarnContainerId;
+    try {
+      yarnContainerId = ContainerId.fromString(containerIdParam);
+      for (YarnContainer yarnContainer : yarnAppState.runningYarnContainers.values()) {
+        if (yarnContainer.id().compareTo(yarnContainerId) == 0) {
+          alive = true;
+        }
+      }
+      if (!alive) {
+        heartbeatInvalidCount.inc();
```
vjagadish1989: Thanks for thoughtfully adding this metric, Super-useful to track orphaned containers.

samza-yarn/src/main/java/org/apache/samza/webapp/YarnContainerHeartbeatServlet.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.samza.job.yarn.YarnContainer;
+import org.apache.samza.metrics.Counter;
+import org.apache.samza.metrics.ReadableMetricsRegistry;
+import org.codehaus.jettison.json.JSONException;
+import org.codehaus.jettison.json.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class YarnContainerHeartbeatServlet extends HttpServlet {
```
vjagadish1989: Would be nice to add public docs.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
vjagadish1989: Thanks for explaining, makes sense.

samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatClient.java
```diff
@@ -0,0 +1,100 @@
...
+ * The expected endpoint on the JobCoordinator side is `/containerHeartbeat`
+ * which takes `executionContainerId` (eg: YARN container Id) as a query parameter to validate against.
+ */
+public class ContainerHeartbeatClient {
+  private static Logger log = LoggerFactory.getLogger(ContainerHeartbeatClient.class);
+  private final String coordinatorUrl;
+  private final String executionEnvContainerId;
+  private final String heartbeatEndpoint = "%scontainerHeartbeat?executionContainerId=%s";
+  private static final int NUMBER_OF_RETRIES = 3;
+  private static final int TIMEOUT = 5000;
```
navina: Use TIMEOUT_MS ?

samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatResponse.java
```diff
@@ -0,0 +1,33 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.container;
+
+public class ContainerHeartbeatResponse {
```
navina: I think the response object should be immutable. You can do it by:
* adding a constructor `ContainerHeartbeatResponse(boolean isAlive)`
* remove `setAlive` method? 
* declare `alive` as final

samza-yarn/src/main/java/org/apache/samza/webapp/YarnContainerHeartbeatServlet.java
```diff
@@ -0,0 +1,86 @@
...
+  public static final String GROUP = "ContainerHeartbeat";
+  private final Counter heartbeatInvalidCount;
+
+  private YarnAppState yarnAppState;
+  private JSONObject jsonResponse;
+
+  public YarnContainerHeartbeatServlet(YarnAppState yarnAppState, ReadableMetricsRegistry registry) {
+    this.yarnAppState = yarnAppState;
+    this.jsonResponse = new JSONObject();
+    this.heartbeatInvalidCount = registry.newCounter(GROUP, "heartbeats-invalid");
```
navina: can you please document the metric here -> [http://samza.apache.org/learn/documentation/latest/container/metrics-table.html](http://samza.apache.org/learn/documentation/latest/container/metrics-table.html)

navina: **nit**: s/heartbeatInvalidCount/heartbeatsInvalidCount/

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnContainerRunner.java
```diff
@@ -110,6 +111,7 @@ public void runContainer(String samzaContainerId, Container container, CommandBu
     log.info("Container ID {} using command {}", samzaContainerId, command);
 
     Map<String, String> env = getEscapedEnvironmentVariablesMap(cmdBuilder);
+    env.put(ShellCommandConfig.ENV_EXECUTION_ENV_CONTAINER_ID(), Util.envVarEscape(container.getId().toString()));
```
navina: You should probably add this environmental variable to ShellCommandBuilder and not here?? Please check how other environmental variables are being passed to the containers.

samza-yarn/src/test/scala/org/apache/samza/job/yarn/TestSamzaYarnAppMasterService.scala
```diff
@@ -75,8 +77,9 @@ class TestSamzaYarnAppMasterService {
     val jobModelManager = JobModelManager(config)
     val samzaState = new SamzaApplicationState(jobModelManager)
     val state = new YarnAppState(-1, ConverterUtils.toContainerId("container_1350670447861_0003_01_000002"), "testHost", 1, 1);
+    val registry = new MetricsRegistryMap()
```
navina: Why is this change needed?

navina:
@abhishekshivanna  Is there any plan to make the heartbeat interval tunable? Or do we have some estimate on what the optimal value should be?

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnContainerRunner.java
abhishekshivanna: The Yarn container Id is only available to be set in two places: here and in YarnClusterResourceManager. Unless I make changes to ShellCommandBuilder to be able to add new configs (via a setter??) I don't this this environment var can be set elsewhere

samza-yarn/src/test/scala/org/apache/samza/job/yarn/TestSamzaYarnAppMasterService.scala
abhishekshivanna: I use the registry to create a new counter in the heartbeat servlet, which is initialized in SamzaYarnAppMasterService.onInit() called by this test. The tests failed due to an NPE before this change.

abhishekshivanna:
@navina regarding making the heartbeat interval tunable via config I think it was bought up in the SEP discussion that we can choose a sensible default for now and if needed make it a tunable in the future.

samza-yarn/src/test/scala/org/apache/samza/job/yarn/TestSamzaYarnAppMasterService.scala
navina: Oh cool. Thanks for fixing it :) It seemed like a one-off change. that's why I asked.

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnContainerRunner.java
navina: Ah my bad. I got confused between Samza's containerId and yarn's container id. You are right. This is the right place to add it. We are passing all of AM container's environmental variables to the container's context and additionally, including the yarn container id.

samza-core/src/test/java/org/apache/samza/container/TestContainerHeartbeatClient.java
```diff
@@ -0,0 +1,81 @@
...
+    client.setThrowException(true);
+    ContainerHeartbeatResponse response = client.requestHeartbeat();
+    Assert.assertFalse(response.isAlive());
+  }
+
+  private class MockContainerHeartbeatClient extends ContainerHeartbeatClient {
+    private String httpOutput;
+    private boolean throwException = false;
+
+    public void setThrowException(boolean throwException) {
```
vjagadish1989: s/setThrowException/`shouldThrowException`

samza-yarn/src/main/java/org/apache/samza/webapp/YarnContainerHeartbeatServlet.java
```diff
@@ -0,0 +1,92 @@
...
+    PrintWriter printWriter = resp.getWriter();
+    String containerIdParam = req.getParameter(YARN_CONTAINER_ID);
+    ContainerHeartbeatResponse response;
+    resp.setContentType(APPLICATION_JSON);
+    boolean alive = false;
+    try {
+      yarnContainerId = ContainerId.fromString(containerIdParam);
+      for (YarnContainer yarnContainer : yarnAppState.runningYarnContainers.values()) {
+        if (yarnContainer.id().compareTo(yarnContainerId) == 0) {
+          alive = true;
```
vjagadish1989: Do we need to `break` here instead of continuing to traverse?

samza-yarn/src/test/java/org/apache/samza/webapp/TestYarnContainerHeartbeatServlet.java
```diff
@@ -0,0 +1,98 @@
...
+import org.codehaus.jackson.map.ObjectMapper;
+import org.eclipse.jetty.servlet.DefaultServlet;
+import org.eclipse.jetty.servlet.ServletHolder;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
```
vjagadish1989: extra new lines

samza-yarn/src/main/java/org/apache/samza/webapp/YarnContainerHeartbeatServlet.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.samza.job.yarn.SamzaAppMasterMetrics;
+import org.apache.samza.job.yarn.YarnAppState;
+import org.apache.samza.job.yarn.YarnContainer;
+import org.apache.samza.metrics.Counter;
+import org.apache.samza.metrics.ReadableMetricsRegistry;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
```
vjagadish1989: Minor: extra new line

vjagadish1989 (approved):
approved. thanks for the contributions Abhishek.

Final round of f/b from last review:

samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatClient.java
```diff
@@ -0,0 +1,112 @@
...
+    } catch (IOException e) {
+      LOG.error("Error in container heart beat protocol. Query url: {} response: {}", heartbeatEndpoint, reply);
+    }
+    response = new ContainerHeartbeatResponse(false);
+    return response;
+  }
+
+  String httpGet(URL url) throws IOException {
+    HttpURLConnection conn;
+    int delayMillis = 1000;
```
abhishekshivanna: So the`delayMillis` is a variable that gets multiplied by the `BACKOFF_MULTIPLIER` on every failure. `1000` is is just the initial value. I could make the initial a constant and make `delayMillis` take its initial value from the constant but i'm not sure it that is required.

samza-yarn/src/main/java/org/apache/samza/webapp/YarnContainerHeartbeatServlet.java
abhishekshivanna: Yes that makes sense. Added it.

samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatClient.java
navina: > I could make the initial a constant
+1 for INIT_DELAY_MS

samza-yarn/src/main/java/org/apache/samza/webapp/YarnContainerHeartbeatServlet.java
```diff
@@ -0,0 +1,92 @@
...
+ * which in YARN's case is the YARN container Id.
+ * This servlet validates the container Id against the list
+ * of running containers maintained in the {@link YarnAppState}.
+ * The returned {@link ContainerHeartbeatResponse#isAlive()} is
+ * <code> true </code> iff. the container Id exists in {@link YarnAppState#runningYarnContainers}.
+ */
+public class YarnContainerHeartbeatServlet extends HttpServlet {
+
+  private static final String YARN_CONTAINER_ID = "executionContainerId";
+  private static final Logger log = LoggerFactory.getLogger(YarnContainerHeartbeatServlet.class);
```
navina: nit: s/log/LOG/

samza-yarn/src/main/java/org/apache/samza/webapp/YarnContainerHeartbeatServlet.java
```diff
@@ -0,0 +1,92 @@
...
+  public static final String APPLICATION_JSON = "application/json";
```
navina: any reason for this to be a *public* variable?

samza-yarn/src/main/java/org/apache/samza/webapp/YarnContainerHeartbeatServlet.java
```diff
@@ -0,0 +1,92 @@
...
+  public static final String GROUP = SamzaAppMasterMetrics.class.getName();
```
navina: same question as above.

PR #164, opened 2017-05-04
SAMZA-1206: Fix TestJMXServer.

shanthoosh:


shanthoosh:
@prateekm @navina 

PR #165, opened 2017-05-04
SAMZA-1263: Samza Fluent: NPE is streams.id.samza.system is missing

jmakes:


jmakes:
@prateekm this ought to fix it!

samza-core/src/test/java/org/apache/samza/runtime/TestAbstractApplicationRunner.java
```diff
@@ -20,17 +20,16 @@
...
 import org.apache.samza.config.MapConfig;
 import org.apache.samza.config.StreamConfig;
-import org.apache.samza.application.StreamApplication;
 import org.apache.samza.job.ApplicationStatus;
 import org.apache.samza.system.StreamSpec;
 import org.junit.Test;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
+import static org.junit.Assert.*;
```
vjagadish1989: - no static imports 
- no wild card imports
of course, no `static wild cards` either ;)

Now, For some 0.13 release humor: 

https://www.pinterest.com/pin/494551602806487660/

jmakes: Thanks, IntelliJ. :-)

But to clarify, my understanding is that those rules don't apply to tests. Correct me if I'm wrong.

prateekm: I also don't mind importing * for mockito/junit static methods in tests. It's a little annoying to import them on a case by case basis.

PR #166, opened 2017-05-05
SAMZA-1150 : Handling Error propagation between ZkJobCoordinator & DebounceTimer

navina:
This PR depends on PR #153 
* Treats all errors in jobcoordinator as FATAL and shuts-down the streamprocessor
* [Bug] Fixed bug reported in SAMZA-1241
* Introduced a callback to be associated with the timer (same callback for every Runnable failure)

**TBD**: some more unit tests


navina:
@xinyuiscool @sborya @prateekm Please review . thanks!

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -72,13 +84,29 @@ synchronized public void scheduleAfterDebounceTime(String actionName, long debou
       futureHandles.remove(actionName);
     }
     // schedule a new task
-    sf = scheduledExecutorService.schedule(runnable, debounceTimeMs, TimeUnit.MILLISECONDS);
-    LOGGER.info("DEBOUNCE: scheduled " + actionName + " in " + debounceTimeMs);
+    sf = scheduledExecutorService.schedule(() -> {
+        try {
+          runnable.run();
+          LOGGER.debug(actionName + " completed successfully.");
+        } catch (Exception e) {
```
xinyuiscool: I think we need to catch the throwble here. Otherwise if's a RuntimeException like SamzaException the thread will be killed without notifying the listener.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -51,10 +50,23 @@
 
   public static final int DEBOUNCE_TIME_MS = 2000;
 
+  private final Callback callback;
```
xinyuiscool: Maybe a better naming for the Callback interface?

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -19,11 +19,28 @@
...
+    val systemNames = config.getSystemNames.toSet
+    // Map the name of each system to the corresponding SystemAdmin
+    val systemAdmins = systemNames.map(systemName => {
+      val systemFactoryClassName = config
+        .getSystemFactory(systemName)
+        .getOrElse(throw new SamzaException("A stream uses system %s, which is missing from the configuration." format systemName))
+      val systemFactory = Util.getObj[SystemFactory](systemFactoryClassName)
+      systemName -> systemFactory.getAdmin(systemName, config)
+    }).toMap
+    systemAdmins
```
xinyuiscool: is this line needed?

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
navina: ScheduledTaskCallback?

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
navina: @xinyuiscool can you please explain why RuntimeException won't get caught with an `Exception` ? the unit test I wrote throws a RuntimeException and it seems to invoke the callback.

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
navina: not really. We can get rid of the local variable actually. I will change that.

navina: My bad. That shouldn't even be there. Nice catch!

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -18,91 +18,74 @@
...
+        LOGGER.error("Received exception from in JobCoordinator Processing!");
+        stop();
+      });
+
     zkController.register();
   }
 
   @Override
-  public void stop() {
+  public synchronized void stop() {
```
sborya: can you please explain why do you need synchronized here?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -148,30 +131,29 @@ public void doOnProcessorChange(List<String> processors) {
...
         // get the new job model
         newJobModel = zkUtils.getJobModel(version);
 
-        log.info("pid=" + processorId + ": new JobModel available. ver=" + version + "; jm = " + newJobModel);
+        LOGGER.info("pid=" + processorId + ": new JobModel available. ver=" + version + "; jm = " + newJobModel);
 
         // update ZK and wait for all the processors to get this new version
-        ZkBarrierForVersionUpgrade barrier = (ZkBarrierForVersionUpgrade) coordinationUtils.getBarrier(
-            JOB_MODEL_UPGRADE_BARRIER);
+        ZkBarrierForVersionUpgrade barrier =
```
sborya: Just wondering, if this is auto-format change or it is your formatting preferences?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -213,27 +195,27 @@ private void generateNewJobModel(List<String> processors) {
...
     JobModel jobModel = JobModelManager.readJobModel(this.config, Collections.emptyMap(), null, streamMetadataCache,
         containerIds);
 
-    log.info("pid=" + processorId + "Generated jobModel: " + jobModel);
+    LOGGER.info("pid=" + processorId + "Generated jobModel: " + jobModel);
 
     // publish the new job model first
     zkUtils.publishJobModel(nextJMVersion, jobModel);
-    log.info("pid=" + processorId + "published new JobModel ver=" + nextJMVersion + ";jm=" + jobModel);
+    LOGGER.info("pid=" + processorId + "published new JobModel ver=" + nextJMVersion + ";jm=" + jobModel);
```
sborya: I know it is existing code. Please remove the job model from this log line. We don't need it twice in the log.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
navina: JobCoordinator can be stopped by StreamProcessor (can happen when the container thread stops/fails or when the user-api invokes a stop on streamprocessor) or the JobCoordinator's failure callback. So, it needs to be synchronized. 
Additionally, I think we need to maintain the state of the jobcoordinator or make stop idempotent. Atm, I didn't want to refactor JC state.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
navina: Ok sure,.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
navina: I think it was spilling over the character limit of 80 on a single line. Don't recollect if I changed it or it was the IDE.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -51,10 +50,23 @@
...
+  private final ScheduledTaskFailureCallback scheduledTaskFailureCallback;
+
   private final ScheduledExecutorService scheduledExecutorService = Executors.newSingleThreadScheduledExecutor(
       new ThreadFactoryBuilder().setNameFormat("debounce-thread-%d").setDaemon(true).build());
   private final Map<String, ScheduledFuture> futureHandles = new HashMap<>();
 
+  // Ideally, this should be only used for testing. But ZkBarrierForVersionUpgrades uses it. This needs to be fixed.
+  // TODO: Timer shouldn't be passed around the components. It should be associated with the JC or the caller of
+  // coordinationUtils.
+  public ScheduleAfterDebounceTime() {
```
sborya: I think the debounceTimer should be associated with the groupId of the coordinationUtils, so the same thread is used for all the action related to the same processor. Also I don't where this constructor is invoked, so may be it can be removed. 
On the other hand, it looks like errorHandling listener should be passed from the point of invocation, so it could be different for different components.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -18,91 +18,74 @@
...
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 
 /**
  * JobCoordinator for stand alone processor managed via Zookeeper.
  */
 public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
-  private static final Logger log = LoggerFactory.getLogger(ZkJobCoordinator.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(ZkJobCoordinator.class);
```
sborya: Please use LOG.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
navina: > I think the debounceTimer should be associated with the groupId of the coordinationUtils, so the same thread is used for all the action related to the same processor.

`ScheduleAfterDebounceTime` is the actual event processing loop of the `ZkJobCoordinator`. I don't want to couple this with any other component that may use `CoodinationUtils`, which is used as a **library**.  It should be left up-to the caller on how it deals with the asynchronous callbacks when using a library. 

> Also I don't where this constructor is invoked, so may be it can be removed.

This constructor is used in `getCoordinationService` of `ZkCoordinationServiceFactory`. I didn't want to re-factor this code along with this PR. We can do it as a part of SAMZA-1128

> it looks like errorHandling listener should be passed from the point of invocation, so it could be different for different components.

It could be different for different components. But at this point, we are differentiating the different error scenarios. That's why I didn't add it to the `schedule` api and, instead added it to the constructor. Let me try adding to the schedule method itself.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
navina: Is LOG a standard convention of a *Logger* object?

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
navina: @sborya I couldn't easily add it to the schedule method because the callback should `stop` the JobCoordinator. So, if I were to replace all submissions to the scheduler, all those components should have access to `stop` method of `ZkJobCoordinator`. This doesn't feel right, esp. in ZkBarrierForVersionUpgrade, we will require a reference to the jobcoordinator. That's clunky and I don't want to add that now. 
I think this refactoring will require changes in SAMZA-1128 :(

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
xinyuiscool: sorry, my bad, I used the wrong class name. It's Error, not RuntimeException (https://docs.oracle.com/javase/7/docs/api/java/lang/Error.html), I think we want to catch those too. Otherwise the thread might die without notifying the listener. So it is safer to catch all the throwable and pass on to the listener.

navina: Ok. Makes sense 👍

PR #167, opened 2017-05-05
SAMZA-1266: Unable to use MetricsSnapshotReporterFactory with fluent API

jmakes:


samza-core/src/main/scala/org/apache/samza/metrics/reporter/MetricsSnapshotReporterFactory.scala
```diff
@@ -47,9 +47,10 @@ class MetricsSnapshotReporterFactory extends MetricsReporterFactory with Logging
       .getJobId
       .getOrElse(1.toString)
 
-    val taskClass = config
+    val taskClass = Option(config
       .getTaskClass
-      .getOrElse(throw new SamzaException("No task class defined for config."))
+      .getOrElse(new ApplicationConfig(config).getAppClass()))
```
prateekm: Very Minor: Maybe:
```
config
  .getTaskClass
  .orElse(Option(new .. getAppClass()))
  .getOrElse(throw)
```

jmakes: 👍 yes, thanks!

PR #173, opened 2017-05-09
SAMZA-1272 : ZkCoordinationUtils deletes the entire Zk tree on reset

navina:
* ZkCoordinationUtils has a reset interface that deletes the entire Zk tree. This is not desirable.
* Also, fixed flakiness in unit test by unique barrier name in each of the unit tests. Otherwise, they share the same path on Zk and fail during concurrent test execution

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
@@ -19,13 +19,6 @@
 
 package org.apache.samza.zk;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.concurrent.TimeUnit;
-
-import com.google.common.annotations.VisibleForTesting;
 import org.I0Itec.zkclient.IZkChildListener;
 import org.I0Itec.zkclient.IZkDataListener;
 import org.I0Itec.zkclient.ZkClient;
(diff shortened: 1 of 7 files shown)


navina:
@sborya @xinyuiscool @prateekm Can you guys please review this PR? Important bug fix

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -300,6 +302,7 @@ public void subscribeToProcessorChange(IZkChildListener listener) {
     zkClient.subscribeChildChanges(keyBuilder.getProcessorsPath(), listener);
   }
 
+  @VisibleForTesting
```
prateekm: Probably don't need this for public methods.

navina: Actually, I am just going to remove it. It is not used anywhere but in test classes. It can just be local to the test class.

xinyuiscool:
Looks good to me.  @sborya : Do we know what's left over under the root after all processors are shut down? I assume all the processor nodes and latch nodes will be cleaned up.

navina:
> Do we know what's left over under the root after all processors are shut down? I assume all the processor nodes and latch nodes will be cleaned up.

Processor nodes are cleaned up automatically because they are ephemeral. Latch / Barrier related nodes are not cleaned up. I believe @sborya is going to address them as a part of [SAMZA-1165](https://issues.apache.org/jira/browse/SAMZA-1165).

PR #176, opened 2017-05-09
SAMZA-1275: Kafka throws when users configure replication.factor for …

jmakes:
…Kafka default stream

### DIFF
--- a/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaStreamSpec.java
+++ b/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaStreamSpec.java
@@ -25,12 +25,16 @@ import java.util.Properties;
 import kafka.log.LogConfig;
 import org.apache.samza.config.KafkaConfig;
 import org.apache.samza.system.StreamSpec;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 /**
  * Extends StreamSpec with the ability to easily get the topic replication factor.
  */
 public class KafkaStreamSpec extends StreamSpec {
+  private static Logger LOG = LoggerFactory.getLogger(KafkaStreamSpec.class);
+
   private static final int DEFAULT_REPLICATION_FACTOR = 2;
 
   /**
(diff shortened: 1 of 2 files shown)


cpettitt-linkedin:
Code changes looks good. It would be great to have a test to verify filtered config is removed from the spec.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaStreamSpec.java
```diff
@@ -62,6 +63,26 @@ private static Properties mapToProperties(Map<String, String> map) {
   }
 
   /**
+   * Filter out properties from the original config that are not supported by Kafka.
+   * For example, we allow users to set replication.factor as a property of the streams
+   * and then parse it out so we can pass it separately as Kafka requires. But Kafka
+   * will also throw if replication.factor is passed as a property on a new topic.
```
navina: QQ: Is this behavior seen only when "creating" new topics?

jmakes: yes

jmakes:
@cpettitt-linkedin Replying here since github won't let me directly

Unit test is coming, was just looking for feedback while I commuted. 

BTW just tested this with the hello-samza and it seems to fix the exception. 

Stay tuned!

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaStreamSpec.java
```diff
@@ -62,6 +67,30 @@ private static Properties mapToProperties(Map<String, String> map) {
...
+   *
+   * @param originalConfig  The original config to filter
+   * @return                The filtered config
+   */
+  private static Map<String, String> filterUnsupportedProperties(Map<String, String> originalConfig) {
```
bharathkk: Can the return type me ImmutableMap instead?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaStreamSpec.java
```diff
@@ -62,6 +67,30 @@ private static Properties mapToProperties(Map<String, String> map) {
...
+      // Kafka requires replication factor, but not as a property, so we have to filter it out.
+      if (!KafkaConfig.TOPIC_REPLICATION_FACTOR().equals(entry.getKey())) {
+        if (LogConfig.configNames().contains(entry.getKey())) {
+          filteredConfig.put(entry.getKey(), entry.getValue());
+        } else {
+          LOG.warn("Property '{}' is not a valid Kafka topic config. It will be ignored.");
+        }
+      }
+    }
+    return filteredConfig;
```
bharathkk: Can we java8 'fy this?

PR #178, opened 2017-05-10
SAMZA-1258. Integration tests. Happy Path.

sborya:
This is reincarnation of PR #157 
It was not possible to merge/rebase that branch, so I had to create another one and reapply the changes.


samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
```diff
@@ -0,0 +1,558 @@
...
+    testStreamProcessor(new String[]{"1"});
+  }
+
+  @Test
+  public void testTwoStreamProcessors() {
+    testStreamProcessor(new String[]{"1", "2"});
+  }
+
+  @Test
+  public void testFiveStreamProcessors() {
```
navina: Can you split the tests into "setup" , "test" and "teardown" sections using @Before and @After annotations? It will be much cleaner and safer to run. 

Looking at this testStreamProcessor line here, I can't tell what your test is and what your assertions are.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
```diff
@@ -0,0 +1,558 @@
...
+   */
+  private void verifyNumMessages(String topic, final Map<Integer, Boolean> expectedValues, int expectedNumMessages) {
+    KafkaConsumer consumer = getKafkaConsumer();
+    consumer.subscribe(Collections.singletonList(topic));
+
+    Map<Integer, Boolean> map = new HashMap<>(expectedValues);
+    int count = 0;
+    int emptyPollCount = 0;
+
+    while (count < expectedNumMessages && emptyPollCount < 5) {
```
fredji97: It would be helpful to put a comment why we use 5 as the threshold for checking against emptyPollCount

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
```diff
@@ -0,0 +1,558 @@
...
+          map.put(Integer.valueOf(val), true);
+          count++;
+        }
+      } else {
+        emptyPollCount++;
+        System.out.println("empty polls " + emptyPollCount);
+      }
+    }
+    // filter out numbers we did not get
+    long numFalse = map.values().stream().filter(v -> !v).count();
```
fredji97: I think the behavior of this function has some dependency on the input parameter which is expectedValues. It assumes that the values for expectedValues map are always false in the input. I think the behavior of the function should not have an assumption on the values of the input.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
```diff
@@ -0,0 +1,558 @@
...
+      t2.join(1000);
+    } catch (InterruptedException e) {
+      Assert.fail("Failed to join finished threads:" + e.getLocalizedMessage());
+    }
+
+    // processor1 will read 40 events, and then processor1 and processor2 will read 80 events together,
+    // but the expected values are the same 0-79 - we should get each value one time.
+    // Meanwhile the number of events we gonna get is 80+40=120
+    Map<Integer, Boolean> expectedValues = new HashMap<>(2 * messageCount);
+    for (int i = 0; i < 2 * messageCount; i++) {
```
fredji97: I would move this line and the following two lines into verifyNumMessages(). And just put the expectedMessageCount as the second input for verifyNumMessages.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
```diff
@@ -0,0 +1,558 @@
...
+  }
+
+  // main test method for happy path with fixed number of processors
+  private void testStreamProcessor(String[] processorIds) {
+    final String testSystem = "test-system";
+    final String inputTopic = "numbers";
+    final String outputTopic = "output";
+    final int messageCount = 40;
+
+    final Map<String, String> map = createConfigs(testSystem, inputTopic, outputTopic, messageCount);
```
sborya: good point. I will change the names of the topics for each test. Same ZK should be fine.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
sborya: it is not easy to do here. Different tests require different setup parts. I will see what I can do, but no guarantees.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
```diff
@@ -0,0 +1,558 @@
...
+  /**
+   * Testing a basic identity stream task - reads data from a topic and writes it to another topic
+   * (without any modifications)
+   *
+   * <p>
+   * The standalone version in this test uses KafkaSystemFactory and it uses a SingleContainerGrouperFactory. Hence,
+   * no matter how many tasks are present, it will always be run in a single processor instance. This simplifies testing
+   */
+
+  public static StreamProcessorLifecycleListener listener = new StreamProcessorLifecycleListener() {
```
sborya: nope, removed.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
```diff
@@ -0,0 +1,558 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestZkStreamProcessor extends StandaloneIntegrationTestHarness {
+  /**
+   * Testing a basic identity stream task - reads data from a topic and writes it to another topic
+   * (without any modifications)
+   *
+   * <p>
+   * The standalone version in this test uses KafkaSystemFactory and it uses a SingleContainerGrouperFactory. Hence,
```
sborya: that's how it is done. The comment is wrong. I'll remove it.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
```diff
@@ -0,0 +1,558 @@
...
+    final Map<String, String> map = createConfigs(testSystem, inputTopic, outputTopic, messageCount);
+
+    // Note: createTopics needs to be called before creating a StreamProcessor. Otherwise it fails with a
+    // TopicExistsException since StreamProcessor auto-creates them.
+    createTopics(inputTopic, outputTopic);
+
+    // create a latch of the size == number of messages
+    TestStreamTask.endLatch = new CountDownLatch(messageCount);
+
+    StreamProcessor[] streamProcessors = new StreamProcessor[processorIds.length];
```
sborya: well, the tests require some much more finer control, which is not generally available through the applicaionRunner.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
sborya: well, I actually had to move it out of there, because only test knows what exact messages expected. Even if in most of them they are the same..

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
sborya: I will add a comment.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
sborya: as I said the list of expected values may change based on the test. So decision what are the values, should be done in the test itself. I can create an auxiliary method to make it look nicer.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
```diff
@@ -0,0 +1,558 @@
...
+    int count = 0;
+    int emptyPollCount = 0;
+
+    while (count < expectedNumMessages && emptyPollCount < 5) {
+      ConsumerRecords records = consumer.poll(5000);
+      if (!records.isEmpty()) {
+        Iterator<ConsumerRecord> iterator = records.iterator();
+        while (iterator.hasNext()) {
+          ConsumerRecord record = iterator.next();
+          //Assert.assertEquals(new String((byte[]) record.value()), String.valueOf(count));
```
sborya: done.

sborya: done.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
```diff
@@ -0,0 +1,291 @@
...
+    produceMessages(messageCount, inputTopic, messageCount);
+
+    // wait until all the events are consumed
+    // make sure it consumes all the messages from the first batch
+    waitUntilConsumedN(0);
+
+    // shutdown both
+    try {
+      synchronized (t1) {
+        t1.notify();
```
xinyuiscool: what's waiting on the thread object?

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
```diff
@@ -0,0 +1,291 @@
...
+      }
+      synchronized (t2) {
+        t2.notify();
```
xinyuiscool: same question above

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
sborya: this is the way to stop the thread. See runThread in the TestZkStreamProcessorBase.java

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
sborya: same answer as above :)

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessorBase.java
```diff
@@ -0,0 +1,270 @@
...
+import org.apache.samza.task.TaskCoordinator;
+import org.apache.samza.test.StandaloneIntegrationTestHarness;
+import org.apache.samza.test.StandaloneTestUtils;
+import org.apache.samza.zk.TestZkUtils;
+import org.junit.Assert;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class TestZkStreamProcessorBase extends StandaloneIntegrationTestHarness {
```
prateekm: There's already a StreamApplicationIntegrationTestHarness, can we refactor and use that one for these? I was thinking of combining that and StandaloneIntegrationTestHarness into a single class. Would be nice to avoid even more specific harnesses.

sborya: I am sure we can. They seem very compatible. But I think better refactoring would be to get rid of one of them completely. How did we end up with two in a first place??

I think the right approach would be to combine them together, and then refactor the tests. We can do it as a part of 0.14.

xinyuiscool:
As discussed with @sborya , the refactoring will be done in 0.14. The test looks good to me.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessor.java
```diff
@@ -0,0 +1,291 @@
...
+
+    // wait until the processor reports that it has started
+    try {
+      countDownLatch2.await(1000, TimeUnit.MILLISECONDS);
+    } catch (InterruptedException e) {
+      Assert.fail("got interrupted while waiting for the 2nd processor to start.");
+    }
+
+    // wait for at least one full debounce time to let the system to publish and distribute the new job model
+    TestZkUtils.sleepMs(3000);
```
prateekm: Is it possible to avoid this or reduce the debounce time here? I'd really prefer not adding more sleeps in integration tests.

PR #179, opened 2017-05-10
SAMZA-1276 : Adding customReporters passed by StreamProcessor into SamzaContainer

navina:
SamzaContainer is not adding custom reporters along side class-loaded reporters from config. This was missed in [SAMZA-1080](https://issues.apache.org/jira/browse/SAMZA-1080).


vjagadish1989 (approved):
approved, looks good to me

PR #180, opened 2017-05-10
SAMZA-1279: Do not process onProcessorChange if list of children is null.

sborya:


samza-core/src/main/java/org/apache/samza/zk/ZkControllerImpl.java
```diff
@@ -94,6 +94,12 @@ public void subscribeToProcessorChange() {
      */
     @Override
     public void handleChildChange(String parentPath, List<String> currentChildren) throws Exception {
+      if (currentChildren == null) {
+        // this may happen only in case of some exception in ZK. Exception should be visible in the logs.
```
prateekm: Update comment to describe when this can happen?

navina: can you rephrase to say- **if the parentPath doesn't exist, ZkClient doesn't throw exception. It will instead pass a "null" list.** instead of *some exception in ZK*

PR #181, opened 2017-05-10
SAMZA-1193. re-enable the test TestZkBarrierForVersionUpgrade.testZkBarrierForVersionUpgrade

sborya:
The code changed significantly. I've rerun the test multiple times both with gradle and intelij. It passed every time. I suggest we enable it back.

### DIFF
--- a/samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
+++ b/samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
@@ -18,6 +18,11 @@
  */
 package org.apache.samza.zk;
 
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 import junit.framework.Assert;
 import org.apache.samza.config.Config;
 import org.apache.samza.config.MapConfig;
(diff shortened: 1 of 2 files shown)


navina:
@sborya what was the change done here, other than changing the barrier name in the tests?

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
```diff
@@ -74,8 +73,7 @@ public static void teardown() {
     zkServer.teardown();
```
fredji97: zkServer.setup is called in Before, but zkServer.teardown is called in AfterClass. You may want to call zkServer.teardown in After for each test, like  zkServer.setup in Before, or make zkServer.setup called in BeforeClass.

sborya:
Moved init stuff into Before and After methods.

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
sborya: Good point. Fixed.

fredji97: thanks for addressing it!

PR #184, opened 2017-05-10
SAMZA-1283: Expose the buffered-message-size metric

xinyuiscool:
Regardless of whether we enable size limit for the consumer buffer, this metric helps to see what's the buffer size and make configuring size limit easier.


vjagadish1989 (approved):
approved.

looks good to me.

one question from the review:

The `size` computation overhead appears to be trivial since the `IncomingMessageEnvelope.size()` is proxying its call to `Message.size()` which is simply returning `java.nio.Buffer.size()`. 

Any idea on why the flag was added in the first place?

xinyuiscool:
Not sure. Maybe it was trying to not have any impact if the use doesn't need size limit. Since this metric helps a lot in memory analysis, let's add it.

PR #185, opened 2017-05-11
SAMZA-1286: Close zk connection in ZkController.stop()

sborya:


prateekm:
@sborya Does this need to go in 0.13?

PR #186, opened 2017-05-11
Increase the plan graph size

xinyuiscool:
Increase the canvas size to a standard 24 inch resolution and also the scaling factor.


vjagadish1989 (approved):
approved. looks good to me.

PR #189, opened 2017-05-12
SAMZA-1289: Default id generator if not configured

xinyuiscool:
Right now in standalone deployment we require the user to provide an id generator. Since most of the time the users can simply use the UUID generator for id generation, we should default it using that.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
+++ b/samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
@@ -153,7 +153,7 @@ public interface MessageStream<M> {
    * The merged {@link MessageStream} contains messages from all {@code streams} in the order they arrive.
    *
    * @param streams {@link MessageStream}s to be merged
-   * @param <T> the common type of messages in the streams
+   * @param <T> the type of messages in each of the streams
    * @return the merged {@link MessageStream}
    * @throws IllegalArgumentException if {@code streams} is empty
    */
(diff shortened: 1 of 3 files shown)


samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatClient.java
```diff
@@ -61,7 +61,7 @@ public ContainerHeartbeatClient(String coordinatorUrl, String executionEnvContai
 
   /**
    * Issues a heartbeat request to the coordinator and
```
jmakes: Sentence shouldn't end with "and"

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -153,6 +153,7 @@
    * The merged {@link MessageStream} contains messages from all {@code streams} in the order they arrive.
    *
    * @param streams {@link MessageStream}s to be merged
+   * @param <T> the common type of messages in the streams
```
vjagadish1989: Minor:  

`s/<T> the common type of messages in the streams/the type of messages in each of the streams/`

samza-core/src/main/java/org/apache/samza/config/ApplicationConfig.java
```diff
@@ -18,6 +18,9 @@
  */
 package org.apache.samza.config;
 
+import org.apache.samza.runtime.UUIDGenerator;
+
+
```
vjagadish1989: Minor: no new line

xinyuiscool: OH, this is needed since it's the only import in this file :)

navina:
@xinyuiscool In this version, we are maintaining backward compatibility with "processor.id" config. So, are you going to default to UUIDGenerator if processor.id is not configured?

xinyuiscool:
right, if processor.id is configured, we will not use the generator, otherwise it's using UUID generator. There is a check for this logic in every place which generates the processorId.

PR #190, opened 2017-05-12
SAMZA-1290: Fix javadoc for ApplicationRuner

jmakes:


jmakes:
@prateekm

samza-api/src/main/java/org/apache/samza/runtime/ApplicationRunner.java
```diff
@@ -29,7 +29,7 @@
 
 
 /**
- * A physical execution engine to deploy the config and jobs to run the {@link org.apache.samza.operators.StreamGraph}
+ * The primary means of interacting with the {@link org.apache.samza.application.StreamApplication} at runtime.
```
prateekm: s/interacting with/running?

jmakes: I was going to say that but it also kills and gets status. Do we consider those be under the umbrella of "running"?

prateekm: Good point. Maybe managing the execution? Interacting is fine too.

PR #195, opened 2017-05-18
SAMZA-1128 : Remove dependency of debounce timer from the CoordinationUtils

navina:
This patch addresses the following, apart from the main bug that barrier implementation uses a different scheduler that the Jobcoordinator's main thread. 
* Removes CoordinationUtils#getBarrier, BarrierForVersionUpgrade interface
* Renamed ZkBarrierForVersionUpgrade to ZkBarrier and introduces a listener ZkBarrierListener
* Simplified the ZkBarrier class and its integration test


navina:
@sborya @xinyuiscool Review please??

samza-core/src/main/java/org/apache/samza/zk/ZkBarrier.java
```diff
@@ -0,0 +1,214 @@
...
+  private final ZkUtils zkUtils;
+  private final BarrierKeyBuilder keyBuilder;
+
+  public enum State {
+    TIMED_OUT, DONE
+  }
+
+  private ZkBarrierListener barrierListener;
+
+  public ZkBarrier(String barrierRoot, ZkUtils zkUtils, ZkBarrierListener barrierListener) {
```
bharathkk: Just curious why this isn't final? do we expect this to change outside the constructor? 
If it is nullable, can we make it `Optional` instead? that way, inside your code you can get rid of null checks and do `barrierListener.ifPresent(...)` It would alleviate concerns over null checks on every code path and force defensive programming inherently. 
Thoughts?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrier.java
```diff
@@ -0,0 +1,214 @@
...
+   * Joins a shared barrier by registering under the barrier sub-tree in ZK
+   *
+   * @param version Version associated with the Barrier
+   * @param participantId Identifier of the participant
+   */
+  public void join(String version, String participantId) {
+    String barrierDonePath = keyBuilder.getBarrierStatePath(version);
+    zkUtils.getZkClient().subscribeDataChanges(barrierDonePath, new ZkBarrierReachedHandler(barrierDonePath, version));
+
+    // TODO: Handle ZkNodeExistsException
```
bharathkk: What do we have to do in order to handle this exception? Not sure if it is complicated. If yes, is it desirable to defer it? If not, can we either do it right away or get rid of the _todo_?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrier.java
```diff
@@ -0,0 +1,214 @@
...
+   * @param version Version associated with the Barrier
+   */
+  public void expire(String version) {
+    zkUtils.getZkClient().writeData(
+        keyBuilder.getBarrierStatePath(version),
+        State.TIMED_OUT);
+
+  }
+  /**
+   * Listener for changes to the list of participants. It is meant to be subscribed only by the creator of the barrier
```
bharathkk: Is there a way for us to enforce it? What happens if non-creators are listeners?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrier.java
```diff
@@ -0,0 +1,214 @@
...
+    public void handleChildChange(String parentPath, List<String> currentChildren) {
+      if (currentChildren == null) {
+        LOG.info("Got ZkBarrierChangeHandler handleChildChange with null currentChildren");
+        return;
+      }
+      LOG.debug("list of children in the barrier = " + parentPath + ":" + Arrays.toString(currentChildren.toArray()));
+      LOG.debug("list of children to compare against = " + parentPath + ":" + Arrays.toString(names.toArray()));
+
+      // check if all the expected participants are in
+      if (CollectionUtils.containsAll(currentChildren, names)) {
```
bharathkk: Correct me if I am wrong, `currentChildren` at the end of barrier will have all the participants right? 
If yes,  a quick check on the size as an additional condition might be performant since `Collectionutils.containsAll(...)` doesn't check for sizes.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierListener.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.zk;
+
+public interface ZkBarrierListener {
```
bharathkk: Can you please add java docs and method level docs as well?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -121,12 +128,33 @@ public void onProcessorChange(List<String> processors) {
         ScheduleAfterDebounceTime.DEBOUNCE_TIME_MS, () -> doOnProcessorChange(processors));
   }
 
-  public void doOnProcessorChange(List<String> processors) {
+  void doOnProcessorChange(List<String> processors) {
     // if list of processors is empty - it means we are called from 'onBecomeLeader'
-    generateNewJobModel(processors);
-    if (coordinatorListener != null) {
-      coordinatorListener.onJobModelExpired();
+    List<String> currentProcessorIds = getActualProcessorIds(processors);
```
bharathkk: What is the expected behaviour when `currentProcessorIds` is empty? Do we still generate job model, create barrier .. etc

samza-core/src/main/java/org/apache/samza/config/ZkConfig.java
```diff
@@ -28,8 +28,10 @@
 
   public static final int DEFAULT_CONNECTION_TIMEOUT_MS = 60000;
   public static final int DEFAULT_SESSION_TIMEOUT_MS = 30000;
-  public static final int DEFAULT_CONSENSUS_TIMEOUT_MS = 40000;
```
sborya: Why is this deleted? This is the new setting name.

samza-core/src/main/java/org/apache/samza/config/ZkConfig.java
```diff
@@ -49,7 +51,7 @@ public int getZkConnectionTimeoutMs() {
     return getInt(ZK_CONNECTION_TIMEOUT_MS, DEFAULT_CONNECTION_TIMEOUT_MS);
   }
 
-  public int getZkBarrierTimeoutMs() {
-    return getInt(ZK_CONSENSUS_TIMEOUT_MS, DEFAULT_CONSENSUS_TIMEOUT_MS);
+  public long getZkBarrierTimeoutMs() {
+    return getLong(ZK_BARRIER_TIMEOUT_MS, DEFAULT_BARRIER_TIMEOUT_MS);
```
sborya: int gets values up to 600 hours. that should be enough for the barrier.

samza-core/src/main/java/org/apache/samza/coordinator/BarrierForVersionUpgrade.java
```diff
@@ -1,48 +0,0 @@
...
-package org.apache.samza.coordinator;
-
-import java.util.List;
-
-
-/**
- * Interface for a barrier - to allow synchronization between different processors to switch to a newly published
- * JobModel.
- */
-public interface BarrierForVersionUpgrade {
```
sborya: This barrier is not designed for generic use, and will be refactored to be part of the JobPublishing. I don't see a reason to rename it.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrier.java
```diff
@@ -0,0 +1,214 @@
...
+import org.apache.commons.collections4.CollectionUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+import java.util.List;
+
+
+/**
+ * ZkBarrier is an implementation of distributed barrier, which guarantees that the expected barrier size and barrier
```
sborya: Nice comments !

samza-core/src/main/java/org/apache/samza/zk/ZkBarrier.java
```diff
@@ -0,0 +1,214 @@
...
+ *  |
+ *  |- barrier_{version1}/
+ *  |   |- barrier_state/
+ *  |   |  ([DONE|TIMED_OUT])
+ *  |   |- barrier_participants/
+ *  |   |   |- {id1}
+ *  |   |   |- {id2}
+ *  |   |   |-  ...
+ */
+public class ZkBarrier {
```
sborya: Same as above. Should be left with the old name.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -1,185 +0,0 @@
...
-
-
-/**
- * This class creates a barrier for version upgrade.
- * Barrier is started by the participant responsible for the upgrade. (start())
- * Each participant will mark its readiness and register for a notification when the barrier is reached. (waitFor())
- * If a timer (started in start()) goes off before the barrier is reached, all the participants will unsubscribe
- * from the notification and the barrier becomes invalid.
- */
-public class ZkBarrierForVersionUpgrade implements BarrierForVersionUpgrade {
```
sborya: Again, I don't understand why do we need to do this renaming now. 
This all will be part of the refactoring of the coordination utils.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -273,7 +269,7 @@ public void publishJobModelVersion(String oldVersion, String newVersion) {
     } catch (Exception e) {
       String msg = "publish job model version failed for new version = " + newVersion + "; old version = " + oldVersion;
       LOG.error(msg, e);
-      throw new SamzaException(msg);
+      throw new SamzaException(e);
```
sborya: Please leave msg.

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
```diff
@@ -1,154 +0,0 @@
...
-import org.apache.samza.coordinator.BarrierForVersionUpgrade;
-import org.apache.samza.coordinator.CoordinationServiceFactory;
-import org.apache.samza.coordinator.CoordinationUtils;
-import org.apache.samza.testUtils.EmbeddedZookeeper;
-import org.junit.After;
-import org.junit.Before;
-import org.junit.Test;
-
-
-public class TestZkBarrierForVersionUpgrade {
```
sborya: why rename?

samza-core/src/main/java/org/apache/samza/config/ZkConfig.java
navina: It was not being used. After rebasing with master (I think you had a few commits), it was not being used. So, I removed it. Let me know if I merged incorrectly. There were some conflicts during rebase.

samza-core/src/main/java/org/apache/samza/config/ZkConfig.java
navina: Hmm.. I don't remember why I changed this :P Ah ha.. And here is where you used `ZK_CONSENSUS_TIMEOUT_MS` . I will change these back.

samza-core/src/main/java/org/apache/samza/coordinator/BarrierForVersionUpgrade.java
navina: @sborya I removed the interface itself because it is not meant for any generic use. I didn't rename it. Can you explain what you think I have renamed :(

samza-core/src/main/java/org/apache/samza/zk/ZkBarrier.java
navina: You want to call this ZkBarrierForVersionUpgrade?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
navina: Same question as above.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
navina: I can use `new SamzaException(msg, e)` . Is that fine?

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
navina: In most codebase I have worked with, unit test classes are named as `Test<ClassName>`. The tests in this class are not unit tests. So, I prefer to name it different and more obvious on reading.

samza-core/src/main/java/org/apache/samza/config/ZkConfig.java
```diff
@@ -28,8 +28,10 @@
 
   public static final int DEFAULT_CONNECTION_TIMEOUT_MS = 60000;
   public static final int DEFAULT_SESSION_TIMEOUT_MS = 30000;
-  public static final int DEFAULT_CONSENSUS_TIMEOUT_MS = 40000;
-  
+
+  public static final String ZK_BARRIER_TIMEOUT_MS = "coordinator.zk.barrier-timeout-ms";
```
navina: Sure

samza-core/src/main/java/org/apache/samza/zk/ZkBarrier.java
navina: Are you referring to `barrierListener` ? I was going back and forth between having it as optional or mandatory. In the end though, it made sense to make it mandatory. I missed the part about making it final. 

`Optional` is a nice idea, although we haven't used it anywhere in our code-base. Let me give that shot. Sounds like a good programming practice. Thanks for the suggestion.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrier.java
navina: If the node already exists , that means we have a duplicate processor in the group. Ideally, the checks for duplicate processorId should happen elsewhere . See SAMZA-1304. 

> If yes, is it desirable to defer it? If not, can we either do it right away or get rid of the todo?
It may not be pretty straightforward to fix this cleanly. Aside from this, there are many other issues/conditions with zk client connection that we don't handle gracefully. So, I am ok with fixing this in 0.13.1 as a part of SAMZA-1304. I will add this JIRA number to the TODO to make sure we don't lose track of the fix.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrier.java
navina: We could make the creator register it as an anonymous class. That way, non-creators don't have an option of even using this class. But in reality, it doesn't prevent anyone from subscribing to any of the znode paths. 

Since this is internal to this implementation of the barrier, I suspect the probability of non-creators becoming listeners is really low.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrier.java
navina: Yes. I didn't know that `Collectionutils.containsAll(...)` doesn't check for size. Will change it. Thanks!

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierListener.java
navina: Sure .. sounds good!

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
navina: There is also another issue of currentProcessorIds not being unique. Both the issues are not addressed yet. I would prefer to tackle them in a separate PR. Thanks for bringing it up. I will create a JIRA to track it.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrier.java
```diff
@@ -0,0 +1,214 @@
...
+   *
+   * @param version Version associated with the Barrier
+   * @param participants List of expected participated for this barrier to complete
+   */
+  public void create(final String version, List<String> participants) {
+    String barrierRoot = keyBuilder.getBarrierRoot();
+    zkUtils.makeSurePersistentPathsExists(new String[]{
+        barrierRoot,
+        keyBuilder.getBarrierPath(version),
+        keyBuilder.getBarrierParticipantsPath(version),
```
navina: Sure.

navina:
@sborya @bharathkk I have addressed most of the comments, I think. Please review them. Thanks!

samza-core/src/main/java/org/apache/samza/zk/ZkKeyBuilder.java
```diff
@@ -51,6 +51,7 @@
 
   static final String PROCESSORS_PATH = "processors";
   static final String JOBMODEL_GENERATION_PATH = "JobModelGeneration";
+  static final String JOB_MODEL_UPGRADE_BARRIER = "jobModelUpgradeBarrier";
 
   public ZkKeyBuilder(String pathPrefix) {
     if (Strings.isNullOrEmpty(pathPrefix)) {
```
bharathkk: It is not related to your change, though i noticed we trim the `pathPrefix` before assignment. Ideally we should trim it before the check to make sure `"    "` isn't considered a valid prefix

navina: Thanks for pointing out. I also found some inaccurate unit tests for this case and another bug in Line 91. Let me create a separate PR for this.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -19,161 +19,162 @@
...
- * If a timer (started in start()) goes off before the barrier is reached, all the participants will unsubscribe
- * from the notification and the barrier becomes invalid.
+ * ZkBarrierForVersionUpgrade is an implementation of distributed barrier, which guarantees that the expected barrier
+ * size and barrier participants match before marking the barrier as complete.
+ * It also allows the caller to expire the barrier.
+ *
+ * This implementation is specifically tailored towards barrier support during jobmodel version upgrades. The participant
+ * responsible for the version upgrade starts the barrier by invoking {@link #create(String, List)}.
+ * Each participant in the list, then, joins the new barrier. When all listed participants {@link #join(String, String)}
+ * the barrier, the creator marks the barrier as done which signals the end of barrier.
```
navina: Ok

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -19,161 +19,162 @@
...
+ * The creator of the barrier can invoke {@link #expire(String)} to indicate that the barrier has timed_out or
```
navina: Ok

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -19,161 +19,162 @@
...
+        zkUtils.getZkClient().writeData(barrierDonePath, State.DONE); // this will trigger notifications
+        zkUtils.getZkClient().unsubscribeChildChanges(barrierDonePath, this);
       }
     }
   }
 
+  /**
+   * Listener for changes to the Barrier state. It is subscribed by all participants of the barrier, including the
+   * participant that creates the barrier.
+   * Barrier state can be of types - DONE or TIMED_OUT. It only registers to receive on valid state change notification.
```
navina: ok

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -19,161 +19,162 @@
...
-
+    public void handleChildChange(String parentPath, List<String> currentChildren) {
       if (currentChildren == null) {
-        LOG.info("Got handleChildChange with null currentChildren");
+        LOG.info("Got ZkBarrierChangeHandler handleChildChange with null currentChildren");
         return;
       }
-      LOG.info("list of children in the barrier = " + parentPath + ":" + Arrays.toString(currentChildren.toArray()));
-      LOG.info("list of children to compare against = " + parentPath + ":" + Arrays.toString(names.toArray()));
+      LOG.debug("list of children in the barrier = " + parentPath + ":" + Arrays.toString(currentChildren.toArray()));
```
navina: ok

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -182,4 +183,30 @@ public void handleDataDeleted(String dataPath)
       LOG.warn("barrier done got deleted at " + dataPath);
     }
   }
+
+  class BarrierKeyBuilder {
```
navina: Yeah. I spent an hour debugging a unit test because there was a typo in one of the keys. I didn't notice and there wasn't a doc for the actual value. So, decided to just add it here for now.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -182,4 +183,30 @@ public void handleDataDeleted(String dataPath)
...
+    private static final String BARRIER_PARTICIPANTS = "/barrier_participants";
+    private static final String BARRIER_STATE = "/barrier_state";
+    private final String barrierRoot;
+    BarrierKeyBuilder(String barrierRoot) {
+      this.barrierRoot = barrierRoot;
```
navina: Ok sounds good

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierListener.java
```diff
@@ -0,0 +1,47 @@
...
+   */
+  void onBarrierStateChanged(String version, ZkBarrierForVersionUpgrade.State state);
+
+  /**
+   * Invoked when Barrier encounters error
+   *
+   * @param version Version associated with the Barrier
+   * @param t Throwable describing the cause of the barrier error
+   */
+  void onBarrierError(String version, Throwable t);
```
navina: Ah.. This method can be called when the barrier throws any exception. For now, we are not actually handling the errors from the barrier. For example, if there is a duplicate processor, "join" will throw an error saying ZkNodeExists. We don't handle it now, but potentially we should be able to handle it later on and cleanly shutdown the processor or handle intelligently. For now, I didn't want to include error handling in ZkBarrier. At least not in the same PR.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -19,161 +19,162 @@
...
     this.zkUtils = zkUtils;
-    barrierPrefix = zkUtils.getKeyBuilder().getJobModelVersionBarrierPrefix(barrierId);
-    this.debounceTimer = debounceTimer;
-    this.barrierTimeoutMS = barrierTimeoutMS;
-  }
-
-  protected long getBarrierTimeOutMs() {
-    return barrierTimeoutMS;
+    this.keyBuilder = new BarrierKeyBuilder(barrierRoot);
+    this.barrierListener = Optional.ofNullable(barrierListener);
```
navina: > Please use different names - having same name, but different types might be misleading.

Are you referring to keyBuilder variable name ?

> (besides, I am not sure on value of having an Optional here, since it doesn't protect you from calling .get() on this object and throws a confusing NoSuchElementException instead of NullPointerException. So it only changes ifNotNull for ifPresent...but ..ok)

I agree there is a possibility of NoSuchElementException when using Optionals. But I am not following what you are suggesting I change? Can you elaborate your suggestion?

samza-core/src/main/java/org/apache/samza/zk/ZkControllerImpl.java
```diff
@@ -114,12 +114,6 @@ public void handleChildChange(String parentPath, List<String> currentChildren) t
   }
 
   class ZkJobModelVersionChangeHandler implements IZkDataListener {
-    /**
-     * called when job model version gets updated
```
navina: Done

navina:
@sborya I addressed all your comments. I accidentally rebased though. Sorry about that.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
sborya: ZkBarrierListener barrierListener and Optional<ZkBarrierListener> barrierListener.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierListener.java
sborya: ok.

PR #197, opened 2017-05-19
SAMZA-1298: Create zk path.

sborya:
if ZK path contains extra path at the end, it needs to be created in ZK at first connect.


samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -84,8 +86,50 @@ public static ZkConnection createZkConnection(String zkConnectString, int sessio
...
-  public static ZkClient createZkClient(ZkConnection zkConnection, int connectionTimeoutMs) {
-    return new ZkClient(zkConnection, connectionTimeoutMs);
+  /**
+   * helper method to create zkClient
+   * @param connectString - zkConnect string
+   * @param sessionTimeoutMS - session timeout
+   * @param connectionTimeoutMs - connection timeout
+   * @return zkClient object
+   */
+  public static ZkClient createZkClient(String connectString, int sessionTimeoutMS, int connectionTimeoutMs) {
```
vjagadish1989: Not sure how much can be done within the scope of this PR. I'm happy to do the refactoring as a part of a later PR, and submit one.

1. Can we move the `initPath` inside the `ZkUtils`? That way, the lifecycle of initializing `ZkClient` is not spread across multiple places? Optionally, `ZkUtils` can take in a `ZkConfig` object in its constructor. 

2. Also, can we get rid off two ways (currently exposed as 2 factory methods) to obtain an instance of `ZkClient` where in one case you obtained a pre-inited object, and in the other you don't?

3. Also, it's slightly strange to have the name `ZkUtils` , Usually `Util` classes are a bunch of static methods, with a private constructor. Instead it should be something like `ZkClientWrapper` or `SamzaZkClient` that is a thin wrapper around `ZkClient` with well-defined responsibilities.

sborya: I moved both createClient and createPath to the CoordinationUtilsFactory, where it fits better.

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
```diff
@@ -18,27 +18,69 @@
...
+  }
+
+  /**
+   * helper method to create zkClient
+   * @param connectString - zkConnect string
+   * @param sessionTimeoutMS - session timeout
+   * @param connectionTimeoutMs - connection timeout
+   * @return zkClient object
+   */
+  public static ZkClient createZkClient(String connectString, int sessionTimeoutMS, int connectionTimeoutMs) {
```
vjagadish1989: Why do we need this method here? Can be removed. It's slightly odd to have a static method that simply proxies it's parameters to an underlying public constructor

Appears to just wrap around a `Zkclient`? Can't users simply do a `new ZkClient(param1, param2, param3)`

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
```diff
@@ -18,27 +18,69 @@
...
+import org.apache.zookeeper.client.ConnectStringParser;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 public class ZkCoordinationServiceFactory implements CoordinationServiceFactory {
+  public final static Logger LOG = LoggerFactory.getLogger(ZkCoordinationServiceFactory.class);
+
   // TODO - Why should this method be synchronized?
   synchronized public CoordinationUtils getCoordinationService(String groupId, String participantId, Config config) {
```
vjagadish1989: Should this be `ZkConfig`?

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
```diff
@@ -18,27 +18,69 @@
...
 import org.apache.samza.config.ZkConfig;
 import org.apache.samza.coordinator.CoordinationServiceFactory;
 import org.apache.samza.coordinator.CoordinationUtils;
+import org.apache.zookeeper.client.ConnectStringParser;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 public class ZkCoordinationServiceFactory implements CoordinationServiceFactory {
+  public final static Logger LOG = LoggerFactory.getLogger(ZkCoordinationServiceFactory.class);
```
vjagadish1989: Minor: should be private?

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
```diff
@@ -18,27 +18,69 @@
...
   // TODO - Why should this method be synchronized?
   synchronized public CoordinationUtils getCoordinationService(String groupId, String participantId, Config config) {
     ZkConfig zkConfig = new ZkConfig(config);
-    ZkClient zkClient;
+
+    ZkClient zkClient = createZkClient(zkConfig.getZkConnect(),
+        zkConfig.getZkSessionTimeoutMs(), zkConfig.getZkConnectionTimeoutMs());
+
+    // make sure the 'path' exists
+    createZkPath(zkConfig.getZkConnect(), zkClient);
```
vjagadish1989: Can this be a private method? Does n't seem to be used anywhere outside this? Also, seems like an odd method to have in a factory class.

vjagadish1989:
Final set of comments, most of these should be minor changes. Do let me know if you need to defer others to a later PR (in the interest of 0.13 release schedule)

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
sborya: sure.

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
sborya: well, this is a factory, I think it should be Config.

vjagadish1989: Makes sense, it was not obvious that this was an overridden method. 

Maybe, add `@Override` in the annotation?

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
sborya: Seems like there are multiple views on the correct location of this method. Since there is a jira for refactoring the whole coordination utils matter, I think we should leave it be here for now.

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
sborya: Seems like there are multiple views on the correct location of this method. Since there is a jira for refactoring the whole coordination utils matter, I think we should leave it be here for now.

sborya:
Location of these methods will be reviewed as the part of SAMZA-1173.

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
vjagadish1989: Okay, I'll let you agree with @navina and address follow-ups as a part of SAMZA-1173

PR #203, opened 2017-05-25
SAMZA-1309. Debounce time config

sborya:


docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -464,6 +464,17 @@
                 <tr>
                     <th colspan="3" class="section" id="task"><a href="../api/overview.html">Task configuration</a></th>
                 </tr>
+                <tr>
+                    <td class="property" id="job.debounce.time.ms">job.debounce.time.ms</td>
+                    <td class="default"> 2000 </td>
+                    <td class="description">
+                        How long the Leader processor will wait before performing some notification based action (like JobModel recalculation) from the moment the notification came.
```
navina: Isn't jobmodel recalculation or rather group membership the only time we use this value? 
I think it should be sufficient to say - "How long the leader processor will wait before recalculating job model on processor change " (may be with better phrasing)

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -464,6 +464,17 @@
...
+                    </td>
+                </tr>
+                <tr>
+                    <th colspan="3" class="section" id="task"><a href="../api/overview.html">Task configuration</a></th>
```
navina: This line is same as Line 465. Is it needed?

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -48,7 +48,7 @@
   // Action name when the Processor membership changes
   public static final String ON_PROCESSOR_CHANGE = "OnProcessorChange";
 
-  public static final int DEBOUNCE_TIME_MS = 2000;
+  public static final int DEFAULT_DEBOUNCE_TIME_MS = 2000;
```
navina: Can you move this definition to JobConfig or ZkConfig? It looks like ZkJobCoordinator is the only place accessing this value. 

If we move to either of the config classes, we can avoid referring to the default value in Line 84 of ZkJobCoordinator. It will be encapsulated within the config class and the api will jobconfig.getDebounceTimeMs().

docs/learn/documentation/versioned/jobs/configuration-table.html
sborya: Should be sufficient, I was thinking about LeaderElection, but is using the same event anyway.
Will change.

docs/learn/documentation/versioned/jobs/configuration-table.html
sborya: ay-yay-yay, good catch. Copy-paste error. Removed.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
sborya: yep, looks better. I will keep it in JobConfig, because, in theory, this value may be needed also for non ZK coordination.

sborya:
@navina addressed the comments.

sborya:
@navina . If all the questions are addressed, can we please commit it? Thanks.

navina:
@sborya I saw this PR being closed this morning and assumed somebody committed it. Are you sure this wasn't committed before? If not, I will commit it now.

PR #206, opened 2017-05-26
SAMZA-1313. Rename StandaloneJobCoordinator to PassthroughJobCoordinator

sborya:


samza-test/src/test/java/org/apache/samza/test/StandaloneTestUtils.java
```diff
@@ -25,7 +25,7 @@
 import java.util.Map;
 
 public class StandaloneTestUtils {
-  private static final String STANDALONE_JOB_COORDINATOR_FACTORY = "org.apache.samza.standalone.StandaloneJobCoordinatorFactory";
+  private static final String PASSTHROUGH_JOB_COORDINATOR_FACTORY = "org.apache.samza.standalone.PassthroughJobCoordinatorFactory";
```
navina: This class name is kind of misleading. But now it is being used in Zk test classes and pass-through test classes. So, let's not change it right away.

PR #207, opened 2017-05-31
SAMZA-1312: Add Control Messages and Intermediate Stream Serde

xinyuiscool:
In this patch, we add the control message types which includes:
* EndOfStreamMessage
* WatermarkMessage

To support in-band data and control messages, we provide a wrapper serde (IntermediateMessageSerde) to serialize/deserialize data/control messages based on message type byte (first byte in the intermediate stream message). The format of the message is defined in SAMZA-1312. The patch integrates this serde with SerdeManager.

Tested in example jobs deployed locally and works as expected.


vjagadish1989 (approved):
looks good to me.

approved.

one f/b:

There appear to be the following information we embed in our `Config` object:
1. Configs specified to us by the user
2. Configs that we clobber/ rewrite in the `ExecutionPlanner`
3. Configs that are generated and are purely internal to the Samza run-time. For instance, `StreamConfig.IS_INTERMEDIATE_FROM_STREAM_ID`
4. Misc. stuff like our generated JSON plan

2,3  corresponds to the serialization of the graph itself. Would be nice to have a plan to separate these. (not necessarily this PR)

PR #211, opened 2017-06-02
SAMZA-1271: Guarantee predictable, deterministic order for operator initialization and finalization

vjagadish1989:
Currently, the order of initialization of operators in the Samza high level API is not deterministic. The non-determinism arises from two primary causes:

- No fixed order of iteration for all subscribed `OperatorSpec`s for a given `MessageStream`
- No fixed order of iteration for all the `OperatorImpl`s in the `OperatorImplGraph`

We aim to provide the following 2 guarantees in this patch:
For any 2 operators A, B in the graph, if B consumes the output of A:
- A is initialized before B is initialized
- A is finalized only after B is finalized

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/operators/functions/ClosableFunction.java
+++ b/samza-api/src/main/java/org/apache/samza/operators/functions/ClosableFunction.java
@@ -26,6 +26,10 @@ import org.apache.samza.annotation.InterfaceStability;
  *
  * <p> Implement {@link #close()} to free resources used during the execution of the function, clean up state etc.
  *
+ * <p> Order of finalization: {@link ClosableFunction}s are invoked in the reverse topological order of operators in the
+ * {@link org.apache.samza.operators.StreamGraph}. For any two operators A and B in the graph, if operator B consumes results
+ * from operator A, then operator B is guaranteed to be closed before operator A.
+ *
  */
 @InterfaceStability.Unstable
 public interface ClosableFunction {
--- a/samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
+++ b/samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
@@ -18,6 +18,7 @@
  */
 package org.apache.samza.operators.impl;
 
+import com.google.common.collect.Lists;
 import org.apache.samza.config.Config;
 import org.apache.samza.operators.MessageStreamImpl;
 import org.apache.samza.operators.StreamGraphImpl;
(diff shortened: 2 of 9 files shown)


vjagadish1989:
Ping @prateekm , @nickpan47 , @xinyuiscool for reviews!

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
```diff
@@ -18,6 +18,7 @@
  */
 package org.apache.samza.operators.impl;
 
+import com.google.common.collect.Lists;
```
prateekm: Minor: Maybe Collections.reverse?

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpecs.java
```diff
@@ -101,6 +106,12 @@ public void init(Config config, TaskContext context) {
       public void init(Config config, TaskContext context) {
         filterFn.init(config, context);
       }
+
+      @Override
+      public void close() {
+        filterFn.close();
+      }
+
```
prateekm: Delete extra newline.

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpecs.java
```diff
@@ -71,6 +71,11 @@ private OperatorSpecs() {}
       public void init(Config config, TaskContext context) {
         mapFn.init(config, context);
       }
+
+      @Override
+      public void close() {
```
prateekm: Good catch, thanks.

samza-api/src/main/java/org/apache/samza/operators/functions/ClosableFunction.java
```diff
@@ -26,6 +26,10 @@
  *
  * <p> Implement {@link #close()} to free resources used during the execution of the function, clean up state etc.
  *
+ * <p> Order of finalization: {@link ClosableFunction}s are invoked in the reverse topological order of operators in the
```
prateekm: "Order of closing", "are closed"

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
vjagadish1989: `Lists.reverse` is nicer as it does not mutate the passed in list. We appear to be having a guava dependency anyways in samza-core

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpecs.java
vjagadish1989: Yeah, the unit tests caught 'em

samza-api/src/main/java/org/apache/samza/operators/functions/ClosableFunction.java
vjagadish1989: Fixed!

PR #214, opened 2017-06-05
SAMZA-1293: Enable partition expansion of input streams (SEP-5)

lindong28:


lindong28:
@navina @nickpan47 The current patch still adds the new method in the existing interface instead of creating a new interface. I think the approach is probably cleaner. Here is my comparison of the two solutions:

Solution 1:
- Add new method to interface `SystemStreamPartitionGrouper`
- Add new config `fix.partition.enabled`
- If fix.partition.enabled = true, use the new method to group partition. Otherwise, use the existing method.
- Enable partition monitor if `fix.partition.enabled` = true.

Solution 2:
- Add new interface `SystemStreamPartitionGrouperWithFixedPartition` with the new method
- Add new classes `GroupByPartitionWithFixedTaskNum`, `GroupBySystemStreamPartitionWithFixedTaskNum`, `GroupByPartitionWithFixedTaskNumFactory` and `GroupBySystemStreamPartitionWithFixedTaskNumFactory`
- If the grouper class implements the new interface, use the new method to group partition. Otherwise, use the existing method.
- User needs to explicitly set `job.coordinator.monitor-partition-change` to true to enable partition monitor.

Previously we wanted to do Solution 2 because 1) we want to keep the same code style as when we added BalancingTaskNameGrouper; and 2) we don't determine which method to call by checking with the new method is defined. 

But now I think Solution 1 is better because 1) number of new classes we need to add is much smaller; 2) use a new config to determine the choice of method without having to peek into the class definition; 3) the use of new config `fix.partition.enabled` is more explicit than re-using existing config `job.coordinator.monitor-partition-change`; 4) Less code duplication between the new grouper classes and the existing grouper classes.

Do you think Solution 1 would be cleaner?

prateekm:
@lindong28 Does this still need to be reviewed/merged? If not, can you close this PR?

lindong28:
@prateekm Sure, I closed the PR now.

PR #216, opened 2017-06-07
SAMZA-1326: Revert to use getGlobalAppId for zk path

sborya:


samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -65,7 +65,7 @@
   public ZkJobCoordinator(Config config) {
     this.config = config;
     ZkConfig zkConfig = new ZkConfig(config);
-    ZkKeyBuilder keyBuilder = new ZkKeyBuilder(new ApplicationConfig(config).getAppId());
+    ZkKeyBuilder keyBuilder = new ZkKeyBuilder(new ApplicationConfig(config).getGlobalAppId());
```
navina: What is the difference between appId and globalAppId?

prateekm: Yeah, doesn't make sense to have a appId and a globalAppId. Shouldn't the appId be unique anyway?

navina: Yeah. That's what I thought. Reading the code, it seems like appId is aka jobId.

PR #217, opened 2017-06-07
SAMZA-1327: create zk namespace if does not exist

sborya:


sborya:
Closing the request, because we decided not to create the name space for now.

PR #218, opened 2017-06-07
SAMZA-1327: fail if namespace specified in the connection string does not exist

sborya:


samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
```diff
@@ -56,31 +52,38 @@ synchronized public CoordinationUtils getCoordinationService(String groupId, Str
...
     try {
-      return new ZkClient(connectString, sessionTimeoutMS, connectionTimeoutMs);
+      zkClient = new ZkClient(connectString, sessionTimeoutMS, connectionTimeoutMs);
     } catch (Exception e) {
       // ZkClient constructor may throw a variety of different exceptions, not all of them Zk based.
       throw new SamzaException("zkClient failed to connect to ZK at :" + connectString, e);
     }
+
+    // make sure the namespace in zk exists (if specified)
+    validateZkNameSpace(connectString, zkClient);
```
xinyuiscool: do we need to create the name space path nodes?

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
```diff
@@ -56,31 +52,38 @@ synchronized public CoordinationUtils getCoordinationService(String groupId, Str
...
-      }
+    if (Strings.isNullOrEmpty(path)) {
+      return; // no namespace path
     }
-  }
 
+    LOG.info("connectString = " + zkConnect + "; path =" + path);
+
+    // if namespace specified (path above) but "/" does not exists, we will fail
+    if (!zkClient.exists("/")) {
```
prateekm: What is this doing? Is it testing that "/" exists under namespace? 
Also, what does it mean if there's no namespace path at line 79?

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
sborya: We decided not to do it. Just fail. So they would be created manually.

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
sborya: "/" is the chrooted root. So if it exists (with namespace) it means the namespace exists.
This is where we got it wrong before :)

prateekm: Oh ok. So what does it mean if the chrooted path is empty? Is that an exception too?

sborya: If there is not chrootpath - do nothing. It is a zookeeper tree rooted at the normal root.
If it is not empy, then the root is actually host:port/namespace/, but access wise it should be accessed as "/", so it must exist.

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
```diff
@@ -56,31 +52,38 @@ synchronized public CoordinationUtils getCoordinationService(String groupId, Str
...
+
+    // make sure the namespace in zk exists (if specified)
+    validateZkNameSpace(connectString, zkClient);
+
+    return zkClient;
   }
 
   /**
-   * if ZkConnectString contains some path at the end, it needs to be created when connecting for the first time.
+   * if ZkConnectString contains namespace path at the end, it needs to be created when connecting for the first time.
```
prateekm: Update documentation.

samza-core/src/test/java/org/apache/samza/zk/TestZkNamespace.java
```diff
@@ -0,0 +1,146 @@
...
+    // test if the new root exists
+    Assert.assertTrue(zkClient.exists("/"));
+    Assert.assertTrue(zkClient.exists("/test"));
+    Assert.assertTrue(zkClient.exists("/test/test1"));
+  }
+
+  @Test
+  public void testValidateFailZkNameSpace() {
+    try {
+      testFailIfNameSpaceMissing("/zkNameSpace");
```
navina: @sborya It is nice that you have tried to refactor common tasks into private methods. But it makes the tests very hard to read. Can you please consider doing away with testFailIfNameSpaceMissing ? 

I also think each of these within testValidateFailZkNameSpace warrants a separate test method , instead of putting all into one test method. Is there any reason we need to that?

samza-core/src/test/java/org/apache/samza/zk/TestZkNamespace.java
```diff
@@ -0,0 +1,146 @@
...
+import org.I0Itec.zkclient.ZkConnection;
+import org.apache.samza.SamzaException;
+import org.apache.samza.testUtils.EmbeddedZookeeper;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+
+public class TestZkNamespace {
```
navina: Given that a lot of us in the team are less familiar with ZK semantics, can you please add some documentation on what a namespace in ZK means?

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
sborya: done.

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
```diff
@@ -33,15 +33,11 @@
 public class ZkCoordinationServiceFactory implements CoordinationServiceFactory {
   private final static Logger LOG = LoggerFactory.getLogger(ZkCoordinationServiceFactory.class);
 
-  // TODO - Why should this method be synchronized?
   synchronized public CoordinationUtils getCoordinationService(String groupId, String participantId, Config config) {
```
sborya: It used to be called from multiple places, but now we call it only from one place. So I don't think we need to sync it.

samza-core/src/test/java/org/apache/samza/zk/TestZkNamespace.java
sborya: added.

samza-core/src/test/java/org/apache/samza/zk/TestZkNamespace.java
sborya: I've updated the negative tests. All the positive tests have a lot of common code, so it is better to have an aux method for them.

PR #221, opened 2017-06-09
SAMZA-1331: Optimize Snapshot class constructor

logarithm:
In some of our workloads (where we need to gather samza metrics five times per minute) `SlidingTimeWindowReservoir.getSnapshot()` method takes up to 10% of processor time.

Almost all of `getSnapshot` time is taken by Collections.sort method. So, the complexity of Snapshot constructor is O(NlogN) + iteration through passed values.

This ticket asks to improve the performance of Snapshot constructor but keep the performance of all other methods at least on the same level.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/metrics/Snapshot.java
+++ b/samza-api/src/main/java/org/apache/samza/metrics/Snapshot.java
@@ -21,19 +21,47 @@ package org.apache.samza.metrics;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 
 /**
  * A statistical snapshot of a collection of values
  */
 public class Snapshot {
   private final ArrayList<Long> values;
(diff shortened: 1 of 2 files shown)


vjagadish1989 (approved):
looks good to me. +1. Thanks for noticing this, and fixing it! 

approved

vjagadish1989:
>> In some of our workloads (where we need to gather samza metrics five times per minute) SlidingTimeWindowReservoir.getSnapshot() method takes up to 10% of processor time.

For my understanding, Do you mean this method takes 10% of the CPU time of the entire container process?

logarithm:
@vjagadish1989 Yes, correct. But as I said we gather metrics 5 times per minute.

vjagadish1989:
Thanks again! I'll commit it in a couple of days if no one has any other feedback.

PR #223, opened 2017-06-12
SAMZA-1324 : Add a metricsreporter lifecycle for JobCoordinator component of StreamProcessor

PawasChhokra:
Added a metrics class for ZK based job coordinator that reports a few metrics.


PawasChhokra:
@navina @sborya

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -235,6 +263,7 @@ private JobModel generateNewJobModel(List<String> processors) {
     @Override
     public void onBecomingLeader() {
       LOG.info("ZkJobCoordinator::onBecomeLeader - I became the leader!");
+      metrics.leaderElection.inc();
```
navina: @PawasChhokra Looks like this metric counts the number of times a processor has become the leader. The metric name "leader-election" can be easily misinterpreted as the number of times the processor participated in leader election. Can you please rename this metric?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -170,6 +196,8 @@ public void onNewJobModelAvailable(final String version) {
     debounceTimer.scheduleAfterDebounceTime(ScheduleAfterDebounceTime.JOB_MODEL_VERSION_CHANGE, 0, () ->
       {
         LOG.info("pid=" + processorId + "new JobModel available");
+        metrics.newJobModel.inc();
```
navina: Can you please explain the significance of the "new-jobmodel-generated" metric? Does it indicate the number of times jobmodel was generated or the number of times this processor was notified about a jobmodel change?

If it is the former, I think this metric should be incremented only by the leader. If it is the latter, perhaps rename the metric as "new-jobmodel-available" ?

docs/learn/documentation/versioned/container/metrics-table.html
```diff
@@ -887,6 +892,46 @@
...
+        <td>reads</td>
+        <td>Number of reads from Zookeeper</td>
+    </tr>
+    <tr>
+        <td>writes</td>
+        <td>Number of writes to Zookeeper</td>
+    </tr>
+    <tr>
+        <td>subscriptions</td>
+        <td>Number of subscriptions to Zookeeper</td>
```
navina: Maybe rephrase as "number of subscriptions to znodes in Zookeeper" ?

docs/learn/documentation/versioned/container/metrics-table.html
```diff
@@ -887,6 +892,46 @@
...
+        <td>zk-connection-error</td>
+        <td>Number of Zookeeper connection errors</td>
+    </tr>
+    <tr>
+        <td>is-leader</td>
+        <td>Denotes if the processor is a leader or not (1/0)</td>
+    </tr>
+    <tr>
+        <td>barrier-creation</td>
+        <td>Number of times a barrier was created by Zookeeper</td>
```
navina: s/created by Zookeeper/created by the leader/

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -93,7 +93,7 @@ public void create(final String version, List<String> participants) {
 
     // subscribe for participant's list changes
     LOG.info("Subscribing for child changes at " + barrierParticipantsPath);
-    zkUtils.getZkClient().subscribeChildChanges(barrierParticipantsPath, new ZkBarrierChangeHandler(version, participants));
+    zkUtils.subscribeChildChanges(barrierParticipantsPath, new ZkBarrierChangeHandler(version, participants));
```
navina: Thanks for cleaning up here! :+1:

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -54,25 +60,29 @@
...
-
   private int debounceTimeMs;
+  long startTime;
 
   public ZkJobCoordinator(Config config) {
     this.config = config;
     ZkConfig zkConfig = new ZkConfig(config);
-    ZkKeyBuilder keyBuilder = new ZkKeyBuilder(new ApplicationConfig(config).getGlobalAppId());
+    ZkKeyBuilder keyBuilder = new ZkKeyBuilder(new ApplicationConfig(config).getAppId());
+    ReadableMetricsRegistry metricsRegistry = new MetricsRegistryMap();
```
navina: One suggestion (I forgot to mention before): Can you instantiate the `metricsRegistry` in the `ZkJobCoordinatorFactory` and pass it in as a constructor parameter?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -83,11 +93,13 @@ public ZkJobCoordinator(Config config) {
         zkUtils,
         new ZkBarrierListenerImpl());
     this.debounceTimeMs = new JobConfig(config).getDebounceTimeMs();
+    this.reporters = MetricsReporterLoader.getMetricsReporters(new MetricsConfig(config), "samza-container-" + processorId);
```
navina: Let's get rid of "samza-container-" and just use the processorId directly ?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -104,14 +116,29 @@ public synchronized void stop() {
       coordinatorListener.onJobModelExpired();
     }
 
+    metrics.isLeader.set(0);
```
navina: Can you please add a comment here explaining why we are resetting the gauge here?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -104,14 +116,29 @@ public synchronized void stop() {
...
     debounceTimer.stopScheduler();
     zkController.stop();
 
+    shutdownMetrics();
     if (coordinatorListener != null) {
       coordinatorListener.onCoordinatorStop();
     }
   }
 
+  public void startMetrics() {
```
navina: I think this method scope should be private.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -104,14 +116,29 @@ public synchronized void stop() {
...
+    for (MetricsReporter reporter:reporters.values()) {
```
navina: nit: add space after ":" 
(surprised there is no checkstyle error for this! )

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -104,14 +116,29 @@ public synchronized void stop() {
...
+      reporter.register("samza-container-" + processorId, metrics.getMetricsRegistry());
```
navina: s/samza-container-/job-coordinator-/

This parameter represent the source of this set of metrics. That is the jobcoordinator component of the processor.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -258,7 +290,9 @@ public void onBarrierCreated(String version) {
 
     public void onBarrierStateChanged(final String version, ZkBarrierForVersionUpgrade.State state) {
       LOG.info("JobModel version " + version + " obtained consensus successfully!");
+      metrics.barrierStateChange.inc();
       if (ZkBarrierForVersionUpgrade.State.DONE.equals(state)) {
+        metrics.singleBarrierRebalancingTime.update(System.nanoTime() - startTime);
```
navina: I think we should update the timer metric even in case of a time out state change. That way, we can correlate the metric value with the state change - if the difference in time == barrier_timeout-ms, then the barrier has timed out. What do you think? 

I also want to point out a potential issue here: `startTime` is initialized in `doProcessorChange` which is only invoked in the leader JC. For all other processors, it will be 0 and result high and incorrect timer value. Perhaps it isn't obvious from the code. But the leader also receives barrierCreation notification. So, you can create the `startTime` within the handler and initialize it in `ZkBarrierListenerImpl#onBarrierCreated`. Does this sound reasonable to you?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorMetrics.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class ZkJobCoordinatorMetrics extends MetricsBase {
+
+  private final ReadableMetricsRegistry metricsRegistry;
+
+  public final Counter reads;
+  public final Counter writes;
+  public final Counter subscriptions;
+  public final Counter zkConnectionError;
+  public final Gauge<Integer> isLeader;
```
navina: In general, it is useful to add some javadoc around member variables (esp. in public facing APIs). However, in this case, most of the variable names are obvious in terms of their functionality. Can you please add some doc for `isLeader`, `barrierCreation`, `barrierError` and `singleBarrierRebalancingTime` ?  Thanks!

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -668,6 +668,7 @@ class SamzaContainer(
     try {
       info("Starting container.")
 
+      var startTime = System.nanoTime()
```
navina: `var` is used for mutable members. In this, you can simply use `val` because you are not going to modify `startTime`

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
```diff
@@ -52,9 +53,10 @@ public static void test() {
   @Before
   public void testSetup() {
     ZkClient zkClient = new ZkClient(testZkConnectionString, ZkConfig.DEFAULT_SESSION_TIMEOUT_MS, ZkConfig.DEFAULT_CONNECTION_TIMEOUT_MS);
-    this.zkUtils = new ZkUtils(new ZkKeyBuilder("group1"), zkClient, ZkConfig.DEFAULT_CONNECTION_TIMEOUT_MS);
+    ZkJobCoordinatorMetrics metrics = new ZkJobCoordinatorMetrics(new MetricsRegistryMap());
```
navina: Maybe use `new NoOpMetricsRegistry()` if you are not going to assert any metrics ?

samza-core/src/test/java/org/apache/samza/zk/TestZkLeaderElector.java
```diff
@@ -433,9 +434,10 @@ public void testAmILeader() {
 
   private ZkUtils getZkUtilsWithNewClient() {
     ZkClient zkClient = ZkCoordinationServiceFactory.createZkClient(testZkConnectionString, SESSION_TIMEOUT_MS, CONNECTION_TIMEOUT_MS);
+    ZkJobCoordinatorMetrics metrics = new ZkJobCoordinatorMetrics(new MetricsRegistryMap());
```
navina: Same comment as above

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
```diff
@@ -67,10 +68,11 @@ public void testSetup() {
       // Do nothing
     }
 
+    ZkJobCoordinatorMetrics metrics = new ZkJobCoordinatorMetrics(new MetricsRegistryMap());
```
navina: Same comment as above

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
PawasChhokra: Sounds reasonable. Changed the timer metric to update after both state changes. Also fixed the startTime issue. Thanks!

docs/learn/documentation/versioned/container/metrics-table.html
PawasChhokra: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
PawasChhokra: You're welcome!

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorMetrics.java
PawasChhokra: Added javadocs.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
PawasChhokra: Changed to val.

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
PawasChhokra: Done.

samza-core/src/test/java/org/apache/samza/zk/TestZkLeaderElector.java
PawasChhokra: Done.

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
PawasChhokra: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
PawasChhokra: Replaced samza-container with job-coordinator.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
PawasChhokra: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
PawasChhokra: Changed to private. Sorry about this.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
PawasChhokra: Added the comment. We are setting the isLeader metric to false here because when the stream processor shuts down, it does not remain the leader anymore

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
PawasChhokra: Removed samza-container.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
PawasChhokra: Instantiated in the ZkJobCoordinatorFactory.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -48,32 +54,33 @@
...
   private JobModel newJobModel;
-
   private int debounceTimeMs;
 
-  public ZkJobCoordinator(Config config) {
+  public ZkJobCoordinator(Config config, MetricsRegistry metricsRegistry) {
     this.config = config;
     ZkConfig zkConfig = new ZkConfig(config);
-    ZkKeyBuilder keyBuilder = new ZkKeyBuilder(new ApplicationConfig(config).getGlobalAppId());
+    ZkKeyBuilder keyBuilder = new ZkKeyBuilder(new ApplicationConfig(config).getAppId());
```
navina: Is the change from getGlobalAppId and getAppId intentional? If not, please change it back as it will cause a severe bug for zk based deployments (as seen during 0.13.0 release testing and fixed in SAMZA-1326: StreamProcessor should use ApplicationConfig.getGlobalAppId()).

It is not very obvious what a "globalAppId" is. sorry about that. Please correct it.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -270,6 +303,7 @@ public void onBarrierStateChanged(final String version, ZkBarrierForVersionUpgra
           // In our consensus model, if the Barrier is timed-out, then it means that one or more initial
           // participants failed to join. That means, they should have de-registered from "processors" list
           // and that would have triggered onProcessorChange action -> a new round of consensus.
+          metrics.singleBarrierRebalancingTime.update(System.nanoTime() - startTime);
```
navina: You can just increment this timer once before the "if" condition, instead of updating it in both if and else case :) Let me know if I am missing something obvious here.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorMetrics.java
```diff
@@ -0,0 +1,83 @@
...
+public class ZkJobCoordinatorMetrics extends MetricsBase {
+
+  private final MetricsRegistry metricsRegistry;
+
+  public final Counter reads;
+  public final Counter writes;
+  public final Counter subscriptions;
+  public final Counter zkConnectionError;
+
+  /**
```
navina: Thanks for adding the docs :+1:

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
```diff
@@ -67,10 +68,11 @@ public void testSetup() {
       // Do nothing
     }
 
+    ZkJobCoordinatorMetrics metrics = new ZkJobCoordinatorMetrics(new NoOpMetricsRegistry());
```
navina: nit: You can avoid a local variable `metrics` by directly using new NoOpMetricsRegistry in the ZkUtils constructor.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
PawasChhokra: Sorry about this. I think this changed on its own when I pulled from the updated repository and merged my code into it. Changed it back to what it was.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
PawasChhokra: Moved it out of the if-else block.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorMetrics.java
PawasChhokra: You're welcome!

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
PawasChhokra: Got rid of the 'metrics' local variable. Thank you :)

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -65,16 +65,25 @@
...
     this.keyBuilder = zkKeyBuilder;
     this.connectionTimeoutMs = connectionTimeoutMs;
     this.zkClient = zkClient;
   }
 
+  public ZkUtils(ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs, ZkJobCoordinatorMetrics metrics) {
+    this.keyBuilder = zkKeyBuilder;
+    this.connectionTimeoutMs = connectionTimeoutMs;
+    this.zkClient = zkClient;
+    this.metrics = metrics;
```
sborya: I've noticed you added a new constructor instead of changing the other one? What happens now with the code that uses the other constructor? 'metics' field will be null? Then this will cause multiple NPEs. Can you check it please.

navina: @sborya I already committed this JIRA a few days back. 
@santhoshvenkat1988 identified this issue and is in the process of fixing it. https://builds.apache.org/view/S-Z/view/Samza/job/samza-pr-checks/663/

PR #224, opened 2017-06-14
Updated test build and README versions to next version

prateekm:
Updated the versions used in smoke tests and the README for overnight tests to point to the next release. Using 0.13.1 instead of 0.13.1-SNAPSHOT here since these tests are usually run against the release branch when release testing, where the version doesn't have -SNAPSHOT in it.

### DIFF
--- a/docs/startup/hello-samza/versioned/index.md
+++ b/docs/startup/hello-samza/versioned/index.md
@@ -61,7 +61,7 @@ Then, you can continue w/ the following command in hello-samza project:
 {% highlight bash %}
 mvn clean package
 mkdir -p deploy/samza
-tar -xvf ./target/hello-samza-0.13.1-SNAPSHOT-dist.tar.gz -C deploy/samza
+tar -xvf ./target/hello-samza-0.13.1-dist.tar.gz -C deploy/samza
 {% endhighlight %}
 
 ### Run a Samza Job
--- a/docs/startup/hello-samza/versioned/index.md
+++ b/docs/startup/hello-samza/versioned/index.md
@@ -61,7 +61,7 @@ Then, you can continue w/ the following command in hello-samza project:
 {% highlight bash %}
 mvn clean package
 mkdir -p deploy/samza
-tar -xvf ./target/hello-samza-0.13.1-dist.tar.gz -C deploy/samza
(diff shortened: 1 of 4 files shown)


prateekm:
@nickpan47 @jmakes

docs/startup/hello-samza/versioned/index.md
```diff
@@ -61,7 +61,7 @@ Then, you can continue w/ the following command in hello-samza project:
 {% highlight bash %}
 mvn clean package
 mkdir -p deploy/samza
-tar -xvf ./target/hello-samza-0.13.1-SNAPSHOT-dist.tar.gz -C deploy/samza
+tar -xvf ./target/hello-samza-0.13.1-dist.tar.gz -C deploy/samza
```
jmakes: This should not be required until the next release. The release process removes the snapshot version with the assumption that you'd be testing against the released samza. The /latest/ copy of this tutorial assumes you're going to use master and build locally, which will create a snapshot version. 

Am I mistaken on this?

samza-test/src/main/python/configs/tests.json
```diff
@@ -1,5 +1,5 @@
 {
-  "samza_executable": "samza-test_2.11-0.13.0-SNAPSHOT.tgz",
+  "samza_executable": "samza-test_2.11-0.13.1.tgz",
```
jmakes: For the same reason above, I think this should be 13.1-SNAPSHOT

docs/startup/hello-samza/versioned/index.md
prateekm: Yeah this one should point to the version on master (0.13.1-SNAPSHOT). Will revert. See description for changing the versions in other files.

docs/startup/hello-samza/versioned/index.md
```diff
@@ -61,7 +61,7 @@ Then, you can continue w/ the following command in hello-samza project:
 {% highlight bash %}
 mvn clean package
 mkdir -p deploy/samza
-tar -xvf ./target/hello-samza-0.13.1-dist.tar.gz -C deploy/samza
+tar -xvf ./target/hello-samza-0.13.1-SNAPSHOT-dist.tar.gz -C deploy/samza
```
jmakes: Rebase needed? It looks like this change is already made in master

PR #225, opened 2017-06-14
SAMZA-1321: Aggregate and propagate end-of-stream and watermark control messages

xinyuiscool:
The patch completes the aggregation and propagation of both watermark and end-of-stream messages across intermediate streams. The main logic this patch adds:

1) EndOfStreamManager aggregates the end-of-stream control messages, propagate the result to to downstream intermediate topics based on the topology of the IO in the StreamGraph.

2) WatermarkManager aggregates the watermark control messages from the upstage tasks, pass it through the operators, and propagate it to downstream.

3) In operator impl, I implemented similar watermark logic as Beam for watermark propagation:
   *  InputWatermark(op) = min { OutputWatermark(op') | op1 is upstream of op}
   *  OutputWatermark(op) = min { InputWatermark(op), OldestWorkTime(op) }

Add quite a few unit tests and integration test. The code is 100% covered as reported by Intellij. Both control messages work as expected.


xinyuiscool:
@prateekm and @vjagadish1989 : please take a look. The number of files look pretty big, but most of them are actually test files. I believe the test code I wrote is much more than the source code :), due to the complexity of small EndOfStreamManager class.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -87,6 +92,7 @@ public final void init(Config config, TaskContext context) throws Exception {
 
     // create the operator impl DAG corresponding to the logical operator spec DAG
     this.operatorImplGraph = new OperatorImplGraph(streamGraph, config, context, clock);
+    this.eosDispatcher = EndOfStreamManager.createDispatcher(streamGraph);
```
nickpan47: So, we create the eosDispatcher multiple times, one per task. Is it due to the fact that we are initializing and creating the streamGraph once per task instance? We can live with it for now. But should revisit this when we have a serialized DAG.

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -597,7 +599,9 @@ private boolean checkEndOfStream() {
         if (envelope.isEndOfStream()) {
           SystemStreamPartition ssp = envelope.getSystemStreamPartition();
           processingSspSet.remove(ssp);
-          pendingEnvelopeQueue.remove();
+          if (!hasIntermediateStreams) {
```
nickpan47: So, if the task has intermediate streams, pendingEnvelopeQueue will not be empty. How do we trigger EoS in this case?

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -97,6 +102,13 @@ class TaskInstance(
     scala.collection.mutable.Map[SystemStreamPartition, Boolean]()
   systemStreamPartitions.foreach(ssp2catchedupMapping += _ -> false)
 
+  val controlManager = new ControlMessageManager(taskName.getTaskName,
+                                                       containerModel,
```
nickpan47: nit: indentation is not aligned.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -150,20 +162,28 @@ class TaskInstance(
...
 
-  def process(envelope: IncomingMessageEnvelope, coordinator: ReadableCoordinator,
+  def process(msgEnvelope: IncomingMessageEnvelope, coordinator: ReadableCoordinator,
     callbackFactory: TaskCallbackFactory = null) {
     metrics.processes.inc
 
-    if (!ssp2catchedupMapping.getOrElse(envelope.getSystemStreamPartition,
-      throw new SamzaException(envelope.getSystemStreamPartition + " is not registered!"))) {
-      checkCaughtUp(envelope)
+    if (!ssp2catchedupMapping.getOrElse(msgEnvelope.getSystemStreamPartition,
```
nickpan47: nit: let's rename it to correct English: ssp2caughtupMapping

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -150,20 +162,28 @@ class TaskInstance(
...
-    if (ssp2catchedupMapping(envelope.getSystemStreamPartition)) {
+    if (ssp2catchedupMapping(msgEnvelope.getSystemStreamPartition)) {
       metrics.messagesActuallyProcessed.inc
 
       trace("Processing incoming message envelope for taskName and SSP: %s, %s"
-        format (taskName, envelope.getSystemStreamPartition))
+        format (taskName, msgEnvelope.getSystemStreamPartition))
+
+      val envelope: IncomingMessageEnvelope = if (msgEnvelope.getMessage.isInstanceOf[ControlMessage]) {
+        controlManager.update(msgEnvelope)
```
nickpan47: This is not conforming to what we discussed earlier in the SEP. I thought that the interception of control messages and the reconciliation / propagation of control message would be done in TaskInstance. It seems that it is now separated in two pieces: update the EoS status of each stream in TaskInstance, and propagate EoS in StreamOperatorTask. Is that the current implementation? Can we elaborate why not do everything in TaskInstance? And even we decided that propagate EoS is not a general function for all task classes in TaskInstance, I would prefer to add a new interface ControlMessageListnerTask to handle onControlMessage(ControlMessage msg), instead of borrowing the IncomingMessageEnvelope which is associated to a single incoming SSP, which the consolidated control message apparently is meant for *all partitions* for the stream to a single task.

samza-core/src/test/java/org/apache/samza/operators/util/TestIOGraphUtil.java
```diff
@@ -0,0 +1,161 @@
...
+
+  @Before
+  public void setup() {
+    ApplicationRunner runner = mock(ApplicationRunner.class);
+    Map<String, String> configMap = new HashMap<>();
+    configMap.put(JobConfig.JOB_NAME(), "test-app");
+    configMap.put(JobConfig.JOB_DEFAULT_SYSTEM(), "test-system");
+    Config config = new MapConfig(configMap);
+
+    /**
```
nickpan47: Thanks for the detailed docs in the test! Appreciated it!

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
```diff
@@ -39,6 +39,7 @@
 import org.apache.samza.container.TaskInstanceExceptionHandler;
```
nickpan47: Can we add the test case with a task with intermediate streams? We should test that the EoS works in the following case:
--topic1--->Task1----output1-->
                   /|\      |
                    |       |
                    Lt3__|

Where t3 is the intermediate topic and EoS is finally triggered.

xinyuiscool:
Will upload the patch soon.

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
xinyuiscool: yes, I have an integration test for that :).

xinyuiscool:
@nickpan47 updated the patch based on our discussion. Now the EoS message won't go through the task, and we don't need seperate dispatchers anymore. The code looks much nicer :). Please take a look when you get a chance. The patch now also includes the watermarks, which is fully functioning.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
```diff
@@ -61,6 +61,7 @@ public TestableStreamProcessor(
     @Override
     SamzaContainer createSamzaContainer(
```
nickpan47: If this is the existing StreamProcessor interface method that are used by existing customers, this would be a breaking change and we should consider to start the 0.14 branch for this PR.

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -90,17 +90,6 @@ public boolean isEndOfStream() {
...
   }
 
-  /**
-   * Builds an end-of-stream envelope for an SSP. This is used by a {@link SystemConsumer} implementation to
-   * indicate that it is at end-of-stream. The end-of-stream message should not delivered to the task implementation.
-   *
-   * @param ssp The SSP that is at end-of-stream.
-   * @return an IncomingMessageEnvelope corresponding to end-of-stream for that SSP.
-   */
-  public static IncomingMessageEnvelope buildEndOfStreamEnvelope(SystemStreamPartition ssp) {
```
cpettitt-linkedin: This is not backwards compatible (I'm using it in Beam, BTW). Can we \@Deprecate with an indication of the preferred alternative?

samza-core/src/main/java/org/apache/samza/control/ControlMessageListener.java
```diff
@@ -0,0 +1,47 @@
...
+
+package org.apache.samza.control;
+
+import org.apache.samza.system.SystemStream;
+import org.apache.samza.task.MessageCollector;
+import org.apache.samza.task.TaskCoordinator;
+
+
+/**
+ * The listener interface for the aggregation result of control messages
```
cpettitt-linkedin: More documentation here would be helpful. Who is expected to use this interface?

samza-core/src/main/java/org/apache/samza/control/ControlMessageUtils.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.samza.system.StreamMetadataCache;
+import org.apache.samza.system.SystemStream;
+import org.apache.samza.system.SystemStreamMetadata;
+import org.apache.samza.task.MessageCollector;
+
+
+/**
+ * This class privates static utils for handling control messages
+ */
+public class ControlMessageUtils {
```
cpettitt-linkedin: Is this intended to be used by users? If not, can we make package private or give some indication via documentation?

samza-core/src/main/java/org/apache/samza/control/EndOfStreamManager.java
```diff
@@ -0,0 +1,164 @@
...
+import org.apache.samza.task.TaskCoordinator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * This class handles the end-of-stream control message. It aggregates the end-of-stream state for each input ssps of
+ * a task, and propagate the eos messages to downstream intermediate streams if needed.
+ */
+public class EndOfStreamManager {
```
cpettitt-linkedin: Can this be package private or marked as internal?

samza-core/src/main/java/org/apache/samza/control/EndOfStreamManager.java
```diff
@@ -0,0 +1,164 @@
...
+
+  public void init(IOGraph ioGraph) {
+    this.ioGraph = ioGraph;
+    this.upstreamTaskCounts = ControlMessageUtils.calculateUpstreamTaskCounts(inputToTasks, ioGraph);
+  }
+
+  public void update(IncomingMessageEnvelope envelope, TaskCoordinator coordinator) {
+    EndOfStreamState state = eosStates.get(envelope.getSystemStreamPartition());
+    EndOfStreamMessage message = (EndOfStreamMessage) envelope.getMessage();
+    state.update(message.getTaskName(), message.getTaskCount());
```
cpettitt-linkedin: Are we guaranteed that update is always called from a single thread? If not we need to ensure visibility.

Also, could we detect the upstream tasks when we build the graph? That would save sending task counts per message. You could also then just verify you've seen each of the upstream tasks via a concurrent set.

samza-core/src/main/java/org/apache/samza/control/IOGraph.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public boolean isOutputIntermediate() {
+      return isOutputIntermediate;
+    }
+  }
+
+  Collection<IONode> nodes;
+  Multimap<SystemStream, IONode> inputToNodes;
+
```
cpettitt-linkedin: final

samza-core/src/main/java/org/apache/samza/control/Watermark.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * A watermark is a monotonically increasing value, which represents the point up to which the
+ * system believes it has received all of the data before the watermark timestamp. Data that arrives
+ * with a timestamp that is before the watermark is considered late.
+ *
+ * <p>This is the aggregate result from the WatermarkManager, which keeps track of the control message
+ * {@link org.apache.samza.message.WatermarkMessage} and aggregate by returning the min of all watermark timestamp
+ * in each partition.
+ */
+public interface Watermark {
```
cpettitt-linkedin: I expected this just to be a value object but that doesn't agree with the `propagate` method, which looks out of place.

I would expect that the user can get and create a watermark, which may be as simple as a primitive (e.g. long). In the fluent API there should be a mechanism for the user to get watermark messages and a mechanism for them to emit watermark messages to all downstream consumers.

samza-core/src/main/java/org/apache/samza/control/Watermark.java
```diff
@@ -0,0 +1,55 @@
...
+   * @param systemStream intermediate stream
+   */
+  void propagate(SystemStream systemStream);
+
+  /**
+   * Create a copy of the watermark with the timestamp
+   * @param timestamp new timestamp
+   * @return new watermark
+   */
+  Watermark copyWithTimestamp(long timestamp);
```
cpettitt-linkedin: As a user I should be able to set an arbitrary watermark. This method suggests that I must first receive a watermark then copy it. If we have the former, this looks to be unnecessary.

samza-core/src/main/java/org/apache/samza/control/WatermarkManager.java
```diff
@@ -0,0 +1,191 @@
...
+  private final String taskName;
+  private final Map<SystemStreamPartition, WatermarkState> watermarkStates;
+  private final Map<SystemStream, Long> watermarkPerStream;
+  private final StreamMetadataCache metadataCache;
+  // mapping from input stream to its downstream tasks
+  private final Multimap<SystemStream, String> inputToTasks;
+  private final MessageCollector collector;
+
+  // mapping from output stream to its upstream task count
+  private Map<SystemStream, Integer> upstreamTaskCounts;
```
cpettitt-linkedin: As Yi mentioned below, it would be great to eliminate init and move to final to the greatest degree possible.

samza-core/src/main/java/org/apache/samza/control/WatermarkManager.java
```diff
@@ -0,0 +1,191 @@
...
+  long getWatermarkTime(SystemStreamPartition ssp) {
+    return watermarkStates.get(ssp).getWatermarkTime();
+  }
+
+  /**
+   * Send the watermark message to all partitions of an intermediate stream
+   * @param timestamp watermark timestamp
+   * @param systemStream intermediate stream
+   */
+  void sendWatermark(long timestamp, SystemStream systemStream, int taskCount) {
```
cpettitt-linkedin: How do users get and set watermarks? Maybe this comes later?

samza-core/src/main/java/org/apache/samza/control/WatermarkManager.java
```diff
@@ -0,0 +1,191 @@
...
+
+  /**
+   * Per ssp state of the watermarks. This class keeps track of the latest watermark timestamp
+   * from each upstream producer tasks, and use the min to update the aggregated watermark time.
+   */
+  final static class WatermarkState {
+    private int expectedTotal = Integer.MAX_VALUE;
+    private final Map<String, Long> timestamps = new HashMap<>();
+    private long watermarkTime = TIME_NOT_EXIST;
+
```
cpettitt-linkedin: If this is updateable from more than one thread then this state needs to be thread safe.

samza-core/src/main/java/org/apache/samza/control/WatermarkManager.java
```diff
@@ -0,0 +1,191 @@
...
+      }
+    }
+
+    long getWatermarkTime() {
+      return watermarkTime;
+    }
+  }
+
+  /**
+   * Implementation of the Watermark. It keeps a reference to the {@link WatermarkManager}
```
cpettitt-linkedin: We seem to be mixing two ideas here... the watermark itself with its mechanism for propagation. It would be cleaner to provide these at the fluent API level (this is how flink works) and to do the right default operation for typical operations. Ultimately the user needs to be able to override this behavior though and I'm not seeing that capability so far in the design.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -167,6 +173,71 @@ public final void onTimer(MessageCollector collector, TaskCoordinator coordinato
...
+    } else {
+      // InputWatermark(op) = min { OutputWatermark(op') | op1 is upstream of op}
+      inputWatermarkMin = prevOperators.stream().map(op -> op.getOutputWatermarkTime()).min(Long::compare).get();
+    }
+
+    if (inputWatermarkTime < inputWatermarkMin) {
+      // advance the watermark time of this operator
+      inputWatermarkTime = inputWatermarkMin;
+      Watermark inputWatermark = watermark.copyWithTimestamp(inputWatermarkTime);
+      long oldestWorkTime = handleWatermark(inputWatermark, collector, coordinator);
```
cpettitt-linkedin: handleWatermark is insufficient for Beam. I need the ability to emit watermark events even when I have not received a watermark. An existing example is that I may be able to process queued up events based on a side input arriving.

samza-core/src/main/java/org/apache/samza/control/ControlMessageListener.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public interface ControlMessageListener {
+
+  /**
+   * Returns the topology of the streams. Any control message listener needs to
+   * provide this topology so Samza can propagate the control message to downstreams.
+   * @return {@link IOGraph} of input to output streams
+   */
+  IOGraph getIOGraph();
```
cpettitt-linkedin: We should indicate in the docs that this may be null, as may be the case in AsyncStreamTaskAdapter.

samza-core/src/main/java/org/apache/samza/operators/impl/OutputOperatorImpl.java
```diff
@@ -69,4 +69,14 @@ protected void handleClose() {
   protected OperatorSpec<M, Void> getOperatorSpec() {
     return outputOpSpec;
   }
+
+  @Override
+  protected long handleWatermark(Watermark inputWatermark,
+      MessageCollector collector,
+      TaskCoordinator coordinator) {
+    if (outputOpSpec.getOpCode() == OperatorSpec.OpCode.PARTITION_BY) {
+      inputWatermark.propagate(outputStream.getStreamSpec().toSystemStream());
```
cpettitt-linkedin: Ideally when we propagate we should not need to be aware of who we are propagating to. This should be internally managed (and again, probably not on the watermark object).

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
xinyuiscool: Fixed with the deprecated annotation.

samza-core/src/main/java/org/apache/samza/control/ControlMessageListener.java
xinyuiscool: Added  more documentation here. Also renamed to ControlMessageListenerTask interface as suggested By Yi. Basically any task that listens to control message should implement this interface.

samza-core/src/main/java/org/apache/samza/control/ControlMessageListener.java
xinyuiscool: Makes sense. I added it in the java doc.

samza-core/src/main/java/org/apache/samza/control/ControlMessageUtils.java
xinyuiscool: Thanks for the catch! It's only for the WatermarkManager/EndOfStreamManager to use. I made it package private.

samza-core/src/main/java/org/apache/samza/control/EndOfStreamManager.java
xinyuiscool: It need to be public since it's created from TaskInstance. I marked it as internal in the java doc.

samza-core/src/main/java/org/apache/samza/control/EndOfStreamManager.java
xinyuiscool: Yes, it's done before the task can kick off any threads/async process. So it's already single threaded.

I created https://issues.apache.org/jira/browse/SAMZA-1352 for your second suggestion. I will submit another pr for that.

samza-core/src/main/java/org/apache/samza/control/IOGraph.java
xinyuiscool: Thanks for the catch! Fixed.

samza-core/src/main/java/org/apache/samza/control/Watermark.java
xinyuiscool: This makes a lot of sense! I was having some trouble of thinking whether this mixed logic makes it easier to use or having watermark and propagation separated. The use cases you mentioned here definitely need them separated so the user can create watermark easily. I creates https://issues.apache.org/jira/browse/SAMZA-1351 for this and will submit another pr for it.

samza-core/src/main/java/org/apache/samza/control/Watermark.java
xinyuiscool: Yes, makes sense to me. https://issues.apache.org/jira/browse/SAMZA-1351 will address it.

samza-core/src/main/java/org/apache/samza/control/WatermarkManager.java
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/control/WatermarkManager.java
xinyuiscool: Yes, I am focusing on the aggregation and propagation for now. Watermark emit from user and high-level API support will be future tasks.

samza-core/src/main/java/org/apache/samza/control/WatermarkManager.java
xinyuiscool: Right now it's single threaded as I commented in TaskInstance. I will make it thread safe once we start supporting running it in multithreading.

samza-core/src/main/java/org/apache/samza/control/WatermarkManager.java
xinyuiscool: Created https://issues.apache.org/jira/browse/SAMZA-1351 for separating watermark from propagation.

Support of Watermark Emission and High-leve API will come in the future (I will discuss with Yi and you for that).

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
xinyuiscool: Right, watermark emission is future task (current pr is for aggregation and propagation). I will discuss with you and Yi about the API to support it.

samza-core/src/main/java/org/apache/samza/operators/impl/OutputOperatorImpl.java
xinyuiscool: Yup, completely agree. https://issues.apache.org/jira/browse/SAMZA-1351 tracks this.

samza-core/src/main/java/org/apache/samza/control/WatermarkManager.java
cpettitt-linkedin: I would imagine that watermark handling would change rather significantly once we support user query / emission. I don't think it makes sense to treat them separately given the impact to the design (e.g. user facing pieces need to change and I would not be surprised if WatermarkManager gets inlined into the operators).

samza-core/src/main/java/org/apache/samza/operators/impl/OutputOperatorImpl.java
nickpan47: Agree with Chris, separating the propagation from the watermark object itself sounds good to me.

samza-core/src/main/java/org/apache/samza/control/WatermarkManager.java
nickpan47: @cpettitt-linkedin are you requesting to add user-inject watermark API together with this PR? And could you give some Beam example on which part of WatermarkManager function will be inlined into operators?
I was thinking:
1) user-inject watermark generation API can be added later, as additional callback methods provided to WatermarkManager
2) WatermarkManager is essentially only doing two things: reconcile watermarks/propagate watermarks. Do you have any use case from Beam that would need user-provided function to do reconcile/propagation of watermark differently than the general rule / algorithm implemented here?

Thanks!

cpettitt-linkedin: There should be some way for the user to control watermark emission (its fine, even good, if we do the right thing for the basic cases). The case where I need to control emission is when I'm pushing back work and then processing it based not on a watermark event but on a side input event (this would come through the normal event flow in Samza) - see DoFnOp from the Samza Beam runner as an example.

The WatermarkManager design is separating the handling of watermarks from the code that knows best how to handle watermarks (the operators). As a contrast, Flink has a cleaner design where watermark handling is baked into the operators with reasonable defaults and the user is able to entirely control watermark emission via an overridden operator. See AbstractStreamOperator where the default is to propagate watermarks, but the user can override processWatermark to change this. They can also emit watermarks from processWatermark or processeElement.

You can continue down the current design, but it is going to be more convoluted than Flink's and does not appear to have a compensating upside. If it's not obvious, please try to port DoFnOp to the proposed Samza API - it should make the design differences pretty clear. I suggest addressing user control of watermark emission now because I expect it will significantly change the user facing API and the implementation.

Happy to discuss further here or over video, but the port of DoFnOp should also be highly illuminating.

PR #228, opened 2017-06-17
SAMZA-1334: fix pre-condition for ContainerAllocator to work properly

nickpan47:
We have observed issues when the LocalityManager reports the container locality mapping while the host-affinity is disabled in ContainerAllocator, in which the ContainerAllocator failed to release extra containers.

Hence, fix is in the form of make sure the pre-condition is met for the ContainerAllocator w/o host-affinity: the localityMap from the JobModel should contain no preferred host info.


vjagadish1989 (approved):
looks good. approved. thanks for the contribution @nickpan47!

PR #229, opened 2017-06-21
SAMZA-1336. Session disconnect propagation.

sborya:
If ZK doesn't receive any communication from a zkClient (including heartbeats), it closes the session with the client. It removes all the ephemeral nodes associated with the client. That's why we need to restore all these nodes - need to re-register. 
We are using ZkClient library to connect to zookeeper. This library allows us to get notification when the session is closed and when a new session is created. So when the new session is created we reset all session related state and re-register.
One weird feature of the library/zookeeper is that when a new session is established, it is still possible to receive old notifications. To avoid this we introduced generation number which we pass into each callback. And if the generation number has changed when the callback was invoked, we will ignore this callback.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/config/TaskConfigJava.java
+++ b/samza-core/src/main/java/org/apache/samza/config/TaskConfigJava.java
@@ -37,7 +37,7 @@ import scala.collection.JavaConverters;
 public class TaskConfigJava extends MapConfig {
   // Task Configs
   private static final String TASK_SHUTDOWN_MS = "task.shutdown.ms";
-  public static final long DEFAULT_TASK_SHUTDOWN_MS = 5000L;
+  public static final long DEFAULT_TASK_SHUTDOWN_MS = 30000L;
 
   // broadcast streams consumed by all tasks. e.g. kafka.foo#1
   public static final String BROADCAST_INPUT_STREAMS = "task.broadcast.inputs";
--- a/samza-core/src/main/java/org/apache/samza/config/TaskConfigJava.java
+++ b/samza-core/src/main/java/org/apache/samza/config/TaskConfigJava.java
@@ -37,7 +37,7 @@ import scala.collection.JavaConverters;
 public class TaskConfigJava extends MapConfig {
   // Task Configs
   private static final String TASK_SHUTDOWN_MS = "task.shutdown.ms";
-  private static final long DEFAULT_TASK_SHUTDOWN_MS = 30000L;
+  public static final long DEFAULT_TASK_SHUTDOWN_MS = 30000L;
 
   // broadcast streams consumed by all tasks. e.g. kafka.foo#1
   public static final String BROADCAST_INPUT_STREAMS = "task.broadcast.inputs";
--- a/samza-core/src/main/java/org/apache/samza/config/TaskConfigJava.java
+++ b/samza-core/src/main/java/org/apache/samza/config/TaskConfigJava.java
@@ -37,7 +37,7 @@ import scala.collection.JavaConverters;
 public class TaskConfigJava extends MapConfig {
   // Task Configs
   private static final String TASK_SHUTDOWN_MS = "task.shutdown.ms";
-  public static final long DEFAULT_TASK_SHUTDOWN_MS = 30000L;
+  private static final long DEFAULT_TASK_SHUTDOWN_MS = 30000L;
 
   // broadcast streams consumed by all tasks. e.g. kafka.foo#1
   public static final String BROADCAST_INPUT_STREAMS = "task.broadcast.inputs";
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -71,7 +71,7 @@ public class StreamProcessor {
   
   // Latch used to synchronize between the JobCoordinator thread and the container thread, when the container is
   // stopped due to re-balancing
-  private volatile CountDownLatch jcContainerShutdownLatch = new CountDownLatch(1);
+  /* package private */volatile CountDownLatch jcContainerShutdownLatch;
   private volatile boolean processorOnStartCalled = false;
 
   @VisibleForTesting
@@ -110,6 +110,10 @@ public class StreamProcessor {
     this(config, customMetricsReporters, (Object) streamTaskFactory, processorListener, null);
   }
 
+  public String toString() {
+    return "Processor" + getCurrentJobCoordinator().getProcessorId();
+  }
+
   /* package private */
   JobCoordinator getJobCoordinator() {
     return Util.
@@ -117,6 +121,11 @@ public class StreamProcessor {
             new JobCoordinatorConfig(config)
                 .getJobCoordinatorFactoryClassName())
         .getJobCoordinator(config);
+  }
+
+  @VisibleForTesting
+  public JobCoordinator getCurrentJobCoordinator() {
+    return jobCoordinator;
   }
 
   StreamProcessor(Config config, Map<String, MetricsReporter> customMetricsReporters, Object taskFactory,
@@ -129,7 +138,10 @@ public class StreamProcessor {
     this.jobCoordinator = (jobCoordinator != null) ? jobCoordinator : getJobCoordinator();
     this.jobCoordinatorListener = createJobCoordinatorListener();
     this.jobCoordinator.setListener(jobCoordinatorListener);
+    //this.jcContainerShutdownLatch= new CountDownLatch(1);
+    //LOGGER.info("latch for pid " + jobCoordinator.getProcessorId() + " = " + jcContainerShutdownLatch);
   }
+
 
   /**
    * Asynchronously starts this {@link StreamProcessor}.
@@ -197,6 +209,7 @@ public class StreamProcessor {
   }
 
   JobCoordinatorListener createJobCoordinatorListener() {
+    final String pid = this.toString();
     return new JobCoordinatorListener() {
 
       @Override
@@ -206,9 +219,10 @@ public class StreamProcessor {
           if (SamzaContainerStatus.NOT_STARTED.equals(status) || SamzaContainerStatus.STARTED.equals(status)) {
             boolean shutdownComplete = false;
             try {
-              LOGGER.info("Shutting down container in onJobModelExpired.");
+              LOGGER.info("Shutting down container in onJobModelExpired for processor:" + pid);
               container.pause();
               shutdownComplete = jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
+              LOGGER.info("Latch " + jcContainerShutdownLatch + " returned. ShutdownCompete=" + shutdownComplete + ";pid=" + pid + ";cd=" + jcContainerShutdownLatch.getCount());
             } catch (IllegalContainerStateException icse) {
               // Ignored since container is not running
               LOGGER.info("Container was not running.", icse);
@@ -216,6 +230,7 @@ public class StreamProcessor {
             } catch (InterruptedException e) {
               LOGGER.warn("Container shutdown was interrupted!" + container.toString(), e);
             }
+            LOGGER.info("Shutting down container done for pid=" + pid + "; complete =" + shutdownComplete);
             if (!shutdownComplete) {
               LOGGER.warn("Container " + container.toString() + " may not have shutdown successfully. " +
                   "Stopping the processor.");
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -67,6 +67,8 @@ public class StreamProcessor {
   private ExecutorService executorService;
 
   private volatile SamzaContainer container = null;
+  private volatile Throwable containerException = null;
+  
   // Latch used to synchronize between the JobCoordinator thread and the container thread, when the container is
   // stopped due to re-balancing
   private volatile CountDownLatch jcContainerShutdownLatch = new CountDownLatch(1);
@@ -274,7 +276,8 @@ public class StreamProcessor {
               } else {
                 LOGGER.warn("JobCoordinatorLatch was null. It is possible for some component to be waiting.");
               }
-              LOGGER.error("Container failed. Stopping the processor.", t);
+              containerException = t;
+              LOGGER.error("Container failed. Stopping the processor.", containerException);
               container = null;
               stop();
             }
@@ -298,7 +301,10 @@ public class StreamProcessor {
           executorService.shutdownNow();
         }
         if (processorListener != null) {
-          processorListener.onShutdown();
+          if (containerException != null)
+            processorListener.onFailure(containerException);
+          else
+            processorListener.onShutdown();
         }
       }
 
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -63,6 +63,7 @@ public class StreamProcessor {
   private final Map<String, MetricsReporter> customMetricsReporter;
   private final Config config;
   private final long taskShutdownMs;
+  private final String processorId;
 
   private ExecutorService executorService;
 
@@ -112,7 +113,7 @@ public class StreamProcessor {
 
   @Override
   public String toString() {
-    return "Processor" + getCurrentJobCoordinator().getProcessorId();
+    return "Processor:" + processorId;
   }
 
   /* package private */
@@ -125,7 +126,7 @@ public class StreamProcessor {
   }
 
   @VisibleForTesting
-  public JobCoordinator getCurrentJobCoordinator() {
+  JobCoordinator getCurrentJobCoordinator() {
     return jobCoordinator;
   }
 
@@ -139,6 +140,8 @@ public class StreamProcessor {
     this.jobCoordinator = (jobCoordinator != null) ? jobCoordinator : getJobCoordinator();
     this.jobCoordinatorListener = createJobCoordinatorListener();
     this.jobCoordinator.setListener(jobCoordinatorListener);
+
+    processorId = jobCoordinator.getProcessorId();
   }
 
   /**
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -110,6 +110,7 @@ public class StreamProcessor {
     this(config, customMetricsReporters, (Object) streamTaskFactory, processorListener, null);
   }
 
+  @Override
   public String toString() {
     return "Processor" + getCurrentJobCoordinator().getProcessorId();
   }
@@ -219,7 +220,7 @@ public class StreamProcessor {
               LOGGER.info("Shutting down container in onJobModelExpired for processor:" + pid);
               container.pause();
               shutdownComplete = jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
-              LOGGER.info("ShutdownCompete=" + shutdownComplete + ";pid=" + pid);
+              LOGGER.info("ShutdownComplete=" + shutdownComplete);
             } catch (IllegalContainerStateException icse) {
               // Ignored since container is not running
               LOGGER.info("Container was not running.", icse);
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -138,10 +138,7 @@ public class StreamProcessor {
     this.jobCoordinator = (jobCoordinator != null) ? jobCoordinator : getJobCoordinator();
     this.jobCoordinatorListener = createJobCoordinatorListener();
     this.jobCoordinator.setListener(jobCoordinatorListener);
-    //this.jcContainerShutdownLatch= new CountDownLatch(1);
-    //LOGGER.info("latch for pid " + jobCoordinator.getProcessorId() + " = " + jcContainerShutdownLatch);
   }
-
 
   /**
    * Asynchronously starts this {@link StreamProcessor}.
@@ -222,7 +219,7 @@ public class StreamProcessor {
               LOGGER.info("Shutting down container in onJobModelExpired for processor:" + pid);
               container.pause();
               shutdownComplete = jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
-              LOGGER.info("Latch " + jcContainerShutdownLatch + " returned. ShutdownCompete=" + shutdownComplete + ";pid=" + pid + ";cd=" + jcContainerShutdownLatch.getCount());
+              LOGGER.info("ShutdownCompete=" + shutdownComplete + ";pid=" + pid);
             } catch (IllegalContainerStateException icse) {
               // Ignored since container is not running
               LOGGER.info("Container was not running.", icse);
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -141,7 +141,7 @@ public class StreamProcessor {
     this.jobCoordinatorListener = createJobCoordinatorListener();
     this.jobCoordinator.setListener(jobCoordinatorListener);
 
-    processorId = jobCoordinator.getProcessorId();
+    processorId = this.jobCoordinator.getProcessorId();
   }
 
   /**
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
@@ -93,7 +93,7 @@ public class ZkBarrierForVersionUpgrade {
 
     // subscribe for participant's list changes
     LOG.info("Subscribing for child changes at " + barrierParticipantsPath);
-    zkUtils.getZkClient().subscribeChildChanges(barrierParticipantsPath, new ZkBarrierChangeHandler(version, participants));
+    zkUtils.subscribeChildChanges(barrierParticipantsPath, new ZkBarrierChangeHandler(version, participants));
 
     barrierListenerOptional.ifPresent(zkBarrierListener -> zkBarrierListener.onBarrierCreated(version));
   }
@@ -106,7 +106,7 @@ public class ZkBarrierForVersionUpgrade {
    */
   public void join(String version, String participantId) {
     String barrierDonePath = keyBuilder.getBarrierStatePath(version);
-    zkUtils.getZkClient().subscribeDataChanges(barrierDonePath, new ZkBarrierReachedHandler(barrierDonePath, version));
+    zkUtils.subscribeDataChanges(barrierDonePath, new ZkBarrierReachedHandler(barrierDonePath, version));
 
     // TODO: Handle ZkNodeExistsException - SAMZA-1304
     zkUtils.getZkClient().createPersistent(
@@ -119,7 +119,7 @@ public class ZkBarrierForVersionUpgrade {
    * @param version Version associated with the Barrier
    */
   public void expire(String version) {
-    zkUtils.getZkClient().writeData(
+    zkUtils.writeData(
         keyBuilder.getBarrierStatePath(version),
         State.TIMED_OUT);
 
@@ -150,8 +150,8 @@ public class ZkBarrierForVersionUpgrade {
       if (currentChildren.size() == names.size() && CollectionUtils.containsAll(currentChildren, names)) {
         String barrierDonePath = keyBuilder.getBarrierStatePath(barrierVersion);
         LOG.info("Writing BARRIER DONE to " + barrierDonePath);
-        zkUtils.getZkClient().writeData(barrierDonePath, State.DONE); // this will trigger notifications
-        zkUtils.getZkClient().unsubscribeChildChanges(barrierDonePath, this);
+        zkUtils.writeData(barrierDonePath, State.DONE); // this will trigger notifications
+        zkUtils.unsubscribeChildChanges(barrierDonePath, this);
       }
     }
   }
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
@@ -131,14 +131,14 @@ public class ZkBarrierForVersionUpgrade {
     private final List<String> names;
 
     public ZkBarrierChangeHandler(String barrierVersion, List<String> names, ZkUtils zkUtils) {
-      super(zkUtils);
+      super(zkUtils, "ZkBarrierChangeHandler");
       this.barrierVersion = barrierVersion;
       this.names = names;
     }
 
     @Override
     public void handleChildChange(String parentPath, List<String> currentChildren) {
-      if (skip("ZkBarrierChangeHandler")) {
+      if (notAValidEvent()) {
         return;
       }
       if (currentChildren == null) {
@@ -177,7 +177,7 @@ public class ZkBarrierForVersionUpgrade {
     @Override
     public void handleDataChange(String dataPath, Object data) {
       LOG.info("got notification about barrier " + barrierStatePath + "; done=" + data);
-      if (notAValidEven())
+      if (notAValidEvent())
         return;
 
       zkUtils.unsubscribeDataChanges(barrierStatePath, this);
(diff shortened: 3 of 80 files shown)


sborya:
@prateekm @navina, could you please review?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -65,11 +65,21 @@
...
+  private int currentGeneration;
 
   public ZkUtils(ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs) {
     this.keyBuilder = zkKeyBuilder;
     this.connectionTimeoutMs = connectionTimeoutMs;
     this.zkClient = zkClient;
+    currentGeneration = 0;
+  }
+
+  public synchronized void incGeneration() {
```
cpettitt-linkedin: Over time I've generally found it a best practice not to synchronize on your object directly because other users may end up doing the same which can lead to surprising locking issues. You can use a private Object as a lock.

However, given that there is no real coordination of state here, you could opt to make currentGeneration an AtomicInt and use CAS to update it. Reads should be cheaper too.

samza-test/src/test/java/org/apache/samza/processor/TestZkStreamProcessorBase.java
```diff
@@ -93,6 +98,41 @@ public void setUp() {
     createTopics(inputTopic, outputTopic);
   }
 
+  protected Thread expireSession1(ZkClient zkClient) {
```
cpettitt-linkedin: This looks like dead code that can be removed. Suspend might be problematic if the code is suspended while invoking a callback (see Thread#suspend docs for details).

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessorSession.java
```diff
@@ -0,0 +1,133 @@
...
+
+    // run the processors in separate threads
+    Thread[] threads = new Thread[processorIds.length];
+    for (int i = 0; i < processorIds.length; i++) {
+      threads[i] = runInThread(streamProcessors[i], TestZkStreamProcessorBase.TestStreamTask.endLatch);
+      threads[i].start();
+      // wait until the processor reports that it has started
+      try {
+        synchronized (startWait[i]) {
+          startWait[i].wait(1000);
```
cpettitt-linkedin: This could wake up early due to a spurious wakeup. See Object#wait for details. You could address this either by adding a condition check or using a CountDownLatch, with the latter being simpler.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessorSession.java
```diff
@@ -0,0 +1,133 @@
...
+
+    // expire zk session of one of the processors
+    int pidToStop = (processorIds.length > 1) ? 1 : 0;
+    expireSession(jobCoordinators[pidToStop].getZkUtils().getZkClient());
+
+    // wait until all other processors stop
+    for (int i = 0; i < processorIds.length; i++) {
+      // wait until all other processor reports that they have stopped
+      try {
+        synchronized (stopWait[i]) {
```
cpettitt-linkedin: Same as above re. wait.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -280,4 +276,50 @@ public void onBarrierError(String version, Throwable t) {
...
+      // increase generation of the ZK connection. All the callbacks from the previous generation will be ignored.
+      zkUtils.incGeneration();
+
+      if (coordinatorListener != null) {
+        coordinatorListener.onJobModelExpired();
+      }
+      // reset all the values that might've been from the previous session (e.g ephemeral node path)
+      zkUtils.unregister();
+
+      if (zkController != null) {
```
cpettitt-linkedin: This check seems unnecessary because zkController is final and initialized in the constructor.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -280,4 +276,50 @@ public void onBarrierError(String version, Throwable t) {
...
+    public void handleSessionEstablishmentError(Throwable error)
+        throws Exception {
+      // this means we cannot connect to zookeeper
+      LOG.info("=================================================handleSessionEstablishmentError received for processor=" + processorId, error);
+      stop();
+    }
+  }
+
+  @VisibleForTesting
+  public ZkUtils getZkUtils() {
```
cpettitt-linkedin: Can this be made package private?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -65,11 +65,21 @@
   private volatile String ephemeralPath = null;
   private final ZkKeyBuilder keyBuilder;
   private final int connectionTimeoutMs;
+  private int currentGeneration;
 
   public ZkUtils(ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs) {
     this.keyBuilder = zkKeyBuilder;
     this.connectionTimeoutMs = connectionTimeoutMs;
     this.zkClient = zkClient;
+    currentGeneration = 0;
```
cpettitt-linkedin: It might be worth a note in the class docs about how currentGeneration is used in ZkUtil.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -57,7 +57,7 @@
 public class StreamProcessor {
   private static final Logger LOGGER = LoggerFactory.getLogger(StreamProcessor.class);
 
-  private final JobCoordinator jobCoordinator;
+  private JobCoordinator jobCoordinator;
```
sborya: done.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -117,6 +117,11 @@ JobCoordinator getJobCoordinator() {
         .getJobCoordinator(config);
   }
 
+  @VisibleForTesting
+  public JobCoordinator getCurrentJobCoordinator() {
```
sborya: Need to remove some tests to the same package. Done.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: Right, done.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: done. AtomicInteger.

samza-test/src/test/java/org/apache/samza/processor/TestZkStreamProcessorBase.java
sborya: Yeah, this was on of the attempts to simulate ZK timeout.. Didn't work anyway.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -163,7 +183,12 @@ String readProcessorData(String fullPath) {
         String fullPath = String.format("%s/%s", processorPath, child);
         processorIds.add(new ProcessorData(readProcessorData(fullPath)).getProcessorId());
       }
-
+      Collections.sort(processorIds, new Comparator<String>() {
```
shanthoosh: Just curious, why're we sorting the processorIds in `getActiveProcessorsIDs` api.

We already have `getSortedActiveProcessorsIDs` which does the same(or does it make sense to merge them both).

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -65,11 +74,21 @@
...
+  private AtomicInteger currentGeneration;
 
   public ZkUtils(ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs) {
     this.keyBuilder = zkKeyBuilder;
     this.connectionTimeoutMs = connectionTimeoutMs;
     this.zkClient = zkClient;
+    currentGeneration = new AtomicInteger(0);
+  }
+
+  public synchronized void incGeneration() {
```
shanthoosh: Just wondering if having synchronized block around AtomicInteger is redundant.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -280,4 +276,50 @@ public void onBarrierError(String version, Throwable t) {
...
+        zkController.register();
+      }
+    }
+
+    @Override
+    public void handleSessionEstablishmentError(Throwable error)
+        throws Exception {
+      // this means we cannot connect to zookeeper
+      LOG.info("=================================================handleSessionEstablishmentError received for processor=" + processorId, error);
+      stop();
```
shanthoosh: Just curious.. We're stopping debounceTimer(other resources) along with zkController. 

On new session establishment, only zkController is registered. DebounceTimer(executorService), others has to be reinstantiated right?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: no, not really, it is called from a test which is in a different package..

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: It is. Removed (it was a leftover from before it was atomic :))

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: No, we don't stop debounce on a new Session establishment. We stop() everything if we get sessionEstablishmentError - which means client couldn't re-establish a session and it is a hard failure.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: I will look at it later.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
shanthoosh: Thanks for the explanation. Makes sense.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessorSession.java
sborya: yep. Done. CountDownLatch.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkStreamProcessorSession.java
sborya: done. CountDownLatch.

samza-core/src/main/java/org/apache/samza/config/TaskConfigJava.java
```diff
@@ -37,7 +37,7 @@
 public class TaskConfigJava extends MapConfig {
   // Task Configs
   private static final String TASK_SHUTDOWN_MS = "task.shutdown.ms";
-  public static final long DEFAULT_TASK_SHUTDOWN_MS = 5000L;
+  public static final long DEFAULT_TASK_SHUTDOWN_MS = 30000L;
```
sborya: Going forward, we need to allow much more time for a container to shutdown. If the container shuts down before that - there is no impact. But for huge jobs 5 sec may be not enough.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -93,7 +91,7 @@ public void create(final String version, List<String> participants) {
 
     // subscribe for participant's list changes
     LOG.info("Subscribing for child changes at " + barrierParticipantsPath);
-    zkUtils.getZkClient().subscribeChildChanges(barrierParticipantsPath, new ZkBarrierChangeHandler(version, participants));
+    zkUtils.getZkClient().subscribeChildChanges(barrierParticipantsPath, new ZkBarrierChangeHandler(version, participants, zkUtils));
```
sborya: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -79,11 +98,12 @@ public void connect() throws ZkInterruptedException {
...
 
-  public static ZkConnection createZkConnection(String zkConnectString, int sessionTimeoutMs) {
-    return new ZkConnection(zkConnectString, sessionTimeoutMs);
+  // reset all zk-session specific state
+  public void unregister() {
+    ephemeralPath = null;
   }
 
-  ZkClient getZkClient() {
+  public ZkClient getZkClient() {
```
sborya: although it is a right thing to do, unfortunately we have some tests in a different package using it.
And I don't know an immediate way to change it.

sborya:
@shanthoosh @navina  please review.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -215,14 +242,60 @@ public void close() throws ZkInterruptedException {
...
+    public GenIZkChildListener(ZkUtils zkUtils) {
+      generation = zkUtils.getGeneration();
+      this.zkUtils = zkUtils;
+    }
+
+    protected boolean skip(String listenerName) {
+      int curGeneration = zkUtils.getGeneration();
+      if (curGeneration != generation) {
+        LOG.warn("SKIPPING handleDataChanged for " + listenerName +
+            " from wrong generation. curGen=" + curGeneration + "; cb gen= " + generation);
```
shanthoosh: What is cb gen(is more like jvm permgen|newgen)?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -315,4 +315,48 @@ public void onBarrierError(String version, Throwable t) {
       stop();
     }
   }
+
+  /// listener to handle session expiration
+  public class ZkSessionStateChangedListener implements IZkStateListener {
```
shanthoosh: We can make this package private(No dependencies outside). 

Curious question: You could have written this in same package/anywhere. Is the main motivation for having it as inner class is to perform these actions (`coordinator.stop()`, `zkController.register|unregister`, zkutils operations) easily(not worry about constructing dependency tree for this)?

samza-core/src/main/java/org/apache/samza/config/TaskConfigJava.java
sborya: Yes, we should make it private. Will do.
Yes, in the future we need to make sure that if the new Job model doesn't change for this specific processor, it shouldn't stop at all. And that should be majority of the case.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: right. Fixed.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -129,8 +138,11 @@ JobCoordinator getJobCoordinator() {
     this.jobCoordinator = (jobCoordinator != null) ? jobCoordinator : getJobCoordinator();
     this.jobCoordinatorListener = createJobCoordinatorListener();
     this.jobCoordinator.setListener(jobCoordinatorListener);
+    //this.jcContainerShutdownLatch= new CountDownLatch(1);
```
sborya: my logic was - we check the latch only when container is not null, and container becomes not null only in onNewJobModel, is it correct?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -129,8 +138,11 @@ JobCoordinator getJobCoordinator() {
...
+    //LOGGER.info("latch for pid " + jobCoordinator.getProcessorId() + " = " + jcContainerShutdownLatch);
   }
 
+
```
sborya: ok.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -206,16 +219,18 @@ public void onJobModelExpired() {
           if (SamzaContainerStatus.NOT_STARTED.equals(status) || SamzaContainerStatus.STARTED.equals(status)) {
             boolean shutdownComplete = false;
             try {
-              LOGGER.info("Shutting down container in onJobModelExpired.");
+              LOGGER.info("Shutting down container in onJobModelExpired for processor:" + pid);
               container.pause();
```
sborya: Not sure if we need it as a metric (too many of them anyway). But may be a good addition for debugging. Feel free to add a jira (and implement it if you want).

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -206,16 +219,18 @@ public void onJobModelExpired() {
...
               shutdownComplete = jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
+              LOGGER.info("Latch " + jcContainerShutdownLatch + " returned. ShutdownCompete=" + shutdownComplete + ";pid=" + pid + ";cd=" + jcContainerShutdownLatch.getCount());
```
sborya: I will delete it. I was using it for debugging and needed hash value to make sure it was the same object.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: fixed.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -215,14 +242,60 @@ public void close() throws ZkInterruptedException {
...
   }
 
+  // Generation enforcing zk listener abstract class.
+  // Helps listeners, which extend it,  to skip old generation events.
+  // We cannot use 'sessionId' for this because it is not available through ZkClient (at leaste without reflection)
+  public abstract static class GenIZkChildListener implements IZkChildListener {
+    private final int generation;
+    private final ZkUtils zkUtils;
+
+    public GenIZkChildListener(ZkUtils zkUtils) {
```
sborya: I can do that.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -215,14 +242,60 @@ public void close() throws ZkInterruptedException {
...
   /**
     * subscribe for changes of JobModel version
     * @param dataListener describe this
     */
-  public void subscribeToJobModelVersionChange(IZkDataListener dataListener) {
+  public void subscribeToJobModelVersionChange(GenIZkDataListener dataListener) {
     LOG.info(" subscribing for jm version change at:" + keyBuilder.getJobModelVersionPath());
     zkClient.subscribeDataChanges(keyBuilder.getJobModelVersionPath(), dataListener);
-    metrics.subscriptions.inc();
+    if (metrics != null) {
```
sborya: So what will happen if it is null? How is it fixed?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: yes. Functionally this class makes sense only for the JobCoordinator, so I didn't see a reason to take it out.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
shanthoosh: It's initialized with proper params (will be non null). 

Even if we want to guard against null, ll be great to verify once in constructor before assignment(just to eliminate repeated checks).

sborya:
@navina @shanthoosh @cpettitt-linkedin 
I've addressed all of the comments - please review/shipit.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -48,7 +48,7 @@ object JobConfig {
   val JOB_CONTAINER_SINGLE_THREAD_MODE = "job.container.single.thread.mode"
   val JOB_INTERMEDIATE_STREAM_PARTITIONS = "job.intermediate.stream.partitions"
   val JOB_DEBOUNCE_TIME_MS = "job.debounce.time.ms"
-  val DEFAULT_DEBOUNCE_TIME_MS = 2000
+  val DEFAULT_DEBOUNCE_TIME_MS = 20000
```
shanthoosh: Any reason to increase default debounce time from `2s` to `20s` ?

samza-core/src/main/java/org/apache/samza/config/TaskConfigJava.java
navina: > This change in worst case will cause long rebalancing phases 

This is definitely how this change will affect the Zk based deployment mode. However, it will affect the existing yarn based deployments. The side-effect may not be significant. But for high throughput jobs, a container downtime for 30 s means no processing for 30s that could cause significant lag. 
I am not in favor of changing existing default config values (http://samza.apache.org/learn/documentation/latest/jobs/configuration-table.html) as part of a larger patch. This kind of change should be treated equivalent to an api change and needs to be evaluated/planned separately. Can we please do that in a separate PR ?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
navina: @sborya I don't see any usages outside of the current package namespace. Am I missing something? Which tests are you referring to?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -315,4 +315,48 @@ public void onBarrierError(String version, Throwable t) {
...
+    @Override
+    public void handleNewSession()
+        throws Exception {
+      LOG.info("Got new session created event for processor=" + processorId);
+
+      // increase generation of the ZK connection. All the callbacks from the previous generation will be ignored.
+      zkUtils.incGeneration();
+
+      if (coordinatorListener != null) {
+        coordinatorListener.onJobModelExpired();
```
navina: onJobModelExpired is a blocking call. Should that be handled directly in the event thread? Esp since shutting down the container is taking longer for certain jobs?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -190,12 +204,16 @@ public void unsubscribeDataChanges(String path, IZkDataListener dataListener) {
 
   public void subscribeDataChanges(String path, IZkDataListener dataListener) {
     zkClient.subscribeDataChanges(path, dataListener);
-    metrics.subscriptions.inc();
+    if (metrics != null) {
```
navina: thanks for adding this check! :+1:

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -215,14 +235,62 @@ public void close() throws ZkInterruptedException {
...
+    private final ZkUtils zkUtils;
+    private final String listenerName;
+
+    public GenIZkDataListener(ZkUtils zkUtils, String listenerName) {
+      generation = zkUtils.getGeneration();
+      this.zkUtils = zkUtils;
+      this.listenerName = listenerName;
+    }
+
+    protected boolean notAValidEven() {
```
navina: typo in method name: s/notAValidEven/notAValidEvent/

in the previous listener, you have used the method name as skip(). Does it make sense to use the same for both?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -215,14 +235,62 @@ public void close() throws ZkInterruptedException {
...
+      int curGeneration = zkUtils.getGeneration();
+      if (curGeneration != generation) {
+        LOG.warn("SKIPPING handleDataChanged for " + listenerName +
```
navina: Perhaps, we can log this on the caller side so that you don't have to pass the listener name??

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -171,12 +186,11 @@ String readProcessorData(String fullPath) {
     String processorPath = keyBuilder.getProcessorsPath();
     List<String> processorIds = new ArrayList<>(znodeIds.size());
     if (znodeIds.size() > 0) {
-
       for (String child : znodeIds) {
         String fullPath = String.format("%s/%s", processorPath, child);
         processorIds.add(new ProcessorData(readProcessorData(fullPath)).getProcessorId());
       }
-
+      Collections.sort(processorIds);
```
navina: Any reason why we are sorting this processorIds here?  Looks like previously we weren't and now we do.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -215,14 +235,62 @@ public void close() throws ZkInterruptedException {
     zkClient.close();
   }
 
+  // Generation enforcing zk listener abstract class.
```
navina: Can you please make this into a multi-line comment ? :)

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -162,18 +164,22 @@ public void handleChildChange(String parentPath, List<String> currentChildren) {
...
+    public ZkBarrierReachedHandler(String barrierStatePath, String version, ZkUtils zkUtils) {
+      super(zkUtils, "ZkBarrierReachedHandler");
       this.barrierStatePath = barrierStatePath;
       this.barrierVersion = version;
     }
 
     @Override
     public void handleDataChange(String dataPath, Object data) {
       LOG.info("got notification about barrier " + barrierStatePath + "; done=" + data);
+      if (notAValidEven())
```
sborya: typo fix done. thanks.
I want to keep it at the info level for at least first release, in case we need to debug it.
For the next version we will review log levels to minimize logging.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -73,34 +78,33 @@ public ZkJobCoordinator(Config config, MetricsRegistry metricsRegistry) {
     this.config = config;
     ZkConfig zkConfig = new ZkConfig(config);
     ZkKeyBuilder keyBuilder = new ZkKeyBuilder(new ApplicationConfig(config).getGlobalAppId());
+    ZkClient zkClient = ZkCoordinationServiceFactory
+        .createZkClient(zkConfig.getZkConnect(), zkConfig.getZkSessionTimeoutMs(), zkConfig.getZkConnectionTimeoutMs());
```
Also changed GenIZkChildListener from ZkUtils (from the previous comment, but github doesn't let me comment)

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -65,34 +73,39 @@
...
     this.connectionTimeoutMs = connectionTimeoutMs;
     this.zkClient = zkClient;
     this.metrics = metrics;
+    currentGeneration = new AtomicInteger(0);
   }
 
   public void connect() throws ZkInterruptedException {
     boolean isConnected = zkClient.waitUntilConnected(connectionTimeoutMs, TimeUnit.MILLISECONDS);
-    if (!isConnected) {
+    if (!isConnected && metrics != null) {
```

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
sborya: we need it to be bigger then Session timeout (which is 6 sec).
Better way would've been to have it 3xSessionTO, but we don't have the setting here.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -110,6 +110,10 @@ public StreamProcessor(Config config, Map<String, MetricsReporter> customMetrics
     this(config, customMetricsReporters, (Object) streamTaskFactory, processorListener, null);
   }
 
+  public String toString() {
```
sborya: It is useful for debugging. I am not sure where else can it be defined.
We cannot have it package private because it overrides a public method.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -206,16 +216,18 @@ public void onJobModelExpired() {
           if (SamzaContainerStatus.NOT_STARTED.equals(status) || SamzaContainerStatus.STARTED.equals(status)) {
             boolean shutdownComplete = false;
             try {
-              LOGGER.info("Shutting down container in onJobModelExpired.");
+              LOGGER.info("Shutting down container in onJobModelExpired for processor:" + pid);
               container.pause();
               shutdownComplete = jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
+              LOGGER.info("ShutdownCompete=" + shutdownComplete + ";pid=" + pid);
```
sborya: fixed and removed pid.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -65,34 +73,39 @@
   private volatile String ephemeralPath = null;
   private final ZkKeyBuilder keyBuilder;
   private final int connectionTimeoutMs;
+  private AtomicInteger currentGeneration;
```

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: I think we need to - this guarantees the same job model for the same list of ids. Sorting node names is not sufficient.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: done for both.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: it would be difficult because these are callbacks.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -315,4 +315,48 @@ public void onBarrierError(String version, Throwable t) {
...
+
+  /// listener to handle session expiration
+  public class ZkSessionStateChangedListener implements IZkStateListener {
+
+    @Override
+    public void handleStateChanged(Watcher.Event.KeeperState state)
+        throws Exception {
+      if (state == Watcher.Event.KeeperState.Expired) {
+        // if the session has expired it means that all the registration's ephemeral nodes are gone.
+        LOG.warn("Got session expired event for processor=" + processorId);
```
sborya: In general yes, but the way it is implemented in zkClient- this call will be made ONLY when a new session is setup, so it is sufficient to do it in the new session handler.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: hmm. it is true, but we HAVE to make sure it is down, before we call register again.. I don't know any other way.

samza-core/src/main/java/org/apache/samza/config/TaskConfigJava.java
navina: @sborya I think we missed to discuss about this comment yesterday. Another thing I noticed is that `TaskConfigJava` and `TaskConfig` have different defaults for `task.shutdown.ms`. Can you please fix this and also, comment on what you think about the impact on existing jobs in yarn that may rely on this default ? 
If we are changing this, we should change it everywhere in the code, update the config documentation and also, enlist it in the release notes.

shanthoosh: @navina
Instead of overriding defaults, wondering if it's acceptable to default `task.shutdown.ms` to `30s` if undefined in standAlone container launcher flow, there by not affecting yarn deployments.

navina: @shanthoosh  (just re-iterating our offline conversation) Imo, `task.shutdown.ms` should not be affected by executing environment. Changing defaults across environment in the code will confuse the user, even if it doesn't directly affect samza-yarn jobs. So, if we want to make the defaults reasonable to avoid unnecessary re-balancing attempts in standalone, it should be fine as long as it is called out clearly everywhere. 

Again, this may still impact large through-put job which can easily start lagging. This is purely due to our architectural changes. In case of Yarn, there is no question of "rebalancing" today. So, we have to find a balance that will work for both architectural choices.

PR #230, opened 2017-06-22
SAMZA-1340 - StreamProcessor does not propagate container failures from StreamTask

navina:
Storing the exception seen from the container in the `SamzaContainerListener#onFailure(Throwable)` in the StreamProcessor.
`JobCoordinator#stop` callback inspects this stored exception and invokes the correct callback for StreamProcessorLifecycleListener.
It is pretty difficult to add all test cases. Suggestion welcome for improving code/testing.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -67,6 +67,7 @@ public class StreamProcessor {
   private ExecutorService executorService;
 
   private volatile SamzaContainer container = null;
+
   // Latch used to synchronize between the JobCoordinator thread and the container thread, when the container is
   // stopped due to re-balancing
   private volatile CountDownLatch jcContainerShutdownLatch = new CountDownLatch(1);
@@ -74,6 +75,7 @@ public class StreamProcessor {
 
   @VisibleForTesting
   JobCoordinatorListener jobCoordinatorListener = null;
+  volatile Throwable containerException = null;
 
   /**
    * Create an instance of StreamProcessor that encapsulates a JobCoordinator and Samza Container
@@ -274,7 +276,8 @@ public class StreamProcessor {
               } else {
                 LOGGER.warn("JobCoordinatorLatch was null. It is possible for some component to be waiting.");
               }
-              LOGGER.error("Container failed. Stopping the processor.", t);
+              containerException = t;
+              LOGGER.error("Container failed. Stopping the processor.", containerException);
               container = null;
               stop();
             }
@@ -298,7 +301,10 @@ public class StreamProcessor {
           executorService.shutdownNow();
         }
         if (processorListener != null) {
-          processorListener.onShutdown();
+          if (containerException != null)
+            processorListener.onFailure(containerException);
+          else
+            processorListener.onShutdown();
         }
       }
 
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -67,7 +67,8 @@ public class StreamProcessor {
   private ExecutorService executorService;
 
   private volatile SamzaContainer container = null;
-
+  private volatile Throwable containerException = null;
+  
   // Latch used to synchronize between the JobCoordinator thread and the container thread, when the container is
   // stopped due to re-balancing
   private volatile CountDownLatch jcContainerShutdownLatch = new CountDownLatch(1);
@@ -75,7 +76,6 @@ public class StreamProcessor {
 
   @VisibleForTesting
   JobCoordinatorListener jobCoordinatorListener = null;
-  volatile Throwable containerException = null;
 
   /**
    * Create an instance of StreamProcessor that encapsulates a JobCoordinator and Samza Container
--- a/samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
+++ b/samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
@@ -25,7 +25,6 @@ import org.apache.samza.config.MapConfig;
 import org.apache.samza.container.RunLoop;
 import org.apache.samza.container.SamzaContainer;
 import org.apache.samza.coordinator.JobCoordinator;
-import org.apache.samza.coordinator.JobCoordinatorListener;
 import org.apache.samza.job.model.ContainerModel;
 import org.apache.samza.job.model.JobModel;
 import org.apache.samza.metrics.MetricsReporter;
@@ -33,8 +32,6 @@ import org.apache.samza.task.StreamTask;
 import org.apache.samza.task.StreamTaskFactory;
 import org.junit.Assert;
 import org.junit.Test;
-import org.mockito.invocation.InvocationOnMock;
-import org.mockito.stubbing.Answer;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -42,7 +39,6 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
-import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.doReturn;
@@ -74,22 +70,22 @@ public class TestStreamProcessor {
       if (container == null) {
         RunLoop mockRunLoop = mock(RunLoop.class);
         doAnswer(invocation ->
-        {
-          try {
-            runLoopStartForMain.countDown();
-            containerStop.await();
-          } catch (InterruptedException e) {
-            System.out.println("In exception" + e);
-            e.printStackTrace();
-          }
-          return null;
-        }).when(mockRunLoop).run();
+          {
+            try {
+              runLoopStartForMain.countDown();
+              containerStop.await();
+            } catch (InterruptedException e) {
+              System.out.println("In exception" + e);
+              e.printStackTrace();
+            }
+            return null;
+          }).when(mockRunLoop).run();
 
         doAnswer(invocation ->
-        {
-          containerStop.countDown();
-          return null;
-        }).when(mockRunLoop).shutdown();
+          {
+            containerStop.countDown();
+            return null;
+          }).when(mockRunLoop).shutdown();
         container = StreamProcessorTestUtils.getDummyContainer(mockRunLoop, mock(StreamTask.class));
       }
       return container;
@@ -184,7 +180,7 @@ public class TestStreamProcessor {
    * through the StreamProcessor
    *
    * Assertions:
-   * - JobCoordinator has been stopped can be asserted from the JobCoordinatorListener callback
+   * - JobCoordinator has been stopped from the JobCoordinatorListener callback
    * - StreamProcessorLifecycleListener#onFailure(Throwable) has been invoked
    */
   @Test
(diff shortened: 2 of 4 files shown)


navina:
@nickpan47 @sborya Can you guys please review this change? Please suggest if there are ways to improve the solution.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
```diff
@@ -165,10 +175,98 @@ public void onFailure(Throwable t) {
...
+  public void testContainerFailureCorrectlyStopsProcessor() {
+    JobCoordinator mockJobCoordinator = mock(JobCoordinator.class);
+    Throwable expectedThrowable =  new SamzaException("Failure in Container!");
+    AtomicReference<Throwable> actualThrowable = new AtomicReference<>();
+    final CountDownLatch runLoopStartedLatch = new CountDownLatch(1);
+    RunLoop failingRunLoop = mock(RunLoop.class);
+    doAnswer(invocation ->
+      {
+        try {
+          Thread.sleep(10);
```
cpettitt-linkedin: Why is this sleep necessary?

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
```diff
@@ -165,10 +175,98 @@ public void onFailure(Throwable t) {
...
+    TestableStreamProcessor processor = new TestableStreamProcessor(
+        new MapConfig(),
+        new HashMap<>(),
+        mock(StreamTaskFactory.class),
+        new StreamProcessorLifecycleListener() {
+          @Override
+          public void onStart() { }
+
+          @Override
+          public void onShutdown() { }
```
cpettitt-linkedin: It would be nice to verify that shutdown is not also called when the processor fails due to a user exception.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
```diff
@@ -165,10 +175,98 @@ public void onFailure(Throwable t) {
...
+              e.printStackTrace();
+            }
+          }).start();
+        return null;
+      }).when(mockJobCoordinator).start();
+
+    try {
+      processor.start();
+
+      // This block is required for the mockRunloop is actually start.
```
cpettitt-linkedin: s/start/started/

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
```diff
@@ -115,19 +130,14 @@ public void onFailure(Throwable t) {
```
cpettitt-linkedin: Similar to below, it is worth asserting that failure is not called on a clean shutdown (though I'm less worried about this case).

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
```diff
@@ -165,10 +175,98 @@ public void onFailure(Throwable t) {
...
+
+      // This block is required for the mockRunloop is actually start.
+      // Otherwise, processor.stop gets triggered before mockRunloop begins to block
+      runLoopStartedLatch.await();
+      Assert.assertTrue(
+          "Container failed and processor listener failed was not invoked within timeout!",
+          processorListenerFailed.await(30, TimeUnit.SECONDS));
+      Assert.assertEquals(expectedThrowable, actualThrowable.get());
+
+    } catch (InterruptedException e) {
```
cpettitt-linkedin: We should not be expecting the test to be interrupted here, so it would be reasonable to throw here instead of swallowing the exception.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
```diff
@@ -165,10 +175,98 @@ public void onFailure(Throwable t) {
...
+          }).start();
+        return null;
+      }).when(mockJobCoordinator).start();
+
+    try {
+      processor.start();
+
+      // This block is required for the mockRunloop is actually start.
+      // Otherwise, processor.stop gets triggered before mockRunloop begins to block
+      runLoopStartedLatch.await();
```
cpettitt-linkedin: Why not wait on the processor to finish here instead? That is what users would do.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
navina: Yeah. it will be good to assert it. I will add that.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
navina: Nice catch. This is not needed here.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
navina: got it.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
navina: Yeah. The blocking call to wait until processor finishes is in the LocalApplicationRunner. So, the alternative is to use the StreamProcessorLifecycleListener knobs to wait on events.  I find the listener to be more convenient. What you are suggesting is also a valid point if I were using LocalApplicationRunner.
Let me take a look and see if there are any integrations tests with LocalApplicationRunner where I can assert this test.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
navina: Sure. Sounds good.

navina:
@cpettitt @nickpan47 @sborya  Can you please review this PR? It is slowing down testing for a lot of us. Thanks!

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
```diff
@@ -43,53 +48,85 @@
...
+  @After
+  public void after() {
+    processorListenerState.clear();
+  }
 
   class TestableStreamProcessor extends StreamProcessor {
     private final CountDownLatch containerStop = new CountDownLatch(1);
     private final CountDownLatch runLoopStartForMain = new CountDownLatch(1);
-    private SamzaContainer containerReference = null;
+    public SamzaContainer container = null;
```
cpettitt-linkedin: Is it possible to make this private? I didn't see an obvious reason why it is public, but it's quite possible I missed it.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
```diff
@@ -43,53 +48,85 @@
 import static org.mockito.Mockito.when;
 
 public class TestStreamProcessor {
+  private Map<String, Boolean> processorListenerState;
```
cpettitt-linkedin: A ConcurrentMap would be appropriate here if there is a chance that we are (or will) invoke a state transition from another thread.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
```diff
@@ -43,53 +48,85 @@
...
+  private static final String ON_START = "onStart";
+  private static final String ON_SHUTDOWN = "onShutdown";
+  private static final String ON_FAILURE = "onFailure";
+
```
cpettitt-linkedin: An enum would be more suitable here because it nicely groups the possible states.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
```diff
@@ -43,53 +48,85 @@
...
+  @Before
+  public void before() {
+    processorListenerState = new HashMap<String, Boolean>() {
+      {
+        put(ON_START, false);
+        put(ON_FAILURE, false);
+        put(ON_SHUTDOWN, false);
+      }
```
cpettitt-linkedin: An alternative here would be to use a Set instead of a Map and assume that membership in the set means that the corresponding event was detected. Clearing out the set is simply a call to .clear(). If you prefer Map then going with an enum and iterating over the values would be less error prone for additional states.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
```diff
@@ -147,28 +181,119 @@ public void onFailure(Throwable t) {
...
-    } catch (InterruptedException e) {
-      e.printStackTrace();
-    }
+    processorListenerStop.await();
+
+    // Assertions on which callbacks are expected to be invoked
+    Assert.assertTrue(processorListenerState.get(ON_START));
+    Assert.assertTrue(processorListenerState.get(ON_SHUTDOWN));
+    Assert.assertFalse(processorListenerState.get(ON_FAILURE));
+  }
```
cpettitt-linkedin: Optional: you could respecify this as Set (or Map) equality on the expected states. It would give you slightly more contextual information in the even of a test failure (e.g. possibly detecting a second hidden failure).

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
navina: It will be from another thread. You are right. It should concurrent.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
navina: I will stick with using the Map for now, as it seems *slightly* easier to read the code and assertions. If it gets too verbose, I will modify them to a set.

samza-core/src/test/java/org/apache/samza/processor/TestStreamProcessor.java
navina: Yes. It should be the same as before with containerReference. Will fix it.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -274,7 +276,8 @@ public void onContainerFailed(Throwable t) {
               } else {
                 LOGGER.warn("JobCoordinatorLatch was null. It is possible for some component to be waiting.");
               }
-              LOGGER.error("Container failed. Stopping the processor.", t);
+              containerException = t;
```
sborya: All this logic becomes very confusing with a different ways of invoking stop() from different points of the code.
May be we should think about some kind of doStop(Throwable t) - to differentiate between different stops, instead trying to figure every time why are we stopping - for an error, or normal.

navina: @sborya I agree it is pretty convoluted how we are doing this. However, @prateekm and I have discussed this option extensively in the past when we are working on [SAMZA-1212](https://issues.apache.org/jira/browse/SAMZA-1212). 
As I explained before , it is not fully clear how to wire-in the source of the stop and cause of the stop of a streamprocessor across the components. I don't believe it is as straightforward as passing in a `Throwable`. It could work. It becomes particularly tricky for bounded jobs or jobs that decide to stop themselves using TaskCoordinator. The approach you are suggesting is equivalent to threading a needle , where we will end up passing around the state/source of the `stop` across the components' api and callbacks. 

The alternative and more straightforward solution, imo, is to clearly define a state model for each components and persist/manage the status in the streamprocessor. That way, it will make it easier for streamprocessor to take remediation steps.

We can discuss more on this. However, this PR is simply a bug-fix. We should scrutinize the refactoring/behavior of the APIs in a separate thread/JIRA. What do you think?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -274,7 +276,8 @@ public void onContainerFailed(Throwable t) {
...
+              LOGGER.error("Container failed. Stopping the processor.", containerException);
               container = null;
               stop();
```
sborya: Another problem, that this will cause the stop() to run in the a different thread (container thread).
May be we should invoke the stop from JobCoordinator on a debounce Thread?

navina: >Another problem, that this will cause the stop() to run in the a different thread (container thread).

Yes. We knew that it will be invoked from a different thread. We assumed that all tasks in the ZkJobCoordinator queue would stop when the executor service is shutdown. However, there isn't a lot we can do other than interrupting the thread. 

> May be we should invoke the stop from JobCoordinator on a debounce Thread?

Yes. That is an option we can explore. Thanks for creating jiras to track this issue.

PR #231, opened 2017-06-22
SAMZA-1337: Use StreamTask with the LocalApplicationRunner

sborya:


samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -201,7 +202,7 @@ private String createProcessorId(Config config) {
     ApplicationConfig appConfig = new ApplicationConfig(config);
     if (appConfig.getProcessorId() != null) {
       return appConfig.getProcessorId();
-    } else if (appConfig.getAppProcessorIdGeneratorClass() != null) {
+    } else if (!StringUtils.isEmpty(appConfig.getAppProcessorIdGeneratorClass())) {
```
bharathkk: Can we replace `isEmpty` w/ `isBlank` so that values like `"   "` are also not accepted?
btw, commons does have `isNotEmpty`, `isNotBlank`. IMO, it makes it more readable and explicit.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -35,12 +37,32 @@
...
+  @Override
+  public void runTask() {
+    JobConfig jobConfig = new JobConfig(this.config);
+
+    // validation
+    String taskName = new TaskConfig(config).getTaskClass().get();
+    if (taskName == null) {
+      throw new SamzaException("Neither APP nor task.class are defined defined");
+    }
+    LOG.info("LocalApplicationRunner will run " + taskName);
```
bharathkk: np: s/LocalApplicationRunner/RemoteApplicationRunner

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -134,6 +136,24 @@ public LocalApplicationRunner(Config config) {
   }
 
   @Override
+  public void runTask() {
```
xinyuiscool: do we need to support stopTask() { }?

navina: Can we not make the interface runTask(TaskFactory) ? That way, we can avoid class-loading from config and users can easily provide a taskfactory that produces a StreamTask or AsynsStreamTask. This will make it easier to integrate with dependency injection systems (like offspring) and it will look more like a library. 
I understand that StreamTask or AsyncStreamTask are two independent interfaces. We should be able to inspect that easily.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -35,12 +37,32 @@
  */
 public class RemoteApplicationRunner extends AbstractApplicationRunner {
 
-  private static final Logger log = LoggerFactory.getLogger(RemoteApplicationRunner.class);
+  private static final Logger LOG = LoggerFactory.getLogger(RemoteApplicationRunner.class);
```
navina: QQ: How is RemoteApplicationRunner invoked? Isn't it through some wrapper shell script and user never sees this class? Need clarification as we have been back-and-forth on this interface discussion.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
sborya: sure.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
sborya: done. thanks.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
sborya: we discussed it, and decided that this is a quick fix for people to try it with least amount of changes, and this API will be hidden with Yi's API changes later. And then it will most likely change to accept a factory, as you described.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -134,6 +136,24 @@ public LocalApplicationRunner(Config config) {
...
+    JobConfig jobConfig = new JobConfig(this.config);
+
+    // validation
+    String taskName = new TaskConfig(config).getTaskClass().get();
```
sborya: done.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -35,12 +37,32 @@
...
   public RemoteApplicationRunner(Config config) {
     super(config);
   }
 
+  @Override
+  public void runTask() {
+    JobConfig jobConfig = new JobConfig(this.config);
+
+    // validation
+    String taskName = new TaskConfig(config).getTaskClass().get();
```
sborya: I think it is too small for this. Will be kind of weird method.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
sborya: I think for the purposes of these PR we will not implement RemoteApplicationRunner for this.

navina: thank you!

xinyuiscool: Hey, if we are not implementing in the Remote, can we remove the runTask() method in ApplciationRunner so we only add it in LocalApplicationRunner?  Thx

navina: @xinyuiscool I prefer it to be in the abstract class itself as it is much cleaner. It is pretty common for interfaces to throw exception when an operation is not supported. Any particular reason you want to keep it only in LocalApplicationRunner?

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -68,6 +68,11 @@ public LocalContainerRunner(JobModel jobModel, String containerId) {
   }
 
   @Override
+  public void runTask() {
+    throw new SamzaException("Running StreamTask is not implemented for LocalContainerRunner");
```
navina: maybe use `UnsupportedOperationException` ?

samza-core/src/test/java/org/apache/samza/runtime/TestLocalApplicationRunner.java
```diff
@@ -205,6 +206,35 @@ public void countDown() {
...
+        StreamProcessorLifecycleListener listener = captor.getValue();
+        listener.onStart();
+        listener.onShutdown();
+        return null;
+      }).when(sp).start();
+
+    LocalApplicationRunner spy = spy(runner);
+    doReturn(sp).when(spy).createStreamProcessor(anyObject(), anyObject(), captor.capture());
+
+    spy.run(null);
```
navina: Shouldn't we be testing spy.runTask here?

PR #232, opened 2017-06-26
SAMZA-1346: GroupByContainerCount.balance() should guard against null…

jmakes:
… LocalityManager

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
@@ -89,6 +89,11 @@ public class GroupByContainerCount implements BalancingTaskNameGrouper {
 
     validateTasks(tasks);
 
+    if (localityManager == null) {
+      log.info("Locality manager is null. Cannot read or write task assignments. Invoking grouper.");
+      return group(tasks);
+    }
+
     TaskAssignmentManager taskAssignmentManager = localityManager.getTaskAssignmentManager();
     List<TaskGroup> containers = getPreviousContainers(taskAssignmentManager, tasks.size());
     if (containers == null || containers.size() == 1 || containerCount == 1) {
--- a/samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
+++ b/samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
@@ -857,4 +857,33 @@ public class TestGroupByContainerCount {
     new GroupByContainerCount(3).balance(taskModels, localityManager); //Should throw
 
   }
+
+  @Test
+  public void testBalancerWithNullLocalityManager() {
+    Set<TaskModel> taskModels = generateTaskModels(3);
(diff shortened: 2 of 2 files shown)


samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
```diff
@@ -857,4 +857,33 @@ public void testBalancerThrowsOnNonIntegerContainerIds() {
...
+      containersMap.put(container.getProcessorId(), container);
+    }
+    assertEquals(3, containers.size());
+    ContainerModel container0 = containersMap.get("0");
+    ContainerModel container1 = containersMap.get("1");
+    ContainerModel container2 = containersMap.get("2");
+    assertNotNull(container0);
+    assertNotNull(container1);
+    assertNotNull(container2);
+    assertEquals("0", container0.getProcessorId());
```
cpettitt-linkedin: Isn't this redundant in the sense that this is by definition per the loop at 868?

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
```diff
@@ -857,4 +857,33 @@ public void testBalancerThrowsOnNonIntegerContainerIds() {
...
+    assertEquals("1", container1.getProcessorId());
+    assertEquals("2", container2.getProcessorId());
+    assertEquals(1, container0.getTasks().size());
+    assertEquals(1, container1.getTasks().size());
+    assertEquals(1, container2.getTasks().size());
+
+    assertTrue(container0.getTasks().containsKey(getTaskName(0)));
```
cpettitt-linkedin: This looks to be the meat of the test if I understand. Couldn't you collapse this and the checks in 881-883 into a set equality test? The benefit would be that if there is any difference we actually see the difference in the test failure.

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
```diff
@@ -89,6 +89,11 @@ public GroupByContainerCount(int containerCount) {
 
     validateTasks(tasks);
 
+    if (localityManager == null) {
```
navina: I just made a comment in the JIRA regarding using `job.host-affinity.enabled` config. If you use that config, that can be passed from the factory to the GroupByContainerCount class. If host-affinity is enabled and localityManager is not defined, then it should throw an exception.

jmakes: Responded in the JIRA

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
jmakes: This is a great suggestion. I've applied it to this test and all the others that compare the output of group() and balance(). 

https://github.com/jmakes/samza/commit/9e0ec1b72c65721b2f5f6682f9a823a11ca36a72

I just need github to pick up the change in this PR. Stay tuned

PR #233, opened 2017-06-26
SAMZA-1347: GroupByContainerIds NPE if containerIds list is null

jmakes:


jmakes:
@navina @sborya take a look!

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerIds.java
```diff
@@ -60,16 +54,19 @@ public GroupByContainerIds(int count) {
   }
 
   public Set<ContainerModel> group(Set<TaskModel> tasks, List<String> containersIds) {
+    if (containersIds == null)
+      return this.group(tasks);
+
+    if (containersIds.isEmpty())
+      throw new IllegalArgumentException("Must have at least one container");
+
```
bharathkk: Not sure if its just me who feels slightly weird about this contract about mismatch is behavior.
Although I agree semantically `null` and `empty` lists mean two different things, I have often seen them being used interchangeably or at the least have similar behavior in many codebases. 
Just putting in a comment to see others take on it.

Feel free to drop my comment if rest of the team is fine with it.

jmakes: Yeah, I could see how it looks weird. I think it's just fall out from the way this grouper was designed. Other groupers only get a container count. This one uses the container count if there's no container list, otherwise it uses the size of the container list. So it is purposely trying to fall back to the old behavior if the container list is null. And when it does that, you can see that it generates a container list from the container count and then calls this method again. So this check is to ensure that in both cases (when a container list is provided by the caller AND when it is generated) we never have 0 containers. An error is thrown for all the other groupers when container count is 0. 

So this was a best effort at catching the same case in this grouper, while still allowing the fall back to the old grouper signature when container list is null.

PR #235, opened 2017-06-28
SAMZA-1324: Fix NullPointerException in ZkUtils api's.

shanthoosh:
Problem:
Read/Write api methods in ZkUtils updates counters/timers in `metrics` field. In a ZkUtils constructor this fields is not initialized properly. Java default for uninitialized field is null resulting in NPE. 

Fix:
Initialize private fields of ZkUtils class with appropriate defaults.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
@@ -40,9 +40,7 @@ public class ZkCoordinationServiceFactory implements CoordinationServiceFactory 
     ZkClient zkClient =
         createZkClient(zkConfig.getZkConnect(), zkConfig.getZkSessionTimeoutMs(), zkConfig.getZkConnectionTimeoutMs());
 
-    ZkJobCoordinatorMetrics zkJobCoordinatorMetrics = new ZkJobCoordinatorMetrics(new NoOpMetricsRegistry());
-
-    ZkUtils zkUtils = new ZkUtils(new ZkKeyBuilder(groupId), zkClient, zkConfig.getZkConnectionTimeoutMs(), zkJobCoordinatorMetrics);
+    ZkUtils zkUtils = new ZkUtils(new ZkKeyBuilder(groupId), zkClient, zkConfig.getZkConnectionTimeoutMs(), new NoOpMetricsRegistry());
 
     return new ZkCoordinationUtils(participantId, zkConfig, zkUtils);
   }
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
@@ -25,13 +25,14 @@ import org.apache.samza.config.Config;
 import org.apache.samza.config.ZkConfig;
 import org.apache.samza.coordinator.CoordinationServiceFactory;
 import org.apache.samza.coordinator.CoordinationUtils;
+import org.apache.samza.util.NoOpMetricsRegistry;
 import org.apache.zookeeper.client.ConnectStringParser;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 
 public class ZkCoordinationServiceFactory implements CoordinationServiceFactory {
-  private final static Logger LOG = LoggerFactory.getLogger(ZkCoordinationServiceFactory.class);
+  private static final Logger LOG = LoggerFactory.getLogger(ZkCoordinationServiceFactory.class);
 
   public CoordinationUtils getCoordinationService(String groupId, String participantId, Config config) {
     ZkConfig zkConfig = new ZkConfig(config);
@@ -39,7 +40,9 @@ public class ZkCoordinationServiceFactory implements CoordinationServiceFactory 
     ZkClient zkClient =
         createZkClient(zkConfig.getZkConnect(), zkConfig.getZkSessionTimeoutMs(), zkConfig.getZkConnectionTimeoutMs());
 
-    ZkUtils zkUtils = new ZkUtils(new ZkKeyBuilder(groupId), zkClient, zkConfig.getZkConnectionTimeoutMs());
+    ZkJobCoordinatorMetrics zkJobCoordinatorMetrics = new ZkJobCoordinatorMetrics(new NoOpMetricsRegistry());
+
+    ZkUtils zkUtils = new ZkUtils(new ZkKeyBuilder(groupId), zkClient, zkConfig.getZkConnectionTimeoutMs(), zkJobCoordinatorMetrics);
 
     return new ZkCoordinationUtils(participantId, zkConfig, zkUtils);
   }
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
@@ -31,6 +31,7 @@ import org.I0Itec.zkclient.ZkConnection;
 import org.I0Itec.zkclient.exception.ZkInterruptedException;
 import org.apache.samza.SamzaException;
 import org.apache.samza.job.model.JobModel;
+import org.apache.samza.metrics.MetricsRegistry;
 import org.apache.samza.serializers.model.SamzaObjectMapper;
 import org.apache.zookeeper.data.Stat;
 import org.codehaus.jackson.map.ObjectMapper;
@@ -65,13 +66,13 @@ public class ZkUtils {
   private volatile String ephemeralPath = null;
   private final ZkKeyBuilder keyBuilder;
   private final int connectionTimeoutMs;
-  private final ZkJobCoordinatorMetrics metrics;
+  private final ZkUtilsMetrics metrics;
 
-  public ZkUtils(ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs, ZkJobCoordinatorMetrics metrics) {
+  public ZkUtils(ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs, MetricsRegistry metricsRegistry) {
     this.keyBuilder = zkKeyBuilder;
     this.connectionTimeoutMs = connectionTimeoutMs;
     this.zkClient = zkClient;
-    this.metrics = metrics;
+    this.metrics = new ZkUtilsMetrics(metricsRegistry);
   }
 
   public void connect() throws ZkInterruptedException {
@@ -263,7 +264,9 @@ public class ZkUtils {
    * @return jobmodel version as a string
    */
   public String getJobModelVersion() {
-    return zkClient.<String>readData(keyBuilder.getJobModelVersionPath());
+    String jobModelVersion = zkClient.readData(keyBuilder.getJobModelVersionPath());
+    metrics.reads.inc();
+    return jobModelVersion;
   }
 
   /**
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
@@ -65,13 +65,7 @@ public class ZkUtils {
   private volatile String ephemeralPath = null;
   private final ZkKeyBuilder keyBuilder;
   private final int connectionTimeoutMs;
-  private ZkJobCoordinatorMetrics metrics;
-
-  public ZkUtils(ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs) {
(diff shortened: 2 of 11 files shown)


shanthoosh:
@navina @PawasChhokra 

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -65,12 +66,10 @@
...
   private final ZkKeyBuilder keyBuilder;
   private final int connectionTimeoutMs;
-  private ZkJobCoordinatorMetrics metrics;
+  private final ZkJobCoordinatorMetrics metrics;
 
   public ZkUtils(ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs) {
-    this.keyBuilder = zkKeyBuilder;
-    this.connectionTimeoutMs = connectionTimeoutMs;
-    this.zkClient = zkClient;
+    this (zkKeyBuilder, zkClient, connectionTimeoutMs, new ZkJobCoordinatorMetrics(new MetricsRegistryMap()));
```
navina: Is this constructor used only for testing? 

If not, it should not be creating a separate ZkJobCoordinatorMetrics instance with a new metrics registry. You should wire-in the MetricsRegistry from the caller and use it in a `ZkUtilsMetrics` class, although I don't see a value of separating the metrics encapsulating class here. Alternative is to wire-in the metrics instance from the caller itself and remove this constructor.

shanthoosh: Q1: 
No, it's used at other places as well(`ZkCoordinationServiceFactory`).

Q2:

Instantiation of the MetricsRegistry don’t change with any configuration parameter provided by client of this constructor(In essence not tunable by any configuration). 

This is more like convenient shorthand, if there’re multiple places if this is called and all of them have to do the same instantiation again and again, I prefer moving it one level up here to minimize redundancy . 

What is the disadvantage of having like this and why should the client be aware of this dependency (it doesn’t affect normal operability of any api)?? 

If by mistake, this param is set to null, some ZkUtils api's will break with NPE. To prevent that from happening, we can  delete the other constructor.

Please let me know your thoughts or if i’m missing something.

shanthoosh: Please check my changes and verify if it makes sense. Though I partly agree with you on approach, lets at-least pass ZkUtilsMetrics into ZkUtils constructor. Passing in ZkJobCoordinatorMetrics into ZkUtils constructor looks bizarre to me.
 
I'm bringing in new Reporters in `ZkCoordinationServiceFactory `, not sure how much resource footprint it will create.

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
```diff
@@ -19,32 +19,57 @@
...
   public CoordinationUtils getCoordinationService(String groupId, String participantId, Config config) {
     ZkConfig zkConfig = new ZkConfig(config);
 
     ZkClient zkClient =
         createZkClient(zkConfig.getZkConnect(), zkConfig.getZkSessionTimeoutMs(), zkConfig.getZkConnectionTimeoutMs());
 
-    ZkUtils zkUtils = new ZkUtils(new ZkKeyBuilder(groupId), zkClient, zkConfig.getZkConnectionTimeoutMs());
+    ZkJobCoordinatorMetrics zkJobCoordinatorMetrics = getJobCoordinatorMetrics(config, participantId);
+
+    ZkUtils zkUtils = new ZkUtils(new ZkKeyBuilder(groupId), zkClient, zkConfig.getZkConnectionTimeoutMs(), zkJobCoordinatorMetrics);
```
navina: I think handling null metrics instance is a good idea, atm.

shanthoosh: `isLeader` metric reported from ZkJobCoordinator will be useful for CoordinationService when we run in production. 
Ex: When streamApplication death happens after acquiring lock and before creating intermediate streams, we could easily track those scenarios with metrics.

navina: I am not opposed to having this `isLeader` metric or any metric for CoordinationService. Here are the reasons why I think the solution provided here won't work or is not comprehensive:
1. Merely creating the metrics reporter and registering the metrics will not actually report any metrics. We need to start/stop the metrics reporter , aka manage its lifecycle. Until we do that, we really don't have any metrics.
2. I just noticed this today. The `participantId` used in ZkCoordinationService is a random UUID. I suspect that is because of the changes from SEP-1 that CoordinationService does not have access to the processorId. So, in reality, we will be getting metrics around a whole bunch of different *processorIds*. I am not sure if you can actually associate a temporary uuid with any processor itself. Please take a look at [LocalApplicationRunner Line 134](https://github.com/apache/samza/blob/4875842b37e0285100e2d1d753a7bc4a1448e897/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java#L134)

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -65,13 +65,7 @@
...
-  private ZkJobCoordinatorMetrics metrics;
-
-  public ZkUtils(ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs) {
-    this.keyBuilder = zkKeyBuilder;
-    this.connectionTimeoutMs = connectionTimeoutMs;
-    this.zkClient = zkClient;
-  }
+  private final ZkJobCoordinatorMetrics metrics;
 
   public ZkUtils(ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs, ZkJobCoordinatorMetrics metrics) {
```
navina: It will be cleaner to simply pass the metrics registry associated with this component and register more granular group of metrics under `ZkUtilsMetrics`. Overloading it with `ZkJobCoordinatorMetrics` is confusing as this component - ZkUtils is also accessed from CoordinationService. 

If you pass MetricsRegistry into this constructor, it can work for CoordinationService as well, whenever we get to completing SAMZA-1153 (pull out entire metric reporter lifecycle into localapplicationrunner). For now, it seems like we should be creating ZkUtils metrics as a subset of ZkJobCoordinatorMetrics and register it with Metricsregistry provided.

```
public ZkUtils (ZkKeyBuilder zkKeyBuilder, ZkClient zkClient, int connectionTimeoutMs, MetricsRegistry metricsRegistry) {
...
this.zkUtilsMetrics = new ZkUtilsMetrics();
metricsRegistry.register(this.zkUtilsMetrics);
...
}
```

If you have more questions on why we are passing around MetricsRegistry and the pattern used to de-couple a component's metrics instance from the reporting itself, take a look at the design in [dropwizard.io](https://github.com/dropwizard/dropwizard)
HTH. Thanks!

shanthoosh: >Overloading it with ZkJobCoordinatorMetrics is confusing as this component - ZkUtils is also accessed from CoordinationService.

Just to be clear. ZKJobCoordinatorMetrics dependency into ZkUtils wasn’t introduced in this patch(as part of this change). It was how it was existent before. 

Just to fix a NPE, I didn’t want to get into refactoring stuff.

I agree with passing in just MetricsRegistry(that’s how it's done in rest of samza viz SamzaContainerMetrics, LocalStoreMonitorMetrics etc). 

Will make the change and update here.

> If you have more questions on why we are passing around MetricsRegistry and the

Thanks for the pointer, I will read about it. 

Though we achieve isolation in terms of registering/creating  metics and publishing metrics , following this pattern introduces dependency between different layers. 

Basic premise here is each layer expects the previous layer creating it to have a metricsRegistry(map) to pass it to current layer. There’s a fan out from top layer (who registers with reporters) to all lower level.  

In chain A-B-C if A registers metricsRegistry with reporters and B doesn’t need MetricsRegistry for reporting, it still has to have it as explicit dependency 
since C might require it to report metrics(which it should receive from B). 

If you already know the reasoning please share it with me(or poet out why this thinking is incorrect).

navina: > Just to fix a NPE, I didn’t want to get into refactoring stuff.

Yes. fixing NPE is what this PR is about. If you are thinking of refactoring how metrics are wired in the new APIs, let's solve it in a separate PR. 

> Though we achieve isolation in terms of registering/creating metics and publishing metrics , following this pattern introduces dependency between different layers.
> Basic premise here is each layer expects the previous layer creating it to have a metricsRegistry(map) to pass it to current layer. There’s a fan out from top layer (who registers with reporters) to all lower level.

Yes. That's right. If the dependencies are skipping across layers, then it is perhaps an opportunity to step-back and revisit the component layout itself, rather than directly solving for dependency injection.

> In chain A-B-C if A registers metricsRegistry with reporters and B doesn’t need MetricsRegistry for reporting, it still has to have it as explicit dependency
since C might require it to report metrics(which it should receive from B).

Yes. That's right. If A manages the component B and B manages C, B will be a passthrough for any dependency of component C on A. There isn't another way around it, afaik unless we have a separate dependency injection system.

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
shanthoosh: >Q1: 

Agree with you. We could tie up the reporters lifecycle with ZkCoordinatorService. We can start reporters in ZkCoordinatorService constructor and stop reporters in reset(tieing reporters in ZkCoordinationService is synonymous to ZkJobCoordinator approach of reporter lifecycle management).

>Q2: 

I was with wrong notion that this is synonymous to processorID. There doesn't seem to be obvious way to address the random participant problem(Utilizing processorId for this doesn't look logical at LocalApplicationRunner level. Currently all stream application have same app id, app name. If acceptable, we can report metrics with (appid,appname) & use hostname to solve this problem partially and create a JIRA for proper solution later).

navina: > I was with wrong notion that this is synonymous to processorID. There doesn't seem to be obvious way to address the random participant problem.

Yeah. Even I thought so until I stumbled into it today. This should hopefully be fixed with SEP-2 although I haven't seen the latest draft. API interaction between ApplicationRunner and StreamProcessors are kind of messed up right now. 

> Utilizing processorId for this doesn't look logical at LocalApplicationRunner level. Currently all stream application have same app id, app name. If acceptable, we can report metrics with (appid,appname) & use hostname to solve this problem partially and create a JIRA for proper solution later).

Yeah. We will have into this issue. But let's do it  hand-in-hand with SEP-2. I think it will be pretty influential in making these design choices.

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationServiceFactory.java
```diff
@@ -19,32 +19,57 @@
...
     return new ZkCoordinationUtils(participantId, zkConfig, zkUtils);
   }
 
   /**
+   * Builds {@link ZkJobCoordinatorMetrics} and registers it with MetricsReporters.
+   * @param config required to build ZkJobCoordinatorMetrics.
+   * @param participantId a unique id that identifies the participant.
+   * @return the built ZkJobCoordinatorMetrics.
+   */
+  private ZkJobCoordinatorMetrics getJobCoordinatorMetrics(Config config, String participantId) {
```
navina: Can we please remove any "JobCoordinatorMetrics" from CoordinatorService layer for now?
1. It is not really being used ( as we discussed ) to generate/report any sensible metrics
2. Creating "JobCoordinatorMetrics" object for CoordinationService is confusing. 

For now, let's just pass null metrics into ZkUtils and handle it in ZkUtils. This is a lot of unused code, which I suspect will again change significantly after SEP-2.

shanthoosh: Agreed. Update the patch based upon offline discussions.

Passed in ZkJobCoordinatorMetrics with NoopsMetricsRegistry, instead of null. Let's create a JIRA to revisit this in the future.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -269,7 +264,9 @@ public JobModel getJobModel(String jobModelVersion) {
    * @return jobmodel version as a string
    */
   public String getJobModelVersion() {
-    return zkClient.<String>readData(keyBuilder.getJobModelVersionPath());
+    String jobModelVersion = zkClient.readData(keyBuilder.getJobModelVersionPath());
```
navina: thanks for identifying the gaps and fixing it! 👍

PR #237, opened 2017-06-29
SAMZA-1355 : Enable standalone integration tests conditionally in build.

shanthoosh:
It's run only when includeSamzaTest gradle project property is set.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -45,3 +45,19 @@ rootProject.children.each {
     it.name = it.name + "_" + scalaVersion
   }
 }
+
+/**
+ * Skips execution of all tasks for the project 'samza-test'.
+ * To include `samza-test`, initiate gradle build by setting the property 'includeSamzaTest'.
+ */
+def propertyName = "includeSamzaTest"
+def projectName = "samza-test"
+gradle.taskGraph.whenReady { taskGraph ->
+  taskGraph.getAllTasks().each { task ->
+    def project = task.getProject()
+    task.onlyIf {
+      // Closure, returns false for samza-test(will get filtered) and true for all other modules(will remain unfiltered).
+      project.hasProperty(propertyName) || !project.getName().contains(projectName)
+    }
+  }
+}
--- a/settings.gradle
+++ b/settings.gradle
@@ -47,18 +47,18 @@ rootProject.children.each {
 }
 
 /**
- * Skips execution of all tasks for the project 'samza-test'.
- * To include `samza-test`, initiate gradle build by setting the property 'includeSamzaTest'.
- * To build with samza-test: './gradlew clean build -PincludeSamzaTest'
+ * Skips execution of all integration tests in project 'samza-test'.
+ * To run integration tests in samza-test: './gradlew clean build -PrunIntegrationTests'
  */
-def propertyName = "includeSamzaTest"
-def projectName = "samza-test"
 gradle.taskGraph.whenReady { taskGraph ->
   taskGraph.getAllTasks().each { task ->
     def project = task.getProject()
     task.onlyIf {
-      // Closure, returns false for samza-test(will get filtered) and true for all other modules(will remain unfiltered).
-      project.hasProperty(propertyName) || !project.getName().contains(projectName)
+      /**
+       * Behaves as pass through filter for all tasks when `runIntegrationTests` property is turned on.
+       * Filters 'test' task of 'samza-test' project otherwise.
+       */
+      project.hasProperty("runIntegrationTests") || !(project.getName().contains("samza-test") && task.getName() == "test")
     }
   }
 }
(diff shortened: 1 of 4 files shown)


shanthoosh:
@navina @sborya 

settings.gradle
```diff
@@ -45,3 +45,19 @@ rootProject.children.each {
     it.name = it.name + "_" + scalaVersion
   }
 }
+
+/**
+ * Skips execution of all tasks for the project 'samza-test'.
+ * To include `samza-test`, initiate gradle build by setting the property 'includeSamzaTest'.
```
sborya: please add an example here on how to do it.

shanthoosh:
@navina 
We decided to disable integration tests by default in local builds and run them by default in jenkins. 

If acceptable, please add property `includeSamzaTest ` to jenkins gradle cmd(I don't have edit permissions in jenkins).

settings.gradle
```diff
@@ -45,3 +45,20 @@ rootProject.children.each {
...
+ * To build with samza-test: './gradlew clean build -PincludeSamzaTest'
+ */
+def propertyName = "includeSamzaTest"
+def projectName = "samza-test"
+gradle.taskGraph.whenReady { taskGraph ->
+  taskGraph.getAllTasks().each { task ->
+    def project = task.getProject()
+    task.onlyIf {
+      // Closure, returns false for samza-test(will get filtered) and true for all other modules(will remain unfiltered).
+      project.hasProperty(propertyName) || !project.getName().contains(projectName)
```
prateekm: How does the first condition `project.hasProperty(propertyName)` work? Is the property inherited by every project when set? If so, preceding comment is incorrect/incomplete.

settings.gradle
```diff
@@ -45,3 +45,20 @@ rootProject.children.each {
...
+ * To include `samza-test`, initiate gradle build by setting the property 'includeSamzaTest'.
+ * To build with samza-test: './gradlew clean build -PincludeSamzaTest'
+ */
+def propertyName = "includeSamzaTest"
+def projectName = "samza-test"
+gradle.taskGraph.whenReady { taskGraph ->
+  taskGraph.getAllTasks().each { task ->
+    def project = task.getProject()
+    task.onlyIf {
+      // Closure, returns false for samza-test(will get filtered) and true for all other modules(will remain unfiltered).
```
shanthoosh: This is not mentioned anywhere in gradle user guide.

Can you please point out how this is gradle convention(I'm okay to change it but after adding space before '('  code looks ugly/weird to me)?

settings.gradle
```diff
@@ -45,3 +45,20 @@ rootProject.children.each {
     it.name = it.name + "_" + scalaVersion
   }
 }
+
+/**
+ * Skips execution of all tasks for the project 'samza-test'.
+ * To include `samza-test`, initiate gradle build by setting the property 'includeSamzaTest'.
+ * To build with samza-test: './gradlew clean build -PincludeSamzaTest'
+ */
+def propertyName = "includeSamzaTest"
```
prateekm: Re 2. Clearly, we disagree, hence my original comment :) First, see comment below about whether we should exclude the whole project or not. Even if we decide to, I think the name is confusing for someone not familiar with the build & project structure. IMHO will be much clearer to indicate its function instead.

settings.gradle
```diff
@@ -45,3 +45,20 @@ rootProject.children.each {
...
+def projectName = "samza-test"
```
prateekm: > There will no benefit in building locally samza-test if it’s not going to be run locally (since it just contains tests, doesn't have src files).

It contains the test source files, which should be compiled as part of the build to detect compilation errors/API changes.

settings.gradle
prateekm: Comment only applies to the second clause in the OR. It's incomplete if the project property is set. Should document both clauses.

settings.gradle
shanthoosh: Okay, makes sense. Now i understand. (my naming was chosen based upon including iit in the build or not. Your suggestion was based upon just compiling and not running it). 

Changing it.

settings.gradle
shanthoosh: Makes sense. Done.

settings.gradle
shanthoosh: I see. Thanks for making it clear.

Updating the comment.

settings.gradle
shanthoosh: Updating the comment.

prateekm:
@shanthoosh any updates on this?

shanthoosh:
@prateekm 
Got caught up with other work(UniqueId for venice, samza-admin fixes). 

Updated the patch now.

Please take a look.

shanthoosh:
@prateekm 
Updated relevant docs.

Please take a look.

PR #240, opened 2017-06-30
SAMZA-1304: Handling duplicate stream processor registration.

shanthoosh:
When a stream processor registers with same processorId as already existing
processor in processor group, it's registration should fail.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ProcessorData.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ProcessorData.java
@@ -19,9 +19,12 @@
 
 package org.apache.samza.zk;
 
+import java.util.Objects;
 import org.apache.samza.SamzaException;
 
-
+/**
+ * Represents processor data stored in zookeeper processors node.
+ */
 public class ProcessorData {
   private final String processorId;
   private final String host;
@@ -50,5 +53,18 @@ public class ProcessorData {
 
   public String getProcessorId() {
     return processorId;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(processorId, host);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (obj == null) return false;
+    if (getClass() != obj.getClass()) return false;
+    final ProcessorData other = (ProcessorData) obj;
+    return processorId.equals(other.processorId) && host.equals(other.host);
   }
 }
--- a/samza-core/src/main/java/org/apache/samza/zk/ProcessorData.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ProcessorData.java
@@ -65,6 +65,6 @@ public class ProcessorData {
     if (obj == null) return false;
     if (getClass() != obj.getClass()) return false;
     final ProcessorData other = (ProcessorData) obj;
-    return processorId.equals(other.processorId) && host.equals(other.host);
+    return Objects.equals(processorId, other.processorId) && Objects.equals(host, other.host);
   }
 }
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
@@ -20,9 +20,10 @@ package org.apache.samza.zk;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
+import java.util.Set;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.samza.config.ApplicationConfig;
 import org.apache.samza.config.Config;
@@ -165,9 +166,7 @@ public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
     // if list of processors is empty - it means we are called from 'onBecomeLeader'
     // TODO: Handle empty currentProcessorIds.
     List<String> currentProcessorIds = getActualProcessorIds(processors);
-    List<String> uniqueProcessorIds = currentProcessorIds.stream()
-                                                         .distinct()
-                                                         .collect(Collectors.toList());
+    Set<String> uniqueProcessorIds = new HashSet<String>(currentProcessorIds);
 
     if (currentProcessorIds.size() != uniqueProcessorIds.size()) {
       LOG.info("Processors: {} has duplicates. Not generating job model.", currentProcessorIds);
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.samza.config.ApplicationConfig;
 import org.apache.samza.config.Config;
@@ -162,8 +163,16 @@ public class ZkJobCoordinator implements JobCoordinator, ZkControllerListener {
 
   void doOnProcessorChange(List<String> processors) {
     // if list of processors is empty - it means we are called from 'onBecomeLeader'
-    // TODO: Handle empty currentProcessorIds or duplicate processorIds in the list
+    // TODO: Handle empty currentProcessorIds.
     List<String> currentProcessorIds = getActualProcessorIds(processors);
+    List<String> uniqueProcessorIds = currentProcessorIds.stream()
+                                                         .distinct()
+                                                         .collect(Collectors.toList());
+
+    if (currentProcessorIds.size() != uniqueProcessorIds.size()) {
+      LOG.info("Processors: {} has duplicates. Not generating job model.", currentProcessorIds);
+      return;
+    }
 
     // Generate the JobModel
     JobModel jobModel = generateNewJobModel(currentProcessorIds);
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
@@ -19,15 +19,18 @@
 
 package org.apache.samza.zk;
 
+import com.google.common.base.Preconditions;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
+import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import org.I0Itec.zkclient.IZkChildListener;
 import org.I0Itec.zkclient.IZkDataListener;
 import org.I0Itec.zkclient.ZkClient;
-import org.I0Itec.zkclient.ZkConnection;
 import org.I0Itec.zkclient.exception.ZkInterruptedException;
 import org.apache.samza.SamzaException;
 import org.apache.samza.job.model.JobModel;
@@ -83,10 +86,6 @@ public class ZkUtils {
     }
   }
 
-  public static ZkConnection createZkConnection(String zkConnectString, int sessionTimeoutMs) {
-    return new ZkConnection(zkConnectString, sessionTimeoutMs);
-  }
-
   ZkClient getZkClient() {
     return zkClient;
   }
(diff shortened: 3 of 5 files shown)


shanthoosh:
@sborya @navina 

samza-core/src/main/java/org/apache/samza/zk/ProcessorData.java
```diff
@@ -51,4 +54,17 @@ public String getHost() {
...
+  public int hashCode() {
+    return Objects.hash(processorId, host);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (obj == null) return false;
+    if (getClass() != obj.getClass()) return false;
+    final ProcessorData other = (ProcessorData) obj;
+    return processorId.equals(other.processorId) && host.equals(other.host);
```
navina: I don't think we should be comparing host for equality. It should only be processors. For example, a case where the duplicate processorId comes on a different host, but it is still a duplicate as far as partitioning/grouping is concerned. Please correct me if I am wrong.

vjagadish1989: Can `processorId` be `null`? Can `host` be null? If so, can handle null checks in here as well.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -162,8 +163,16 @@ public void onProcessorChange(List<String> processors) {
 
   void doOnProcessorChange(List<String> processors) {
     // if list of processors is empty - it means we are called from 'onBecomeLeader'
-    // TODO: Handle empty currentProcessorIds or duplicate processorIds in the list
+    // TODO: Handle empty currentProcessorIds.
     List<String> currentProcessorIds = getActualProcessorIds(processors);
+    List<String> uniqueProcessorIds = currentProcessorIds.stream()
```
vjagadish1989: This appears verbose. Might be simpler to use a `HashSet` here?

```
Set<String> uniqueProcessorIds = new HashSet<>(currentProcessorIds)
```

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -110,17 +109,92 @@ public ZkKeyBuilder getKeyBuilder() {
...
+   * @return the sequential ephemeral path created in zookeeper.
+   */
+  String registerProcessor(final ProcessorData processorData) {
+    // Create ephemeral path in zookeeper corresponding to this processor.
+    ephemeralPath = zkClient.createEphemeralSequential(keyBuilder.getProcessorsPath() + "/", processorData.toString());
+    LOG.info("Created ephemeral path: {} for processor: {} in zookeeper.", ephemeralPath, processorData);
+    ProcessorNode processorNode = new ProcessorNode(processorData, ephemeralPath);
+    String processorId = processorData.getProcessorId();
+    // Determine if there are duplicate processors with this.processorId after registration.
+    if (getProcessorCount(processorId) > 1 && shouldStopProcessor(processorNode)) {
```
vjagadish1989: Do we actually need both these checks here?

 It seems that we can get rid off the `getProcessorCount()` check completely since the `shouldStopProcessor` is a more stronger guarantee?

As an aside, this will save an extra Zk read, in addition to simplifying some tests and removing the `Preconditions.checkState` assertion.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -110,17 +109,92 @@ public ZkKeyBuilder getKeyBuilder() {
...
+   * Fetches all the ephemeral processor nodes of a standalone job from zookeeper.
+   * @return a list of {@link ProcessorNode}, where each ProcessorNode represents a registered stream processor.
+   */
+  private List<ProcessorNode> getAllProcessorNodes() {
+    List<String> processorZNodes = getSortedActiveProcessorsZnodes();
+    LOG.debug("Active ProcessorZNodes in zookeeper: {}.", processorZNodes);
+    return processorZNodes.stream()
+                          .map(processorZNode -> {
+                              String ephemeralProcessorPath = String.format("%s/%s", keyBuilder.getProcessorsPath(), processorZNode);
+                              String data = zkClient.readData(ephemeralProcessorPath);
```
navina: Is it possible to use `ZkUtils#readProcessorData` instead of using zkclient directly ?
Metrics is lost here since we are not using wrapper methods for zkClient. There are other places in this PR where we are doing it this way. It may also be the same in the existing code. I think we should create wrapper methods for every zkclient api that we use for the sake of metrics.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -324,4 +398,45 @@ public void subscribeToProcessorChange(IZkChildListener listener) {
     zkClient.subscribeChildChanges(keyBuilder.getProcessorsPath(), listener);
     metrics.subscriptions.inc();
   }
+
+  /**
+   * Represents zookeeper processor node.
+   */
+  private static class ProcessorNode {
```
navina: Can you explain the motivation for introducing `ProcessorNode` here? Is it possible to achieve the same by adding the ephemeralProcessorPath to `ProcessorData` ?

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
```diff
@@ -174,6 +182,47 @@ public void handleDataDeleted(String dataPath)
...
+    ProcessorData processorData1 = new ProcessorData(testHostName, testProcessId);
+    // Register processor 1 which is not duplicate, this registration should succeed.
+    zkUtils.registerProcessorAndGetId(processorData1);
+
+    ZkUtils zkUtils1 = getZkUtils();
+    zkUtils1.connect();
+    ProcessorData duplicateProcessorData = new ProcessorData(testHostName, testProcessId);
+    // Registration of the duplicate processor should fail.
+    // Expect the following block to fail with IllegalStateException.
+    expectedException.expect(IllegalStateException.class);
```
vjagadish1989: I did not know about this `ExpectedException` idiom :) useful!

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
navina: Do we really need a test here for `shouldStopProcessor`?  If I understand correctly, this instance will read the list and look for duplicates only *after* registration. So, if the count is greater than 1, won't it always have to delete its own registered znode? 

I may be missing something. Can you please verify / explain?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -110,17 +109,92 @@ public ZkKeyBuilder getKeyBuilder() {
...
+                                                         .collect(Collectors.toCollection(TreeSet::new));
+    return !sortedProcessorPaths.first().equals(processor.getEphemeralPath());
+  }
+
+  /**
+   * Find the number of processors having the provided processorId.
+   * @param processorId to use for equality check.
+   * @return the count of matched processors.
+   */
+  private long getProcessorCount(final String processorId) {
```
navina: Can you modify this method to return the list of duplicate processors, if any, instead of the count? 
In Line 140, you are reading invoking `getAllProcessorNodes` twice - once in getProcessorsCount and shouldStopProcessor. Each call to `getAllProcessorNodes` will read the entire list and its data. It seems inefficient and also, potentially, you can get different lists.

samza-core/src/main/java/org/apache/samza/zk/ProcessorData.java
shanthoosh: This is just honoring `Object.equals` contract. 

The way in which we want to determine equality(without host affinity) now shouldn't affect ideal implementation. 

For determining duplicate processors, equality check is on processorID alone(Lines 165, 179 in ZkUtils).

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
shanthoosh: When multiple clients(`C1, C2... Cn`) are trying to register with same processorId (`P1`) at same time, `shouldStopProcessor` picks a client to live and stops rest of them(instead of stoping all).

Current implementation will pick client with lowest `zkProcessorPath` to live and stop others.

This behavior is verified in `TestZkUtils` integration test.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
shanthoosh: Sure, changing it.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
shanthoosh: Reasons: 

* This is a `DAO` and mimics the exact way how processors and it's data are organized in zookeeper. There's a proper logical mapping between this to how it exists in zookeeper. I think ephemeralPath doesn't belong to ProcessorData. In future when the codebase evolves, will help readability.

* ProcessorNode is a place for storing creationTime, numberOfChildren and other statistics along with ACL related info in future.

Makes sense? Please share your thoughts.

samza-core/src/main/java/org/apache/samza/zk/ProcessorData.java
shanthoosh: Fixed. Switched to `java.util.Objects`

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
shanthoosh: Makes sense. 

Initial idea was to generate jobmodel with unique processorIds(not swallow & return). Since it was a complicated change, it's punted for this release based upon discussion with boris. This was reminiscence of that.
 
Fixed.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
shanthoosh: Rational is to make registration inexpensive and fail fast in error scenarios. 

Preconditions check was added for that, registering a duplicate processor and having duplicate check later would have triggered processor change in leader, which could potentially take the processor group to rebalancing/barrier phase depending upon timing of events. 

Cost is extra two zookeeper `ls` on `/rootapp/processors` directory and benefit is avoiding reshuffle in worst case(just for duplicate processor addition). 

Since Zookeeper is highly performant for read heavy workloads, this wouldn't be a big cost.

Makes sense?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
shanthoosh: Sure, that is a good optimization avoiding redundant reads.

Changing it.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
navina: I think it is nice to use Data Access Object (DAO) pattern here. However, it still doesn't explain why it should be a completely separate class. I think the intention behind `ProcessorData` was similar.

>  I think ephemeralPath doesn't belong to ProcessorData.

I am not following why ephemeralPath cannot belong to ProcessorData. Can you please explain your thoughts?

> In future when the codebase evolves, will help readability.

Code readability is undeniably important. But it is not helping with readability right now, at least for me. 

> ProcessorNode is a place for storing creationTime, numberOfChildren and other statistics along with ACL related info in future

It's nice to plan for future. These will be good stats and data to have in the local object state. It still doesn't explain why we need 2 different classes. Why can't be just maintain one class `ProcessorNode` or `ProcessorData`?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
vjagadish1989: @shanthoosh My observation was that the one of the checks is redundant. (since it is a strict super-set of the other). I also don't think we are failing-fast (since, we'll have to do a read eitherways).

The first `processorId` to register succeeds, and the 2nd registration with the same `processorId` fails. All you need is one consistent Znode read on the list-of-processors to do this.

navina: > Current implementation will pick client with lowest zkProcessorPath to live and stop others.

How is the processor "stopping" the rest of them? 

Iiuc, every processor is doing this check. Since every client in C1, C2, ... Cn is going to have a consistent view of the entire list (as guaranteed by ZK), the client itself knows whether it is the lowest `zkProcessorPath` or not. Shouldn't that be sufficient for the client to delete and disconnect?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
shanthoosh: Discussed offline. 
`ProcessorData` represents actual data(processorId, hostname) stored in zkNode. 
`ProcessorNode` holds auxiliary details about zkNode(currently processorpath, in future : metadata, parentNode, childNodes, ACL) along with `ProcessorData` .

shanthoosh: Discussed offline. 
ProcessorData represents actual data(processorId, hostname) stored in zkNode. ProcessorNode holds auxiliary details about zkNode(currently processorpath, in future : metadata, parentNode, childNodes, ACL) along with ProcessorData.

Two different logical things, hence they’re separate.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -110,17 +108,73 @@ public ZkKeyBuilder getKeyBuilder() {
...
+  }
+
+  /**
+   * Determines if a processor should be stopped.
+   * @param processor to check for stop condition in processors group.
+   * @return true if processor should be stopped. false otherwise.
+   */
+  private boolean shouldStopProcessor(final ProcessorNode processor) {
+    /**
+     * If there are multiple processors registered with same processorId,
```
navina: can you please move this documentation detail above the function ?

nit: can you rename this method `isProcessorDuplicate` instead of `shouldStopProcessor`? Functionality is more intuitive when named so.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -110,17 +108,73 @@ public ZkKeyBuilder getKeyBuilder() {
...
     } else {
-      LOG.info("existing path for " + data +  " is " +  ephemeralPath);
-      return ephemeralPath;
+      LOG.info("Ephemeral path: {} exists for processor: {} in zookeeper.", ephemeralPath, data);
     }
+    return ephemeralPath;
+  }
+
+  /**
+   * Determines if a processor should be stopped.
```
navina: The method actually determines if the given processorNode is a duplicate of any of the existing nodes. The action we take on duplicate is stopping the processors. Can you please update the doc?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -110,17 +108,73 @@ public ZkKeyBuilder getKeyBuilder() {
...
   public synchronized String registerProcessorAndGetId(final ProcessorData data) {
+    String processorId = data.getProcessorId();
     if (ephemeralPath == null) {
-      ephemeralPath =
-          zkClient.createEphemeralSequential(
-              keyBuilder.getProcessorsPath() + "/", data.toString());
-
-      LOG.info("newly generated path for " + data +  " is " +  ephemeralPath);
-      return ephemeralPath;
+      /**
```
vjagadish1989: 1.Comments below are unnecessary.
2.Consider adding an `@throws` annotation to this method's javadoc instead?

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
```diff
@@ -174,6 +182,27 @@ public void handleDataDeleted(String dataPath)
     Assert.assertTrue(testWithDelayBackOff(() -> "newProcessor".equals(res.getRes()), 2, 1000));
   }
 
+  /**
+   * Create two duplicate processors with same processorId.
+   * Second creation should fail with exception.
+   */
+  @Test
+  public void testRegisterProcessorAndGetIdShouldFailForDuplicateProcessorRegistration() {
```
vjagadish1989: s/testRegisterProcessorAndGetIdShouldFailForDuplicateProcessorRegistration/testDuplicateProcessorRegistrationShouldFail maybe?

samza-test/src/test/java/org/apache/samza/test/processor/TestZkLocalApplicationRunner.java
```diff
@@ -256,10 +258,9 @@ public void shouldFailWhenNewProcessorJoinsWithSameIdAsExistingProcessor() throw
     publishKafkaEvents(inputKafkaTopic, NUM_KAFKA_EVENTS, PROCESSOR_IDS[2]);
     kafkaEventsConsumedLatch = new CountDownLatch(NUM_KAFKA_EVENTS);
     StreamApplication streamApp3 = new TestStreamApplication(inputKafkaTopic, outputKafkaTopic, null, null, kafkaEventsConsumedLatch);
+    // Fail when the duplicate processor joins.
+    expectedException.expect(Exception.class);
```
vjagadish1989: Should we expect a more specific exception here?

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
```diff
@@ -174,6 +182,27 @@ public void handleDataDeleted(String dataPath)
...
+    final String testProcessId = "testProcessorId";
+    ProcessorData processorData1 = new ProcessorData(testHostName, testProcessId);
+    // Register processor 1 which is not duplicate, this registration should succeed.
+    zkUtils.registerProcessorAndGetId(processorData1);
+
+    ZkUtils zkUtils1 = getZkUtils();
+    zkUtils1.connect();
+    ProcessorData duplicateProcessorData = new ProcessorData(testHostName, testProcessId);
+    // Registration of the duplicate processor should fail.
+    // Expect the following block to fail with IllegalStateException.
```
vjagadish1989: Can remove L:201

vjagadish1989 (approved):
looks good. thanks! 

approved

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
shanthoosh: @navina

When there’re multiple duplicate processors(P1,P2,P3..Pn) this marks processor with lowest sequential id as not duplicate, rest as duplicate.

I don’t think this  is synonymous to `Collection.contains` to rename(which should be symmetric for all processors). If we rename that way, it also conveys wrong meaning that we kill all duplicate processors (P1, P2, .. Pn). 

Naming to  `isValidRegisteredProcessor` if it's okay

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
shanthoosh: Moving java doc above the function.

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
shanthoosh: Done

samza-test/src/test/java/org/apache/samza/test/processor/TestZkLocalApplicationRunner.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
shanthoosh: No value addition in documenting, since SamzaException is unchecked and stream processor should die(Not a recoverable exception).

New checked exception isn't added for consistency( majority of codebase uses this for custom exception).

vjagadish1989: Makes sense. Thank you for checking and removing the unnecessary comments!

PR #241, opened 2017-06-30
SAMZA-1356: Improve monitoring for state restore

jmakes:


vjagadish1989 (approved):
looks good. approved. thank you for the contribution

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngineMetrics.scala
```diff
@@ -34,8 +34,8 @@ class KeyValueStorageEngineMetrics(
   val deletes = newCounter("deletes")
   val flushes = newCounter("flushes")
 
-  val restoredMessages = newCounter("messages-restored")
-  val restoredBytes = newCounter("messages-bytes")
+  val restoredMessages = newGauge("messages-restored", 0)
```
shanthoosh: Sorry to peak my nose into this. I joined offline discussion late, so I don't have complete context. 

Can you please explain why are we changing from counter to gauge?

jmakes: There should be no offline context needed. Please review the JIRA and if the reasoning is still unclear, let me know and I will update it.  Thanks!

shanthoosh: Counters are not `reported` as deltas from samza  [`JmxReporter`, `MetricsSnapshotReporter`, `AmfReporter`]. I think problem is LI metrics renderer engine `plots` the delta from the counter(diff between two different reports). It wasn't obvious from JIRA and code(replacing counter functionality with gauge). It's clear now. Thanks.

jmakes: Fair point. I will update the JIRA to make it clearer.

Thanks

jmakes:
@cpettitt-linkedin I was doing manual test, which is why this PR is still open. Couldn't test it in an LI cluster during moratorium.

I thought about adding a unit test to verify the type of the metrics, but it didn't seem to add much value.

cpettitt-linkedin:
Why not start up a test store and verify metrics? I assume we're not doing that if this change did not break any tests.

PR #242, opened 2017-07-11
SAMZA-1358: fix the bug in validating task.class empty string when ap…

nickpan47:
…p.class is configured

Another bug due to scala/java differences.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
+++ b/samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
@@ -32,6 +32,8 @@ import java.util.concurrent.ExecutorService;
 
 import scala.runtime.AbstractFunction0;
 
+import static org.apache.samza.util.ScalaToJavaUtils.defaultValue;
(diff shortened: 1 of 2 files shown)


samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
```diff
@@ -162,7 +164,8 @@ public static StreamApplication createStreamApplication(Config config) {
     ApplicationConfig appConfig = new ApplicationConfig(config);
     if (appConfig.getAppClass() != null && !appConfig.getAppClass().isEmpty()) {
       TaskConfig taskConfig = new TaskConfig(config);
-      if (taskConfig.getTaskClass() != null && !taskConfig.getTaskClass().isEmpty()) {
+      String taskClassName = taskConfig.getTaskClass().getOrElse(defaultValue(null));
```
bharathkk: np: Given we are anyways converting it to null, how about using `taskConfig.getTaskClass().orNull`

PR #243, opened 2017-07-14
SAMZA-1359: Handle phantom container notifications cleanly during an RM fail-over

vjagadish1989:
1. Improved our container handling logic to be resilient to phantom notifications.
2. Added a new metric to Samza's ContainerProcessManager module that tracks the number of such invalid notifications.
3. Add a couple of tests that simulate this exact scenario above that we encountered during the cluster upgrade. (container starts -> container fails -> legitimate notification for the failure - container re-start -> RM fail-over -> phantom notification with a different exit code)
4. As an aside, there are a whole bunch of tests in ContainerProcessManager that rely on Thread.sleep to ensure that threads get to run in a certain order. Removed this non-determinism and made them predictable.


samza-core/src/main/scala/org/apache/samza/metrics/ContainerProcessManagerMetrics.scala
```diff
@@ -52,6 +52,7 @@ class ContainerProcessManagerMetrics(
     val mFailedContainers = newGauge("failed-containers", () => state.failedContainers.get())
     val mReleasedContainers = newGauge("released-containers", () => state.releasedContainers.get())
     val mContainers = newGauge("container-count", () => state.containerCount)
+    val mInvalidNotifications = newGauge("invalid-notifications", () => state.invalidNotifications.get())
```
jmakes: From what I can tell, these notifications are not "invalid". They are "duplicate", "redundant" or "ignored", but not "invalid" 

I'd suggest a rename, but it's your call.

samza-core/src/test/java/org/apache/samza/clustermanager/MockContainerAllocator.java
```diff
@@ -23,16 +23,29 @@
...
   }
 
+  /**
+   * Causes the current thread to block until the provided number of containers have started.
+   *
+   * @param numExpectedContainers  the number of containers expected to start
+   * @throws InterruptedException  if the current thread is interrupted while waiting
+   */
+  void awaitContainersStart(int numExpectedContainers) throws InterruptedException {
+    latch = new CountDownLatch(numExpectedContainers);
```
jmakes: How does the code guarantee that this latch reassignment doesn't happen AFTER the latch.countDown in runStreamProcessor()?

samza-core/src/main/scala/org/apache/samza/metrics/ContainerProcessManagerMetrics.scala
vjagadish1989: Makes sense, I'll rename them to be duplicate.

samza-core/src/test/java/org/apache/samza/clustermanager/MockContainerAllocator.java
vjagadish1989: 1. `awaitContainersStart` is not meant to be used in a multi-threaded context. 

2. The only guarantee is that, we will wait until 2 containers have started after the call to `awaitContainersStart`. The caller should ensure that `awaitContainerStart` is called before any action that triggers the startup. This is just an alternative to doing a `Thread.sleep()`

jmakes: Certainly better than Thread.sleep(). Thanks for explaining and +10000 to getting rid of Thread.sleep(). Appreciate it!

cpettitt-linkedin: I would suggest against an unbounded await. How about some reasonable timeout and then fail the test?

Instead of reassigning the count down latch you could use a semaphore and acquire the expected number of permits here and release them as containers start up.

vjagadish1989:
@santhoshvenkat1988 @jmakes Review please!

PR #245, opened 2017-07-20
SAMZA-1366: ScriptRunner should allow callers to control the child pr…

jmakes:
…ocess environment.

### DIFF
--- a/samza-rest/src/main/java/org/apache/samza/rest/script/ScriptRunner.java
+++ b/samza-rest/src/main/java/org/apache/samza/rest/script/ScriptRunner.java
@@ -43,17 +43,7 @@ public class ScriptRunner {
   private static final int DEFAULT_SCRIPT_CMD_TIMEOUT_S = 30;
   private int scriptTimeout = DEFAULT_SCRIPT_CMD_TIMEOUT_S;
 
-  // Dont pass down the current environment vars by default
-  private boolean forwardEnvironmentVars = false;
   private final Map<String, String> environment = new HashMap<>();
-
-  public ScriptRunner() {
-
-  }
-
-  public ScriptRunner(boolean forwardEnvironmentVars) {
-    this.forwardEnvironmentVars = forwardEnvironmentVars;
-  }
 
   protected long getScriptTimeoutS() {
     return scriptTimeout;
@@ -112,10 +102,7 @@ public class ScriptRunner {
     log.debug("Building process with command {}", command);
     ProcessBuilder pb =  new ProcessBuilder(command);
 
-    if (!forwardEnvironmentVars) {
-      pb.environment().clear();
-    }
-
+    pb.environment().clear();
     pb.environment().putAll(environment);
     return pb;
   }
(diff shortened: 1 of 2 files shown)


jmakes:
@santhoshvenkat1988 @vjagadish1989  take a look?

samza-rest/src/main/java/org/apache/samza/rest/script/ScriptRunner.java
```diff
@@ -126,4 +153,34 @@ public void run() {
...
+    return forwardEnvironmentVars;
+  }
+
+  /**
+   * Set the flag indicating whether this runner will forward the current environment variables to the child process.
+   *
+   * @param forwardEnvironmentVars  true to forward the current environment to the child process,
+   *                                false to start with an empty environment.
+   */
+  public void setForwardEnvironmentVars(boolean forwardEnvironmentVars) {
```
vjagadish1989: A couple of alternative approaches 

- Do we need this setter? Can get rid off this to ensure that the `forwardEnvVars` is initialized during construction.

- Alternately, it feels natural to add `forwardEnvVars` to the `runScript` API since it's designed to run a script with IO inherited from the current Java process anyway.

vjagadish1989:
@jmakes : LGTM, just a question, and it's shippable!

samza-rest/src/main/java/org/apache/samza/rest/script/ScriptRunner.java
jmakes: Passing it at construction isn't doable without significant work, because the ScriptJob proxy reuses the same runner over and over. 

The second approach sounds promising. It gave me another idea. 
I expect it to be rare that we would want to forward the env vars. Maybe I'll just clear them always and then the caller can use .environment() to get the map and copy the current env into it, if they really want to forward. 

Let me try that and post another diff. 

Thanks for the suggestion!

samza-rest/src/main/java/org/apache/samza/rest/script/ScriptRunner.java
```diff
@@ -84,13 +90,21 @@ public int runScript(String scriptPath, ScriptOutputHandler outputHandler, Strin
...
     List<String> command = new ArrayList<>(args.length + 1);
     command.add(scriptPath);
     command.addAll(Arrays.asList(args));
 
     log.debug("Building process with command {}", command);
-    return new ProcessBuilder(command);
+    ProcessBuilder pb =  new ProcessBuilder(command);
+
+    pb.environment().clear();
+    pb.environment().putAll(environment);
```
shanthoosh: Minor suggestion(Please change if it makes sense): 

Just wondering, if it's okay to take in environmentMap as runScript param(just like commandArgs or even as constructor param) and use it for processcreation(if we want to support client of this api running script with custom env).

I think how setting environment  will affect process creation will be simple that way. Otherwise  a client wanting to modify env will have do getMap() and modify it and then call runScript.  

Or let's disallow clients to even modify envMap(hiding the get method).

vjagadish1989 (approved):
looks great. approved.
thank you for the contribution

PR #246, opened 2017-07-20
Samza-1364: Handle ZKExceptions in zkCoordinationUtils.reset.

shanthoosh:
In some cases LocalAppRunner.waitForFinish indefinitely blocks after LocalApplicationRunner.kill. Last step in LocalAppRunner.kill(streamApp) is zkClient.close()[zkClient belongs to ZkCoordinationService].

ApplicationRunner.kill triggers listeners chain and in final listener zkClient.close throws ZkInterruptedException(RuntimeException) & it's swallowed in listeners preventing shutdownLatch update in LocalApplicationRunner(required for proper shutdown).

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationUtils.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationUtils.java
@@ -18,13 +18,18 @@
  */
 package org.apache.samza.zk;
 
+import org.I0Itec.zkclient.exception.ZkInterruptedException;
 import org.apache.samza.config.ZkConfig;
 import org.apache.samza.coordinator.CoordinationUtils;
 import org.apache.samza.coordinator.Latch;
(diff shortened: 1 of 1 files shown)


samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationUtils.java
```diff
@@ -37,7 +42,12 @@ public ZkCoordinationUtils(String processorId, ZkConfig zkConfig, ZkUtils zkUtil
 
   @Override
   public void reset() {
-    zkUtils.close();
+    try {
+      zkUtils.close();
+    } catch (ZkInterruptedException ex) {
+      // Swallowing due to occurrence in last stage of lifecycle(Not actionable).
```
sborya: nit. s/last/the last/

shanthoosh: Fixed.

PR #247, opened 2017-07-20
SAMZA-1368: make sure new job model will be generated in case of barrier timeout.

sborya:


samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -310,11 +311,17 @@ public void onBarrierStateChanged(final String version, ZkBarrierForVersionUpgra
...
-          // In our consensus model, if the Barrier is timed-out, then it means that one or more initial
-          // participants failed to join. That means, they should have de-registered from "processors" list
-          // and that would have triggered onProcessorChange action -> a new round of consensus.
-          LOG.info("Barrier for version " + version + " timed out.");
+          // no-op for non-leaders
+          // for leader: make sure we do not stop - so generate a new job model
+          LOG.warn("Leader: barrier for version " + version + " timed out.");
+          if (leaderElector.amILeader()) {
+            debounceTimer.scheduleAfterDebounceTime(ScheduleAfterDebounceTime.ON_PROCESSOR_CHANGE, debounceTimeMs, () ->
+              {
```
shanthoosh: This guards against no processors change(for some reason processor didn't remove itself) on barrier timeout. 

On contrary, we could have redundant rebalancing phases (one at barrier timeout, one when processor gets deleted. We can't handle this cleanly until there's pause functionality in SamzaContainer). Let's create a JIRA for this.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -310,11 +310,17 @@ public void onBarrierStateChanged(final String version, ZkBarrierForVersionUpgra
...
       } else {
         if (ZkBarrierForVersionUpgrade.State.TIMED_OUT.equals(state)) {
-          // no-op
-          // In our consensus model, if the Barrier is timed-out, then it means that one or more initial
-          // participants failed to join. That means, they should have de-registered from "processors" list
-          // and that would have triggered onProcessorChange action -> a new round of consensus.
-          LOG.info("Barrier for version " + version + " timed out.");
+          // no-op for non-leaders
+          // for leader: make sure we do not stop - so generate a new job model
+          LOG.warn("Leader: barrier for version " + version + " timed out.");
```
shanthoosh: Minor suggestion(log message format): 

s/Leader: barrier for version /Barrier for version. 

At this point, we're not sure if it's leader or not.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -310,11 +310,17 @@ public void onBarrierStateChanged(final String version, ZkBarrierForVersionUpgra
...
+          if (zkController.isLeader()) {
```
shanthoosh: After the if, let's log that we're scheduling `PROCESSOR_CHANGE` debounce (on barrier timeout from `leader`).

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java

PR #248, opened 2017-07-24
SAMZA-1361: OperatorImplGraph is using wrong keys to store/retrieve OperatorImpl in the map

prateekm:


prateekm:
@nickpan47 @vjagadish1989 Please take a look.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -170,13 +170,13 @@ public ContextManager getContextManager() {
...
    *
-   * @return  a set of all available {@link OperatorSpec}s
+   * @return  all available {@link OperatorSpec}s
    */
   public Collection<OperatorSpec> getAllOperatorSpecs() {
     Collection<InputOperatorSpec> inputOperatorSpecs = inputOperators.values();
     Set<OperatorSpec> operatorSpecs = new HashSet<>();
-
     for (InputOperatorSpec inputOperatorSpec: inputOperatorSpecs) {
+      operatorSpecs.add(inputOperatorSpec);
```
vjagadish1989: thanks!

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
```diff
@@ -129,7 +129,7 @@ public void close() {
    */
   OperatorImpl createAndRegisterOperatorImpl(OperatorSpec prevOperatorSpec, OperatorSpec operatorSpec,
       Config config, TaskContext context) {
-    if (!operatorImpls.containsKey(operatorSpec) || operatorSpec instanceof JoinOperatorSpec) {
+    if (!operatorImpls.containsKey(operatorSpec.getOpName()) || operatorSpec instanceof JoinOperatorSpec) {
```
nickpan47: This relies on the guarantee that getOpName() returns the unique ID for an operatorSpec in a user defined operator graph. Instead, using the operatorSpec object reference directly as key does not require that. Curious about what's the advantage here using String as the key?

prateekm: @nickpan47 There are 2 PartialJoinOperatorImpl instances corresponding to the same JoinOperatorSpec, which we need to distinguish between. Hence using their names (E.g. join-2-L and join-2-R) as the map keys.

nickpan47: Got it. As discussed offline, please make the requirements of the getOpName() return value from OperatorSpec clear s.t. implementation is required to return a unique ID string in a user defined DAG.

prateekm: @nickpan47 Updated documentation for that method, please take a look.

nickpan47:
Note: there are checkstyleTest failures in the Jenkins build. Please fix those as well.

Thanks!

prateekm:
@nickpan47 checkstyle failures were from an unrelated commit, should be fixed now. Will push a dummy commit to verify.

PR #249, opened 2017-07-25
Fix log messages from StreamProcessor(onJobModelExpired event).

shanthoosh:
Log messages published in onJobModelExpired event have `processorId` as null. `processorId` is cached as final var in jobCoordinatorListener method. JLS for final fields/variables states that they're initialized before the constructor. This sets local final variable copy as null(since it relies upon value of instance variable to be set in constructor).
Changes
* Use processorId directly in `createCoordinatorListener` method.
* Remove StreamProcessor.toString since it has no usages.


shanthoosh:
@navina Please take a look when you get a chance. 

Before fix:

```
INFO org.apache.samza.processor.StreamProcessor - Shutting down container done for pid=Processor:null; complete =false
```

After fix:
```
INFO org.apache.samza.processor.StreamProcessor - Shutting down container done for pid=0000000001; complete =true
```

PR #253, opened 2017-07-27
SAMZA-1365: Calling zkClient.close from zkWatch impl blocks indefinitely.

shanthoosh:


shanthoosh:
@navina 

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -333,6 +333,8 @@ public void onBarrierError(String version, Throwable t) {
   /// listener to handle session expiration
   class ZkSessionStateChangedListener implements IZkStateListener {
 
+    private static final String ZK_SESSION_ERROR = "ZK_SESSION_ERROR";
```
navina: nit: Can we move this definition to ScheduleAfterDebounceTime?

PR #255, opened 2017-08-02
SAMZA-1382: added Zk communication protocol version verification

sborya:


samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -438,11 +442,48 @@ public void publishJobModelVersion(String oldVersion, String newVersion) {
...
+      throw new SamzaException("ZK Protocol mismatch. Expected " + ZK_PROTOCOL_VERSION + "; found " + version);
+    }
+  }
+
   /**
    * verify that given paths exist in ZK
    * @param paths - paths to verify or create
    */
-  public void makeSurePersistentPathsExists(String[] paths) {
+  public void validatePaths(String[] paths) {
```
fredji97: It is minor, but to me this new name is not very accurate. I would call it assurePathExists.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -438,11 +442,48 @@ public void publishJobModelVersion(String oldVersion, String newVersion) {
         "(actual data version after update = " + stat.getVersion() + ")");
   }
 
+  // validate Zk protocol currently used is the same as in this participant
+  private void validateZkVersion() {
+    String rootPath = keyBuilder.getRootPath();
+    if (!zkClient.exists(rootPath)) {
```
fredji97: I would recommend moving Line 447-458 to a separate function called validateRootPath since this is not for version validation. I think it is better to have single responsibility for each function.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -438,11 +442,48 @@ public void publishJobModelVersion(String oldVersion, String newVersion) {
...
+
   /**
    * verify that given paths exist in ZK
    * @param paths - paths to verify or create
    */
-  public void makeSurePersistentPathsExists(String[] paths) {
+  public void validatePaths(String[] paths) {
+
+    // validate the version of ZK protocol
+    validateZkVersion();
```
fredji97: I would recommend moving this line out. In the outside, we make three calls
validateRootPath();
validateZkVersion();
assurePathExists() or validatePath();

We may not prefer to embed one inside another.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: actually this check is for version validation. It is just that there could be two different scenarios:
1. RootPath already exists - then we need to read and validate the version.
2. RootPath doesn't exist - then we need to create one and write the version into it.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: Done. Using validateZkVersion separately.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
sborya: I think they are mostly the same.

vjagadish1989 (approved):
Looks good to me. approved. Ship it!!

PR #254, opened 2017-08-02
Samza-1379: Create Azure Client

PawasChhokra:
@navina 
**PR 1: AzureClient + AzureConfig** (current PR)

PR #256, opened 2017-08-02
SAMZA-1376: Create a leasing utility class for blobs in Azure

PawasChhokra:
@navina 
**PR 2: LeaseBlobManager** (current PR)


samza-azure/src/main/java/org/apache/samza/AzureConfig.java
```diff
@@ -0,0 +1,72 @@
...
+    ApplicationConfig appConfig = new ApplicationConfig(config);
+    //Remove all non-alphanumeric characters from id as table name does not allow them.
+    String id = appConfig.getGlobalAppId().replaceAll("[^A-Za-z0-9]", "");
+    containerName = "samzacontainer" + id;
+    blobName = "samzablob" + id;
+    tableName = "samzatable" + id;
+  }
+
+  public String getAzureConnect() {
+    if (!containsKey(AZURE_STORAGE_CONNECT)) {
```
bharathkk: It is usually good practice to throw errors as early as possible.
How about we do this within constructor?

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
```diff
@@ -0,0 +1,95 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Helper class for lease blob operations.
+ */
+public class LeaseBlobManager {
+
+  public static final Logger LOG = LoggerFactory.getLogger(LeaseBlobManager.class);
+  private CloudPageBlob leaseBlob;
```
bharathkk: np: mark it final as well

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
```diff
@@ -0,0 +1,95 @@
...
+
+  public LeaseBlobManager(CloudPageBlob leaseBlob) {
+    this.leaseBlob = leaseBlob;
+  }
+
+  /**
+   * Acquires a lease on a blob.
+   * If the blob does not exist, throws an exception.
```
bharathkk: What exception gets thrown? It looks from the code, you catch the `StorageException` and log the error based on the `HttpStatus`.
Did you intend to rethrow `SamzaException` within the else block?

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
```diff
@@ -0,0 +1,95 @@
...
+   * @param leaseTimeInSec The time in seconds you want to acquire the lease for.
+   * @param leaseId Proposed ID you want to acquire the lease with, null if not proposed.
+   * @return String that represents lease ID. Null if lease is not acquired.
+   */
+  public String acquireLease(int leaseTimeInSec, String leaseId) {
```
bharathkk: I would suggest you to return `Optional<String>` instead which will force to clients program defensively.

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
```diff
@@ -0,0 +1,95 @@
...
+   * Renews the lease on the blob.
+   * @param leaseId ID of the lease to be renewed.
+   * @return True if lease was renewed successfully, false otherwise.
+   */
+  public boolean renewLease(String leaseId) {
+    try {
+      leaseBlob.renewLease(AccessCondition.generateLeaseCondition(leaseId));
+      return true;
+    } catch (StorageException e) {
+      LOG.error("Wasn't able to renew lease.", new SamzaException(e));
```
bharathkk: Why are we wrapping the exception here?

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
```diff
@@ -0,0 +1,95 @@
...
+   * Releases the lease on the blob.
+   * @param leaseId ID of the lease to be released.
+   * @return True if released successfully, false otherwise.
+   */
+  public boolean releaseLease(String leaseId) {
+    try {
+      leaseBlob.releaseLease(AccessCondition.generateLeaseCondition(leaseId));
+      return true;
+    } catch (StorageException e) {
+      LOG.error("Wasn't able to release lease.", new SamzaException(e));
```
bharathkk: same as above. why wrap the exception for logging?

samza-azure/src/main/java/org/apache/samza/AzureClient.java
```diff
@@ -0,0 +1,64 @@
...
+  AzureClient(String storageConnectionString) {
+    try {
+      account = CloudStorageAccount.parse(storageConnectionString);
+      blobClient = account.createCloudBlobClient();
+      tableClient = account.createCloudTableClient();
+    } catch (IllegalArgumentException | URISyntaxException e) {
+      LOG.error("\nConnection string {} specifies an invalid URI.", storageConnectionString);
+      LOG.error("Please confirm the connection string is in the Azure connection string format.");
+      throw new SamzaException(e);
+    } catch (InvalidKeyException e) {
```
bharathkk: np: Not super picky. Just so you know, the same `catch` block can be used and upi check for `Exception` instance and log the error.

samza-azure/src/main/java/org/apache/samza/AzureConfig.java
```diff
@@ -0,0 +1,72 @@
...
+
+package org.apache.samza;
+
+import org.apache.samza.config.ApplicationConfig;
+import org.apache.samza.config.Config;
+import org.apache.samza.config.ConfigException;
+import org.apache.samza.config.MapConfig;
+
+
+public class AzureConfig extends MapConfig {
```
bharathkk: Can you add class level javadocs?

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
PawasChhokra: Yes, I intended to throw the exception in the else block. Fixed this.

samza-azure/src/main/java/org/apache/samza/AzureConfig.java
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
PawasChhokra: The lease ID for our case is not user defined. It is provided by Azure leasing operation when an entity acquires the lease. It is null initially for all processors.

navina:
@PawasChhokra Can you rebase with upstream ? I closed your previous PR #254

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
PawasChhokra: Fixed. It should be thrown as a storage exception.

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
PawasChhokra: Same as above.

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
```diff
@@ -0,0 +1,95 @@
...
+  public LeaseBlobManager(CloudPageBlob leaseBlob) {
+    this.leaseBlob = leaseBlob;
+  }
+
+  /**
+   * Acquires a lease on a blob.
+   * If the blob does not exist, throws an exception.
+   * @param leaseTimeInSec The time in seconds you want to acquire the lease for.
+   * @param leaseId Proposed ID you want to acquire the lease with, null if not proposed.
+   * @return String that represents lease ID. Null initially.
```
navina: It doesn't look like you are throwing any exceptions. So, "null sugnifies both first attempt to acquire lease and also, failure to acquire the lease?

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
```diff
@@ -0,0 +1,95 @@
...
+   * Renews the lease on the blob.
+   * @param leaseId ID of the lease to be renewed.
+   * @return True if lease was renewed successfully, false otherwise.
+   */
+  public boolean renewLease(String leaseId) {
+    try {
+      leaseBlob.renewLease(AccessCondition.generateLeaseCondition(leaseId));
+      return true;
+    } catch (StorageException storageException) {
+      LOG.error("Wasn't able to renew lease.", storageException);
```
navina: You should include the "leaseId" in the log messages so that it provides more context.

samza-azure/src/test/java/org/apache/samza/MyTestExample.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.samza.operators.MessageStream;
+import org.apache.samza.operators.StreamGraph;
+import org.apache.samza.operators.functions.SinkFunction;
+import org.apache.samza.runtime.LocalApplicationRunner;
+import org.apache.samza.task.MessageCollector;
+import org.apache.samza.task.TaskCoordinator;
+import org.apache.samza.util.CommandLine;
+
+
+public class MyTestExample implements StreamApplication {
```
navina: Is this Test class relevant for the leasing utility class? If not, let's add this with a later patch?

samza-azure/src/main/java/org/apache/samza/AzureClient.java
```diff
@@ -38,19 +38,19 @@
   private final CloudTableClient tableClient;
   private final CloudBlobClient blobClient;
 
-  AzureClient(String storageConnectionString) {
+  AzureClient(String storageConnectionString) throws URISyntaxException, InvalidKeyException {
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
PawasChhokra: I meant to throw an exception in the else part. Fixed that.
And yes, the leaseID is null at the first attempt to acquire lease. If the lease was acquired successfully, then the leaseID will be not null. It will be null otherwise.

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
PawasChhokra: Done.

samza-azure/src/test/java/org/apache/samza/MyTestExample.java
PawasChhokra: Removed.

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
```diff
@@ -0,0 +1,99 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Helper class for lease blob operations.
+ */
+public class LeaseBlobManager {
+
+  public static final Logger LOG = LoggerFactory.getLogger(LeaseBlobManager.class);
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
```diff
@@ -0,0 +1,99 @@
...
+   */
+  public String acquireLease(int leaseTimeInSec, String leaseId) {
+    try {
+      String id = leaseBlob.acquireLease(leaseTimeInSec, leaseId);
+      LOG.info("Acquired lease with lease id = " + id);
+      return id;
+    } catch (StorageException storageException) {
+      int httpStatusCode = storageException.getHttpStatusCode();
+      if (httpStatusCode == HttpStatus.CONFLICT_409) {
+        LOG.info("The blob you're trying to acquire is leased already.");
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
```diff
@@ -0,0 +1,99 @@
...
+      } else if (httpStatusCode == HttpStatus.NOT_FOUND_404) {
+        LOG.error("The blob you're trying to lease does not exist.", storageException);
+        throw new AzureException(storageException);
+      } else {
+        LOG.error("Error acquiring lease!", storageException);
```
PawasChhokra: Yes, it is functionally correct for those cases because someone else is already the leader if the blob is leased already. Also, for any other storage exception, the caller of the function will handle the behaviour.

samza-azure/src/main/java/org/apache/samza/AzureException.java
```diff
@@ -0,0 +1,40 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza;
+
+public class AzureException extends RuntimeException {
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/AzureClient.java
```diff
@@ -46,11 +52,11 @@
     } catch (IllegalArgumentException | URISyntaxException e) {
       LOG.error("\nConnection string {} specifies an invalid URI.", storageConnectionString);
       LOG.error("Please confirm the connection string is in the Azure connection string format.");
-      throw new SamzaException(e);
+      throw new AzureException(e);
```
PawasChhokra: Done.

.gitignore
```diff
@@ -29,3 +29,4 @@ out/
 *.patch
 **.pyc
 samza-shell/src/main/visualizer/plan.json
+samza-azure/src/test/java/org/apache/samza/MyTestExample.java
```
PawasChhokra: Done.

PR #257, opened 2017-08-03
SAMZA-1380: Create Utility Class for interacting with Azure Blob Storage

PawasChhokra:
PR 2: LeaseBlobManager
**PR 3: BlobUtils + JobModelBundle** (current PR)


samza-azure/src/main/java/org/apache/samza/BlobUtils.java
```diff
@@ -0,0 +1,240 @@
...
+  public BlobUtils(AzureClient client, String containerName, String blobName, long length) {
+    this.blobClient = client.getBlobClient();
+    try {
+      this.container = blobClient.getContainerReference(containerName);
+      container.createIfNotExists();
+      this.blob = container.getPageBlobReference(blobName);
+      if (!blob.exists()) {
+        blob.create(length);
+      }
+    } catch (URISyntaxException e) {
```
navina: When is URISyntaxException thrown? I suspect it is thrown by AzureClient. You may have to rebase this branch as well and modify the exception handling? 

Also, you are only logging the exception and not throwing. This will cause a cascade of failures on the caller side :)

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
```diff
@@ -0,0 +1,240 @@
...
+    try {
+      JobModelBundle bundle = new JobModelBundle(prevJM, currJM, prevJMV, currJMV);
+      byte[] data = SamzaObjectMapper.getObjectMapper().writerWithDefaultPrettyPrinter().writeValueAsBytes(bundle);
+      byte[] pageData = Arrays.copyOf(data, (int) JOB_MODEL_BLOCK_SIZE);
+      InputStream is = new ByteArrayInputStream(pageData);
+      blob.uploadPages(is, 0, JOB_MODEL_BLOCK_SIZE, AccessCondition.generateLeaseCondition(leaseId), null, null);
+      LOG.info("Uploaded {} jobModel to blob", bundle.getCurrJobModel());
+
+    } catch (Exception e) {
+      LOG.error("JobModel publish failed for version = " + currJMV, new SamzaException(e));
```
navina: You should throw this exception , apart from logging it. 
nit: when logging, you can simply log the original exception `e` You can throw the wrapped exception.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
```diff
@@ -0,0 +1,240 @@
...
+   * Called by the leader.
+   * @param prevJM Previous job model version that the processor was operating on.
+   * @param currJM Current job model version that the processor is operating on.
+   * @param prevJMV Previous job model version that the processor was operating on.
+   * @param currJMV Current job model version that the processor is operating on.
+   * @param leaseId LeaseID of the lease that the processor holds on the blob. Null if there is no lease.
+   */
+  public void publishJobModel(JobModel prevJM, JobModel currJM, String prevJMV, String currJMV, String leaseId) {
+    try {
+      JobModelBundle bundle = new JobModelBundle(prevJM, currJM, prevJMV, currJMV);
```
navina: Does it make sense to change the JobModelBundle from an immutable object to a mutable one? I am not sure how this is currently being used. But if it doesn't have concurrent access, then it should be simpler to create a mutable class `JobModelBundle` and provide only the current job model and version for `#publishJobModel`. JobModelBundle can automatically handle maintaining the previous and current job model. 

Just a thought. You can make the decision now or later based on how you are using this method.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
```diff
@@ -0,0 +1,240 @@
...
+      LOG.info("Uploaded {} jobModel to blob", bundle.getCurrJobModel());
+
+    } catch (Exception e) {
+      LOG.error("JobModel publish failed for version = " + currJMV, new SamzaException(e));
+    }
+  }
+
+  /**
+   * Reads the current job model from the blob.
+   * @return Current job model published on the blob.
```
navina: Update documentation: It can also return null when jobmodel is not found on the blob

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
```diff
@@ -0,0 +1,240 @@
...
+   */
+  public JobModel getJobModel() {
+    LOG.info("Reading the job model from blob.");
+    JobModelBundle jmBundle = getJobModelBundle();
+    if (jmBundle == null) {
+      LOG.error("Job Model details don't exist on the blob.");
+      return null;
+    }
+    JobModel jm = jmBundle.getCurrJobModel();
+    if (jm == null) {
```
navina: Just curious: is this case possible, when jobModelBundle is available, but no current JobModel?

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
```diff
@@ -0,0 +1,240 @@
...
+   * Called only by the leader.
+   * @param state Barrier state to be published to the blob.
+   * @param leaseId LeaseID of the lease that the processor holds on the blob. Null if there is no lease.
+   */
+  public void publishBarrierState(String state, String leaseId) {
+    try {
+      byte[] data = SamzaObjectMapper.getObjectMapper().writerWithDefaultPrettyPrinter().writeValueAsBytes(state);
+      byte[] pageData = Arrays.copyOf(data, (int) BARRIER_STATE_BLOCK_SIZE);
+      InputStream is = new ByteArrayInputStream(pageData);
+      blob.uploadPages(is, JOB_MODEL_BLOCK_SIZE, BARRIER_STATE_BLOCK_SIZE, AccessCondition.generateLeaseCondition(leaseId), null, null);
```
navina: How does AccessCondition work? Does that make this `uploadPages` a conditional write?

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
```diff
@@ -0,0 +1,240 @@
...
+    return jmVersion;
+  }
+
+  /**
+   * Writes the barrier state to the blob.
+   * Write is successful only if the lease ID passed is valid and the processor holds the lease.
+   * Called only by the leader.
+   * @param state Barrier state to be published to the blob.
+   * @param leaseId LeaseID of the lease that the processor holds on the blob. Null if there is no lease.
+   */
```
navina: Update javadoc regarding the exception thrown

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
```diff
@@ -0,0 +1,240 @@
...
+  public void publishBarrierState(String state, String leaseId) {
+    try {
+      byte[] data = SamzaObjectMapper.getObjectMapper().writerWithDefaultPrettyPrinter().writeValueAsBytes(state);
+      byte[] pageData = Arrays.copyOf(data, (int) BARRIER_STATE_BLOCK_SIZE);
+      InputStream is = new ByteArrayInputStream(pageData);
+      blob.uploadPages(is, JOB_MODEL_BLOCK_SIZE, BARRIER_STATE_BLOCK_SIZE, AccessCondition.generateLeaseCondition(leaseId), null, null);
+      LOG.info("Uploaded barrier state {} to blob", state);
+    } catch (Exception e) {
+      LOG.error("Barrier state " + state + " publish failed", new SamzaException(e));
```
navina: Same comment as before. You should log the original exception and throw the wrapper exception.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
PawasChhokra: Fixed to throw an exception. Also, the URISyntaxException is thrown when the blob name or the container name provided are invalid. Added that to the documentation.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
PawasChhokra: What you're saying makes sense. I will think more about this while adding the Azure Job Coordinator, and incorporate it in the same.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
PawasChhokra: Fixed to throw the exception. I have made a wrapper for AzureException to do the same.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
PawasChhokra: Updated.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
PawasChhokra: Not possible. Removed the check.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
PawasChhokra: Updated.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
PawasChhokra: ploadPages is only successfull when the AccessCondition provided has an active and valid lease ID. It fails otherwise.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
PawasChhokra: Fixed.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
navina: Can you add a comment at Line 152 about `AccessCondition` for reference?
Does failure to `#uploadPages` always throw `StorageException` or `IOException`? 

if the `leaseId` is null, would it be better to do nothing at all, instead of throwing exception?

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
```diff
@@ -0,0 +1,260 @@
...
+
+  /**
+   * Writes the list of live processors in the system to the blob.
+   * Write is successful only if the lease ID passed is valid and the processor holds the lease.
+   * Called only by the leader.
+   * @param processors List of live processors to be published on the blob.
+   * @param leaseId LeaseID of the valid lease that the processor holds on the blob. Null if there is no lease.
+   * @throws AzureException If an Azure storage service error or IO exception occurred.
+   */
+  public void publishLiveProcessorList(List<String> processors, String leaseId) {
```
navina: Perhaps useful to not do anything if leaseId is null ? 

Alternatively, the method can return a boolean indicating a successful operation or not. If you think it is appropriate for your usage, update it.

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
```diff
@@ -46,7 +46,7 @@ public LeaseBlobManager(CloudPageBlob leaseBlob) {
    * @return String that represents lease ID.  Null if acquireLease is unsuccessful because the blob is leased already.
    * @throws AzureException If a Azure storage service error occurred. This includes the case where the blob you're trying to lease does not exist.
    */
-  public String acquireLease(int leaseTimeInSec, String leaseId) {
+  public String acquireLease(int leaseTimeInSec, String leaseId) throws StorageException {
```
navina: Aren't you throwing `AzureException` ? :)

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
PawasChhokra: Done.

.gitignore
```diff
@@ -28,4 +28,5 @@ docs/learn/documentation/*/rest/javadocs
 out/
 *.patch
 **.pyc
+<<<<<<< HEAD
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/LeaseBlobManager.java
PawasChhokra: Fixed.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
PawasChhokra: Fixed. If leaseId is null, the function will do nothing and return false.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
PawasChhokra: Fixed.

samza-azure/src/main/java/org/apache/samza/BlobUtils.java
```diff
@@ -0,0 +1,260 @@
...
+   * @param containerName Name of container inside which we want the blob to reside.
+   * @param blobName Name of the blob to be managed.
+   * @param length Length of the page blob.
+   * @throws AzureException If an Azure storage service error occurred, or when the container name or blob name is invalid.
+   */
+  public BlobUtils(AzureClient client, String containerName, String blobName, long length) {
+    this.blobClient = client.getBlobClient();
+    try {
+      this.container = blobClient.getContainerReference(containerName);
+      container.createIfNotExists();
```
PawasChhokra: I have handled this by making sure that the blob is created only if it does not exist already. Updated the code accordingly.

navina: Awesome!

PR #259, opened 2017-08-03
SAMZA-1374: Implement Leader Election using Lease Blob in Azure

PawasChhokra:
PR 2: LeaseBlobManager
PR 3: BlobUtils + JobModelBundle
PR 4: TableUtils + ProcessorEntity
**PR 5: AzureLeaderElector** (current PR)


samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  /**
+   * Releases the lease in order to resign leadership. It also stops all schedulers scheduled by the leader.
+   */
+  @Override
+  public void resignLeadership() {
+    if (isLeader.get()) {
+      boolean status = leaseBlobManager.releaseLease(leaseId.get());
+      if (status) {
```
navina: What happens if `status` is false?

I think we should set `isLeader` state to false, irrespective of  whether release was successful or not.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
```diff
@@ -0,0 +1,118 @@
...
+        isLeader.set(false);
+        leaseId = null;
+        renewLeaseSF.cancel(true);
```
navina: Best to use defensive coding and do null checks for `renewLeaseSF` and `livenessSF`

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
```diff
@@ -0,0 +1,118 @@
...
+   * Tries to become the leader by acquiring a lease on the blob.
+   * Invokes the listener on becoming the leader.
+   */
+  @Override
+  public void tryBecomeLeader() {
+    leaseId.getAndSet(leaseBlobManager.acquireLease(LEASE_TIME_IN_SEC, leaseId.get()));
+    if (leaseId.get() != null) {
+      LOG.info("Became leader!");
+      isLeader.set(true);
+      leaderElectorListener.onBecomingLeader();
```
navina: Null check for `leaderElectorListener`

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
```diff
@@ -0,0 +1,118 @@
...
+    this.leaderElectorListener = listener;
+  }
+
+  /**
+   * Tries to become the leader by acquiring a lease on the blob.
+   * Invokes the listener on becoming the leader.
+   */
+  @Override
+  public void tryBecomeLeader() {
+    leaseId.getAndSet(leaseBlobManager.acquireLease(LEASE_TIME_IN_SEC, leaseId.get()));
```
navina: acquireLease throws `AzureException` under when the blob doesn't exist and when there is any other `StorageException`. I think you should handle them here.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
```diff
@@ -0,0 +1,118 @@
...
+  @Override
+  public boolean amILeader() {
+    return isLeader.get();
+  }
+
+  public String getLeaseId() {
+    return leaseId.get();
+  }
+
+  public void setRenewLeaseScheduledFuture(ScheduledFuture sf) {
```
navina: I am curious to see how this `ScheduledFuture` is used. Until then, can you please add a line or two of docs explaining what `livenessSF` and `renewLeaseSF` is meant for?

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
PawasChhokra: Done. I have removed the status right now and will add it in a later PR with a proper retry mechanism.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
PawasChhokra: On re-evaluation, I believe I don't need these variables. Hence, removed them.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
PawasChhokra: Same as above.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
```diff
@@ -0,0 +1,111 @@
...
+ * Every processor requires a valid active lease ID in order to perform successful write and delete operations on the blob.
+ * Read operations from the blob are not dependent on the lease ID.
+ */
+public class AzureLeaderElector implements LeaderElector {
+
+  private static final Logger LOG = LoggerFactory.getLogger(AzureLeaderElector.class);
+  private static final int LEASE_TIME_IN_SEC = 60;
+  private final LeaseBlobManager leaseBlobManager;
+  private LeaderElectorListener leaderElectorListener = null;
+  private AtomicReference<String> leaseId;
```
shanthoosh: Please mark the two atomic fields(leaseId, isLeader) as final.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
```diff
@@ -0,0 +1,111 @@
...
+   * Tries to become the leader by acquiring a lease on the blob.
+   * The acquireLease operation has a retry policy where upon failure, the operation is tried 3 times at 5 second intervals.
+   * Invokes the listener on becoming the leader.
+   */
+  @Override
+  public void tryBecomeLeader() {
+    try {
+      leaseId.getAndSet(leaseBlobManager.acquireLease(LEASE_TIME_IN_SEC, leaseId.get()));
+      if (leaseId.get() != null) {
+        LOG.info("Became leader!");
```
shanthoosh: Will be helpful to log the `leaseId` here. Please add it if you think it makes sense.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * Releases the lease in order to resign leadership. It also stops all schedulers scheduled by the leader.
+   * The releaseLease operation has a retry policy where upon failure, the operation is tried 3 times at 5 second intervals.
+   */
+  @Override
+  public void resignLeadership() {
+    if (isLeader.get()) {
+      leaseBlobManager.releaseLease(leaseId.get());
```
shanthoosh: Better to log at info/debug level that this thread is resigning the leadership along with `leaseId`.

navina:
@PawasChhokra Can you please fix the unused import in AzureClient Line 23? Otherwise, our checkstyle will not allow for a successful build -https://builds.apache.org/job/samza-pr-checks/807/console Thanks!

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
PawasChhokra: Done. Thank you!

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
```diff
@@ -0,0 +1,111 @@
...
+ * Class to facilitate leader election in Azure.
+ * The processor that acquires the lease on the blob becomes the leader.
+ * The lease ID is null initially. It is generated by Azure when the processor acquires the lease, and updated accordingly.
+ * Every processor requires a valid active lease ID in order to perform successful write and delete operations on the blob.
+ * Read operations from the blob are not dependent on the lease ID.
+ */
+public class AzureLeaderElector implements LeaderElector {
+
+  private static final Logger LOG = LoggerFactory.getLogger(AzureLeaderElector.class);
+  private static final int LEASE_TIME_IN_SEC = 60;
```
PawasChhokra: Can you elaborate more on why you think this should be configurable? I believe we don't need to expose the lease interval for leader election to the user.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
```diff
@@ -0,0 +1,111 @@
...
+
+
+/**
+ * Class to facilitate leader election in Azure.
+ * The processor that acquires the lease on the blob becomes the leader.
+ * The lease ID is null initially. It is generated by Azure when the processor acquires the lease, and updated accordingly.
+ * Every processor requires a valid active lease ID in order to perform successful write and delete operations on the blob.
+ * Read operations from the blob are not dependent on the lease ID.
+ */
+public class AzureLeaderElector implements LeaderElector {
```
PawasChhokra: I plan to add tests in the next PR. Will update you then :)

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
```diff
@@ -0,0 +1,111 @@
...
+      leaseId.getAndSet(leaseBlobManager.acquireLease(LEASE_TIME_IN_SEC, leaseId.get()));
+      if (leaseId.get() != null) {
+        LOG.info("Became leader!");
+        isLeader.set(true);
+        if (leaderElectorListener != null) {
+          leaderElectorListener.onBecomingLeader();
+        }
+      }
+    } catch (AzureException e) {
+      LOG.error("Error while trying to acquire lease.", e);
```
PawasChhokra: This exception occurs after a predefined number of retries according to the blob operations retry policy. Hence, I have swallowed the exception here and the Azure JC will take care of the corresponding behaviour.

samza-azure/src/main/java/org/apache/samza/AzureClient.java
```diff
@@ -47,7 +51,14 @@
   AzureClient(String storageConnectionString) {
     try {
       account = CloudStorageAccount.parse(storageConnectionString);
+
       blobClient = account.createCloudBlobClient();
+      // Set retry policy for operations on the blob. In this case, every failed operation on the blob will be retried thrice, after 5 second intervals.
```
PawasChhokra: The exception handling is exactly the same for creating the client side references for Azure services. Also, this instance is only created once for the duration of the application and used accordingly. Hence, I believe it doesn't make sense to decouple the client creation. Let me know if you have more feedback on this.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
shanthoosh: It's operational constant which can be figured out only after deploying it & consuming events in azure.

This value might be sufficient for some clients and might not be for others.

It's better to allow clients to define it(after having a reasonable default. Think of it like job.debounce.time or barrier.wait.time)

Otherwise if someone want's different blob lease interval, we need to provide new samza build (since it's hardcoded in code). Makes sense?

PR #260, opened 2017-08-03
Fix flaky, slow integration tests in TestZkStreamProcessor and TestZkStreamProcessorSession

shanthoosh:
Fix flaky and slow integration tests in TestZkStreamProcessor and TestZkStreamProcessorSession
Reason for failures:

There’re three configurable wait times in rebalancing phase in samza standalone before consensus is acheived and processing resumes with updated jobModel.

* debounceTime (Specified by `job.debounce.time.ms`. Upon processor change, leader waits for this interval before generating jobModel expecting stabilization in processors group(new arrival, deletion etc)).
* taskShutdownMs (Specified by `task.shutdown.ms`. Wait time for SamzaContainer shutdown in StreamProcessor).
* barrierWaitTimeOutMs (Specified by `job.coordinator.zk.consensus.timeout.ms`. Wait time for all processors in the group to join the barrier after creation).

Above wait times affects rebalancing phase duration. All these wait time have defaults in order of 40-60 seconds and not set to low values.

Flaky tests expects processors to come back up after rebalancing phase and drain message sources(Accomplished by checking a latch.count. RemoteApplicationRunner integration tests does exact same thing by checking if kafka input queue is drained directly with similar logic).

In worst case rebalancing phases can last upto 3-4 minutes(Making these tests sometime take 10 minutes at worst case).

Change:

Set all the above timeouts to 2 seconds(Sufficient for tests and verified by local build).

Benefits:

* Faster build time(Average runtime of these individual tests were reduced from 1m56s to 14s)
* More predicability in assertions(Didn’t fail even once in 30-40 attempts locally).

NOTE: If this doesn’t fix TestZkStreamProcessor and TestZkStreamProcessorSession,
longer term fix should be to use message markers in input source and 
shutdown taskCoordinator upon receiving them from TaskImpl(Or use 
bounded collection based pluggable InMemorySystemConsumer/InMemorySystemProducer).

### DIFF
--- a/samza-test/src/test/java/org/apache/samza/processor/TestZkStreamProcessorBase.java
+++ b/samza-test/src/test/java/org/apache/samza/processor/TestZkStreamProcessorBase.java
@@ -39,8 +39,10 @@ import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.samza.config.ApplicationConfig;
 import org.apache.samza.config.Config;
+import org.apache.samza.config.JobConfig;
 import org.apache.samza.config.JobCoordinatorConfig;
 import org.apache.samza.config.MapConfig;
+import org.apache.samza.config.TaskConfigJava;
 import org.apache.samza.config.ZkConfig;
 import org.apache.samza.coordinator.JobCoordinator;
 import org.apache.samza.coordinator.JobCoordinatorFactory;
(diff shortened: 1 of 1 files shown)


shanthoosh:
@sborya Please take a look when you get a chance.

shanthoosh:
@navina 
Default value for timeouts are very high (order of 30s) [I don't have complete context behind the decision].
Let's revisit it when there're production problems[Highly likely to happen].

navina:
> Let's revisit it when there're production problems[Highly likely to happen]

@shanthoosh If it is highly likely to happen in production, please bring it up before the release :) Let's discuss at tomorrow's meeting. Make sure to bring it up. Thanks!

PR #261, opened 2017-08-03
SAMZA-1378: Introduce and Implement Scheduler Interface for Polling in Azure

PawasChhokra:
PR 2: LeaseBlobManager
PR 3: BlobUtils + JobModelBundle
PR 4: TableUtils + ProcessorEntity
PR 5: AzureLeaderElector 
PR 6: Added all schedulers (current PR)


samza-azure/src/main/java/org/apache/samza/scheduler/HeartbeatScheduler.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Override
+  public ScheduledFuture scheduleTask() {
+    return scheduler.scheduleWithFixedDelay(() -> {
+        try {
+          LOG.info("Updating heartbeat");
+          table.updateHeartbeat(currentJMVersion.get(), processorId);
+        } catch (Exception e) {
+          LOG.error("Exception in Heartbeat Scheduler.", e);
```
navina: It is perhaps useful to provide a handler for exceptions. Otherwise, we are just swallowing the exceptions. 
Option 1: change `schedulerTask` interface to take in a lambda for error handling
Option 2: if every submitted task to this scheduler is going to use the same error handler, simply pass lambda/handler in the constructor

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
```diff
@@ -0,0 +1,402 @@
...
+  private JobModel jobModel = null;
+
+  /**
+   * Creates an instance of Azure job coordinator, along with references to Azure leader elector, Azure Blob and Azure Table.
+   * @param config User defined config
+   */
+  public AzureJobCoordinator(Config config) {
+    this.config = config;
+    processorId = createProcessorId(config);
+    currentJMVersion = new AtomicReference<>(INITIAL_STATE);
```
navina: Why is the initial value for `currentJMVersion` 0 ? Can you explain how you are using this variable?

samza-azure/src/main/java/org/apache/samza/scheduler/LeaderBarrierCompleteScheduler.java
```diff
@@ -0,0 +1,103 @@
...
+  private final AtomicBoolean barrierTimeout;
+  private SchedulerStateChangeListener listener = null;
+
+  public LeaderBarrierCompleteScheduler(TableUtils table,
+      String nextJMVersion, List<String> blobProcessorList, long startTime, AtomicBoolean barrierTimeout) {
+    this.table = table;
+    this.nextJMVersion = nextJMVersion;
+    this.blobProcessorSet = new HashSet<>(blobProcessorList);
+    this.startTime = startTime;
+    this.barrierTimeout = barrierTimeout;
```
navina: It took me a some time to understand how you were using `barrierTimeout` variable :)

Looks like you are creating a new instance of `LeaderBarrierCompleteScheduler` each time there is need to monitor the barrier state. However, I think you should `shutdown` this scheduler once it's operation is completed. You are canceling the current future. If that is the case, is there a need to create a new scheduler instance for every rebalance barrier attempt? 

We can discuss more in-person if my point is not coming across clearly.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
```diff
@@ -107,5 +107,4 @@ public boolean amILeader() {
   public LeaseBlobManager getLeaseBlobManager() {
     return this.leaseBlobManager;
   }
-
 }
```
navina: Can you please add a log line for the `else` condition above Line#73 ? 
I don't remember if we discussed this before. You should throw an exception when there is an `AzureException`.

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
```diff
@@ -0,0 +1,402 @@
...
+   */
+  private SchedulerStateChangeListener createLeaderLivenessListener() {
+    return () -> {
+      LOG.info("Existing leader died.");
+      azureLeaderElector.tryBecomeLeader();
+    };
+  }
+
+
+  private JobModel generateNewJobModel(List<String> processors) {
```
navina: nit: If this is used only in one place, it is ok to inline it.

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
```diff
@@ -0,0 +1,402 @@
...
+    JobModel prevJobModel = jobModel;
+    AtomicBoolean barrierTimeout = new AtomicBoolean(false);
+
+    nextJMVersion = Integer.toString(Integer.valueOf(prevJMVersion) + 1);
+    if (currentProcessorIds.isEmpty()) {
+      currentProcessorIds = new ArrayList<>(table.getActiveProcessorsList(currentJMVersion));
+    } else {
+      //Check if previous barrier not reached, then previous barrier times out.
+      String blobJMV = leaderBlob.getJobModelVersion();
+      if (blobJMV != null && Integer.valueOf(blobJMV) > Integer.valueOf(prevJMVersion)) {
```
navina: what happens when blobJMV == null? I think you should handle that case as well.

navina: It is perhaps noting that the invariant blobJMV >= prevJMVersion. Is that right?

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
```diff
@@ -0,0 +1,402 @@
...
+        prevJMVersion = blobJMV;
+        prevJobModel = leaderBlob.getJobModel();
+        nextJMVersion = Integer.toString(Integer.valueOf(blobJMV) + 1);
```
navina: I think you can safely generate `nextJMVersion` right before the JobModel generation, instead of setting it in Line 254 and here

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
```diff
@@ -0,0 +1,402 @@
...
+        leaderBlob.publishBarrierState(BarrierState.TIMEOUT.name() + " " + blobJMV, azureLeaderElector.getLeaseId().get());
+      }
+    }
+
+    // Generate the new JobModel
+    JobModel newJobModel = generateNewJobModel(currentProcessorIds);
+    LOG.info("pid=" + processorId + "Generated new Job Model. Version = " + nextJMVersion);
+
+    // Publish the new job model
+    leaderBlob.publishJobModel(prevJobModel, newJobModel, prevJMVersion, nextJMVersion, azureLeaderElector.getLeaseId().get());
```
navina: I think each of the `publish*` method has a return value indicating the status of the operation. You should ideally add appropriate behaviors when the operation fails.

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
```diff
@@ -0,0 +1,402 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Class that provides coordination mechanism for Samza standalone in Azure.
+ * Handles processor lifecycle through Azure blob and table storage. Orchestrates leader election.
+ * The leader job coordinator generates partition mapping, writes shared data to the blob and manages rebalancing.
+ */
+public class AzureJobCoordinator implements JobCoordinator {
+
```
PawasChhokra: Removed.

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
PawasChhokra: The blobJMV will only be null when the first processor is spawned and for that case, it will get into the if block, because the list of currentProcessorIds will be empty (as it is called from onBecomingLeader()). For every other case when the processor is not able to retrieve the blobJMV, it will throw an exception in getJobModelVersion().

PawasChhokra: It is to be noted that the local variable currentJMVersion is only set when rebalancing completes successfully in onNewJobModelConfirmed(). The expected initial value of prevJMVersion/currentJMVersion when entering this function should be the same as the blobJMV. This is because this function is only invoked by the leader, and it is the leader that will introduce the difference between these two values by publishing the new blobJMV > prevJMVersion.

samza-azure/src/main/java/org/apache/samza/AzureLeaderElector.java
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/scheduler/HeartbeatScheduler.java
```diff
@@ -0,0 +1,79 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(HeartbeatScheduler.class);
+  private static final long HEARTBEAT_DELAY_SEC = 5;
+  private static final ThreadFactory PROCESSOR_THREAD_FACTORY =
+      new ThreadFactoryBuilder().setNameFormat("HeartbeatScheduler-%d").build();
+  private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(5, PROCESSOR_THREAD_FACTORY);
+  private final String processorId;
+  private final TableUtils table;
+  private final AtomicReference<String> currentJMVersion;
+  private SchedulerStateChangeListener listener = null;
```
PawasChhokra: Removed.

samza-azure/src/main/java/org/apache/samza/scheduler/LeaderBarrierCompleteScheduler.java
PawasChhokra: Makes sense. I have changed the code so that the scheduler is shutdown after completing its operation.

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
```diff
@@ -0,0 +1,483 @@
...
+      versionUpgradeDetected.getAndSet(false);
+      String state;
+      if (barrierTimeout.get()) {
+        LOG.error("Barrier timed out for version {}", nextJMVersion);
+        state = BarrierState.TIMEOUT.name() + " " + nextJMVersion;
+      } else {
+        LOG.info("Leader detected barrier completion.");
+        state = BarrierState.END.name() + " " + nextJMVersion;
+      }
+      if (!leaderBlob.publishBarrierState(state, azureLeaderElector.getLeaseId().get())) {
```
nickpan47: Seems that this can only be performed by the leader. What if during the wait for barrier completion, leader has lost its lease to the leaderBlob? I did not see a logic here to handle that case.

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
```diff
@@ -0,0 +1,483 @@
...
+    Set<SystemStreamPartition> allSystemStreamPartitions = getInputStreamPartitions();
+    SystemStreamPartitionGrouper grouper = getSystemStreamPartitionGrouper();
+    Map<TaskName, Set<SystemStreamPartition>> groups = grouper.group(allSystemStreamPartitions);
+    LOG.info("SystemStreamPartitionGrouper " + grouper.toString() + " has grouped the SystemStreamPartitions into " + Integer.toString(groups.size()) +
+        " tasks with the following taskNames: {}", groups.keySet());
+    return groups.size();
+  }
+
+  /**
+   * Called only by the leader, either when the processor becomes the leader, or when the list of live processors changes.
```
nickpan47: Same question here, what if the leader lost the lease to leaderBlob while doOnProcessorChange() is running? Any leader-specific actions should be stopped until the leader resumes the lease, or completely aborted if another process assume the leader position.

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
```diff
@@ -0,0 +1,483 @@
...
+
+      // Schedule a task to renew the lease after a fixed time interval
+      LOG.info("Starting scheduler to keep renewing lease held by the leader.");
+      renewLease.scheduleTask();
+
+      doOnProcessorChange(new ArrayList<>());
+
+      // Start scheduler to check for change in list of live processors
+      LOG.info("Starting scheduler to check for change in list of live processors in the system.");
+      liveness.setStateChangeListener(createLivenessListener(liveness.getLiveProcessors()));
```
nickpan47: I think that there is an issue when you have separate scheduler for some action that only leader performs, while the leader liveness / renewLease is on another scheduler. You will either have to a) implement a shared flag indicating whether the current process is still the leader when trying to perform leader actions in those schedulers, or b) re-structure the schedulers to combine leader actions in the scheduler to detect leader liveness/ renewLease, in which you only perform those leader actions after confirming the leadership is still valid in the renewLease scheduler. I think option b) is probably easier.

samza-azure/src/main/java/org/apache/samza/BarrierState.java
```diff
@@ -0,0 +1,27 @@
+/*
```
nickpan47: Quick comment on the package organization of all source code files in samza-azure. Please put the classes into a proper sub-package name, instead of clubbing them all under org.apache.samza. For example, AzureJobCoordinator definitely should go to package org.apache.samza.coordinator

samza-azure/src/main/java/org/apache/samza/ProcessorEntity.java
```diff
@@ -38,11 +39,13 @@ public ProcessorEntity(String jobModelVersion, String processorId) {
     this.partitionKey = jobModelVersion;
     this.rowKey = processorId;
     this.isLeader = false;
+    Random rand = new Random();
+    this.liveness = rand.nextInt(10000) + 2;
```
nickpan47: Maybe I misunderstood the liveness value's meaning here. What is the random value used for? The length of the heartbeat interval in seconds? Or just a random value? If not for the length of heartbeat interval, what's the significance of the random value here?

samza-azure/src/main/java/org/apache/samza/ProcessorEntity.java
```diff
@@ -38,11 +39,13 @@ public ProcessorEntity(String jobModelVersion, String processorId) {
...
   }
 
   /**
    * Updates heartbeat by updating the liveness value in the table.
-   * @param value
+   * @param value Random integer value
```
nickpan47: Some explanation why do we need a random integer here as liveness?

samza-azure/src/main/java/org/apache/samza/TableUtils.java
```diff
@@ -44,35 +44,38 @@
...
    * @param isLeader Denotes whether the processor is a leader or not.
    * @throws AzureException If an Azure storage service error occurred.
    */
-  public void addProcessorEntity(String jmVersion, String pid, int liveness, boolean isLeader) {
+  public void addProcessorEntity(String jmVersion, String pid, boolean isLeader) {
     ProcessorEntity entity = new ProcessorEntity(jmVersion, pid);
     entity.setIsLeader(isLeader);
-    entity.setLiveness(liveness);
+    Random rand = new Random();
+    entity.setLiveness(rand.nextInt(10000) + 2);
```
nickpan47: Please do explain what's the usage of this liveness value.

samza-azure/src/main/java/org/apache/samza/scheduler/RenewLeaseScheduler.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.samza.LeaseBlobManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Scheduler class to keep renewing the lease once an entity has acquired it.
+ * Renews every 45 seconds.
+ * All time units are in SECONDS.
+ */
+public class RenewLeaseScheduler implements TaskScheduler {
```
nickpan47: It does not seems that you have very different scheduler functions among all the scheduler classes defined here. Rather, it seems like you can have a single defined scheduler, but different Runnable to be scheduled at fixed rate instead? Also, why do you want to use multiple threads for RenewLeaseScheduler? It seems to me that the RenewLease tasks are very important to be serialized s.t. we don't have race conditions in maintaining the lease renew from the same process. Imagine a situation that you use RenewLeaseScheduler to renew the lease to the leaderBlob, the renew operation would be a critical operation and would be nice to run single threaded to avoid race conditions.

build.gradle
```diff
@@ -134,6 +134,7 @@ project(':samza-api') {
   }
 }
 
+
```
PawasChhokra: Removed.

samza-azure/src/main/java/org/apache/samza/ProcessorEntity.java
PawasChhokra: The liveness is a random value that is written to an attribute of the row in order to update its last modified since timestamp. The random value is there just to update the row in some sense, to assert to the leader that the processor is alive.

samza-azure/src/main/java/org/apache/samza/ProcessorEntity.java
PawasChhokra: It doesn't have to be a random integer necessarily. It could be anything. What else would you suggest we do?

samza-azure/src/main/java/org/apache/samza/TableUtils.java
PawasChhokra: Explained above. Let me know what you have in mind for the same.

samza-azure/src/main/java/org/apache/samza/scheduler/RenewLeaseScheduler.java
PawasChhokra: In my next commit, I have addressed this issue. I have also added proper exception handling for the case when the leader is not able to renew the lease. It gracefully resigns the leadership and another processor takes over as the leader.

samza-azure/src/main/java/org/apache/samza/scheduler/TaskScheduler.java
```diff
@@ -0,0 +1,35 @@
...
+
+
+/**
+ * Interface for scheduling tasks for Azure Job Coordinator.
+ */
+public interface TaskScheduler {
+
+  ScheduledFuture scheduleTask();
+
+  void setStateChangeListener(SchedulerStateChangeListener listener);
```
PawasChhokra: As discussed with you offline, I am going to change the scheduling interfaces to have one scheduler with multiple threads that schedule these tasks and another scheduler that acts as a worker queue.

samza-azure/src/main/java/org/apache/samza/BarrierState.java
PawasChhokra: Will do. Thanks!

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
PawasChhokra: I have added that logic in my next commit. Whenever the leader is unable to renew the lease, it gracefully resigns the leadership and another processor becomes the leader. This new leader will detect the change in list of live processors again and trigger a new rebalancing cycle.

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
PawasChhokra: Explained above. Will push the commit in a while.

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
PawasChhokra: If there is failure in renewing the lease, I am updating the isLeader value of that processor to be false and resigning the leadership. Also it will shutdown all the schedulers that the leader is supposed to run. This will indicate to the other processors that the current leader died and someone else will become the leader and start all these tasks again.

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
PawasChhokra: I have changed it to the previous value which was "unassigned".

samza-azure/src/main/java/org/apache/samza/scheduler/HeartbeatScheduler.java
PawasChhokra: I have added a Consumer function to the constructor of every scheduler for error handling;

samza-azure/src/main/java/org/apache/samza/AzureJobCoordinator.java
navina: The reason I asked about the blobJMV being null is because it is possible for `BlobUtils#getJobModelVersion` to return null. Shouldn't that be considered as an errorCondition ?

samza-azure/src/main/java/org/apache/samza/coordinator/AzureJobCoordinator.java
```diff
@@ -0,0 +1,509 @@
...
+    renewLease = null;
+    liveness = null;
+  }
+
+  @Override
+  public void start() {
+
+    LOG.info("Starting Azure job coordinator.");
+    streamMetadataCache = StreamMetadataCache.apply(METADATA_CACHE_TTL_MS, config);
+    table.addProcessorEntity(INITIAL_STATE, processorId, azureLeaderElector.amILeader());
```
navina: shouldn't `azureLeaderElector.amILeader()` always return `false` at this point?

samza-azure/src/main/java/org/apache/samza/coordinator/scheduler/HeartbeatScheduler.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+      }, HEARTBEAT_DELAY_SEC, HEARTBEAT_DELAY_SEC, TimeUnit.SECONDS);
+  }
+
+  @Override
+  public void setStateChangeListener(SchedulerStateChangeListener listener) {}
+
+  @Override
+  public void shutdown() {
+    scheduler.shutdownNow();
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/coordinator/scheduler/HeartbeatScheduler.java
```diff
@@ -0,0 +1,79 @@
...
+    this.currentJMVersion = currentJMVersion;
+    processorId = pid;
+    this.errorHandler = errorHandler;
+  }
+
+  @Override
+  public ScheduledFuture scheduleTask() {
+    return scheduler.scheduleWithFixedDelay(() -> {
+        try {
+          LOG.info("Updating heartbeat for processor ID: " + processorId + " and job model version: " + currentJMVersion.get());
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/coordinator/data/ProcessorEntity.java
```diff
@@ -38,11 +39,13 @@ public ProcessorEntity(String jobModelVersion, String processorId) {
...
+    Random rand = new Random();
+    this.liveness = rand.nextInt(10000) + 2;
   }
 
   /**
    * Updates heartbeat by updating the liveness value in the table.
-   * @param value
+   * @param value Random integer value
    */
   public void setLiveness(int value) {
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/coordinator/AzureJobCoordinator.java
PawasChhokra: True. Changed it to false.

samza-azure/src/main/java/org/apache/samza/util/TableUtils.java
```diff
@@ -44,35 +47,38 @@
...
   /**
    * Add a row which denotes an active processor to the processor table.
    * @param jmVersion Job model version that the processor is operating on.
    * @param pid Unique processor ID.
-   * @param liveness Random heartbeat value.
    * @param isLeader Denotes whether the processor is a leader or not.
    * @throws AzureException If an Azure storage service error occurred.
    */
-  public void addProcessorEntity(String jmVersion, String pid, int liveness, boolean isLeader) {
+  public void addProcessorEntity(String jmVersion, String pid, boolean isLeader) {
```
PawasChhokra: Added a doc in the ProcessorEntity class.

samza-azure/src/main/java/org/apache/samza/coordinator/scheduler/LeaderBarrierCompleteScheduler.java
```diff
@@ -0,0 +1,117 @@
...
+    this.currentJMVersion = currentJMVersion;
+  }
+
+  @Override
+  public ScheduledFuture scheduleTask() {
+    return scheduler.scheduleWithFixedDelay(() -> {
+        try {
+          if (!table.getEntity(currentJMVersion.get(), processorId).getIsLeader()) {
+            LOG.info("Not the leader anymore. Shutting down LeaderBarrierCompleteScheduler.");
+            scheduler.shutdownNow();
```
PawasChhokra: I have added the scheduler code in an if-else block and am shutting down the scheduler only in the listener now. This should address your review for now. Will do the rest of the refactoring as a part of a different JIRA.

samza-azure/src/main/java/org/apache/samza/coordinator/scheduler/LeaderLivenessCheckScheduler.java
```diff
@@ -0,0 +1,120 @@
...
+
+
+/**
+ * Scheduler class invoked by each processor to check if the leader is alive.
+ * Checks every 30 seconds.
+ * If a processor row hasn't been updated since 30 seconds, the system assumes that the processor has died.
+ * All time units are in SECONDS.
+ */
+public class LeaderLivenessCheckScheduler implements TaskScheduler {
+
```
PawasChhokra: Removed.

samza-azure/src/main/java/org/apache/samza/coordinator/scheduler/LivenessCheckScheduler.java
```diff
@@ -0,0 +1,107 @@
...
+  private final AtomicReference<List<String>> liveProcessorsList;
+  private final Consumer<String> errorHandler;
+  private SchedulerStateChangeListener listener = null;
+  private final String processorId;
+
+  public LivenessCheckScheduler(Consumer<String> errorHandler, TableUtils table, BlobUtils blob, AtomicReference<String> currentJMVersion, final String pid) {
+    this.table = table;
+    this.blob = blob;
+    this.currentJMVersion = currentJMVersion;
+    liveProcessorsList = new AtomicReference<>(null);
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/util/LeaseBlobManager.java
```diff
@@ -54,7 +55,7 @@ public String acquireLease(int leaseTimeInSec, String leaseId) {
     } catch (StorageException storageException) {
       int httpStatusCode = storageException.getHttpStatusCode();
       if (httpStatusCode == HttpStatus.CONFLICT_409) {
-        LOG.info("The blob you're trying to acquire is leased already.", storageException);
+        LOG.info("The blob you're trying to acquire is leased already.");
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/AzureClient.java
```diff
@@ -44,21 +45,25 @@
   /**
    * Creates a reference to the Azure Storage account according to the connection string that the client passes.
    * Also creates references to Azure Blob Storage and Azure Table Storage.
-   * @param storageConnectionString Connection string to conenct to Azure Storage Account, format: "DefaultEndpointsProtocol=<https>;AccountName=<>;AccountKey=<>"
+   * @param storageConnectionString Connection string to connect to Azure Storage Account
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/coordinator/scheduler/JMVersionUpgradeScheduler.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  @Override
+  public ScheduledFuture scheduleTask() {
+    return scheduler.scheduleWithFixedDelay(() -> {
+        try {
+          LOG.info("Checking for job model version upgrade");
+          // Read job model version from the blob.
+          String blobJMV = blob.getJobModelVersion();
+          LOG.info("Blob JMV: {}", blobJMV);
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/coordinator/scheduler/JMVersionUpgradeScheduler.java
```diff
@@ -0,0 +1,98 @@
...
+          String blobBarrierState = blob.getBarrierState();
+          String currentJMV = currentJMVersion.get();
+          LOG.info("currentJMV: {}", currentJMV);
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/coordinator/scheduler/LeaderBarrierCompleteScheduler.java
```diff
@@ -0,0 +1,117 @@
...
+            listener.onStateChange();
+          }
+          LOG.info("Leader checking for barrier state");
+          // Get processor IDs listed in the table that have the new job model verion.
+          Iterable<ProcessorEntity> tableList = table.getEntitiesWithPartition(nextJMVersion);
+          Set<String> tableProcessors = new HashSet<>();
+          for (ProcessorEntity entity : tableList) {
+            tableProcessors.add(entity.getRowKey());
+          }
+          LOG.info("blobProcessorSet = {}", blobProcessorSet);
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/coordinator/scheduler/LeaderBarrierCompleteScheduler.java
```diff
@@ -0,0 +1,117 @@
...
+          LOG.info("tableProcessors = {}", tableProcessors);
```
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/coordinator/scheduler/LeaderLivenessCheckScheduler.java
```diff
@@ -0,0 +1,120 @@
...
+      if (entity.getIsLeader()) {
+        leader = entity;
+      }
+    }
+    int currJMVInt = 0;
+    if (!currJMV.equals(initialState)) {
+      currJMVInt = Integer.valueOf(currJMV);
+    }
+    if (Integer.valueOf(blobJMV) > currJMVInt) {
+      LOG.info("Leader info 2 in LeaderLivenessCheckScheduker: {}", leader);
```
PawasChhokra: Removed.

samza-azure/src/main/java/org/apache/samza/coordinator/scheduler/HeartbeatScheduler.java
```diff
@@ -0,0 +1,79 @@
...
+          table.updateHeartbeat(currentJMVersion.get(), processorId);
+        } catch (Exception e) {
+          errorHandler.accept("Exception in Heartbeat Scheduler. Stopping the processor...");
+        }
+      }, HEARTBEAT_DELAY_SEC, HEARTBEAT_DELAY_SEC, TimeUnit.SECONDS);
+  }
+
+  @Override
+  public void setStateChangeListener(SchedulerStateChangeListener listener) {}
```
PawasChhokra: Will do :)

PR #262, opened 2017-08-04
SAMZA-1173: Refactor zk coordination utils

sborya:


samza-core/src/main/java/org/apache/samza/coordinator/LeaderElector.java
```diff
@@ -56,4 +56,6 @@
    * @return True, if the caller is the current leader. False, otherwise
    */
   boolean amILeader();
+
+  void close();
```
xinyuiscool: could you provide some documentation on close()? is it true after close() there won't be no more leader?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -207,40 +199,36 @@ public void waitForFinish() {
...
         leaderElector.setLeaderElectorListener(() -> {
             getStreamManager().createStreams(intStreams);
             initLatch.countDown();
           });
         leaderElector.tryBecomeLeader();
         initLatch.await(LATCH_TIMEOUT_MINUTES, TimeUnit.MINUTES);
+
+        // close the connections
+        leaderElector.close();
+        initLatch.close();
```
xinyuiscool: what will happen if a new StreamProcessor joins after the app is started and running? Does it trigger leader election?

sborya:
closed in favor of PR 284

PR #263, opened 2017-08-08
SAMZA-1384: Race condition with async commit affects checkpoint correctness

jmakes:


jmakes:
@prateekm @nickpan47 @xinyuiscool @vjagadish1989

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -245,14 +263,12 @@ class OffsetManager(
 
       // invoke checkpoint listeners
       //partitionOffsets.groupBy(_._1.getSystem).foreach {
-      partitionOffsets.groupBy { case (ssp, _) => ssp.getSystem }.foreach {
+      checkpoint.getOffsets.asScala.groupBy { case (ssp, _) => ssp.getSystem }.foreach {
         case (systemName:String, offsets: Map[SystemStreamPartition, String]) => {
```
prateekm: Nitpick: Unrelated, Since checkpoint.offsets is already a java.util.Map, I think this is doing unnecessary Java -> Scala -> Java conversion. Can you try making this `offsets: util.Map[SSP, String]` and remove the `offsets.asJava` below and see if it still compiles/works?

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -218,7 +220,7 @@ class TaskInstance(
 
     trace("Checkpointing offsets for taskName: %s" format taskName)
```
prateekm: Maybe update message to log the offsets being checkpointed.

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
```diff
@@ -368,8 +367,10 @@ public void testCommitSingleTask() throws Exception {
     task0ProcessedMessagesLatch.await();
     task1ProcessedMessagesLatch.await();
 
-    verify(offsetManager).checkpoint(taskName0);
-    verify(offsetManager, never()).checkpoint(taskName1);
+    verify(offsetManager).createCheckpoint(taskName0);
```
prateekm: Slightly cleaner to verify that TaskInstance's commit is called/not called instead of interacting with the OffsetManager directly (since the contract AsyncRunLoop is with TaskInstance, not OffsetManager.)

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
```diff
@@ -368,8 +367,10 @@ public void testCommitSingleTask() throws Exception {
...
+    verify(offsetManager).writeCheckpoint(taskName0, any());
+    verify(offsetManager, never()).createCheckpoint(taskName1);
+    verify(offsetManager, never()).writeCheckpoint(taskName1, any());
   }
 
   //@Test
```
prateekm: Looks like these tests don't run at all? Shanthoosh, are these fixed yet.

shanthoosh: @prateekm 
I will fix this in a separate patch.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -217,25 +216,44 @@ class OffsetManager(
...
 
   /**
-   * Checkpoint all offsets for a given TaskName using the CheckpointManager.
-   */
-  def checkpoint(taskName: TaskName) {
+    * Gets a snapshot of all the current offsets for the specified task. This is useful to
+    * ensure there are no concurrent updates to the offsets between when this method is
+    * invoked and the corresponding call to [[OffsetManager.writeCheckpoint()]]
+    */
+  def createCheckpoint(taskName: TaskName): Checkpoint = {
```
jmakes: So I originally called it getCheckpoint() but found myself asking "does it get the value of the latest written checkpoint, or does it create a new one?"

In the spirit of compromise, how about getNewCheckpoint() or buildCheckpoint()?

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -217,25 +216,44 @@ class OffsetManager(
...
     if (checkpointManager != null || checkpointListeners.nonEmpty) {
-      debug("Checkpointing offsets for taskName %s." format taskName)
+      debug("Getting checkpoint offsets for taskName %s." format taskName)
 
+      // Which ssps are assigned to this task?
       val sspsForTaskName = systemStreamPartitions.getOrElse(taskName, throw new SamzaException("No such SystemStreamPartition set " + taskName + " registered for this checkpointmanager")).toSet
```
jmakes: Agreed. Fixed the error.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -217,25 +216,44 @@ class OffsetManager(
...
+      // What are all the latests offsets for this task?
```
jmakes: Actually didn't mean to leave these comments in. :-)
Removed.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
jmakes: Doing that in the OffsetManager now. Since that message is at level DEBUG, it should appear whenever this one does. So prefer not to print all those offsets twice.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
jmakes: scala.MatchError: (test-system2,Map(SystemStreamPartition [test-system2, test-stream2, 0] -> 100)) (of class scala.Tuple2)
        at org.apache.samza.checkpoint.OffsetManager$$anonfun$writeCheckpoint$4.apply(OffsetManager.scala:264)

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -20,20 +20,19 @@
 package org.apache.samza.checkpoint
 
 
+import java.util
```
jmakes: It's needed for the hashmap. Not sure why intellij decided to import the whole package.
Made it more specific.

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
jmakes: Agree, but to fix this we'd have to convert all TaskInstances in this test suite to be mocks

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
prateekm: buildCheckpoint() sounds pretty reasonable.

PR #264, opened 2017-08-09
SAMZA-1372: Change Latch Interface to Lock Interface for Samza Standalone with ZK

PawasChhokra:


samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -114,4 +114,13 @@ default boolean createStream(StreamSpec streamSpec) {
   default void validateStream(StreamSpec streamSpec) throws StreamValidationException {
     throw new UnsupportedOperationException();
   }
+
+  /**
+   * Check if the stream described by the spec already exists.
+   * @param streamSpec The spec, or blueprint for the physical stream on the system.
+   * @return true if stream exists already, false otherwise
```
sborya: nit. the stream

samza-core/src/main/java/org/apache/samza/coordinator/CoordinationUtils.java
```diff
@@ -41,4 +41,6 @@
   LeaderElector getLeaderElector(); // leaderElector is unique based on the groupId
 
   Latch getLatch(int size, String latchId);
+
+  Lock getLock();
```
sborya: I'd rather named it CoordinationLock (or something), to reduce confusion between this and java concurrency lock.

samza-core/src/main/java/org/apache/samza/coordinator/Lock.java
```diff
@@ -0,0 +1,38 @@
...
+  boolean lock();
+
+  /**
+   * Releases the lock
+   */
+  void unlock();
+
+  boolean hasLock();
+
+  void setLockListener(LockListener listener);
```
sborya: I personally think that lock should be synchronous utility. It makes programming much easier.
lock = getLock();

lock.lock()
.....
lock.release/unlock()

samza-core/src/main/java/org/apache/samza/execution/StreamManager.java
```diff
@@ -75,4 +75,19 @@ public void createStreams(List<StreamSpec> streams) {
 
     return streamToPartitionCount;
   }
+
+  /**
+   * Check if the streams described by the specs already exist.
+   * @param streams A list of stream specs, whose existence we need to check for
```
sborya: nit. s/A/a/

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -249,6 +251,44 @@ public void waitForFinish() {
...
+  /**
+   * Create intermediate streams using {@link org.apache.samza.execution.StreamManager}.
+   * If {@link CoordinationUtils} is provided, this function will first acquire a lock, and then create the streams.
+   * All the runner processes will wait till the time they acquire the lock. On acquiring lock, they will check if the streams have already been created.
+   * If streams exist already, they will unlock and proceed normally. If streams don't exist, they create the streams and then unlock.
+   * @param intStreams list of intermediate {@link StreamSpec}s
+   * @throws Exception exception for latch timeout
+   */
+  private void createStreamsWithLock(List<StreamSpec> intStreams) throws Exception {
+    if (!intStreams.isEmpty()) {
```
sborya: nit. I prefer if(intStreams.isEmpty()) return;  to avoid unnecessary indentation. But that's personal preference.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -249,6 +251,44 @@ public void waitForFinish() {
...
+      if (coordinationUtils != null) {
+        Lock initLock = coordinationUtils.getLock();
+        LockListener lockListener = null;
+        if (coordinationUtils.getClass().getName().equals("org.apache.samza.zk.ZkCoordinationUtils")) {
```
sborya: I know this is a shortcut, but I think it'd be better to put this logic into the coordinationUtils, not here. LocalApplicationRunner should not know about Zk stuff.

samza-core/src/test/java/org/apache/samza/zk/TestZkLock.java
```diff
@@ -0,0 +1,23 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.zk;
+
+public class TestZkLock {
```
sborya: not very efficient test :)

samza-core/src/main/java/org/apache/samza/zk/ZkLock.java
```diff
@@ -0,0 +1,129 @@
...
+   * Tries to acquire a lock in order to create intermediate streams. On failure to acquire lock, it keeps trying until the lock times out.
+   * Creates a sequential ephemeral node to acquire the lock. If the path of this node has the lowest sequence number, the processor has acquired the lock.
+   */
+  @Override
+  public boolean lock() {
+    nodePath = zkUtils.getZkClient().createEphemeralSequential(lockPath + "/", participantId);
+
+    //Start timer for timeout
+    ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor(new ThreadFactoryBuilder().setNameFormat("ZkLock-%d").build());
+    scheduler.schedule(() -> {
```
shanthoosh: As we'd discussed offline, this exception will not get propagated to the other main 
thread which is running the while loop(Since the exception thrown by the
threadPool's  thread doesn't reach the main thread's while loop).

So essentially all the followers will go into infinite loop (as long as there's a leader). 

Please share a separate state (something like CountdownLatch, AtomicBoolean, 
volatile boolean) between the two threads to indicate that the periodic 
polling should stop. 

Please correct me if i'm wrong.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -249,6 +251,44 @@ public void waitForFinish() {
     }
   }
 
+
+  private void createIntermediateStreams(List<StreamSpec> intStreams) throws Exception {
+    boolean streamsExist = getStreamManager().checkIfStreamsExist(intStreams);
```
bharathkk: Why don't we get the subset of `intStreams` that are not created and only try creating with them? In case of stream already exists, we do the work twice; once for checking if it exists and other time during the actual creation which would again result in no op.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -114,4 +114,13 @@ default boolean createStream(StreamSpec streamSpec) {
...
+   */
+  default boolean existStream(StreamSpec streamSpec) {
```
prateekm: Don't have context on this PR, just wondering why we need a new method in the SystemAdmin API for this? Is this just for convenience or do we not currently have a way to check if a stream exists?

vjagadish1989:
@navina : 
Is this for merge into master? Or is the plan to merge this to a separate branch?

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
vjagadish1989: Would n't the `getStreamMetadata` API on `SystemAdmin` be sufficient?

navina:
@vjagadish1989  This is for 0.14.0 branch.

samza-core/src/main/java/org/apache/samza/coordinator/CoordinationUtils.java
PawasChhokra: Renamed it to distributed lock :)

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
PawasChhokra: It seems like we won't be needing this function at all. Hence removed.

samza-core/src/main/java/org/apache/samza/coordinator/Lock.java
PawasChhokra: Removed the listener interface.

samza-core/src/main/java/org/apache/samza/execution/StreamManager.java
PawasChhokra: It seems like we won't be needing this function at all. Hence removed.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
PawasChhokra: Removed.

samza-core/src/main/java/org/apache/samza/zk/ZkLock.java
PawasChhokra: Following this discussion, I have changed the lock timeout implementation. Kindly take a look at it and let me know if you have any comments. Thank you!

samza-core/src/main/java/org/apache/samza/coordinator/Lock.java
```diff
@@ -0,0 +1,38 @@
...
+
```
PawasChhokra: Removed.

samza-core/src/main/java/org/apache/samza/coordinator/Lock.java
```diff
@@ -0,0 +1,38 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.coordinator;
+
```
PawasChhokra: Added :)

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
PawasChhokra: Will keep in mind for the future, thank you!

samza-core/src/main/java/org/apache/samza/coordinator/LockListener.java
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.coordinator;
+
+public interface LockListener {
```
PawasChhokra: Removed the class.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -272,4 +312,28 @@ StreamProcessor createStreamProcessor(
...
+        try {
+          createIntermediateStreams(intStreams);
+          LOG.info("Created intermediate streams successfully!");
+        } catch (Exception e) {
+          onError();
+        }
+        initLock.unlock();
+      }
+
+      @Override
```
PawasChhokra: Removed.

samza-core/src/main/java/org/apache/samza/zk/ZkLock.java
```diff
@@ -0,0 +1,129 @@
...
+        if (!hasLock.get()) {
+          throw new SamzaException("Timed out while acquiring lock for creating intermediate streams.");
+        }
+      }, zkConfig.getZkSessionTimeoutMs(), TimeUnit.MILLISECONDS);
```
PawasChhokra: Defined a lock timeout value in LocalApplicationRunner instead of using ZkConfig.

samza-core/src/main/java/org/apache/samza/zk/ZkLock.java
```diff
@@ -0,0 +1,129 @@
...
+  @Override
+  public boolean hasLock() {
+    return hasLock.get();
+  }
+
+  @Override
+  public void setLockListener(LockListener listener) {
+    this.zkLockListener = listener;
+  }
+
```
PawasChhokra: Removed.

samza-core/src/main/java/org/apache/samza/zk/ZkLock.java
```diff
@@ -0,0 +1,129 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Lock primitive for Zookeeper.
+ */
+public class ZkLock implements Lock {
+
+  public static final Logger LOG = LoggerFactory.getLogger(ZkLock.class);
+  private final static String LOCK_PATH = "lock";
```
PawasChhokra: Renamed to DistributedLock.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
PawasChhokra: Discussed offline.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
PawasChhokra: It seems like we won't be needing this function at all. Hence removed.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -114,4 +114,4 @@ default boolean createStream(StreamSpec streamSpec) {
   default void validateStream(StreamSpec streamSpec) throws StreamValidationException {
     throw new UnsupportedOperationException();
   }
-}
+}
```
navina: nit: can we try to get rid of these minor changes that are seemingly unrelated to the patch?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -222,25 +219,27 @@ public void waitForFinish() {
...
-          });
-        leaderElector.tryBecomeLeader();
-        initLatch.await(LATCH_TIMEOUT_MINUTES, TimeUnit.MINUTES);
+        DistributedLock initLock = coordinationUtils.getLock();
+        boolean hasLock = initLock.lock(LOCK_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        if (hasLock) {
+          getStreamManager().createStreams(intStreams);
+        } else {
+          LOG.error("Timed out while trying to acquire lock");
+          coordinationUtils.reset();
```
navina: Can you wrap lines from Line 253 in a try-finally block? You should invoke reset even in the case where lock has been successfully acquired and critical section is complete. Adding reset in finally block will guarantee this behavior.

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedLock.java
```diff
@@ -0,0 +1,108 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Distributed lock primitive for Zookeeper.
+ */
+public class ZkDistributedLock implements DistributedLock {
+
+  public static final Logger LOG = LoggerFactory.getLogger(ZkDistributedLock.class);
+  private final static String LOCK_PATH = "distributed_lock";
```
navina: Can you make this configurable by making it a constructor argument?
Ideally, this should be unique for different deployments of the streamapplication. Can we make this based on the hashcode of intermediate streams list for now? Check out how @bharathkk  is doing it in PR #265

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedLock.java
```diff
@@ -0,0 +1,108 @@
...
+      if (children.size() == 0 || index == -1) {
+        throw new SamzaException("Looks like we are no longer connected to Zk. Need to reconnect!");
+      }
+      // Acquires lock when the node has the lowest sequence number and returns.
+      if (index == 0) {
+        LOG.info("Acquired lock for participant id: {}", participantId);
+        return true;
+      } else {
+        // Keep trying to acquire the lock
+        long currentTime = System.currentTimeMillis();
```
navina: nit: you can get rid of the local variable `currentTime`

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedLock.java
```diff
@@ -0,0 +1,108 @@
...
+        if ((currentTime - startTime) >= lockTimeout) {
```
navina: Perhaps you can make this the condition for your while loop? 
See if you can accommodate this change. If it doesn't work out, it's fine.

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedLock.java
```diff
@@ -0,0 +1,108 @@
...
+   * Unlocks, by deleting the ephemeral sequential node created to acquire the lock.
+   */
+  @Override
+  public void unlock() {
+    if (nodePath != null) {
+      zkUtils.getZkClient().delete(nodePath);
+      nodePath = null;
+      LOG.info("Ephemeral lock node deleted. Unlocked!");
+    } else {
+      LOG.error("Ephemeral lock node you want to delete doesn't exist");
```
navina: Perhaps change this `warn` since it doesn't actually indicate an error condition.

PawasChhokra: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedLock.java
PawasChhokra: Done.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
PawasChhokra: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedLock.java
PawasChhokra: Made it configurable. I have reused Bharath's code for this. Kindly let me know if it's okay.

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedLock.java
PawasChhokra: Done :)

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
PawasChhokra: Done.

samza-core/src/main/java/org/apache/samza/coordinator/DistributedLock.java
```diff
@@ -0,0 +1,39 @@
...
+   * @param unit Time Unit of the timeout defined above.
+   * @return true if lock is acquired successfully, false if it times out.
+   */
+  boolean lock(long timeout, TimeUnit unit);
+
+  /**
+   * Releases the lock
+   */
+  void unlock();
+}
```
sborya: I think it would be helpful to have a close() method in the lock to be able to close the connection (separately from the unlock)

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -59,9 +58,10 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(LocalApplicationRunner.class);
   private static final String APPLICATION_RUNNER_ZK_PATH_SUFFIX = "/ApplicationRunnerData";
-  // Latch timeout is set to 10 min
-  private static final long LATCH_TIMEOUT_MINUTES = 10;
-  private static final long LEADER_ELECTION_WAIT_TIME_MS = 1000;
+  // Lock timeout is set to 10 seconds here, as we don't want to introduce a new config value currently.
+  private static final long LOCK_TIMEOUT_SECONDS = 10;
+  private static final String ZK_COORDINATION_CLASS = "org.apache.samza.zk.ZkJobCoordinatorFactory";
```
nickpan47: We should avoid more and more ZK-specific names/details to bleed into the general implementation. LocalApplicationRunner is not just for ZK-based implementation. Hence, I would recommend not to hard-code any actual ZK implementation classes here in the LocalApplicationRunner.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -59,9 +58,10 @@
...
+  private static final String ZK_COORDINATION_SERVICE_CLASS = "org.apache.samza.zk.ZkCoordinationServiceFactory";
```
nickpan47: Same here. We should only use the general CoordinatorServiceFactory interface in LocalApplicationRunner, *not* the ZK-specific implementation.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -214,9 +214,9 @@ public void waitForFinish() {
     String jobCoordinatorFactoryClassName = config.get(JobCoordinatorConfig.JOB_COORDINATOR_FACTORY, "");
 
     // TODO: we will need a better way to package the configs with application runner
-    if (ZkJobCoordinatorFactory.class.getName().equals(jobCoordinatorFactoryClassName)) {
+    if (ZK_COORDINATION_CLASS.equals(jobCoordinatorFactoryClassName)) {
       ApplicationConfig appConfig = new ApplicationConfig(config);
-      return new ZkCoordinationServiceFactory().getCoordinationService(
+      return Util.<CoordinationServiceFactory>getObj(ZK_COORDINATION_SERVICE_CLASS).getCoordinationService(
```
nickpan47: Please remove the ZK-specific code here.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -225,31 +225,28 @@ public void waitForFinish() {
...
+   * If {@link CoordinationUtils} is provided, this function will first acquire a lock, and then create the streams.
+   * All the runner processes will either wait till the time they acquire the lock, or timeout after the specified time.
+   * After stream creation, they will unlock and proceed normally.
    * @param planId a unique identifier representing the plan used for coordination purpose
    * @param intStreams list of intermediate {@link StreamSpec}s
-   * @throws TimeoutException exception for latch timeout
+   * @throws TimeoutException exception for lock timeout
    */
-  /* package private */ void createStreams(String planId, List<StreamSpec> intStreams) throws TimeoutException {
+   /* package private */ void createStreams(String planId, List<StreamSpec> intStreams) throws TimeoutException {
```
nickpan47: nit: unnecessary extra leading white space.

samza-core/src/main/java/org/apache/samza/zk/ZkLock.java
```diff
@@ -0,0 +1,102 @@
...
+    //Start timer for timeout
+    long startTime = System.currentTimeMillis();
+    long lockTimeout = TimeUnit.MILLISECONDS.convert(timeout, unit);
+
+    while ((System.currentTimeMillis() - startTime) < lockTimeout) {
+      List<String> children = zkUtils.getZkClient().getChildren(lockPath);
+      int index = children.indexOf(ZkKeyBuilder.parseIdFromPath(nodePath));
+
+      if (children.size() == 0 || index == -1) {
+        throw new SamzaException("Looks like we are no longer connected to Zk. Need to reconnect!");
```
nickpan47: I don't see any error handling logic here (e.g. try to re-connect within the session expiration, etc.). Is it on the contract to the caller of DistributedLock.lock() to handle those exceptions? If yes, it would be necessary to define the exception types the DistributedLock.lock() method would throw and clearly state in the document what the caller supposed to do when catch those exceptions. Otherwise, the implementer of lock() should just handle the specific exceptions internally (i.e. re-try) until timeout.

samza-core/src/main/java/org/apache/samza/zk/ZkLock.java
```diff
@@ -0,0 +1,102 @@
...
+      }
+      // Acquires lock when the node has the lowest sequence number and returns.
+      if (index == 0) {
+        LOG.info("Acquired lock for participant id: {}", participantId);
+        return true;
+      } else {
+        try {
+          Thread.sleep(random.nextInt(1000));
```
nickpan47: I thought that in ZK implementation, we can use watch() on children?

samza-core/src/main/java/org/apache/samza/zk/ZkLock.java
```diff
@@ -0,0 +1,102 @@
...
+    return false;
+  }
+
+  /**
+   * Unlocks, by deleting the ephemeral sequential node created to acquire the lock.
+   */
+  @Override
+  public void unlock() {
+    if (nodePath != null) {
+      zkUtils.getZkClient().delete(nodePath);
```
nickpan47: So, I would recommend to think it through about the zkClient lifecycle needed in this lock implementation. Are we assuming that the zkClient is initialized by some other class and just passed in to this lock implementation? Or are we assuming that the lock would internally own the lifecycle of the zkClient? If we are assuming the first case, I think that this DistributedLock implementation lack of handling on the failure cases: what if the zkClient connection to the ZK server was dropped *after* the user has called lock() and proceeded into the critical section? If the disconnect time is longer than the session timeout, there would be race conditions. How do the users of the DistributedLock being notified of this issue after calling the lock()?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
navina: @nickpan47 I agree. but we have to do this string matching because we decided that the user will not configure a `CoordinationServiceFactory` and that we will derive it from `job.coordinator.factory` configuration. Are you recommending that we introduce another configuration ?

samza-core/src/main/java/org/apache/samza/coordinator/DistributedLock.java
shanthoosh: @nickpan47 

When there’s a failure in unlock, a client of ZkDistributedLock will not be able to distinguish if it’s because of releasing resources or due to unlock implementation. A lock owner wanting to release a lock to other processor in group, will not be able to conclude effectively if unlock has happened or not in case of exceptions(Due to two possibilities and will not be able to handle exception effectively).

Also, logically with just lock, unlock operations a client can call lock() after unlock(), essentially trying to join again which wouldn’t be possible if resource establishment happens in constructor/factory.create method(instead of separate init() lifecycle method). 

Since any DistributedLock implementation will require resource establishment with a RemoteService(Zookeeper, AzureBlobStore, any), extending Autocloseable and having init() lifecycle methods, we would be clearly defining boundaries and communicating that intent with our interface(Similar to `SystemProducer`, `SystemConsumer`). 

I’m not suggesting that we should do it, just interested to know your opinion. 

Can you please point out why this reasoning is incorrect(not required now).

Please share your thoughts.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
nickpan47: @navina I am OK if the config has a default, which is ZK-specific. I am against the hard-code ZK-specific class names in the LocalApplicationRunner implementation. I would suggest that in the LocalApplicationRunner, we load the CoordinatorServiceFactory via Utils.getInstance(JobConfig.getJobCoordinatorFactory()), w/o hard-code the ZK job coordinator class names in the LocalApplicationRunner.

samza-core/src/main/java/org/apache/samza/coordinator/DistributedLock.java
nickpan47: @shanthoosh please see my comments on the lock interface definition. You point here is exactly what I tried to say in the interface definition. You can have the lock itself maintain a heart-beat/life-cycle/lease to the remote resource, which may not require init()/close() in that case. However, it seems necessary to define a callback to the lock interface s.t. user can react to loss of possession of remote resource.

prateekm:
@PawasChhokra Does this still need to be merged? If not, can we close this PR?

PawasChhokra:
@prateekm This has been implemented already. Closing the pull request now. Thanks for the reminder!

PR #266, opened 2017-08-10
SAMZA-1387: Unable to Start Samza App Because Regex Check

jmakes:


jmakes:
@fji @nickpan47 @prateekm take a look. 
This change is almost identical to #213, but for coordinator streams rather than changelog.

fredji97:
There is a check failure, could you please take a look? Thanks!

Fred

jmakes:
@fredji97 Yeah it's a known flaky test called testShouldStartAndRestore

@shanthoosh, you've emailed about this before. Is there a ticket for it already?

PR #268, opened 2017-08-11
SAMZA-1389: Fix ZkProcessorLatch await(timeout, TimeUnit) api.

shanthoosh:
Use passed in timeUnit value for zkClient.waitUnitExists method rather than hardcoding with  `TimeUnit.MILLISECONDS`.


shanthoosh:
@navina @sborya @xinyuiscool

PR #270, opened 2017-08-14
SAMZA-1390: Update SamzaMonitorService to spawn deamon threads

shanthoosh:
Observed in LinkedIn production setup that samza-rest jvm process doesn’t stop after main
thread death(due to jetty server failures) since non-deamon threads spawned for
`SamzaMonitorService` are alive.

This affects samza-rest jvm process lifecycle management. To fix this, plugging
in ThreadFactory which sets Thread name format & marks them as daemon.


shanthoosh:
@jmakes @bringhurst @abhishekshivanna

PR #272, opened 2017-08-15
SAMZA-1392: KafkaSystemProducer performance and correctness with conc…

jmakes:
…urrent sends and flushes


jmakes:
@nickpan47 @prateekm @xinyuiscool @shanthoosh

jmakes:
Folks,
It looks like I've got some tests to fix. But please take a pass through the code in the meantime!

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -210,7 +198,9 @@ class TaskInstance(
 
     trace("Flushing producers for taskName: %s" format taskName)
 
-    collector.flush
+    exceptionHandler.maybeHandle { // Allow users to decide whether to swallow producer exceptions (in kafka whole batch will be lost!)
+      collector.flush
```
xinyuiscool: One question about this: now we wrap the flush with the exception handler, so we expect the user to provide their own to handle this exception? I am wondering what is the steps user needs to do if they want to do something here.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -208,25 +179,30 @@ class KafkaSystemProducer(systemName: String,
...
+        // Throw the original exception if possible. If not, throw exception for null producer.
+        if (exception != null) {
+          throw new SystemProducerException("Flush failed. A batch of messages were unsent!", exception)
         } else {
-          trace("Flushed %s." format (source))
+          throw new SystemProducerException("Kafka system producer is not available.")
         }
       }
+
+      currentProducer.flush()
```
prateekm: It's possible that the producer is closed-but-not-null here (same reason why line 191 can happen). Maybe try-catch and rethrow as SystemProducerException?

prateekm: It's possible that we have a saved exception and a working new Producer here. We can probably pre-check for exception and throw before trying to flush the producer to avoid out-of-order delivery during flush. Doesn't prevent all OOO delivery (see comment above), but won't hurt. It does look awkward, but refactoring might help.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -98,14 +80,13 @@ class KafkaSystemProducer(systemName: String,
 
   def register(source: String) {
     if(sources.putIfAbsent(source, new SourceData) != null) {
-      throw new SamzaException("%s is already registered with the %s system producer" format (source, systemName))
+      throw new SystemProducerException("%s is already registered with the %s system producer" format (source, systemName))
     }
   }
 
   def closeAndNullifyCurrentProducer(currentProducer: Producer[Array[Byte], Array[Byte]]) {
```
prateekm: Minor, Unrelated to PR: Personally prefer placing private/helper methods after public methods (or methods that use them). +1 for updating if you agree ;)

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -47,22 +39,13 @@ class KafkaSystemProducer(systemName: String,
...
-    var latestFuture: Future[RecordMetadata] = null
-    /**
      * exceptionInCallback: to store the exception in case of any "ultimate" send failure (ie. failure
      * after exhausting max_retries in Kafka producer) in the I/O thread, we do not continue to queue up more send
      * requests from the samza thread. It helps the samza thread identify if the failure happened in I/O thread or not.
      *
      * In cases of multiple exceptions in the callbacks, we keep the first one before throwing.
      */
-    var exceptionInCallback: AtomicReference[SamzaException] = new AtomicReference[SamzaException]()
+    var exceptionInCallback: AtomicReference[SystemProducerException] = new AtomicReference[SystemProducerException]()
```
prateekm: QQ: Does this need to be a var? Maybe make a val to avoid accidentally nulling the reference instead of calling getAndSet(null).

samza-kafka/src/test/java/org/apache/samza/system/kafka/MockKafkaProducer.java
```diff
@@ -253,4 +233,32 @@ public boolean isDone() {
       return true;
     }
   }
+
+  private class FlushRunnable implements Runnable {
```
nickpan47: +1! Like this pattern!

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemProducer.scala
```diff
@@ -168,21 +165,42 @@ class TestKafkaSystemProducer {
...
     producer.stop()
   }
 
+  /**
+    * If there's an exception, we should:
+    * 1. Close the producer (from the one-and-only kafka send thread) to prevent subsequent sends from going out of order.
+    * 2. Nullify the producer to cause it to be recreated
+    * 3. Throw the exception from systemProducer.flush() to prevent a checkpoint
+    *
+    * Assumptions:
```
nickpan47: +1. This is really good documentation! 👍

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
jmakes: Good question. 

If they want to ignore producer exceptions (and understand that they will drop whole batches of output messages if they do), they just add the config:
task.ignored.exceptions=org.apache.samza.system.SystemProducerException

See the following unit test for an example:
org.apache.samza.container.TestTaskInstance#testProducerExceptionsCanBeIgnored

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -320,14 +316,17 @@ class KafkaProducerConfig(val systemName: String,
...
-      && producerProperties.get(ProducerConfig.RETRIES_CONFIG).asInstanceOf[String].toInt < RETRIES_DEFAULT) {
-      warn("Samza does not provide producer failure handling. Consider setting '%s' to a large value, like Int.MAX." format ProducerConfig.RETRIES_CONFIG)
-    } else {
-      // Retries config is set to Max so that when all attempts fail, Samza also fails the send. We do not have any special handler
-      // for producer failure
+    if (!producerProperties.containsKey(ProducerConfig.RETRIES_CONFIG)) {
+      debug("%s undefined. Defaulting to %s." format(ProducerConfig.RETRIES_CONFIG, RETRIES_DEFAULT))
       producerProperties.put(ProducerConfig.RETRIES_CONFIG, RETRIES_DEFAULT)
     }
+    producerProperties.get(ProducerConfig.RETRIES_CONFIG).toString.toInt // Verify int
```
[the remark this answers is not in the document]
jmakes: We only explicitly put the default in the map. In order to check the user-provided values, we'd have to filter the properties for this config and then check it if it exists. 
Keeping the code here, while a bit clunky, allows us to:
1. Keep the related checks (default and type) for these properties adjacent in code
2. Check the type after we already know that the value is defined so we don't have to check for existence. 

That said, I'm open to alternative approaches if they can be done cleanly.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -78,17 +61,16 @@ class KafkaSystemProducer(systemName: String,
...
             producer = null
           }
-        }
-        currentProducer.close
 
-        sources.asScala.foreach {p =>
-          if (p._2.exceptionInCallback.get() == null) {
-            flush(p._1)
-          }
+          currentProducer.close
```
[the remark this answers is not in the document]
jmakes: Agreed. Will address this in the next commit for this PR

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -130,12 +111,6 @@ class KafkaSystemProducer(systemName: String,
...
 
-    val exception = sourceData.exceptionInCallback.getAndSet(null)
-    if (exception != null) {
-      metrics.sendFailed.inc
-      throw exception  // in case the caller catches all exceptions and will try again
-    }
-
     // lazy initialization of the producer
     if (producer == null) {
       producerLock.synchronized {
```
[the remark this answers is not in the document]
jmakes: Great find. Thank you.
Take a look at the latest and let me know if you still see issues like this.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -21,99 +21,71 @@ package org.apache.samza.system.kafka
...
-     */
-    var exceptionInCallback: AtomicReference[SamzaException] = new AtomicReference[SamzaException]()
-  }
-
-  @volatile var producer: Producer[Array[Byte], Array[Byte]] = null
+  /**
+    * Represents a fatal error that caused the producer to close.
+    */
+  val fatalException: AtomicReference[SystemProducerException] = new AtomicReference[SystemProducerException]()
+  var producer: Producer[Array[Byte], Array[Byte]] = null
```
prateekm: IIUC this needs to be volatile (or its reads need to be synchronized on producerLock too). Otherwise sender/flushing threads aren't guaranteed to see the new producer created by the Kafka I/O thread in the callback (i.e., there's no memory barrier).

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -125,31 +97,17 @@ class KafkaSystemProducer(systemName: String,
       throw new IllegalArgumentException(StreamNameNullOrEmptyErrorMsg)
     }
 
-    val sourceData = sources.get(source)
-    if (sourceData == null) {
+    if (!sources.contains(source)) {
```
prateekm: Just wondering: This happens on every send. Does this still provide value besides a defensive check against our (i.e. framework devs as opposed to user) coding errors?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +119,96 @@ class KafkaSystemProducer(systemName: String,
...
+      // Close producer to ensure messages queued in-flight are not sent and hence, avoid re-ordering
+      // This works because the callback thread is singular and no sends can complete until the callback returns.
+      if (isFatalException) {
+        producer.close(0, TimeUnit.MILLISECONDS)
+        fatalException.compareAndSet(null, producerException)
+      }
+    }
+  }
+
+  private def isFatalException(exception: Exception): Boolean = {
```
prateekm: SerializationException is only thrown synchronously on producer.send. I think IllegalState and ProducerClosed exceptions should also be considered fatal exceptions here. If that's true, it'll be cleaner to special case the catch for SerializationException, and remove this method and the isFatalException parameter from handleSendException.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +119,96 @@ class KafkaSystemProducer(systemName: String,
...
+            producer = getProducer()
+          }
         }
       }
+    } else {
+      // If there is an exception in the callback, it means that the Kafka producer has exhausted the max-retries
+      // Close producer to ensure messages queued in-flight are not sent and hence, avoid re-ordering
+      // This works because the callback thread is singular and no sends can complete until the callback returns.
+      if (isFatalException) {
+        producer.close(0, TimeUnit.MILLISECONDS)
```
prateekm: Producer#close can throw. We should probably try/catch/finally here to guarantee fatal exception is raised in flush.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +119,96 @@ class KafkaSystemProducer(systemName: String,
...
+    if (dropProducerExceptions) {
+      warn("Ignoring producer exception. All messages in the failed produce request will be dropped!", producerException)
+
+      if (isFatalException) {
+        producerLock.synchronized {
+          // Prevent each callback from recreating producer for the same failed event.
+          if (currentProducer == producer) {
+            info("Creating a new producer for system %s." format systemName)
+            producer = getProducer()
+          }
```
prateekm: Do we need to close currentProducer after recreating a new one, or are we intentionally avoiding it to prevent some out of order delivery scenarios?

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -210,7 +207,9 @@ class TaskInstance(
 
     trace("Flushing producers for taskName: %s" format taskName)
 
-    collector.flush
+    exceptionHandler.maybeHandle { // Allow users to decide whether to swallow producer exceptions
```
prateekm: Do we still need this if we're adding the top level task.drop.producer.error config?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -21,99 +21,71 @@ package org.apache.samza.system.kafka
...
-            // only nullify the member producer if it is still the same object, no point nullifying new producer
-            producer = null
-          }
+    info("Stopping producer for system: " + this.systemName)
+
+    // stop() should not happen often so no need to optimize locking
+    producerLock.synchronized {
+      try {
+        if (producer != null) {
+          producer.close
```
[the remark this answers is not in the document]
jmakes: No, close() includes a flush as long as it's not a forceful (timeout==0) close.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -21,99 +21,71 @@ package org.apache.samza.system.kafka
...
+    */
+  val fatalException: AtomicReference[SystemProducerException] = new AtomicReference[SystemProducerException]()
+  var producer: Producer[Array[Byte], Array[Byte]] = null
   var producerLock: Object = new Object
   val StreamNameNullOrEmptyErrorMsg = "Stream Name should be specified in the stream configuration file."
-  val sources: ConcurrentHashMap[String, SourceData] = new ConcurrentHashMap[String, SourceData]
+  val sources: mutable.Set[String] = new mutable.HashSet[String]
 
   def start(): Unit = {
+    producer = getProducer()
```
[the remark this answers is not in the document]
jmakes: Great question. 

Both start() and stop() are called once per producer by SystemProducers. The difference is start() is always guaranteed to occur in a clean state before any send() or flush() calls. It's harder to make a similar "clean state" guarantee for stop(). 

It wouldn't hurt to add the lock to start(). But it would mostly be for symmetry, as I can't imagine it ever providing value.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
jmakes: Another great question. I never bothered to look into the history to see whether this check was to enforce the proper state transitions or if it was just to make sure the map is prepopulated in a synchronous way. Looks like it provides no value now. I'll remove it!

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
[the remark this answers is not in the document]
prateekm: I see. I remembered the behavior for when it's called from the I/O thread (is a force close). Maybe worth a comment saying that this flushes too.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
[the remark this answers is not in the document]
prateekm: Makes sense, stop isn't guaranteed to be clean. Let's leave as is.

jmakes:
@prateekm can you take another look? 

I feel pretty good about the change now. 
Fault testing has shown proper behavior with and without task.drop.producer.errors enabled for both user thread and callback thread exceptions. 

Performance testing shows a minor improvement for single threaded and compression disabled. It shows a big improvement for concurrency with compression enabled. I'll share some numbers on the JIRA tomorrow.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -97,7 +97,8 @@ class KafkaSystemFactory extends SystemFactory with Logging {
       systemName,
       new ExponentialSleepStrategy(initialDelayMs = producerConfig.reconnectIntervalMs),
       getProducer,
-      metrics)
+      metrics,
+      dropProducerExceptions = config.getDropProducerError)
```
prateekm: Minor: can remove the parameter name.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -21,134 +21,100 @@ package org.apache.samza.system.kafka
...
+import org.apache.samza.system.SystemProducerException
 import org.apache.samza.util.ExponentialSleepStrategy
 import org.apache.samza.util.KafkaUtil
 import org.apache.samza.util.Logging
 import org.apache.samza.util.TimerUtils
 
-import scala.collection.JavaConverters._
+object KafkaSystemProducer {
+  // Send messages
+  val SendMsgFormat = "Enqueuing message: %s, %s."
```
prateekm: Strongly prefer keeping these inline. These log/exception messages also serve as inline method documentation, and IMO it's more readable to avoid the indirection.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -21,134 +21,100 @@ package org.apache.samza.system.kafka
...
+  val StreamNameNullOrEmptyErrorMsg = "Stream Name should be specified in the stream configuration file."
```
prateekm: Not from this PR, but I think this particular message is inaccurate since output streams aren't specified in configs.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -21,134 +21,100 @@ package org.apache.samza.system.kafka
...
+  val FailedSendMsgFormat = "Failed to send message for Source: %s on System:%s Topic:%s Partition:%s"
+  val DropErrorMsg = "Ignoring producer exception. All messages in the failed producer request will be dropped!"
+  val NewProducerMsgFormat = "Creating a new producer for system %s."
+
+  // Flush messages
+  val FailedFlushMsg = "Flush failed. One or more batches of messages were not sent!"
+
+  // Stop messages
+  val StopMsg = "Stopping producer for system: "
+  val UnhandledSendErrorMsg = "Observed an earlier send() error while closing producer"
```
prateekm: Minor: Maybe s/earlier/unhandled?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -21,134 +21,100 @@ package org.apache.samza.system.kafka
...
+        if (producer != null) {
+          producer.close  // Also performs the equivalent of a flush()
         }
-        currentProducer.close
 
-        sources.asScala.foreach {p =>
-          if (p._2.exceptionInCallback.get() == null) {
-            flush(p._1)
-          }
+        // Log error for debugging purposes.
```
prateekm: Minor: can remove.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -21,134 +21,100 @@ package org.apache.samza.system.kafka
...
+        val exception = fatalException.get()
+        if (exception != null) {
+          error(KafkaSystemProducer.UnhandledSendErrorMsg, exception)
         }
+      } catch {
+        case e: Exception => error(KafkaSystemProducer.ErrorClosingMsg + systemName, e)
+      } finally {
+        producer = null
```
prateekm: This is the only place where producer is nullified. Might be clearer to update the null check message in send/flush to say something like "send/flush called after producer close".

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -21,134 +21,100 @@ package org.apache.samza.system.kafka
...
-          producer = getProducer()
-        }
-      }
+      throw new SystemProducerException(KafkaSystemProducer.PreviousErrorMsg, globalProducerException)
     }
 
     val currentProducer = producer
     if (currentProducer == null) {
-      throw new SamzaException("Kafka system producer is not available.")
+      throw new SystemProducerException(KafkaSystemProducer.NullProducerMsg)
```
prateekm: Need sendFailed.inc?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +127,102 @@ class KafkaSystemProducer(systemName: String,
...
-          })
-        sourceData.latestFuture = futureRef
-      }
+      currentProducer.send(record, new Callback {
+        def onCompletion(metadata: RecordMetadata, exception: Exception): Unit = {
+          if (exception == null) {
+            //send was successful.
+            metrics.sendSuccess.inc
+          }
+          else {
```
prateekm: Code style: Previous line.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +127,102 @@ class KafkaSystemProducer(systemName: String,
...
-                             exception)
-              }
-            }
-          })
-        sourceData.latestFuture = futureRef
-      }
+      currentProducer.send(record, new Callback {
+        def onCompletion(metadata: RecordMetadata, exception: Exception): Unit = {
+          if (exception == null) {
+            //send was successful.
```
prateekm: Minor: remove comment.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +127,102 @@ class KafkaSystemProducer(systemName: String,
...
+          }
+        }
+      })
       metrics.sends.inc
     } catch {
-      case e: Exception => {
-        error("Closing the producer because of an exception in send: ", e)
-
-        closeAndNullifyCurrentProducer(currentProducer)
+      case e : Exception =>
```
prateekm: Minor: s/e/originalException (clarifies that isFatalException is on originalException as opposed to producerException.)

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +127,102 @@ class KafkaSystemProducer(systemName: String,
...
+      }
 
-        //if there is an exception thrown from the previous callbacks just before flush, we have to fail the container
-        if (sourceData.exceptionInCallback.get() != null) {
-          metrics.flushFailed.inc
-          throw sourceData.exceptionInCallback.get()
-        } else {
-          trace("Flushed %s." format (source))
+      // Flush only throws InterruptedException, all other errors are handled in send() callbacks
+      producer.flush()
```
prateekm: 1. currentProducer = producer to avoid NPE w/ concurrent stop.
2. Wondering if it's ok to flush a closed producer?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +127,102 @@ class KafkaSystemProducer(systemName: String,
...
+      }
+      trace("Flushed %s." format source)
+    }
+  }
+
+
+  private def handleSendException(currentProducer: Producer[Array[Byte], Array[Byte]], producerException: SystemProducerException, isFatalException: Boolean) = {
+    metrics.sendFailed.inc
+    error(producerException)
+    // The SystemProducer contract is synchronous, so there's no way for us to guarantee that an exception will
```
prateekm: Minor: s/contract/send API

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +127,102 @@ class KafkaSystemProducer(systemName: String,
...
+    // be handled by the Task before we recreate the producer, and if it isn't handled, a concurrent send() from another
+    // source could send on the new producer before handling its own exceptions and produce out of order messages.
```
prateekm: Minor: What is "its own exceptions" referring to?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +127,102 @@ class KafkaSystemProducer(systemName: String,
...
+    // So we have to handle it right here in the SystemProducer.
+    if (dropProducerExceptions) {
+      warn(KafkaSystemProducer.DropErrorMsg, producerException)
```
prateekm: This can potentially create a large amount of logs when the Kafka cluster being written to is unavailable.
We should consider adding a throttling logger for messages like these, but for now we can remove the producerException from the message (it's already logged as error above).

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +127,102 @@ class KafkaSystemProducer(systemName: String,
...
+
+      if (isFatalException) {
+        producerLock.synchronized {
+          // Prevent each callback from recreating producer for the same failed event.
+          if (currentProducer == producer) {
+            info(KafkaSystemProducer.NewProducerMsgFormat format systemName)
+            currentProducer.close(0, TimeUnit.MILLISECONDS)
```
prateekm: I think we need a try/catch around producer.close().

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +127,102 @@ class KafkaSystemProducer(systemName: String,
...
+          }
         }
       }
+    } else {
+      // If there is an exception in the callback, it means that the Kafka producer has exhausted the max-retries
+      // Close producer to ensure messages queued in-flight are not sent and hence, avoid re-ordering
+      // This works because the callback thread is singular and no sends can complete until the callback returns.
+      if (isFatalException) {
+        fatalException.compareAndSet(null, producerException)
+        producer.close(0, TimeUnit.MILLISECONDS)
```
prateekm: I think we need a try/catch around producer.close().

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +127,102 @@ class KafkaSystemProducer(systemName: String,
...
+            info(KafkaSystemProducer.NewProducerMsgFormat format systemName)
+            currentProducer.close(0, TimeUnit.MILLISECONDS)
+            producer = getProducer()
+          }
         }
       }
+    } else {
+      // If there is an exception in the callback, it means that the Kafka producer has exhausted the max-retries
+      // Close producer to ensure messages queued in-flight are not sent and hence, avoid re-ordering
+      // This works because the callback thread is singular and no sends can complete until the callback returns.
```
prateekm: Minor: s/callback thread is singular/there is 1 callback thread.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +127,102 @@ class KafkaSystemProducer(systemName: String,
...
+      if (isFatalException) {
+        fatalException.compareAndSet(null, producerException)
+        producer.close(0, TimeUnit.MILLISECONDS)
+      }
+    }
+  }
+
+  // A fatal exception is one that corrupts the producer or otherwise makes it unusable.
```
prateekm: Minor: Javadoc format (/** .... */).

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +127,102 @@ class KafkaSystemProducer(systemName: String,
...
+  // We want to handle non-fatal exceptions differently because they can often be handled by the user
+  // and that's preferable because it allows users that drop exceptions a way to do that with less
```
prateekm: Minor: s/allows/gives

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -161,72 +127,102 @@ class KafkaSystemProducer(systemName: String,
...
+    // The SystemProducer contract is synchronous, so there's no way for us to guarantee that an exception will
+    // be handled by the Task before we recreate the producer, and if it isn't handled, a concurrent send() from another
+    // source could send on the new producer before handling its own exceptions and produce out of order messages.
+    // So we have to handle it right here in the SystemProducer.
+    if (dropProducerExceptions) {
+      warn(KafkaSystemProducer.DropErrorMsg, producerException)
+
+      if (isFatalException) {
+        producerLock.synchronized {
+          // Prevent each callback from recreating producer for the same failed event.
```
prateekm: Minor: s/failed/failure

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -663,6 +663,21 @@
                 </tr>
 
                 <tr>
+                    <td class="property" id="task-drop-producer-errors">task.drop.producer.errors</td>
+                    <td class="default">false</td>
+                    <td class="description">
+                        If true, producer errors will be logged and ignored. If false, producer errors will be propagated
```
prateekm: Should clarify that this depends on SystemProducer implementation.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -663,6 +663,21 @@
...
+                        upward, which will usually fail the container. Failing the container is a safety precaution to
```
prateekm: "upward, either to the task, or to the run loop, in which case it will fail the container."

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -663,6 +663,21 @@
...
+                        ensure the latest checkpoints only reflect the events that have been completely and successfully
+                        processed. However, some applications prefer to remain running at all costs, even if that means
+                        lost messages. Setting this property to true will enable applications to recover from producer
+                        errors at the expense of one or many (in the case of batching producers) dropped messages.
```
prateekm: Should clarify that this needs to be done in conjunction with (i.e. is not a replacement for) catching exceptions in collector.send().

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -663,6 +663,21 @@
...
+                        If you enable this, it is highly recommended that you configure alerting on the
+                        'producer-send-failed' metric.
```
prateekm: ".. metric, since the producer might drop messages indefinitely"

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
jmakes: IIUC then the config would be used as the clock parameter for KafkaSystemProducer. Doesn't compile when I tried it.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
jmakes: Agree. Fixed.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
jmakes: That wouldn't always be accurate, since the exception may have been thrown to the sender many times (and swallowed by user code) before finally propagating up to the container from flush(). In that case, we're just logging the error (which the application has already seen and ignored) on shutdown for troubleshooting purposes.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
jmakes: Could also happen if send/flush were called before start(). 

Either way, it's likely a samza producer lifecycle bug. I'm not sure getting more descriptive will help much.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
jmakes: I thought about doing that, but according to the doc in KafkaSystemProducerMetrics, this metric is tracking failures in the KafkaProducer as opposed to "sends" which is tracking calls to the KafkaSystemProducer. I didn't want to change the semantics without good reason. 

Thoughts?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
jmakes: 2. Yep it's fine. I checked with the kafka team.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
jmakes: Phrase came from the version where we were tracking exceptions per source. Clarified it for this patch. 
Thanks!

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
jmakes: Ok, never can be too safe in an error handling routine.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
jmakes: Done

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemProducer.scala
```diff
@@ -165,52 +162,383 @@ class TestKafkaSystemProducer {
...
   }
 
   @Test
   def testKafkaProducerFlushMsgsWhenStop {
-    val msg1 = new OutgoingMessageEnvelope(new SystemStream("test", "test"), "a".getBytes)
-    val msg2 = new OutgoingMessageEnvelope(new SystemStream("test", "test"), "b".getBytes)
-    val msg3 = new OutgoingMessageEnvelope(new SystemStream("test", "test"), "c".getBytes)
+    val msg1 = new OutgoingMessageEnvelope(systemStream, "a".getBytes)
+    val msg2 = new OutgoingMessageEnvelope(systemStream, "b".getBytes)
+    val msg3 = new OutgoingMessageEnvelope(systemStream, "c".getBytes)
```
nickpan47: It would be nice to use the startDelayedSendThread mechanism to test race conditions between flush() w/ error against send() as well.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -320,14 +316,17 @@ class KafkaProducerConfig(val systemName: String,
       producerProperties.put(ProducerConfig.MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION, MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION_DEFAULT)
     }
 
-    if (producerProperties.containsKey(ProducerConfig.RETRIES_CONFIG)
-      && producerProperties.get(ProducerConfig.RETRIES_CONFIG).asInstanceOf[String].toInt < RETRIES_DEFAULT) {
-      warn("Samza does not provide producer failure handling. Consider setting '%s' to a large value, like Int.MAX." format ProducerConfig.RETRIES_CONFIG)
-    } else {
-      // Retries config is set to Max so that when all attempts fail, Samza also fails the send. We do not have any special handler
-      // for producer failure
+    if (!producerProperties.containsKey(ProducerConfig.RETRIES_CONFIG)) {
```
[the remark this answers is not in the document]
jmakes: Checked with the Kafka team. Your understanding is correct, it will ultimately fail with expired batches after a couple mins. 
They said the behavior will be roughly equivalent, since the batch will likely expire before 20 retries or INT_MAX retries. 
To err on the side of caution, I'll change it back to the previous value

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemProducer.scala
jmakes: The tests above test this case but with more determinism. It's part of the reason they're so long. :-)

PR #274, opened 2017-08-16
SAMZA-1396 TestZkLocalApplicationRunner tests fails after SAMZA-1385 (release blocker)

navina:
* Fixes ZkPath issues
* Fixes appname / jobname mismatch


navina:
@bharathkk @shanthoosh @nickpan47 @sborya  : Please take a look at this fix.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorFactory.java
```diff
@@ -60,7 +60,7 @@ private ZkUtils getZkUtils(Config config, MetricsRegistry metricsRegistry) {
     return new ZkUtils(keyBuilder, zkClient, zkConfig.getZkConnectionTimeoutMs(), metricsRegistry);
   }
 
-  private String getJobCoordinationZkPath(Config config) {
+  public static String getJobCoordinationZkPath(Config config) {
```
bharathkk: imo, this feels like a utility method and should be in its own class.
I am fine the way it is right now to unblock release

PR #275, opened 2017-08-16
SAMZA-1397: log.debug loop to run only if debug is enabled

sborya:


samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -109,8 +109,10 @@ public void register(String sourceSuffix) {
     }
     containerToHostMapping = Collections.unmodifiableMap(allMappings);
 
-    for (Map.Entry<String, Map<String, String>> entry : containerToHostMapping.entrySet()) {
-      log.debug(String.format("Locality for container %s: %s", entry.getKey(), entry.getValue()));
+    if (log.isDebugEnabled()) {
```
vjagadish1989: @sborya : Is this patch purely for efficiency?

PR #277, opened 2017-08-17
SAMZA-1386: Inline End-of-stream and Watermark logic inside OperatorImpl

xinyuiscool:
This patch contains the following changes:
1. Refactor watermark and end-of-stream logic. The aggregation/handling has been moved from WatermarkManager/EndOfStreamManager to be inline inside OperatorImpl. This is for keeping the logic in one place.
2. Now subclass of OperatorImpl will override handleWatermark() to do its specific handling, such as fire trigger.
3. Add emitWatermark() in OperatorImpl so subclass can call it to emit watermark upon receiving a message or watermark.


xinyuiscool:
@cpettitt-linkedin and @nickpan47 : please take a look. Thanks!

samza-core/src/main/java/org/apache/samza/operators/impl/WatermarkStates.java
```diff
@@ -0,0 +1,113 @@
...
+    boolean updated = state.update(watermarkMessage.getTimestamp(), watermarkMessage.getTaskName());
+
+    if (updated) {
+      long minTimestamp = watermarkStates.entrySet().stream()
+          .filter(entry -> entry.getKey().getSystemStream().equals(ssp.getSystemStream()))
+          .map(entry -> entry.getValue().getWatermarkTime())
+          .min(Long::compare)
+          .get();
+      Long curWatermark = watermarks.get(ssp.getSystemStream());
+      if (curWatermark == null || curWatermark < minTimestamp) {
```
cpettitt-linkedin: I believe the first condition would never hold in the first implementation. In that case it would seem clearer to assert that.

samza-core/src/main/java/org/apache/samza/operators/impl/WatermarkStates.java
```diff
@@ -0,0 +1,113 @@
...
+
+      return preWatermarkTime != watermarkTime;
+    }
+
+    long getWatermarkTime() {
+      return watermarkTime;
+    }
+  }
+
+  private final Map<SystemStreamPartition, WatermarkState> watermarkStates;
```
cpettitt-linkedin: Same comment as above re. thread-safety.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -128,6 +150,17 @@ public final void onMessage(M message, MessageCollector collector, TaskCoordinat
     this.handleMessageNs.update(endNs - startNs);
 
     results.forEach(rm -> this.registeredOperators.forEach(op -> op.onMessage(rm, collector, coordinator)));
+
+    InitableFunction transformFn = getOperatorSpec().getTransformFn();
+    if (transformFn != null && transformFn instanceof WatermarkFunction) {
```
cpettitt-linkedin: First condition is redundant with the second.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -128,6 +150,17 @@ public final void onMessage(M message, MessageCollector collector, TaskCoordinat
...
+      // check whether there is new watermark emitted from the user function
+      long outputWm = ((WatermarkFunction) transformFn).getOutputWatermark();
+      if (outputWatermark < outputWm) {
```
cpettitt-linkedin: I'd probably add a warn to the log if outputWatermark > outputWm. It is OK for outputWatermark == outputWm and the current behavior is correct.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -174,68 +207,139 @@ public final void onTimer(MessageCollector collector, TaskCoordinator coordinato
...
-      if (outputWatermarkTime < outputWatermarkMin) {
-        // populate the watermark to downstream
-        outputWatermarkTime = outputWatermarkMin;
-        Watermark outputWatermark = watermark.copyWithTimestamp(outputWatermarkTime);
+      inputWatermark = inputWatermarkMin;
+      LOG.trace("Advance input watermark to {} in operator {}", inputWatermark, getOperatorName());
+
+      final long outputWm;
+      InitableFunction transformFn = getOperatorSpec().getTransformFn();
+      if (transformFn != null && transformFn instanceof WatermarkFunction) {
```
cpettitt-linkedin: The second condition implies the first.

samza-core/src/main/java/org/apache/samza/operators/impl/WatermarkStates.java
```diff
@@ -0,0 +1,113 @@
...
+
+    boolean update(long timestamp, String taskName) {
+      final long preWatermarkTime = watermarkTime;
+      if (taskName != null) {
+        timestamps.put(taskName, timestamp);
+      }
+
+      if (timestamps.size() == expectedTotal) {
+        Optional<Long> min = timestamps.values().stream().min(Long::compare);
+        watermarkTime = min.orElse(timestamp);
```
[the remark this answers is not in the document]
xinyuiscool: Exactly. For a source, the expectedTotal is 0. Added the documentation.

samza-core/src/main/java/org/apache/samza/operators/impl/WatermarkStates.java
xinyuiscool: Added the doc on the WatermarkStates class for thread-safety.

samza-core/src/main/java/org/apache/samza/operators/impl/WatermarkStates.java
xinyuiscool: That's correct. I added the assert.

samza-core/src/main/java/org/apache/samza/operators/impl/WatermarkStates.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Set;
+import org.apache.samza.system.SystemStream;
+import org.apache.samza.system.SystemStreamPartition;
+import org.apache.samza.system.WatermarkMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class manages the state of the watermarks in a task. Internally it keeps track of the latest
+ * watermark timestamp from each upstream task, and use the min as the consolidated watermark time.
```
nickpan47: This is a bit confusing. The WatermarkState is per task, while inputWatermark/outputWatermark are local variables in each operatorImpl. So, can I assume that the logic is:
1) the first input operator impl invokes aggregateWatermark() method and records the current min watermark per ssp
2) based on the current min watermark per ssp, the onWatermark() method checks the corresponding operatorImpl's inputWatermark and tries to advance watermark if current min watermark per ssp > the operator's inputWatermark
3) the operatorImpl may generate outputWatermark and propagate it downstream, which is used by the downstream operatorImpl to update their inputWatermark and generate outputWatermark
So, at the end, when the watermarks are propagated at the boundary of a task, can we assume that the outputWatermark at the boundary is >= the shared WatermarkState per ssp in the task? I would suggest to add an example with numbers to explain the relationship between the WatermarkState here in the task, and the inputWatermark/outputWatermark as local variables in the OperatorImpl.

samza-core/src/test/java/org/apache/samza/operators/impl/TestOperatorImplGraph.java
```diff
@@ -255,4 +272,134 @@ public Object apply(Object message) {
...
+    MessageStream m2 = streamGraph.getInputStream("input2", msgBuilder).filter(m -> true);
+    MessageStream m3 = streamGraph.getInputStream("input3", msgBuilder).filter(m -> true).partitionBy(m -> "hehe").map(m -> m);
+    Function mockFn = mock(Function.class);
+    OutputStream<Object, Object, Object> om1 = streamGraph.getOutputStream("output1", mockFn, mockFn);
+    OutputStream<Object, Object, Object> om2 = streamGraph.getOutputStream("output2", mockFn, mockFn);
+
+    m1.join(m2, mock(JoinFunction.class), Duration.ofHours(2)).partitionBy(m -> "haha").sendTo(om1);
+    m3.join(m2, mock(JoinFunction.class), Duration.ofHours(1)).sendTo(om2);
+
+    Multimap<SystemStream, SystemStream> outputToInput = OperatorImplGraph.getIntermediateToInputStreams(streamGraph);
```
nickpan47: The method name here is a bit hard to follow. Is the function trying to get the map that keyed by intermediate stream and the value is a set of input topics linking to it through a single job? Wouldn't it be better named as getIntermediateToInputStreamsMap()? And, why is it limited to intermediate topics as keys? I think in general, this would be getOutputToInputStreamsMap().

samza-core/src/test/java/org/apache/samza/operators/impl/TestOperatorImplGraph.java
```diff
@@ -255,4 +272,134 @@ public Object apply(Object message) {
...
+
+  @Test
+  public void testGetOutputToInputStreams() {
+    Map<String, String> configMap = new HashMap<>();
+    configMap.put(JobConfig.JOB_NAME(), "test-app");
+    configMap.put(JobConfig.JOB_DEFAULT_SYSTEM(), "test-system");
+    Config config = new MapConfig(configMap);
+
+    /**
+     * the graph looks like the following. number of partitions in parentheses. quotes indicate expected value.
```
nickpan47: Like this! 👍

samza-core/src/test/java/org/apache/samza/operators/impl/TestOperatorImplGraph.java
```diff
@@ -255,4 +272,134 @@ public Object apply(Object message) {
...
+    assertTrue(inputs.contains(input1.toSystemStream()));
+    assertTrue(inputs.contains(input2.toSystemStream()));
+
+    inputs = outputToInput.get(int2.toSystemStream());
+    assertEquals(inputs.size(), 1);
+    assertEquals(inputs.iterator().next(), input3.toSystemStream());
+  }
+
+  @Test
+  public void testGetProducerTaskCountForIntermediateStreams() {
```
nickpan47: Can we have a similar asc ii graph for this test case as well? Thanks a lot in advance!

samza-core/src/test/java/org/apache/samza/operators/impl/TestOperatorImplGraph.java
```diff
@@ -255,4 +272,134 @@ public Object apply(Object message) {
...
+    intermediateToInputStreams.put(int1, input1);
+    intermediateToInputStreams.put(int1, input2);
+    intermediateToInputStreams.put(int1, input3);
+
+    intermediateToInputStreams.put(int2, input2);
+    intermediateToInputStreams.put(int2, input3);
+
+    Map<SystemStream, Integer> counts = OperatorImplGraph.getProducerTaskCountForIntermediateStreams(
+        streamToConsumerTasks, intermediateToInputStreams);
+    assertTrue(counts.get(int1) == 3);
```
nickpan47: I think that a more illustrative test case could have intermediateToInputStreams only has two input streams (input1 and input2) mapping to int1. Thus, streamToConsumerTasks should still be counts.get(int1) == 3 and counts.get(int2) ==2, even intermediateToInputStreams.get(int1).size()==2

samza-core/src/test/java/org/apache/samza/operators/impl/TestWatermarkStates.java
```diff
@@ -0,0 +1,101 @@
...
+    Set<SystemStreamPartition> ssps = new HashSet<>();
+    SystemStreamPartition inputPartition0 = new SystemStreamPartition(input, new Partition(0));
+    SystemStreamPartition intPartition0 = new SystemStreamPartition(intermediate, new Partition(0));
+    SystemStreamPartition intPartition1 = new SystemStreamPartition(intermediate, new Partition(1));
+    ssps.add(inputPartition0);
+    ssps.add(intPartition0);
+    ssps.add(intPartition1);
+
+    Map<SystemStream, Integer> producerCounts = new HashMap<>();
+    producerCounts.put(intermediate, 2);
```
nickpan47: Same question as in EoS test cases, should it be (intermediate, 1) for two different tasks, instead of (intermediate, 2) for both tasks?

samza-core/src/test/java/org/apache/samza/operators/impl/TestWatermarkStates.java
```diff
@@ -0,0 +1,101 @@
...
+    assertEquals(watermarkStates.getWatermarkPerSSP(intPartition1), WATERMARK_NOT_EXIST);
+    assertEquals(watermarkStates.getWatermark(intermediate), WATERMARK_NOT_EXIST);
+
+    watermarkMessage = new WatermarkMessage(4L, "task 1");
+    updated = watermarkStates.update(watermarkMessage, intPartition1);
+    assertTrue(updated);
+    assertEquals(watermarkStates.getWatermarkPerSSP(intPartition1), 4L);
+    assertEquals(watermarkStates.getWatermark(intermediate), 3L);
+
+    watermarkMessage = new WatermarkMessage(8L, "task 1");
```
nickpan47: minor: Would be nice to add some comments here to describe the test case. For example, advance input watermark, advance output watermark, test input watermark remains the same, test output watermark remains the same, etc.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -53,26 +45,57 @@ import org.apache.samza.task.WindowableTask
...
-      for (containerModel <- jobModel.getContainers.values) {
-        for (taskModel <- containerModel.getTasks.values) {
-          for (ssp <- taskModel.getSystemStreamPartitions) {
-            streamToTasks.put(ssp.getSystemStream, taskModel.getTaskName.toString)
-          }
-        }
-      }
-    }
-    return streamToTasks
+class TaskContextImpl(
```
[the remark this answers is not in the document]
xinyuiscool: Sure. Translated into Java.

samza-core/src/test/java/org/apache/samza/operators/impl/TestControlMessageSender.java
```diff
@@ -17,24 +17,14 @@
...
 import java.util.Set;
 import org.apache.samza.Partition;
-import org.apache.samza.container.TaskName;
-import org.apache.samza.message.ControlMessage;
-import org.apache.samza.system.OutgoingMessageEnvelope;
-import org.apache.samza.system.StreamMetadataCache;
-import org.apache.samza.system.StreamSpec;
-import org.apache.samza.system.SystemStream;
-import org.apache.samza.system.SystemStreamMetadata;
+import org.apache.samza.system.*;
```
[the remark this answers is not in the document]
xinyuiscool: Fixed.

samza-core/src/test/java/org/apache/samza/operators/impl/TestEndOfStreamStates.java
```diff
@@ -0,0 +1,78 @@
...
+    Set<SystemStreamPartition> ssps = new HashSet<>();
+    SystemStreamPartition inputPartition0 = new SystemStreamPartition(input, new Partition(0));
+    SystemStreamPartition intPartition0 = new SystemStreamPartition(intermediate, new Partition(0));
+    SystemStreamPartition intPartition1 = new SystemStreamPartition(intermediate, new Partition(1));
+    ssps.add(inputPartition0);
+    ssps.add(intPartition0);
+    ssps.add(intPartition1);
+
+    Map<SystemStream, Integer> producerCounts = new HashMap<>();
+    producerCounts.put(intermediate, 2);
```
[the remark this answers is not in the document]
xinyuiscool: Oh, this is just a number I come up with. I assume there are 2 upstream tasks producing to this int topic. It doesn't need to relate to the input topic, if we are thinking about multi-stage.

samza-core/src/test/java/org/apache/samza/operators/impl/TestEndOfStreamStates.java
```diff
@@ -0,0 +1,78 @@
...
+    assertFalse(endOfStreamStates.isEndOfStream(intermediate));
+    assertFalse(endOfStreamStates.allEndOfStream());
+
+    IncomingMessageEnvelope envelope = IncomingMessageEnvelope.buildEndOfStreamEnvelope(inputPartition0);
+    endOfStreamStates.update((EndOfStreamMessage) envelope.getMessage(), envelope.getSystemStreamPartition());
+    assertTrue(endOfStreamStates.isEndOfStream(input));
+    assertFalse(endOfStreamStates.isEndOfStream(intermediate));
+    assertFalse(endOfStreamStates.allEndOfStream());
+
+    EndOfStreamMessage eos = new EndOfStreamMessage("task 0");
```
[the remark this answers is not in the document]
xinyuiscool: same, can be tasks from previous stage. It doesn't require upstage tasks have to live in the same job.

samza-core/src/test/java/org/apache/samza/operators/impl/TestEndOfStreamStates.java
```diff
@@ -0,0 +1,78 @@
...
+    endOfStreamStates.update(eos, intPartition0);
+    endOfStreamStates.update(eos, intPartition1);
+    assertFalse(endOfStreamStates.isEndOfStream(intermediate));
+    assertFalse(endOfStreamStates.allEndOfStream());
+
+    eos = new EndOfStreamMessage("task 1");
```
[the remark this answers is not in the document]
xinyuiscool: Oh, there is a confusion here. The tasks are the upstream tasks producing to this topic, not the current task that the topic is assigned to consume.

samza-core/src/test/java/org/apache/samza/operators/impl/TestOperatorImplGraph.java
xinyuiscool: Fixed the name. Currently seems this is the only use case for this function, so I didn't make it general to avoid another filtering afterwards. I think we can make it general if future use cases come.

samza-core/src/test/java/org/apache/samza/operators/impl/TestOperatorImplGraph.java
xinyuiscool: Added.

samza-core/src/test/java/org/apache/samza/operators/impl/TestOperatorImplGraph.java
xinyuiscool: yup, that's a more interesting case. Changed the test to that.

samza-core/src/test/java/org/apache/samza/operators/impl/TestWatermarkStates.java
xinyuiscool: same explanation above.

samza-core/src/test/java/org/apache/samza/operators/impl/TestWatermarkStates.java
xinyuiscool: Added the comments.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -193,7 +194,7 @@ class OffsetManager(
    */
   def update(taskName: TaskName, systemStreamPartition: SystemStreamPartition, offset: String) {
     lastProcessedOffsets.putIfAbsent(taskName, new ConcurrentHashMap[SystemStreamPartition, String]())
-    if (offset != null) {
+    if (offset != null && !offset.equals(IncomingMessageEnvelope.END_OF_STREAM_OFFSET)) {
```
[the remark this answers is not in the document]
nickpan47: OK.... So, the side effect of not persisting this is some dup of messages from last checkpoint? I guess that we can live with it now. I recommend to open a JIRA to track this. Ideally, we want to avoid those known repeats.

samza-core/src/test/java/org/apache/samza/operators/impl/TestEndOfStreamStates.java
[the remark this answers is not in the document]
nickpan47: This is exactly what got me confused in this test case. If inputPartition0 is the first input stream to the application, there can only be one task that is processing inputPartition0. Hence, only 1 task producing to intPartition{0,1}. Hence, I don't get how we can get 2 upstream tasks in this test case?

samza-core/src/test/java/org/apache/samza/operators/impl/TestEndOfStreamStates.java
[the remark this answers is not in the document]
nickpan47: Ah, so, there are x number of upstream tasks producing to input.par-0. But I am still not clear about how do you calculate the producer count for the intermediate stream? If input has only one partition, there can only be one task that is processing the input and producing to the two partitions in the intermediate stream. How do we get producer count 2 for intermediate stream?

samza-core/src/main/java/org/apache/samza/operators/impl/WatermarkStates.java
```diff
@@ -0,0 +1,113 @@
...
+ */
+class WatermarkStates {
+
+  private final static class WatermarkState {
+    private int expectedTotal = Integer.MAX_VALUE;
+    private final Map<String, Long> timestamps = new HashMap<>();
```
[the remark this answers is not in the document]
xinyuiscool: yes, I made it thread safe after seeing your and Chris's comments. Now it will not having any issue even if user configure in-task parallelism (in case of keyed executors based on partitions). So basically update() for each watermark state is under a lock, and the watermark is thread-safe to read from the state.

samza-core/src/test/java/org/apache/samza/operators/impl/TestEndOfStreamStates.java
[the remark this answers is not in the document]
xinyuiscool: As discussed offline, this is part of the whole DAG and we don't assume this current task is also producing to the intermediate topic. These 2 will be upstream tasks that might be in a different container.

PR #279, opened 2017-08-18
SAMZA-1401: update NOTICE for 0.13.1 release with RocksDB BSD+patents license

fredji97:
Please see details in SAMZA-1401.


NOTICE
```diff
@@ -4,3 +4,4 @@ Copyright 2014 The Apache Software Foundation
 This product includes software developed at The Apache Software
 Foundation (http://www.apache.org/).
 
+This product uses RocksDB which carries Facebook BSD+patents license.
```
sborya: Do we need to mention Aug 31 deadline?

PR #280, opened 2017-08-18
SAMZA-1375: Implement Lock for Azure using Lease Blob

PawasChhokra:


samza-azure/src/main/java/org/apache/samza/coordinator/AzureLock.java
```diff
@@ -0,0 +1,100 @@
...
+
+    while ((System.currentTimeMillis() - startTime) < lockTimeout) {
+      try {
+        leaseId.getAndSet(leaseBlobManager.acquireLease(LEASE_TIME_IN_SEC, leaseId.get()));
+      } catch (AzureException e) {
+        return false;
+      }
+      if (leaseId.get() != null) {
+        LOG.info("Acquired lock!");
+        hasLock.set(true);
```
navina: Instead of `#set(boolean)`, use `#getAndSet(boolean)` which atomically sets the value.

samza-azure/src/main/java/org/apache/samza/coordinator/AzureCoordinationServiceFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+package org.apache.samza.coordinator;
+
+import org.apache.samza.config.Config;
+
+public class AzureCoordinationServiceFactory implements CoordinationServiceFactory {
+
+  @Override
+  public CoordinationUtils getCoordinationService(String groupId, String participantId, Config config) {
+    return new AzureCoordinationUtils(config);
```
navina: It's kind of weird that we are not using `groupId` and `participantId`. Isn't that required for your table name, lock name etc?

samza-azure/src/main/java/org/apache/samza/coordinator/AzureCoordinationUtils.java
```diff
@@ -0,0 +1,56 @@
...
+  public AzureCoordinationUtils(Config config) {
+    azureConfig = new AzureConfig(config);
+    this.client = new AzureClient(azureConfig.getAzureConnect());
+  }
+
+  @Override
+  public void reset() {}
+
+  @Override
+  public LeaderElector getLeaderElector() {
```
navina: Why is this returning null? Shouldn't it return the `AzureLeaderElector` ?

samza-azure/src/main/java/org/apache/samza/coordinator/AzureCoordinationUtils.java
```diff
@@ -0,0 +1,56 @@
...
+    return null;
+  }
+
+  @Override
+  public Latch getLatch(int size, String latchId) {
+    return null;
```
navina: This makes sense. Although to be clear, I would throw an "UnsupportedOperationException"

samza-azure/src/main/java/org/apache/samza/coordinator/AzureLock.java
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Unlocks, by releasing the lease on the blob.
+   */
+  @Override
+  public void unlock() {
+    boolean status = leaseBlobManager.releaseLease(leaseId.get());
+    if (status) {
+      LOG.info("Unlocked successfully.");
+      hasLock.set(false);
```
navina: same as above. Use `getAndSet(boolean)`

samza-azure/src/main/java/org/apache/samza/coordinator/AzureLock.java
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/coordinator/AzureLock.java
PawasChhokra: Done.

samza-azure/src/main/java/org/apache/samza/coordinator/AzureCoordinationServiceFactory.java
PawasChhokra: The lock name is currently the lock path which is the hash of the ExecutionPlan. This is passed as a string to the getLock method in AzureCoordinationUtils. Apart from this, I need the container name and the the blob name which are generated using the globalAppId in AzureConfig. Let me know if this is not the right way to do it and I'm misinterpreting something here.

samza-azure/src/main/java/org/apache/samza/coordinator/AzureCoordinationUtils.java
PawasChhokra: I assumed that we won't be needing this function once we refactor the API. Do you think I should still return an AzureLeaderElector anyway?

samza-azure/src/main/java/org/apache/samza/coordinator/AzureCoordinationUtils.java
PawasChhokra: Done.

PR #281, opened 2017-08-21
SAMZA-1404: Add warning in case of potential process starvation due to longer window method

bharathkk:
Currently, we use the average windowNs as the lower bound for window trigger time to determine if user needs to warned. We could potentially make this complicated by also including average commit ns and some delta to be more accurate.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
+++ b/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
@@ -443,6 +443,20 @@ public class AsyncRunLoop implements Runnable, Throttleable {
             long startTime = clock.nanoTime();
             task.window(coordinator);
             containerMetrics.windowNs().update(clock.nanoTime() - startTime);
+
+            // A window() that executes for more than task.window.ms, will starve the next process() call
+            // when the application has job.thread.pool.size > 1. This is due to prioritizing window() ahead of process()
+            // to guarantee window() will fire close to its trigger interval time.
+            // We warn the users if the average window execution time is greater than equals to window trigger interval.
+            long lowerBoundForWindowTriggerTimeInMs = TimeUnit.NANOSECONDS
+                .toMillis((long) containerMetrics.windowNs().getSnapshot().getAverage());
+            if (windowMs <= lowerBoundForWindowTriggerTimeInMs) {
+              log.warn(
+                  "window() call might potentially starve process calls."
+                      + " Consider setting task.window.ms > {} ms",
+                  lowerBoundForWindowTriggerTimeInMs);
+            }
+
             coordinatorRequests.update(coordinator);
 
             state.doneWindow();


samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -443,6 +443,18 @@ public void run() {
             long startTime = clock.nanoTime();
             task.window(coordinator);
             containerMetrics.windowNs().update(clock.nanoTime() - startTime);
+
+            // We have scenario when the task window method takes longer time than window trigger time, there is a
+            // potential for process method starvation. Ideally, we need to merge the thread pool for commit, window with
+            // process to guarantee fairness. In the interim, we want to warn users about potential starvation.
```
xinyuiscool: This comment is not very accurate. Commit,window and process are running in the same thread pool if user sets the job.thread.pool.size. The reason why window is always running is that we give priority to fire window when we schedule the next operation for a task ( see AsyncTaskState.nextOp()). This is to guarantee the window will fire on time.

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -443,6 +443,18 @@ public void run() {
...
+            double lowerBoundForWindowTriggerTimeInMs =
```
xinyuiscool: this seems not very accurate. One large window might skew the avg even the process are invoked after it. And this logic shouldn't be here on the critical path of run loop. I would suggest moving this logic to window() itself, after updating the windowNs, do a count and check there. Seems more accurate that way.

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
bharathkk: Updated the comment.

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -443,6 +443,18 @@ public void run() {
             long startTime = clock.nanoTime();
             task.window(coordinator);
             containerMetrics.windowNs().update(clock.nanoTime() - startTime);
+
+            // We have scenario when the task window method takes longer time than window trigger time, there is a
```
bharathkk: Done

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -443,6 +443,18 @@ public void run() {
...
+            // potential for process method starvation. Ideally, we need to merge the thread pool for commit, window with
+            // process to guarantee fairness. In the interim, we want to warn users about potential starvation.
+            double lowerBoundForWindowTriggerTimeInMs =
+                containerMetrics.windowNs().getSnapshot().getAverage() / 1e6;
+            if (windowMs <= lowerBoundForWindowTriggerTimeInMs) {
+              log.warn(
```
bharathkk: I am still going to leave it as potential starvation scenario since the upcoming window method could end up not starving the process method. I will update the warn statement to use task.window.ms

vjagadish1989: Thank you, my comment was more on the verbosity of the original log line.

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -443,6 +443,19 @@ public void run() {
             long startTime = clock.nanoTime();
             task.window(coordinator);
             containerMetrics.windowNs().update(clock.nanoTime() - startTime);
+
+            // A window() that executes for more than task.window.ms, will starve the next process() call
+            // when the application has job.thread.pool.size > 1. This is due to prioritizing window() ahead of process()
+            // to guarantee window() will fire close to its trigger interval time.
+            // We warn the users if the average window execution time is greater than equals to window trigger interval.
+            double lowerBoundForWindowTriggerTimeInMs =
+                containerMetrics.windowNs().getSnapshot().getAverage() / 1e6;
```
shanthoosh: Could extract `1e6` as static constant.

sborya: please use constant! (with doc preferably)

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -443,6 +443,19 @@ public void run() {
...
+            if (windowMs <= lowerBoundForWindowTriggerTimeInMs) {
+              log.warn(
+                  "Window might potentially starve process calls."
```
sborya: s/Window/window() call/

PR #282, opened 2017-08-22
Introduces CoordinationUtilsFactory to create different implementations of CoordinationUtils. Also adds close() method to CoordinationUtils and to each Utility API separately.

sborya:
Some refactoring and cleanup.


samza-core/src/main/java/org/apache/samza/zk/ZkProcessorLatch.java
```diff
@@ -66,8 +67,19 @@ public void await(long timeout, TimeUnit timeUnit) throws TimeoutException {
...
     String path = zkUtils.getZkClient().createPersistentSequential(latchPath + "/", participantId);
-    LOGGER.debug("ZKProcessorLatch countDown created " + path);
+    LOG.debug("ZKProcessorLatch countDown created " + path);
+  }
+
+  @Override
+  public void close() {
+    try {
+      if (zkUtils != null)
+        zkUtils.close();
```
navina: We should not be closing zkUtils here since it is being instantiated outside of this class and passed as constructor argument.

samza-core/src/main/scala/org/apache/samza/coordinator/Latch.java
```diff
@@ -30,4 +30,9 @@
 public interface Latch {
   void await(long timeout, TimeUnit tu) throws TimeoutException;
   void countDown();
+
+  /**
+   * implementation specific cleanup
+   */
+  void close();
```
navina: Imo, `close` should not belong to the interface api. It is tough to accurately describe the contract the implementation class has to guarantee. In general, if we define the state-changes for the `Latch` interface, then we can reason about whether there is a need for this method to belong to the interface definition. For example, if there was such a thing as `open` for a latch, then, having a corresponding `close` seems reasonable. This is my opinion. 
Will be great if others can chime-in. Thanks!

navina: nit: I think we should add some javadocs for `await` and `countDown` methods.

sborya: But we need to provide some way for cleanup for all of these utilities. So close() does that.

samza-core/src/main/java/org/apache/samza/zk/ZkProcessorLatch.java
sborya: I understand this concern. I can move close() to the CoordinationUitls. The only downside for it is that we can close all the utilities only together.

PR #283, opened 2017-08-25
SAMZA-1406: Fix potential orphaned containers problem in stand alone.

shanthoosh:
Changes:

* Switch from `executorService.shutdown()` to `executorService.shutdownNow()` in `stopScheduler()` implementation. Along with cancelling all the pending tasks, this switch will also interrupt the current running thread.

* Check for the interrupted flag in `ScheduleAfterDebounceTime` scheduled task and trigger the stop sequence.  

* Update stop sequence of `ScheduleAfterDebounceTime` on task exception.

* Move zookeeper specific constants outside of `ScheduleAfterDebounceTime` into `ZkJobCoordinator`(The overall purpose of this `ScheduleAfterDebounceTime` class is to become the common scheduler for both AzureTasks, ZkTasks. This cleanup takes few steps in that general direction).

* Unit test to verify the callback triggering on exception.

* General code cleanup and java doc added to fields and methods. 

NOTE:  Scenario that triggers orphaned containers is the last run of `ScheduleAfterDebounceTime` which brings up the container. Though we handle the interrupt and kill container, during the small window between interrupt occurrence and killing container, there’ll be orphaned containers.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
@@ -20,99 +20,159 @@
 package org.apache.samza.zk;
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
-
-import java.util.HashMap;
-import java.util.Map;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
-
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * This class allows scheduling a Runnable actions after some debounce time.
  * When the same action is scheduled it needs to cancel the previous one. To accomplish that we keep the previous
- * future in a map, keyed by the action name. Here we predefine some actions, which are used in the
- * ZK based standalone app.
+ * future in a map, keyed by the action name.
  */
 public class ScheduleAfterDebounceTime {
-  public static final Logger LOG = LoggerFactory.getLogger(ScheduleAfterDebounceTime.class);
-  public static final long TIMEOUT_MS = 1000 * 10; // timeout to wait for a task to complete
+  private static final Logger LOG = LoggerFactory.getLogger(ScheduleAfterDebounceTime.class);
+  private static final String DEBOUNCE_THREAD_NAME_FORMAT = "debounce-thread-%d";
 
-  // Here we predefine some actions which are used in the ZK based standalone app.
-  // Action name when the JobModel version changes
-  public static final String JOB_MODEL_VERSION_CHANGE = "JobModelVersionChange";
-
-  // Action name when the Processor membership changes
-  public static final String ON_PROCESSOR_CHANGE = "OnProcessorChange";
+  // timeout to wait for a task to complete.
+  private static final int TIMEOUT_MS = 1000 * 10;
 
   /**
-   *
-   * cleanup process is started after every new job model generation is complete.
-   * It deletes old versions of job model and the barrier.
-   * How many to delete (or to leave) is controlled by @see org.apache.samza.zk.ZkJobCoordinator#NUM_VERSIONS_TO_LEAVE.
-   **/
-  public static final String ON_ZK_CLEANUP = "OnCleanUp";
+   * {@link ScheduledTaskCallback} associated with the scheduler. OnError method of the
+   * callback will be invoked on first scheduled task failure.
+   */
+  private Optional<ScheduledTaskCallback> scheduledTaskCallback;
 
-  private final ScheduledTaskFailureCallback scheduledTaskFailureCallback;
+  // Responsible for scheduling delayed actions.
+  private final ScheduledExecutorService scheduledExecutorService;
 
-  private final ScheduledExecutorService scheduledExecutorService = Executors.newSingleThreadScheduledExecutor(
-      new ThreadFactoryBuilder().setNameFormat("debounce-thread-%d").setDaemon(true).build());
+  /**
+   * A map from actionName to {@link ScheduledFuture} of task scheduled for execution.
+   */
   private final Map<String, ScheduledFuture> futureHandles = new HashMap<>();
 
-  // Ideally, this should be only used for testing. But ZkBarrierForVersionUpgrades uses it. This needs to be fixed.
-  // TODO: Timer shouldn't be passed around the components. It should be associated with the JC or the caller of
-  // coordinationUtils.
   public ScheduleAfterDebounceTime() {
-    this.scheduledTaskFailureCallback = null;
+    ThreadFactory threadFactory = new ThreadFactoryBuilder().setNameFormat(DEBOUNCE_THREAD_NAME_FORMAT).setDaemon(true).build();
+    this.scheduledExecutorService = Executors.newSingleThreadScheduledExecutor(threadFactory);
   }
 
-  public ScheduleAfterDebounceTime(ScheduledTaskFailureCallback errorScheduledTaskFailureCallback) {
-    this.scheduledTaskFailureCallback = errorScheduledTaskFailureCallback;
+  public void setScheduledTaskCallback(ScheduledTaskCallback scheduledTaskCallback) {
+    this.scheduledTaskCallback = Optional.ofNullable(scheduledTaskCallback);
   }
 
-  synchronized public void scheduleAfterDebounceTime(String actionName, long debounceTimeMs, Runnable runnable) {
-    // check if this action has been scheduled already
-    ScheduledFuture sf = futureHandles.get(actionName);
-    if (sf != null && !sf.isDone()) {
-      LOG.info("cancel future for " + actionName);
-      // attempt to cancel
-      if (!sf.cancel(false)) {
+  /**
+   * Performs the following operations in sequential order.
+   * <ul>
+   *    <li> Makes best effort to cancel any existing task in task queue associated with the action.</li>
+   *    <li> Schedules the incoming action for later execution and records its future.</li>
+   * </ul>
+   *
+   * @param actionName the name of scheduleable action.
+   * @param delayInMillis the time from now to delay execution.
+   * @param runnable the action to execute.
+   */
+  public synchronized void scheduleAfterDebounceTime(String actionName, long delayInMillis, Runnable runnable) {
+    // 1. Try to cancel any existing scheduled task associated with the action.
+    tryCancelScheduledAction(actionName);
+
+    // 2. Schedule the action.
+    ScheduledFuture scheduledFuture = scheduledExecutorService.schedule(getScheduleableAction(actionName, runnable), delayInMillis, TimeUnit.MILLISECONDS);
+
+    LOG.info("Scheduled action: {} to run after: {} milliseconds.", actionName, delayInMillis);
+    futureHandles.put(actionName, scheduledFuture);
+  }
+
+  /**
+   * Stops the scheduler. After this invocation no further schedule calls will be accepted
+   * and all pending enqueued tasks will be cancelled.
+   */
+  public synchronized void stopScheduler() {
+    scheduledExecutorService.shutdownNow();
+
+    // Clear the existing future handles.
+    futureHandles.clear();
+  }
+
+  /**
+   * Tries to cancel the task that belongs to {@code actionName} submitted to the queue.
+   *
+   * @param actionName the name of action to cancel.
+   */
+  private void tryCancelScheduledAction(String actionName) {
+    ScheduledFuture scheduledFuture = futureHandles.get(actionName);
+    if (scheduledFuture != null && !scheduledFuture.isDone()) {
+      LOG.info("Attempting to cancel the future of action: {}", actionName);
+      // Attempt to cancel
+      if (!scheduledFuture.cancel(false)) {
         try {
-          sf.get(TIMEOUT_MS, TimeUnit.MILLISECONDS);
+          scheduledFuture.get(TIMEOUT_MS, TimeUnit.MILLISECONDS);
         } catch (Exception e) {
           // we ignore the exception
-          LOG.warn("cancel for action " + actionName + " failed with ", e);
+          LOG.warn("Cancelling the future of action: {} failed.", actionName, e);
         }
       }
       futureHandles.remove(actionName);
     }
-    // schedule a new task
-    sf = scheduledExecutorService.schedule(() -> {
-        try {
-          runnable.run();
-          LOG.debug(actionName + " completed successfully.");
-        } catch (Throwable t) {
-          LOG.error(actionName + " threw an exception.", t);
-          if (scheduledTaskFailureCallback != null) {
-            scheduledTaskFailureCallback.onError(t);
-          }
+  }
+
+  /**
+   * Decorate the executable action with exception handlers to facilitate cleanup on failures.
+   *
+   * @param actionName the name of the scheduleable action.
+   * @param runnable the action to execute.
+   * @return the executable action decorated with exception handlers.
+   */
+  private Runnable getScheduleableAction(String actionName, Runnable runnable) {
+    return () -> {
+      try {
+        runnable.run();
+        /*
+         * Expects all run() implementations <b>not to swallow the interrupts.</b>
+         * This thread is interrupted from an external source(mostly executor service) to die.
+         */
+        if (Thread.currentThread().isInterrupted()) {
+          LOG.warn("Action: {} is interrupted.", actionName);
+          doCleanUpOnTaskException(new InterruptedException());
+        } else {
+          LOG.debug("Action: {} completed successfully.", actionName);
         }
-      },
-     debounceTimeMs,
-     TimeUnit.MILLISECONDS);
-    LOG.info("scheduled " + actionName + " in " + debounceTimeMs);
-    futureHandles.put(actionName, sf);
+      } catch (Exception exception) {
+        LOG.error("Execution of action: {} failed.", actionName, exception);
+        doCleanUpOnTaskException(exception);
+      }
+    };
   }
 
-  public void stopScheduler() {
-    // shutdown executor service
-    scheduledExecutorService.shutdown();
+  /**
+   * Handler method to invoke on a exception during an scheduled task execution and which
+   * the following operations in sequential order.
+   * <ul>
+   *   <li> Stops the scheduler.</li>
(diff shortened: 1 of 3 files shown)


shanthoosh:
@sborya @navina 

Please take a look.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,164 @@
...
+    // 1. Stop the scheduler. If a task execution fails or a task is interrupted, scheduler will not accept/execute any new tasks.
+    stopScheduler();
+
+    // 2. Invoke the callback if its defined.
+    scheduledTaskCallback.ifPresent(callback -> callback.onError(exception));
   }
 
-  interface ScheduledTaskFailureCallback {
+  /**
+   * A ScheduledTaskCallback is fired to client of {@link ScheduleAfterDebounceTime},
```
navina: Rephrase : " ScheduledTaskCallback is invoked on the first occurrence of an exception when executing a task" , perhaps? 

Thanks for adding docs, btw.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,164 @@
...
+      try {
+        runnable.run();
+        LOG.debug("Action: {} completed successfully.", actionName);
+        /*
+         * Expects all run() implementations <b>not to swallow the interrupts.</b>
+         * This thread is interrupted from an external source(mostly executor service) to die.
+         */
+        if (Thread.currentThread().isInterrupted()) {
+          LOG.warn("Action: {} is interrupted.", actionName);
+          doCleanUpOnTaskException(new InterruptedException());
```
navina: minor suggestion: It is perhaps simpler to set a local variable for storing the exception value. Inspect its value in `finally` clause and then, take action.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,164 @@
...
 
-  // Ideally, this should be only used for testing. But ZkBarrierForVersionUpgrades uses it. This needs to be fixed.
-  // TODO: Timer shouldn't be passed around the components. It should be associated with the JC or the caller of
-  // coordinationUtils.
-  public ScheduleAfterDebounceTime() {
-    this.scheduledTaskFailureCallback = null;
+  /**
+   * To achieve mutual exclusion when multiple threads are scheduling debounceable actions.
+   */
+  private final Object lock = new Object();
```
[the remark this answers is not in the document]
sborya: +1

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,164 @@
...
+
+      // 2. Schedule the action.
+      ScheduledFuture scheduledFuture = scheduledExecutorService.schedule(getScheduleableAction(actionName, runnable), delayInMillis, TimeUnit.MILLISECONDS);
+
+      LOG.info("Scheduled action: {} to run after: {} milliseconds.", actionName, delayInMillis);
+      futureHandles.put(actionName, scheduledFuture);
+    }
   }
 
-  synchronized public void scheduleAfterDebounceTime(String actionName, long debounceTimeMs, Runnable runnable) {
```
sborya: again, not sure why remove synchronized.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,164 @@
...
+    this.scheduledExecutorService = Executors.newSingleThreadScheduledExecutor(threadFactory);
+    this.scheduledTaskCallback = Optional.ofNullable(scheduledTaskCallback);
   }
 
-  public ScheduleAfterDebounceTime(ScheduledTaskFailureCallback errorScheduledTaskFailureCallback) {
-    this.scheduledTaskFailureCallback = errorScheduledTaskFailureCallback;
+  /**
+   * Performs the following operations in sequential order.
+   * <ul>
+   *    <li> Makes best possible attempt to cancel any existing task in task queue associated with the action.</li>
```
sborya: s/best possible/best effort/

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,164 @@
...
+   * @return the executable action decorated with exception handlers.
+   */
+  private Runnable getScheduleableAction(String actionName, Runnable runnable) {
+    return () -> {
+      try {
+        runnable.run();
+        LOG.debug("Action: {} completed successfully.", actionName);
+        /*
+         * Expects all run() implementations <b>not to swallow the interrupts.</b>
+         * This thread is interrupted from an external source(mostly executor service) to die.
```
sborya: this comment is not clear. Please rephrase.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,164 @@
...
+   *
+   * @param actionName the name of the scheduleable action.
+   * @param runnable the action to execute.
+   * @return the executable action decorated with exception handlers.
+   */
+  private Runnable getScheduleableAction(String actionName, Runnable runnable) {
+    return () -> {
+      try {
+        runnable.run();
+        LOG.debug("Action: {} completed successfully.", actionName);
```
sborya: should be in the else block.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,164 @@
...
+          LOG.warn("Action: {} is interrupted.", actionName);
+          doCleanUpOnTaskException(new InterruptedException());
         }
-      },
-     debounceTimeMs,
-     TimeUnit.MILLISECONDS);
-    LOG.info("scheduled " + actionName + " in " + debounceTimeMs);
-    futureHandles.put(actionName, sf);
+      } catch (Exception exception) {
+        LOG.error("Execution of action: {} failed.", new Object[]{actionName, exception});
```
sborya: please pass exception as the third parameter.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
sborya: +1

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,164 @@
...
+    scheduledTaskCallback.ifPresent(callback -> callback.onError(exception));
   }
 
-  interface ScheduledTaskFailureCallback {
+  /**
+   * A ScheduledTaskCallback is fired to client of {@link ScheduleAfterDebounceTime},
+   * on first occurrence of exception when executing a task. Provides plausible
+   * hook for handling failures in an asynchronous scheduled task execution.
+   */
+  interface ScheduledTaskCallback {
```
sborya: please change to new ScheduleTaskCallback in the caller (ZkJobCoordinator).

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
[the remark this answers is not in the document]
shanthoosh: Functionally they're same(some token to use to lock on).

 I prefer to lock on internal private field which is hidden and no other block/code will be able to lock on/through it.

If we lock on “this”(synchronized), then any other code which has access to the object of `ScheduleAfterDebounceTime`, could be able to use that to lock on it.

When it happens could lead to deadlocks, or just excessive locking(Synchronization achieved by `ScheduleAfterDebounceTime` object on any other method foo outside of `ScheduleAfterDebounceTime` class could potentially block execution of schedule methods).

Some classes in samza itself follow this pattern(`PollingScanDiskSpaceMonitor`, `TaskCallbackManager`, `StreamPartitionCountMonitor`, `StatisticsMonitorImpl` to name a few).

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,164 @@
...
-    if (sf != null && !sf.isDone()) {
-      LOG.info("cancel future for " + actionName);
-      // attempt to cancel
-      if (!sf.cancel(false)) {
+  /**
+   * Stops the scheduler. After this invocation no further schedule calls will be accepted
+   * and all pending enqueued tasks will be cancelled.
+   */
+  public void stopScheduler() {
+    scheduledExecutorService.shutdownNow();
```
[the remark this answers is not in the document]
shanthoosh: `scheduledExecutorService.shutdownNow` is thread safe.

We could move shutdown within lock for symmetry.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
shanthoosh: Let's discuss this offline.

I'm not sure of benefits gained by replacing a lambda with anonymous inner class implementation.

shanthoosh:
@navina 
>Is it possible to add some kind of testing for orphaned containers problem itself ?

I wanted to do that before sending the patch for review. 

Scenario to test is interruption of a particular task (SamzaContainerLauncher) which gets triggered as effect on zk event sequence chain.

But all logic to spawn SamzaContainer in standalone is hidden within `ZkJobCoordinator` (It's hard to pull that which in itself would require huge refactoring).

I was thinking to mimic `ZkJobCoordinator` logic to see if we interrupt a running samza conatiner within ScheduleDebounce thread. But that wouldn't test any of our logic.

Also, this scenario is very hard to test/control from public LocalApplicationRunner / ZkJobCoordinator api's.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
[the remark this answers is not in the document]
navina: `shutdownNow` may be thread-safe. But when you are using synchronization primitives like `lock`, it is important to keep the all code within the critical section under the lock, even if some of them are therad-safe. Here, you want to make the operations of canceling & submitting task to the scheduler and cleaning up the scheduler resources to be mutually exclusive. It is clearer to include it within the lock.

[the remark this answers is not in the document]
shanthoosh: @navina 
I should have been more explicit/clear in the comment.

Yes, I've already moved it within the lock.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,156 @@
...
+        doCleanUpOnTaskException(exception);
+      }
+    };
   }
 
-  public void stopScheduler() {
-    // shutdown executor service
-    scheduledExecutorService.shutdown();
+  /**
+   * Handler method to invoke on an exception during an scheduled task execution and does
```
sborya: s/an/a/
s/does/which/
s/./:/

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,156 @@
...
+   * the following operations in sequential order.
+   * <ul>
+   *   <li> Stops the scheduler.</li>
+   *   <li> Invokes the onError handler method if taskCallback is defined.</li>
+   * </ul>
+   *
+   * @param exception the exception happened during task execution.
+   */
+  private void doCleanUpOnTaskException(Exception exception) {
+    // 1. Stop the scheduler. If a task execution fails or a task is interrupted, scheduler will not accept/execute any new tasks.
```
sborya: s/a task/the task/

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,156 @@
...
+    stopScheduler();
+
+    // 2. Invoke the callback if its defined.
+    scheduledTaskCallback.ifPresent(callback -> callback.onError(exception));
   }
 
-  interface ScheduledTaskFailureCallback {
+  /**
+   * A ScheduledTaskCallback is invoked on first occurrence of exception
```
sborya: It is either generic callback interface or error specific one, so either rename the interface back to ScheduledTaskFailureCallback or in the comment write ScheduledTaskCallback::onError().

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
sborya: +1

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
shanthoosh: done.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
shanthoosh: done.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
shanthoosh: Done. Updated the comment.

shanthoosh: @navina 
Myself & boris discussed offline, there're two approaches to not take in ScheduledTaskCallbak as constructor param (It looks kind of weird to take in callback as param as constructor param).
i) have setScheduledTaskCallback method here in this class
ii) Or taken in scheduleTaskCallBack as a param to schedule method itself. 

ii) involves touching/modifying in lot of places, hence going with i) which is reasonable as well. Do you agree with ii)

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
shanthoosh: @navina  @sborya 
I also wanted to do that, we need to declare that exception variable as effectively final since this happens in a lambda(either as an array or a Wrapper, which i think is a kill).
I tried adding it that way initially and it looks weird. 
Please take a look at the test (for same thing that i'm doing there)

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
navina: @shanthoosh thanks for listing the approaches. I agree that ii) is the ideal solution. For now, it's fine to keep it in setScheduledTaskCallback.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,159 @@
...
+   *   <li> Invokes the onError handler method if taskCallback is defined.</li>
+   * </ul>
+   *
+   * @param exception the exception happened during task execution.
+   */
+  private void doCleanUpOnTaskException(Exception exception) {
+    // 1. Stop the scheduler. If the task execution fails or a task is interrupted, scheduler will not accept/execute any new tasks.
+    stopScheduler();
+
+    // 2. Invoke the callback if its defined.
```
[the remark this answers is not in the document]
shanthoosh: Removed.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -20,99 +20,159 @@
...
+   * the following operations in sequential order.
+   * <ul>
+   *   <li> Stops the scheduler.</li>
+   *   <li> Invokes the onError handler method if taskCallback is defined.</li>
+   * </ul>
+   *
+   * @param exception the exception happened during task execution.
+   */
+  private void doCleanUpOnTaskException(Exception exception) {
+    // 1. Stop the scheduler. If the task execution fails or a task is interrupted, scheduler will not accept/execute any new tasks.
```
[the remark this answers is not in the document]
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
shanthoosh: I agree/tried initially to follow the way you are recommending. As we'd discussed offline, declaring it as array or reference (just to make it effectively final) is the reason which make it more convoluted. This way is simplar.

navina:
@shanthoosh Merged this patch to 0.14.0 branch. You can close this PR. Thanks!

prateekm:
@shanthoosh looks like this is merged, let's close this PR.

nickpan47:
@shanthoosh please close this one.

shanthoosh:
@nickpan47 @prateekm 
Done.

PR #286, opened 2017-08-25
SAMZA-1408 update the branch doc for 0.13.1 release

fredji97:


docs/_config.yml
```diff
@@ -23,6 +23,6 @@ redcarpet:
   extensions: ['with_toc_data', 'smart']
 exclude: [_docs]
 baseurl: http://samza.apache.org
-version: latest
+version: 0.13.1
```
nickpan47: need '' marks.

nickpan47: Note that we don't need change in the doc version since we never track the minor version docs in the release. We simply overwrite the previous 0.13.0 doc version w/ the new 0.13.1 under the same 0.13 version.

nickpan47:
@fredji97 do we still need to merge this one? Or can one close it now?

fredji97:
@nickpan47 this one was merged in the past for 0.13.1 release. we can close it now.

PR #289, opened 2017-08-28
SAMZA-1410 make tests pass in different Junit versions for testPlanIdWithShuffledStreamSpecs and testGeneratePlanIdWithDifferentStreamSpec

fredji97:
More details are in https://issues.apache.org/jira/browse/SAMZA-1410.

gradlew build and test passed.


fredji97:
Thanks @sborya ! The original import lines were importing Assert class with *, so there is nothing to remove there. 

Depending on the code style and best practice, most recommend not to use * to do wildcard matching (which I prefer), but some may think it is better to keep the existing style for consistency purpose. Anyway, we can discus about that one and determine an import practice horizontally for all files. For this PR, I think it is OK to not touch since it is out of the scope of the purpose.

fredji97:
@sborya Thanks. Reformatted the lines to make lines shorter.

PR #290, opened 2017-08-30
SAMZA-1413: Config for CoordinationUtilsFactory class name.

sborya:


samza-core/src/main/java/org/apache/samza/config/JobCoordinatorConfig.java
```diff
@@ -22,33 +22,31 @@
...
 import org.apache.samza.SamzaException;
 import org.apache.samza.zk.ZkCoordinationUtilsFactory;
-import org.apache.samza.zk.ZkJobCoordinatorFactory;
-
 
 public class JobCoordinatorConfig extends MapConfig {
   public static final String JOB_COORDINATOR_FACTORY = "job.coordinator.factory";
   public static final String JOB_COORDINATION_UTILS_FACTORY = "job.coordination.utils.factory";
-  public final static String DEFAULT_COORDINATION_UTILS_FACTORY = ZkCoordinationUtilsFactory.class.getName();
+  public final static String DEFAULT_COORDINATION_UTILS_FACTORY = ZkCoordinationUtilsFactory.class.getName(); //default
```
navina: nit: you can remove the default comment :)

samza-core/src/main/java/org/apache/samza/config/JobCoordinatorConfig.java
```diff
@@ -22,33 +22,31 @@
...
 
   public String getJobCoordinationUtilsFactoryClassName() {
-    String jobCoordinatorFactoryClassName = get(JobCoordinatorConfig.JOB_COORDINATOR_FACTORY, "");
-
-    String className = get(JOB_COORDINATION_UTILS_FACTORY, "");
+    String className = get(JOB_COORDINATION_UTILS_FACTORY, DEFAULT_COORDINATION_UTILS_FACTORY);
 
-    if (!Strings.isNullOrEmpty(className)) {
-      return className;
+    if (Strings.isNullOrEmpty(className)) {
```
navina: className would never be null or empty if you have a default, right?

samza-core/src/main/java/org/apache/samza/config/JobCoordinatorConfig.java
```diff
@@ -22,33 +22,31 @@
...
+      throw new SamzaException("Failed to read config value for " + JOB_COORDINATION_UTILS_FACTORY + " = " + className);
```
fredji97: this exception can be modified to be more accurate. Basically, the exception happens if the user defines JOB_COORDINATION_UTILS_FACTORY in config but put the empty string. If not configured, it is fine because it is using default, but if configured, an exception is thrown because of the wrong configuration.

samza-core/src/main/java/org/apache/samza/config/JobCoordinatorConfig.java
sborya: done.

samza-core/src/main/java/org/apache/samza/config/JobCoordinatorConfig.java
sborya: i guess you're right. Fixed.

samza-core/src/main/java/org/apache/samza/config/JobCoordinatorConfig.java
sborya: changed.

samza-core/src/main/java/org/apache/samza/config/JobCoordinatorConfig.java
navina: Ah.. I see. That sounds good!

PR #291, opened 2017-08-30
Remove latch sharing between task and tests to enable parallel execution

bharathkk:
Refactoring the tests to enable parallel execution. Also, removed unnecessary sleeping during execution. This PR will serve as a prototype to get some feedback on this testing pattern which eliminates state sharing between the tests and opens up the avenue for parallel execution.

This change brings down the test time for these 4 tests to **14 sec** compared to **44 secs**.

### DIFF
--- a/samza-test/src/test/java/org/apache/samza/test/processor/IdentityStreamTask.java
+++ b/samza-test/src/test/java/org/apache/samza/test/processor/IdentityStreamTask.java
@@ -23,19 +23,14 @@ import org.apache.samza.config.Config;
 import org.apache.samza.system.IncomingMessageEnvelope;
 import org.apache.samza.system.OutgoingMessageEnvelope;
 import org.apache.samza.system.SystemStream;
-import org.apache.samza.task.ClosableTask;
 import org.apache.samza.task.InitableTask;
 import org.apache.samza.task.MessageCollector;
 import org.apache.samza.task.StreamTask;
 import org.apache.samza.task.TaskContext;
 import org.apache.samza.task.TaskCoordinator;
 
-import java.util.concurrent.CountDownLatch;
 
-public class IdentityStreamTask implements StreamTask , InitableTask , ClosableTask  {
-  // static field since there's no other way to share state b/w a task instance and
-  // stream processor when constructed from "task.class".
-  static CountDownLatch endLatch = new CountDownLatch(1);
+public class IdentityStreamTask implements StreamTask , InitableTask  {
   private int processedMessageCount = 0;
   private int expectedMessageCount;
   private String outputTopic;
@@ -59,14 +54,7 @@ public class IdentityStreamTask implements StreamTask , InitableTask , ClosableT
             incomingMessageEnvelope.getMessage()));
     processedMessageCount++;
     if (processedMessageCount == expectedMessageCount) {
-      endLatch.countDown();
+      taskCoordinator.shutdown(TaskCoordinator.RequestScope.ALL_TASKS_IN_CONTAINER);
     }
-  }
-
-  @Override
-  public void close() throws Exception {
-    // need to create a new latch after each test since it's a static field.
-    // tests are assumed to run sequentially.
-    endLatch = new CountDownLatch(1);
   }
 }
--- a/samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
+++ b/samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
@@ -53,7 +53,9 @@ import org.junit.Test;
 import scala.Option$;
 
 import static org.mockito.Matchers.anyObject;
-import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.mock;
 
 
 public class TestStreamProcessor extends StandaloneIntegrationTestHarness {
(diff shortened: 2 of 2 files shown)


bharathkk:
@navina @nickpan47 @sborya

samza-test/src/test/java/org/apache/samza/test/processor/IdentityStreamTask.java
```diff
@@ -59,14 +54,7 @@ public void process(
             incomingMessageEnvelope.getMessage()));
     processedMessageCount++;
     if (processedMessageCount == expectedMessageCount) {
-      endLatch.countDown();
+      taskCoordinator.shutdown(TaskCoordinator.RequestScope.ALL_TASKS_IN_CONTAINER);
```
navina: awesome change :)

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
```diff
@@ -47,26 +49,14 @@
...
 import org.junit.Assert;
-import org.junit.Before;
 import org.junit.Test;
+import scala.Option$;
 
 import static org.mockito.Matchers.anyObject;
-import static org.mockito.Mockito.doNothing;
-import static org.mockito.Mockito.mock;
-import static org.apache.samza.test.processor.IdentityStreamTask.endLatch;
+import static org.mockito.Mockito.*;
```
navina: Let's import only the required packages and avoid "*" ?

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
```diff
@@ -247,4 +220,70 @@ private void verifyNumMessages(String topic, int expectedNumMessages) {
...
+          4096L,
+          "org.apache.kafka.clients.consumer.RangeAssignor",
+          30000,
+          SecurityProtocol.PLAINTEXT,
+          Option$.MODULE$.empty(),
+          Option$.MODULE$.empty(),
+          Option$.MODULE$.empty());
+    }
+
+    private void initProcessListener() {
```
navina: nit: perhaps rename to initProcessorListener

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
```diff
@@ -247,4 +220,70 @@ private void verifyNumMessages(String topic, int expectedNumMessages) {
 
     Assert.assertEquals(count, expectedNumMessages);
   }
+
+  private static class Mocks {
```
navina: It will be awesome if we can rename to something other than Mock. Sounds like you are trying to wrap the producer, consumer, processor and listener into a class. What are we benefitting from this? 

It sound like you want a stream processor factory for testing purposes. Producer and consumer are already in a util method. Perhaps you simply want to move initConsumer and initProducer as a private method in TestStreamProcessor, instead of tying it together with "Mocks". Please correct me if I am wrong here.

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
```diff
@@ -247,4 +220,70 @@ private void verifyNumMessages(String topic, int expectedNumMessages) {
...
+    CountDownLatch latch;
```
vjagadish1989: minor : prefer this to be `shutdownLatch` to make intent explicit

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
vjagadish1989: +1 from my side too or can rename it to something other than `Mocks`

samza-test/src/test/java/org/apache/samza/test/processor/IdentityStreamTask.java
vjagadish1989: Any reason why the `shutdown()` is faster than `latch` in this case?

bharathkk: ```shutdown()``` is much cleaner because we don't have to share latch between task and test which opens the door for parallel execution. In terms of speed, I believe there isn't big difference.

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
bharathkk: done

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
bharathkk: Discussed offline. I will rename this to something meaningful instead of mocks. You are right in some sense this is a wrapper class to consolidate all the dependencies required to unit test stream processor.

I could have them as private methods and have test methods invoke them to get instances of producer & consumer. IMO, it isn't any different except the current implementation makes it slightly readable in terms of setting up dependencies for my test in one place.
Let me know if you feel strongly about moving them as private methods.

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
bharathkk: done

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamProcessor.java
bharathkk: done

vjagadish1989:
Discussed with @bharathkk offline. 
- The big difference in speed from 14s->4s comes from the removal of `Thread.sleep(10,000)`. 
- Using  `shutdown` is a first step to remove state sharing between tests and potentially enabling parallel execution later.

+1, Looks great! 👍

PR #292, opened 2017-09-06
SAMZA-1415: Add clearStream API in SystemAdmin and remove deprecated APIs

xinyuiscool:
The patch does the following:

1) add clearStream() APi in SystemAdmin. Currently it's only supported in Kafka with broker configuring delete.topic.enable=true.

2) remove the deprecated APIs including createChangeLogStream(), validateChangelogStream() and createCoordinatorStream().


xinyuiscool:
Tested locally with hello-samza. Both coordinator and changelog streams are created as expected.

@jmakes : could you please help review it?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -491,32 +498,45 @@ class KafkaSystemAdmin(
...
 
   /**
-    * Validates a stream in Kafka. Should not be called before createStream(),
-    * since ClientUtils.fetchTopicMetadata(), used by different Kafka clients, is not read-only and
-    * will auto-create a new topic.
+    * Exception to be thrown when the change log stream creation or validation has failed
     */
-  override def validateChangelogStream(topicName: String, numKafkaChangelogPartitions: Int) = {
-    validateStream(new KafkaStreamSpec(CHANGELOG_STREAMID, topicName, systemName, numKafkaChangelogPartitions))
+  class KafkaChangelogException(s: String, t: Throwable) extends SamzaException(s, t) {
```
jmakes: Not a big deal, but since you're already cleaning up this class so much, what do you think about removing this exception and using StreamValidationException in its place?

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemAdminJava.java
```diff
@@ -54,7 +55,7 @@ public void testCreateCoordinatorStreamDelegatesToCreateStream() {
   public void testCreateChangelogStreamDelegatesToCreateStream() {
```
jmakes: Suggest updating the name for this test, since createChangelogStream no longer exists so the delegation is not needed. Instead, this test seems to primarily verify that the changelog properties are used when the StreamSpec is a changelog StreamSpec.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemAdminJava.java
```diff
@@ -94,22 +96,23 @@ public void testCreateChangelogStreamDelegatesToCreateStream_specialCharsInTopic
...
+      assertEquals(STREAM, internalSpec.getPhysicalName());
+      assertEquals(REP_FACTOR, ((KafkaStreamSpec) internalSpec).getReplicationFactor());
+      assertEquals(PARTITIONS, internalSpec.getPartitionCount());
+      assertEquals(changeLogProps, ((KafkaStreamSpec) internalSpec).getProperties());
+
+      return internalSpec;
+    }).when(admin).toKafkaSpec(Mockito.any());
   }
 
   @Test
```
jmakes: Doesn't seem like this test is needed anymore. There's no delegation.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemAdminJava.java
```diff
@@ -121,14 +124,13 @@ public void testValidateChangelogStreamDelegatesToValidateStream() {
...
+    admin.createStream(spec);
     admin.validateStream(spec);
-    admin.validateChangelogStream(spec.getPhysicalName(), spec.getPartitionCount());
 
     Mockito.verify(admin).createStream(Mockito.any());
-    Mockito.verify(admin, Mockito.times(3)).validateStream(Mockito.any());
+    Mockito.verify(admin, Mockito.times(1)).validateStream(Mockito.any());
   }
 
   @Test
```
jmakes: Doesn't seem like this test is needed anymore. There's no delegation.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemAdminJava.java
```diff
@@ -63,29 +64,30 @@ public void testCreateChangelogStreamDelegatesToCreateStream() {
...
+      assertEquals(REP_FACTOR, ((KafkaStreamSpec) internalSpec).getReplicationFactor());
+      assertEquals(PARTITIONS, internalSpec.getPartitionCount());
+      assertEquals(changeLogProps, ((KafkaStreamSpec) internalSpec).getProperties());
+
+      return internalSpec;
+    }).when(admin).toKafkaSpec(Mockito.any());
   }
 
   @Test
   public void testCreateChangelogStreamDelegatesToCreateStream_specialCharsInTopicName() {
```
jmakes: Not related to your change, but it seems like there should be an equivalent test for coordinator streams with special characters. I thought there was one, but can't find it in this patch. 

I even see it on my local copy.

testCreateCoordinatorStreamDelegatesToCreateStream_specialCharsInTopicName

If it doesn't exist here, would you mind adding it?

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemAdminJava.java
xinyuiscool: Removed the test.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemAdminJava.java
xinyuiscool: Removed the test.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -87,7 +87,7 @@ default void validateStream(StreamSpec streamSpec) throws StreamValidationExcept
    * Clear the stream described by the spec.
    * @param streamSpec  The spec for the physical stream on the system.
    */
```
jmakes: Please add doc for return value

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -50,38 +50,6 @@
   Map<String, SystemStreamMetadata> getSystemStreamMetadata(Set<String> streamNames);
 
   /**
-   * An API to create a change log stream
```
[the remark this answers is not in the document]
jmakes: Talk to @navina. I think the release notes will go in a wiki, but I'm not sure.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
xinyuiscool: Fixed.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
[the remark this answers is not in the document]
navina: Yes. But I am yet to create that page :) will ping @xinyuiscool once I have it ready. I am noting down the PRs that need release/upgrade notes. Thanks!

PR #293, opened 2017-09-08
SAMZA-1109: Allow setting Serdes for fluent API operators in code

prateekm:


prateekm:
+ @cpettitt-linkedin too.

samza-core/src/main/scala/org/apache/samza/serializers/JsonSerde.scala
```diff
@@ -22,26 +22,36 @@ package org.apache.samza.serializers
 import org.apache.samza.SamzaException
 import org.apache.samza.serializers.model.SamzaObjectMapper
 import org.codehaus.jackson.`type`.TypeReference
-import org.codehaus.jackson.map.ObjectMapper
 import org.apache.samza.config.Config
 
 class JsonSerde[T] extends Serde[T] {
-  val mapper = SamzaObjectMapper.getObjectMapper()
+  @transient lazy val mapper = SamzaObjectMapper.getObjectMapper()
+  var clazzOption: Option[Class[T]] = None
```
cpettitt-linkedin: Not super familiar with Scala semantics. Could this be made a val?

samza-test/src/main/resources/log4j.xml
```diff
@@ -26,14 +26,14 @@
     <param name="MaxFileSize" value="256MB" />
     <param name="MaxBackupIndex" value="1" />
     <layout class="org.apache.log4j.PatternLayout">
-      <param name="ConversionPattern" value="%d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %c{1} [%p] %m%n" />
```
cpettitt-linkedin: Not clear why the logging format is changing based on the pull request description.

samza-test/src/test/java/org/apache/samza/test/operator/RepartitionWindowApp.java
```diff
@@ -42,16 +43,20 @@
 
   @Override
   public void init(StreamGraph graph, Config config) {
+    graph.setDefaultKeySerde(new StringSerde());
+    // will fail with class cast in partitionBy if (optional) PageView.class param not provided
+    graph.setDefaultMsgSerde(new JsonSerde<>(PageView.class));
```
cpettitt-linkedin: Given the likelihood that messages are going to change in some way in the pipeline and that this is not type safe, have we considered defaulting not for the whole pipeline but by type?

One other consideration is that neglecting to set this and an explicit serializer later would give an info entry in the log and a class cast exception. It seems that it would be better to require the user to explicitly use a pass-through SerDe and error out early if not SerDe is set. Also, if we really do expose the ClassCastException to the user it would be good to augment that with some info about what to look for (e.g. make sure serde matches type).

jmakes:
Just one immediate comment before diving into code:

> I've made a choice here to disallow setting system/stream serdes in config to make configs simpler. Imho the 'serdes only in code' and 'only 2 levels of serde' model is much simpler to understand. The risk is that there might be streams that aren't part of the user application code, but need their serdes to be set in configuration. An example is the logging topic. Do you think this is a reasonable choice?

👍  for this. I think you've identified the only reason serdes should have ever been in config and imo we can find a way to configure them in code for the high-level-api

I assume nothing's changed for low level api. It's still config driven for all serdes, right?

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -65,46 +73,82 @@ public StreamGraphImpl(ApplicationRunner runner, Config config) {
...
 
   @Override
-  public <K, V, M> OutputStream<K, V, M> getOutputStream(String streamId,
-      Function<? super M, ? extends K> keyExtractor, Function<? super M, ? extends V> msgExtractor) {
-    if (keyExtractor == null) {
-      throw new IllegalArgumentException("keyExtractor can't be null for an output stream.");
+  public <K, V, M> MessageStream<M> getInputStream(String streamId,
+      BiFunction<? super K, ? super V, ? extends M> msgBuilder) {
+    if (defaultKeySerde instanceof NoOpSerde) {
+      LOGGER.info("Using NoOpSerde as the default key serde for input stream " + streamId +
```
jmakes: Had a comment above asking for this. Very thorough, thanks.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -65,46 +73,82 @@ public StreamGraphImpl(ApplicationRunner runner, Config config) {
...
-        new OutputStreamImpl<>(streamSpec, (Function<M, K>) keyExtractor, (Function<M, V>) msgExtractor));
+        new OutputStreamImpl<>(streamSpec, keySerde, msgSerde,
+            (Function<M, K>) keyExtractor, (Function<M, V>) msgExtractor));
     return outputStreams.get(streamSpec);
   }
 
   @Override
+  public <K, V, M> OutputStream<K, V, M> getOutputStream(String streamId,
+      Function<? super M, ? extends K> keyExtractor, Function<? super M, ? extends V> msgExtractor) {
+    if (defaultKeySerde instanceof NoOpSerde) {
```
jmakes: These checks are repeated a few times. Refactoring them would help us keep the logic (particularly the log messages) consistent.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -126,31 +172,45 @@ public StreamGraph withContextManager(ContextManager contextManager) {
...
+  <K, V, M> IntermediateMessageStreamImpl<K, V, M> getIntermediateStream(String streamName,
+      Function<? super M, ? extends K> keyExtractor, Function<? super M, ? extends V> msgExtractor,
+      BiFunction<? super K, ? super V, ? extends M> msgBuilder) {
+    if (defaultKeySerde instanceof NoOpSerde) {
+      LOGGER.info("Using NoOpSerde as the default key serde for intermediate stream " + streamName +
+          ". Keys will not be (de)serialized");
+    }
+    if (defaultMsgSerde instanceof NoOpSerde) {
+      LOGGER.info("Using NoOpSerde as the default msg serde for intermediate stream " + streamName +
+          ". Keys will not be (de)serialized");
```
jmakes: Typo: "Messages will not be..."

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
jmakes: Related question, are the small variances in log messages intentional?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -232,8 +232,27 @@ object SamzaContainer extends Logging {
 
       (serdeName, serde)
     }).toMap
+    info("Got serdes from factories: %s" format serdesFromFactories.keys)
+
+    val serdesFromSerializedInstances = config.subset(SerializerConfig.SERIALIZER_PREFIX format "").asScala
+        .filter { case (key, value) => key.endsWith(SerializerConfig.SERIALIZED_INSTANCE_SUFFIX) }
+        .flatMap { case (key, value) =>
+          val serdeName = key.replace(SerializerConfig.SERIALIZED_INSTANCE_SUFFIX, "")
+          debug(s"Trying to deserialize serde instance for $serdeName")
```
jmakes: TIL: scala string interpolation is pretty neat 👍

samza-test/src/test/java/org/apache/samza/test/operator/RepartitionWindowApp.java
prateekm: The major use cases for these default serdes in my mind were:
1. For the keySerde, when the key type is always the same (e.g. String/Integer).
2. For the msgSerde, when the data is avro/json/pegasus/protobuf and the Serde can figure out the schema from the message contents.
For DAGs of low to medium complexity, I don't think default serdes per type would be much more concise than specifying them inline. It's also non-trivial to infer the concrete type M for a given MessageStream at compile/planning time, so I don't know of a simple implementation other than checking the message type at runtime. Maybe we can add them later if we find that we need them?

Re: default serde == NoOpSerde vs null (and throwing on use). This version optimizes for users of Systems that do the serialization themselves. E.g., at LinkedIn, this is the case for the Kafka SystemProducer/SystemConsumer where avro is enforced and user-provided serdes are disallowed/ignored. I didn't want to make this yet another configuration. Maybe we can make this null by default and patch it internally to be NoOpSerde instead? @jmakes @nickpan47, any thoughts on this?

Agree that we should add more information around ClassCastException where possible. Currently it's hard to find the source operator for the exception from the stacktrace.

prateekm:
@jmakes 
`I assume nothing's changed for low level api. It's still config driven for all serdes, right?`
Yeah, should not affect low level api.

samza-test/src/test/java/org/apache/samza/test/operator/RepartitionWindowApp.java
jmakes: Agree with the use cases and with these methods being optional s.t. if they add value, use them, if not, don't.

I'm not sure I understand the difference between PassthroughSerde and NoOpSerde. They sound the same to me. 

Do you think it's useful to expose a typed NoOpSerde<> so users that do their own serialization can hint the type info expected by their implementation? e.g. at LinkedIn we could enforce NoOpSerde<IndexedRecord>. I honestly didn't look close enough at the code to verify whether this is useful, just throwing it out there.

I don't think we have much data to say whether or not the defaults will be more or less useful. With avro specific records, you could imagine each external stream having a different type and therefore another serde instance. However, I think a number of jobs at LI still use generic records, and for them this might be more useful.

prateekm: @jmakes In this context, pass through == byte[] in byte[] out, whereas no-op == don't attempt to serialize at all. Maybe NoSerde is better?

The NoOpSerde is already typed in this PR. However, note that if you set the default serde, you don't get any type inference when creating input/output streams. You have to provide the type of the MessageStream contents manually and correctly based on the default serde.

prateekm:
@cpettitt-linkedin @jmakes @nickpan47 @vjagadish1989, please take a look again.

Major Changes:
1. Changed the StreamGraph and MessageStream APIs such that they now work with both keyed and unkeyed streams. Only the 'repartition' (formerly partitionBy) operator enforces any key-value type restrictions. The rest depend on the type of Serde provided for the input/output stream.
2. Moved Serde classes from samza-core to samza-api.
3. Changed the moved JsonSerde to use the default ObjectMapper instead of SamzaObjectMapper. See TODO 1 for further work required for this.
4. Removed the restriction that users can't specify serdes in config. This is required for sinkTo to work since we don't control what streams they can output to. There are workarounds, but they're kinda hacky. Serdes provided in high level api code still take priority over those provided in configs.

To do:
1. Create an SamzaJsonSerde that uses SamzaObjectMapper internally and use it for coordinator/logger etc. TBD what the "json" serde config maps to. Probably the internal one, in which case we need a new name for the new serde.
2. Add more information about ClassCastExceptions where possible.
3. Tests.

samza-core/src/main/java/org/apache/samza/operators/impl/InputOperatorImpl.java
```diff
@@ -35,13 +35,12 @@
  *
  * @param <K> the type of key in the incoming message
  * @param <V> the type of message in the incoming message
- * @param <M> the type of input message
  */
-public final class InputOperatorImpl<K, V, M> extends OperatorImpl<Pair<K, V>, M> {
+public final class InputOperatorImpl<K, V> extends OperatorImpl<KV<K, V>, Object> { // Object == KV<K,V> | V
```
jmakes: Is there another InputOperator* for unkeyed inputs?

samza-api/src/main/scala/org.apache.samza.serializers/KVSerde.scala
```diff
@@ -0,0 +1,55 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.serializers
+
+import org.apache.samza.SamzaException
+import org.apache.samza.operators.KV
+
+object KVSerde {
```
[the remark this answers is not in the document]
prateekm: Just because the existing Serde implementations are in Scala and I didn't want to change them all to Java while moving them to a different package. I'll do that in a separate PR.

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -133,17 +135,24 @@ public void sink(SinkFunction<? super M> sinkFn) {
...
+      Function<? super M, ? extends V> valueExtractor, KVSerde<K, V> serde) {
     int opId = this.graph.getNextOpId();
     String opName = String.format("%s-%s", OperatorSpec.OpCode.PARTITION_BY.name().toLowerCase(), opId);
-    IntermediateMessageStreamImpl<K, M, M> intermediateStream =
-        this.graph.getIntermediateStream(opName, keyExtractor, m -> m, (k, m) -> m);
-    OutputOperatorSpec<M> partitionByOperatorSpec = OperatorSpecs.createPartitionByOperatorSpec(
-        intermediateStream.getOutputStream(), opId);
+    IntermediateMessageStreamImpl<KV<K, V>> intermediateStream = this.graph.getIntermediateStream(opName, serde);
+    PartitionByOperatorSpec<M, K, V> partitionByOperatorSpec =
+        OperatorSpecs.createRepartitionOperatorSpec(
```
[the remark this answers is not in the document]
prateekm: I'll revert this, was part of an experimental renaming.

samza-core/src/main/java/org/apache/samza/operators/impl/InputOperatorImpl.java
prateekm: Not really. The messages we get as an input to this operator are always keyed, with the key being that of InputMessageEnvelope. This operator is where we decide whether to propagate the key further, based on whether the input stream Serde is keyed or not.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -65,46 +71,51 @@ public StreamGraphImpl(ApplicationRunner runner, Config config) {
...
+    Preconditions.checkNotNull(serde, "serde must not be null for an input stream.");
+    Preconditions.checkState(!inputOperators.containsKey(runner.getStreamSpec(streamId)),
+        "getInputStream must not be called multiple times with the same streamId: " + streamId);
 
     StreamSpec streamSpec = runner.getStreamSpec(streamId);
+    KV<Serde, Serde> kvSerdes = getKVSerdes(streamId, serde);
+    boolean isKeyedInput = serde instanceof KVSerde;
     inputOperators.put(streamSpec,
-        new InputOperatorSpec<>(streamSpec, (BiFunction<K, V, M>) msgBuilder, this.getNextOpId()));
+        new InputOperatorSpec<>(streamSpec, kvSerdes.getKey(), kvSerdes.getValue(), isKeyedInput, this.getNextOpId()));
```
vjagadish1989: Can this `instance of` check be done in the constructor. You would n't need the extra param to the constructor then?

Same for `OutputStream`s

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -116,38 +127,31 @@ public StreamGraph withContextManager(ContextManager contextManager) {
...
    * @param streamName the name of the stream to be created. Will be prefixed with job name and id to generate the
    *                   logical streamId.
-   * @param keyExtractor the {@link Function} to extract the outgoing key from the intermediate message
-   * @param msgExtractor the {@link Function} to extract the outgoing message from the intermediate message
-   * @param msgBuilder the {@link BiFunction} to convert the incoming key and message to a message
-   *                   in the intermediate {@link MessageStream}
-   * @param <K> the type of key in the intermediate message
-   * @param <V> the type of message in the intermediate message
+   * @param serde the {@link Serde} to use for the message in the intermediate stream. If null, the default serde
+   *              is used.
```
vjagadish1989: Minor: can add `{@link #setDefaultSerde}` to provide a default serde

samza-core/src/main/java/org/apache/samza/operators/impl/PartitionByOperatorImpl.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.samza.system.SystemStream;
+import org.apache.samza.task.MessageCollector;
+import org.apache.samza.task.TaskContext;
+import org.apache.samza.task.TaskCoordinator;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.function.Function;
+
+
```
vjagadish1989: minor: no new line

samza-core/src/main/java/org/apache/samza/operators/spec/PartitionByOperatorSpec.java
```diff
@@ -0,0 +1,76 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.operators.spec;
+
+import org.apache.samza.operators.KV;
+
+import java.util.function.Function;
+
+
```
vjagadish1989: minor: no new line

samza-core/src/test/java/org/apache/samza/execution/TestExecutionPlanner.java
```diff
@@ -145,13 +147,21 @@ private StreamGraphImpl createStreamGraphWithJoin() {
...
 
     StreamGraphImpl streamGraph = new StreamGraphImpl(runner, config);
-    BiFunction msgBuilder = mock(BiFunction.class);
-    MessageStream m1 = streamGraph.getInputStream("input1", msgBuilder).map(m -> m);
-    MessageStream m2 = streamGraph.getInputStream("input2", msgBuilder).partitionBy(m -> "haha").filter(m -> true);
-    MessageStream m3 = streamGraph.getInputStream("input3", msgBuilder).filter(m -> true).partitionBy(m -> "hehe").map(m -> m);
-    Function mockFn = mock(Function.class);
-    OutputStream<Object, Object, Object> output1 = streamGraph.getOutputStream("output1", mockFn, mockFn);
-    OutputStream<Object, Object, Object> output2 = streamGraph.getOutputStream("output2", mockFn, mockFn);
+    streamGraph.setDefaultSerde(KVSerde.of(new NoOpSerde<>(), new NoOpSerde<>()));
```
vjagadish1989: Do we need this explicitly set to its default value?

samza-core/src/test/java/org/apache/samza/operators/TestMessageStreamImpl.java
```diff
@@ -172,43 +172,84 @@ public void testSendTo() {
...
     assertEquals(OpCode.SEND_TO, registeredOpSpec.getOpCode());
-    assertEquals(mockOutputOpSpec, ((OutputOperatorSpec) registeredOpSpec).getOutputStream());
+    assertEquals(mockOutputStreamImpl, ((OutputOperatorSpec) registeredOpSpec).getOutputStream());
+
+    // same behavior as above so nothing new to assert. but ensures that this variant compiles.
+    MessageStreamImpl<KV<String, TestMessageEnvelope>> keyedInputStream = new MessageStreamImpl<>(mockGraph, mockOpSpec);
+    OutputStreamImpl<KV<String, TestMessageEnvelope>> mockKeyedOutputStreamImpl = mock(OutputStreamImpl.class);
+    keyedInputStream.sendTo(mockKeyedOutputStreamImpl);
+
+    // can't unit test it, but the following variants should not compile
```
vjagadish1989: thanks!

vjagadish1989 (approved):
looks great. approved.

Big thanks for drastically simplifying the API, and verifying it with an exhaustive suite of tests!

samza-api/src/main/java/org/apache/samza/operators/KV.java
```diff
@@ -0,0 +1,48 @@
...
+
+
+/**
+ * A key and value pair.
+ *
+ * @param <K> type of the key
+ * @param <V> type of the value
+ */
+public class KV<K, V> {
+  public final K key;
```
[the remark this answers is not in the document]
prateekm: It's sometimes to do a kv.key instead of kv.getKey(). It's ok to allow this usage since the class is immutable. Hence keeping them public.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -182,11 +182,25 @@
...
+   * @param keyExtractor the {@link MapFunction} to extract the message and partition key from the input message
+   * @param valueExtractor the {@link MapFunction} to extract the value from the input message
+   * @param serde the {@link KVSerde} to use for (de)serializing the key and value.
    * @return the repartitioned {@link MessageStream}
    */
-  <K> MessageStream<M> partitionBy(Function<? super M, ? extends K> keyExtractor);
+  <K, V> MessageStream<KV<K, V>> partitionBy(Function<? super M, ? extends K> keyExtractor,
+      Function<? super M, ? extends V> valueExtractor, KVSerde<K, V> serde);
 
+  /**
```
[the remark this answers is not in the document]
prateekm: The documentation for this method is complex enough that I'd like to keep it in one place. This pattern is not uncommon for overloaded methods.

samza-api/src/main/java/org/apache/samza/operators/StreamGraph.java
```diff
@@ -19,48 +19,83 @@
...
 
 /**
- * Provides access to {@link MessageStream}s and {@link OutputStream}s used to describe the processing logic.
+ * Provides access to {@link MessageStream}s and {@link OutputStream}s used to describe application logic.
  */
 @InterfaceStability.Unstable
 public interface StreamGraph {
 
   /**
+   * Sets the default {@link Serde} to use for (de)serializing messages.
```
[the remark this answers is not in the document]
prateekm: Yes, but only before creating any I/O streams. Last writer wins (which is the expected behavior). Worth documenting?

samza-api/src/main/java/org/apache/samza/operators/StreamGraph.java
```diff
@@ -19,48 +19,83 @@
...
+   * <p>.
+   * If the the default serde is set, it must be set <b>before</b> creating any input or output streams.
+   * The default value for the default serde is a KVSerde&lt;NoOpSerde, NoOpSerde&gt;. This means that any streams
+   * created when using the the default value for the default serde will be recieve messages of type KV&lt;Object, Object&gt;.
```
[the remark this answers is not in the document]
prateekm: How about:

If no explicit or default serdes are provided, a NoOpSerde is used for keys and values. This means that any streams created without an explicit or default serde should be cast to MessageStream&ltKV&lt;Object, Object&gt&gt.

samza-api/src/main/java/org/apache/samza/operators/KV.java
```diff
@@ -0,0 +1,48 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.operators;
+
+
```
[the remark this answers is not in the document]
prateekm: Convention (not always followed) seems to be two new lines after imports, so maybe leave as is?

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
prateekm: These constructor doesn't get the KVSerde, only the key and value serdes separately. This is because the defaulting for the key/message serde based on the default serde happens here. Maybe there's another way to avoid this?

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
prateekm: This is a protected method called by MessageStreamImpl (i.e. us), so we shouldn't be setting the default serde before calling. Hence this is only documenting that the default serde (whether set or the NoOp) will be used. Will add the link if we ever expose this in public API.

samza-api/src/main/java/org/apache/samza/operators/StreamGraph.java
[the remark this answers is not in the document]
vjagadish1989: Not worth documenting then. i was just wondering!

samza-api/src/main/java/org/apache/samza/operators/StreamGraph.java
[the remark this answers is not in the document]
vjagadish1989: like this much better, thanks! If you're fine please feel free to update it.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
vjagadish1989: makes sense, probably not worth doing then

vjagadish1989 (approved):
looks good to me. approved. thanks!

samza-api/src/main/java/org/apache/samza/operators/KV.java
```diff
@@ -0,0 +1,48 @@
...
+  public static <K, V> KV<K, V> of(K key, V value) {
+    return new KV<>(key, value);
+  }
+
+  public KV(K key, V value) {
+    this.key = key;
+    this.value = value;
+  }
+
+  public K getKey() {
```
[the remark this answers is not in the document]
prateekm: Since this is a public API, I thought it OK to allow users to use it however they prefer.

samza-api/src/main/java/org/apache/samza/operators/StreamGraph.java
```diff
@@ -19,48 +19,85 @@
...
+ * Provides access to {@link MessageStream}s and {@link OutputStream}s used to describe application logic.
  */
 @InterfaceStability.Unstable
 public interface StreamGraph {
 
   /**
+   * Sets the default {@link Serde} to use for (de)serializing messages.
+   * <p>.
+   * If the default serde is set, it must be set <b>before</b> creating any input or output streams.
+   * If no explicit or default serdes are provided, a NoOpSerde is used for keys and values. This means that any
```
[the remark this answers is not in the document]
prateekm: No, I meant to call this out explicitly. The documentation that this is referring to says:
"If the provided {@code serde} is null, uses the default serde provided via {@link StreamGraph#setDefaultSerde}, which must be a KVSerde."

samza-api/src/main/scala/org.apache.samza.serializers/NoOpSerde.scala
```diff
@@ -0,0 +1,37 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.serializers
+
+/**
+  * A marker serde class to indicate that messages should not be serialized or deserialized.
+  * This is the same behavior as when no serde is provided, and is intended for use cases where
+  * a Serde parameter or configuration is required.
```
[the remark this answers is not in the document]
prateekm: I'll write some documentation for "Serialization in High Level API" before the release. The general summary is:
For any streams (e.g. input/output/intermediate) or stores (e.g. join/window) created explictly in the high level API, the only way to specify the serde is in code, either explicitly or using the default serde.
For non-explicit streams (e.g., those produced to in sink) or stores (e.g. in a map with state), the only way to specify serde is in configs.

samza-api/src/main/scala/org.apache.samza.serializers/NoOpSerde.scala
```diff
@@ -0,0 +1,37 @@
...
+  * This is different than [[ByteSerde]] which is a pass-through serde for byte arrays.
+  *
+  * @tparam T type of messages which should not be serialized or deserialized
+  */
+class NoOpSerde[T] extends Serde[T] {
+
+  override def fromBytes(bytes: Array[Byte]): T =
```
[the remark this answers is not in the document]
prateekm: We overwrite the serde configs for any streams created using getInput/Output/IntermediateStream with the serde defined in code (either explicitly or using the default), so this can not happen.

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
```diff
@@ -127,11 +134,88 @@ public JobConfig generateConfig(String executionPlanJson) {
...
+    // for each unique serde instance, generate a unique name and serialize to config
+    HashSet<Serde> serdes = new HashSet<>(keySerdes.values());
+    serdes.addAll(msgSerdes.values());
+    SerializableSerde<Serde> serializableSerde = new SerializableSerde<>();
+    Base64.Encoder base64Encoder = Base64.getEncoder();
+    Map<Serde, String> serdeUUIDs = new HashMap<>();
+    serdes.forEach(serde -> {
+        String serdeName = serdeUUIDs.computeIfAbsent(serde,
+            s -> serde.getClass().getSimpleName() + "-" + UUID.randomUUID().toString());
+        configs.putIfAbsent(String.format(SerializerConfig.SERDE_SERIALIZED_INSTANCE(), serdeName),
```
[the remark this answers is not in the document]
prateekm: In general, we try to preserve the object graph for serdes. E.g., if two instances of an AvroSerde want to share a SchemaRegistryClient, we should be able to do so after deserialization. The JsonSerdeV2 class instance is a similar example, we want the type information to be available in the serde at runtime.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -257,11 +282,15 @@ object SamzaContainer extends Logging {
...
+        .flatMap(systemStream => {
           val serdeName = getSerdeName(systemStream).get
           val serde = serdes.getOrElse(serdeName,
-            throw new SamzaException("buildSystemStreamSerdeMap: No class defined for serde: %s." format serdeName))
-          (systemStream, serde)
+            throw new SamzaException("buildSystemStreamSerdeMap: No serde found for name: %s." format serdeName))
+
+          // respect explicitly set no-op serdes in high level API
+          Option(serde)
+            .filter(!_.isInstanceOf[NoOpSerde[Any]])
```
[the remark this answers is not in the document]
prateekm: That works too, but I'd prefer to explicitly declare a serde instead of passing nulls around, which is error prone. That said, I'll rename this to NoSerde to clarify its intent.

samza-core/src/main/java/org/apache/samza/operators/spec/InputOperatorSpec.java
```diff
@@ -18,35 +18,46 @@
...
+public class InputOperatorSpec<K, V> extends OperatorSpec<KV<K, V>, Object> { // Object == KV<K, V> | V
 
   private final StreamSpec streamSpec;
-  private final BiFunction<K, V, M> msgBuilder;
+  private final Serde<K> keySerde;
+  private final Serde<V> valueSerde;
+  private final boolean isKeyedInput;
 
-  public InputOperatorSpec(StreamSpec streamSpec, BiFunction<K, V, M> msgBuilder, int opId) {
+  public InputOperatorSpec(StreamSpec streamSpec,
```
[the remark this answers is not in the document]
prateekm: We talked offline. There's not much meaningful type checking we can do between spec and impl DAGs in the current implementation since the upper bound for message types in specs is Object. The best we can do for now is make the types _within_ the spec and impl DAG consistent. We'll revisit when we have the StreamDescriptor/StreamApplication API.

PR #296, opened 2017-09-15
SAMZA-1416 : Better logging around the exception where class loading failed in initializing the SystemFactory for a input/output system

dnishimura:
Also added test coverage for the Util.getObj method.
@nickpan47 @jmakes

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/util/Util.scala
+++ b/samza-core/src/main/scala/org/apache/samza/util/Util.scala
@@ -81,10 +81,17 @@ object Util extends Logging {
    * Instantiate a class instance from a given className.
    */
   def getObj[T](className: String) = {
-    Class
-      .forName(className)
-      .newInstance
-      .asInstanceOf[T]
+    try {
+      Class
+        .forName(className)
+        .newInstance
+        .asInstanceOf[T]
+    } catch {
+      case e: Throwable => {
+        error("Unable to instantiate a class instance for %s." format className, e)
+        throw e
+      }
+    }
   }
 
   /**
--- a/samza-core/src/main/java/org/apache/samza/config/JavaSystemConfig.java
+++ b/samza-core/src/main/java/org/apache/samza/config/JavaSystemConfig.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.samza.SamzaException;
 import org.apache.samza.system.SystemAdmin;
 import org.apache.samza.system.SystemFactory;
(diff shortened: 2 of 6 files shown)


samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -81,10 +81,17 @@ object Util extends Logging {
...
-      .forName(className)
-      .newInstance
-      .asInstanceOf[T]
+    try {
+      Class
+        .forName(className)
+        .newInstance
+        .asInstanceOf[T]
+    } catch {
+      case e: Throwable => {
```
navina: Is it better to handle specific exceptions here, rather than `Throwables` ?

dnishimura: Normally I would catch specific exceptions if the exception handling was more extensive, but the intent is just to log the name of the class (which is currently not reported in the logs) that caused the error before a throw. On the other hand, specific exceptions could be caught with slightly different error messages, but I'm not sure what other information could be logged that's not already conveyed in the stack trace reported from the exception. Thoughts? Thanks for reviewing!

jmakes:
This patch looks useful, but looking at the description in the JIRA, I don't think it's sufficient.

1. When the system factory config exists but is an empty string, the class name is an empty string. So the log message in this patch won't be very informative. 
2. It would be better if the log message informed the user which system has the misconfigured system factory. 

One way to do that is to update SystemConfig.getSystemFactory() to validate against empty strings. 

Another option is to leverage some variant of ScalaMapConfig.getNonEmptyOption(), though that might cause us to silently ignore a configured system. So some adaptation would be required to instead print a log message.

dnishimura:
@jmakes Thanks for the feedback. I pushed another commit to utilize the `getNonEmptyOption()` method to return `None` if the class name is an empty string. This triggers the appropriate logging via `SamzaException`s and logs the misconfigured system factory.

jmakes:
@dnishimura Looking better, but if we really want to put this issue to rest, it looks like we'll need a similar change in JavaSystemConfig.java

*sigh* we really shouldn't have both and the Samza project has vowed to use Java for everything new. But it looks like at least for now there's some duplication in this area.

Apologies for the trouble, but would you mind either making the analogous change in JavaSystemConfig or doing a little refactoring so there's only one method to fetch the systemFactory?

dnishimura:
@jmakes I was wondering why there are Scala and Java system config classes. I'll attempt some refactoring so other newbies don't have to walk the same path.

dnishimura:
@jmakes I removed duplicate code between `SystemConfig` and `JavaSystemConfig`. It'll probably be a bigger job to completely merge the two, but hopefully this is sufficient enough for this PR.

dnishimura:
Hm don't know this commit failed. Log shows: `Failed to mkdirs: /home/jenkins/jenkins-slave/workspace/samza-pr-checks` as the error.

PR #297, opened 2017-09-19
SAMZA-1417: Clear and recreate intermediate and metadata streams for batch processing

xinyuiscool:
For each run of a batch application, we need to clear the internal streams from the previous run and recreate new ones. This patch introduces the following:
1) bounded flag in StreamSpec
2) app.mode (BATCH/STREAM) in the application config. An application is a batch app iff all the input streams are bounded.
3) app.runId and use it to generate the internal topics for each run.

run.id generation is not addressed in this pr. There will be another patch to resolve it for both yarn and standalone. For now, this patch only works for yarn.


xinyuiscool:
@jmakes : could you please help take a look? Thanks!

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
```diff
@@ -60,12 +69,17 @@ void addTargetNode(JobNode targetNode) {
...
-    } else {
-      return streamSpec.copyWithPartitionCount(partitions);
+  StreamSpec getStreamSpec() {
+    StreamSpec spec = (partitions == PARTITIONS_UNKNOWN) ?
+        streamSpec : streamSpec.copyWithPartitionCount(partitions);
+
+    if (isIntermediate) {
+      ApplicationConfig appConfig = new ApplicationConfig(config);
+      if (appConfig.getAppMode() == ApplicationConfig.ApplicationMode.BATCH && appConfig.getRunId() != null) {
+        spec = spec.copyWithPhysicalName(spec.getPhysicalName() + "-" + appConfig.getRunId());
```
jmakes: This logic is somewhat similar to the logic added to JavaStreamConfig. 

Is there any way to put it in a central place? 

Not sure if I like this idea, but perhaps a BatchStreamSpec class that either extends StreamSpec (like KafkaStreamSpec does) or just provides some static methods. Thoughts?

samza-core/src/main/java/org/apache/samza/execution/StreamManager.java
```diff
@@ -75,4 +85,52 @@ public void createStreams(List<StreamSpec> streams) {
...
+      //Find checkpoint stream and clean up
+      TaskConfig taskConfig = new TaskConfig(prevConfig);
+      String checkpointManagerFactoryClass = taskConfig.getCheckpointManagerFactory().getOrElse(defaultValue(null));
+      if (checkpointManagerFactoryClass != null) {
+        CheckpointManager checkpointManager = ((CheckpointManagerFactory) Util.getObj(checkpointManagerFactoryClass))
+            .getCheckpointManager(prevConfig, new MetricsRegistryMap());
+        checkpointManager.clearCheckpoints();
+      }
+
+      //Find changelog streams and remove them
```
jmakes: This is only sufficient if host affinity is disabled. Where do we enforce that?

samza-core/src/main/java/org/apache/samza/execution/StreamManager.java
```diff
@@ -75,4 +85,52 @@ public void createStreams(List<StreamSpec> streams) {
...
+        String changelog = storageConfig.getChangelogStream(store).getOrElse(defaultValue(null));
+        if (changelog != null) {
+          LOGGER.info("Clear store {} changelog {}", store, changelog);
+          SystemStream systemStream = Util.getSystemStreamFromNames(changelog);
+          StreamSpec spec = StreamSpec.createChangeLogStreamSpec(systemStream.getStream(), systemStream.getSystem(), 1);
+          sysAdmins.get(spec.getSystemName()).clearStream(spec);
+        }
+      }
+    } catch (Exception e) {
+      LOGGER.warn("Fail to clear internal streams from previous run. Please clean up manually.", e);
```
jmakes: I think we should rethrow here, since we can't guarantee correctness. Thoughts?

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
```diff
@@ -37,6 +37,7 @@ object StreamConfig {
   val BOOTSTRAP =               SAMZA_PROPERTY + "bootstrap"
   val PRIORITY =                SAMZA_PROPERTY + "priority"
   val IS_INTERMEDIATE =         SAMZA_PROPERTY + "intermediate"
+  val IS_BOUNDED =              SAMZA_PROPERTY + "bounded"
```
jmakes: This might be ok for now, but it feels like we should determine whether a stream is bounded from metadata, or perhaps its type. (e.g. I can't think of how an unbounded HDFS file could exist)

samza-core/src/main/java/org/apache/samza/execution/StreamManager.java
xinyuiscool: good point. I added the check inside ExecutionPlanner.validateConfig().

samza-core/src/main/java/org/apache/samza/execution/StreamManager.java
xinyuiscool: this should be fine. We always generate a new set of topics (appending the run.id so they are different from previous runs). I actually expect clearStreams not supported in many scenarios, where we might rely on retention to get the streams cleared.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
xinyuiscool: I assume that's the job of StreamDescriptor :). It should generate the StreamSpec correctly.

This is also used in Beam where I translate BoundedSource to a bounded StreamSpec.

samza-core/src/main/java/org/apache/samza/execution/StreamManager.java
jmakes: Ahh, forgot about that. 👍 

So this is just to free up resources rather than to ensure correctness. I'd suggest mentioning that in a comment.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
jmakes: I should have been more specific. We absolutely need the notion of whether a stream is bounded or not. 

Rather, my concern was about adding such a property to config.

PR #299, opened 2017-09-21
SAMZA-1427: use systemFactory in checkpoint manager.

sborya:


samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,71 @@ class KafkaCheckpointManager(
...
-              KafkaUtil.maybeThrowException(errorCode)
+    val ssp: SystemStreamPartition = new SystemStreamPartition(systemName, checkpointTopic, new Partition(0))
+    val systemConsumer = getSystemConsumer()
+    val offset = getEarliestOffset(checkpointTopic, new Partition(0))
+    systemConsumer.register(ssp, offset); // checkpoint stream should always be read from the beginning
+    systemConsumer.start();
+
+    var count = 0
+    try {
+      var done = false;
```
fredji97: remove ; by convention.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,71 @@ class KafkaCheckpointManager(
...
+      while (!done) {
+        // Map[SystemStreamPartition, List[IncomingMessageEnvelope]]
+        val envelops = systemConsumer.poll(Collections.singleton(ssp), 1000)
+        debug("CheckpointMgr read %d envelops" format envelops.size());
```
fredji97: remove ; by convention, including the following several lines;

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,71 @@ class KafkaCheckpointManager(
...
   /**
-   * Common code for reading both changelog partition mapping and change log
+   * Code for reading checkpoint log
    *
    * @param entryType What type of entry to look for within the log key's
    * @param handleEntry Code to handle an entry in the log once it's found
    */
   private def readLog(entryType:String, shouldHandleEntry: (KafkaCheckpointLogKey) => Boolean,
                       handleEntry: (ByteBuffer, KafkaCheckpointLogKey) => Unit): Unit = {
-    retryBackoff.run[Unit](
```
fredji97: why do you remove retryBackoff?

samza-kafka/src/test/scala/org/apache/samza/checkpoint/kafka/TestKafkaCheckpointManager.scala
```diff
@@ -19,22 +19,22 @@
...
 package org.apache.samza.checkpoint.kafka
 
-import kafka.admin.AdminUtils
-import kafka.common.{InvalidMessageSizeException, UnknownTopicOrPartitionException}
-import kafka.message.InvalidMessageException
-import kafka.server.{KafkaConfig, KafkaServer, ConfigType}
-import kafka.utils.{CoreUtils, TestUtils, ZkUtils}
-import kafka.integration.KafkaServerTestHarness
+import _root_.kafka.admin.AdminUtils
+import _root_.kafka.common.{InvalidMessageSizeException, UnknownTopicOrPartitionException}
```
fredji97: What issue are you seeing without _root_?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
```diff
@@ -62,16 +64,23 @@ class KafkaCheckpointManagerFactory extends CheckpointManagerFactory with Loggin
...
+    val (systemName: String, systemFactory : SystemFactory) =  org.apache.samza.util.Util.getCheckpointSystemStreamAndFactory(config)
 
     val producerConfig = config.getKafkaSystemProducerConfig(
       systemName,
       clientId,
       INJECTED_PRODUCER_PROPERTIES)
-    val connectProducer = () => {
-      new KafkaProducer[Array[Byte], Array[Byte]](producerConfig.getProducerProperties)
+
+    val systemProducerF = () => {
```
fredji97: would it be better to call it getSystemProducer instead of systemProducerF?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: It used to retry direct requests to the Kafka brokers, but systemConsumer makes poll() requests to the BlockingEnvelopeMap, and, since it is local only call, there is no point in retries.
Same if we go to a high level consumer in Kafka. It will take care of retries.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done.

samza-kafka/src/test/scala/org/apache/samza/checkpoint/kafka/TestKafkaCheckpointManager.scala
sborya: I don't see any. This is done by "optimize import" in Intelij.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
sborya: Yep, already done it.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -21,23 +21,20 @@ package org.apache.samza.checkpoint.kafka
...
 
+import _root_.kafka.consumer.SimpleConsumer
+import _root_.kafka.utils.ZkUtils
 import org.apache.kafka.common.utils.Utils
-import org.apache.kafka.clients.producer.{Producer, ProducerRecord}
-import org.apache.samza.SamzaException
 import org.apache.samza.checkpoint.{Checkpoint, CheckpointManager}
 import org.apache.samza.container.TaskName
 import org.apache.samza.serializers.CheckpointSerde
+import org.apache.samza.system._
```
vjagadish1989: No blanket imports

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -66,16 +65,17 @@ class KafkaCheckpointManager(
   import org.apache.samza.checkpoint.kafka.KafkaCheckpointManager._
 
   var taskNames = Set[TaskName]()
-  var producer: Producer[Array[Byte], Array[Byte]] = null
+  var systemProducer: SystemProducer = null
   var taskNamesToOffsets: Map[TaskName, Checkpoint] = null
 
-  var startingOffset: Option[Long] = None // Where to start reading for each subsequent call of readCheckpoint
+  //var startingOffset: Option[Long] = None // Where to start reading for each subsequent call of readCheckpoint
```
vjagadish1989: Remove unused code

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -66,16 +65,17 @@ class KafkaCheckpointManager(
...
   val kafkaUtil: KafkaUtil = new KafkaUtil(retryBackoff, connectZk)
 
   KafkaCheckpointLogKey.setSystemStreamPartitionGrouperFactoryString(systemStreamPartitionGrouperFactoryString)
 
   info("Creating KafkaCheckpointManager with: clientId=%s, checkpointTopic=%s, systemName=%s" format(clientId, checkpointTopic, systemName))
 
+
```
vjagadish1989: remove extra new line

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
     checkpoints.toMap /* of the immutable kind */
   }
 
+  private def getEarliestOffset(topic: String, partition: Partition): (String, String) = {
```
vjagadish1989: Minor: Prefer not returning tuples and returning `partitionMetadata`

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+        "-1"
+      else
+        partitionMetaData.getNewestOffset()
+
+    (partitionMetaData.getOldestOffset, newestOffset)
+  }
 
   /**
-   * Common code for reading both changelog partition mapping and change log
+   * Code for reading checkpoint log
```
vjagadish1989: Minor: (rewrite docs as follows)
`Reads an entry from the checkpoint log and invokes the provided lambda on that entry`

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+    systemConsumer.register(ssp, oldestOffset); // checkpoint stream should always be read from the beginning
+    systemConsumer.start();
+
+    var count = 0
+    try {
+      var done = false
+      var currentOffset = -1L
+      val newestOffsetLong = newestOffset.toLong
+      while (currentOffset < newestOffsetLong) {
+        // Map[SystemStreamPartition, List[IncomingMessageEnvelope]]
```
vjagadish1989: remove comment

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+        val envelopes = systemConsumer.poll(Collections.singleton(ssp), 1000)
```
vjagadish1989: Use a final variable here instead of a hard coded value for timeout. Should we provide a more generous timeout?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+
+        val messages: util.List[IncomingMessageEnvelope] = envelopes.get(ssp)
+        val messagesSize = if (messages != null) messages.size else -1
+        info("CheckpointMgr read %s envelopes (%s messages) from ssp %s"
+                     format (envelopes.size(), messagesSize, ssp))
```
vjagadish1989: Why is envelopes.size useful to log? Wouldn't it always be one given that we are polling for a single SSP?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
```diff
@@ -91,8 +99,10 @@ class KafkaCheckpointManagerFactory extends CheckpointManagerFactory with Loggin
       socketTimeout,
       consumerConfig.socketReceiveBufferBytes,
       consumerConfig.fetchMessageMaxBytes,            // must be > buffer size
+      systemConsumerF,
```
vjagadish1989: s/systemConsumerF/systemConsumerFn

Similar for producer, and admin

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
```diff
@@ -91,8 +99,10 @@ class KafkaCheckpointManagerFactory extends CheckpointManagerFactory with Loggin
...
+      systemAdminF,
       new ClientUtilTopicMetadataStore(producerConfig.bootsrapServers, clientId, socketTimeout),
-      connectProducer,
+      systemProducerF,
```
vjagadish1989: Why do we take in functions for `systemAdminF`, `systemProducerF` and `systemConsumerF`. Is n't what the `SystemFactory` is doing with its methods for `createProducer`, `createConsumer` and `createAdmin`? It might be simpler to take in the `SystemFactory` right?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -21,23 +21,20 @@ package org.apache.samza.checkpoint.kafka
...
-import java.util.Properties
-
-import kafka.api._
-import kafka.common.{ErrorMapping, InvalidMessageSizeException, TopicAndPartition, UnknownTopicOrPartitionException}
-import kafka.consumer.SimpleConsumer
-import kafka.message.InvalidMessageException
-import kafka.utils.ZkUtils
+import java.util.{Collections, Properties}
 
+import _root_.kafka.consumer.SimpleConsumer
```
vjagadish1989: Why do we still need the `SimpleConsumer` import here?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -88,21 +88,26 @@ class KafkaCheckpointManager(
...
       (exception, loop) => {
         warn("Failed to write %s partition entry %s: %s. Retrying." format(CHECKPOINT_LOG4J_ENTRY, key, exception))
         debug("Exception detail:", exception)
-        if (producer != null) {
-          producer.close
+        if (systemProducer != null) {
+          systemProducer.stop
         }
-        producer = null
+        systemProducer = null
```
vjagadish1989: Given that we are no longer using the `SimpleConsumer` can we get rid off the `getConsumer` code completely? that will simplify this class further

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+        val envelopes = systemConsumer.poll(Collections.singleton(ssp), 1000)
+
+        val messages: util.List[IncomingMessageEnvelope] = envelopes.get(ssp)
+        val messagesSize = if (messages != null) messages.size else -1
+        info("CheckpointMgr read %s envelopes (%s messages) from ssp %s"
+                     format (envelopes.size(), messagesSize, ssp))
+        if (envelopes.isEmpty || messages.isEmpty) {
+          // fully read the log
+          info("Got empty list of messages")
+          //done = true
```
vjagadish1989: remove commented code

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
-              }
-              KafkaUtil.maybeThrowException(errorCode)
+    val ssp: SystemStreamPartition = new SystemStreamPartition(systemName, checkpointTopic, new Partition(0))
+    val systemConsumer = getSystemConsumer()
+    val (oldestOffset, newestOffset) = getEarliestOffset(checkpointTopic, new Partition(0))
+    systemConsumer.register(ssp, oldestOffset); // checkpoint stream should always be read from the beginning
+    systemConsumer.start();
+
+    var count = 0
+    try {
```
vjagadish1989: Refactor this by adding the try inside the while loop? Also, the variable `done` does not appear to be used? Are we using it to exit the loop?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
-
-              val checkpointKey = KafkaCheckpointLogKey.fromBytes(Utils.readBytes(response.message.key))
-
-              if (!shouldHandleEntry(checkpointKey)) {
-                debug("Skipping " + entryType + " entry with key " + checkpointKey)
-              } else {
-                handleEntry(response.message.payload, checkpointKey)
-              }
+            else {
+              val bbuf: java.nio.ByteBuffer = ByteBuffer.wrap(msg.getMessage.asInstanceOf[Array[Byte]])
```
vjagadish1989: Prefer variables to be named what they actually mean semantically (instead of what their types are):
s/bbuf/checkpointPayLoad

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
-              }
-
-              val checkpointKey = KafkaCheckpointLogKey.fromBytes(Utils.readBytes(response.message.key))
-
-              if (!shouldHandleEntry(checkpointKey)) {
-                debug("Skipping " + entryType + " entry with key " + checkpointKey)
-              } else {
-                handleEntry(response.message.payload, checkpointKey)
-              }
+            else {
```
vjagadish1989: code style: `else` in prev line

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
-            warn("While trying to read last %s entry for topic %s and partition 0: %s. Retrying." format(entryType, checkpointTopic, e))
-            debug("Exception detail:", e)
         }
       }
-    ).getOrElse(throw new SamzaException("Failed to get entries for " + entryType + " from topic " + checkpointTopic))
-
+    } finally {
+      systemConsumer.stop();
+    }
+    info("done reading %s msgs from checkpoint stream  %s:%s" format(count, systemName, checkpointTopic))
```
vjagadish1989: s/done/Done, s/msg/messages
Re-write log line for consistency:
"Done reading 10 messages from checkpoint system: foo stream: bar"

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+          for (msg: IncomingMessageEnvelope <- messages) {
+            val key = msg.getKey.asInstanceOf[Array[Byte]]
+            currentOffset = msg.getOffset().toLong
+            if (key == null)
+              throw new KafkaUtilException("While reading checkpoint stream encountered message without key.")
+
+            val checkpointKey = KafkaCheckpointLogKey.fromBytes(key)
+
+            if (!shouldHandleEntry(checkpointKey)) {
+              info("Skipping " + entryType + " entry with key " + checkpointKey)
```
vjagadish1989: Use format vs string concat consistently in log lines.  (Prefer format)

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+        val messagesSize = if (messages != null) messages.size else -1
+        info("CheckpointMgr read %s envelopes (%s messages) from ssp %s"
+                     format (envelopes.size(), messagesSize, ssp))
+        if (envelopes.isEmpty || messages.isEmpty) {
+          // fully read the log
+          info("Got empty list of messages")
+          //done = true
+        }
+        else {
+          count += messages.size()
```
vjagadish1989: s/count/msgCount

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+        val messages: util.List[IncomingMessageEnvelope] = envelopes.get(ssp)
+        val messagesSize = if (messages != null) messages.size else -1
+        info("CheckpointMgr read %s envelopes (%s messages) from ssp %s"
+                     format (envelopes.size(), messagesSize, ssp))
+        if (envelopes.isEmpty || messages.isEmpty) {
+          // fully read the log
+          info("Got empty list of messages")
+          //done = true
+        }
+        else {
```
vjagadish1989: code style: `else` in prev line

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+          count += messages.size()
+          // check the key
+          for (msg: IncomingMessageEnvelope <- messages) {
+            val key = msg.getKey.asInstanceOf[Array[Byte]]
+            currentOffset = msg.getOffset().toLong
+            if (key == null)
```
vjagadish1989: Prefer to wrap in `{}` to make code immune to accidental refactors later

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
```diff
@@ -62,16 +64,23 @@ class KafkaCheckpointManagerFactory extends CheckpointManagerFactory with Loggin
     val jobName = config.getName.getOrElse(throw new SamzaException("Missing job name in configs"))
     val jobId = config.getJobId.getOrElse("1")
 
-    val systemName = config
-      .getCheckpointSystem
-      .getOrElse(throw new SamzaException("no system defined for Kafka's checkpoint manager."))
+    val (systemName: String, systemFactory : SystemFactory) =  org.apache.samza.util.Util.getCheckpointSystemStreamAndFactory(config)
```
vjagadish1989: Can't we import `Util` here instead of a fully qualified name here?

samza-kafka/src/test/scala/org/apache/samza/checkpoint/kafka/TestKafkaCheckpointManager.scala
```diff
@@ -19,22 +19,21 @@
...
 
-import kafka.admin.AdminUtils
-import kafka.common.{InvalidMessageSizeException, UnknownTopicOrPartitionException}
-import kafka.message.InvalidMessageException
-import kafka.server.{KafkaConfig, KafkaServer, ConfigType}
-import kafka.utils.{CoreUtils, TestUtils, ZkUtils}
-import kafka.integration.KafkaServerTestHarness
-
-import org.apache.kafka.common.security.JaasUtils
+import _root_.kafka.admin.AdminUtils
```
vjagadish1989: Why's this `_root_` convention needed?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -88,21 +88,26 @@ class KafkaCheckpointManager(
...
         loop.done
       },
 
       (exception, loop) => {
         warn("Failed to write %s partition entry %s: %s. Retrying." format(CHECKPOINT_LOG4J_ENTRY, key, exception))
         debug("Exception detail:", exception)
-        if (producer != null) {
-          producer.close
+        if (systemProducer != null) {
+          systemProducer.stop
```
vjagadish1989: iiuc, multiple threads can't call write checkpoint on the checkpoint manager right? So, there's no synchronization needed on the producer? (even with async commit enabled)
cc: @prateekm to confirm

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
prateekm: Can extract Collections.singleton(ssp) outside the while loop.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
prateekm: @vjagadish1989 From SystemConsumer:
"If no messages are available for a SystemStreamPartition that was supplied in the polling set, the map will not contain a key for the SystemStreamPartition."

However, messages == null sounds like a bug.

Also, s/messagesSize/numMessages

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+      while (currentOffset < newestOffsetLong) {
+        // Map[SystemStreamPartition, List[IncomingMessageEnvelope]]
+        val envelopes = systemConsumer.poll(Collections.singleton(ssp), 1000)
+
+        val messages: util.List[IncomingMessageEnvelope] = envelopes.get(ssp)
+        val messagesSize = if (messages != null) messages.size else -1
+        info("CheckpointMgr read %s envelopes (%s messages) from ssp %s"
+                     format (envelopes.size(), messagesSize, ssp))
+        if (envelopes.isEmpty || messages.isEmpty) {
+          // fully read the log
```
prateekm: Don't think the "fully read the log" comment is true. E.g. if there were no results within poll timeout (maybe due to a slow network fetch) but we're not caught up with the ssp. Maybe remove this comment. Also add more context (e.g. currentOffset, newestOffset) to the info log.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+          //done = true
+        }
+        else {
+          count += messages.size()
+          // check the key
+          for (msg: IncomingMessageEnvelope <- messages) {
+            val key = msg.getKey.asInstanceOf[Array[Byte]]
+            currentOffset = msg.getOffset().toLong
+            if (key == null)
+              throw new KafkaUtilException("While reading checkpoint stream encountered message without key.")
```
prateekm: Add currentOffset to exception message.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
prateekm: Also log currentOffset for skipped message.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
prateekm: Maybe checkpointMessage? Or 'checkpointPayload'. Don't need type annotation.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+  }
 
   /**
-   * Common code for reading both changelog partition mapping and change log
+   * Code for reading checkpoint log
    *
    * @param entryType What type of entry to look for within the log key's
    * @param handleEntry Code to handle an entry in the log once it's found
    */
   private def readLog(entryType:String, shouldHandleEntry: (KafkaCheckpointLogKey) => Boolean,
```
prateekm: Not related to this PR: Not sure if this entryType constant/indirection is useful. Looks like it's just used within this class as a constant for the string "checkpoint log" in a couple of log lines, could inline and make the code simpler.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+    val ssp: SystemStreamPartition = new SystemStreamPartition(systemName, checkpointTopic, new Partition(0))
+    val systemConsumer = getSystemConsumer()
+    val (oldestOffset, newestOffset) = getEarliestOffset(checkpointTopic, new Partition(0))
+    systemConsumer.register(ssp, oldestOffset); // checkpoint stream should always be read from the beginning
+    systemConsumer.start();
+
+    var count = 0
+    try {
+      var done = false
+      var currentOffset = -1L
```
prateekm: Would be good to extract a constant for this -1L (e.g. UNKNOWN_OFFSET) and use it for this and null newestOffset since there's a relationship b/w the two.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
prateekm: I don't think this should be a util method, can we move that method here?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
prateekm: Prefer not extracting these functions. +1 for Jagadish's suggestion for taking a SystemAdmin. If not, they seem simple/short enough to inline.
If you do decide to extract them, better to rename these to get/createSystemXYZ, and remove the {} and make it a one liner. E.g.:
val createSystemConsumer =  () => systemFactory.getSystemConsumer(...)

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
```diff
@@ -62,16 +64,23 @@ class KafkaCheckpointManagerFactory extends CheckpointManagerFactory with Loggin
...
       INJECTED_PRODUCER_PROPERTIES)
-    val connectProducer = () => {
-      new KafkaProducer[Array[Byte], Array[Byte]](producerConfig.getProducerProperties)
+
+    val systemProducerF = () => {
+      systemFactory.getProducer(systemName, config, new NoOpMetricsRegistry())
+    }
+
+    val systemConsumerF = () => {
+      systemFactory.getConsumer(systemName, config, new NoOpMetricsRegistry())
```
prateekm: Can we share the same NoOpMetricsRegistry?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
prateekm: Good catch @vjagadish1989, this is a bug. We can have concurrent commit for concurrently executing TaskInstances, even without async commit enabled. We need to make this producer recycling thread safe.

If I'm reading this correctly, we retry forever until we succeed? Since there aren't that many errors that can be fixed by recreating the producer, we probably don't need to recycle it (should verify).

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
prateekm: 1000 millis should be fine. We'll be retrying until we hit newestOffset anyway.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -55,8 +52,10 @@ class KafkaCheckpointManager(
                               socketTimeout: Int,
                               bufferSize: Int,
                               fetchSize: Int,
+                              val getSystemConsumer: () => SystemConsumer,
```
prateekm: Don't need the 'val' in params.

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -19,29 +19,21 @@
 
 package org.apache.samza.util
 
-import java.net._
-import java.io._
+import java.io.{InputStreamReader, _}
```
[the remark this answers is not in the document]
sborya: this is Intelij preference, done by 'optimize import'. I will try to change it, but it may get changed back by Intelij again.

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -248,6 +240,25 @@ object Util extends Logging {
   }
 
   /**
+   * Get the Checkpoint System and system factory from the configuration
+   * @param config
+   * @return system name and system factory
+   */
+  def getCheckpointSystemStreamAndFactory(config: Config) = {
```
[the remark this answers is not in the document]
sborya: Well, I did it this way to keep consistency with the corresponding methods for other systems (e.g. getCoordinatorSystemStreamAndFactory).

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -248,6 +240,25 @@ object Util extends Logging {
...
+    val CHECKPOINT_SYSTEM = "task.checkpoint.system"
+
+    val systemName = config.get(CHECKPOINT_SYSTEM.toString, "")
```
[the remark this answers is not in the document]
sborya: no, of course not.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: cleaned up.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: but we do need lots of classes from there: 
SystemConsumer, SystemAdmin, SystemProducer, SystemStream, OutgoingMessageEnvelope and much more..

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: I thought it was needed if we want this to be a member of the class. Looks like it is not required.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: Done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: Done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -88,21 +88,26 @@ class KafkaCheckpointManager(
...
-          producer = connectProducer()
+        if (systemProducer == null) {
+          systemProducer = getSystemProducer()
+          systemProducer.register(taskName.getTaskName)
+          systemProducer.start
         }
 
-        producer.send(new ProducerRecord(checkpointTopic, 0, keyBytes, msgBytes)).get()
+        val systemStream = new SystemStream(systemName, checkpointTopic)
+        val envelope: OutgoingMessageEnvelope = new OutgoingMessageEnvelope(systemStream, keyBytes, msgBytes)
```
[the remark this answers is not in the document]
sborya: done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: hmm, since it is existing code, I guess this problem existed before. 
If we keep recycling, then the only way would be to synchronize the whole thing (all accesses to the producer),
but if we remove the recycling then we only need to sync the initialization (to avoid double creation).
Hopefully the new producer is resilient enough and doesn't need to be recycled.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: sure.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: agree, removed.

dnishimura:
Forgot license header and a couple other things. Adding those things in now.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: removed 'done' and added a try/catch. I don't do anything the try/catch, except printing the exception.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: Done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -187,92 +174,83 @@ class KafkaCheckpointManager(
...
+    val systemConsumer = getSystemConsumer()
+    val (oldestOffset, newestOffset) = getEarliestOffset(checkpointTopic, new Partition(0))
+    systemConsumer.register(ssp, oldestOffset); // checkpoint stream should always be read from the beginning
+    systemConsumer.start();
+
+    var count = 0
+    try {
+      var done = false
+      var currentOffset = -1L
+      val newestOffsetLong = newestOffset.toLong
```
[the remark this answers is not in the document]
sborya: Now I am returning partitionMetadata, extract and convert it to Long.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: It was a reminder of the returned type. Moved it to the actual type to make code more readable.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done.

sborya: added attempts, so trying 1000x10 attempts at max. Don't feel very comfortable leaving it as an endless loop.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: wait. 'envelopes' can be empty, but if it is empty, then 'messages' will be null. If 'messages' is not null, it still theoretically could be empty (although that would be weird).

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: yeah, I've removed it already.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: ok.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: this is my Intelij's idea of formatting.. not sure why..

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: I was so surprised that handle() method takes ByteBuffer instead of byte[], that I even named the variable after it.
Done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: Done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
sborya: well, I put it in Utils because its Coordinator counterpart was there, but, I guess, it makes more sense for it to be in this class, so I will move and refactor it.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
sborya: Done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
sborya: ok.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
sborya: that's kind of bigger refactoring - if you pass the systemFactory, then you also need to pass config. If you pass the config, then you also can avoid passing multiple/most of the other arguments and derive them from the config inside the method.

samza-kafka/src/test/scala/org/apache/samza/checkpoint/kafka/TestKafkaCheckpointManager.scala
sborya: I am not even sure what this _root_ is needed for. This is some Intelij's "import optimize" magic. 
Removed.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -86,61 +82,34 @@ class KafkaCheckpointManager(
...
     val msgBytes = serde.toBytes(checkpoint)
+    val systemStream = new SystemStream(systemName, checkpointTopic)
+    val envelope = new OutgoingMessageEnvelope(systemStream, keyBytes, msgBytes)
+
     retryBackoff.run(
       loop => {
-        if (producer == null) {
-          producer = connectProducer()
+        if (systemProducer == null) {
+          systemProducer.synchronized {
```
vjagadish1989: Is n't this a bug? Would you get an NPE? You seem to be synchronizing on a null `systemProducer`?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -86,61 +82,34 @@ class KafkaCheckpointManager(
...
+            if (systemProducer == null) {
+              systemProducer = getSystemProducer()
```
vjagadish1989: Why do we need a double `null` check here again? All you need is a synchronized read-modify-write back to the variable right? Can we simplify this?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -183,96 +152,100 @@ class KafkaCheckpointManager(
...
-    readLog(CHECKPOINT_LOG4J_ENTRY, shouldHandleEntry, handleCheckpoint)
+    readLog(shouldHandleEntry, handleCheckpoint)
     checkpoints.toMap /* of the immutable kind */
   }
 
+  private def getEarliestOffset(topic: String, partition: Partition): SystemStreamPartitionMetadata = {
+    val systemAdmin = getSystemAdmin()
+    val metaDataMap: java.util.Map[String, SystemStreamMetadata] = systemAdmin.getSystemStreamMetadata(Collections.singleton(topic))
+    val checkpointMetadata: SystemStreamMetadata = metaDataMap.get(topic)
+    if(checkpointMetadata == null)
```
vjagadish1989: minor: spacing with `if(`.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -183,96 +152,100 @@ class KafkaCheckpointManager(
...
+      throw new SamzaException("Cannot get metadata for system=%s, topic=%s" format (systemName, topic))
```
vjagadish1989: Prefer `{}` even for one line statements to make code immune to accidental refactors later

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -183,96 +152,100 @@ class KafkaCheckpointManager(
       checkpoints.put(taskName, checkpoint) // replacing any existing, older checkpoints as we go
     }
 
-    readLog(CHECKPOINT_LOG4J_ENTRY, shouldHandleEntry, handleCheckpoint)
+    readLog(shouldHandleEntry, handleCheckpoint)
     checkpoints.toMap /* of the immutable kind */
   }
 
+  private def getEarliestOffset(topic: String, partition: Partition): SystemStreamPartitionMetadata = {
```
vjagadish1989: Name is misleading. `getSSPMetadata` ?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -183,96 +152,100 @@ class KafkaCheckpointManager(
...
+    val partitionMetaData = checkpointMetadata.getSystemStreamPartitionMetadata().get(partition)
+    if(partitionMetaData == null)
+      throw new SamzaException("Cannot get partitionMetaData for system=%s, topic=%s" format (systemName, topic))
+
+    return partitionMetaData
+  }
 
   /**
-   * Common code for reading both changelog partition mapping and change log
+   * Reads an entry from the checkpoint log and invokes the provided lambda on that entry.
```
vjagadish1989: Minor: s/on that entry/on it

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -183,96 +152,100 @@ class KafkaCheckpointManager(
...
-  private def readLog(entryType:String, shouldHandleEntry: (KafkaCheckpointLogKey) => Boolean,
+  private def readLog(shouldHandleEntry: (KafkaCheckpointLogKey) => Boolean,
                       handleEntry: (ByteBuffer, KafkaCheckpointLogKey) => Unit): Unit = {
-    retryBackoff.run[Unit](
-      loop => {
-        val consumer = getConsumer()
-
-        val topicAndPartition = new TopicAndPartition(checkpointTopic, 0)
 
+    val UNKNOWN_OFFSET: String = "-1"
```
vjagadish1989: Do we need the return type "String" here?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -183,96 +152,100 @@ class KafkaCheckpointManager(
...
-          info("Get latest offset %s for topic %s and partition 0." format(latestOffset, checkpointTopic))
-
-          if (offset < 0) {
-            info("Got offset 0 (no messages in %s) for topic %s and partition 0, so returning empty collection. If you expected the checkpoint topic to have messages, you're probably going to lose data." format (entryType, checkpointTopic))
-            return
+          systemConsumer.poll(sspToPoll, POLL_TIMEOUT)
+        } catch {
+          case e: Exception => {
+            // these exceptions are most likely intermediate
+            warn("Got %s exception while polling the consumer for checkpoints." format e.getMessage)
```
vjagadish1989: should you print the entire stack trace for the exception instead of just the message?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -183,96 +152,100 @@ class KafkaCheckpointManager(
...
+            if (attempts == 0) throw new SamzaException("Multiple attempts failed while reading the checkpoints. Giving up.")
```
vjagadish1989: prefer to wrap the underlying exception here.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -183,96 +152,100 @@ class KafkaCheckpointManager(
...
+            attempts -= 1
+            emptyEnvelops
```
vjagadish1989: s/emptyEnvelops/emptyEnvelopes

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -286,15 +259,11 @@ class KafkaCheckpointManager(
   }
 
   def stop = {
-    if (producer != null) {
-      producer.close
+    if (systemProducer != null) {
+      systemProducer.stop
```
vjagadish1989: Should synchronize all accesses to `systemProducer` for memory visibility.
Consider two threads: T1, T2. T1 recycles the producer reference, which is not visible to T2. T2 calls `stop` on the previous producer instance causing a leak?

samza-kafka/src/test/scala/org/apache/samza/checkpoint/kafka/TestKafkaCheckpointManager.scala
```diff
@@ -69,23 +68,39 @@ class TestKafkaCheckpointManager extends KafkaServerTestHarness {
...
     metadataStore = new ClientUtilTopicMetadataStore(brokers, "some-job-name")
+
+    config.put(SystemConfig.SYSTEM_FACTORY format systemName, "org.apache.samza.system.kafka.KafkaSystemFactory")
+    config.put(org.apache.samza.config.KafkaConfig.CHECKPOINT_SYSTEM, systemName);
+    config.put(JobConfig.JOB_NAME, "some-job-name");
+    config.put(JobConfig.JOB_ID, "i001");
+    config.put("systems.%s.producer.%s" format (systemName, ProducerConfig.BOOTSTRAP_SERVERS_CONFIG), brokers)
+    config.put("systems.%s.consumer.zookeeper.connect" format systemName, zkConnect)
+    val cfg: SystemConfig = new SystemConfig(new MapConfig(config))
+    val (sName: String, systemConsumerFactory : SystemFactory) =  KafkaCheckpointManagerFactory.getCheckpointSystemStreamAndFactory(cfg)
```
vjagadish1989: s/sName/systemName

samza-kafka/src/test/scala/org/apache/samza/checkpoint/kafka/TestKafkaCheckpointManager.scala
```diff
@@ -248,8 +305,10 @@ class TestKafkaCheckpointManager extends KafkaServerTestHarness {
     socketTimeout = 30000,
     bufferSize = 64 * 1024,
     fetchSize = 300 * 1024,
+    getSystemConsumer = systemConsumerF,
+    getSystemAdmin = systemAdminF,
     metadataStore = metadataStore,
-    connectProducer = () => new KafkaProducer(producerConfig.getProducerProperties),
+    getSystemProducer = systemProducerF,
```
vjagadish1989: s/systemProducerF/systemProducerFn

Like wise, replace for all occurences of consumer and admin

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -248,6 +240,25 @@ object Util extends Logging {
   }
 
   /**
+   * Get the Checkpoint System and system factory from the configuration
```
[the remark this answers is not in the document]
sborya: ok

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -19,29 +19,21 @@
...
 
-import java.net._
 import java.io._
 import java.lang.management.ManagementFactory
-import java.util.zip.CRC32
-import org.apache.samza.{SamzaException, Partition}
-import org.apache.samza.system.{SystemFactory, SystemStreamPartition, SystemStream}
+import java.net._
 import java.util.Random
+import java.util.zip.CRC32
```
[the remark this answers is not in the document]
sborya: there are some existing code which is using it.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: Of course. Fixed.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: this allows you to avoid getting lock in most cases.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: now it does. changed.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: you are better then checkStyle! Done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: k

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: k

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done. missed it.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: yep. Done.

samza-kafka/src/test/scala/org/apache/samza/checkpoint/kafka/TestKafkaCheckpointManager.scala
sborya: there is already  a systemName variable above. Changed to something else.

samza-kafka/src/test/scala/org/apache/samza/checkpoint/kafka/TestKafkaCheckpointManager.scala
sborya: done.

vjagadish1989 (changes requested):
looks great.

final round of review/feedback.

```
+        if (systemProducer == null) {
     synchronized {
 +            if (systemProducer == null) {
 +              systemProducer = getSystemProducer()
 +              systemProducer.register(taskName.getTaskName)
 +              systemProducer.start
 +            }
 +          }


```
This double checked locking idiom used here is broken when used with a compiler that can re-order instructions (or when run on a shared memory multi-processor). 

There is still no guarantee that the first check gets the most up-date value of `systemProducer`.

Can we make `systemProducer` volatile to fix this?

sborya:
oh, interesting. I will add volatile.

PR #300, opened 2017-09-21
SAMZA-1428: Fix scala 2.10 compilation issue with java 8 interface st…

jmakes:
…atic methods


jmakes:
@sborya take a look!

Also, because of the PR number, I gotta say: 
"THIS.... IS.... SPARTA!!!!"

sborya:
And because yesterday was a "Speak as a pirate day" you should say:
"This .. is .. SPARRRRRTA!!!

PR #303, opened 2017-09-26
SAMZA-1423: Implement time series storage for joins and windows

vjagadish1989:
Notable changes:
* New interface for storing and retrieving time series data.
* New store and serde implementation for use in windows and joins

Pending:
* Documentation, and minor clean-ups
* Wire-up of stores from ExecutionPlanner
* Usage of the store to implement various windows and joins


vjagadish1989:
@prateekm @nickpan47 @jmakes @xinyuiscool - Please review :-)

samza-api/src/main/java/org/apache/samza/storage/kv/ClosableIterator.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Iterator;
+
+/**
+ * An iterator that can be closed.
+ *
+ * <p> Implement close to free resources assigned to the iterator such as open file handles, persistent state etc.
+ *
+ * @param <V> the type of value returned by this iterator
+ */
+public interface ClosableIterator<V> extends Iterator<V> {
```
jmakes: Seems like it would be better to leverage the java Closeable interface so we can support AutoCloseable for try-with-resources
https://docs.oracle.com/javase/7/docs/api/java/lang/AutoCloseable.html

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesValue.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * An immutable pair of a value, and its corresponding timestamp.
+ *
+ * <p> Iterators on {@link TimeSeriesStore}s always return {@link TimeSeriesValue}s
+ *
+ * @param <V> the type of the value
+ */
+public class TimeSeriesValue<V> {
+  private final V value;
+  private final Long timestamp;
```
jmakes: Should this timestamp always match the timestamp in the key? If so, does it need to be in both the key and the value?

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
```diff
@@ -0,0 +1,185 @@
...
+    TimeSeriesKey<K> fromKey = new TimeSeriesKey(key, startTimestamp, 0);
+    TimeSeriesKey<K> toKey = new TimeSeriesKey(key, endTimeStamp, 0);
+
+    KeyValueIterator<TimeSeriesKey<K>, V> range = kvStore.range(fromKey, toKey);
+    return new TimeSeriesStoreIterator<>(range);
+  }
+
+  @Override
+  public void remove(K key, Long startTimestamp, Long endTimeStamp) {
+    validateRanges(startTimestamp, endTimeStamp);
```
prateekm: Minor: validateRange

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
```diff
@@ -0,0 +1,185 @@
...
+    KeyValueIterator<TimeSeriesKey<K>, V> range = kvStore.range(fromKey, toKey);
+    while (range.hasNext()) {
+      keysToDelete.add(range.next().getKey());
+    }
+
+    kvStore.deleteAll(keysToDelete);
+  }
+
+  @Override
+  public void close() {
```
prateekm: Do we need flush on this store too?

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
```diff
@@ -0,0 +1,185 @@
...
+    @Override
+    public boolean hasNext() {
+      return wrappedIterator.hasNext();
+    }
+
+    @Override
+    public TimeSeriesValue<V> next() {
+      Entry<TimeSeriesKey<K>, V> next = wrappedIterator.next();
+
+      TimeSeriesValue<V> kv = new TimeSeriesValue<>(next.getValue(), next.getKey().getTimestamp());
```
prateekm: Minor: 'return new TimeSeriesValue...'

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStore.java
```diff
@@ -0,0 +1,74 @@
...
+   *
+   * Values returned by the iterator are ordered by their timestamp. Values with the same timestamp are
+   * returned in their order of insertion.
+   *
+   * @param key the key to look up in the store
+   * @param startTimestamp the start timestamp of the range, inclusive
+   * @param endTimeStamp the end timestamp of the range, exclusive
+   * @throws IllegalArgumentException when startTimeStamp &gt; endTimeStamp, or when either of them is negative
+   */
+  ClosableIterator<TimeSeriesValue<V>> get(K key, Long startTimestamp, Long endTimeStamp);
```
prateekm: Do we need range(K k1, K k2)/all?

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
```diff
@@ -0,0 +1,185 @@
...
+      wrappedIterator.close();
+    }
+
+    @Override
+    public boolean hasNext() {
+      return wrappedIterator.hasNext();
+    }
+
+    @Override
+    public TimeSeriesValue<V> next() {
```
prateekm: Should this return the key as well? See question about `range`.

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesValue.java
```diff
@@ -0,0 +1,63 @@
...
+package org.apache.samza.operators.impl.store;
+
+/**
+ * An immutable pair of a value, and its corresponding timestamp.
+ *
+ * <p> Iterators on {@link TimeSeriesStore}s always return {@link TimeSeriesValue}s
+ *
+ * @param <V> the type of the value
+ */
+public class TimeSeriesValue<V> {
```
prateekm: Can we use the KV class from the serde PR when it's available?

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
```diff
@@ -0,0 +1,185 @@
...
+
+public class TimeSeriesStoreImpl<K, V> implements TimeSeriesStore<K, V> {
+
+  private final KeyValueStore<TimeSeriesKey<K>, V> kvStore;
+
+  /**
+   * Since timestamps are at the granularity of milliseconds, multiple entries added in the same
+   * millisecond are distinguished by a monotonically increasing sequence number.
+   */
+  private int seqNum = 0;
```
prateekm: I think it's reasonable to use an AtomicInteger here instead of an int + lock: 
The concern is ordering of values after the seq num wraps. We get 2G seq nums before we wrap. Given that we have 8 byte timestamp + 4 byte seqNum overhead per key, this is at the very minimum (12 bytes * 2 G) = 24 GB for a store with empty keys and values. Even an (int, int) key value store would have 8 * 24 = 192 GB storage at a minimum (there's additional rocksDB ttl timestamp and write amplification overhead), and this is per TaskInstance.

Just throwing it out there since we were also considering storing a version number with the key: we can use 7 LSB from a long seq num + a 1 byte version num.

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesKey.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || !getClass().equals(o.getClass())) return false;
+
+    TimeSeriesKey<?> that = (TimeSeriesKey<?>) o;
+
+    if (timestamp != that.timestamp) return false;
```
prateekm: Good call.

samza-api/src/main/java/org/apache/samza/storage/kv/ClosableIterator.java
prateekm: Can we use KeyValueIterator instead?

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
```diff
@@ -0,0 +1,185 @@
...
+    validateRanges(startTimestamp, endTimeStamp);
+    TimeSeriesKey<K> fromKey = new TimeSeriesKey(key, startTimestamp, 0);
+    TimeSeriesKey<K> toKey = new TimeSeriesKey(key, endTimeStamp, 0);
+
+    KeyValueIterator<TimeSeriesKey<K>, V> range = kvStore.range(fromKey, toKey);
+    return new TimeSeriesStoreIterator<>(range);
+  }
+
+  @Override
+  public void remove(K key, Long startTimestamp, Long endTimeStamp) {
```
prateekm: Maybe add a `remove(K key)` too?

samza-api/src/main/java/org/apache/samza/storage/kv/ClosableIterator.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.apache.samza.storage.kv;
+
+import java.util.Iterator;
+
+/**
+ * An iterator that can be closed.
+ *
+ * <p> Implement close to free resources assigned to the iterator such as open file handles, persistent state etc.
```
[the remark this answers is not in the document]
vjagadish1989: good find , I will fix it uniformly everywhere! :)

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesKey.java
```diff
@@ -0,0 +1,79 @@
...
+    if (seqNum != that.seqNum) return false;
+    return key != null ? key.equals(that.key) : that.key == null;
+  }
+
+  @Override
+  public String toString() {
+    return "TimeSeriesKey{" +
+            "key=" + key +
```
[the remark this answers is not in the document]
vjagadish1989: Wil fix, thanks!

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
vjagadish1989: Since all access to the store are based on a key `K` and a timestamp, it's not necessary for the key to be returned again as a part of the iterator.

samza-api/src/main/java/org/apache/samza/storage/kv/ClosableIterator.java
vjagadish1989: We agreed that since all accesses to the store are by a particular key anyway, there's no need to return a KVIterator. The user of the store knows the key when methods are invoked.

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesKeySerde.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  @Override
+  public byte[] toBytes(TimeSeriesKey<K> timeSeriesKey) {
+    K key = timeSeriesKey.getKey();
+    long timestamp = timeSeriesKey.getTimestamp();
+    int seqNum = timeSeriesKey.getSeqNum();
+
+    byte[] serializedKey = keySerde.toBytes(key);
+    int keySize = serializedKey == null ? 0 : serializedKey.length;
```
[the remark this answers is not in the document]
vjagadish1989: Null bytes for key are needed to implement non-keyed tumbling and session windows. Basically, in that case, the start range of the timestamp will uniquely identify a window. The start range of the timestamp in combination with the sequence number will uniquely identify an entry in the store

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStore.java
vjagadish1989: There's no usecase currently that requires a range scan across keys, or an iterator across all keys in the store. We agreed that we'll add it should a scenario arise

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
vjagadish1989: Thank you for the suggestion.

After discussion we agreed the following:
- The wrapping is a non-issue if we use an atomic long
- We'll use a version for the MSB, and use the next 7 bytes for storing the long value in the sequence number

samza-api/src/main/java/org/apache/samza/storage/kv/ClosableIterator.java
vjagadish1989: 1. While the `AutoClosable` idiom is useful for file handles, it expects the caller to catch the checked `Exception` each time they call `iterator.close()`. Almost always, In Samza we 'd want to shutdown when there's an exception during `iterator.close()`, So, a runtime exception makes more sense.

2. Furthermore,  `iterator.close()` on our existing `KVStore` API does not throw any checked `Exception`. I'd prefer to lean on the side of consistency, and refactor it later if there's enough consensus. Happy to discuss offline, and follow-up!

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesValue.java
vjagadish1989: There are no such requirements. 

- On the write side, you'd call `put(k, v, timestamp)`
- On the query side, you ask the store to give you all values for a key within a timestamp range. The return value is an iterator that returns the value along with the timestamp they were added to the store. 

There's no need to store the timestamp in both the key and the value.

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesValue.java
vjagadish1989: Well, I ended up renaming the `TimeSeriesValue<V>` to `TimestampedValue<V>`. 

In general, prefer new classes (with their own semantic meanings) over `KV<K, V>` or `Pair<K,V>`.

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
```diff
@@ -0,0 +1,200 @@
...
+ *    |  serialized-key bytes |  timestamp       | version    |       sequence number  | serialized value      |
+ *    |                       |                  |            |                        |                       |
+ *    +-----------------------+------------------+------------+------------------------+-----------------------+
+ *    +----------------------+--------8 bytes----+----1 bytes-+---------7 bytes--------+----value size----------
+ *    +----------------------------------STORE KEY-------------------------------------+---STORE VAL-----------+
+ *  </pre>
+ *  An 8 byte timestamp, a one byte version and a 7 byte sequence number are appended to the provided key and this
+ *  combination is used as the key in the k-v store. The provided value is stored as is.
+ *
+ *  <p> This class is thread-safe and concurrent reads/writes are expected.
```
prateekm: Does the KVStore API/impl provide read after write consistency across threads as well? I.e. do we have any guarantees  about when a value written by T1 will be visible to T2?

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStore.java
```diff
@@ -0,0 +1,86 @@
...
+   * @param val the value to insert
+   * @param timestamp the timestamp in milliseconds
+   */
+  void put(K key, V val, long timestamp);
+
+  /**
+   * Returns an iterator over values for the given key in the provided time-range - [{@code startTimestamp}, {@code endTimestamp})
+   *
+   * Values returned by the iterator are ordered by their timestamp. Values with the same timestamp are
+   * returned in their order of insertion. The iterator must be closed after use by calling {@link #close}.
```
prateekm: IMHO the "Iterator must be closed" warning is worth putting on a newline and putting a &lt;b&gt; around must. Forgetting to do so will lead to memory leaks.

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStore.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * A key-value store that allows entries to be queried and stored based on time ranges.
+ *
+ * Operations on the store can be invoked from multiple threads. Hence, implementations are expected to be thread-safe.
+ *
+ * @param <K> the type of key in the store
+ * @param <V> the type of value in the store
+ */
+public interface TimeSeriesStore<K, V> {
```
xinyuiscool: should V be something extends fromTimestampedValue<V>?

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
```diff
@@ -0,0 +1,200 @@
...
+    // number is always zero. This ensures that only the most recent value is retained.
+    if (appendMode) {
+      seqNum.getAndIncrement();
+    }
+    TimeSeriesKey<K> timeSeriesKey = new TimeSeriesKey<>(key, timestamp, seqNum.get());
+    kvStore.put(timeSeriesKey, val);
+  }
+
+  @Override
+  public ClosableIterator<TimestampedValue<V>> get(K key, long startTimestamp, long endTimestamp) {
```
xinyuiscool: Why not close the iterator automatically when it reaches to the end? Now we have to remind user to close it after using.

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
```diff
@@ -0,0 +1,200 @@
...
+    if (endTimestamp < 0) {
+      throw new IllegalArgumentException(String.format("End timestamp :%d is less than zero", endTimestamp));
+    }
+
+    if (endTimestamp < startTimestamp) {
+      throw new IllegalArgumentException(String.format("End timestamp :%d is less than start timestamp: %d", endTimestamp, startTimestamp));
+    }
+  }
+
+  private static class TimeSeriesStoreIterator<K, V> implements ClosableIterator<TimestampedValue<V>> {
```
xinyuiscool: as mentioned above, is it possible to automatically close it when it reaches the end?

build.gradle
```diff
@@ -167,6 +167,7 @@ project(":samza-core_$scalaVersion") {
     compile "org.scala-lang:scala-library:$scalaLibVersion"
     compile "org.slf4j:slf4j-api:$slf4jVersion"
     testCompile project(":samza-api").sourceSets.test.output
+    testCompile project(":samza-kv-rocksdb_$scalaVersion")
```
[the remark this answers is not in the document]
vjagadish1989: That would mean that all tests for operators (window/ join) that instantiate stores would now be a part of `samza-kv-rocksdb`. Let's sync up offline since the scope of that seems to be larger than this PR.

Also, I believe circular dependencies in `testCompile` tasks are fine in gradle.

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesKey.java
```diff
@@ -0,0 +1,80 @@
...
+
+  public K getKey() {
+    return key;
+  }
+
+  public long getTimestamp() {
+    return timestamp;
+  }
+
+  public byte getVersion() {
```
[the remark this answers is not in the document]
vjagadish1989: Thanks, will fix it!

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStore.java
vjagadish1989: Not really, `<V>` should be completely agnostic to types here for this to be a general purpose API. However, the return type from `get` should be an `Iterator<TimeStampedValue<V>>` (as it's being done)

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
vjagadish1989: - Closing iterators automatically means that `iterator.hasNext()` is stateful - The invocation of `hasNext` immediately after the iterator has been closed will return an `InvalidStateException` . 
- Alternately, If we choose to return `false` during `hasNext()`, we should have a way to distinguish these 2 cases:
* `false` returned due to no elements to iterate upon
* `false` returned due to iterator being closed.

For these reasons, and to maintain consistency with our already existing iterator APIs that are built on top of RocksDb, I'd prefer to avoid auto-closing iterators.

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
```diff
@@ -0,0 +1,200 @@
...
+ * Data is serialized and organized into K-V pairs as follows:
+ *  <pre>
+ *    +-----------------------+------------------+------------+------------------------+-----------------------+
+ *    |  serialized-key bytes |  timestamp       | version    |       sequence number  | serialized value      |
+ *    |                       |                  |            |                        |                       |
+ *    +-----------------------+------------------+------------+------------------------+-----------------------+
+ *    +----------------------+--------8 bytes----+----1 bytes-+---------7 bytes--------+----value size----------
+ *    +----------------------------------STORE KEY-------------------------------------+---STORE VAL-----------+
+ *  </pre>
+ *  An 8 byte timestamp, a one byte version and a 7 byte sequence number are appended to the provided key and this
```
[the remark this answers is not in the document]
vjagadish1989: Thanks for catching this!

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStoreImpl.java
vjagadish1989: The RocksDb docs state that the iterators are "thread-safe". We'll have to find out more on whether they mean: 
- Concurrent operations on the store will not corrupt state
- Operations on the store "happen-before" other operations? That directly translates to visibility across threads.

[2] is a much stronger guarantee on [1]

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesStore.java
vjagadish1989: +1 for the extra emphasis

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesKey.java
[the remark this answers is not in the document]
vjagadish1989: I accidentally committed this without addressing this feedback. However, I ended up doing a follow-up patch that addresses your comment - https://github.com/apache/samza/commit/92894e5e283d605e6243177cc5ed11c3139aa7bc

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimeSeriesKeySerde.java
```diff
@@ -0,0 +1,96 @@
...
+      buf.get(keyBytes);
+      key = keySerde.fromBytes(keyBytes);
+    }
+
+    long timeStamp = buf.getLong();
+    long seqNum = buf.getLong();
+    long version = seqNum & ~SEQUENCE_NUM_MASK;
+
+    if (version != 0) {
+      throw new SamzaException("Version is not zero. Sequence number: " + seqNum);
```
prateekm: This check should be `version != TimeSeriesKey.VERSION`.

PR #305, opened 2017-09-28
SAMZA-1431: Enable SonarCloud integration with Travis-CI.

dnishimura:
These are modifications needed to integrate Travis-CI with SonarCloud. Also, the JaCoCo plugin is enabled in gradle for code coverage reporting to SonarCloud.

For reference, here is the request to Apache Infra to enable Travis-CI: https://issues.apache.org/jira/browse/INFRA-15132

@nickpan47 @isolis

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,7 +105,6 @@ allprojects {
 subprojects {
   apply plugin: 'eclipse'
   apply plugin: 'project-report'
-  apply plugin: 'java'
   apply plugin: 'jacoco'
 
   tasks.withType(Test) {
--- a/build.gradle
+++ b/build.gradle
@@ -105,6 +105,8 @@ allprojects {
 subprojects {
   apply plugin: 'eclipse'
   apply plugin: 'project-report'
+  apply plugin: 'java'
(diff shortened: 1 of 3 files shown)


dnishimura:
Forgot license header and a couple other things. Adding them now.

dnishimura:
I think all the Travis-CI gotchas are fixed. Ready for review!

build.gradle
```diff
@@ -105,6 +105,8 @@ allprojects {
 subprojects {
   apply plugin: 'eclipse'
   apply plugin: 'project-report'
+  apply plugin: 'java'
```
jmakes: 2 issues here:
1. If this is needed, then I believe you can remove the `apply plugin: 'java'` line from each of the projects
2. I don't know much about this, but since the scala plugin extends the java plugin, could this affect the projects that depend on scala (particularly those in which we force joint compilation)?

dnishimura: Turns out that line is not needed since each subproject already has the java or scala plugin. Removed in the latest commit.

PR #306, opened 2017-09-28
SAMZA-1429: Add callback success/failure metrics to async tasks

dnishimura:
@jmakes & @nickpan47 please take a look when you get the chance.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
+++ b/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
@@ -39,7 +39,6 @@ import org.apache.samza.container.SamzaContainerMetrics;
 import org.apache.samza.container.TaskInstance;
 import org.apache.samza.container.TaskInstanceMetrics;
 import org.apache.samza.container.TaskName;
-import org.apache.samza.metrics.Counter;
 import org.apache.samza.util.HighResolutionClock;
 import org.apache.samza.system.IncomingMessageEnvelope;
 import org.apache.samza.system.SystemConsumers;
@@ -69,8 +68,6 @@ public class AsyncRunLoop implements Runnable, Throttleable {
   private final long commitMs;
   private final long callbackTimeoutMs;
   private final SamzaContainerMetrics containerMetrics;
-  private final Counter containerMetricsCompletedCounter;
-  private final Counter containerMetricsFailureCounter;
   private final ScheduledExecutorService workerTimer;
   private final ScheduledExecutorService callbackTimer;
   private final ThrottlingScheduler callbackExecutor;
@@ -94,8 +91,6 @@ public class AsyncRunLoop implements Runnable, Throttleable {
     this.threadPool = threadPool;
     this.consumerMultiplexer = consumerMultiplexer;
     this.containerMetrics = containerMetrics;
-    this.containerMetricsCompletedCounter = containerMetrics.newCounter("async-callback-complete-calls");
-    this.containerMetricsFailureCounter = containerMetrics.newCounter("async-callback-failure-calls");
     this.windowMs = windowMs;
     this.commitMs = commitMs;
     this.maxConcurrency = maxConcurrency;
@@ -532,8 +527,7 @@ public class AsyncRunLoop implements Runnable, Throttleable {
         public void run() {
           try {
             state.doneProcess();
-            state.taskMetricsCompletedCounter.inc();
-            containerMetricsCompletedCounter.inc();
+            state.taskMetrics.asyncCallbackCompleted().inc();
             TaskCallbackImpl callbackImpl = (TaskCallbackImpl) callback;
             containerMetrics.processNs().update(clock.nanoTime() - callbackImpl.timeCreatedNs);
             log.trace("Got callback complete for task {}, ssp {}",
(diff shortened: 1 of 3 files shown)


samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -91,6 +94,8 @@ public AsyncRunLoop(Map<TaskName, TaskInstance> taskInstances,
     this.threadPool = threadPool;
     this.consumerMultiplexer = consumerMultiplexer;
     this.containerMetrics = containerMetrics;
+    this.containerMetricsCompletedCounter = containerMetrics.newCounter("async-callback-complete-calls");
```
nickpan47: Please follow the pattern we had to declare/add metric in SamzaContainerMetrics (i.e. instantiate the metric inside SamzaContainerMetrics class and just refer to it from outside like containerMetrics.callbackCompletes.inc()

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -595,10 +604,14 @@ public void onFailure(TaskCallback callback, Throwable t) {
     private final Set<SystemStreamPartition> processingSspSet;
     private final TaskName taskName;
     private final TaskInstanceMetrics taskMetrics;
+    private final Counter taskMetricsCompletedCounter;
```
nickpan47: Same here. Please follow the current pattern of instantiating metric in TaskInstanceMetrics

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
```diff
@@ -200,6 +201,49 @@ public void setup() {
     when(consumerMultiplexer.pollIntervalMs()).thenReturn(10);
   }
 
+  @Test
```
nickpan47: Can we also add the callback failure test cases as well?

dnishimura:
@jmakes I'll defer to @nickpan47, but it looks like you're correct that `messages-in-flight` will keep increasing if the user never calls the callback methods. Also good point on 2.

nickpan47:
@jmakes from the incidence that I saw, messages-in-flight stays as a constant positive number due to the fact that the process stalls when this number equals task.max.concurrency, which does not help much since in a full-throttle process, messages-in-flight may be a constant as well.

Good point on 2! I missed that. But I think that callbacks-completed metrics may still be helpful in the following situation (i.e. the one I caught previously during debugging some user code):
# task.callback.timeout.ms is not set
# user missed the call TaskCallback.complete() on completion
Mainly, if the container stuck w/ process-calls = 0 and callback-completed = 0 and messages-in-flight != 0, we can conclude that the container got stuck due to callback.complete() is never triggered. Make sense?

jmakes:
@nickpan47 

> user missed the call TaskCallback.complete() on completion
Mainly, if the container stuck w/ process-calls = 0 and callback-completed = 0 and messages-in-flight != 0, we can conclude that the container got stuck due to callback.complete() is never triggered. Make sense?

Well, process-calls is incremented before invoking the user process(), so I don't think that scenario can occur, exactly. Instead process-calls = 1, and messages-in-flight = 1. 

To me, if process-calls is not increasing and messages-in-flight stays non-zero for the same duration, we can conclude that it's stuck, right?

nickpan47:
Synced up with @jmakes offline. We agreed that this task level callback-completed would be useful to debug that a certain task stuck w/ failing. Hence, we are only adding this one metric in TaskInstanceMetrics and removing the failure counts.

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
dnishimura: Done

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
dnishimura: Done!

dnishimura:
@nickpan47 I removed the async container level metrics and left only the async completed metric at the task level per your discussion with @jmakes. Ready for another review!

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
```diff
@@ -200,6 +201,45 @@ public void setup() {
...
+
+    TestTask task0 = new TestTask(true, true, false, task0ProcessedMessages);
+    TestTask task1 = new TestTask(true, true, false, task1ProcessedMessages);
+    TaskInstance t0 = createTaskInstance(task0, taskName0, ssp0);
+    TaskInstance t1 = createTaskInstance(task1, taskName1, ssp1);
+
+    Map<TaskName, TaskInstance> tasks = new HashMap<>();
+    tasks.put(taskName0, t0);
+    tasks.put(taskName1, t1);
+    //task0.callbackHandler = buildOutofOrderCallback(task0);
```
nickpan47: Remove this line if not used.

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
```diff
@@ -200,6 +201,45 @@ public void setup() {
     when(consumerMultiplexer.pollIntervalMs()).thenReturn(10);
   }
 
+  @Test
```
nickpan47: Would be nice if we also add a test case to illustrate that callback-completed was not incremented by mal-implemented task code.

PR #307, opened 2017-09-29
SAMZA-1434: Fix issues found in Hadoop

xinyuiscool:
Fix the following bugs found when running Samza on hadoop:

1. Hdfs allows output partitions to be 0 (empty folder)
2. Add null check for the changelog topic generation
3. Call getStreamSpec() instead of using streamSpec member in StreamEdge. This is due to getStreamSpec will do more transformation.
4. Bound the auto-generated intermediate topic partition by a certain count (256).

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
+++ b/samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
@@ -158,8 +158,9 @@ public class StreamSpec {
     validateLogicalIdentifier("streamId", id);
     validateLogicalIdentifier("systemName", systemName);
 
-    if (partitionCount < 1) {
-      throw new IllegalArgumentException("Parameter 'partitionCount' must be greater than 0");
+    // partition count being 0 is a valid use case in Hadoop when the output stream is an empty folder
+    if (partitionCount < 0) {
+      throw new IllegalArgumentException("Parameter 'partitionCount' must be >= 0");
     }
 
     this.id = id;
--- a/samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
+++ b/samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
@@ -50,6 +50,8 @@ import org.slf4j.LoggerFactory;
 public class ExecutionPlanner {
   private static final Logger log = LoggerFactory.getLogger(ExecutionPlanner.class);
 
+  private static final int MAX_INFERRED_PARTITIONS = 256;
+
   private final Config config;
   private final StreamManager streamManager;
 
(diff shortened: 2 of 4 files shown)


xinyuiscool:
@sborya : please help take a look!

vjagadish1989 (approved):
looks great.

approved.

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -158,8 +158,9 @@ public StreamSpec(String id, String physicalName, String systemName, int partiti
     validateLogicalIdentifier("streamId", id);
     validateLogicalIdentifier("systemName", systemName);
 
-    if (partitionCount < 1) {
```
jmakes: Please add this original validation to KafkaStreamSpec.java. It is still the proper validation for kafka.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -253,9 +255,10 @@ private static void calculateIntStreamPartitions(JobGraph jobGraph, Config confi
     if (partitions < 0) {
       // use the following simple algo to figure out the partitions
       // partition = MAX(MAX(Input topic partitions), MAX(Output topic partitions))
+      // partition will be further bounded by MAX_INFERRED_PARTITIONS. This is important when running in hadoop.
       int maxInPartitions = maxPartition(jobGraph.getSources());
       int maxOutPartitions = maxPartition(jobGraph.getSinks());
-      partitions = Math.max(maxInPartitions, maxOutPartitions);
+      partitions = Math.min(Math.max(maxInPartitions, maxOutPartitions), MAX_INFERRED_PARTITIONS);
```
jmakes: There should at least be a WARN message if MAX_INFERRED_PARTITIONS, because artificially choosing a max can affect correctness(joins) and stability (too much throughput per partition).

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -253,9 +255,10 @@ private static void calculateIntStreamPartitions(JobGraph jobGraph, Config confi
     if (partitions < 0) {
       // use the following simple algo to figure out the partitions
       // partition = MAX(MAX(Input topic partitions), MAX(Output topic partitions))
+      // partition will be further bounded by MAX_INFERRED_PARTITIONS. This is important when running in hadoop.
```
jmakes: Please elaborate why it's important for hadoop. The current statement is to vague to be useful for code maintainers

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
sborya: +1

PR #308, opened 2017-09-29
SAMZA-1438: SystemProducer, Consumer and Admin interfaces for EventHubs

dxichen:


samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubEntityConnection.java
```diff
@@ -0,0 +1,132 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(EventHubEntityConnection.class);
+  final Map<Integer, String> offsets = new TreeMap<>();
+  final Map<Integer, PartitionReceiveHandler> handlers = new TreeMap<>();
+  private final String namespace;
+  private final String entityPath;
+  private final String sasKeyName;
+  private final String sasKey;
+  private final String consumerName;
+  private final EventHubConfig config;
+  private final Map<Integer, PartitionReceiver> receivers = new TreeMap<>();
```
vjagadish1989: Currently, the `EventHubEntityConnection` class has some stateful/mutating operations -
including, some synchronized methods:
```
+  synchronized void addPartition(int partitionId, String offset, PartitionReceiveHandler handler) {}
+  synchronized void connectAndStart() {}
+  synchronized void stop() {}
```
Can we consider separating these and making the class immutable:
```
 +  private final String namespace;
 +  private final String entityPath;
 +  private final String sasKeyName;
 +  private final String sasKey;
 +  private final String consumerName;
 +  private final EventHubConfig config;
```
Also, I'd remove these state-ful fields from this class:
```
 +  private final Map<Integer, PartitionReceiver> receivers = new TreeMap<>();
 +  private EventHubClientWrapper ehClientWrapper;
 +  private final EventHubClientFactory eventHubClientFactory = new EventHubClientFactory();
```

Do you think we can move this map of partition -> PartitionReceiver instances into the Consumer itself since all of these are related to receiving events? Another nice thing is that you can get rid off the `EventHubEntitityConnectionFactory` completely for wiring these stateful components

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubSystemFactory.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @Override
+  public SystemConsumer getConsumer(String systemName, Config config, MetricsRegistry registry) {
+    return new EventHubSystemConsumer(new EventHubConfig(config, systemName),
+            new EventHubEntityConnectionFactory(), registry);
+  }
+
+  @Override
+  public SystemProducer getProducer(String systemName, Config config, MetricsRegistry registry) {
+    return new EventHubSystemProducer(systemName, new EventHubConfig(config, systemName), registry);
```
vjagadish1989: - Can we inject the `EventHubWrapperFactory` into the `EventHubSystemProducer`, `EventHubSystemConsumer` and `EventHubSystemAdmin`
instead of class-loading them via config? It'd be much cleaner with types that way. Use-cases at LinkedIn that require  tunneling can inject their tunnel-aware versions of the EventHubClientWrapper.

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -0,0 +1,242 @@
...
+      errors = readErrors;
+      this.serde = serde;
+    }
+
+    @Override
+    public void onReceive(Iterable<EventData> events) {
+      if (events != null) {
+
+        events.forEach(event -> {
+            byte[] decryptedBody = event.getBytes();
```
vjagadish1989: Remove references to encryption and de-cryption everywhere? since they seem orthogonal to the record payloadss

dxichen: Removed the Serde configs and added the Stream to Serde mapping to `EventHubSystemFactory`. (Passthrough Serde by default)

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubEntityConnection.java
dxichen: Refactored in order to remove `EventHubConnectionEnitity` https://github.com/apache/samza/pull/308/commits/1c1677faf360d1287717abd4c3ceba8776d365eb

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventDataWrapper.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public EventData getEventData() {
+    return eventData;
+  }
+
+  /**
+   * @return the body of decrypted body of the message. In case no encryption is setup for this topic,
+   * just returns the body of the message.
+   */
+  public byte[] getDecryptedBody() {
```
vjagadish1989: `s/getDecryptedBody/getBody`

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubClientWrapper.java
```diff
@@ -0,0 +1,45 @@
...
+* KIND, either express or implied.  See the License for the
+* specific language governing permissions and limitations
+* under the License.
+*/
+
+package org.apache.samza.system.eventhub;
+
+import com.microsoft.azure.eventhubs.EventHubClient;
+
+public interface EventHubClientWrapper {
```
vjagadish1989: Add java docs:
`Wraps the {@link EventHubClient} with lifecycle hooks for initialization and close`

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubConfig.java
```diff
@@ -0,0 +1,232 @@
...
+  public static final String CONFIG_STREAM_SAS_TOKEN = "systems.%s.streams.%s.eventhubs.sas.token";
+
+  public static final String CONFIG_STREAM_CONSUMER_GROUP = "systems.%s.streams.%s.eventhubs.consumer.group";
+  public static final String DEFAULT_CONFIG_STREAM_CONSUMER_GROUP = EventHubClient.DEFAULT_CONSUMER_GROUP_NAME;
+
+  public static final String CONFIG_STREAM_CONSUMER_START_POSITION = "systems.%s.streams.%s.eventhubs.start.position";
+  public static final String DEFAULT_CONFIG_STREAM_CONSUMER_START_POSITION = StartPosition.LATEST.name();
+
+  public static final String CONFIG_PRODUCER_PARTITION_METHOD = "systems.%s.eventhubs.partition.method";
+  public static final String DEFAULT_CONFIG_PRODUCER_PARTITION_METHOD = EventHubSystemProducer
```
vjagadish1989: Can't we internally translate calls to `send()` into one of these partition schemes? For instance, if a specific partitionId is provided, we write it to the partition. If not, we use the eventhub's hash partitioner?

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubConfig.java
```diff
@@ -0,0 +1,232 @@
...
+          .PartitioningMethod.EVENT_HUB_HASHING.name();
+
+  public static final String CONFIG_SEND_KEY_IN_EVENT_PROPERTIES = "systems.%s.eventhubs.send.key";
+  public static final Boolean DEFAULT_CONFIG_SEND_KEY_IN_EVENT_PROPERTIES = false;
+
+  public static final String CONFIG_CONNECTION_SHUTDOWN_TIMEOUT_MILLIS = "systems.%s.eventhubs.shutdown.timeout";
```
vjagadish1989: Would be good to confirm if all these configs need to be exposed as top-level configs? If we can simply propagate eventhub configs down to the client, it'll save us from maintaining these configs

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubConfig.java
```diff
@@ -0,0 +1,232 @@
...
+  public static final String CONFIG_FETCH_RUNTIME_INFO_TIMEOUT_MILLIS = "systems.%s.eventhubs.runtime.info.timeout";
+  public static final long DEFAULT_CONFIG_FETCH_RUNTIME_INFO_TIMEOUT_MILLIS = Duration.ofMinutes(1L).toMillis();
+
+  public static final String CONFIG_CONSUMER_BUFFER_CAPACITY = "systems.%s.eventhubs.receive.queue.size";
+  public static final int DEFAULT_CONFIG_CONSUMER_BUFFER_CAPACITY = 100;
+
+
+  private final String systemName;
+
+  public EventHubConfig(Map<String, String> config, String systemName) {
```
vjagadish1989: Is there a reason for `EventHubConfig` to take in a systemName? Seems fairly system agnostic.

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubSystemFactory.java
```diff
@@ -0,0 +1,65 @@
...
+    Map<String, Serde<byte[]>> serdes = new HashMap<>();
+    List<String> streamList = config.getStreamList();
+    streamList.forEach((streamName) -> serdes.put(streamName, new ByteSerde()));
+    return serdes;
+  }
+
+  @Override
+  public SystemConsumer getConsumer(String systemName, Config config, MetricsRegistry registry) {
+    EventHubConfig eventHubConfig = new EventHubConfig(config, systemName);
+    return new EventHubSystemConsumer(eventHubConfig, new EventHubClientWrapperFactory(), getSerdesMap(eventHubConfig),
```
vjagadish1989: Can we call `EventHubClientWrapper` as `SamzaEventHubClient`?  the name "wrapper" might be slightly confusing?

samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+      });
+    ehRuntimeInfos.forEach((streamName, ehRuntimeInfo) -> {
+        if (!streamPartitions.containsKey(streamName)) {
+          try {
+            EventHubRuntimeInformation ehInfo = ehRuntimeInfo.get(eventHubConfig.getRuntimeInfoWaitTimeMS(),
+                    TimeUnit.MILLISECONDS);
+            streamPartitions.put(streamName, ehInfo.getPartitionIds());
+          } catch (InterruptedException | ExecutionException e) {
+            String msg = String.format("Error while fetching EventHubRuntimeInfo for System:%s, Stream:%s",
```
vjagadish1989: Cant we use the same path for all exceptions?  We appear to be handling exceptions the same way in either of these `catch` code paths.

samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
```diff
@@ -0,0 +1,164 @@
...
+            LOG.error(msg, e);
+            throw new SamzaException(msg);
+          } catch (TimeoutException e) {
+            String msg = String.format("Timed out while fetching EventHubRuntimeInfo for System:%s, Stream:%s",
+                    systemName, streamName);
+            LOG.error(msg, e);
+            throw new SamzaException(msg);
+          }
+        }
+        requestedMetadata.put(streamName, new SystemStreamMetadata(streamName,
```
vjagadish1989: Where is `close` called? If not, we should close the client in `finally` block to avoid memory leaks?

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -0,0 +1,321 @@
...
+
+  public static final String START_OF_STREAM = PartitionReceiver.START_OF_STREAM; // -1
+  public static final String END_OF_STREAM = "-2";
+  public static final String AGGREGATE = "aggregate";
+  public static final String EVENT_READ_RATE = "eventReadRate";
+  public static final String EVENT_BYTE_READ_RATE = "eventByteReadRate";
+  public static final String READ_LATENCY = "readLatency";
+  public static final String READ_ERRORS = "readErrors";
+
+  private static Counter aggEventReadRate = null;
```
vjagadish1989: Can these counters be per-instance variables? or do they have to be shared? another nice thing is that you can get rid off this `aggregate` lock.

cc: @lhaiesp to confirm

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -0,0 +1,321 @@
...
+  private static Counter aggEventByteReadRate = null;
+  private static SamzaHistogram aggReadLatency = null;
+  private static Counter aggReadErrors = null;
+
+  private Map<String, Counter> eventReadRates;
```
vjagadish1989: consider making these maps final

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -0,0 +1,321 @@
...
+    eventReadRates = streamList.stream()
+            .collect(Collectors.toMap(Function.identity(), x -> registry.newCounter(x, EVENT_READ_RATE)));
+    eventByteReadRates = streamList.stream()
+            .collect(Collectors.toMap(Function.identity(), x -> registry.newCounter(x, EVENT_BYTE_READ_RATE)));
+    readLatencies = streamList.stream()
+            .collect(Collectors.toMap(Function.identity(), x -> new SamzaHistogram(registry, x, READ_LATENCY)));
+    readErrors = streamList.stream()
+            .collect(Collectors.toMap(Function.identity(), x -> registry.newCounter(x, READ_ERRORS)));
+
+    // Locking to ensure that these aggregated metrics will be created only once across multiple system consumers.
```
vjagadish1989: Do we really need an aggregate view of the metric? 

@lhaiesp

vjagadish1989:
@prateekm for review.

cc:
@lhaiesp @srinipunuru as well since they are knowledgeable about Eventhubs.

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubConfig.java
dxichen: the `partitionId` is obtained by the partitionKey on the outgoingEnvelope. This id could be used as the partition on eventhub or the seed for the consistent hashing on eventhub. 

It could be ambiguous if the user wants to use event hub hash by providing the hash via the `partitionId` if the partition method is not specified.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -0,0 +1,337 @@
...
+    eventHubClients.values().forEach(ehClient -> ehClient.close(config.getShutdownWaitTimeMS()));
+    eventHubClients.clear();
+  }
+
+  @Override
+  public synchronized void register(String streamName) {
+    LOG.info("Trying to register {}.", streamName);
+    if (isStarted) {
+      String msg = "Cannot register once the producer is started.";
+      LOG.error(msg);
```
prateekm: General comment: Usually these exceptions get logged by the caller/uncaught exception handler anyway. Would strongly prefer not over-logging (here and everywhere else).

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -0,0 +1,337 @@
...
+      throw new SamzaException(msg);
+    }
+
+    EventHubClientWrapper ehClient = eventHubClientWrapperFactory
+            .getEventHubClientWrapper(config.getStreamNamespace(streamName), config.getStreamEntityPath(streamName),
```
prateekm: Extract variables.

Formatting: There are several instances where we're inlining verbose method calls and line wrapping half-way through the statement. IMHO it's much more readable to extract variables instead.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -0,0 +1,337 @@
...
+      SamzaException e = new SamzaException(sendExceptionOnCallback);
+      sendExceptionOnCallback = null;
+      pendingFutures.clear();
+      LOG.error("One of the previous sends failed.");
+      throw e;
+    }
+
+    EventData eventData = createEventData(destination, envelope);
+    int eventDataLength =  eventData.getBytes() == null ? 0 : eventData.getBytes().length;
+    eventWriteRate.get(destination).inc();
```
prateekm: @vjagadish1989 would be good to check that these metrics aren't duplicating those that samza emits anyway.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -0,0 +1,337 @@
...
+    aggEventByteWriteRate.inc(eventDataLength);
+    EventHubClientWrapper ehClient = eventHubClients.get(destination);
+
+    Instant startTime = Instant.now();
+
+    CompletableFuture<Void> sendResult = sendToEventHub(destination, eventData, getEnvelopePartitionId(envelope),
+            ehClient.getEventHubClient());
+
+    Instant endTime = Instant.now();
+    long latencyMs = Duration.between(startTime, endTime).toMillis();
```
prateekm: Maybe just use currentTimeMillis/nanoTime and subtraction for simplicity?

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -0,0 +1,337 @@
...
+        return aVoid;
+      });
+  }
+
+  private CompletableFuture<Void> sendToEventHub(String streamName, EventData eventData, Object partitionKey,
+                                                 EventHubClient eventHubClient) {
+    if (partitioningMethod == PartitioningMethod.EVENT_HUB_HASHING) {
+      return eventHubClient.send(eventData, convertPartitionKeyToString(partitionKey));
+    } else if (partitioningMethod == PartitioningMethod.PARTITION_KEY_AS_PARTITION) {
+      if (!(partitionKey instanceof Integer)) {
```
prateekm: Null check too (or is it guaranteed to be not-null?).

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -0,0 +1,337 @@
...
+        String msg = "Partition key should be of type Integer";
+        LOG.error(msg);
+        throw new SamzaException(msg);
+      }
+
+      PartitionSender sender = getPartitionSender(streamName, (int) partitionKey, eventHubClient);
+      return sender.send(eventData);
```
prateekm: Is there a built in back-pressure mechanism (producer buffer size configuration etc.) somewhere? Would we run OOM if the producer can't keep up?

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -0,0 +1,337 @@
...
+    } else {
+      throw new SamzaException("Unknown partitioning method " + partitioningMethod);
+    }
+  }
+
+  private Object getEnvelopePartitionId(OutgoingMessageEnvelope envelope) {
```
prateekm: Unnecessary, can inline.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -0,0 +1,337 @@
...
+    } else if (partitionKey instanceof Integer) {
+      return String.valueOf(partitionKey);
+    } else if (partitionKey instanceof byte[]) {
+      return new String((byte[]) partitionKey, Charset.defaultCharset());
+    } else {
+      throw new SamzaException("Unsupported key type: " + partitionKey.getClass().toString());
+    }
+  }
+
+  private PartitionSender getPartitionSender(String streamName, int partition, EventHubClient eventHubClient) {
```
prateekm: getOrCreate

prateekm: Why is the creation happening during send? Should it happen once during start instead?

build.gradle
```diff
@@ -191,16 +191,25 @@ project(':samza-azure') {
...
     compile "com.microsoft.azure:azure-storage:5.3.1"
+    compile "com.microsoft.azure:azure-eventhubs:0.14.5"
     compile "com.fasterxml.jackson.core:jackson-core:2.8.8"
+    compile "io.dropwizard.metrics:metrics-core:3.1.2"
     compile project(':samza-api')
     compile project(":samza-core_$scalaVersion")
     compile "org.slf4j:slf4j-api:$slf4jVersion"
     testCompile "junit:junit:$junitVersion"
+    testCompile "org.mockito:mockito-all:$mockitoVersion"
+    testCompile "org.powermock:powermock-api-mockito:$powerMockVersion"
```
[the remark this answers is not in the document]
dxichen: Unfortunately, I need powermock in order to mock the EventHubClient (private ctor) and PartitionReceiver(final class) for unit tests.

build.gradle
```diff
@@ -191,16 +191,25 @@ project(':samza-azure') {
...
+    testCompile "org.powermock:powermock-core:$powerMockVersion"
+    testCompile "org.powermock:powermock-module-junit4:$powerMockVersion"
   }
   checkstyle {
     configFile = new File(rootDir, "checkstyle/checkstyle.xml")
     toolVersion = "$checkstyleVersion"
   }
+  test {
+    exclude 'org/apache/samza/system/eventhub/producer/**'
```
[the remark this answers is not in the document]
dxichen: Excluding it since it is an integration test that requires access to an external EventHub instance

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventDataWrapper.java
```diff
@@ -0,0 +1,53 @@
...
+*/
+
+package org.apache.samza.system.eventhub;
+
+import com.microsoft.azure.eventhubs.EventData;
+
+/**
+ * Simpler wrapper of {@link EventData} events with the decrypted payload
+ */
+public class EventDataWrapper {
```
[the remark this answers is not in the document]
dxichen: `EventData` cannot be extended due to private ctor. This is used as a workaround to pass to both the metadata from `EventData` and the decrypted body as a message to IME.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Mon Oct 31 23:13:44 PDT 2016
+#Thu Sep 07 09:55:54 PDT 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.8-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.8-all.zip
```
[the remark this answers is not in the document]
dxichen: Nope, not sure why it's there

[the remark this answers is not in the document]
lhaiesp: This actually helps you get all the gradle dependencies. So you that you can browse through the gradle plugins from your IDE.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -0,0 +1,337 @@
...
+    // Auto remove the future from the list when they are complete.
+    sendResult.handle((aVoid, throwable) -> {
+        long callbackLatencyMs = Duration.between(endTime, Instant.now()).toMillis();
+        sendCallbackLatency.get(destination).update(callbackLatencyMs);
+        aggSendCallbackLatency.update(callbackLatencyMs);
+        if (throwable != null) {
+          sendErrors.get(destination).inc();
+          aggSendErrors.inc();
+          LOG.error("Send message to event hub: {} failed with exception: ", destination, throwable);
+          sendExceptionOnCallback = throwable;
```
prateekm: If this handler isn't called on the send thread (I assume it's called on the thread that completed the future?), this exception may still not be visible to the sender threads even if the send method is synchronized:

https://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html#synchronization
"Synchronization ensures that memory writes by a thread before or during a synchronized block are made visible in a predictable manner to other threads which synchronize <b>on the same monitor</b>."

dxichen: Good catch, I will to sync on the `sendExceptionOnCallback` throwable instead.

build.gradle
```diff
@@ -191,16 +191,26 @@ project(':samza-azure') {
...
+    testCompile "org.mockito:mockito-all:$mockitoVersion"
+    testCompile "org.powermock:powermock-api-mockito:$powerMockVersion"
+    testCompile "org.powermock:powermock-core:$powerMockVersion"
+    testCompile "org.powermock:powermock-module-junit4:$powerMockVersion"
   }
   checkstyle {
     configFile = new File(rootDir, "checkstyle/checkstyle.xml")
     toolVersion = "$checkstyleVersion"
   }
+  test {
```
shanthoosh: Just curious. Any reason to exclude these tests from the build?

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -0,0 +1,342 @@
...
+          break;
+        case LATEST:
+          offset = END_OF_STREAM;
+          break;
+        default:
+          throw new SamzaException("Unknown starting position config " +
+                  config.getStartPosition(systemName, systemStreamPartition.getStream()));
+      }
+    }
+    streamPartitionOffsets.put(systemStreamPartition, offset);
```
shanthoosh: Please correct me if i'm wrong here.

I think this has to use offsetComparator and pick the lowest offset. 

If we've two tasks with checkpointed offsets T1(SP1: START_OF_STREAM), T2(SP1: END_OF_STREAM), register should choose the lowest one and store in streamPartitionOffsets for a partition(This register call will be called once for each TaskInstance). If we just overwrite, we could pick End (depending upon register call order).

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -0,0 +1,342 @@
...
+            .collect(Collectors.toMap(Function.identity(), x -> registry.newCounter(x, EVENT_READ_RATE)));
+    eventByteReadRates = streamNames.stream()
+            .collect(Collectors.toMap(Function.identity(), x -> registry.newCounter(x, EVENT_BYTE_READ_RATE)));
+    readLatencies = streamNames.stream()
+            .collect(Collectors.toMap(Function.identity(), x -> new SamzaHistogram(registry, x, READ_LATENCY)));
+    readErrors = streamNames.stream()
+            .collect(Collectors.toMap(Function.identity(), x -> registry.newCounter(x, READ_ERRORS)));
+
+    // Locking to ensure that these aggregated metrics will be created only once across multiple system consumers.
+    synchronized (AGGREGATE) {
```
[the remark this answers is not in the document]
dxichen: Nice catch, will use `Object` instead

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -0,0 +1,342 @@
...
+      if (aggEventReadRate == null) {
+        aggEventReadRate = registry.newCounter(AGGREGATE, EVENT_READ_RATE);
+        aggEventByteReadRate = registry.newCounter(AGGREGATE, EVENT_BYTE_READ_RATE);
+        aggReadLatency = new SamzaHistogram(registry, AGGREGATE, READ_LATENCY);
```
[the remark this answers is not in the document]
dxichen: I was planning to add the histogram as is for now and have a new PR later on to properly add the histogram metric.

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
dxichen: Good catch, I was relying on the `register` being only called once per SSP

samza-azure/src/main/java/org/apache/samza/system/eventhub/metrics/SamzaHistogram.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Creates a {@link Histogram} metric using {@link ExponentiallyDecayingReservoir}
+ * Keeps a {@link Gauge} for each percentile
+ */
+public class SamzaHistogram {
```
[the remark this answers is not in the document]
dxichen: As mentioned above, I was planning to have another PR for the metrics histogram update, thanks for the comment!

samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
```diff
@@ -0,0 +1,191 @@
...
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+public class EventHubSystemAdmin implements SystemAdmin {
```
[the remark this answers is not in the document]
dxichen: Seems like there will be no problem in implementing the extended version, but there might be something I'm missing. Maybe we can discuss this offline.

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubClientManager.java
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * Returns the EventHubClient instance of the wrapper so its methods can be invoked directly.
+   *
+   * @return EventHub client instance of the wrapper
+   */
+  EventHubClient getEventHubClient();
+
+  /**
+   * Timed synchronous connection close to the EventHub.
```
vjagadish1989: Reword docs?

Tries to close the {@link EventHubClient} waiting for the provided timeout. Set timeout to {@link EventHubClient.BLOCK_UNTIL_CLOSE} to block indefinitely till the client is closed.

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubClientManager.java
```diff
@@ -0,0 +1,48 @@
...
+   *
+   * @param timeoutMs
+   *          Time in Milliseconds to wait for individual components to
```
vjagadish1989: reword docs:
@timeoutMs the close timeout in milliseconds

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubSystemFactory.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.samza.system.eventhub.consumer.EventHubSystemConsumer;
+import org.apache.samza.system.eventhub.producer.EventHubSystemProducer;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class EventHubSystemFactory implements SystemFactory {
+
+  private Map<String, Serde<byte[]>> getSerdesMap(EventHubConfig config, String systemName) {
```
vjagadish1989: This need not be a `Serde` right? Since, this is a byte in byte out purely. Would be simpler to define an `Interceptor` interface that takes in a `byte[]` and returns a `byte[]`. We'll have a default pass-through interceptor that we'll use.

samza-azure/src/main/java/org/apache/samza/system/eventhub/SamzaEventHubClientManager.java
```diff
@@ -0,0 +1,101 @@
...
+      LOG.error(msg, e);
+      throw new SamzaException(msg, e);
+    }
+  }
+
+  public EventHubClient getEventHubClient() {
+    return eventHubClient;
+  }
+
+
```
vjagadish1989: Minor: Use override to explicitly indicate that it's overriding a super-type's method. Likewise for other overridden methods

samza-azure/src/main/java/org/apache/samza/system/eventhub/SamzaEventHubClientManager.java
```diff
@@ -0,0 +1,101 @@
...
+  public void close(long timeoutMS) {
+    try {
+      if (timeoutMS <= 0) {
```
vjagadish1989: Define a constant here, and use that instead.

samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
```diff
@@ -0,0 +1,199 @@
...
+  private final Map<String, String[]> streamPartitions = new HashMap<>();
+
+  public EventHubSystemAdmin(String systemName, EventHubConfig eventHubConfig,
+                             EventHubClientManagerFactory eventHubClientManagerFactory) {
+    this.systemName = systemName;
+    this.eventHubConfig = eventHubConfig;
+    this.eventHubClientManagerFactory = eventHubClientManagerFactory;
+  }
+
+  private static String getNextOffset(String currentOffset) {
```
vjagadish1989: any reason this should be static?

samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
```diff
@@ -0,0 +1,199 @@
...
+      });
+    return sspMetadataMap;
+  }
+
+  public static Integer compareOffsets(String offset1, String offset2) {
+    if (offset1 == null || offset2 == null) {
+      return null;
+    }
+    try {
+      if (offset1.equals(EventHubSystemConsumer.END_OF_STREAM)) {
```
vjagadish1989: we appear to be not comparing the actual offsets (we appear to be handling cases only where offset1 and offset2 are either null or EOS. Is that intentional? Should we convert them to Long and compare?

samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
```diff
@@ -0,0 +1,199 @@
...
+      return offset2.equals(EventHubSystemConsumer.END_OF_STREAM) ? -1 :
+              Long.compare(Long.parseLong(offset1), Long.parseLong(offset2));
+    } catch (NumberFormatException exception) {
+      return null;
+    }
+  }
+
+  @Override
+  public Integer offsetComparator(String offset1, String offset2) {
+    return compareOffsets(offset1, offset2);
```
vjagadish1989: inline the above method here.

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubIME.java
```diff
@@ -0,0 +1,51 @@
...
+import com.microsoft.azure.eventhubs.EventData.SystemProperties;
+import org.apache.samza.system.IncomingMessageEnvelope;
+import org.apache.samza.system.SystemStreamPartition;
+
+import java.util.Map;
+
+/**
+ * Extension of {@link IncomingMessageEnvelope} which contains {@link EventData} system and user properties metadata
+ */
+public class EventHubIME extends IncomingMessageEnvelope {
```
vjagadish1989: s/EventHubIME/EventHubMessageEnvelope

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubIME.java
```diff
@@ -0,0 +1,51 @@
...
+  private SystemProperties systemProperties;
+  private Map<String, Object> userProperties;
```
vjagadish1989: Should we just return `EventData` here instead?

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -0,0 +1,375 @@
...
+  @Override
+  public Map<SystemStreamPartition, List<IncomingMessageEnvelope>> poll(Set<SystemStreamPartition> systemStreamPartitions, long timeout) throws InterruptedException {
+    if (perminentEventHubError != null) {
+      String msg = String.format("Received a non transient error from event hub partition receiver (ssp=%s)", failedSSPReceiver);
+      throw new SamzaException(msg, perminentEventHubError);
+    }
+    return super.poll(systemStreamPartitions, timeout);
+  }
+
+  private void renewPartitionReceiver(SystemStreamPartition ssp, Throwable throwable) {
```
vjagadish1989: Can log the exception at the caller, and not take the throwable here.

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -0,0 +1,375 @@
...
+
+    try {
+      PartitionReceiver receiver = eventHubClientManager.getEventHubClient()
+              .createReceiverSync(consumerGroup, partitionId.toString(), offset,
+                      !offset.equals(EventHubSystemConsumer.START_OF_STREAM));
+
+      // Timeout for EventHubClient receive
+      receiver.setReceiveTimeout(DEFAULT_EVENTHUB_RECEIVER_TIMEOUT);
+      receiver.setReceiveHandler(streamPartitionHandlers.get(ssp));
+      streamPartitionReceivers.put(ssp, receiver);
```
vjagadish1989: Should the receiver be started? Or is it done automatically>

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -0,0 +1,375 @@
...
+      this.readLatency = readLatency;
+      errorRate = readErrors;
+      this.serde = serde;
+    }
+
+    @Override
+    public void onReceive(Iterable<EventData> events) {
+      if (events != null) {
+        events.forEach(event -> {
+            byte[] decryptedBody = event.getBytes();
```
vjagadish1989: Remove all references to encryption/decryption. Can use a message interceptor instead

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -0,0 +1,375 @@
...
+                        namespace, entityPath, partitionId), e);
+      }
+      LOG.debug(String.format("Connection successfully started for namespace=%s, entity=%s ", namespace, entityPath));
+
+    }
+  }
+
+  @Override
+  public Map<SystemStreamPartition, List<IncomingMessageEnvelope>> poll(Set<SystemStreamPartition> systemStreamPartitions, long timeout) throws InterruptedException {
+    if (perminentEventHubError != null) {
```
vjagadish1989: s/perminent/permenant

samza-azure/src/main/java/org/apache/samza/system/eventhub/metrics/SamzaHistogram.java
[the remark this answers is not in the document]
vjagadish1989: @shanthoosh  You are right. this PR is intentionally kept separate from that.

vjagadish1989 (approved):
looks great. approved.

thank you for the contribution.

PR #309, opened 2017-10-03
SAMZA-1056: Added wiring for High Level API state stores, their serdes and changelogs.

prateekm:
Provided join operator access to durable state stores.


samza-core/src/main/java/org/apache/samza/operators/functions/PartialJoinFunction.java
```diff
@@ -47,23 +48,6 @@
    *
    * @return the key value store containing the state for this stream
    */
-  KeyValueStore<K, PartialJoinMessage<M>> getState();
+  KeyValueStore<K, TimestampedValue<M>> getState();
```
vjagadish1989: thanks! `TimestampedValue` is much nicer than `PartialJoinMessage`!

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
```diff
@@ -71,54 +63,29 @@
...
-    if (otherMessage != null && otherMessage.getReceivedTimeMs() > now - ttlMs) {
-      RM joinResult = thisPartialJoinFn.apply(message, otherMessage.getMessage());
+    if (otherMessage != null && otherMessage.getTimestamp() > now - ttlMs) {
+      JM joinResult = thisPartialJoinFn.apply(message, otherMessage.getValue());
       return Collections.singletonList(joinResult);
     }
     return Collections.emptyList();
   }
 
   @Override
```
vjagadish1989: Do we rely on rocksDb ttl for expiry? Instead of doing point-deletes on the store?

samza-core/src/main/java/org/apache/samza/operators/functions/PartialJoinFunction.java
vjagadish1989: Should this be 
`KVStore<TimeSeriesKey<K>, V>`? Why do we need to store the timestamp as a part of the value?  
You can also perhaps use the same operator for doing windowed-joins later?

prateekm: Because we need to look up by key for doing joins b/w the two streams, and timestamps are not part of the key. I don't want to use a timeline store for this version since all queries would be range queries which is not caching-friendly. We can refactor if necessary when we do other kind of joins.

samza-core/src/main/java/org/apache/samza/operators/impl/PartialJoinOperatorImpl.java
prateekm: Yeah, we configure RocksDb TTL  + changelog TTL == ttlMs.

prateekm:
@nickpan47 @jmakes FYI

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimestampedValueSerde.java
```diff
@@ -0,0 +1,50 @@
...
+  private final Serde<V> vSerde;
+
+  public TimestampedValueSerde(Serde<V> vSerde) {
+    this.vSerde = vSerde;
+  }
+
+  @Override
+  public TimestampedValue<V> fromBytes(byte[] bytes) {
+    ByteBuffer bb = ByteBuffer.wrap(bytes);
+    byte[] vBytes = new byte[bytes.length - 8 /* long ts bytes */];
```
vjagadish1989: Minor: use a constant `TIMESTAMP_LEN` here instead of 8?

samza-core/src/main/java/org/apache/samza/operators/impl/store/TimestampedValueSerde.java
```diff
@@ -0,0 +1,50 @@
...
+
+public class TimestampedValueSerde<V> implements Serde<TimestampedValue<V>> {
+  private final Serde<V> vSerde;
+
+  public TimestampedValueSerde(Serde<V> vSerde) {
+    this.vSerde = vSerde;
+  }
+
+  @Override
+  public TimestampedValue<V> fromBytes(byte[] bytes) {
```
vjagadish1989: Would be good to verify the behavior with empty `vBytes` with a test!

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -231,6 +230,17 @@ public void runApplication(StreamApplication streamApplication, String appName,
...
+    // This is to prevent tests from taking a long time to stop after they're done. The issue is that
+    // tearDown currently doesn't call runner.kill(app), and shuts down the Kafka and ZK servers immediately.
+    // The test process then exits, triggering the SamzaContainer shutdown hook, which in turn tries to flush any
+    // store changelogs, which then get stuck trying to produce to the stopped Kafka server.
+    // Calling runner.kill doesn't work since RemoteApplicationRunner creates a new ThreadJob instance when
+    // kill is called. We can't use LocalApplicationRunner since ZkJobCoordinator doesn't currently create
+    // changelog streams. Hence we just force an unclean shutdown here to. This _should be_ OK
+    // since the test method has already executed by the time the shutdown hook is called. The side effect is
+    // that buffered state (e.g. changelog contents) might not be flushed correctly after the test run.
+    configs.put("task.shutdown.ms", "100");
```
vjagadish1989: can't we simply set it to `1` then?

vjagadish1989 (approved):
approved. looks good to me.

PR #313, opened 2017-10-04
SAMZA-1439: Address late review feedback from SAMZA-1434

jmakes:


jmakes:
@xinyuiscool @sborya can you guys take a look?

PR #314, opened 2017-10-04
SAMZA-1442: use systemAdmin.validateStream in KafkaCheckpointManager

sborya:


samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -252,8 +253,30 @@ class KafkaCheckpointManager(
   }
 
   override def start {
+    /*
     kafkaUtil.createTopic(checkpointTopic, 1, replicationFactor, checkpointTopicProperties)
     kafkaUtil.validateTopicPartitionCount(checkpointTopic, systemName, metadataStore, 1, failOnCheckpointValidation)
+    */
+    val CHECKPOINT_STREAMID = "unused-temp-checkpoint-stream-id"
+    val spec = new KafkaStreamSpec(CHECKPOINT_STREAMID,
```
jmakes: `val spec = KafkaStreamSpec.fromSpec(StreamSpec.createCheckpointStreamSpec(checkpointTopic, systemName))`

Then you don't need to introduce another dummy streamId

samza-kafka/src/main/scala/org/apache/samza/util/KafkaUtil.scala
```diff
@@ -133,54 +133,6 @@ class KafkaUtil(val retryBackoff: ExponentialSleepStrategy = new ExponentialSlee
   }
 
   /**
-   * Common code to validate partition count in a topic
```
jmakes: org.apache.samza.system.StreamSpec#createCheckpointStreamSpec

samza-kafka/src/main/scala/org/apache/samza/util/KafkaUtil.scala
```diff
@@ -133,54 +133,6 @@ class KafkaUtil(val retryBackoff: ExponentialSleepStrategy = new ExponentialSlee
...
-   *
```
jmakes: Why not delete the createTopic method too?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
jmakes: Or you could roll this line into a KafkaStreamSpec static constructor that does the same thing but also takes replication factor:

KafkaStreamSpec.createCheckpointStreamSpec(topicName, systemName, replicationFactor)

That would still allow you to uphold two principles:
1. Immutable StreamSpecs
2. Only one constant for the checkpoint streamId

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -252,8 +253,30 @@ class KafkaCheckpointManager(
   }
 
   override def start {
+    /*
     kafkaUtil.createTopic(checkpointTopic, 1, replicationFactor, checkpointTopicProperties)
```
fredji97: just delete these two lines.

samza-kafka/src/main/scala/org/apache/samza/util/KafkaUtil.scala
sborya: hm. You are right. It can be deleted too. Will do.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
sborya: done.

PR #317, opened 2017-10-06
SAMZA-1447: Swapping out CLI JobStatusProvider for REST based implementation in samza-rest

abhishekshivanna:
Removing the YarnCliJobStatusProvider since forking a new shell for every request on the JobsResource endpoint is resource intensive.


samza-rest/src/main/java/org/apache/samza/rest/model/ApplicationInfo.java
```diff
@@ -0,0 +1,295 @@
...
+ */
+package org.apache.samza.rest.model;
+
+import java.util.Objects;
+import org.codehaus.jackson.annotate.JsonProperty;
+
+
+public class ApplicationInfo {
+  @JsonProperty("finishTime")
+  private long _finishTime;
```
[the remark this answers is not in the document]
prateekm: We don't follow the _ prefix for field in rest of the samza codebase. Even if that's the convention in samza-rest, consider making the field name match the property name and removing the JsonProperty annotation in this class.

samza-rest/src/main/java/org/apache/samza/rest/model/ApplicationInfo.java
```diff
@@ -0,0 +1,295 @@
...
+  private int _runningContainers;
+
+  @JsonProperty("memorySeconds")
+  private int _memorySeconds;
+
+  @JsonProperty("vcoreSeconds")
+  private int _vcoreSeconds;
+
+  ApplicationInfo() {
+
```
prateekm: Remove newline.

samza-rest/src/main/java/org/apache/samza/rest/model/ApplicationInfo.java
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  public ApplicationInfo(long finishTime, String amContainerLogs, String trackingHistory, String state, String user,
```
prateekm: Do we really care about all of these fields? Maybe only keep the ones we want and allow Jackson to ignore extra fields?

samza-rest/src/main/java/org/apache/samza/rest/model/ApplicationInfo.java
```diff
@@ -0,0 +1,295 @@
...
+    _queue = queue;
+    _allocatedMB = allocatedMB;
+    _allocatedVCores = allocatedVCores;
+    _runningContainers = runningContainers;
+    _memorySeconds = memorySeconds;
+    _vcoreSeconds = vcoreSeconds;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
prateekm: Let's not override hashcode and equals. There's no particular reason to expect content based equality for this class.

samza-rest/src/main/java/org/apache/samza/rest/model/ApplicationInfoWrapper.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Objects;
+
+
+@JsonRootName("apps")
+public class ApplicationInfoWrapper {
+  @JsonProperty("app")
+  private List<ApplicationInfo> apps;
+
+  ApplicationInfoWrapper() {
+
```
prateekm: Remove newline.

samza-rest/src/main/java/org/apache/samza/rest/model/ApplicationInfoWrapper.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package org.apache.samza.rest.model;
+
+import org.codehaus.jackson.annotate.JsonProperty;
+import org.codehaus.jackson.map.annotate.JsonRootName;
+import java.util.List;
+import java.util.Objects;
+
+
+@JsonRootName("apps")
```
prateekm: Just wondering: Why is this required?

samza-rest/src/main/java/org/apache/samza/rest/model/ApplicationInfoWrapper.java
```diff
@@ -0,0 +1,60 @@
...
+public class ApplicationInfoWrapper {
+  @JsonProperty("app")
```
prateekm: Misleading name, should be "apps" since the value is a list.
Even better would be to make this a "application ID" -> ApplicationInfo Map so that we can look up by the id directly (whatever that is).

samza-rest/src/main/java/org/apache/samza/rest/model/ApplicationInfoWrapper.java
```diff
@@ -0,0 +1,60 @@
...
+  public ApplicationInfoWrapper(List<ApplicationInfo> apps) {
+    this.apps = apps;
+  }
+
+  public List<ApplicationInfo> getApps() {
+    return apps;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
prateekm: Don't need to override equals and hashcode here, it's the default behaviour.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnCliJobStatusProvider.java
```diff
@@ -1,154 +0,0 @@
...
-import org.apache.samza.rest.script.ScriptRunner;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-
-/**
- * An implementation of the {@link JobStatusProvider} that retrieves
- * the job status from the YARN command line interface.
- */
-public class YarnCliJobStatusProvider implements JobStatusProvider {
```
prateekm: Check with @jmakes if it's OK to just delete this class. Do we need to keep it around for compatibility?

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
```diff
@@ -0,0 +1,164 @@
...
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(YarnRestJobStatusProvider.class);
+  private final String _apiEndpoint;
+  private final HttpClient _httpClient;
+  private final ObjectMapper _objectMapper;
+
+  public YarnRestJobStatusProvider(JobsResourceConfig config) {
+    YarnJobResourceConfig yarnConfig = new YarnJobResourceConfig(config);
+    this._objectMapper = new ObjectMapper();
+    this._objectMapper.configure(DeserializationConfig.Feature.UNWRAP_ROOT_VALUE, true);
```
prateekm: Why is this required?

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
```diff
@@ -0,0 +1,164 @@
...
+public class YarnRestJobStatusProvider implements JobStatusProvider {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(YarnRestJobStatusProvider.class);
+  private final String _apiEndpoint;
+  private final HttpClient _httpClient;
+  private final ObjectMapper _objectMapper;
+
+  public YarnRestJobStatusProvider(JobsResourceConfig config) {
+    YarnJobResourceConfig yarnConfig = new YarnJobResourceConfig(config);
+    this._objectMapper = new ObjectMapper();
```
prateekm: Can make static final.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
```diff
@@ -0,0 +1,164 @@
...
+    this._objectMapper.configure(DeserializationConfig.Feature.UNWRAP_ROOT_VALUE, true);
+    this._httpClient = new HttpClient();
+    this._apiEndpoint = String.format("http://%s/ws/v1/cluster/apps?states=RUNNING,NEW,NEW_SAVING,SUBMITTED,ACCEPTED",
+        yarnConfig.getYarnResourceManagerEndpoint());
+  }
+
+  private static final String JOB_NAME_ID_FORMAT = "%s_%s";
```
prateekm: Group class fields together before constructor, and static fields before object fields.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
```diff
@@ -0,0 +1,164 @@
...
+
+  @Override
+  public void getJobStatuses(Collection<Job> jobs) throws IOException, InterruptedException {
```
prateekm: It's usually bad practice to mutate passed in data. Can we return a Map<JobName, Status> instead? Also, if we do that, we don't need a Collection<Job>, just a Collection<String> of qualifiedJobName.

prateekm: Is this called per job or once in a while for all jobs in the cluster? Or something in between?

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
```diff
@@ -0,0 +1,164 @@
...
+          job.setStatus(samzaStatus);
+        }
+      }
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to retrieve node info.", e);
+    }
+  }
+
+  @Override
+  public Job getJobStatus(JobInstance jobInstance)
```
prateekm: This is potentially very inefficient. How often is this called?

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
```diff
@@ -0,0 +1,164 @@
...
+    switch (yarnState) {
+      case RUNNING:
+        return JobStatus.STARTED;
+      case NEW:
+      case NEW_SAVING:
+      case SUBMITTED:
+      case ACCEPTED:
+        return JobStatus.STARTING;
+      case FINISHED:
+      case FAILED:
```
prateekm: We don't have a FAILED job status?

samza-rest/src/main/java/org/apache/samza/rest/resources/YarnJobResourceConfig.java
```diff
@@ -0,0 +1,52 @@
...
+   * Specifies the host and port of the YARN ResourceManager against against which the API
+   * calls are made.
+   */
+  private static final String CONFIG_YARN_RESOURCE_MANAGER_API_ENDPOINT = "yarn.resourcemanager.api.endpoint";
+
+  /**
+   * Specifies the default host and port of the YARN ResourceManager assuming that samza-rest is running on the
+   * Resource Manager host.
+   */
+  private static final String CONFIG_DEFAULT_YARN_RESOURCE_MANAGER_API_ENDPOINT = "localhost:8088";
```
prateekm: Why default this to localhost? Also, why is the "assuming that samza-rest is running..." required?

samza-rest/src/main/java/org/apache/samza/rest/resources/YarnJobResourceConfig.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * @see YarnJobResourceConfig#CONFIG_YARN_RESOURCE_MANAGER_API_ENDPOINT
```
prateekm: Add description to javadoc.

samza-rest/src/main/java/org/apache/samza/rest/resources/YarnJobResourceConfig.java
```diff
@@ -0,0 +1,52 @@
...
+public class YarnJobResourceConfig extends JobsResourceConfig {
+  public YarnJobResourceConfig(JobsResourceConfig config) {
+    super(config);
+  }
+
+  /**
+   * Specifies the host and port of the YARN ResourceManager against against which the API
+   * calls are made.
+   */
+  private static final String CONFIG_YARN_RESOURCE_MANAGER_API_ENDPOINT = "yarn.resourcemanager.api.endpoint";
```
jmakes: This should be documented in the config table in jobs.md as an optional config.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
```diff
@@ -0,0 +1,164 @@
...
+  private final String _apiEndpoint;
+  private final HttpClient _httpClient;
+  private final ObjectMapper _objectMapper;
+
+  public YarnRestJobStatusProvider(JobsResourceConfig config) {
+    YarnJobResourceConfig yarnConfig = new YarnJobResourceConfig(config);
+    this._objectMapper = new ObjectMapper();
+    this._objectMapper.configure(DeserializationConfig.Feature.UNWRAP_ROOT_VALUE, true);
+    this._httpClient = new HttpClient();
+    this._apiEndpoint = String.format("http://%s/ws/v1/cluster/apps?states=RUNNING,NEW,NEW_SAVING,SUBMITTED,ACCEPTED",
```
jmakes: This is a bit fragile to changes in the YarnApplicationState enum. It hasn't changed since 2013, but it's still possible on a major version bump.

Is there a performance hit for requesting all states?
If so, maybe stick with this approach but verify that the YarnApplicationState doesn't have any new values unknown to this class (because if it does, they may not be handled properly).

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
```diff
@@ -0,0 +1,164 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(YarnRestJobStatusProvider.class);
+  private final String _apiEndpoint;
+  private final HttpClient _httpClient;
+  private final ObjectMapper _objectMapper;
+
+  public YarnRestJobStatusProvider(JobsResourceConfig config) {
+    YarnJobResourceConfig yarnConfig = new YarnJobResourceConfig(config);
+    this._objectMapper = new ObjectMapper();
+    this._objectMapper.configure(DeserializationConfig.Feature.UNWRAP_ROOT_VALUE, true);
+    this._httpClient = new HttpClient();
```
jmakes: Why not use YarnClient?
https://github.com/apache/hadoop/blob/trunk/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-client/src/main/java/org/apache/hadoop/yarn/client/api/YarnClient.java

Then we wouldn't need the ApplicationInfo classes at all. We could leverage Yarn's ApplicationReport.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnCliJobStatusProvider.java
jmakes: @prateekm, we chatted about it. Looks like this new impl is better in every way. Ok to delete!

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
jmakes: The REST API provides both the simplified JobStatus and the cluster specific status detail. The simplified status is primarily focused on whether or not the job is running or not. It's not really used for diagnosing problems. 

This area of the code was actually my first contribution to Samza, and I have to admit, it's a bit naive in retrospect. But this patch is mostly about producing equivalent behavior with better performance, so I wouldn't recommend changing it at the moment.

samza-rest/src/main/java/org/apache/samza/rest/model/ApplicationInfo.java
```diff
@@ -0,0 +1,294 @@
+/*
```
jmakes: If we keep these classes, recommend nesting this class and the Wrapper class in a /yarn package, since they're specific to YARN.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
vjagadish1989: +1 to changing this interface. As an aside, it will also simplify most of your logic inside the implementation.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
prateekm: Makes sense, we can always add the failed status if we need it.

samza-rest/src/main/java/org/apache/samza/rest/model/ApplicationInfoWrapper.java
abhishekshivanna: YARN has the following response to the /cluster/apps endpoint: https://hadoop.apache.org/docs/stable/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html#Cluster_Applications_API

So this is in order to prevent the need to create a wrapper for the root level object with the key "apps".
The following explains the solution with an example: https://stackoverflow.com/questions/11704255/jackson-json-deserialization-with-root-element

samza-rest/src/main/java/org/apache/samza/rest/model/ApplicationInfoWrapper.java
abhishekshivanna: its unfortunate, but "app" is the name of the LIST of apps that is returned from the API. https://hadoop.apache.org/docs/stable/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html#Cluster_Applications_API

Since the returned JSON data is a list, i'm not sure if making this a Map will work for jackson ser/deser.

samza-rest/src/main/java/org/apache/samza/rest/model/ApplicationInfo.java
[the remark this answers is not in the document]
abhishekshivanna: Agreed. I will rename and remove the annotations.

samza-rest/src/main/java/org/apache/samza/rest/model/ApplicationInfo.java
abhishekshivanna: Agreed. Will ignore fields that are not used.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
abhishekshivanna: This is required for unwrapping the top level key called "apps". Explained in my other comment.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/SimpleYarnJobProxy.java
```diff
@@ -78,7 +79,7 @@ public void stop(JobInstance jobInstance)
     }
 
     String scriptPath = getScriptPath(jobInstance, STOP_SCRIPT_NAME);
-    int resultCode = scriptRunner.runScript(scriptPath, YarnCliJobStatusProvider.getQualifiedJobName(jobInstance));
+    int resultCode = scriptRunner.runScript(scriptPath, YarnRestJobStatusProvider.getQualifiedJobName(jobInstance));
```
shanthoosh: Just a minor suggestion. 

It looks odd to house getQualifiedName method in YarnRestJobStatusProvider class and invoke it just at scriptRunner.runScript as a param.
It could be better to either move that string generation to util method or move it to even jobInstance.toString(there by removing getQualifiedName method).

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
abhishekshivanna: Ideally this would be used when ever the status of a set of jobs needs to be checked. In the specific case of local store gc monitor, this is called once per monitor run for a given host.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
abhishekshivanna: Is called if there is any uses cases for checking a single job's status. Currently only the start/stop in samza-rest uses it.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
```diff
@@ -0,0 +1,164 @@
...
+      case FINISHED:
+      case FAILED:
+      case KILLED:
+      default:
+        return JobStatus.STOPPED;
+    }
+  }
+
+  /**
+   * This method initiates http get request on the request url and returns the
```
vjagadish1989:
```
Issues a HTTP Get request to the provided url and returns the response
@param requestUrl the request url
@return the response
...
```

samza-rest/src/main/java/org/apache/samza/rest/resources/YarnJobResourceConfig.java
```diff
@@ -0,0 +1,53 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.rest.resources;
+
+/**
+ * Configurations for the YARN based {@link JobsResource} endpoint.
+ */
+public class YarnJobResourceConfig extends JobsResourceConfig {
```
vjagadish1989: Can we remove this class and have a single variable instead / inline this in JobsResourceConfig

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
jmakes: ^ Bump because github decided this one is outdated and it's not

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
```diff
@@ -0,0 +1,164 @@
...
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(YarnRestJobStatusProvider.class);
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  private static final String JOB_NAME_ID_FORMAT = "%s_%s";
+  private final String apiEndpoint;
+  private final HttpClient httpClient;
+
+  public YarnRestJobStatusProvider(JobsResourceConfig config) {
+    YarnJobResourceConfig yarnConfig = new YarnJobResourceConfig(config);
+    this.httpClient = new HttpClient();
```
jmakes: Why not use YarnClient?
https://github.com/apache/hadoop/blob/trunk/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-client/src/main/java/org/apache/hadoop/yarn/client/api/YarnClient.java

Then we wouldn't need the ApplicationInfo classes at all. We could leverage Yarn's ApplicationReport.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
```diff
@@ -0,0 +1,164 @@
...
+    OBJECT_MAPPER.configure(DeserializationConfig.Feature.UNWRAP_ROOT_VALUE, true);
+    this.apiEndpoint = String.format("http://%s/ws/v1/cluster/apps?states=RUNNING,NEW,NEW_SAVING,SUBMITTED,ACCEPTED",
```
jmakes: This is a bit fragile to changes in the YarnApplicationState enum. It hasn't changed since 2013, but it's still possible on a major version bump.

Is there a performance hit for requesting all states?
If so, maybe stick with this approach but verify that the YarnApplicationState doesn't have any new values unknown to this class (because if it does, they may not be handled properly).

samza-rest/src/main/java/org/apache/samza/rest/resources/YarnJobResourceConfig.java
abhishekshivanna: JobResourceConfig is agnostic to which cluster manager runs the jobs. Therefore having a YARN related config in JobResourceConfig doesn't seem right.

vjagadish1989: thanks for clarifying.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
abhishekshivanna: Built a working prototype of the same provider using YarnClient. It seems to work perfectly fine as long as the RM address points to an Active RM.
If the RM endpoint is configured to a Passive RM, YarnClient throws a `java.net.ConnectException: Connection refused`.

On the other hand, hitting the YARN Http REST endpoint on a Passive RM returns a HTTP 302 redirect. Therefore, it always works regardless of which RM's (Active/Passive) JobsResource endpoint is hit, but this would require us to maintain models similar to what YarnClient already gives us by default.

Not sure how to proceed here.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
abhishekshivanna: I was only trying to reduce the number of records we had to iterate over. Removing the state filters doesn't have any significant performance hits.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/YarnRestJobStatusProvider.java
jmakes: Discussed in person. There doesn't seem to be any simple workaround for the YarnClient approach, so I'm ok with the HttpClient. It's still better than the CLI implementation.

vjagadish1989 (approved):
looks good to me approved.

vjagadish1989:
Thanks for the patch, @abhishekshivanna ! Merged to open source

PR #318, opened 2017-10-06
SAMZA-272, SAMZA-1440, SAMZA-1269: Fixed thread interrupt tests in TestExponentialSleepStrategy.

prateekm:


samza-core/src/test/scala/org/apache/samza/util/TestExponentialSleepStrategy.scala
```diff
@@ -116,45 +118,57 @@ class TestExponentialSleepStrategy {
...
       def run {
         try { operation } catch { case e: Exception => exception = Some(e) }
       }
     })
-    interruptee.start
-    Thread.sleep(10) // give the thread a chance to make some progress before we interrupt it
-    interruptee.interrupt
-    interruptee.join
+    interruptee.start()
+    latch.await()
```
jmakes: Would prefer the following so the test doesn't hang

assertTrue("Latch timed out", latch.await(someHighNumber, someUnit))

PR #322, opened 2017-10-11
Clean up interrupted thread bugs

isolis:
Call Thread.currentThread().interrupt(); when capturing InterruptedException


vjagadish1989 (approved):
looks good to me, approved thanks!

PR #324, opened 2017-10-11
SAMZA-1454: Globally unique and user settable IDs for stateful operators

prateekm:


prateekm:
cc @vjagadish1989, @nickpan47 for review.
cc @jmakes, @xinyuiscool, FYI.

vjagadish1989:
@prateekm Personally, I'd prefer names be a required parameter as well for now. It keeps the API simple instead of having 4 different variants of `partitionby`. 

We can certainly add another variant with auto-generated names later.

prateekm:
Made the operator IDs required for stateful operation after offline discussion with @nickpan47 and @cpettitt-linkedin.

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -186,8 +193,26 @@ public ContextManager getContextManager() {
...
-    return this.opId++;
+  /**
+   * Gets the unique ID for the next operator in the graph. The ID is of the following format:
+   * jobName-jobId-opCode-(opName|nextOpNum);
+   *
+   * @param opCode the {@link OpCode} of the next operator
+   * @param opName the optional user-provided name of the next operator or null
+   * @return the unique ID for the next operator in the graph
+   */
+  /* package private */ String getNextOpId(OpCode opCode, String opName) {
```
vjagadish1989: s/opName/userProvidedName since it's more explicit

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpecs.java
```diff
@@ -43,6 +44,22 @@
   private OperatorSpecs() {}
 
   /**
+   * Creates an {@link InputOperatorSpec} for consuming input.
+   * @param streamSpec  the stream spec for the input stream
```
vjagadish1989: Minor: new line between description and `@param`

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpecs.java
```diff
@@ -203,7 +220,7 @@ public void close() {
    */
   public static <K, M, OM, JM> JoinOperatorSpec<K, M, OM, JM> createJoinOperatorSpec(
       OperatorSpec<?, M> leftInputOpSpec, OperatorSpec<?, OM> rightInputOpSpec, JoinFunction<K, M, OM, JM> joinFn,
-      Serde<K> keySerde, Serde<M> messageSerde, Serde<OM> otherMessageSerde, long ttlMs, int opId) {
+      Serde<K> keySerde, Serde<M> messageSerde, Serde<OM> otherMessageSerde, long ttlMs, String opId) {
```
vjagadish1989: Now that the `OperatorSpecs` appears to just return new instances from the passed-in parameters, we can get rid off it in a later refactor

samza-core/src/test/java/org/apache/samza/execution/TestExecutionPlanner.java
```diff
@@ -111,7 +111,7 @@ private StreamGraphImpl createSimpleGraph() {
     MessageStream<KV<Object, Object>> input1 = streamGraph.getInputStream("input1");
     OutputStream<KV<Object, Object>> output1 = streamGraph.getOutputStream("output1");
     input1
-        .partitionBy(m -> m.key, m -> m.value)
+        .partitionBy(m -> m.key, m -> m.value, "p3")
```
vjagadish1989: s/p3/p1

samza-core/src/test/java/org/apache/samza/operators/TestJoinOperator.java
```diff
@@ -77,6 +78,25 @@ public void join() throws Exception {
...
+    StreamApplication app = new StreamApplication() {
+      @Override
+      public void init(StreamGraph graph, Config config) {
+        IntegerSerde integerSerde = new IntegerSerde();
+        KVSerde<Integer, Integer> kvSerde = KVSerde.of(integerSerde, integerSerde);
+        MessageStream<KV<Integer, Integer>> inStream = graph.getInputStream("instream", kvSerde);
+
+        SystemStream outputSystemStream = new SystemStream("outputSystem", "outputStream");
+        inStream
+            .join(inStream, new TestJoinFunction(), integerSerde, kvSerde, kvSerde, JOIN_TTL, "join")
```
vjagadish1989: minor: sink is probably unnecessary for this test

samza-core/src/test/java/org/apache/samza/operators/TestJoinOperator.java
```diff
@@ -77,6 +78,25 @@ public void join() throws Exception {
...
+            .sink((m, mc, tc) -> mc.send(new OutgoingMessageEnvelope(outputSystemStream, m)));
+      }
+    };
+
+    createStreamOperatorTask(new SystemClock(), app);
```
vjagadish1989: Perhaps, add a comment that this line should throw?

samza-core/src/test/java/org/apache/samza/operators/TestWindowOperator.java
```diff
@@ -93,7 +93,8 @@ public void testTumblingWindowsDiscardingMode() throws Exception {
     TestClock testClock = new TestClock();
     StreamOperatorTask task = new StreamOperatorTask(sgb, runner, testClock);
     task.init(config, taskContext);
-    MessageCollector messageCollector = envelope -> windowPanes.add((WindowPane<Integer, Collection<IntegerEnvelope>>) envelope.getMessage());
+    MessageCollector messageCollector =
```
vjagadish1989: 👍

samza-core/src/test/java/org/apache/samza/example/PageViewCounterExample.java
```diff
@@ -56,7 +56,7 @@
     pageViewEvents
         .window(Windows.keyedTumblingWindow(m -> m.memberId, Duration.ofSeconds(10), initialValue, foldLeftFn, null, null)
             .setEarlyTrigger(Triggers.repeat(Triggers.count(5)))
-            .setAccumulationMode(AccumulationMode.DISCARDING))
+            .setAccumulationMode(AccumulationMode.DISCARDING), "window")
```
vjagadish1989: trivial: window/tumblingWindow

vjagadish1989 (approved):
approved. Looks pretty good, minor feedback!

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpecs.java
prateekm: +1, it's always been this way.

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -72,54 +74,62 @@ public MessageStreamImpl(StreamGraphImpl graph, OperatorSpec<?, M> operatorSpec)
...
+    String opId = this.graph.getNextOpId(OpCode.SINK, null);
+    SinkOperatorSpec<M> op = OperatorSpecs.createSinkOperatorSpec(sinkFn, opId);
     this.operatorSpec.registerNextOperatorSpec(op);
   }
 
   @Override
   public void sendTo(OutputStream<M> outputStream) {
-    OutputOperatorSpec<M> op = OperatorSpecs.createSendToOperatorSpec(
-        (OutputStreamImpl<M>) outputStream, this.graph.getNextOpId());
+    String opId = this.graph.getNextOpId(OpCode.SEND_TO, null);
```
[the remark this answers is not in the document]
prateekm: Not sure if that'll be any cleaner, since the version with explicit name will have to accept nulls anyway (so that the version without names can call it). Do you prefer one over the other?

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -72,54 +74,62 @@ public MessageStreamImpl(StreamGraphImpl graph, OperatorSpec<?, M> operatorSpec)
 
   @Override
   public <TM> MessageStream<TM> map(MapFunction<? super M, ? extends TM> mapFn) {
-    OperatorSpec<M, TM> op = OperatorSpecs.createMapOperatorSpec(mapFn, this.graph.getNextOpId());
+    String opId = this.graph.getNextOpId(OpCode.MAP, null);
```
nickpan47: minor: it seems that getNextOpId(code, null) was repeated many times. Might be better to add another getNextOpId(code) to avoid typing null repeatedly.  But, you call.

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -72,54 +74,62 @@ public MessageStreamImpl(StreamGraphImpl graph, OperatorSpec<?, M> operatorSpec)
...
-    OutputOperatorSpec<M> op = OperatorSpecs.createSendToOperatorSpec(
-        (OutputStreamImpl<M>) outputStream, this.graph.getNextOpId());
+    String opId = this.graph.getNextOpId(OpCode.SEND_TO, null);
+    OutputOperatorSpec<M> op = OperatorSpecs.createSendToOperatorSpec((OutputStreamImpl<M>) outputStream, opId);
     this.operatorSpec.registerNextOperatorSpec(op);
   }
 
   @Override
-  public <K, WV> MessageStream<WindowPane<K, WV>> window(Window<M, K, WV> window) {
+  public <K, WV> MessageStream<WindowPane<K, WV>> window(Window<M, K, WV> window, String userProvidedId) {
```
nickpan47: nit: userDefinedId?

samza-core/src/main/java/org/apache/samza/operators/StreamGraphImpl.java
```diff
@@ -186,8 +193,26 @@ public ContextManager getContextManager() {
...
+    String nextOpId = String.format("%s-%s-%s-%s",
+        config.get(JobConfig.JOB_NAME()),
```
nickpan47: Shouldn't this be appName? If we create multi-job applications later, this will mandate migration of state/streams.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -258,7 +258,7 @@ public final void aggregateEndOfStream(EndOfStreamMessage eos, SystemStreamParti
    * @param collector message collector
    * @param coordinator task coordinator
    */
-  private final void onEndOfStream(MessageCollector collector, TaskCoordinator coordinator) {
+  private void onEndOfStream(MessageCollector collector, TaskCoordinator coordinator) {
```
nickpan47: Why do we remove the final declarative here? I thought that the reason Xinyu marked this as final is no operator should override this, since it is the system-controlled behavior.

samza-core/src/test/java/org/apache/samza/operators/TestWindowOperator.java
```diff
@@ -79,7 +79,7 @@ public void setup() throws Exception {
     when(taskContext.getSystemStreamPartitions()).thenReturn(ImmutableSet
         .of(new SystemStreamPartition("kafka", "integers", new Partition(0))));
     when(taskContext.getMetricsRegistry()).thenReturn(new MetricsRegistryMap());
-    when(taskContext.getStore("window-3")).thenReturn(new TestInMemoryStore<>(storeKeySerde, storeValSerde));
+    when(taskContext.getStore("null-null-window-w1")).thenReturn(new TestInMemoryStore<>(storeKeySerde, storeValSerde));
```
nickpan47: nit: would be nice to set the actual mock strings instead of using null here.

samza-core/src/test/java/org/apache/samza/operators/impl/TestOperatorImplGraph.java
```diff
@@ -221,14 +222,14 @@ public void testJoinChain() {
...
     MessageStream<Object> inputStream2 = streamGraph.getInputStream("input2", new NoOpSerde<>());
     inputStream1.join(inputStream2, mockJoinFunction,
-        mock(Serde.class), mock(Serde.class), mock(Serde.class), Duration.ofHours(1));
+        mock(Serde.class), mock(Serde.class), mock(Serde.class), Duration.ofHours(1), "j1");
 
     TaskContextImpl mockTaskContext = mock(TaskContextImpl.class);
     when(mockTaskContext.getMetricsRegistry()).thenReturn(new MetricsRegistryMap());
     KeyValueStore mockLeftStore = mock(KeyValueStore.class);
-    when(mockTaskContext.getStore(eq("join-2-L"))).thenReturn(mockLeftStore);
+    when(mockTaskContext.getStore(eq("null-null-join-j1-L"))).thenReturn(mockLeftStore);
```
nickpan47: Same here.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -105,15 +105,19 @@
    * <p>
    * Use the {@link org.apache.samza.operators.windows.Windows} helper methods to create the appropriate windows.
    * <p>
-   * <b>Warning:</b> As of version 0.13.0, messages in windows are kept in memory and will be lost during restarts.
+   * The {@code id} is used as part of the unique ID for any state stores and streams created by this transform.
```
nickpan47: In the test code, it seems that the same id + different opCode is allowed in the same application. It would be nice to clarify what "part of the unique ID" means here. Are we explicitly asking user to supply a fully-qualified unique ID in the application here as id? Or just unique ID for each type of transform in his application? Either way, need to be explicit here. The implementation may not need to change at this point, but the user expectation needs to be set clearly.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
prateekm: Private methods are final anyway, so was redundant. But I'll add it back so that this doesn't get changed on a future refactor.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
prateekm: I'd prefer just saying "unique in the application", since that's easy to understand and implies unique ID per operator type. Will clarify in documentation.

PR #328, opened 2017-10-16
SAMZA-1457: Set retention for internal streams for Batch application

xinyuiscool:
For intermediate streams, checkpoint and changelog, we need to set a short retention period for batch.


samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaStreamSpec.java
```diff
@@ -167,6 +167,10 @@ public StreamSpec copyWithPartitionCount(int partitionCount) {
     return new KafkaStreamSpec(getId(), getPhysicalName(), getSystemName(), partitionCount, getReplicationFactor(), getProperties());
   }
 
+  public KafkaStreamSpec copyWithProperties(Properties properties) {
```
vjagadish1989: minor: add java docs

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
```diff
@@ -48,9 +49,24 @@ object KafkaCheckpointManagerFactory {
     } else {
       new KafkaConfig(config).getCheckpointSegmentBytes()
     }
-    (new Properties /: Map(
-      "cleanup.policy" -> "compact",
-      "segment.bytes" -> String.valueOf(segmentBytes))) { case (props, (k, v)) => props.put(k, v); props }
+
+    val isStreamMode = if (config != null) {
```
vjagadish1989: since,`STREAM` is the default anyways, can we rewrite this to be more concise?

val isStreamMode = appConfig.getAppMode == ApplicationMode.STREAM

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
```diff
@@ -48,9 +49,24 @@ object KafkaCheckpointManagerFactory {
...
+      val appConfig = new ApplicationConfig(config)
+      appConfig.getAppMode == ApplicationMode.STREAM
+    } else {
+      true
+    }
+
+    if (isStreamMode) {
```
vjagadish1989: Can rewrite this to something simpler?
For instance:
```
    val props = new Properties()
    props.putAll(ImmutableMap.of(
      ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, brokers,
      ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, defaultSerializer))
    props
```

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
```diff
@@ -48,9 +49,24 @@ object KafkaCheckpointManagerFactory {
...
+      (new Properties /: Map(
+        "cleanup.policy" -> "compact",
+        "segment.bytes" -> String.valueOf(segmentBytes))) { case (props, (k, v)) => props.put(k, v); props }
+    } else {
+      (new Properties /: Map(
+        "cleanup.policy" -> "compact,delete",
```
vjagadish1989: Can we consolidate this logic, and the one in `KafkaConfig` in one place?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -139,7 +139,12 @@ class KafkaSystemAdmin(
    * Replication factor for the Changelog topic in kafka
    * Kafka properties to be used during the Changelog topic creation
    */
-  topicMetaInformation: Map[String, ChangelogInfo] = Map[String, ChangelogInfo]()) extends ExtendedSystemAdmin with Logging {
+  topicMetaInformation: Map[String, ChangelogInfo] = Map[String, ChangelogInfo](),
+
+  /**
+   * Kafka properties to be used during the intermediate topic creation
+   */
+  intermediateStreamProperties: Map[String, Properties] = Map()) extends ExtendedSystemAdmin with Logging {
```
vjagadish1989: s/intermediateStreamProperties/streamProperties
Can also consolidate this with other streams like coordinator and use a single map perhaps?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
xinyuiscool: Good suggestion. I moved the logic there and now it looks much cleaner.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
xinyuiscool: I think this is beyond the scope of this pr. The refactoring should be separate patch and needs more testing than what I have it here.

vjagadish1989: In that case do we really need this extra stateful parameter at all? Can't we simply use the properties from the `StreamSpec` class?

The `createStream(StreamSpec)` can internally construct the right `KafkaStreamSpec` by parsing the properties from the passed in `StreamSpec` instance ?

xinyuiscool: Passing in specific kafka properties in StreamSpec and then construct KafkaStreamSpec from it seems not a good idea. As far as I understand, StreamSpec should be general and can be materialized into different physical stream.

samza-kafka/src/test/scala/org/apache/samza/checkpoint/kafka/TestKafkaCheckpointManager.scala
```diff
@@ -56,7 +56,7 @@ class TestKafkaCheckpointManager extends KafkaServerTestHarness {
 
   val checkpointTopic = "checkpoint-topic"
   val serdeCheckpointTopic = "checkpoint-topic-invalid-serde"
-  val checkpointTopicConfig = KafkaCheckpointManagerFactory.getCheckpointTopicProperties(null)
+  val checkpointTopicConfig = new org.apache.samza.config.KafkaConfig(new MapConfig()).getCheckpointTopicProperties()
```
vjagadish1989: need not use the fully qualified names here.

xinyuiscool: Actually we are using the KafkaConfig from kafka-server in this class so I have to use the full name here.

PR #331, opened 2017-10-17
SAMZA-1465: Performance regression for KafkaCheckpointManager

jmakes:


samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -219,16 +221,17 @@ class KafkaCheckpointManager(
 
         val messages: util.List[IncomingMessageEnvelope] = envelopes.get(ssp)
         val messagesNum = if (messages != null) messages.size else 0
-        info("CheckpointMgr read %s envelopes (%s messages) from ssp %s. Current offset is %s, newest is %s"
+        debug("CheckpointMgr read %s envelopes (%s messages) from ssp %s. Current offset is %s, newest is %s"
```
prateekm: While we're at it, let's fix this message too:

`"Read % messages from ssp %s. Current offset is %s, newest is %s", format(messagesNum, ssp, currentOffset, newestOffset)`
Note the order of params.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -181,73 +183,40 @@ class KafkaCheckpointManager(
...
-    var attempts = 10
-    val POLL_TIMEOUT = 1000L
+    info("Reading from checkpoint system:%s topic:%s" format(systemName, checkpointTopic))
 
     val ssp: SystemStreamPartition = new SystemStreamPartition(systemName, checkpointTopic, new Partition(0))
-    val systemConsumer = getSystemConsumer()
     val partitionMetadata = getSSPMetadata(checkpointTopic, new Partition(0))
-    // offsets returned are strings
-    val newestOffset = if (partitionMetadata.getNewestOffset == null) UNKNOWN_OFFSET else partitionMetadata.getNewestOffset
     val oldestOffset = partitionMetadata.getOldestOffset
```
sborya: this line should be moved into the lazy consumer instantiation block (together with 189)

sborya: should we move this line into the if (consumer == null) block?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -181,73 +183,41 @@ class KafkaCheckpointManager(
...
+    var msgCount = 0
+    while (iterator.hasNext) {
+      val msg = iterator.next
+      msgCount += 1
+
+      val key = msg.getKey.asInstanceOf[Array[Byte]]
+      val offset = msg.getOffset
+      if (key == null) {
+        throw new KafkaUtilException("While reading checkpoint (currentOffset=%s) stream encountered message without key."
+                                             format offset)
```
prateekm: Minor: Let's fix formatting, line width.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -181,73 +183,41 @@ class KafkaCheckpointManager(
...
-              handleEntry(checkpointPayload, checkpointKey)
-            }
-          }
-        }
+      val checkpointKey = KafkaCheckpointLogKey.fromBytes(key)
+
+      if (!shouldHandleEntry(checkpointKey)) {
+        info("Skipping checkpoint log entry at offset %s with key %s." format(offset, checkpointKey))
+      } else {
+        // handleEntry requires ByteBuffer
```
prateekm: Can delete comment.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -64,14 +64,14 @@ class KafkaCheckpointManager(
                               checkpointTopicProperties: Properties = new Properties) extends CheckpointManager with Logging {
 
   var taskNames = Set[TaskName]()
-  @volatile var  systemProducer: SystemProducer = null
+  @volatile var systemProducer: SystemProducer = null
+  var systemConsumer: SystemConsumer = null
```
prateekm: Maybe instead of the comment on line 120 say something like "This class is thread safe for writing but not for reading checkpoints. This is currently OK since checkpoints are only on the main thread."

PR #336, opened 2017-10-25
StreamOperatorTask does not need to be final.

sborya:


prateekm:
@sborya this was intentional. Why does this need to be not final?

PR #337, opened 2017-10-25
TestExecutionPlanner compilation error

sborya:


prateekm:
Also don't think this should be required with the latest master since it compiles fine for me. Are you seeing issues compiling with Scala 2.10?

PR #338, opened 2017-10-25
SAMZA-1470: Wrong job status returned by YarnRestJobStatusProvider when there are multiple app attempts

jmakes:


jmakes:
Eww look at that commit history. Closing this PR in favor of #339

PR #341, opened 2017-10-26
Checkpoint manager implementation with Azure Table

dxichen:
@vjagadish1989 
WIP, will add tests


samza-azure/src/main/java/org/apache/samza/checkpoint/azure/AzureCheckpointManager.java
```diff
@@ -0,0 +1,139 @@
...
+    try {
+      cloudTable.execute(batchOperation);
+    } catch (StorageException e) {
+      LOG.error("Connection string {} specifies an invalid key.", storageConnectionString);
+      throw new AzureException(e);
+    }
+  }
+
+  @Override
+  public Checkpoint readLastCheckpoint(TaskName taskName) {
```
vjagadish1989: Would be nice to add a diagram with the fields in the relational table, and how checkpoints corresponding to each of the tasks are stored and retrieved.

samza-azure/src/main/java/org/apache/samza/checkpoint/azure/AzureCheckpointManager.java
```diff
@@ -0,0 +1,139 @@
...
+      offsets.put(new SystemStreamPartition(systemName, streamName, new Partition(partitionId)),
+              taskCheckpointEntity.getOffset());
+    }
+    return new Checkpoint(offsets);
+  }
+
+  @Override
+  public void stop() {
+    try {
+      cloudTable.deleteIfExists();
```
vjagadish1989: If we delete the table when `checkpointManager.stop` is called, then does this mean that the persisted checkpoint data is lost?

We want to retain checkpoint data to persist even after the job is stopped, because during a restart we want to load from our previous checkpoint.

samza-azure/src/main/java/org/apache/samza/checkpoint/azure/AzureCheckpointManager.java
```diff
@@ -0,0 +1,139 @@
...
+            .where(partitionFilter);
+    Map<SystemStreamPartition, String> offsets = new HashMap<>();
+    for (TaskCheckpointEntity taskCheckpointEntity : cloudTable.execute(partitionQuery)) {
+      String systemName = taskCheckpointEntity.getRowKey();
+      String streamName = taskCheckpointEntity.getStreamName();
+      int partitionId = taskCheckpointEntity.getPartitionId();
+      offsets.put(new SystemStreamPartition(systemName, streamName, new Partition(partitionId)),
+              taskCheckpointEntity.getOffset());
+    }
+    return new Checkpoint(offsets);
```
vjagadish1989: Would be nice to wrap this up in an `ImmutableMap` so that we guard against further mutations on the map. 
```new Checkpoint(ImmutableMap.of(offsets)))```

File: build.gradle [hunk 4]
```diff
@@ -202,6 +202,10 @@ project(':samza-azure') {
     configFile = new File(rootDir, "checkstyle/checkstyle.xml")
     toolVersion = "$checkstyleVersion"
   }
+  test {
+    // Exclude integration tests that require connection to Azure Storage
+    exclude 'org/apache/samza/checkpoint/azure/*ITest*'
```
vjagadish1989: Would be nice to use one consistent way to ignore integration tests

samza-azure/src/main/java/org/apache/samza/checkpoint/azure/AzureCheckpointManager.java
```diff
@@ -0,0 +1,248 @@
...
+  @Override
+  public void clearCheckpoints() {
+    LOG.debug("Clearing all checkpoints in Azure table");
+
+    for (TaskName taskName : taskNames) {
+      String partitionQueryKey = taskName.toString();
+
+      // Generate table query
+      String partitionFilter = TableQuery.generateFilterCondition(
+              PARTITION_KEY,
```
vjagadish1989: Use 4 space indents consistently.

samza-azure/src/main/java/org/apache/samza/checkpoint/azure/AzureCheckpointManager.java
```diff
@@ -0,0 +1,248 @@
...
+        TaskCheckpointEntity taskCheckpoint = new TaskCheckpointEntity(taskName.toString(),
+                serializeSystemStreamPartition(ssp));
+        taskCheckpoint.setOffset(offset);
+
+        // Add to batch operation
+        batchOperation.insertOrReplace(taskCheckpoint);
+
+        // Execute when batch reaches capacity
+        if (batchOperation.size() >= MAX_WRITE_BATCH_SIZE) {
+          try {
```
vjagadish1989: consider consolidating this and the block in L:131-138

```
if (batchOperation.size() >= MAX_WRITE_BATCH_SIZE || !iterator.hasNext()) {
  table.execute(batchOperation)
}
```
This will get rid off the special redundant code to process the last batch

samza-azure/src/main/java/org/apache/samza/checkpoint/azure/AzureCheckpointManager.java
```diff
@@ -0,0 +1,248 @@
...
+    if (!taskNames.contains(taskName)) {
+      throw new SamzaException("writing checkpoint of unregistered task");
+    }
+
+    TableBatchOperation batchOperation = new TableBatchOperation();
+    checkpoint.getOffsets().forEach((ssp, offset) -> {
+        // Create table entity
+        TaskCheckpointEntity taskCheckpoint = new TaskCheckpointEntity(taskName.toString(),
+                serializeSystemStreamPartition(ssp));
+        taskCheckpoint.setOffset(offset);
```
vjagadish1989: Prefer to have this initialized during construction, as opposed to having a setter to mutate it later.

PR #342, opened 2017-10-26
SAMZA-1471: SystemConsumers should not poll ssp that hit end of stream

lhaiesp:
When SystemConsumers poll from SSPs that have hit end of stream, obviously there will be no data return and the poll will exhaust the timeout. This would cause performance issue.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/system/SystemConsumers.scala
+++ b/samza-core/src/main/scala/org/apache/samza/system/SystemConsumers.scala
@@ -28,6 +28,7 @@ import org.apache.samza.util.{Logging, TimerUtils}
 import org.apache.samza.system.chooser.MessageChooser
 import org.apache.samza.SamzaException
 import java.util.ArrayDeque
+import java.util.Collections
 import java.util.HashSet
 import java.util.HashMap
 import java.util.Queue
@@ -186,7 +187,7 @@ class SystemConsumers (
     if (IncomingMessageEnvelope.END_OF_STREAM_OFFSET.equals(offset)) {
       info("Stream : %s is already at end of stream" format (systemStreamPartition))
       endOfStreamSSPs.add(systemStreamPartition)
-      return;
+      return
     }
 
     metrics.registerSystemStreamPartition(systemStreamPartition)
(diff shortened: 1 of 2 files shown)


prateekm:
cc. @vjagadish1989 to see if this overlaps with our earlier discussion about stopping SSPIterator on EOS.

vjagadish1989:
@prateekm @lhaiesp @xinyuiscool 

Can we keep emptySSPBySystem and endOfStreamSSPs separate concepts? (instead of removing an end of stream SSP from the set of empty SSPs)

Atleast the semantic meaning of emptySSPs == the set of SSPs that we have no buffered messages for, and endOfStreamSSPs == the set of SSPs that are at EOS is maintained.

If the only intention is to prevent that SSP from being polled we can add a check during poll to filter out end of stream SSPs. In other words, an SSP is polled iff. both these conditions are true:
1. If we don't have buffered messages for it
2. If we have not reached end of stream

xinyuiscool:
@vjagadish1989 : right, the logic an be simplified by adding a simple check in poll().

@lhaiesp : can we make this change instead?

vjagadish1989 (approved):
Thanks Hai for the contribution! 

approved. looks great.

lhaiesp:
Local build and tests pass:
BUILD SUCCESSFUL

Total time: 10 mins 45.518 secs

samza-core/src/main/scala/org/apache/samza/system/SystemConsumers.scala
```diff
@@ -258,7 +259,14 @@ class SystemConsumers (
 
     trace("Getting fetch map for system: %s" format systemName)
 
-    val systemFetchSet = emptySystemStreamPartitionsBySystem.get(systemName)
```
vjagadish1989: Can we keep `emptySSPBySystem` and `endOfStreamSSPs` separate concepts? (instead of removing an end of stream SSP from the set of empty SSPs)

Atleast the semantic meaning of emptySSPs == the set of SSPs that we have no buffered messages for, and endOfStreamSSPs == the set of SSPs that are at EOS is maintained.

If the only intention is to prevent that SSP from being polled we can add a check during poll to filter out end of stream SSPs. In other words, an SSP is polled iff. both these conditions are true:

If we don't have buffered messages for it
If we have not reached end of stream

PR #343, opened 2017-10-26
SAMZA-1473 Fix handling of initial values for aggregating windows

vjagadish1989:
- Handle initial values for windows with foldLeftFunctions configured
- Improve trace level logging


vjagadish1989:
@prateekm for review. 

@jmakes @nickpan47 for FYI

PR #345, opened 2017-10-31
SAMZA-1477: Fix issues found by BEAM tests

xinyuiscool:
A bunch of issues were found by BEAM tests, which includes:

1) WatermarkFunction needs to be able to return output after processWatermark()
2) control message doesn't implement the equals() and hashcode()
3) Some kafka system related code is not scala 2.10 compatible for tests.

### DIFF
--- a/samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.scala
+++ b/samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.scala
@@ -217,7 +217,7 @@ private[kafka] class KafkaSystemConsumer(
               case _ => debug("Ignoring refresh for %s because we already added it from another thread." format head)
             }
           }
-          tpToRefresh.drop(1)
+          tpToRefresh.tail
         }
 
         while (!tpToRefresh.isEmpty) {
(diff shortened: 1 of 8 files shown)


samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.scala
```diff
@@ -193,8 +193,8 @@ private[kafka] class KafkaSystemConsumer(
 
         // addTopicPartition one at a time, leaving the to-be-done list intact in case of exceptions.
         // This avoids trying to re-add the same topic partition repeatedly
-        def refresh(tp: List[TopicAndPartition]) = {
-          val head :: rest = tpToRefresh
+        def refresh() = {
```
prateekm: Why is this change required?

xinyuiscool: It doesn't compile with scala 2.10 in unit tests. Specifically, the scalaTest version we are using ("3.0.1") has some compatibility issues with 2.10.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.scala
```diff
@@ -217,11 +217,11 @@ private[kafka] class KafkaSystemConsumer(
               case _ => debug("Ignoring refresh for %s because we already added it from another thread." format head)
             }
           }
-          rest
+          tpToRefresh.drop(1)
```
prateekm: Minor: Slightly cleaner to do tpToRefresh.tail. Is that equivalent here?

xinyuiscool: sure, that works too. Let me change to it.

PR #346, opened 2017-10-31
Admin GetPartitionsCount should return count and not fake SamzaPartitionsMetadata

sborya:


samza-api/src/main/java/org/apache/samza/system/ExtendedSystemAdmin.java
```diff
@@ -26,7 +26,7 @@
  * TODO: Merge this interface method with SystemAdmin when we upgrade to JDK 1.8
  */
 public interface ExtendedSystemAdmin extends SystemAdmin {
-  Map<String, SystemStreamMetadata> getSystemStreamPartitionCounts(Set<String> streamNames, long cacheTTL);
+  Map<String, Integer> getSystemStreamPartitionCounts(Set<String> streamNames, long cacheTTL);
```
nickpan47: Are we planning to use this in 0.14 or a minor version? This is a breaking change in API and if we are breaking it, I would rather remove this interface and merge this method in the standard SystemAdmin interface.

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -80,7 +80,7 @@ class StreamMetadataCache (
           val systemAdmin = systemAdmins
             .getOrElse(systemName, throw new SamzaException("Cannot get metadata for unknown system: %s" format systemName))
           val streamToMetadata = if (partitionsMetadataOnly && systemAdmin.isInstanceOf[ExtendedSystemAdmin]) {
-            systemAdmin.asInstanceOf[ExtendedSystemAdmin].getSystemStreamPartitionCounts(systemStreams.map(_.getStream).asJava, cacheTTLms)
+            systemAdmin.asInstanceOf[ExtendedSystemAdmin].getSystemStreamMetadata(systemStreams.map(_.getStream).asJava)
```
jmakes: This makes the `if` part of the conditional identical to the `else`. 

I think the original intent was for the `if` to use the faster "getSystemStreamPartitionCounts" method. So this will introduce a performance regression for any code paths that invoke this method with partitionsMetadataOnly==true.  We can't do that. 

I'm in favor of removing the partitionsMetadataOnly argument altogether and changing the callers that used that flag to use the "getSystemStreamPartitionCounts" method directly.

PR #347, opened 2017-11-01
SAMZA-1478; Delete unneeded data from intermediate Kafka topic on offset commit

lindong28:


lindong28:
@nickpan47 @prateekm do you have time to review this patch?

lindong28:
@xinyuiscool @jmakes Do you have time to review this patch?

gradle/dependency-versions.gradle
```diff
@@ -30,7 +30,7 @@
   jodaTimeVersion = "2.2"
   joptSimpleVersion = "3.2"
   junitVersion = "4.8.1"
-  kafkaVersion = "0.10.1.1"
+  kafkaVersion = "0.11.0.1"
```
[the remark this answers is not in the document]
lindong28: I confirmed with Becket that 0.11.* client and work broker version 0.10.* and above. Therefore this change of Kafka version from 0.10.1.1 to 0.11.0.1 should be backward compatible as far as the protocol between client and broker is concerned.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -22,12 +22,24 @@
 import java.util.Map;
 import java.util.Set;
 
+
```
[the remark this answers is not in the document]
lindong28: Sure. Fixed now.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -22,12 +22,24 @@
...
  * Helper interface attached to an underlying system to fetch information about
  * streams, partitions, offsets, etc. This interface is useful for providing
  * utility methods that Samza needs in order to interact with a system.
  */
 public interface SystemAdmin {
+
+  /**
+   * Connect to the underlying system and prepare to begin serving requests
+   */
+  default void start() {};
```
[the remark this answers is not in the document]
lindong28: Yes. It currently applies to Kafka. I added these two methods in SystemAdmin interface because it seems to make sense to have start() and stop() in general, similar to SystemProducer() and SystemConsumer(). If this sounds reasonable in the long term, and there is no short term concern since we provided an empty default method for existing users, maybe we could just add these two methods?

If we want to be conservative and prefer not to make change to the interface, I will move them to KafkaSystemAdmin. What do you think?

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -84,12 +96,24 @@ default void validateStream(StreamSpec streamSpec) throws StreamValidationExcept
...
+
+  /**
+   * Clear each partition up to (and include) the corresponding offset for all system stream partitions specified in the map
+   *
+   * @param offsets a map from system stream partition to offset
+   * @return {@code true} if the stream was successfully cleared.
+   *         {@code false} if clearing stream failed.
+   */
+  default boolean clearStream(Map<SystemStreamPartition, String> offsets) {
+    return false;
```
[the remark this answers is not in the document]
lindong28: Good point. We currently only use Kafka for intermediate stream, it should be OK to throw Exception here. I will update the code to do this.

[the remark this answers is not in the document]
lindong28: I also renamed the name of the newly added method from `clearStream` to `deleteRecords` so that it is more explicit what it does and user can differentiate it from the existing clearStream(), which delete the topic rather than records.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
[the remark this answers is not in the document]
prateekm: I agree with @lindong28, should be ok to add as a default no-op method for SystemAdmin since it makes sense in the context. 

I think EventHubSystemAdmin could benefit from these lifecycle methods too. @vjagadish1989 can verify.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -84,12 +95,22 @@ default void validateStream(StreamSpec streamSpec) throws StreamValidationExcept
...
    * @param streamSpec  The spec for the physical stream on the system.
    * @return {@code true} if the stream was successfully cleared.
    *         {@code false} if clearing stream failed.
    */
   default boolean clearStream(StreamSpec streamSpec) {
     throw new UnsupportedOperationException();
   }
+
+  /**
+   * Delete records up to (and include) the corresponding offset for all system stream partitions specified in the map
```
prateekm: I think 'trim' is pretty descriptive of what this is doing, so wouldn't mind calling it that.

Else, 'records' is a Kafka specific term. Samza refers to these as simply messages, so maybe deleteMessages? Or clearUpTo(Map<SSP, String> offset)?

Minor: "Delete records up to (and **including**) the **provided ssp offsets**."

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -900,7 +900,7 @@ class SamzaContainer(
 
   def startTask {
     info("Initializing stream tasks.")
-
+    offsetManager.systemAdmins.values.foreach(_.start())
```
prateekm: This is wrong, shouldn't be starting SystemAdmins by looking into the offsetManager field. This should be tied either to the lifecycle of the entity that created this, i.e. the SamzaContainer or the entity that owns it, i.e. the offsetManager.

Admins should also be started and stopped everywhere they're being created/used in the framework and not just for OffsetManager. If that's too much work, I'd actually agree with Yi and not make a generic start/stop API.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -203,6 +204,18 @@ class TaskInstance(
     trace("Checkpointing offsets for taskName: %s" format taskName)
 
     offsetManager.writeCheckpoint(taskName, checkpoint)
+
+    if (checkpoint != null) {
+      checkpoint.getOffsets.asScala.filter { case (ssp, _) =>
```
prateekm: Minor: formatting:

```
checkpoint.getOffsets.asScala
    .filter { ... }
    .grouBy { ... }
    .foreach { ... }
```

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -542,6 +562,20 @@ class KafkaSystemAdmin(
...
   /**
+    * @inheritdoc
+    *
+    * Delete records of partitions in Kafka up to (and include) the corresponding offsets. This only works with Kafka cluster 0.11 or later.
+    * Otherwise it's a no-op.
+    */
+  override def deleteRecords(offsets: util.Map[SystemStreamPartition, String]) {
+    if (adminClient == null)
+      throw new SamzaException("KafkaSystemAdmin has not started yet")
+    adminClient.deleteRecordsBefore(offsets.asScala.toSeq.map { case (systemStreamPartition, offset) =>
```
prateekm: Prefer extracting variable for the argument to deleteRecordsBefore.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -542,6 +562,20 @@ class KafkaSystemAdmin(
   }
 
   /**
+    * @inheritdoc
+    *
+    * Delete records of partitions in Kafka up to (and include) the corresponding offsets. This only works with Kafka cluster 0.11 or later.
```
prateekm: Minor: "up to (and **including**) the **provided** offsets. **newline** This only "

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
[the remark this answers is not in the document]
vjagadish1989: I'd be +1 to @lindong28's suggestion as well. For instance, the `EventHubSystemAdmin` would benefit from doing a one-time initialization of its `EventHubClient` during `init`.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
vjagadish1989: +1 to @nickpan47 and @prateekm 's suggestions.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
lindong28: Thanks @prateekm for your review! I have updated the code and the comment as suggested.

I think a few words are very similar, e.g. clear, delete, trim, purge. I don't have a strong preference among them as long as we use the name consistently. I used the name `deleteMessages` for now. I will ask you in person later regarding the name.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
lindong28: Thank you all for the review. I have update the code to address this issue.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
lindong28: Sure. I have updated the code as suggested.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
lindong28: Sure. It is fixed now.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
lindong28: Sure. It is fixed now.

lindong28:
@prateekm @nickpan47 @vjagadish1989 Thanks for your review! I have addressed all comments and fixed a flaky test. Can you review the patch again? Do you think we can include this patch in the next Apache Samza release?

lindong28:
Jenkins keeps failing with "Unknown license: /home/jenkins/jenkins-slave/workspace/samza-pr-checks/samza-test/hs_err_pid5042.log". Does anyone know how to fix it?

vjagadish1989:
@lindong28 : Is that file checked in to the remote repo? Or is hs_err_pid5042 a local file? If it's the latter, can you delete it? It's probably related to a test JVM failure?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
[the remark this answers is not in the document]
lindong28: Since the states and the long-lasting client connections in SystemAdmin are only created when we call SystemAdmin.start(), then it should be OK as of this patch to only call SystemAdmin.start() when we start SamzaContainer, right?

We can have a separate patch to go over all existing usage of SystemAdmin to manage them and call SystemAdmin.start() properly. By splitting the work into two patches, we can reduce the amount of rebase needed for this patch. Does this sound OK?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -148,6 +149,25 @@ class KafkaSystemAdmin(
 
   import KafkaSystemAdmin._
 
+  @volatile var adminClient: AdminClient = null
```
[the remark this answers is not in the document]
lindong28: AdminClient can also be used to create topic, delete topic, list all topics, get topic metadata, etc.

This is volatile because SystemAdmin.deleteRecords() and SystemAdmin.start() may potentially be called by different threads. Does this make sense?

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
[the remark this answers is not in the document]
nickpan47: @prateekm @vjagadish1989 my point is: the current usage pattern of creating the SystemAdmin instances is not quite following the life-cycle management of a stateful component. I did a quick grep in the code base for the calls to SystemFactory.getAdmin() calls, and there are many places:
yipan-mn1:samza_master yipan$ git grep "getAdmin" | grep -vi "test" | cut -d':' -f1 | sort -u
samza-api/src/main/java/org/apache/samza/system/SystemFactory.java
samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubSystemFactory.java
samza-core/src/main/java/org/apache/samza/config/JavaSystemConfig.java
samza-core/src/main/java/org/apache/samza/standalone/PassthroughJobCoordinator.java
samza-core/src/main/java/org/apache/samza/storage/StorageRecovery.java
samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
samza-core/src/main/scala/org/apache/samza/coordinator/stream/CoordinatorStreamSystemFactory.scala
samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
samza-core/src/main/scala/org/apache/samza/system/filereader/FileReaderSystemFactory.scala
samza-elasticsearch/src/main/java/org/apache/samza/system/elasticsearch/ElasticsearchSystemFactory.java
samza-hdfs/src/main/scala/org/apache/samza/system/hdfs/HdfsSystemFactory.scala
samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManagerFactory.scala
samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala

For now, each invocation of SystemFactory.getAdmin() is creating a new instance of admin. If we are going to add the life-cycle methods to SystemAdmin objects, we either need to add life-cycle management code for each instance of SystemAdmin object created, or need to manage a single set of shared SystemAdmin objects (one per system) and control the life-cycle methods in a central place.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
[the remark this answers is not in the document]
nickpan47: Thanks! Makes sense.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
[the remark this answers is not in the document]
nickpan47: @lindong28 I have concerns on this fast path due to the following reason: ideally, if we want to introduce a concept of life-cycle management on SystemAdmin, we should already clearly define what operations are allowed in each different stage in the life-cycle of the SystemAdmin. For example, no admin operations should be allowed before the SystemAdmin is started. W/ the fast path you propose here, we don't have a clear view of which operation is allowed in which stage any more (i.e. there are some admin operations you can do before start and some admin operations that you can't).

lindong28:
@nickpan47 Do you have time to review this patch?

lindong28:
@vjagadish1989 Here is the PR. Thanks in advance for your review!

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -226,10 +233,12 @@ public void runApplication(StreamApplication streamApplication, String appName,
     configs.put("systems.kafka.samza.key.serde", "string");
     configs.put("systems.kafka.samza.msg.serde", "string");
     configs.put("systems.kafka.samza.offset.default", "oldest");
+    configs.put("systems.kafka.samza.delete.messages.enabled", "true");
     configs.put("job.coordinator.system", "kafka");
     configs.put("job.default.system", "kafka");
     configs.put("job.coordinator.replication.factor", "1");
     configs.put("task.window.ms", "1000");
+    configs.put("task.checkpoint.factory", TestStreamManager.MockCheckpointManagerFactory.class.getName());
```
vjagadish1989: why is this needed?

samza-test/src/test/java/org/apache/samza/test/operator/TestRepartitionJoinWindowApp.java
```diff
@@ -76,6 +84,27 @@ public void testRepartitionJoinWindowApp() throws Exception {
       Assert.assertTrue(key.equals("u1") || key.equals("u2"));
       Assert.assertEquals("2", value);
     }
+
+    // Verify that messages in the intermediate stream will be deleted in 10 seconds
```
vjagadish1989: I would also add a negative test to ensure that if `deleteMessages` is set to false, we are able to consume messages from the intermediate stream.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -554,14 +573,31 @@ class KafkaSystemAdmin(
...
+    */
+  override def deleteMessages(offsets: util.Map[SystemStreamPartition, String]) {
+    if (!running)
+      throw new SamzaException("KafkaSystemAdmin has not started yet")
+    val nextOffsets = offsets.asScala.toSeq.map { case (systemStreamPartition, offset) =>
+      (new TopicPartition(systemStreamPartition.getStream, systemStreamPartition.getPartition.getPartitionId), offset.toLong + 1)
+    }.toMap
+
+    if (deleteMessagesEnabled)
+      adminClient.deleteRecordsBefore(nextOffsets)
```
vjagadish1989: Is there any perf-overhead with this method since we do it at each commit? If so, would it be possible to invoke this only if `deleteRecordsBefore` supported on the broker? You can make this determination once during startup.

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -320,7 +320,7 @@ private boolean markProcessed() {
       this.task = task;
       this.callbackManager = new TaskCallbackManager(this, callbackTimer, callbackTimeoutMs, maxConcurrency, clock);
       Set<SystemStreamPartition> sspSet = getWorkingSSPSet(task);
-      this.state = new AsyncTaskState(task.taskName(), task.metrics(), sspSet, task.hasIntermediateStreams());
```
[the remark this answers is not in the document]
lindong28: Alternatively I can add back the method `hasIntermediateStreams` in `TaskInstance`.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -554,14 +573,31 @@ class KafkaSystemAdmin(
...
   }
 
+  /**
+    * @inheritdoc
+    *
+    * Delete records up to (and including) the provided ssp offsets for all system stream partitions specified in the map
+    * This only works with Kafka cluster 0.11 or later. Otherwise it's a no-op.
+    */
+  override def deleteMessages(offsets: util.Map[SystemStreamPartition, String]) {
+    if (!running)
```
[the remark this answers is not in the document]
lindong28: Thanks. I will add this. I will also follow this code style in the future Samza patch.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -554,14 +573,31 @@ class KafkaSystemAdmin(
...
+      throw new SamzaException("KafkaSystemAdmin has not started yet")
+    val nextOffsets = offsets.asScala.toSeq.map { case (systemStreamPartition, offset) =>
+      (new TopicPartition(systemStreamPartition.getStream, systemStreamPartition.getPartition.getPartitionId), offset.toLong + 1)
+    }.toMap
+
```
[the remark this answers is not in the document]
lindong28: You are right. Thanks for the suggestion. I will make this change.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -554,14 +573,31 @@ class KafkaSystemAdmin(
...
 
+  /**
+    * @inheritdoc
+    *
+    * Delete records up to (and including) the provided ssp offsets for all system stream partitions specified in the map
+    * This only works with Kafka cluster 0.11 or later. Otherwise it's a no-op.
+    */
+  override def deleteMessages(offsets: util.Map[SystemStreamPartition, String]) {
+    if (!running)
+      throw new SamzaException("KafkaSystemAdmin has not started yet")
```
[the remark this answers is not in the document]
lindong28: Sure. Does this look OK?

```
throw new SamzaException(s"KafkaSystemAdmin has not started yet. Fail to delete messages for systems ${offsets.keySet().asScala.map(_.getSystem).toSet}")
```

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
lindong28: This is because if we do not define checkpoint factory and if we do not have checkpoint listener, `OffsetManager.buildCheckpoint()` will return null. And if we do not have checkpoint, we will not delete messages in `TaskInstance.commit()`.

samza-test/src/test/java/org/apache/samza/test/operator/TestRepartitionJoinWindowApp.java
lindong28: Sure. I will add the negative test.

lindong28: I added a negative test but it is not easy to run it together with the positive test. These two tests interactive with each other because they use the same Kafka topics.

I can refactor `TestRepartitionJoinWindowApp` and `RepartitionJoinWindowApp` so that it uses different topics for different runs. Not sure if it is needed. What do you think?

lindong28:
@vjagadish1989 Thanks much for taking time to review the patch! I have updated the patch and addressed most of your comments.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -203,6 +204,15 @@ class TaskInstance(
     trace("Checkpointing offsets for taskName: %s" format taskName)
 
     offsetManager.writeCheckpoint(taskName, checkpoint)
+
+    if (checkpoint != null) {
+      checkpoint.getOffsets.asScala
+        .filter { case (ssp, _) => intermediateStreams.contains(ssp.getStream) } // Only delete data of intermediate streams
+        .groupBy { case (ssp, _) => ssp.getSystem }
+        .foreach { case (systemName: String, offsets: Map[SystemStreamPartition, String]) =>
+          systemAdmins.getSystemAdmin(systemName).deleteMessages(offsets.asJava)
```
jmakes: IIUC, because of the default implementation, this will throw for all systems except Kafka. That doesn't seem correct.

lindong28: Good catch @jmakes. I have updated the patch to fix it by not throwing exception in the default implementation.

vjagadish1989: Great find @jmakes! This was missed in my earlier review!

samza-test/src/test/scala/org/apache/samza/test/harness/AbstractIntegrationTestHarness.scala
```diff
@@ -17,10 +17,13 @@
  * under the License.
  */
 package org.apache.samza.test.harness
-import java.util.Properties
+import java.util.{Properties, UUID}
```
nickpan47: nit: why adding UUID here? I don't see the usage of UUID in this file.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -149,6 +151,17 @@ public ApplicationStatus status(StreamApplication app) {
     }
   }
 
+  // Used only by test
```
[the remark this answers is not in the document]
lindong28: Thanks. Fixed now.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -149,6 +151,17 @@ public ApplicationStatus status(StreamApplication app) {
...
+  public List<StreamSpec> getIntermediateStreams(StreamApplication app) {
+    String runId = String.valueOf(System.currentTimeMillis()) + "-" + UUID.randomUUID().toString().substring(0, 8);
```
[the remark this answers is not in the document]
lindong28: Very good point. I have fixed it as suggested.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -67,7 +68,7 @@ class TaskInstance(
     scala.collection.mutable.Map[SystemStreamPartition, Boolean]()
   systemStreamPartitions.foreach(ssp2CaughtupMapping += _ -> false)
 
-  val hasIntermediateStreams = config.getStreamIds.exists(config.getIsIntermediate(_))
+  val intermediateStreams: Set[String] = config.getStreamIds.filter(config.getIsIntermediate).toSet
```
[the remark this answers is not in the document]
lindong28: Sure. I have renamed `config.getIsIntermediate` to `config.isIntermediateStream`

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -98,8 +100,9 @@
 public class StreamApplicationIntegrationTestHarness extends AbstractIntegrationTestHarness {
   private KafkaProducer producer;
   private KafkaConsumer consumer;
+  protected KafkaSystemAdmin systemAdmin;
   private StreamApplication app;
-  private ApplicationRunner runner;
+  protected RemoteApplicationRunner runner;
```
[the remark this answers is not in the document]
lindong28: I am using `RemoteApplicationRunner` here because I need to call `RemoteApplicationRunner.getIntermediateStreams()`. Due to the existing limit in Samza test suite, we have to use `RemoteApplicationRunner` instead of `LocalApplicationRunner`. Prateek left comment in `StreamApplicationIntegrationTestHarness` that says "We can't use LocalApplicationRunner since ZkJobCoordinator doesn't currently create". I think Prateek can explain more deeply why we have to use `RemoteApplicationRunner`.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -247,7 +254,7 @@ public void runApplication(StreamApplication streamApplication, String appName,
     }
 
     app = streamApplication;
-    runner = ApplicationRunner.fromConfig(new MapConfig(configs));
+    runner = (RemoteApplicationRunner) ApplicationRunner.fromConfig(new MapConfig(configs));
```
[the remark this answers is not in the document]
lindong28: Due to the existing limit in Samza test suite, we have to use `RemoteApplicationRunner` instead of `LocalApplicationRunner`. Prateek left comment in `StreamApplicationIntegrationTestHarness` that says "We can't use LocalApplicationRunner since ZkJobCoordinator doesn't currently create". I think Prateek can explain more deeply why we have to use `RemoteApplicationRunner`.

samza-test/src/test/scala/org/apache/samza/test/harness/AbstractIntegrationTestHarness.scala
lindong28: Thanks for the detailed review. I have updated the patch to remove the UUID.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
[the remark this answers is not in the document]
nickpan47: HI, @lindong28 thanks for the response. I should have made my point more explicit. The comment here is not whether the current integration test is actually using which ApplicationRunner class via configuration. The request here is not limiting the ApplicationRunner implementation class that could be used in this test harness.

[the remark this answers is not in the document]
lindong28: Thanks @nickpan47. I have updated the patch to move getIntermediateStreams() into AbstractApplicationRunner. Thus all AbstractApplicationRunner can be used in the test harness.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -554,14 +575,34 @@ class KafkaSystemAdmin(
...
+    deleteMessagesCalled = true
+
+    if (!running) {
+      throw new SamzaException(s"KafkaSystemAdmin has not started yet for system $systemName")
+    }
+    if (deleteMessagesEnabled) {
+      val nextOffsets = offsets.asScala.toSeq.map { case (systemStreamPartition, offset) =>
+        (new TopicPartition(systemStreamPartition.getStream, systemStreamPartition.getPartition.getPartitionId), offset.toLong + 1)
+      }.toMap
+      adminClient.deleteRecordsBefore(nextOffsets)
```
nickpan47: Just out-of-curiosity, what if the adminClient failed during the SystemAdmin's life-cycle? Do we throw-away the failed adminClient and recreate a new one? If not, how do we recover? Or are we saying that the exception from deleteMessages() would definitely fail the SamzaContainer? I would prefer that this deleteMessages() can be a best-effort operation. i.e. if there is a network connection issue that does not allow deleteMessages() now, in next round of commit, when the network connection issue is gone, we should be able to deleteMessages() right?

lindong28: Good point. AdminClient has a background thread that keeps trying to send record and reconnect to broker if needed etc. If AdminClient is disconnected from broker, adminClient.deleteRecordsBefore() will still enqueue request into a request queue and return a future. If AdminClient connects to broker before timeout, then request will be sent and executed. Otherwise, a TimeoutException will be returned if user calls future.get(). Samza would not notice any failure since we do not call get() on the future returned by adminClient.deleteRecordsBefore().

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -38,6 +39,8 @@ import scala.collection.JavaConverters._
 
 object KafkaSystemAdmin extends Logging {
 
+  // Used only for test
+  @volatile var deleteMessagesCalled = false
   val CLEAR_STREAM_RETRIES = 3
 
   /**
```
nickpan47: Just one question: I thought that the introduction of AdminClient would replace our need to instantiate any SimpleConsumer instance, or using KafkaUtils to get topic metadata and fetch offsets? It could be addressed in a separate PR, but I just want to make sure that's the plan and there is no blockers to do that.

samza-test/src/test/java/org/apache/samza/test/operator/StreamApplicationIntegrationTestHarness.java
```diff
@@ -262,6 +269,9 @@ public void setNumEmptyPolls(int numEmptyPolls) {
    */
   @Override
   public void tearDown() {
+    systemAdmin.stop();
```
nickpan47: +1. Like that!

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
lindong28: Thanks for all the review @nickpan47! Yes, AdminClient can be used to get topic metadata. But we would need to use KafkaConsumer.offsetsForTimes to fetch offset for e.g. earliest offset or latest offset.

lindong28:
@nickpan47 @jmakes Can we merge this patch for the next release?

nickpan47:
@jmakes and @prateekm , Dong has finished the revision and this patch is ready to be committed. Are you guys still -1 on this patch? Could you quickly check and respond? Thanks!

samza-test/src/test/java/org/apache/samza/test/operator/RepartitionJoinWindowApp2.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.samza.test.operator.data.AdClick;
+import org.apache.samza.test.operator.data.PageView;
+import org.apache.samza.test.operator.data.UserPageAdClick;
+
+import java.time.Duration;
+
+/**
+ * A {@link StreamApplication} that demonstrates a partitionBy, stream-stream join and a windowed count.
+ */
+public class RepartitionJoinWindowApp2 implements StreamApplication {
```
prateekm: Why copy this? If this is meant to be a separate test maybe rename it?

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -171,4 +174,15 @@ final void writePlanJsonFile(String planJson) {
       log.warn("Failed to write execution plan json to file", e);
     }
   }
+
+  @VisibleForTesting
+  public List<StreamSpec> getIntermediateStreams(StreamApplication app) {
```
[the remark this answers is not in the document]
lindong28: Yeah it exists only for testing. Not sure how that can be moved to test class though. I am happy to make the change if it is doable.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -171,4 +174,15 @@ final void writePlanJsonFile(String planJson) {
...
+    try {
+      ExecutionPlan plan = getExecutionPlan(app);
+      return plan.getIntermediateStreams();
+    } catch (Throwable t) {
+      throw new SamzaException("Failed to run application", t);
```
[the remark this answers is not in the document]
lindong28: Thanks. I have updated the patch to fix the comment.

samza-core/src/main/scala/org/apache/samza/config/SystemConfig.scala
```diff
@@ -30,6 +30,7 @@ object SystemConfig {
   val SYSTEM_PREFIX = JavaSystemConfig.SYSTEM_PREFIX + "%s."
   val SYSTEM_FACTORY = JavaSystemConfig.SYSTEM_FACTORY_FORMAT
   val CONSUMER_OFFSET_DEFAULT = SYSTEM_PREFIX + "samza.offset.default"
+  val DELETE_MESSAGES_ENABLED = SYSTEM_PREFIX + "samza.delete.messages.enabled"
```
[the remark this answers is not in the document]
lindong28: Sure. I have renamed it as suggested. Let me check how to add it in the configuration reference.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -218,6 +219,15 @@ class TaskInstance(
     trace("Checkpointing offsets for taskName: %s" format taskName)
 
     offsetManager.writeCheckpoint(taskName, checkpoint)
+
+    if (checkpoint != null) {
+      checkpoint.getOffsets.asScala
+        .filter { case (ssp, _) => intermediateStreams.contains(ssp.getStream) } // Only delete data of intermediate streams
```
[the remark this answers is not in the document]
lindong28: We can delete committed messages from intermediate stream because, in case the job fails, the job can be restarted again and regenerate all the data in the intermediate stream. However, if we delete committed messages from input stream, then if the job fails, the restarted job may not be able to reproduce the result because data in the input streams is deleted.

And yet, we can add this feature in low level API. We would need users to explicitly mark their input streams as "intermediate". I guess the goal of doing this only in high level API is that our long term goal is for most users to use the high level API. Maybe we can implement this in low level API in a separate patch if needed?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -38,6 +39,8 @@ import scala.collection.JavaConverters._
 
 object KafkaSystemAdmin extends Logging {
 
+  // Used only for test
+  @volatile var deleteMessagesCalled = false
```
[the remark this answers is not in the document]
lindong28: Sure. Added now.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -144,20 +147,38 @@ class KafkaSystemAdmin(
...
+
+  /**
+   * Whether deleteMessages() API can be used
+   */
+  deleteMessagesEnabled: Boolean = false) extends ExtendedSystemAdmin with Logging {
 
   import KafkaSystemAdmin._
 
   @volatile var running = false
+  @volatile var adminClient: AdminClient = null
```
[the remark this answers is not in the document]
lindong28: The type needs to be explicitly specified because Scala can not derive the type of the variable if the value is null.

samza-test/src/test/java/org/apache/samza/test/operator/RepartitionJoinWindowApp2.java
lindong28: `RepartitionJoinWindowApp` is used by `testRepartitionJoinWindowAppAndDeleteMessagesOnCommit()` and `RepartitionJoinWindowApp2` is used by `testRepartitionJoinWindowAppWithoutDeletionOnCommit()`. `RepartitionJoinWindowApp2` is almost the same as `RepartitionJoinWindowApp` except for the topic names used in the app. If these two unit tests use the same app with the same Kafka topics, the second test will fail. I am not able to get around this problem. That is why I added `RepartitionJoinWindowApp2`.

prateekm: Possible to make stream names constructor params for that class in that case?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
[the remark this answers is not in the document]
prateekm: Makes sense, thanks.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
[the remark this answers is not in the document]
prateekm: Possible to expose getExecutionPlanner instead?

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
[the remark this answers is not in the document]
prateekm: As we discussed offline, let's decouple this feature from intermediate streams, since it can apply to all input streams. We can auto-write this config for intermediate streams from the Planner for high level API apps. @nickpan47 also agrees.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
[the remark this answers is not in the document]
lindong28: @prateekm Yeah good point. I have updated the patch as suggested.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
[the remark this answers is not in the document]
lindong28: @prateekm Sure. I have updated the patch that allows user to configure whether a stream's committed messages can be deleted.

samza-test/src/test/java/org/apache/samza/test/operator/RepartitionJoinWindowApp2.java
lindong28: @prateekm Sure. I have updated the patch as suggested.

nickpan47:
@prateekm and @lindong28 what else needs to be addressed here for this one to be merged? Thanks!

lindong28:
@nickpan47 Thanks! I have updated the code to (hopefully) address all comments by @prateekm. However, I am not able to debug the test failure properly in China - it looks like the `./gradlew test` will take super long time to finish when I am in China. I guess it is because the test will try to access something online in US and the access is slow from China. I don't know how to fix this problem. So most likely I will fix the test after I get back US.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
```diff
@@ -37,6 +37,7 @@ object StreamConfig {
   val BOOTSTRAP =               SAMZA_PROPERTY + "bootstrap"
   val PRIORITY =                SAMZA_PROPERTY + "priority"
   val IS_INTERMEDIATE =         SAMZA_PROPERTY + "intermediate"
+  val IS_COMMITTED_MESSAGES_DELETABLE = SAMZA_PROPERTY + "committed.messages.deletable"
```
prateekm: delete.committed.messages?

samza-core/src/main/scala/org/apache/samza/config/SystemConfig.scala
```diff
@@ -31,6 +31,11 @@ object SystemConfig {
   val SYSTEM_FACTORY = JavaSystemConfig.SYSTEM_FACTORY_FORMAT
   val CONSUMER_OFFSET_DEFAULT = SYSTEM_PREFIX + "samza.offset.default"
 
+  // If true, automatically delete committed messages from streams whose committed messages are deletable.
```
prateekm: Should be included in the configuration reference, and in javadoc in the config class.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -218,6 +221,15 @@ class TaskInstance(
     trace("Checkpointing offsets for taskName: %s" format taskName)
 
     offsetManager.writeCheckpoint(taskName, checkpoint)
+
+    if (checkpoint != null) {
+      checkpoint.getOffsets.asScala
+        .filter { case (ssp, _) => committedMessagesDeletableStreams.contains(ssp.getStream) } // Only delete data of intermediate streams
```
prateekm: Minor: Faster to iterate over the streams with committedMessagesDeletable and look up in checkpointedOffsets.

samza-test/src/test/java/org/apache/samza/test/operator/BroadcastAssertApp.java
```diff
@@ -29,15 +29,18 @@
 
 import java.util.Arrays;
 
-import static org.apache.samza.test.operator.RepartitionJoinWindowApp.PAGE_VIEWS;
-
 public class BroadcastAssertApp implements StreamApplication {
 
+  public static final String INPUT_TOPIC_PROP = "inputTopic";
```
prateekm: INPUT_TOPIC_NAME

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemAdmin.scala
```diff
@@ -59,32 +63,34 @@ object TestKafkaSystemAdmin extends KafkaServerTestHarness {
...
-  def generateConfigs() = {
+  override def generateConfigs(): Seq[KafkaConfig] = {
     val props = TestUtils.createBrokerConfigs(numBrokers, zkConnect, true)
     props.map(KafkaConfig.fromProps)
   }
 
-  @BeforeClass
-  override def setUp {
-    super.setUp
+  @Before
```
prateekm: Does this significantly increase test times?

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala
lindong28: Sure. Renamed as suggested.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
lindong28: I am not sure how to achieve this given that `checkpoint.getOffsets` is keyed by SystemStreamPartition. Could you provide example code? Thanks.

samza-test/src/test/java/org/apache/samza/test/operator/BroadcastAssertApp.java
lindong28: Done

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemAdmin.scala
lindong28: The purpose of this change is not to make test faster. It is to make each test independent of each other. If we use BeforeClass and AfterClass tags, then these methods need to be static and the variables instantiated (i.e. producer, systemAdmin) needs to be static as well. These variables will then be instantiated exactly once for the test suite and used by all unit tests in this test suite. Thus kind allows interference between tests and make make test development difficult.

During development of this patch the interference must have caused some problem which required me to make this change. I don't exactly remember what that problem is now. But in general it seems like good practice to make unit test independent of each other. What do you think?

samza-core/src/main/scala/org/apache/samza/config/SystemConfig.scala
lindong28: Sure. I have updated `configuration-table.html` to include the two newly-added configs. Not sure how to include this in the java doc of the config class. Is there example config whose usage is included in the javadoc in the config class?

lindong28:
@nickpan47 @prateekm I am able to debug and fix the test failure after using virtual private network. The patch has been updated to address the latest comments. Can you take another look? Thanks for the review!

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemAdmin.scala
prateekm: These tests are more integration tests than unit tests since they bring up ZK and Kafka. This class has at least 8 tests, so wondering how much of a time difference starting ZK + Kafka 8x vs 1x makes. If it adds say ~ a minute to build times, I'd try to find a workaround.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
prateekm: Never mind, types weren't clear from this snippet.

That reminds me, streamId might not be the same as the system stream name. I think what you need here is the physical name corresponding to the stream id. Can you check if this does the right thing if streamId != physical name?

lindong28: @prateekm @jmakes Thanks. I have updated the patch as suggested.

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemAdmin.scala
lindong28: @prateekm Thanks. I have updated the patch to fix the issue.

PR #348, opened 2017-11-01
SAMZA-1479 : Refactor KafkaCheckpointManager, KafkaCheckpointLogKey and their tests

vjagadish1989:
Notable changes:
* Rewrite `KafkaCheckpointLogKey` into two classes - an immutable class, and a SerDe
* Remove dependency on static setters in the `KafkaCheckpointLogKey`
* Change lifecycle of components in KafkaCheckpointManager 
  - It's safe to start producers and consumers during `start` as opposed to lazy loading them during writes, and reads.
  - Initialize systemProducer and systemConsumer during construction
* Simplify logic for ignoring checkpoint validations
* Re-write checkpointManager#readLog() to use a simpler API.  
* Remove unnecessary complexity after the migration from 0.8
* Remove unnecessary locking in startup, and shut-down
* Remove dependencies on SimpleConsumer configs like bufferSize, fetchSize, socketTimeout
* Refactor KafkaCheckpointManagerFactory and remove static getCheckpointSystemNameAndFactory
* Bug-fix : Register the taskName correctly (instead of using a dummy string for the taskName)

* Add unit tests to verify more checkpoint scenarios
* Consolidate unit tests into utils for creating producer, consumer and admin instances
* Convert/consolidate most long-running integration tests into unit tests


vjagadish1989:
Most of changes are in:
samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala

samza-kafka/src/test/scala/org/apache/samza/checkpoint/kafka/TestKafkaCheckpointManager.scala
but github collapsed the diff.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
+      }
 
-      if (!shouldHandleEntry(checkpointKey)) {
-        info("Skipping checkpoint log entry at offset %s with key %s." format(offset, checkpointKey))
-      } else {
-        val checkpointPayload = ByteBuffer.wrap(msg.getMessage.asInstanceOf[Array[Byte]])
-        handleEntry(checkpointPayload, checkpointKey)
+      val checkpointKey = checkpointKeySerde.fromBytes(keyBytes)
+      val expectedGrouperFactory = checkpointKey.getGrouperFactoryClassName
+      if (!grouperFactory.equals(expectedGrouperFactory)) {
```
prateekm: It'll be good to have this validation be behind the "validateCheckpoint" flag as well. Currently there's no good way to change the grouper for the job.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
+    val partition: Partition = ssp.getPartition
 
-      if (systemConsumer != null) {
-        systemConsumer.stop
-        systemConsumer = null
-      }
+    val metaDataMap = systemAdmin.getSystemStreamMetadata(Collections.singleton(topic))
+    val checkpointMetadata: SystemStreamMetadata = metaDataMap.get(topic)
+    if (checkpointMetadata == null) {
+      throw new SamzaException("Cannot get metadata for system=%s, topic=%s" format(checkpointSystem, topic))
```
prateekm: Nitpick: 
Not your change but:
s/Cannot get/Got null metadata. They have different implications.
Same for partitionMetaData below. Should also not use field name in message (e.g., "partition metadata instead of partitionMetaData")

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
-      if (systemProducer != null) {
-        systemProducer.stop
-        systemProducer = null
-      }
+  /**
+    * Returns the oldest available offset for the provided [[SystemStreamPartition]].
+    */
+  private def getOldestOffset(ssp: SystemStreamPartition): String = {
+    val topic: String = ssp.getSystemStream.getStream
+    val partition: Partition = ssp.getPartition
```
prateekm: Nitpick: Don't need type annotation

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
+  *
+  * <p> This class is thread safe for writing but not for reading checkpoints. This is currently OK since checkpoints
+  * are only read on the main thread.
+  */
+class KafkaCheckpointManager(checkpointSpec: KafkaStreamSpec,
+  systemFactory: SystemFactory,
+  failOnCheckpointValidation: Boolean,
+  config: Config,
+  metricsRegistry: MetricsRegistry,
+  checkpointMsgSerde: Serde[Checkpoint] = new CheckpointSerde) extends CheckpointManager with Logging {
```
prateekm: Maybe checkpointSerde?

prateekm: Just noticed: why is the msg serde here while the key serde is hardcoded?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
   var taskNamesToOffsets: Map[TaskName, Checkpoint] = null
-  val systemAdmin = getSystemAdmin()
-
-  val kafkaUtil: KafkaUtil = new KafkaUtil(retryBackoff, connectZk)
-
-
-  KafkaCheckpointLogKey.setSystemStreamPartitionGrouperFactoryString(systemStreamPartitionGrouperFactoryString)
 
-  info("Creating KafkaCheckpointManager with: clientId=%s, checkpointTopic=%s, systemName=%s" format(clientId, checkpointTopic, systemName))
+  info("Creating KafkaCheckpointManager for checkpointTopic=%s, systemName=%s" format(checkpointTopic, checkpointSystem))
```
prateekm: Nitpick: "Created" or move as first line before any field declarations.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
+  *
+  * <p> The Kafka partition provides an abstraction of a log to which all [[Checkpoint]]s are appended to. The
+  * checkpoints written to the log are keyed by their corresponding taskName.
+  *
+  * <p> This class is thread safe for writing but not for reading checkpoints. This is currently OK since checkpoints
+  * are only read on the main thread.
+  */
+class KafkaCheckpointManager(checkpointSpec: KafkaStreamSpec,
+  systemFactory: SystemFactory,
+  failOnCheckpointValidation: Boolean,
```
prateekm: This is not a great field name. Maybe rename everything upstream up until the config to "validateCheckpoints" or "failOnInvalidCheckpoints"?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
-   * @param taskName Specific Samza taskName of which to write a checkpoint of.
-   * @param checkpoint Reference to a Checkpoint object to store offset data in.
-   **/
-  override def writeCheckpoint(taskName: TaskName, checkpoint: Checkpoint) {
-    val key = KafkaCheckpointLogKey.getCheckpointKey(taskName)
-    val keyBytes = key.toBytes()
-    val msgBytes = serde.toBytes(checkpoint)
-    val systemStream = new SystemStream(systemName, checkpointTopic)
-    val envelope = new OutgoingMessageEnvelope(systemStream, keyBytes, msgBytes)
+    * @inheritdoc
```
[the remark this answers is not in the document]
prateekm: This doesn't seem to work for some reason, ignore this one.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
-      info("No TaskName to checkpoint mapping provided.  Reading for first time.")
-      taskNamesToOffsets = readCheckpointsFromLog()
+      info("Reading checkpoints for the first time")
+      taskNamesToOffsets = readCheckpoints()
     } else {
-      info("Already existing checkpoint mapping.  Merging new offsets")
-      taskNamesToOffsets ++= readCheckpointsFromLog()
+      info("Updating existing checkpoint mappings")
+      taskNamesToOffsets ++= readCheckpoints()
+      info("Updating existing checkpoint mappings" + taskNamesToOffsets)
```
prateekm: Minor: "**Updated** checkpoint mappings: " + taskNamestoOffsets

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
 
   /**
-   * Read through entire log, discarding changelog mapping, and building map of TaskNames to Checkpoints
-   */
-  def readCheckpointsFromLog(): Map[TaskName, Checkpoint] = {
-    val checkpoints = mutable.Map[TaskName, Checkpoint]()
+    * @inheritdoc
+    */
+  override def writeCheckpoint(taskName: TaskName, checkpoint: Checkpoint) {
+    val key = new KafkaCheckpointLogKey(grouperFactory, taskName, KafkaCheckpointLogKey.CHECKPOINT_TYPE)
```
prateekm: This can throw a validation exception. Maybe catch, add more context and rethrow as SamzaException?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
-   * Reads an entry from the checkpoint log and invokes the provided lambda on it.
-   *
-   * @param handleEntry Code to handle an entry in the log once it's found
-   */
-  private def readLog(shouldHandleEntry: (KafkaCheckpointLogKey) => Boolean,
-                      handleEntry: (ByteBuffer, KafkaCheckpointLogKey) => Unit): Unit = {
-    info("Reading from checkpoint system:%s topic:%s" format(systemName, checkpointTopic))
+    * Returns the checkpoints from the log.
+    *
+    * <p> The underlying {@link SystemConsumer} is stateful and tracks its offsets. Hence, each invocation of this method
```
prateekm: Minor: [[SystemConsumer]]

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
+        throw new SamzaException(s"Encountered a checkpoint message without key. Topic:$checkpointTopic " +
+          s"Offset:${checkpointEnvelope.getOffset}")
+      }
 
-      if (!shouldHandleEntry(checkpointKey)) {
-        info("Skipping checkpoint log entry at offset %s with key %s." format(offset, checkpointKey))
-      } else {
-        val checkpointPayload = ByteBuffer.wrap(msg.getMessage.asInstanceOf[Array[Byte]])
-        handleEntry(checkpointPayload, checkpointKey)
+      val checkpointKey = checkpointKeySerde.fromBytes(keyBytes)
```
prateekm: Catch and add context (e.g. offset, the string representation of bytes etc.)?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
+      taskNamesToOffsets = readCheckpoints()
     } else {
-      info("Already existing checkpoint mapping.  Merging new offsets")
-      taskNamesToOffsets ++= readCheckpointsFromLog()
+      info("Updating existing checkpoint mappings")
+      taskNamesToOffsets ++= readCheckpoints()
+      info("Updating existing checkpoint mappings" + taskNamesToOffsets)
     }
 
     val checkpoint = taskNamesToOffsets.get(taskName).getOrElse(null)
```
prateekm: Minor: Maybe getOrElse(taskName, null)

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
+import org.apache.samza.config.{Config, JobConfig}
 import org.apache.samza.container.TaskName
+import org.apache.samza.serializers.Serde
+import org.apache.samza.metrics.MetricsRegistry
 import org.apache.samza.serializers.CheckpointSerde
-import org.apache.samza.system.SystemStreamMetadata.SystemStreamPartitionMetadata
-import org.apache.samza.system.kafka.{KafkaSystemAdmin, KafkaStreamSpec}
-import org.apache.samza.system.{StreamSpec, SystemAdmin, _}
+import org.apache.samza.system.kafka.KafkaStreamSpec
+import org.apache.samza.system._
```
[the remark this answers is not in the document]
vjagadish1989: There are multiple of these imports. So, maybe, leave them as is?
`{IncomingMessageEnvelope, OutgoingMessageEnvelope, SystemAdmin, SystemConsumer, SystemFactory, SystemProducer, SystemStreamMetadata, SystemStreamPartition, SystemStreamPartitionIterator}`

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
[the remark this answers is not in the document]
vjagadish1989: IntelliJ does not appear to show the docs on the class otherwise?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
 
     if (taskNamesToOffsets == null) {
-      info("No TaskName to checkpoint mapping provided.  Reading for first time.")
-      taskNamesToOffsets = readCheckpointsFromLog()
+      info("Reading checkpoints for the first time")
+      taskNamesToOffsets = readCheckpoints()
     } else {
-      info("Already existing checkpoint mapping.  Merging new offsets")
-      taskNamesToOffsets ++= readCheckpointsFromLog()
+      info("Updating existing checkpoint mappings")
```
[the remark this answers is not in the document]
vjagadish1989: Is there a simple way to get the startingOffset from the consumer or the iterator?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
+    systemAdmin.clearStream(checkpointSpec)
   }
 
-  private def getSSPMetadata(topic: String, partition: Partition): SystemStreamPartitionMetadata = {
-    val metaDataMap: java.util.Map[String, SystemStreamMetadata] = systemAdmin.getSystemStreamMetadata(Collections.singleton(topic))
-    val checkpointMetadata: SystemStreamMetadata = metaDataMap.get(topic)
-    if (checkpointMetadata == null) {
-      throw new SamzaException("Cannot get metadata for system=%s, topic=%s" format(systemName, topic))
+  override def stop = {
+    if (systemProducer != null) {
```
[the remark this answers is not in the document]
vjagadish1989: We already throw an exception if checkpointMetadata is null... Do you suggest also logging an error?

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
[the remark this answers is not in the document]
prateekm: I assumed we still save the startingOffset. That isn't the case anymore, so ignore this.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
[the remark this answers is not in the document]
prateekm: I was talking about the systemProducer != null and systemConsumer != null in the stop method. If they are null, that's unexpected and should be logged as an error so that we know this happens.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
-    } catch {
-      case e : StreamValidationException =>
-             if (failOnCheckpointValidation) {
-               throw e
-             } else {
-               warn("Checkpoint stream validation partially failed. Ignoring it because failOnCheckpointValidation=" + failOnCheckpointValidation)
-             }
-      case e1 : Exception => throw e1
+      val checkpointBytes = checkpointEnvelope.getMessage.asInstanceOf[Array[Byte]]
+      val checkpoint = checkpointMsgSerde.fromBytes(checkpointBytes)
```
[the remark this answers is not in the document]
vjagadish1989: Looks like it's not needed. Thanks to your suggestion, I wrote an unit test to verify this case:
```  def testNullCheckpointSerde: Unit = {
    val checkpointBytes = null.asInstanceOf[Array[Byte]]
    val checkpointSerde = new CheckpointSerde
    val checkpoint = checkpointSerde.fromBytes(checkpointBytes)
    assertNull(checkpoint)
  }
```

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -19,258 +19,212 @@
...
+      checkpoints.put(checkpointKey.getTaskName, checkpoint)
     }
+    info(s"Read $numMessagesRead messages from system:$checkpointSystem topic:$checkpointTopic")
```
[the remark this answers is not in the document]
vjagadish1989: >> I think it'll be very useful if we log progress in the loop

Great suggestion, added a progress marker! Also logged the offset just in case we want to track. 

>> We can also log a warning if numMessagesRead happens to be much larger (for some multiplier) than num tasks. This would imply that the topic isn't compacted.

I'm not sure if the warning / heuristic multiplier is particularly useful. We'll have to look at logs anyways to debug these issues.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
vjagadish1989: Since, the string representation of bytes is now logged in the serde exception, I logged the offset here. Thanks for this feedback to improve debuggability!

samza-kafka/src/main/java/org/apache/samza/checkpoint/kafka/KafkaCheckpointLogKey.java
```diff
@@ -0,0 +1,110 @@
...
+   */
+  private final String type;
+
+  public KafkaCheckpointLogKey(String type, TaskName taskName, String grouperFactoryClassName) {
+    Preconditions.checkNotNull(grouperFactoryClassName);
+    Preconditions.checkNotNull(taskName);
+    Preconditions.checkNotNull(type);
+    Preconditions.checkState(!grouperFactoryClassName.isEmpty(), "Empty grouper factory class provided");
+
+    Preconditions.checkState(type.equals(CHECKPOINT_KEY_TYPE), String.format("Invalid type provided for checkpoint key. " +
```
jmakes: The old behavior was just to ignore messages with type != checkpoint. In talking with Yi, this came from a release where we tried to combine checkpoints and some other message types in one stream (but removed it because of performance issues)

I think users should be able to ignore this exception with config.failOnCheckpointValidation in case they happen to have an old checkpoint topic that has some messages with type != checkpoint.

It's admittedly a really unlikely scenario, but the consequence is severe because you would lose your checkpoints after upgrading.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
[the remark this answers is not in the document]
vjagadish1989: Thanks, changed it in the revised draft.

vjagadish1989:
Thanks @prateekm for the diligent review. committed!

PR #350, opened 2017-11-02
SAMZA-1480: TaskStorageManager improperly initializes changelog consu…

jmakes:
…mer position when restoring a store from disk

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
+++ b/samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
@@ -68,7 +68,7 @@ class TaskStorageManager(
   }
 
   var changeLogOldestOffsets: Map[SystemStream, String] = Map()
-  val fileOffset: util.Map[SystemStreamPartition, String] = new util.HashMap[SystemStreamPartition, String]()
+  val fileOffsets: util.Map[SystemStreamPartition, String] = new util.HashMap[SystemStreamPartition, String]()
   val offsetFileName = "OFFSET"
 
   def apply(storageEngineName: String) = taskStores(storageEngineName)
@@ -104,7 +104,9 @@ class TaskStorageManager(
       } else {
         val offset = readOffsetFile(loggedStorePartitionDir)
         info("Read offset %s for the store %s from logged storage partition directory %s." format(offset, storeName, loggedStorePartitionDir))
-        fileOffset.put(new SystemStreamPartition(changeLogSystemStreams(storeName), partition), offset)
+        if (offset != null) {
+          fileOffsets.put(new SystemStreamPartition(changeLogSystemStreams(storeName), partition), offset)
+        }
       }
     })
   }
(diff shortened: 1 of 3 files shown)


jmakes:
@prateekm early patch in case we need a hot fix.
Still needs unit tests. 
Will try to get to those Friday

samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
```diff
@@ -228,11 +230,28 @@ class TaskStorageManager(
 
     for ((storeName, systemStream) <- changeLogSystemStreams) {
       val systemStreamPartition = new SystemStreamPartition(systemStream, partition)
+      val admin = systemAdmins.getOrElse(systemStream.getSystem,
+        throw new SamzaException("Unable to get systemAdmin for store " + storeName + " and systemStream" + systemStream))
```
prateekm: Minor: 
"system admin for store: " 
"and system stream: "

samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
```diff
@@ -228,11 +230,28 @@ class TaskStorageManager(
...
-        Option(fileOffset.get(systemStreamPartition))
-          .getOrElse(changeLogOldestOffsets
-            .getOrElse(systemStream, throw new SamzaException("Missing a change log offset for %s." format systemStreamPartition)))
+      val fileOffset = fileOffsets.get(systemStreamPartition)
+      val oldestOffset = changeLogOldestOffsets
+        .getOrElse(systemStream, throw new SamzaException("Missing a change log offset for %s." format systemStreamPartition));
+      val offset = if (fileOffset != null) {
+        // File offset was the last message written to the changelog that is also reflected in the store,
+        // so we start with the NEXT offset
+        val resumeOffset = admin.getOffsetsAfter(Map(systemStreamPartition -> fileOffset).asJava).get(systemStreamPartition)
```
prateekm: Minor: IMHO it's worth extracting this as a private helper method (e.g. getStartingOffset(oldestOffset, fileOffset)) with documentation + tests.

samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
```diff
@@ -228,11 +230,28 @@ class TaskStorageManager(
...
+        // If the offset we plan to use is older than the oldest offset, just use the oldest offset.
```
prateekm: Minor: Prefer moving this comment to the else block.

samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
jmakes: Totally agree. I've actually already done this on my local copy. Will update the PR when the unit tests are done.

PR #355, opened 2017-11-09
SAMZA-1487: Disable Flaky Zk Integration tests.

shanthoosh:


shanthoosh:
@prateekm Please take a look when you get a chance.

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
```diff
@@ -23,11 +23,6 @@
 import org.apache.samza.config.ZkConfig;
 import org.apache.samza.testUtils.EmbeddedZookeeper;
 import org.apache.samza.util.NoOpMetricsRegistry;
-import org.junit.After;
```
nickpan47: nit: since we are tentatively disabling this test, let's comment those import lines instead of delete them.

shanthoosh: Makes sense. Updated the patch.

PR #356, opened 2017-11-09
SAMZA-1490 - Fix TestRepartitionJoinWindowApp

lindong28:


lindong28:
@prateekm Could you review this patch?

samza-test/src/test/java/org/apache/samza/test/operator/TestRepartitionJoinWindowApp.java
```diff
@@ -71,11 +68,7 @@ public void testRepartitionJoinWindowApp() throws Exception {
       String key = message.key();
       String value = message.value();
       Assert.assertTrue(key.equals("u1") || key.equals("u2"));
-      if ("u1".equals(key)) {
-        Assert.assertEquals("3", value);
-      } else {
-        Assert.assertEquals("4", value);
-      }
+      Assert.assertEquals("2", value);
```
prateekm: Please update the documentation at line 45-47 too, so that we can understand what the expected results are.

lindong28: Sure. I updated the comment now.

PR #357, opened 2017-11-14
SAMZA-1492: Add exception counter to LocalStoreMonitor.

shanthoosh:
Add storeGCExceptionCounter to LocalStoreMonitor(as a part of LocalStoreMonitorMetrics) to enable alerts setup in an production environment.


shanthoosh:
@jmakes Please take a look.

PR #358, opened 2017-11-14
SAMZA-1495: Set intermediate streams as higher priority by default

prateekm:
Most changes in StreamConfig are formatting fixes.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
+++ b/samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
@@ -125,7 +125,7 @@ public class StreamEdge {
     if (isIntermediate()) {
       config.put(String.format(StreamConfig.IS_INTERMEDIATE_FOR_STREAM_ID(), spec.getId()), "true");
       config.put(String.format(StreamConfig.CONSUMER_OFFSET_DEFAULT_FOR_STREAM_ID(), spec.getId()), "oldest");
-      config.put(String.format(StreamConfig.PRIORITY_FOR_STREAM_ID(), spec.getId()), "1");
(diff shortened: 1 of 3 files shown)


prateekm:
cc @vjagadish1989 @xinyuiscool for review

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
```diff
@@ -125,6 +125,7 @@ Config generateConfig() {
     if (isIntermediate()) {
       config.put(String.format(StreamConfig.IS_INTERMEDIATE_FOR_STREAM_ID(), spec.getId()), "true");
       config.put(String.format(StreamConfig.CONSUMER_OFFSET_DEFAULT_FOR_STREAM_ID(), spec.getId()), "oldest");
+      config.put(String.format(StreamConfig.PRIORITY_FOR_STREAM_ID(), spec.getId()), "1");
```
vjagadish1989: If a user already has prioritized streams, would this put the intermediate stream at the same priority as their input?. Alternately, You could set to INT_MAX here to avoid conflicting with configured priorities?

vjagadish1989: Consider adding an assert to https://github.com/apache/samza/blob/master/samza-core/src/test/java/org/apache/samza/execution/TestStreamEdge.java as well if possible?

vjagadish1989 (approved):
approved. added couple of questions from my review! Thanks.

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
prateekm: Makes sense, will change to INT_MAX

PR #359, opened 2017-11-14
SAMZA-1501: Validate operator IDs so that they don't contain special characters and spaces

prateekm:


prateekm:
cc @vjagadish1989 and @xinyuiscool for review

vjagadish1989 (approved):
looks good. thank you for the contributions, approved.

PR #360, opened 2017-11-14
Extend AllSspToSingleTaskGrouper to work with Yarn and ZK JobCoordinator

atoomula:


samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -19,51 +19,116 @@
...
+
     if (ssps == null) {
       throw new SamzaException("ssp set cannot be null!");
     }
     if (ssps.size() == 0) {
       throw new SamzaException("Cannot process stream task with no input system stream partitions");
     }
 
-    final TaskName taskName = new TaskName(String.format("Task-%s", String.valueOf(containerId)));
+    if (isPassthroughJobCoordinator) {
```
vjagadish1989: @prateekm  proposed that having a different grouper - `AllSspToAllTaskGrouper` would be cleaner for this scenario. That way, we don't muddle the existing `Grouper` implementation. I agree with @prateekm that it's a good idea! Thoughts?

atoomula: If we leave implementation part aside, why do we want to have a grouper which works for only one job coordinator and yet another grouper which is similar to the existing one but works with other coordinators ? As long as the behavior for passthru does not change, we should be ok, right ? Having too many groupers is also something that we want to avoid.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -19,51 +19,116 @@
...
+
   @Override
   public SystemStreamPartitionGrouper getSystemStreamPartitionGrouper(Config config) {
-    return new AllSspToSingleTaskGrouper(config.getInt(JobConfig.PROCESSOR_ID()));
+    int taskCount;
+    int containerId = 0;
+    boolean isPassthroughJobCoordinator = false;
+    String jobCoordinatorFactoryClassName = config.get(JobCoordinatorConfig.JOB_COORDINATOR_FACTORY);
+
+    if (jobCoordinatorFactoryClassName != null && !jobCoordinatorFactoryClassName.isEmpty()) {
```
nickpan47: Ultimately, the whole section here is to find the Set&lt;String&gt; processorNames. I would suggest to fold all this into a single method getProcessorNames() and pass that to the grouper's constructor.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -19,51 +19,116 @@
...
+        taskCount = getNumProcessors(config);
+      } else if (jobCoordinatorFactoryClassName.equals(PASSTHROUGH_JOB_COORDINATOR_FACTORY)) {
+        if (config.get(JobConfig.PROCESSOR_ID()) == null) {
+          throw new ConfigException("Could not find " + JobConfig.PROCESSOR_ID() + " in Config!");
+        }
+        isPassthroughJobCoordinator = true;
+        containerId = config.getInt(JobConfig.PROCESSOR_ID());
+        taskCount = 1;
+      } else {
+        throw new ConfigException("AllSspToSingleTaskGrouperFactory is not yet supported with "
```
nickpan47: Thanks for adding this! +1

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -19,51 +19,116 @@
...
+            + jobCoordinatorFactoryClassName);
+      }
+    } else {
+      // must be cluster-based jobCoordinator (Yarn)
```
nickpan47: I would recommend to add a log line here since we are making a guess.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -19,51 +19,116 @@
...
   @Override
   public SystemStreamPartitionGrouper getSystemStreamPartitionGrouper(Config config) {
-    return new AllSspToSingleTaskGrouper(config.getInt(JobConfig.PROCESSOR_ID()));
+    int taskCount;
+    int containerId = 0;
+    boolean isPassthroughJobCoordinator = false;
+    String jobCoordinatorFactoryClassName = config.get(JobCoordinatorConfig.JOB_COORDINATOR_FACTORY);
+
+    if (jobCoordinatorFactoryClassName != null && !jobCoordinatorFactoryClassName.isEmpty()) {
+      if (jobCoordinatorFactoryClassName.equals(ZK_JOB_COORDINATOR_FACTORY)) {
```
nickpan47: A further modification to make the code simpler would be to add an interface method getProcessorNames() to JobCoordinatorFactory class and let each implementation class to implement it (or throw exception by default). That would avoid all the if-then-else cases here and avoid copying ZkJobCoordinatorFactory code here.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
nickpan47: Likewise, do we have any validation of usage of this grouper against input stream configuration? (i.e. this grouper does not work w/ broadcast stream, shouldn't we validate/check it either in group() or in JobCoordinator when we invoking group()?)

samza-core/src/test/java/org/apache/samza/container/grouper/stream/TestAllSspToSingleTaskGrouper.java
```diff
@@ -0,0 +1,108 @@
...
+    partitions.add(aa1);
+    partitions.add(aa2);
+    partitions.add(ab0);
+    expectedResult.put(new TaskName("Task-0"), partitions);
+    expectedResult.put(new TaskName("Task-1"), partitions);
+
+    assertEquals(expectedResult, result);
+  }
+
+  @Test
```
nickpan47: Can't we also add the test for ZkJobCoordinator as well?

samza-core/src/test/java/org/apache/samza/container/grouper/stream/TestAllSspToSingleTaskGrouper.java
```diff
@@ -0,0 +1,108 @@
...
+    partitions.add(aa0);
+    partitions.add(aa1);
+    partitions.add(aa2);
+    partitions.add(ab0);
+    expectedResult.put(new TaskName("Task-1"), partitions);
+
+    assertEquals(expectedResult, result);
+  }
+
+  @Test(expected = ConfigException.class)
```
nickpan47: How about other test cases? Like PassthroughJobCoordinator failed to get processorId, ZkJobCoordinator failed to get the processor count, etc.?

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -19,51 +19,116 @@
...
+      taskCount = config.getInt(YARN_CONTAINER_COUNT);
+    }
+
+    Validate.isTrue(taskCount > 0, "Task count cannot be <= 0");
+
+    return new AllSspToSingleTaskGrouper(isPassthroughJobCoordinator, containerId, taskCount);
+  }
+
+  private int getNumProcessors(Config config) {
+    ZkConfig zkConfig = new ZkConfig(config);
```
nickpan47: This puts a requirement that when getSystemStreamPartitionGrouper() is called, the ZkJobCoordinator leader election has completed. Let's make sure we document somewhere (i.e. depending on where we move this code to).

samza-test/src/test/java/org/apache/samza/test/processor/TestZkLocalApplicationRunner.java
```diff
@@ -268,6 +285,99 @@ public void shouldStopNewProcessorsJoiningGroupWhenNumContainersIsGreaterThanNum
     assertEquals(NUM_KAFKA_EVENTS, processedMessagesLatch.getCount());
   }
 
+  /**
```
nickpan47: 👍 like this detailed test docs! Thanks!

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -19,51 +19,116 @@
...
  */
+
 class AllSspToSingleTaskGrouper implements SystemStreamPartitionGrouper {
+  private final boolean isPassthroughJobCoordinator;
+  private final int taskCount;
   private final int containerId;
 
-  public AllSspToSingleTaskGrouper(int containerId) {
+  AllSspToSingleTaskGrouper(boolean isPassthroughJobCoordinator, int containerId, int taskCount) {
+    this.isPassthroughJobCoordinator = isPassthroughJobCoordinator;
```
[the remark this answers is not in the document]
atoomula: Sure. I added getProcessorNames API to JobCoordinator and simplified this part of the code.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -19,51 +19,116 @@
...
- * This is useful, in case of using load-balanced consumers like the new Kafka consumer, Samza doesn't control the
- * partitions being consumed by a task. Hence, it is assumed that there is only 1 task that processes all messages,
- * irrespective of which partition it belongs to.
- * This also implies that container and tasks are synonymous when this grouper is used. Taskname(s) has to be globally
- * unique within a given job.
+ * AllSspToSingleTaskGrouper creates TaskInstances equal to the number of containers and assigns all partitions to be
+ * consumed by each TaskInstance. This is useful, in case of using load-balanced consumers like the high-level Kafka
+ * consumer and Kinesis consumer, where Samza doesn't control the partitions being consumed by the task.
  *
- * Note: This grouper does not take in broadcast streams yet.
```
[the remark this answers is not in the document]
atoomula: I have added validation for broadcast streams in the grouper factory. It is up to the individual job coordinators to see if they can support getProcessorNames API, otherwise throw an exception. Is there a way to validate if the consumers are load-balanced ? On the flip side, do we need to state that this is supported ONLY for load-balanced consumers ? This could be used with non-LB consumers as well and it does what it exactly states.

samza-core/src/test/java/org/apache/samza/container/grouper/stream/TestAllSspToSingleTaskGrouper.java
atoomula: PassthroughJobCoordinator should have processorId even if it is not explicitly added in the config due to the default processorId generator. Added a test for it.

As described above, I will talk to you about the UT for ZkJobCoordinator.

samza-core/src/test/java/org/apache/samza/container/grouper/stream/TestAllSspToSingleTaskGrouper.java
atoomula: Testing ZkJobCoordinator is tricky with getProcessorNames API in job-coordinator. I'm not sure if testing by mocking getSSPGrouper would be of any further value given that we already have tests for yarn and passthru. Let me check with you to see if you have any ideas around this.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
atoomula: I have added a comment in the code. I looked around to see if there is a way to verify that but couldn't find any.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
atoomula: Sure

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
atoomula: Added validation for broadcast stream.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
atoomula: +1! Added the API just like you suggested.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -73,62 +66,40 @@
...
-        containerId = config.getInt(JobConfig.PROCESSOR_ID());
-        taskCount = 1;
-      } else {
-        throw new ConfigException("AllSspToSingleTaskGrouperFactory is not yet supported with "
-            + jobCoordinatorFactoryClassName);
-      }
-    } else {
-      // must be cluster-based jobCoordinator (Yarn)
-      taskCount = config.getInt(YARN_CONTAINER_COUNT);
+      JobCoordinator jc = Util.
```
nickpan47: There is still a case missing here: for cluster-based jobCoordinator, there is no configuration for the JC factory (yet-to-be-done refactoring work). Hence, a default method to create processorNames from the YARN_CONTAINER_COUNT is still needed now.

atoomula: Are you referring to the code that is already there on line 100-103 which gets the processorNames from YARN_CONTAINER_COUNT ?

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -19,51 +19,87 @@
...
+  private static final String YARN_CONTAINER_COUNT = "yarn.container.count";
+
   @Override
   public SystemStreamPartitionGrouper getSystemStreamPartitionGrouper(Config config) {
-    return new AllSspToSingleTaskGrouper(config.getInt(JobConfig.PROCESSOR_ID()));
+    if (config.containsKey(TaskConfigJava.BROADCAST_INPUT_STREAMS)) {
+      throw new ConfigException("The job configured with AllSspToSingleTaskGrouper cannot have broadcast streams.");
+    }
+
+    String jobCoordinatorFactoryClassName = config.get(JobCoordinatorConfig.JOB_COORDINATOR_FACTORY);
```
nickpan47: Generally, please do not refer to the constant config names directly outside the scope of the config. Use TaskConfigJava.getBroadcastSystemStreams() and JobCoordinatorConfig.getJobCoordinatorFactoryClassName() instead.

samza-core/src/main/java/org/apache/samza/standalone/PassthroughJobCoordinator.java
```diff
@@ -154,4 +155,9 @@ private String createProcessorId(Config config) {
               ApplicationConfig.APP_PROCESSOR_ID_GENERATOR_CLASS));
```
nickpan47: Same here, use method instead of directly referring to the config name.

samza-core/src/test/java/org/apache/samza/container/grouper/stream/TestAllSspToSingleTaskGrouper.java
```diff
@@ -0,0 +1,144 @@
...
+  private SystemStreamPartition aa2 = new SystemStreamPartition("SystemA", "StreamA", new Partition(2));
+  private SystemStreamPartition ab0 = new SystemStreamPartition("SystemA", "StreamB", new Partition(0));
+  private AllSspToSingleTaskGrouperFactory grouperFactory = new AllSspToSingleTaskGrouperFactory();
+
+  @Test
+  public void testLocalStreamGroupedCorrectlyForYarn() {
+    HashSet<SystemStreamPartition> allSSPs = new HashSet<>();
+    HashMap<String, String> configMap = new HashMap<>();
+
+    configMap.put("yarn.container.count", "2");
```
nickpan47: Please use "job.container.count" instead. yarn.container.count is deprecated.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
nickpan47: Ah, got it! I only checked the commit-after-last-review and somehow the diff was skipped. lgtm!

samza-core/src/test/java/org/apache/samza/container/grouper/stream/TestAllSspToSingleTaskGrouper.java
atoomula: Sure.

samza-core/src/main/java/org/apache/samza/standalone/PassthroughJobCoordinator.java
atoomula: Sure. These are not my changes. But refactored code a bit here.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
atoomula: Sure.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -19,51 +19,87 @@
...
+    });
 
-    return Collections.singletonMap(taskName, ssps);
+    return groupedMap;
   }
 }
 
 public class AllSspToSingleTaskGrouperFactory implements SystemStreamPartitionGrouperFactory {
+  private static final Logger LOG = LoggerFactory.getLogger(AllSspToSingleTaskGrouperFactory.class);
+  private static final String YARN_CONTAINER_COUNT = "yarn.container.count";
```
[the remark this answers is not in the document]
atoomula: Sure.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -19,51 +19,83 @@
...
+
+    try {
+      String jobCoordinatorFactoryClassName = new JobCoordinatorConfig(config).getJobCoordinatorFactoryClassName();
+      JobCoordinator jc = Util.<JobCoordinatorFactory>getObj(jobCoordinatorFactoryClassName).getJobCoordinator(config);
+      return new AllSspToSingleTaskGrouper(jc.getProcessorNames());
+    } catch (ConfigException ex) {
+      LOG.info("Guessing cluster-based jobCoordinator(Yarn).");
+      final Set<String> processorNames = new HashSet<>();
+      IntStream.range(0,
+          config.getInt(JobConfig.JOB_CONTAINER_COUNT())).forEach(i -> processorNames.add(String.valueOf(i)));
```
nickpan47: Why are we still using the constant name here, instead of new JobConfig(config).getContainerCount()?

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -19,51 +19,83 @@
...
   @Override
   public SystemStreamPartitionGrouper getSystemStreamPartitionGrouper(Config config) {
-    return new AllSspToSingleTaskGrouper(config.getInt(JobConfig.PROCESSOR_ID()));
+    if (!(new TaskConfigJava(config).getBroadcastSystemStreams().isEmpty())) {
+      throw new ConfigException("The job configured with AllSspToSingleTaskGrouper cannot have broadcast streams.");
+    }
+
+    try {
+      String jobCoordinatorFactoryClassName = new JobCoordinatorConfig(config).getJobCoordinatorFactoryClassName();
+      JobCoordinator jc = Util.<JobCoordinatorFactory>getObj(jobCoordinatorFactoryClassName).getJobCoordinator(config);
```
nickpan47: Just noticed here. Inside the grouper factory, we should not create another JobCoordinator instance, or, at least this JobCoordinator instance that is used to get processor names should not be exposed to the user. Otherwise, we run into the risk that there are two JC instances created and joining the leader election in the same process.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -167,6 +172,19 @@ public String getProcessorId() {
     return processorId;
   }
 
+  @Override
+  public Set<String> getProcessorNames() {
+    final Set<String> processorNames = new HashSet<>();
+    // Leader election could have happened by this time.
```
nickpan47: could? Or should?
And it would be nice to fail here if leader election has not been done. This would require some additional work in ZkJobCoordinator. Please open another JIRA to track it.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
atoomula: Added jira ticket and comments as discussed.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
atoomula: I missed this. Thanks for pointing out.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
atoomula: Opened jira to check if leader election is done and added it to comments.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -167,6 +172,20 @@ public String getProcessorId() {
...
+  public Set<String> getProcessorNames() {
+    final Set<String> processorNames = new HashSet<>();
+    // Leader election should have happened by this time.
+    // SAMZA-1522 Add an API to check if leader election is done.
+    ZkConfig zkConfig = new ZkConfig(config);
+    ZkClient zkClient =
+        createZkClient(zkConfig.getZkConnect(), zkConfig.getZkSessionTimeoutMs(), zkConfig.getZkConnectionTimeoutMs());
+    ZkKeyBuilder keyBuilder = new ZkKeyBuilder(getJobCoordinationZkPath(config));
+    int numProcessors = zkClient.countChildren(keyBuilder.getProcessorsPath());
+    IntStream.range(0, numProcessors).forEach(i -> processorNames.add(String.valueOf(i)));
```
shanthoosh: Just curious and want to understand this better.
A.  We already have processorId with which each processor joins the group. Wouldn't that processorIDList be sufficient here?
B.  If we simply count children and generate processNames based upon that, we wouldn't be able to distinguish different processor processorId groups(say processorIds: [1,2,3,4] vs processorIds:  [1,2,3,5])?

vjagadish1989:
@atoomula Can we resolve these conflicts? Is this PR already superseded by another PR? If so, please can you help close this one! Thank you :-)

atoomula:
@vjagadish1989 This is superseded by https://github.com/apache/samza/pull/383. I'm closing this one. Thanks!

PR #361, opened 2017-11-16
SAMZA-1504: Allow user to register container-level metrics

xinyuiscool:
This change allows user to register the metrics on the per-container basis.

Tested in beam runner and works as expected.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/container/SamzaContainerContext.java
+++ b/samza-api/src/main/java/org/apache/samza/container/SamzaContainerContext.java
@@ -20,6 +20,7 @@
 package org.apache.samza.container;
 
 import org.apache.samza.config.Config;
+import org.apache.samza.metrics.MetricsRegistry;
 
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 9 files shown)


samza-api/src/main/java/org/apache/samza/container/SamzaContainerContext.java
```diff
@@ -31,20 +32,38 @@
...
+      MetricsRegistry metricsRegistry) {
     this.id = id;
     this.config = config;
     this.taskNames = Collections.unmodifiableCollection(taskNames);
+    this.metricsRegistry = metricsRegistry;
+  }
+
+  /**
+   * See also {@link SamzaContainerContext#SamzaContainerContext(String, Config, Collection, MetricsRegistry)}.
+   * NOTE: Used for testing only.
```
prateekm: Can we make this protected or remove public then?

xinyuiscool: Actually the tests in other packages creates container context so I wasn't able to make it pacakge private. But on a second thought, this constructor is kind of confusing and make stuff ugly. I removed it and add the metric registry to all the tests now.

PR #362, opened 2017-11-16
SAMZA-1507: Create changelog streams in Leader(ZkJobCoordinator) for stateful operators.

shanthoosh:
Verified with a test standalone job. Will add integration test for this as a part of fixing and reenabling standalone integration tests.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
@@ -20,12 +20,13 @@ package org.apache.samza.zk;
 
 import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashSet;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import org.I0Itec.zkclient.IZkStateListener;
+import java.util.Objects;
 import java.util.Set;
+import org.I0Itec.zkclient.IZkStateListener;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.samza.config.ApplicationConfig;
 import org.apache.samza.config.Config;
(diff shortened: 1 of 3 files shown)


shanthoosh:
@sborya @prateekm

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -188,6 +189,7 @@ void doOnProcessorChange(List<String> processors) {
 
     // Generate the JobModel
     JobModel jobModel = generateNewJobModel(currentProcessorIds);
+    JobModelManager.createChangeLogStreams(new StorageConfig(config), jobModel.maxChangeLogStreamPartitions);
```
prateekm: Where is the maxChangeLogStreamPartitions being set on the JobModel?

shanthoosh: Set in JobModel constructor.

It's populated in JobModelManager.readJobModel(which generates changeLogPartitionId of the tasks) and JobModel constructor sets it to maximum changeLogPartitionId of all tasks.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -279,7 +283,19 @@ private String createProcessorId(Config config) {
    * Generate new JobModel when becoming a leader or the list of processor changed.
    */
   private JobModel generateNewJobModel(List<String> processors) {
-    return JobModelManager.readJobModel(this.config, Collections.emptyMap(), null, streamMetadataCache, processors);
+    Map<TaskName, Integer> changeLogToPartitionMap = new HashMap<>();
```
sborya: I think we should try to avoid doing it on every job model generation.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -279,7 +290,20 @@ private String createProcessorId(Config config) {
    * Generate new JobModel when becoming a leader or the list of processor changed.
    */
   private JobModel generateNewJobModel(List<String> processors) {
-    return JobModelManager.readJobModel(this.config, Collections.emptyMap(), null, streamMetadataCache, processors);
+    String zkJobModelVersion = zkUtils.getJobModelVersion();
+    // If JobModel exists in zookeeper && cached JobModel version is unequal to JobModel version stored in zookeeper.
+    if (zkJobModelVersion != null && !Objects.equals(cachedJobModelVersion, zkJobModelVersion)) {
```
sborya: why do you use Objects.equals?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -188,6 +195,10 @@ void doOnProcessorChange(List<String> processors) {
 
     // Generate the JobModel
     JobModel jobModel = generateNewJobModel(currentProcessorIds);
+    if (!hasCreatedChangeLogStreams) {
```
sborya: instead of boolean we should check number of partitions. Start with value 0. If number of partitions matches - skip creation.

PR #363, opened 2017-11-17
SAMZA-1482: add config documentation for auto-restart/fail behavior o…

nickpan47:
…n partition count changes

### DIFF
--- a/docs/learn/documentation/versioned/jobs/configuration-table.html
+++ b/docs/learn/documentation/versioned/jobs/configuration-table.html
@@ -241,8 +241,9 @@
                     <td class="description">
                         The frequency at which the input streams' partition count change should be detected. When the input
                         partition count change is detected, Samza will automatically restart a stateless job or fail a
-                        stateful job. This check can be tuned to be pretty low as partition increase is not a common event.
-                        You can completely disable partition count monitoring by setting this value to 0 or a negative integer,
+                        stateful job. A longer time interval is recommended for jobs w/ large number of input system stream
+                        partitions, since gathering partition count may incur measurable overhead to the job. You can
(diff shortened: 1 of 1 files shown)


docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -231,20 +231,19 @@ <h1>Samza Configuration Reference</h1>
...
 
                 <tr>
                     <td class="property" id="job-coordinator-monitor-partition-change-frequency-ms">job.coordinator.<br />monitor-partition-change.frequency.ms</td>
                     <td class="default">300000</td>
                     <td class="description">
-                        The frequency at which the input streams' partition count change should be detected. This check
-                        can be tuned to be pretty low as partition increase is not a common event.
+                        The frequency at which the input streams' partition count change should be detected. When the input
+                        partition count change is detected, Samza will automatically restart a stateless job or fail a
+                        stateful job. This check can be tuned to be pretty low as partition increase is not a common event.
```
vjagadish1989: "This check can be tuned to be pretty low as partition increase is not a common event."

Do we need this? (It's slightly confusing as a lower value of this field would mean aggressive partition monitoring)

vjagadish1989 (approved):
approved.looks good to me. added one question from my review.

docs/learn/documentation/versioned/jobs/configuration-table.html
nickpan47: Make sense. Does the following reads better?

"A longer time interval value is recommended for jobs w/ large number of input system stream partitions, since gathering partition count may incur measurable overhead to the job."

vjagadish1989: +1, reads better.

PR #364, opened 2017-11-20
SAMZA-1505: Fix CheckpointTool writing only one ssp per task

xinyuiscool:
Currently when using CheckpointTool to write checkpoints, it only writes a checkpoint of a single ssp per task. By debugging the code, looks like the flatMap() on the checkpoint of Optional tuple(taskname -> Map(ssp -> offset)) merges the results by key taskname. This patch stores the results explicitly in a list and then groupBy() on it, which fixes the problem.

### DIFF
--- a/samza-core/src/test/scala/org/apache/samza/checkpoint/TestCheckpointTool.scala
+++ b/samza-core/src/test/scala/org/apache/samza/checkpoint/TestCheckpointTool.scala
@@ -21,7 +21,6 @@ package org.apache.samza.checkpoint
 
 import java.util
 
-import com.sun.scenario.effect.Offset
 import org.apache.samza.Partition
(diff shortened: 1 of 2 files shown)


xinyuiscool:
@prateekm Please help me take a look. Thanks!

samza-core/src/test/scala/org/apache/samza/checkpoint/TestCheckpointTool.scala
```diff
@@ -19,7 +19,11 @@
 
 package org.apache.samza.checkpoint
 
+import java.util
+
+import com.sun.scenario.effect.Offset
```
bharathkk: unused import?

xinyuiscool: Thanks for the catch! Fixed it.

PR #367, opened 2017-11-23
SAMZA-1508: JobRunner should not return success until the job is healthy

jmakes:


jmakes:
@xinyuiscool @nickpan47 @prateekm take a look when you get a chance!

prateekm:
@jmakes Does this still need to be reviewed/merged?

jmakes:
@prateekm Yes. I haven't pressed it because there is a tools bug in LI preventing us from adopting it. But it's very much needed.

jmakes:
@prateekm @nickpan47 @vjagadish looks like we're finally going to be able to proceed with this patch. Can I get a review, please?

samza-yarn/src/main/java/org/apache/samza/webapp/ApplicationMasterRestClient.java
```diff
@@ -0,0 +1,111 @@
...
+    String jsonString = getEntityAsJson("/am", "AM state");
+    return jsonMapper.readValue(jsonString, new TypeReference<Map<String, Object>>() {});
+  }
+
+  /**
+   * @return  the config as a map of key to value
+   * @throws IOException if there was an error fetching the config from the servlet.
+   */
+  public Map<String, Object> getConfig() throws IOException {
+    String jsonString = getEntityAsJson("/config", "config");
```
prateekm: Does this mask sensitive configs?

samza-yarn/src/main/scala/org/apache/samza/job/yarn/YarnJob.scala
```diff
@@ -130,14 +130,14 @@ class YarnJob(config: Config, hadoopConfig: Configuration) extends StreamJob {
...
-    Running
+    getStatus
   }
 
   def getStatus: ApplicationStatus = {
     getAppId match {
       case Some(appId) =>
         logger.info("Getting status for applicationId %s" format appId)
-        client.status(appId).getOrElse(null)
+        client.status(appId).get // Should throw if None
```
prateekm: Should this log something before throwing?

samza-yarn/src/main/scala/org/apache/samza/webapp/ApplicationMasterRestServlet.scala
```diff
@@ -19,44 +19,41 @@
...
+    metricsRegistry.getGroups.asScala.foreach(group => {
       val groupMap = new HashMap[String, Object]
 
-      registry.getGroup(group).asScala.foreach {
+      metricsRegistry.getGroup(group).asScala.foreach {
         case (name, metric) =>
           metric.visit(new MetricsVisitor() {
             def counter(counter: Counter) =
-              groupMap.put(counter.getName, counter.getCount: java.lang.Long)
+              groupMap.put(counter.getName, counter.getCount: lang.Long)
```
prateekm: Should this be .asInstanceOf[Long]? Also, are the annotations required?

samza-yarn/src/main/java/org/apache/samza/webapp/ApplicationMasterRestClient.java
jmakes: It doesn't and I don't think it should. This class is merely a client that retrieves the values of the AM endpoint. Any masking should either be done on the AM (if we don't want any sensitive configs to be exposed e.g. via curl), or on top of the client to prevent logging the sensitive configs, for example.

samza-yarn/src/main/scala/org/apache/samza/job/yarn/YarnJob.scala
jmakes: There's some logging in the client.status method. None value should never be returned, so I'll make that an exception, as it would represent a bug that users can't really do anything about.

samza-yarn/src/main/scala/org/apache/samza/webapp/ApplicationMasterRestServlet.scala
jmakes: My scala foo is weak, but after looking into it. It looks like ascription does a better job of disambiguation than casting. I'd rather not trade compile-time errors for runtime-errors just for the sake of convention. Is there something more you know about this behavior that I'm missing?

prateekm: No, just asking why they're inconsistent.

Looks like the annotation is required since the original return type is `long` which gets converted to scala.Long which does not extend Object. Type annotation makes it box to a java.util.Long. So makes sense as is.

samza-yarn/src/main/java/org/apache/samza/webapp/ApplicationMasterRestClient.java
prateekm: Makes sense. Agree that AM API should mask these configs.

PR #369, opened 2017-11-28
SAMZA-1513: Doc updates for persistent windows, joins and serdes.

vjagadish1989:
@jmakes @prateekm @nickpan47 for review.

### DIFF
--- a/docs/startup/preview/index.md
+++ b/docs/startup/preview/index.md
@@ -127,10 +127,10 @@ Since the 0.13.0 release, Samza provides a new high level API that simplifies yo
 
 Check out some examples to see the high-level API in action.
 
-1.  [Pageview AdClick Joiner](https://github.com/apache/samza-hello-samza/blob/c87ed565fbaebf2ac88376143c65e9f52f7a8801/src/main/java/samza/examples/cookbook/PageViewAdClickJoiner.java) demonstrates joining a stream of PageViews with a stream of AdClicks, e.g. to analyze which pages get the most ad clicks.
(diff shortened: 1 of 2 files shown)


docs/startup/preview/index.md
```diff
@@ -247,12 +246,14 @@ MessageStream<String> shortWords = words.filter(word -> word.size() < 3);
 
 ### PartitionBy
 Re-partitions this MessageStream using the key returned by the provided keyExtractor and returns the transformed MessageStream. Messages are sent through an intermediate stream during repartitioning.
```
jmakes: Since this is the first place we use operatorId, do you think it would be helpful to briefly explain what it's for?
Or at least point them to the javadoc for an explanation.

vjagadish1989:
Address @jmakes's feedback.

docs/startup/preview/index.md
vjagadish1989: Good catch. I've updated the docs to add info on operatorId. Thanks!

docs/startup/preview/index.md
```diff
@@ -254,6 +254,9 @@ MessageStream<KV<String, PageView>> partitionedPageViews =
                                         pageView -> pageView, // value extractor
                                         KVSerde.of(new StringSerde(), new JsonSerdeV2<>(PageView.class)), // serdes
                                         "partitioned-page-views") // operator ID
+
+The operator ID should be unique for an operator within the application and is used to identify the streams and stores created by the operator.
```
jmakes: Suggestion. Take or leave it:
"The operator ID should be unique for an operator within the application and is used to identify the streams and stores created by the operator so they can be reused even if the application logic changes."

PR #370, opened 2017-11-28
SAMZA-1516: Another round of issues found by BEAM tests

xinyuiscool:
A couple of more fixes: 1. fix a bug of identifying input streams for an operator. 2. for partitionBy, set the partitionKey to 0L when key is null.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
+++ b/samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
@@ -156,12 +156,12 @@ public class OperatorImplGraph {
   OperatorImpl createAndRegisterOperatorImpl(OperatorSpec prevOperatorSpec, OperatorSpec operatorSpec,
       SystemStream inputStream, Config config, TaskContext context) {
 
+    final OperatorImpl operatorImpl;
     if (!operatorImpls.containsKey(operatorSpec.getOpId()) || operatorSpec instanceof JoinOperatorSpec) {
       // Either this is the first time we've seen this operatorSpec, or this is a join operator spec
       // and we need to create 2 partial join operator impls for it. Initialize and register the sub-DAG.
-      OperatorImpl operatorImpl = createOperatorImpl(prevOperatorSpec, operatorSpec, config, context);
+      operatorImpl = createOperatorImpl(prevOperatorSpec, operatorSpec, config, context);
       operatorImpl.init(config, context);
-      operatorImpl.registerInputStream(inputStream);
 
       // Note: The key here is opImplId, which may not equal opId for some impls (e.g. PartialJoinOperatorImpl).
       // This is currently OK since we don't need to look up a partial join operator impl again during traversal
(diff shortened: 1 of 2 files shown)


samza-core/src/main/java/org/apache/samza/operators/impl/PartitionByOperatorImpl.java
```diff
@@ -77,7 +77,8 @@ protected void handleInit(Config config, TaskContext context) {
       TaskCoordinator coordinator) {
     K key = keyFunction.apply(message);
     V value = valueFunction.apply(message);
-    collector.send(new OutgoingMessageEnvelope(systemStream, null, key, value));
+    Long partitionKey = key == null ? 0L : null;
```
fredji97: after checking with Xinyu, Kafka has a tricky behavior to do round robin  when both the partition key and key are null. It might be better to add a line of comment in code as well for future maintenance reference.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
```diff
@@ -175,8 +176,15 @@ OperatorImpl createAndRegisterOperatorImpl(OperatorSpec prevOperatorSpec, Operat
...
-      // and registered, so we do not need to traverse the DAG further.
-      return operatorImpls.get(operatorSpec.getOpId());
+      // and registered. We still need to traverse the DAG further to register the input streams.
+      OperatorImpl operatorImpl = operatorImpls.get(operatorSpec.getOpId());
+      operatorImpl.registerInputStream(inputStream);
+
+      Collection<OperatorSpec> registeredSpecs = operatorSpec.getRegisteredOperatorSpecs();
+      registeredSpecs.forEach(registeredSpec -> {
+          createAndRegisterOperatorImpl(operatorSpec, registeredSpec, inputStream, config, context);
+        });
```
vjagadish1989: wondering if this logic is needed here? Is the following equivalent?

I'd flip the statement that adds it to the map, and add it to the `operatorImpls ` after the recursive call returns?

```
  OperatorImpl createAndRegisterOperatorImpl(OperatorSpec prevOperatorSpec, OperatorSpec operatorSpec,
      SystemStream inputStream, Config config, TaskContext context) {
    if (!operatorImpls.containsKey(operatorSpec.getOpId()) || operatorSpec instanceof JoinOperatorSpec) {
      OperatorImpl operatorImpl = createOperatorImpl(prevOperatorSpec, operatorSpec, config, context);
      operatorImpl.init(config, context);
      operatorImpl.registerInputStream(inputStream);

      Collection<OperatorSpec> registeredSpecs = operatorSpec.getRegisteredOperatorSpecs();
      registeredSpecs.forEach(registeredSpec -> {
          OperatorImpl nextImpl = createAndRegisterOperatorImpl(operatorSpec, registeredSpec, inputStream, config, context);
          operatorImpl.registerNextOperator(nextImpl);
        });
      operatorImpls.put(operatorImpl.getOpImplId(), operatorImpl);
      return operatorImpl;
    } else {
      return operatorImpls.get(operatorSpec.getOpId());
    }
  }
```

samza-core/src/main/java/org/apache/samza/operators/impl/PartitionByOperatorImpl.java
vjagadish1989: This behavior in Kafka is pretty reasonable. A `null` key actually means that a key is absent (as opposed to a key being null) and the message is non-keyed. Imagine all non-keyed messages going to the same partition 0? That would be very sub-optimal. For stream-processing use-cases, we want to use `null` for `group`ing by key. Hence, we need this to get around this.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
xinyuiscool: This logic is needed since I need to further traverse the graph even if the operator is already registered. That way the input stream can be registered to the downstream operators.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
```diff
@@ -175,8 +176,15 @@ OperatorImpl createAndRegisterOperatorImpl(OperatorSpec prevOperatorSpec, Operat
       return operatorImpl;
     } else {
       // the implementation corresponding to operatorSpec has already been instantiated
-      // and registered, so we do not need to traverse the DAG further.
-      return operatorImpls.get(operatorSpec.getOpId());
+      // and registered. We still need to traverse the DAG further to register the input streams.
+      OperatorImpl operatorImpl = operatorImpls.get(operatorSpec.getOpId());
+      operatorImpl.registerInputStream(inputStream);
```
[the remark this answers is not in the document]
xinyuiscool: Sure, I can extract the registerInputStream outside the if .. else. SHould be straightforward to do.

PR #372, opened 2017-11-29
SAMZA-1518: Add CPU utilization and thread count to JVM metrics

jmakes:


samza-core/src/main/scala/org/apache/samza/metrics/JvmMetrics.scala
```diff
@@ -63,6 +68,10 @@ class JvmMetrics(val registry: MetricsRegistry) extends MetricsHelper with Runna
   val cGcCount = newCounter("gc-count")
   val cGcTimeMillis = newCounter("gc-time-millis")
 
+  val gProcessCpuUsage = if (osMXBean.isInstanceOf[OperatingSystemMXBean]) newGauge("process-cpu-usage", 0.0) else null
```
vjagadish1989: Do we need these metrics to be null? Or, Do you think it's convenient to report zero values for metrics instead of not reporting them at all?

samza-core/src/main/scala/org/apache/samza/metrics/JvmMetrics.scala
```diff
@@ -153,4 +163,16 @@ class JvmMetrics(val registry: MetricsRegistry) extends MetricsHelper with Runna
...
     gThreadsTerminated.set(threadsTerminated)
   }
+
+  private def updateOperatingSystemMetrics {
+    if (osMXBean.isInstanceOf[OperatingSystemMXBean]) {
+      val operatingSystemMXBean = osMXBean.asInstanceOf[OperatingSystemMXBean]
+      gProcessCpuUsage.set(operatingSystemMXBean.getProcessCpuLoad * PCT)
+      gSystemCpuUsage.set(operatingSystemMXBean.getSystemCpuLoad * PCT)
+
+      if (osMXBean.isInstanceOf[UnixOperatingSystemMXBean]) {
```
vjagadish1989: Can we consolidate these `isInstanceOf` checks to a single flag `isUnix` and set it during construction?

samza-core/src/main/scala/org/apache/samza/metrics/JvmMetrics.scala
```diff
@@ -153,4 +163,16 @@ class JvmMetrics(val registry: MetricsRegistry) extends MetricsHelper with Runna
...
+        gOpenFileDescriptorCount.set(osMXBean.asInstanceOf[UnixOperatingSystemMXBean].getOpenFileDescriptorCount)
```
vjagadish1989: Since Samza only supports *nix systems, I wonder if we can directly check for `UnixOperatingSystemMxBean` (this extends from `OperatingSystemMXBean` anyways?) 
It should work for posix systems like *nix, and os-x.

vjagadish1989 (approved):
Looks great. approved

samza-core/src/main/scala/org/apache/samza/metrics/JvmMetrics.scala
jmakes: I think it's less confusing to not emit the metrics if they're not functional.

Makes for ugly code, but a better user experience.

samza-core/src/main/scala/org/apache/samza/metrics/JvmMetrics.scala
jmakes: Spoken like a non-MSFT employee :-)

I prefer it this way, just in case future versions of the JVM start returning a different type. It's frustrating that we have to do these checks and casts in a virtual machine, but since we do, I'd like them to be as robust as possible.

docs/learn/documentation/versioned/container/metrics-table.html
```diff
@@ -345,6 +345,18 @@ <h1>Samza Metrics Reference</h1>
         <td>threads-terminated</td>
         <td>Current number of terminated threads</td>
     </tr>
+    <tr>
+        <td>process-cpu-usage</td>
+        <td>Current CPU usage of the JVM process as a percentage from 0 to 100. A negative number indicates the value was not available from the operating system.</td>
+    </tr>
+    <tr>
+        <td>system-cpu-usage</td>
+        <td>Current CPU usage of the JVM process as a percentage from 0 to 100. A negative number indicates the value was not available from the operating system.</td>
```
prateekm: "CPU usage of the *system* as "

PR #373, opened 2017-11-30
SAMZA-1407 upgrade junit version to 4.12

fredji97:
"./gradlew clean check" passed


vjagadish1989:
@fredji97 : Please can you resolve the conflict and we should be good to merge this.

PR #375, opened 2017-12-01
SAMZA-1506: Fix for robust ContainerHeartbeatMonitor exception handling.

abhishekshivanna:
The Fix includes the following changes:
- Catch all exceptions inside the heartbeat thread and not just
  IOException.
- A time based force kill when the heartbeat is invalid,
  this makes the monitor immune to threads that may keep the
  container stuck in the shutdown sequence. When the timeout
  occurs, a System.exit(1) is called.
- Increasing number of retries for failed heartbeats from 3 to 6.
  This prevents short intermittent network failurs from causing the
  containers to be invalidated.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatMonitor.java
+++ b/samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatMonitor.java
@@ -27,10 +27,12 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+
 public class ContainerHeartbeatMonitor {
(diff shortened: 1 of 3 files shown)


samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatMonitor.java
```diff
@@ -50,6 +52,11 @@ public void start() {
     scheduler.scheduleAtFixedRate(() -> {
         ContainerHeartbeatResponse response = containerHeartbeatClient.requestHeartbeat();
         if (!response.isAlive()) {
+          scheduler.schedule(() -> {
+              // On timeout of container shutting down, force exit.
+              LOG.error("Gracefully shutdown timeout expired. Force exiting.");
```
jmakes: Minor:
s/Gracefully/Graceful

PR #376, opened 2017-12-01
Updated Serde related documentation and error messages for High Level API

prateekm:
Updated and clarified the documentation and error messages related to Serdes for Input/Output/PartitionBy streams.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
+++ b/samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
@@ -188,8 +188,9 @@ public interface MessageStream<M> {
    * input to the job.
    * <p>
    * Uses the provided {@link KVSerde} for serialization of keys and values. If the provided {@code serde} is null,
-   * uses the default serde provided via {@link StreamGraph#setDefaultSerde}, which must be a KVSerde.
-   * If no default serde has been provided <b>before</b> calling this method, no-op serdes are used for keys and values.
+   * uses the default serde provided via {@link StreamGraph#setDefaultSerde}, which must be a KVSerde. If the default
+   * serde is not a {@link KVSerde}, a runtime exception will be thrown. If no default serde has been provided
+   * <b>before</b> calling this method, a {@code KVSerde<NoOpSerde, NoOpSerde>} is used.
    * <p>
    * The number of partitions for this intermediate stream is determined as follows:
    * If the stream is an eventual input to a {@link #join}, and the number of partitions for the other stream is known,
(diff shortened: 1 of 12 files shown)


prateekm:
cc @vjagadish1989 @dxichen @jmakes for review

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -200,8 +203,11 @@
...
    * state from the previous version, and changed for the new operator to discard the state from the previous version.
+   * <p>
+   * Unlike {@link #sendTo}, messages with a null key are all sent to partition 0.
    *
-   * @param keyExtractor the {@link Function} to extract the message and partition key from the input message
+   * @param keyExtractor the {@link Function} to extract the message and partition key from the input message.
+   *                     Messages with a null key are all sent to partition 0.
    * @param valueExtractor the {@link Function} to extract the value from the input message
    * @param serde the {@link KVSerde} to use for (de)serializing the key and value.
    * @param id the unique id of this operator in this application
```
dxichen: For the description of this method without the KVSerde, I think we should add the effect of having a null KVSerde rather than just saying it is the same as this method with null. (L: 223)

samza-core/src/main/java/org/apache/samza/operators/stream/IntermediateMessageStreamImpl.java
```diff
@@ -39,12 +41,19 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(IntermediateMessageStreamImpl.class);
   private final OutputStreamImpl<M> outputStream;
+  private final boolean isKeyed;
 
   public IntermediateMessageStreamImpl(StreamGraphImpl graph, InputOperatorSpec<?, M> inputOperatorSpec,
       OutputStreamImpl<M> outputStream) {
     super(graph, (OperatorSpec<?, M>) inputOperatorSpec);
     this.outputStream = outputStream;
+    if (inputOperatorSpec.isKeyed() != outputStream.isKeyed()) {
+      LOGGER.error("Input and output streams for intermediate stream {} aren't keyed consistently. Input: {}, Output: {}",
```
vjagadish1989: Is this error actionable by the user? they should also know that there is an input stream, and output stream corresponding to every intermediate stream?

vjagadish1989 (approved):
Thanks for the improved error messages. Super-helpful! Looks good to me. approved

samza-core/src/main/java/org/apache/samza/operators/stream/IntermediateMessageStreamImpl.java
prateekm: No, this would be a framework error. Just a defensive check for us to help debug in case this happens.

PR #377, opened 2017-12-01
SAMZA-1532: Eventhub connector fix

dxichen:
Key fixes @vjagadish1989 @lhaiesp @srinipunuru 
- Switched Producer source vs destination assumptions in `send`, `register`
- Check `OME.key` if `OME.partitionId` is null for to get partitionId
- Upcoming offset changed the `END_OF_STREAM` rather than `newestOffset` + 1, eventHub returns an error if the offset does not exist in the system
- Made the NewestOffset+1 as upcoming offset, consumer checks if the offset is valid on startup
- Differentiated between streamNames and streamIds in configs, consumer, producer
- Checkpoint table named after job name
- Checkpoint prints better message for invalid key on write

QOL 
- How to ignore integration tests
- Improved logging

EDIT:
- Also added Round Robin producer partitioning

### DIFF
--- a/samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
@@ -76,6 +76,7 @@ public class EventHubSystemProducer implements SystemProducer {
   private static final Object AGGREGATE_METRICS_LOCK = new Object();
 
   public enum PartitioningMethod {
+    ROUND_ROBIN,
     EVENT_HUB_HASHING,
     PARTITION_KEY_AS_PARTITION
   }
@@ -134,6 +135,7 @@ public class EventHubSystemProducer implements SystemProducer {
   public synchronized void start() {
     LOG.debug("Starting system producer.");
 
+    // Create partition senders if required
     if (PartitioningMethod.PARTITION_KEY_AS_PARTITION.equals(partitioningMethod)) {
       // Create all partition senders
       eventHubClients.forEach((streamId, samzaEventHubClient) -> {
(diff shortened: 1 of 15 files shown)


samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -187,48 +183,51 @@ public synchronized void start() {
   }
 
   @Override
-  public synchronized void send(String destination, OutgoingMessageEnvelope envelope) {
+  public synchronized void send(String source, OutgoingMessageEnvelope envelope) {
+    LOG.info(String.format("Trying to send %s", envelope));
```
vjagadish1989: Should be at a lower log level as this would log every single message.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -109,27 +108,23 @@ public EventHubSystemProducer(EventHubConfig config, String systemName,
     this.registry = registry;
     this.systemName = systemName;
     this.partitioningMethod = config.getPartitioningMethod(systemName);
-    this.eventHubClientManagerFactory = eventHubClientManagerFactory;
     this.interceptors = interceptors;
+
+    List<String> streamNames = config.getStreams(systemName);
+    // Create and initiate connections to Event Hubs
```
vjagadish1989: 1. We should be clear on what the semantic meaning of `streamId` and `physicalName` is. In Samza, a `streamId` is the identifier for the stream. It is used to identify a stream (including stream-specific properties) in config. 
2. We are key-ing this map of eventHubClients by their streamId, but looking them up by their physicalName later when send is called. Should be consistent with the identifiers we use.
Take a look at `StreamConfig` to obtain a physicalName for a given streamId.
3. Also, see if we can get rid off this property - "systems.eventhubs.streams". The set of streams should be derivable from other ways by using `StreamConfig.getStreamIds()`. Having said that, I'm fine if we have to keep it for compatibility reasons for now.

vjagadish1989 (approved):
looks good. approved, thanks!

PR #379, opened 2017-12-05
SAMZA-1523 Cleanup table entries before shutting down the processor

navina:
Modified the `TableUtils#deleteProcessorEntity` to provide an option to disable optimistic locking during a call to Azure Table Storage service. 

@sborya @PawasChhokra @nickpan47   Review please?

### DIFF
--- a/samza-azure/src/main/java/org/apache/samza/coordinator/AzureJobCoordinator.java
+++ b/samza-azure/src/main/java/org/apache/samza/coordinator/AzureJobCoordinator.java
@@ -19,22 +19,10 @@
 
 package org.apache.samza.coordinator;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Random;
-import java.util.Set;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Consumer;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.samza.AzureClient;
-import org.apache.samza.config.AzureConfig;
-import org.apache.samza.coordinator.data.BarrierState;
 import org.apache.samza.config.ApplicationConfig;
+import org.apache.samza.config.AzureConfig;
 import org.apache.samza.config.Config;
 import org.apache.samza.config.ConfigException;
 import org.apache.samza.config.JobConfig;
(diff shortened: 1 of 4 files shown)


samza-azure/src/main/java/org/apache/samza/coordinator/AzureJobCoordinator.java
```diff
@@ -443,18 +440,24 @@ private void onNewJobModelAvailable(final String nextJMVersion) {
...
-      table.deleteProcessorEntity(INITIAL_STATE, processorId);
-    }
+    String prevVersion = currentJMVersion.get();
 
-    //Start heartbeating to new entry only when barrier reached.
+    //Start heart-beating to new entry only when barrier reached.
     //Changing the current job model version enables that since we are heartbeating to a row identified by the current job model version.
     currentJMVersion.getAndSet(nextJMVersion);
 
+    // Delete previous value
```
sborya: Not directly related to these changes, but it would be nicer if this functionality was abstracted into a separate class (AzureUtils or something) instead of bleeding out all these details. Something like:
azureUtil.cleanupVersion(pervVersion()
azureUtil.cleanupInitialState() and so on.

navina: Yeah I agree. I didn't want to make more refactoring in this patch and just focus on the bug here. I am working on another patch with refactoring and code cleanup. There is a lot of it!

navina:
Thanks for the review @sborya  !

samza-azure/src/main/java/org/apache/samza/coordinator/AzureJobCoordinator.java
```diff
@@ -443,18 +440,24 @@ private void onNewJobModelAvailable(final String nextJMVersion) {
...
+    ProcessorEntity entity = table.getEntity(prevVersion, processorId);
+    if (entity != null) {
+      entity.setEtag("*");
+      table.deleteProcessorEntity(entity);
```
shanthoosh: Please correct me if i'm wrong here.

The heart beat thread block could still be executed before this block gets triggered trying to delete the azure table row resulting in the same exception mentioned in the JIRA. Since there're no synchronization/communication between these two regions(thread), I think there're no ordering guarantees.

I think it's much simpler to remove deletes from all the followers and do the GC only on the leader itself based upon a processor liveness timeout configuration(similar to zk session timeout). The followers will simply join with creating an entry and will not delete on exit(leader will do the GC(delete all the processors from previous generations) on each job model generation or once when a leader gets elected).

If it's essential it could be changed as a part of this JIRA or add a todo with JIRA for later(since it happens during stream-processor shutdown).

navina: @shanthoosh You are right in that there are no ordering guarantees here. Setting the etag to "*" only forces deletion irrespective of another thread accessing the table. So, it doesn't solve the synchronization problem. For now, it simply prevents the processor from throwing up exceptions due to locking when it is shutting down. 

> I think it's much simpler to remove deletes from all the followers and do the GC only on the leader itself based upon a processor liveness timeout configuration(similar to zk session timeout).
Yes. That is the solution I would like to see as well. However, given how we have partitioned the tables today, it seems like a very expensive operation. I will create a JIRA as a follow-up for this to fix the synchronization issues and clean-up of the tables.

PR #380, opened 2017-12-06
SAMZA-1528: Change ClusterResourceManager to use the async NMClient 

vjagadish1989:
- Rewrite container handling to be asynchronous
- Verified various failure scenarios using Unit tests, and deployments of a local Samza job.


vjagadish1989:
@jmakes @nickpan47 @prateekm for review.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -383,6 +394,50 @@ public void onResourcesCompleted(List<SamzaResourceStatus> resourceStatuses) {
...
+  @Override
+  public void onStreamProcessorLaunchFailure(SamzaResource resource, Throwable t) {
+    log.error("Got a launch failure for SamzaResource {} with exception {}", resource, t);
+    // 1. Release resources for containers that failed back to YARN
+    log.info("Releasing unstartable container {}", resource.getResourceID());
+    clusterResourceManager.releaseResources(resource);
+
+    // 2. Obtain the Samza container Id for the pending container on this resource.
+    String containerId = getPendingContainerId(resource.getResourceID());
+    log.info("Failed container ID: {} for resourceId: {}", containerId, resource.getResourceID());
```
xinyuiscool: Do we also need to remove the containerId from the state.pendingContainers?

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnClusterResourceManager.java
```diff
@@ -262,7 +280,7 @@ public void releaseResources(SamzaResource resource) {
    */
 
   @Override
-  public void launchStreamProcessor(SamzaResource resource, CommandBuilder builder) throws SamzaContainerLaunchException {
+  public void launchStreamProcessor(SamzaResource resource, CommandBuilder builder)  {
```
jmakes: Need to poke around a bit more, but my gut feel is that we need a try-catch for this entire method, or even higher in the call stack to ensure that any exception in container launch causes onStreamProcessorLaunchFailure(). Also documentation on the abstract method javadoc, that no exceptions should be thrown in implementations of this method and instead the failure callback should be called. 

OR

AbstractContainerAllocator should catch all (client side) exceptions and invoke the failure callback directly.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
jmakes: @xinyuiscool github isn't automatically expanding the changes in YarnClusterManager. Expand that diff and you should see the callback invocation.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterResourceManager.java
```diff
@@ -105,15 +105,21 @@ public ClusterResourceManager(Callback callback) {
   public abstract void releaseResources(SamzaResource resource);
 
   /***
-   * Requests the launch of a StreamProcessor with the specified context on the resource.
+   * Requests the launch of a StreamProcessor with the specified context on the resource asynchronously.
```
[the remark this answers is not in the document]
vjagadish1989: Thanks for taking a look Jake!
1. An instance of the `ClusterResourceManager` is always created by passing a `callback`. This is why it's an abstract class.

2. Each `SamzaResource` is a unique identifier for a request. A success is indicated by invoking callback.onSuccess(Resource) and callback.onFailure(Resource, exception) right?
Are you saying that it's not sufficient and proposing a per-method callback or a future instead?

3. The current design is heavily mirrors the design of Hadoop's `NMClientAsync` and `AMRMClientAsync` for the NM and RM respectively.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
vjagadish1989: Thanks Jake! This behavior is annoying in github.

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnClusterResourceManager.java
vjagadish1989: Good point. The current patch catches all sources of exceptions in the places they occur, and invokes the callbacks (wherever the sync-client threw a `SamzaContainerLaunchException`).

But, you're right - We need to be defensive, and I'd be +1 to adding an uber try-catch for safety. I'll also update the java-docs.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterResourceManager.java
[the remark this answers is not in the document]
jmakes: Yes, I was suggesting that having the callback as a method parameter is more flexible. That way they can customize how the callbacks are handled. e.g. as a caller of this method, I may want to monitor the success/failure in order to update my own state and then pass the result success/failure on to some delegate.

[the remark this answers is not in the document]
vjagadish1989: Here's how I view the requirements for this design:
1. Provide a means to start a `StreamProcessor` on a resource.
2. Provide a way to know if request [1] was successful and the container started.
3. Provide a way to know if request [1] was unsuccessful, and if so, what the error was.
Additionally, It'd be nice to integrate with Yarn (and Mesos later) with minimal effort - since, these are the most popular cluster managers and should serve as good validations that our API is reasonable.

>>  I was suggesting that having the callback as a method parameter is more flexible. That way they can customize how the callbacks are handled. e.g. as a caller of this method, I may want to monitor the success/failure in order to update my own state and then pass the result success/failure on to some delegate.

Agreed on the flexibility. Wouldn't this mean that we will have to track the per-method callback state in `ClusterResourceManager` and `ContainerProcessManager` - just to adapt the APIs of the underlying ClusterManager to this model? 

Given the above requirements, and from my initial analysis of Yarn and Mesos's APIs, I was not convinced that this was absolutely necessary - since, both of them don't support it.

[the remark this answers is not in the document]
jmakes: If you feel strongly about this, I won't stand in the way. I just find it confusing (and expect it will be hard to maintain the code) when the request/response flow is:
Request: AbstractContainerAllocator -> ClusterManager.launchStreamProcessor()
Response: ClusterManager -> ClusterResourceManager.onStreamProcessorLaunch*()

How would an engineer implementing a new ClusterManager (e.g. Mesos) know which object to invoke the callback on?

That would be like you asking me to make a code change and then expecting that I'll tell Yi when I've finished your request.

PR #383, opened 2017-12-12
SAMZA-1502: Make AllSspToSingleTaskGrouper work with Yarn and ZK JobCoordinator

atoomula:
Sending a fresh review as I lost the earlier diffs. This is the new approach that we discussed by adding the processor list in the config and passing it to grouper.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
@@ -19,35 +19,40 @@
 
 package org.apache.samza.container.grouper.stream;
 
-import org.apache.samza.SamzaException;
-import org.apache.samza.config.Config;
-import org.apache.samza.config.JobConfig;
-import org.apache.samza.container.TaskName;
-import org.apache.samza.system.SystemStreamPartition;
-
-import java.util.Collections;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-/**
- * AllSspToSingleTaskGrouper, as the name suggests, assigns all partitions to be consumed by a single TaskInstance
- * This is useful, in case of using load-balanced consumers like the new Kafka consumer, Samza doesn't control the
- * partitions being consumed by a task. Hence, it is assumed that there is only 1 task that processes all messages,
- * irrespective of which partition it belongs to.
- * This also implies that container and tasks are synonymous when this grouper is used. Taskname(s) has to be globally
- * unique within a given job.
- *
- * Note: This grouper does not take in broadcast streams yet.
- */
-class AllSspToSingleTaskGrouper implements SystemStreamPartitionGrouper {
-  private final int containerId;
+import org.apache.samza.SamzaException;
+import org.apache.samza.config.Config;
+import org.apache.samza.config.ConfigException;
+import org.apache.samza.config.JobConfig;
+import org.apache.samza.config.TaskConfigJava;
+import org.apache.samza.container.TaskName;
+import org.apache.samza.system.SystemStreamPartition;
 
-  public AllSspToSingleTaskGrouper(int containerId) {
-    this.containerId = containerId;
+
+/**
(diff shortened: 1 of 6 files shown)


samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -104,7 +107,15 @@ object JobModelManager extends Logging {
 
     val streamMetadataCache = new StreamMetadataCache(systemAdmins = systemAdmins, cacheTTLms = 0)
     val previousChangelogPartitionMapping = changelogManager.readChangeLogPartitionMapping()
-    val jobModelManager = getJobModelManager(config, previousChangelogPartitionMapping, localityManager, streamMetadataCache, null)
+
+    val processorList = new ListBuffer[String]()
+    val containerCount = new JobConfig(config).getContainerCount
+    for (i <- 0 to containerCount) {
```
shanthoosh: I think you need to use `until` here. 
`to` is inclusive of both the ranges `[from, to]`

samza-core/src/test/java/org/apache/samza/container/grouper/stream/TestAllSspToSingleTaskGrouper.java
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.samza.config.MapConfig;
+import org.apache.samza.container.TaskName;
+import org.apache.samza.system.SystemStreamPartition;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+
+public class TestAllSspToSingleTaskGrouper {
+  private SystemStreamPartition aa0 = new SystemStreamPartition("SystemA", "StreamA", new Partition(0));
```
shanthoosh: It's better to name this as sytemStreamPartition0/ssp0  or some other meaningful name rather than aa*.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AllSspToSingleTaskGrouperFactory.java
```diff
@@ -19,51 +19,66 @@
...
 public class AllSspToSingleTaskGrouperFactory implements SystemStreamPartitionGrouperFactory {
   @Override
   public SystemStreamPartitionGrouper getSystemStreamPartitionGrouper(Config config) {
-    return new AllSspToSingleTaskGrouper(config.getInt(JobConfig.PROCESSOR_ID()));
+    if (!(new TaskConfigJava(config).getBroadcastSystemStreams().isEmpty())) {
+      throw new ConfigException("The job configured with AllSspToSingleTaskGrouper cannot have broadcast streams.");
+    }
+
+    String processors = config.get(JobConfig.PROCESSOR_LIST());
+    List<String> processorList = Arrays.asList(processors.split(","));
```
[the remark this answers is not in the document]
atoomula: Added the check and throw an exception if the list is empty.

samza-core/src/test/java/org/apache/samza/container/grouper/stream/TestAllSspToSingleTaskGrouper.java
atoomula: This mirrors system(a) stream(a/b) partition(0-3).

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
atoomula: Yup. Realized this while testing and fixed it.

shanthoosh:
@atoomula 
I think you have to inject processorList into config from ZkJobCoordinator (container.count configuration doesn't apply in standalone) before generating the job model.

Existing change injects processList only for Yarn and PassThroughJobCoordinator scenarios.

atoomula:
@shanthoosh zkJobCoordinator already calls readJobModel with valid list of containerIds. In readJobModel, we just inject this list of containerIds into config. This effort is to build the processor list which is absent for Yarn and PassthruJC. ZkJC already builds the processor list. Let me know if I'm missing something.

shanthoosh:
@atoomula 
Makes sense. Looks good to me.

Let's document the it's mandatory to use AllSSPToSingleTaskGrouper configuration for kinesis.

PR #385, opened 2017-12-12
SAMZA-1534: Fix the visualization in job graph with the new PartitionBy Op

xinyuiscool:
Seems the stream and the partitionBy op has the same id. So in rendering I added the stream as the id for the node. Also resolved the run.id collision issue.


vjagadish1989 (approved):
approved. looks good. ship it!

PR #386, opened 2017-12-12
Support for UDFs in where clauses

srinipunuru:
The existing version of the udf implementation doesn't seem to support udfs in the where clauses because the Type of the object returned is "ANY" and when you do a 
`select * from kafka.topic where regexMatch('.*foo', Name)` it fails in the query validation, because calcite doesn't know the type of regexMatch. 

To solve the problem, We made the scalarUdf generic with a strongly typed return type.

https://issues.apache.org/jira/browse/SAMZA-1535

This PR can be merged into trunk not the 0.14.


xinyuiscool:
LGTM

PR #388, opened 2017-12-15
SAMZA-1537: StreamAppender can deadlock due to locks held by Kafka an…

jmakes:
…d Log4j


jmakes:
@nickpan47 @vjagadish1989 @prateekm for review

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
```diff
@@ -206,13 +257,53 @@ protected void setupSystem() {
...
+
+            OutgoingMessageEnvelope outgoingMessageEnvelope =
+                new OutgoingMessageEnvelope(systemStream, keyBytes, serializedLogEvent);
+            systemProducer.send(SOURCE, outgoingMessageEnvelope);
+
+            if (Thread.interrupted()) {
+              break;
+            }
+          } catch (InterruptedException e) {
+            break; // Someone wants us to stop, so stop.
```
vjagadish1989: Can just invoke a `Thread.currentThread().interrupt()` and let the loop's while condition be, `while(!Thread.currentThread().isInterrupted())` instead of a `while(true)`

samza-log4j/src/test/java/org/apache/samza/logging/log4j/TestStreamAppender.java
```diff
@@ -104,26 +108,115 @@ public void testSystemProducerAppenderInAM() {
...
+    // Wait for messages
+    assertTrue("Thread did not send all messages. Count: " + allMessagesSent.getCount(),
+        allMessagesSent.await(60, TimeUnit.SECONDS));
+    assertEquals(expectedMessagesSent, MockSystemProducer.messagesReceived.size());
+  }
+
+  private void logAndVerifyMessages(List<String> messages) throws InterruptedException {
+    // Set up latch
+    final CountDownLatch allMessagesSent = new CountDownLatch(messages.size());
+    MockSystemProducer.listeners.add((source, envelope) -> allMessagesSent.countDown());
```
vjagadish1989: Nice set of unit tests! thanks!

vjagadish1989 (approved):
looks good to me. approved.

thank you. 

added some f/b from my review.

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
```diff
@@ -110,16 +125,41 @@ public void append(LoggingEvent event) {
...
+          }
+          metrics.bufferFillPct.set(Math.round(100 * logQueue.size() / DEFAULT_QUEUE_SIZE));
         }
       } catch (Exception e) {
         System.err.println("[StreamAppender] Error sending log message:");
         e.printStackTrace();
       } finally {
         recursiveCall.set(false);
       }
+    } else if (metrics != null) {
```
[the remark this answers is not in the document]
jmakes: Yes, because metrics are initialized in setupSystem, which may not have been called for this code path, but has definitely been called for the others.

vjagadish1989 (approved):
approved. (again) ship it!

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
```diff
@@ -206,13 +259,51 @@ protected void setupSystem() {
 
     setSerde(log4jSystemConfig, systemName, streamName);
 
-    systemProducer = systemFactory.getProducer(systemName, config, new MetricsRegistryMap());
+    systemProducer = systemFactory.getProducer(systemName, config, metricsRegistry);
```
nickpan47: A quick question: here we are creating a new producer for StreamAppender, which is not part of the SystemProducers. Hence, not controlled by collector.flush(). For logs, I assume that we would like systemProducer to be highly available, more than no-loss or order-guarantee, as we configure for input/output systems. Can we auto-configure this systemProducer to drop producer errors?

jmakes: Great idea:
https://issues.apache.org/jira/browse/SAMZA-1540

PR #390, opened 2017-12-20
SAMZA-1539: KafkaProducer potential hang on close() when task.drop.pr…

jmakes:
…oducer.errors==true

### DIFF
--- a/samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
+++ b/samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
@@ -134,6 +134,7 @@ class KafkaSystemProducer(systemName: String,
       if (currentProducer == null) {
         if (dropProducerExceptions) {
           // No producer to flush, but we're ignoring exceptions so just return.
+          warn("Skipping flush because the Kafka producer is null.")
           metrics.flushFailed.inc
           return
         }
--- a/samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
+++ b/samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
@@ -28,7 +28,6 @@ import org.apache.kafka.clients.producer.Producer
 import org.apache.kafka.clients.producer.ProducerRecord
 import org.apache.kafka.clients.producer.RecordMetadata
 import org.apache.kafka.common.PartitionInfo
-import org.apache.kafka.common.errors.SerializationException
 import org.apache.samza.system.OutgoingMessageEnvelope
 import org.apache.samza.system.SystemProducer
 import org.apache.samza.system.SystemProducerException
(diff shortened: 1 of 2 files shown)


jmakes:
@prateekm @nickpan47 @vjagadish1989 for review

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -172,49 +179,56 @@ class KafkaSystemProducer(systemName: String,
...
-        }
-      }
-    } else {
-      // If there is an exception in the callback, it means that the Kafka producer has exhausted the max-retries
-      // Close producer to ensure messages queued in-flight are not sent and hence, avoid re-ordering
-      // This works because there is only 1 callback thread and no sends can complete until the callback returns.
-      if (isFatalException) {
-        fatalException.compareAndSet(null, producerException)
+      // Prevent each callback from closing and nulling producer for the same failure.
+      if (currentProducer == producerRef.get()) {
```
fredji97: If both currentProducer and producerRef.get() are null here, it can have null pointer exception. I would recommend to have a null check to make the code more robust.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -46,32 +45,30 @@ class KafkaSystemProducer(systemName: String,
...
   def stop() {
     info("Stopping producer for system: " + this.systemName)
 
-    // stop() should not happen often so no need to optimize locking
-    producerLock.synchronized {
-      try {
-        if (producer != null) {
-          producer.close  // Also performs the equivalent of a flush()
-        }
+    stopped = true
```
fredji97: would it make more sense to put stopped = true after stop is done?

jmakes: No. The flag is used to prevent subsequent producer creations. That should happen immediately when stop() is invoked.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
jmakes: 1. The java doc specifies that currentProducer must not be null
2. If currentProducer was null, the right thing to do is throw an exception because we can't properly handle the exception, so then the debate is NPE or other exception. I don't see the value.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
fredji97: Yeah, the logic makes sense. It might be better to call it stopTriggered or something else instead of stopped, to avoid the confusion, but that's minor and optional.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemProducer.scala
```diff
@@ -125,18 +119,24 @@ class KafkaSystemProducer(systemName: String,
...
-
   def flush(source: String) {
     updateTimer(metrics.flushNs) {
       metrics.flushes.inc
 
-      val currentProducer = producer
+      val currentProducer = producerRef.get()
       if (currentProducer == null) {
+        if (dropProducerExceptions) {
+          // No producer to flush, but we're ignoring exceptions so just return.
```
sborya: do we need a WARN here?

jmakes: Could be useful. Added it.

PR #391, opened 2017-12-20
Merge script improvement - use colons instead of semicolons

jmakes:


jmakes:
Hey committers, what say you?
Our convention seems to be colons. How about we make it Python official?
@xinyuiscool @nickpan47 @vjagadish1989 @sborya @prateekm

vjagadish1989 (approved):
approved, looks good to me

vjagadish1989:
Would also be nice to do the same for our hello-samza merge script too consistently.

jmakes:
@vjagadish1989 this was already done on the hello-samza merge script at the same time the script was added to that repo. 
In fact, this PR was created by diff'ing the 2 scripts. :-)

PR #397, opened 2018-01-02
SAMZA-1548; Add start() and stop() to SystemAdmin

lindong28:
This patch adds start() and stop() to SystemAdmin interface. This can be useful for e.g. kafka.admin.AdminClient which needs to be started before it can be used.

Since we add this method in interface and expect AdminClient to be stateful and probably has its own thread, there will be higher cost to instantiate a new SystemAdmin. Thus we probably want to re-use the SystemAdmin instances instead of creating SystemAdmin on demand when needed. Therefore, this patch also adds SystemAdmins class to help manage a map from system to SystemAdmin, similar to the existing SystemProducers class in Samza.


lindong28:
@nickpan47 @prateekm Do you know who can review this patch? Thanks!

sborya:
Could you please add description into either the ticket or the PR (or both) with what you are trying to achieve. 
Thanks!

lindong28:
@sborya Sure! I have updated the description for both the pull request and the JIRA ticket.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -28,6 +28,17 @@
  * utility methods that Samza needs in order to interact with a system.
  */
 public interface SystemAdmin {
+
+  /*
+   * Connect to the underlying system and prepare to begin serving requests
```
prateekm: Update documentation. Don't need to mention connecting and serving requests, just say something generic like start the SystemAdmin. Same for stop.

samza-azure/src/main/java/org/apache/samza/coordinator/AzureJobCoordinator.java
```diff
@@ -90,6 +93,7 @@
   private RenewLeaseScheduler renewLease;
   private LeaderBarrierCompleteScheduler leaderBarrierScheduler;
   private StreamMetadataCache streamMetadataCache = null;
+  private SystemAdmins systemAdmins = null;
```
prateekm: Where is this used?

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamSystemConsumer.java
```diff
@@ -124,6 +124,7 @@ public void start() {
     }
     log.info("Starting coordinator stream system consumer.");
     systemConsumer.start();
+    systemAdmin.start();
```
prateekm: Should be started and stopped by the object that passed in this SystemAdmin, not this object.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamSystemProducer.java
```diff
@@ -82,6 +82,7 @@ public void start() {
     }
     log.info("Starting coordinator stream producer.");
     systemProducer.start();
+    systemAdmin.start();
```
prateekm: Should be started and stopped by the owner of this SystemAdmin, not this object.

samza-core/src/main/java/org/apache/samza/storage/StorageRecovery.java
```diff
@@ -61,13 +62,12 @@
...
-  private HashMap<String, SystemFactory> systemFactories = new HashMap<String, SystemFactory>();
-  private HashMap<String, SystemAdmin> systemAdmins = new HashMap<String, SystemAdmin>();
-  private Map<String, ContainerModel> containers = new HashMap<String, ContainerModel>();
-  private List<TaskStorageManager> taskStorageManagers = new ArrayList<TaskStorageManager>();
+  private HashMap<String, SystemStream> changeLogSystemStreams = new HashMap<>();
+  private HashMap<String, StorageEngineFactory<?, ?>> storageEngineFactories = new HashMap<>();
+  private Map<String, ContainerModel> containers = new HashMap<>();
+  private List<TaskStorageManager> taskStorageManagers = new ArrayList<>();
   private Logger log = LoggerFactory.getLogger(StorageRecovery.class);
+  private SystemAdmins systemAdmins = null;
```
prateekm: If this is started and stopped in run, do we need to save it as a field?

samza-core/src/main/java/org/apache/samza/storage/StorageRecovery.java
```diff
@@ -207,7 +188,9 @@ private void getChangeLogMaxPartitionNumber() {
    */
   @SuppressWarnings({ "unchecked", "rawtypes" })
   private void getTaskStorageManagers() {
-    StreamMetadataCache streamMetadataCache = new StreamMetadataCache(Util.javaMapAsScalaMap(systemAdmins), 5000, SystemClock.instance());
+    Map<String, SystemAdmin> systemAdminMap = new JavaSystemConfig(jobConfig).getSystemAdmins();
+    StreamMetadataCache streamMetadataCache = new StreamMetadataCache(Util.javaMapAsScalaMap(systemAdminMap), 5000, SystemClock.instance());
+    systemAdmins = new SystemAdmins(Util.javaMapAsScalaMap(systemAdminMap));
```
prateekm: Seems weird that this is set as a side effect in some unrelated method. Can this happen in the constructor?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -120,13 +126,15 @@
         LOG.error("Received exception from in JobCoordinator Processing!", throwable);
         stop();
       });
+    Map<String, SystemAdmin> systemAdminMap = new JavaSystemConfig(config).getSystemAdmins();
+    systemAdmins = new SystemAdmins(Util.javaMapAsScalaMap(systemAdminMap));
```
prateekm: If most usages of this Map constructor parameter are in Java, maybe cleaner to make it a java.util.Map instead?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -334,7 +335,7 @@ object SamzaContainer extends Logging {
 
     val chooserFactory = Util.getObj[MessageChooserFactory](chooserFactoryClassName)
 
-    val chooser = DefaultChooser(inputStreamMetadata, chooserFactory, config, samzaContainerMetrics.registry, systemAdmins)
+    val chooser = DefaultChooser(inputStreamMetadata, chooserFactory, config, samzaContainerMetrics.registry, systemAdminMap)
```
prateekm: Not sure why we're passing the systemAdminMap in some places and systemAdmins in others. How bad is it to remove SystemAdmins?

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -242,9 +249,8 @@ private JobModelManager buildJobModelManager(Config coordinatorSystemConfig, Met
     return jobModelManager;
   }
 
-  private StreamPartitionCountMonitor getPartitionCountMonitor(Config config) {
-    Map<String, SystemAdmin> systemAdmins = new JavaSystemConfig(config).getSystemAdmins();
-    StreamMetadataCache streamMetadata = new StreamMetadataCache(Util.javaMapAsScalaMap(systemAdmins), 0, SystemClock.instance());
+  private StreamPartitionCountMonitor getPartitionCountMonitor(Config config, Map<String, SystemAdmin> systemAdminMap) {
+    StreamMetadataCache streamMetadata = new StreamMetadataCache(Util.javaMapAsScalaMap(systemAdminMap), 0, SystemClock.instance());
```
prateekm: Should this take SystemAdmins instead of systemAdminMap?

samza-core/src/main/java/org/apache/samza/standalone/PassthroughJobCoordinator.java
```diff
@@ -112,20 +110,10 @@ public void setListener(JobCoordinatorListener listener) {
...
-        throw new SamzaException(String.format("A stream uses system %s, which is missing from the configuration.", systemName));
-      }
-      SystemFactory systemFactory = Util.<SystemFactory>getObj(systemFactoryClassName);
-      systemAdmins.put(systemName, systemFactory.getAdmin(systemName, this.config));
-    }
-
-    StreamMetadataCache streamMetadataCache = new StreamMetadataCache(
-        Util.<String, SystemAdmin>javaMapAsScalaMap(systemAdmins), 5000, SystemClock.instance());
-
+    Map<String, SystemAdmin> systemAdminMap = systemConfig.getSystemAdmins();
```
prateekm: This is in a lot of places. Would it be cleaner to add a helper/constructor in SystemAdmins that does this?

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
lindong28: This documentation was based on the documentation of SystemConsumer.start(), which says "Tells the SystemConsumer to connect to the underlying system, and prepare to begin serving messages when poll is invoked".

Sure, I updated the documentation of start() to "Start up API for bringing up a single instance of SystemAdmin.". And I updated the documentation of stop() to "Shutdown API for shutting down a single instance of the SystemAdmin". Does this look good?

samza-azure/src/main/java/org/apache/samza/coordinator/AzureJobCoordinator.java
lindong28: I think it is used in start() and stop().

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
lindong28: I think it is better to use SystemAdmins in StreamMetadataCache. Since it is mostly cosmetic change to the constructor of `StreamMetadataCache`, I have not done this in the current patch to make the review easier. Do you prefer me to make the change here? I can do it.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamSystemConsumer.java
lindong28: `systemAdmin` is started here because the `systemAdmin` is instantiated in `CoordinatorStreamSystemFactory.getCoordinatorStreamSystemConsumer()`. The caller of this method actually does not know about SystemAdmin. How about I change the constructor of `CoordinatorStreamSystemConsumer` to be `CoordinatorStreamSystemConsumer(config: Config, registry: MetricsRegistry)` and remove the existing class `CoordinatorStreamSystemFactory`?

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamSystemProducer.java
lindong28: This is started here for the same reason as `CoordinatorStreamSystemConsumer`. I think it may be reasonable to remove `CoordinatorStreamSystemFactory` and just instantiate SystemAdmin in CoordinatorStreamSystemProducer directly. Does this sound OK?

samza-core/src/main/java/org/apache/samza/storage/StorageRecovery.java
lindong28: The reason it is kept as a field is that this SystemAdmin is instantiated in `getTaskStorageManagers()` and it is shared by multiple `taskStorageManager`. I don't have a good way to avoid saving it as a field. What do you think?

samza-core/src/main/java/org/apache/samza/storage/StorageRecovery.java
lindong28: Good point. I have updated the patch to instantiate SystemAdmin in the constructor.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
lindong28: Good point. I have made the change as suggested.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
lindong28: I can replace SystemAdminMap with SystemAdmins everywhere. It just requires more refactor and cosmetic change in this patch. Do you prefer me to do that?

samza-core/src/main/java/org/apache/samza/standalone/PassthroughJobCoordinator.java
lindong28: Not sure if I completely understand your comments. But I have updated the constructor of SystemAdmins to simplify the code. Can you see if it addresses the problem?

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
lindong28: Fixed as suggestd.

samza-azure/src/main/java/org/apache/samza/coordinator/AzureJobCoordinator.java
lindong28: Fixed as suggested.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
lindong28: Fixed as suggested.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamSystemProducer.java
lindong28: CoordinatorStreamSystemProducer constructor now takes config as the parameter.

samza-core/src/main/java/org/apache/samza/standalone/PassthroughJobCoordinator.java
lindong28: Fixed as discussed.

samza-core/src/main/java/org/apache/samza/storage/StorageRecovery.java
lindong28: Fixed as suggested.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
lindong28: Fixed as discussed.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamSystemConsumer.java
lindong28: CoordinatorStreamSystemConsumer constructor now takes config as the parameter.

lindong28:
@prateekm Thanks much for taking time to review this patch! I have updated the patch to address all comments as we discussed. There is a lot of refactor. The check-all.sh has passed on my local machine.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -153,7 +152,8 @@ public ClusterBasedJobCoordinator(Config coordinatorSystemConfig) {
     config = jobModelManager.jobModel().getConfig();
     hasDurableStores = new StorageConfig(config).hasDurableStores();
     state = new SamzaApplicationState(jobModelManager);
-    partitionMonitor = getPartitionCountMonitor(config);
+    systemAdmins = new SystemAdmins(config);
```
prateekm: Let's add the same comment as above: Should be started/stopped before ...

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -50,10 +50,12 @@
   private static final Logger log = LoggerFactory.getLogger(AbstractApplicationRunner.class);
 
   private final StreamManager streamManager;
+  private final SystemAdmins systemAdmins;
 
   public AbstractApplicationRunner(Config config) {
     super(config);
-    this.streamManager = new StreamManager(new JavaSystemConfig(config).getSystemAdmins());
+    systemAdmins = new SystemAdmins(config);
```
prateekm: Minor: Use `this.` consistently in the class.

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -72,6 +72,7 @@ public void runTask() {
 
   @Override
   public void run(StreamApplication streamApp) {
+    super.run(streamApp);
```
prateekm: Kill is never called here?

samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Map;
+import org.apache.samza.SamzaException;
+import org.apache.samza.config.Config;
+import org.apache.samza.config.JavaSystemConfig;
+
+
+public class SystemAdmins {
+  private final Map<String, SystemAdmin> systemAdminMap;
+
+  public Map<String, SystemAdmin> systemAdminMap() {
```
prateekm: Is this used anywhere? If not, let's remove. If you keep it, rename as a getter 'getSystemAdminsMap' and move below the constructors.

samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
```diff
@@ -0,0 +1,66 @@
...
+    return systemAdminMap;
+  }
+
+  public SystemAdmins(Config config) {
+    JavaSystemConfig systemConfig = new JavaSystemConfig(config);
+    this.systemAdminMap = systemConfig.getSystemAdmins();
+  }
+
+  public SystemAdmins() {
```
prateekm: Why is this constructor required? Can we remove it?

samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  // Used only for test
+  public SystemAdmins(Map<String, SystemAdmin> systemAdminMap) {
+    this.systemAdminMap = systemAdminMap;
+  }
+
+  public void start() {
+    for (SystemAdmin systemAdmin: systemAdminMap.values())
+      systemAdmin.start();
```
prateekm: Minor: Braces around single line loops and conditions in this class, just in case.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -141,7 +142,7 @@ class OffsetManager(
    * SystemAdmins that are used to get next offsets from last checkpointed
    * offsets. Map is from system name to SystemAdmin class for the system.
    */
-  val systemAdmins: Map[String, SystemAdmin] = Map(),
+  val systemAdmins: SystemAdmins = new SystemAdmins(),
```
prateekm: Does this work correctly? Looks like OffsetManager#apply is explicitly passing a null SystemAdmins if it's not passed one. Maybe change the default there to a new SystemAdmins() instead of null.

samza-core/src/main/scala/org/apache/samza/coordinator/stream/CoordinatorStreamSystemFactory.scala
```diff
@@ -1,50 +0,0 @@
...
-import org.apache.samza.system.{SystemFactory, SystemStream}
-import org.apache.samza.util.Util
-
-/**
- * A helper class that does wiring for CoordinatorStreamSystemConsumer and
- * CoordinatorStreamSystemProducer. This factory should only be used in
- * situations where the underlying SystemConsumer/SystemProducer does not
- * exist.
- */
-class CoordinatorStreamSystemFactory {
```
prateekm: Thanks for cleaning this up.

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -19,27 +19,10 @@
...
 
-import org.apache.samza.config.Config
-import org.apache.samza.util.{Util, Logging, Clock, SystemClock}
+import org.apache.samza.util.{Logging, Clock, SystemClock}
 import org.apache.samza.SamzaException
 import scala.collection.JavaConverters._
-import org.apache.samza.config.SystemConfig.Config2System
 
-object StreamMetadataCache {
-  def apply(cacheTtlMs: Int = 5000, config: Config): StreamMetadataCache = {
```
prateekm: Thanks for cleaning this up.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -256,8 +256,8 @@ class BootstrappingChooser(
       null
     }
     val offsetToCheck = if (systemStreamPartitionMetadata == null) {
-      // Use null for offsetToCheck in cases where the partition metadata was 
-      // null. A null partition metadata implies that the stream is not a 
+      // Use null for offsetToCheck in cases where the partition metadata was
```
prateekm: Minor: Prefer reverting formatting changes to maintain comment history.

samza-core/src/main/scala/org/apache/samza/system/chooser/DefaultChooser.scala
```diff
@@ -251,7 +255,7 @@ class DefaultChooser(
    * Defines a mapping from SystemStream name to SystemAdmin.
    * This is useful for determining if a bootstrap SystemStream is caught up.
    */
-  systemAdmins: Map[String, SystemAdmin] = Map()) extends MessageChooser with Logging {
+  systemAdmins: SystemAdmins = new SystemAdmins()) extends MessageChooser with Logging {
```
prateekm: Are these empty default SystemAdmins there for tests? If so, I'd prefer to just make them required and explicitly pass on in tests.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -28,6 +28,17 @@
  * utility methods that Samza needs in order to interact with a system.
  */
 public interface SystemAdmin {
+
+  /*
+   * Start SystemAdmin
```
prateekm: "Start/Stop this SystemAdmin"

lindong28: Sure. I added "this" to the Java doc of start() and stop().

samza-azure/src/main/java/org/apache/samza/coordinator/AzureJobCoordinator.java
```diff
@@ -124,9 +127,11 @@ public AzureJobCoordinator(Config config) {
 
   @Override
   public void start() {
-
     LOG.info("Starting Azure job coordinator.");
-    streamMetadataCache = StreamMetadataCache.apply(METADATA_CACHE_TTL_MS, config);
+
+    systemAdmins = new SystemAdmins(config);
```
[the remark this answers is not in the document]
lindong28: Sure. I added comment `The systemAdmins should be started before streamMetadataCache can be used. And it should be stopped when coordinator is stopped.`

samza-azure/src/main/java/org/apache/samza/coordinator/AzureJobCoordinator.java
```diff
@@ -124,9 +127,11 @@ public AzureJobCoordinator(Config config) {
...
+    streamMetadataCache = new StreamMetadataCache(systemAdmins, METADATA_CACHE_TTL_MS, SystemClock.instance());
+    systemAdmins.start();
```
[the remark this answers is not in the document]
lindong28: Good point. I have made this change.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
lindong28: Sure. I added comment `The systemAdmins should be started before partitionMonitor can be used. And it should be stopped when this coordinator is stopped.`

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
lindong28: Sure. Fixed as suggested.

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
lindong28: Yeah I think so. `LocalContainerRunner.kill(...)` will throw `UnsupportedOperationException`.

samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
lindong28: It is currently used in an info level log which existed before this patch. I have renamed this method as suggested. I can also remove this method and the info level log if you prefer so.

samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
lindong28: It is used to provide the default value for SystemAdmins in a few constructors, e.g. OffsetManager and BootstrappingChooser. Alternatively I can also just use `new SystemAdmins(new java.util.HashMap[String, SystemAdmin]())`. Not sure which style is cleaner. What do you think?

samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
lindong28: Sure. I have updated the code as suggested.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
lindong28: Good point. I have updated the `OffsetManager#apply` as suggested.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
lindong28: It looks like IntellIj made these changes automatically. However, if I open `Annotate` in IntellIj to check the git history, the owner of this comment is not override to me even with this formatting changes. So it may be OK?

If it is not OK, I will edit the scala file without using IntellIj before updating the patch.

samza-core/src/main/scala/org/apache/samza/system/chooser/DefaultChooser.scala
lindong28: Yes this default value is only used for tests. If we were to make them required, should we also make other parameters required as well? I can make this change if you prefer.

lindong28:
Thanks so much for taking time to review the patch @prateekm.

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
prateekm: Are there any issues with leaving the SystemAdmins running in this case?

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
prateekm: If annotate works then it's fine, thanks for checking.

samza-core/src/main/scala/org/apache/samza/system/chooser/DefaultChooser.scala
prateekm: If it's not too much work, making this param alone required would be fine. I mostly want to get rid of the SystemAdmins default constructor.

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
lindong28: I think there is no issue with leaving SystemAdmins running. It seems reasonable SystemAdmins to keep running because `streamManager.getStreamPartitionCounts()` can potentially be run anytime when the LocalContainerRunner is running. What do you think?

lindong28:
@prateekm I have addressed all comments and rebased the patch onto the latest mater branch. Can you take another look? Thanks!

PR #398, opened 2018-01-03
SAMZA-1542 refactor config classes in samza-core java to hide non-necessary interfaces

fredji97:
With previous inheritance, it exposed everything from Config class, which is a bad practice and is error prone. We are refactoring it to use composition over inheritance make it more robust.


vjagadish1989:
Hey @fredji97 ,
Thanks for this PR. While I'm +1 to your idea, I'm trying to ensure that the open-source codebase doesn't remain in a state where some classes follow inheritance and others follow composition for a long time. Ideally, I would love to make all changes in one-shot throughout the Samza code-base. It need not have to be a single PR but would be nice if you have all necessary PRs ready and if they can be merged in quick succession. cc: @prateekm @nickpan47

PR #404, opened 2018-01-11
SAMZA-1500: Added metrics for RocksDB state store memory usage

prateekm:
Approximate RocksDB memory usage = Configured Block Cache size + MemTable size + Indexes and Bloom Filters size = 
rocksdb.block-cache-size + rocksdb.size-all-mem-tables + rocksdb.estimate-table-readers-mem


prateekm:
cc @vjagadish1989

PR #405, opened 2018-01-12
SAMZA-1556: Adding support for multi level sources in queries

srinipunuru:
Right now Samza SQL supports queries with just two levels i.e. `select * from foo.bar`. But there can be sources that are identified though multiple levels. for e.g. `select * from kafka.clusterName.topicName`. 

This change adds the support for sql queries with sources that have more than two levels.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlSystemStreamConfig.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlSystemStreamConfig.java
@@ -19,8 +19,13 @@
 
 package org.apache.samza.sql.interfaces;
 
+import com.google.common.base.Joiner;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
 import org.apache.commons.lang.Validate;
 import org.apache.samza.config.Config;
+import org.apache.samza.config.MapConfig;
 import org.apache.samza.system.SystemStream;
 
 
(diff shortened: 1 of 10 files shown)


samza-sql/src/main/java/org/apache/samza/sql/planner/QueryPlanner.java
```diff
@@ -71,9 +70,9 @@
 
   private final Map<String, SqlSystemStreamConfig> systemStreamConfigBySource;
   private final Collection<UdfMetadata> udfMetadata;
-  private final Map<SystemStream, RelSchemaProvider> relSchemaProviders;
+  private final Map<String, RelSchemaProvider> relSchemaProviders;
 
-  public QueryPlanner(Map<SystemStream, RelSchemaProvider> relSchemaProviders,
+  public QueryPlanner(Map<String, RelSchemaProvider> relSchemaProviders,
```
misanc: Can you add a comment what is the key of the map "String"

samza-sql/src/main/java/org/apache/samza/sql/planner/QueryPlanner.java
```diff
@@ -85,16 +84,24 @@ public RelRoot plan(String query) {
...
       for (SqlSystemStreamConfig ssc : systemStreamConfigBySource.values()) {
+        SchemaPlus previousLevelSchema = rootSchema;
+        List<String> sourceParts = ssc.getSourceParts();
+        RelSchemaProvider relSchemaProvider = relSchemaProviders.get(ssc.getSource());
 
-        RelSchemaProvider relSchemaProvider = relSchemaProviders.get(ssc.getSystemStream());
-        SchemaPlus systemSchema =
-            systemSchemas.computeIfAbsent(ssc.getSystemName(), s -> rootSchema.add(s, new AbstractSchema()));
-        RelDataType relationalSchema = relSchemaProvider.getRelationalSchema();
+        for (int sourcePartIndex = 0; sourcePartIndex < sourceParts.size() - 1; sourcePartIndex++) {
```
misanc: nit: you can do a normal loop

for(String sourcePart : sourceParts) {

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlSystemStreamConfig.java
```diff
@@ -38,20 +42,43 @@
...
-  public SqlSystemStreamConfig(String systemName, String streamName, Config systemConfig) {
+  private Config config;
+
+  private List<String> sourceParts;
+
+  public SqlSystemStreamConfig(String systemName, String streamName, String source, Config systemConfig) {
+    this(systemName, streamName, source, Arrays.asList(systemName, streamName), systemConfig);
+  }
 
+  public SqlSystemStreamConfig(String systemName, String streamName, String source, List<String> sourceParts,
```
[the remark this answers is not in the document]
srinipunuru: I wanted to keep the logic of converting the source String into source parts in the source resolver plugin. So that plugins can parse however they want based on the business needs. But i do see the problem here that we are parsing the redundant information. Let me take in the sourceParts and create a canonical source String that can be used to lookup the RelSchemaProviders and converters.

vjagadish1989 (approved):
approved. Ship it!

PR #407, opened 2018-01-17
SAMZA-1558: State restore metrics should be duplicated and deprecated…

jmakes:
… to avoid type conflicts


jmakes:
@prateekm for review.

PR #412, opened 2018-01-23
SAMZA-1562: TaskStorageManager should delete any local store it canno…

jmakes:
…t open.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
+++ b/samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
@@ -121,10 +121,17 @@ class TaskStorageManager(
     * @return true if the logged store is valid, false otherwise.
     */
   private def isLoggedStoreValid(storeName: String, loggedStoreDir: File): Boolean = {
-    val changeLogDeleteRetentionInMs = changeLogDeleteRetentionsInMs.getOrElse(storeName,
-                                                                               StorageConfig.DEFAULT_CHANGELOG_DELETE_RETENTION_MS)
-    persistedStores.contains(storeName) && isOffsetFileValid(loggedStoreDir) &&
-      !isStaleLoggedStore(loggedStoreDir, changeLogDeleteRetentionInMs)
+    try {
+      val changeLogDeleteRetentionInMs = changeLogDeleteRetentionsInMs.getOrElse(storeName,
+        StorageConfig.DEFAULT_CHANGELOG_DELETE_RETENTION_MS)
+      persistedStores.contains(storeName) && isOffsetFileValid(loggedStoreDir) &&
+        !isStaleLoggedStore(loggedStoreDir, changeLogDeleteRetentionInMs)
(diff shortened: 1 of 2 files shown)


jmakes:
@prateekm

samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
```diff
@@ -289,11 +296,21 @@ class TaskStorageManager(
...
+          val systemStreamPartition = new SystemStreamPartition(systemStream, partition)
+          val systemConsumer = storeConsumers(storeName)
+          val systemConsumerIterator = new SystemStreamPartitionIterator(systemConsumer, systemStreamPartition)
+          store.restore(systemConsumerIterator)
+        } catch {
+          case e: Exception => {
+            val storePartitionDir = TaskStorageManager.getStorePartitionDir(loggedStoreBaseDir, storeName, taskName)
+            error("Exception restoring logged store %s. Deleting potentially-malformed files from path %s" format(storeName, storePartitionDir), e)
+            System.out.println("TSM removing file " + storePartitionDir.toString)
+            Util.rm(storePartitionDir)
```
prateekm: This is dangerous, since the store open may fail for recoverable reasons. E.g., lock file held by an orphaned container. We can't really tell why the open failed since they all throw a RocksDbException, and relying on exception messages is error-prone. For now, I'd prefer to only fix the OFFSET file.

jmakes: I'm not sure I follow the logic. 

Are you saying it's safer to assume that all exceptions restoring the local store are recoverable than to assume they're unrecoverable?

What's the fix for the unrecoverable case, then? Manual deletion? I'd like to do better.

jmakes: @prateekm what if it just deletes the offset file, rather than the whole directory (same as what GC does)?

A running job (orphan) will recreate the offset file on the next commit. If the offset file is not recreated before the next restore, the store will be deleted at that time. 

It's safer, but doesn't completely eliminate the possibility of deleting a store out from under a running orphan. But if there is an orphan, data integrity is already compromised.

prateekm: > Are you saying it's safer to assume that all exceptions restoring the local store are recoverable
Yeah, IMHO it's safer to leave the store alone until we know for sure it's unrecoverable. I was suggesting that we can fix known-to-be-unrecoverable cases automatically, e.g. the corrupt offset file, and the rest will require manual intervention (they're not common). 

I don't have context on what exactly the corrupt OFFSET file issue is. Can you explain?

PS: While you're at it, let's fix Util#readDataFromFile to log the exception too.

prateekm:
@jmakes can we close this PR?

jmakes:
@prateekm it was never approved. Do you have more feedback?

PR #413, opened 2018-01-23
SAMZA-1555: Move creation of checkpoint and changelog streams to the Job Coordinators

dnishimura:
**Overview**
The purpose of this PR is to consolidate the creation of the changelog and checkpoint streams into the JobCoordinators. In the current state, the changelog stream is created from the JobModelManager and the checkpoint stream is created within the OffsetManager. The issue with creating the checkpoint in the OffsetManager is that the first call happens from the first SamzaContainer that runs and each subsequent SamzaContainer run will attempt to create the checkpoint stream.

**Motivations**
There are three driving forces for this refactoring. The first motivation is to assign the creation of the changelog and checkpoint streams to the JobCoordinators where it is most appropriate. This was discussed in more detail with @nickpan47  . The second motivation is to have any potential failure to stream creation happen no later than during job coordination. The third motivation is to accommodate future security work to provide a robust way to set ACLs on streams.

Follow on to this PR will be: https://issues.apache.org/jira/browse/SAMZA-1564


samza-core/src/main/scala/org/apache/samza/job/local/ThreadJobFactory.scala
```diff
@@ -38,6 +41,14 @@ class ThreadJobFactory extends StreamJobFactory with Logging {
     info("Creating a ThreadJob, which is only meant for debugging.")
     val coordinator = JobModelManager(config)
     val jobModel = coordinator.jobModel
+    val metricsRegistry = new MetricsRegistryMap()
+
+    //create necessary checkpoint and changelog streams, if not created
+    CheckpointManagerUtil.createAndInit(coordinator.jobModel, metricsRegistry)
+    val changelogPartitionManager = ChangelogPartitionManager.fromConfig(config, getClass.getSimpleName, metricsRegistry)
```
nickpan47: Is this due to the reason that you want to skip the initialization of coordinator systems? But in fromConfig() we do it anyways?

samza-api/src/main/java/org/apache/samza/checkpoint/CheckpointManager.java
```diff
@@ -26,6 +26,14 @@
  * implementation-specific location.
  */
 public interface CheckpointManager {
+  /**
+   * Bootstrap any required resources needed prior to start.
+   */
+  default void init() { }
```
nickpan47: Why do we need an additional init() here? We already have start().

dnishimura: The `init()` method gets called in job coordinators so that the `KafkaCheckpointManager` can create the checkpoint stream there. A better name could be `createStreams()`, but I wasn't sure if that was generic enough for the CheckpointManager interface. Any suggestions on a name or does `createStreams()` work?
The problem with the `start()` method is that it gets called in the `OffsetManager.start()` method which is called in the `SamzaContainer.run()` method. That happens too late for checkpoint creation.

samza-core/src/main/java/org/apache/samza/checkpoint/CheckpointManagerUtil.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.apache.samza.checkpoint;
+
+import org.apache.samza.config.TaskConfig;
+import org.apache.samza.job.model.JobModel;
+import org.apache.samza.metrics.MetricsRegistry;
+import org.apache.samza.util.Util;
+import scala.Option;
+
+public class CheckpointManagerUtil {
```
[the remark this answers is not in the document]
dnishimura: I can't remember, but I think I was using something else from `JobModel` before I did some more refactoring. I'll  do the decoupling.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -97,6 +104,11 @@
    */
   private final JobModelManager jobModelManager;
 
+  /**
+   * A ChangelogPartitionManager to handle creation of changelog stream and map changelog stream partitions
+   */
+  private final ChangelogPartitionManager changelogPartitionManager;
```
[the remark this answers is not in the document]
dnishimura: That's legacy. ChangelogStreamManager does make more sense.

samza-core/src/main/java/org/apache/samza/storage/ChangelogPartitionManager.java
```diff
@@ -86,4 +98,72 @@ public void writeChangeLogPartitionMapping(Map<TaskName, Integer> changelogEntri
...
+  public void createChangeLogStreams(JobModel jobModel) {
+    createChangeLogStreams(getSource(), jobModel);
+  }
+
+  /**
+   * Create and validate changelog streams
+   * @param source Source name that creates the changelog stream
+   * @param jobModel JobModel with changelog information
+   */
+  public static void createChangeLogStreams(String source, JobModel jobModel) {
```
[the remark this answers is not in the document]
dnishimura: Agreed. Will decouple JobModel.

samza-core/src/main/java/org/apache/samza/storage/ChangelogPartitionManager.java
```diff
@@ -86,4 +98,72 @@ public void writeChangeLogPartitionMapping(Map<TaskName, Integer> changelogEntri
...
+          String accesslogStream = storageConfig.getAccessLogStream(systemStream.getStream());
+          StreamSpec accesslogSpec = new StreamSpec(accesslogStream, accesslogStream, systemStream.getSystem(),
+              jobModel.maxChangeLogStreamPartitions);
+          systemAdmin.createStream(accesslogSpec);
+          systemAdmin.validateStream(accesslogSpec);
+        }
+      });
+  }
+
+  public static ChangelogPartitionManager fromConfig(Config config, String source, MetricsRegistryMap registryMap) {
```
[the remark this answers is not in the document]
dnishimura: Agreed. Will look into it.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -195,10 +195,14 @@ void doOnProcessorChange(List<String> processors) {
...
     // Generate the JobModel
     JobModel jobModel = generateNewJobModel(currentProcessorIds);
-    if (!hasCreatedChangeLogStreams) {
-      JobModelManager.createChangeLogStreams(new StorageConfig(config), jobModel.maxChangeLogStreamPartitions);
-      hasCreatedChangeLogStreams = true;
+
+    // Create checkpoint and changelog streams if they don't exist
+    if (!hasCreatedStreams) {
+      CheckpointManagerUtil.createAndInit(jobModel, metrics.getMetricsRegistry());
+      ChangelogPartitionManager.createChangeLogStreams(this.getClass().getSimpleName(), jobModel);
```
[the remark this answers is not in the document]
dnishimura: True. Will look into consolidating the APIs.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -91,17 +90,22 @@ object JobModelManager extends Logging {
...
 
     val localityManager = new LocalityManager(coordinatorSystemProducer, coordinatorSystemConsumer)
     // We don't need to start() localityManager as they share the same instances with checkpoint and changelog managers.
     // TODO: This code will go away with refactoring - SAMZA-678
 
     localityManager.start()
 
+    val config = coordinatorSystemConsumer.getConfig
+    info("Got config: %s" format config)
+    val changelogManager = new ChangelogPartitionManager(coordinatorSystemProducer, coordinatorSystemConsumer, SOURCE)
```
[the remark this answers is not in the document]
dnishimura: changelogManager wasn't added, but just moved down further. I do agree though both localityManager and changelogManager should be moved out. I'll look into doing this.

samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
```diff
@@ -141,6 +144,14 @@ private void populateContainerLocalityMappings() {
     return containers;
   }
 
+  public Map<TaskName, Integer> getChangelogTaskPartitionMappings() {
+    return changelogTaskPartitionMappings;
```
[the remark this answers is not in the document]
dnishimura: Agreed.

samza-core/src/main/java/org/apache/samza/storage/ChangelogPartitionManager.java
```diff
@@ -86,4 +98,72 @@ public void writeChangeLogPartitionMapping(Map<TaskName, Integer> changelogEntri
...
+    Map<String, SystemStream> storeNameSystemStreamMapping =
+        storageConfig.getStoreNames()
+            .stream()
+            .filter(name -> StringUtils.isNotBlank(storageConfig.getChangelogStream(name)))
+            .collect(Collectors.toMap(name -> name,
+                name -> Util.getSystemStreamFromNames(storageConfig.getChangelogStream(name))));
+
+    // Get SystemAdmin for changelog store's system and attempt to create the stream
+    JavaSystemConfig systemConfig = new JavaSystemConfig(jobModel.getConfig());
+    Map<String, SystemAdmin> systemAdminMapping = systemConfig.getSystemAdmins();
```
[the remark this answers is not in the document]
dnishimura: Just saw his PR get accepted a few hours ago. I'll merge and incorporate his changes.

samza-core/src/main/scala/org/apache/samza/job/local/ThreadJobFactory.scala
dnishimura: Initially I wanted to hide the initialization, but it's probably a bad way to do it. Will fix it in refactoring the changelogManager interface.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -128,10 +132,7 @@ object JobModelManager extends Logging {
       taskName -> Integer.valueOf(taskModel.getChangelogPartition.getPartitionId)
     }}.toMap ++ previousChangelogPartitionMapping.asScala
     info("Saving task-to-changelog partition mapping: %s" format newChangelogPartitionMapping)
-    changelogManager.writeChangeLogPartitionMapping(newChangelogPartitionMapping.asJava)
-
-    createChangeLogStreams(config, jobModel.maxChangeLogStreamPartitions)
-    createAccessLogStreams(config, jobModel.maxChangeLogStreamPartitions)
+    jobModel.setChangelogTaskPartitionMappings(newChangelogPartitionMapping.asJava);
```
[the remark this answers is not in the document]
dnishimura: I debated both sides because the JobModel defines the task to partition, but it is cleaner like you said to completely decouple changelog from JobModel. Will look into your suggestion.

dnishimura:
@nickpan47 I incorporated your suggested changes and also created a CoordinatorStream class to encapsulate creation and management of the coordinator stream producer and consumer. Please review when you get a chance. Thanks!

samza-core/src/main/java/org/apache/samza/checkpoint/CheckpointManagerUtil.java
```diff
@@ -0,0 +1,50 @@
...
+ * Checkpoint manager utility class.
+ */
+public class CheckpointManagerUtil {
+
+  /**
+   * Creates and initlizes the checkpoint manager from config.
+   *
+   * @param config Configuration with checkpoint manager properties
+   * @param metricsRegistry Metrics registry for the checkpoint manager to use.
+   * @return CheckpointManager.
```
prateekm: Document that (and when) this can return null.

samza-core/src/main/java/org/apache/samza/checkpoint/CheckpointManagerUtil.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.samza.util.Util;
+import scala.Option;
+
+/**
+ * Checkpoint manager utility class.
+ */
+public class CheckpointManagerUtil {
+
+  /**
+   * Creates and initlizes the checkpoint manager from config.
```
prateekm: Minor: initializes.

It's a little weird that this only does init but not start and stop. How about we:
1. Move the classloading to the config class.
2. Let the caller do the init (or #createStreams), start and stop?

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -147,8 +161,17 @@ public ClusterBasedJobCoordinator(Config coordinatorSystemConfig) {
...
 
-    //build a JobModelReader and perform partition assignments.
-    jobModelManager = buildJobModelManager(coordinatorSystemConfig, metrics);
+    // Initialize CoordinatorStreamSystem to share between JobModelManager and ChangelogPartitionManager
+    CoordinatorStreamSystemConsumer coordinatorSystemConsumer = new CoordinatorStreamSystemConsumer(coordinatorSystemConfig, metrics);
+    CoordinatorStreamSystemProducer coordinatorSystemProducer = new CoordinatorStreamSystemProducer(coordinatorSystemConfig, metrics);
+
+    //build a JobModelManager and ChangelogPartitionManager and perform partition assignments.
+    CoordinatorStream coordinatorStream = new CoordinatorStream(coordinatorSystemConfig, metrics, COORDINATOR_STREAM_SOURCE);
+    coordinatorStream.startConsumer();
```
prateekm: Should just be a coordinatorStream#start and stop.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -147,8 +161,17 @@ public ClusterBasedJobCoordinator(Config coordinatorSystemConfig) {
 
     metrics = new MetricsRegistryMap();
 
-    //build a JobModelReader and perform partition assignments.
-    jobModelManager = buildJobModelManager(coordinatorSystemConfig, metrics);
+    // Initialize CoordinatorStreamSystem to share between JobModelManager and ChangelogPartitionManager
+    CoordinatorStreamSystemConsumer coordinatorSystemConsumer = new CoordinatorStreamSystemConsumer(coordinatorSystemConfig, metrics);
```
prateekm: Where is this used?

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStream.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.samza.config.Config;
+import org.apache.samza.metrics.MetricsRegistry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Encapsulates the creation and management of the coordinator stream producer and consumer.
+ */
+public class CoordinatorStream {
```
prateekm: What's the difference between this class and AbstractCoordinatorStreamManager? Looks like that one also maintains the producer/consumer?

samza-core/src/main/java/org/apache/samza/storage/ChangelogPartitionManager.java
```diff
@@ -74,16 +98,76 @@ public void register(TaskName taskName) {
...
+    // Get changelog store config
+    JavaStorageConfig storageConfig = new JavaStorageConfig(config);
+    Map<String, SystemStream> storeNameSystemStreamMapping = storageConfig.getStoreNames()
+        .stream()
+        .filter(name -> StringUtils.isNotBlank(storageConfig.getChangelogStream(name)))
+        .collect(Collectors.toMap(name -> name,
+            name -> Util.getSystemStreamFromNames(storageConfig.getChangelogStream(name))));
+
+    // Get SystemAdmin for changelog store's system and attempt to create the stream
+    SystemAdmins systemAdmins = new SystemAdmins(config);
```
prateekm: I think this will load all systems. You might want to load just the changelog system admin.

samza-core/src/main/java/org/apache/samza/storage/ChangelogPartitionManager.java
```diff
@@ -39,10 +48,25 @@
...
   private boolean isCoordinatorConsumerRegistered = false;
 
-  public ChangelogPartitionManager(CoordinatorStreamSystemProducer coordinatorStreamProducer,
-      CoordinatorStreamSystemConsumer coordinatorStreamConsumer,
-      String source) {
-    super(coordinatorStreamProducer, coordinatorStreamConsumer, source);
+  /**
+   * Construct Changelog manager without a coordinator stream.
+   *
+   *  @param source Name of caller.
```
prateekm: What is the source used for?

samza-core/src/main/scala/org/apache/samza/config/StorageConfig.scala
```diff
@@ -51,7 +50,7 @@ class StorageConfig(config: Config) extends ScalaMapConfig(config) with Logging
   def getStorageMsgSerde(name: String) = getOption(StorageConfig.MSG_SERDE format name)
 
   def getAccessLogEnabled(storeName: String) = {
-    getBoolean(ACCESSLOG_ENABLED format storeName, false)
+    new JavaStorageConfig(config).getAccessLogEnabled(storeName);
```
prateekm: Minor: Don't need `;` in scala, here and elsewhere.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -64,79 +63,36 @@ object JobModelManager extends Logging {
...
 
-    localityManager.start()
+    val config = coordinatorSystemConsumer.getConfig
 
     // Map the name of each system to the corresponding SystemAdmin
     val systemAdmins = new SystemAdmins(config)
     val streamMetadataCache = new StreamMetadataCache(systemAdmins, 0)
-    val previousChangelogPartitionMapping = changelogManager.readChangeLogPartitionMapping()
 
     val processorList = new ListBuffer[String]()
```
prateekm: Not your code, but would be cleaner to use either: 
val x = (1 to containerCount).toList
val x = List.range(0, containerCount)

samza-core/src/main/scala/org/apache/samza/job/local/ThreadJobFactory.scala
```diff
@@ -36,8 +40,21 @@ import org.apache.samza.util.Logging
 class ThreadJobFactory extends StreamJobFactory with Logging {
   def getJob(config: Config): StreamJob = {
     info("Creating a ThreadJob, which is only meant for debugging.")
-    val coordinator = JobModelManager(config)
+
+    val metricsRegistry = new MetricsRegistryMap()
+    val coordinatorStream = new CoordinatorStream(config, metricsRegistry, getClass.getSimpleName)
+    coordinatorStream.startConsumer()
+    coordinatorStream.startProducer()
```
prateekm: Where are these stopped?

samza-core/src/main/scala/org/apache/samza/job/local/ProcessJobFactory.scala
```diff
@@ -19,28 +19,43 @@
...
 
     if (containerCount > 1) {
       throw new SamzaException("Container count larger than 1 is not supported for ProcessJobFactory")
     }
-    
-    val coordinator = JobModelManager(config)
+
+    val metricsRegistry = new MetricsRegistryMap()
+    val coordinatorStream = new CoordinatorStream(config, metricsRegistry, getClass.getSimpleName)
+    coordinatorStream.startConsumer()
```
prateekm: Where are these stopped?

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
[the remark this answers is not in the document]
prateekm: I wouldn't mind changing the class name. It should be internal to Samza.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
dnishimura: I think I originally did this because the `LocalityManager` in SamzaContainer was just using the producer, so I figured I'd give an option to start both or just one.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStream.java
dnishimura: This is where it got a little confusing for me. The consumer and producer lifecycles were partially handled outside of the `AbstractCoordinatorStreamManager` and partially in it. Also, I wasn't sure what the purpose of having this abstract class other than some code re-use. The name was kind of confusing too because the subclasses are not really variations to a coordinator stream manager, but instead they are "managers that use the coordinator stream". Composition might be better? What do you think? If so, I can move all the behavior to `CoordinatorStream`. Just a thought.

samza-core/src/main/java/org/apache/samza/storage/ChangelogPartitionManager.java
dnishimura: Source is a member of the parent class and it looks like it's mostly used for logging, except for some non-Kafka consumer/producer implementations.

samza-core/src/main/java/org/apache/samza/storage/ChangelogPartitionManager.java
dnishimura: I thought `StorageConfig` just has the changelog systems, but I'll double check.

samza-core/src/main/scala/org/apache/samza/config/StorageConfig.scala
dnishimura: Good catch.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
dnishimura: Good idea. I'll clean that up.

samza-core/src/main/scala/org/apache/samza/job/local/ProcessJobFactory.scala
dnishimura: The subclasses of `AbstractCoordinationManager` do it, but I see where you're going at. See the comment above regarding the CoordinatorStream.

samza-core/src/main/scala/org/apache/samza/job/local/ThreadJobFactory.scala
dnishimura: Ditto from above.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
dnishimura: Nowhere 😄 . Forgot to delete them. Good catch!

dnishimura:
@nickpan47 and @prateekm I incorporated your suggested changes. Please review when you get the chance. Thanks!

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStream.java
prateekm: After looking at the current code, here's what I'd recommend:

1. Change AbstractCoordinatorStreamManager to a concrete class (CoordinatorStreamManager). 
3. Make CoordinatorStreamSystemConsumer and CoordinatorStreamSystemProducer private to the CoordinatorStreamManager. CoordinatorStreamManager should have two constructors: one that instantiates everything from Config (for general use), and one that accepts a SystemConsumer and SystemProducer and wraps it with the CoordinatorStreamSystemConsumer and CoordinatorStreamSystemProducer (for testing/custom lifecycle management).
3. CoordinatorStreamManager should proxy all public methods in the CSSC/CSSP.
4. LocalityManager/ChangelogManager/TaskAssignmentManager should accept a CoordinatorStreamManager instance instead of extending the abstract class.
4. Remove the CoordinatorStream class.

It looks like a fairly straightforward refactor. If you prefer to do it in a separate PR, that's OK too. @nickpan47, does this seem reasonable?

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -35,56 +33,54 @@
...
-   *
-   * @throws UnsupportedOperationException in the case if a {@link TaskName} is passed
+   * Ensures the coordinator stream is started.
    */
-  @Override
-  public void register(TaskName taskName) {
-    throw new UnsupportedOperationException("TaskName cannot be registered with LocalityManager");
+  public void start() {
+    if (!writeOnly) {
+      this.coordinatorStreamManager.startCoordinatorStreamConsumer();
```
prateekm: Other classes should not be able to manage lifecycle for CoordinatorStreamManager components.
Can we make the producer and consumer and their register/start/stop methods private to the CoordinatorStreamManager? 

Also, The coordinatorStreamManager should be started/stopped by the class that created it, not the classes that use it. Is that possible?

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamManager.java
```diff
@@ -0,0 +1,197 @@
...
+  public CoordinatorStreamManager(Config coordinatorSystemConfig, MetricsRegistry metricsRegistry, String source) {
+    coordinatorStreamConsumer = new CoordinatorStreamSystemConsumer(coordinatorSystemConfig, metricsRegistry);
+    coordinatorStreamProducer = new CoordinatorStreamSystemProducer(coordinatorSystemConfig, metricsRegistry);
+    this.source = source;
+  }
+
+  /**
+   * Fully starts, registers and bootstraps the underlying coordinator stream producer and consumer.
+   */
+  public void registerStartBootstrapAll() {
```
prateekm: This is a weird method. Should these be separate register, start and bootstrap methods?

dnishimura:
@prateekm committed changes from your most recent suggestions. Thanks!

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamManager.java
```diff
@@ -129,69 +119,19 @@ public void bootstrapCoordinatorStreamConsumer() {
    */
   public Set<CoordinatorStreamMessage> getBootstrappedStream(String source) {
     if (coordinatorStreamConsumer == null) {
-      throw new UnsupportedOperationException(String.format("CoordinatorStreamConsumer is not initialized in the CoordinatorStreamManager. "
-          + "input source: %s", source));
+      throw new UnsupportedOperationException(String.format("CoordinatorStreamConsumer is not initialized in the CoordinatorStreamManager. "));
```
prateekm: IllegalStateException is more conventional in this scenario.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamManager.java
```diff
@@ -129,69 +119,19 @@ public void bootstrapCoordinatorStreamConsumer() {
...
-      coordinatorStreamProducer.register(source);
-    }
-  }
-
   /**
    * Returns the config for the coordinator stream consumer.
    * @return Config of the coordinator stream consumer.
    */
-  public Config getCoordinatorStreamConsumerConfig() {
+  public Config getConfig() {
```
prateekm: Why does this need to be a public method? 
Is the javadoc supposed to be "Get config **from** the coordinator stream"?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -841,8 +842,11 @@ class SamzaContainer(
   def startLocalityManager {
     if(localityManager != null) {
       info("Registering localityManager for the container")
-      localityManager.start
-      localityManager.register(String.valueOf(containerContext.id))
+      // Ideally coordinator stream manager should be directly, but the SamzaContainer constructor parameter list is
```
prateekm: Should be ok to add 1 more param, keeps the dependencies clearer. The list too long is a symptom of a different problem. This way you can also remove the public getCoordinatorStreamManager() from LocalityManager.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -186,6 +209,18 @@ public void run() {
...
 
+      //create necessary checkpoint and changelog streams, if not created
+      JobModel jobModel = jobModelManager.jobModel();
+      CheckpointManager checkpointManager = new TaskConfigJava(config).getCheckpointManager(metrics);
+      if (checkpointManager != null) {
+        checkpointManager.createStream();
+      }
+      changelogStreamManager.createChangeLogStreams(jobModel.getConfig(), jobModel.maxChangeLogStreamPartitions);
+
+      // Remap change
```
prateekm: Minor: Incomplete comment?

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -186,6 +209,18 @@ public void run() {
...
+      Map prevPartitionMappings = changelogStreamManager.readPartitionMapping();
+      changelogStreamManager.updatePartitionMapping(prevPartitionMappings, jobModel.getTaskPartitionMappings());
```
prateekm: QQ: Do you know why changelogStreamManager#updatePartitionMapping  doesn't read the mapping itself?

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -35,56 +33,25 @@
...
  * assignment information from the coordinator stream
  * */
-public class LocalityManager extends AbstractCoordinatorStreamManager {
+public class LocalityManager {
   private static final Logger log = LoggerFactory.getLogger(LocalityManager.class);
   private Map<String, Map<String, String>> containerToHostMapping = new HashMap<>();
   private final TaskAssignmentManager taskAssignmentManager;
   private final boolean writeOnly;
+  private CoordinatorStreamManager coordinatorStreamManager;
+  private static final String CONTAINER_PREFIX = "SamzaContainer-";
```
prateekm: Nitpick: Sort class fields, here and in other classes. Usual convention is statics before final before non-final, groups separated by a newline.

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -35,56 +33,25 @@
  * Locality Manager is used to persist and read the container-to-host
  * assignment information from the coordinator stream
  * */
-public class LocalityManager extends AbstractCoordinatorStreamManager {
+public class LocalityManager {
   private static final Logger log = LoggerFactory.getLogger(LocalityManager.class);
   private Map<String, Map<String, String>> containerToHostMapping = new HashMap<>();
   private final TaskAssignmentManager taskAssignmentManager;
   private final boolean writeOnly;
+  private CoordinatorStreamManager coordinatorStreamManager;
```
prateekm: Can this be final?

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
prateekm: Also, why does this say "SamzaContainer" instead of this class' name?

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -35,56 +33,25 @@
...
-   * @param sourceSuffix the source suffix which is a container id
-   */
-  public void register(String sourceSuffix) {
-    if (!this.writeOnly) {
-      registerCoordinatorStreamConsumer();
-    }
-    registerCoordinatorStreamProducer(getSource() + sourceSuffix);
+  public LocalityManager(CoordinatorStreamManager coordinatorStreamManager, boolean writeOnly) {
+    this.coordinatorStreamManager = coordinatorStreamManager;
+    this.writeOnly = writeOnly;
```
prateekm: Is this flag still required?

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -35,56 +33,25 @@
...
-   *
-   * @param sourceSuffix the source suffix which is a container id
-   */
-  public void register(String sourceSuffix) {
-    if (!this.writeOnly) {
-      registerCoordinatorStreamConsumer();
-    }
-    registerCoordinatorStreamProducer(getSource() + sourceSuffix);
+  public LocalityManager(CoordinatorStreamManager coordinatorStreamManager, boolean writeOnly) {
+    this.coordinatorStreamManager = coordinatorStreamManager;
```
prateekm: Does this need to register itself with the CoordinatorStreamManager as well?

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamManager.java
```diff
@@ -0,0 +1,137 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.coordinator.stream;
+
+import org.apache.samza.config.Config;
+import org.apache.samza.coordinator.stream.messages.CoordinatorStreamMessage;
+
+import java.util.Set;
```
prateekm: Nitpick: Sort imports.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamManager.java
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Class which handles the common functionality for coordinator stream consumer and producer
+ */
+public class CoordinatorStreamManager {
+  private Logger log = LoggerFactory.getLogger(CoordinatorStreamManager.class);
+  private CoordinatorStreamSystemProducer coordinatorStreamProducer;
+  private CoordinatorStreamSystemConsumer coordinatorStreamConsumer;
+
+  /**
+   * Creates a new {@link CoordinatorStreamManager} with a given coordinator stream producer, consumer and with a given source.
```
prateekm: Minor: Update javadoc (doesn't take source anymore)

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamManager.java
```diff
@@ -0,0 +1,137 @@
...
+    return coordinatorStreamConsumer.getBootstrappedStream(source);
+  }
+
+  /**
+   * Returns the config for the coordinator stream consumer.
+   * @return Config of the coordinator stream consumer.
+   */
+  public Config getConfig() {
+    if (coordinatorStreamConsumer == null) {
+      throw new UnsupportedOperationException(String.format("CoordinatorStreamConsumer is not initialized in the CoordinatorStreamManager. "));
```
prateekm: Minor: IllegalStateException.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamManager.java
dnishimura: Good catch. thanks.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamSystemConsumer.java
```diff
@@ -290,4 +290,21 @@ public boolean hasNewMessages(SystemStreamPartitionIterator iterator) {
     return iterator.hasNext();
   }
 
+  /**
+   * Has the SystemConsumer started.
+   *
+   * @return True if started, otherwise false.
+   */
+  public boolean isStarted() {
```
prateekm: I don't think isBootstrapped should be a flag. The 'bootstrap' method is a misnomer, it should be called 'refresh' or something. It's meant to be called multiple times, once before each use, to get the latest stream state.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamSystemProducer.java
```diff
@@ -148,4 +148,13 @@ public void writeConfig(String source, Config config) {
     }
     systemProducer.flush(source);
   }
+
+  /**
+   * Has the SystemProducer started.
+   *
+   * @return True if started, otherwise false
+   */
+  public boolean isStarted() {
```
prateekm: Is this still required?

samza-core/src/main/java/org/apache/samza/standalone/PassthroughJobCoordinator.java
```diff
@@ -75,6 +77,10 @@ public void start() {
     JobModel jobModel = null;
     try {
       jobModel = getJobModel();
+      CheckpointManager checkpointManager = new TaskConfigJava(jobModel.getConfig()).getCheckpointManager(null);
+      if (checkpointManager != null) {
+        checkpointManager.createStream();
```
prateekm: Does this not need to create changelog/coordinator streams?

samza-core/src/main/java/org/apache/samza/storage/ChangelogStreamManager.java
```diff
@@ -0,0 +1,161 @@
...
+import org.apache.samza.system.StreamSpec;
+import org.apache.samza.system.SystemAdmin;
+import org.apache.samza.system.SystemStream;
+import org.apache.samza.util.Util;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * The Changelog manager is used to persist and read the changelog information from the coordinator stream.
```
prateekm: Update javadoc. Do you mean changelog partition mapping information?

samza-core/src/main/java/org/apache/samza/storage/ChangelogStreamManager.java
```diff
@@ -0,0 +1,161 @@
...
+  // This is legacy for changelog. Need to investigate what happens if you use a different source name
+  private static final String SOURCE = "JobModelManager";
+
+  /**
+   * Construct Changelog manager without a coordinator stream.
+   *
+   *  @param source Name of caller.
+   */
+  public ChangelogStreamManager(String source) {
+    coordinatorStreamManager = null;
```
prateekm: How does this work? Where is this used?

samza-core/src/main/java/org/apache/samza/storage/ChangelogStreamManager.java
```diff
@@ -0,0 +1,161 @@
...
+      changelogMapping.put(new TaskName(changelogMapEntry.getTaskName()), changelogMapEntry.getPartition());
+      log.debug("TaskName: {} is mapped to {}", changelogMapEntry.getTaskName(), changelogMapEntry.getPartition());
+    }
+    return changelogMapping;
+  }
+
+  /**
+   * Write the taskName to partition mapping.
+   * @param changelogEntries The entries that needs to be written to the coordinator stream, the map takes the taskName
+   *                       and it's corresponding changelog partition.
```
prateekm: Nitpick: formatting (align with "The" on the previousline.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -64,79 +61,31 @@ object JobModelManager extends Logging {
...
-   * c) Recomputes changelog partition mapping based on jobModel and job's configuration
-   * and writes it to the coordinator stream.
-   * d) Builds JobModelManager using the jobModel read from coordinator stream.
-   * @param coordinatorSystemConfig A config object that contains job.name
-   *                                job.id, and all system.&lt;job-coordinator-system-name&gt;.*Ch
-   *                                configuration. The method will use this config to read all configuration
-   *                                from the coordinator stream, and instantiate a JobModelManager.
+   * b) Recomputes changelog partition mapping based on jobModel and job's configuration.
+   * c) Builds JobModelManager using the jobModel read from coordinator stream.
+   * @param coordinatorStreamManager
```
prateekm: Nitpick: Add description to params and return.

samza-core/src/main/scala/org/apache/samza/job/local/ProcessJobFactory.scala
```diff
@@ -19,28 +19,45 @@
...
+    coordinatorStreamManager.register(getClass.getSimpleName)
+    coordinatorStreamManager.start
+    coordinatorStreamManager.bootstrap
+    val changelogManager = new ChangelogStreamManager(coordinatorStreamManager)
+
+    val coordinator = JobModelManager(coordinatorStreamManager, changelogManager.readPartitionMapping())
+    val jobModel = coordinator.jobModel
+    changelogManager.writePartitionMapping(jobModel.getTaskPartitionMappings)
+
+    //create necessary checkpoint and changelog streams, if not created
```
prateekm: Minor: Update comment (remove "if not created")

samza-core/src/main/scala/org/apache/samza/job/local/ThreadJobFactory.scala
```diff
@@ -36,8 +38,25 @@ import org.apache.samza.util.Logging
...
+    coordinatorStreamManager.register(getClass.getSimpleName)
+    coordinatorStreamManager.start
+    coordinatorStreamManager.bootstrap
+    val changelogPartitionManager = new ChangelogStreamManager(coordinatorStreamManager)
+
+    val coordinator = JobModelManager(coordinatorStreamManager, changelogPartitionManager.readPartitionMapping())
     val jobModel = coordinator.jobModel
+    changelogPartitionManager.writePartitionMapping(jobModel.getTaskPartitionMappings)
+
+    //create necessary checkpoint and changelog streams, if not created
```
prateekm: Minor: update comment.

samza-core/src/main/java/org/apache/samza/storage/ChangelogStreamManager.java
dnishimura: 👍

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
dnishimura: 👍

samza-core/src/main/scala/org/apache/samza/job/local/ProcessJobFactory.scala
dnishimura: 👍

samza-core/src/main/scala/org/apache/samza/job/local/ThreadJobFactory.scala
dnishimura: 👍

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamSystemConsumer.java
dnishimura: Thanks for clarifying. I'll remove that flag.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -186,6 +209,18 @@ public void run() {
       //initialize JobCoordinator state
       log.info("Starting Cluster Based Job Coordinator");
 
+      //create necessary checkpoint and changelog streams, if not created
+      JobModel jobModel = jobModelManager.jobModel();
+      CheckpointManager checkpointManager = new TaskConfigJava(config).getCheckpointManager(metrics);
```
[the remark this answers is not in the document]
prateekm: @dnishimura Yup, just a question for Yi.

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
prateekm: My bad, I thought this was the "SOURCE" like the other constants. Makes sense.

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
prateekm: I think I misunderstood, looks like other Managers don't need to register _themselves_ with the CoordinatorStreamManager. We just need to call register on the CoordinatorStreamManager. In that case this makes sense. 

Although, in that case, why does the register on CSM exist? Couldn't its job be done in the constructor?

samza-core/src/main/java/org/apache/samza/storage/ChangelogStreamManager.java
prateekm: Yeah, it does that sometimes if the difference is too large. Thanks for fixing.

samza-core/src/main/java/org/apache/samza/storage/ChangelogStreamManager.java
prateekm: Since this's the only state in this class, this would imply that whatever methods ZkJobCoordinator is using can be made static. Might be cleaner than adding this constructor.

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
dnishimura: Most cases, yes, but if you look at `SamzaContainer`, the source name to register with is not known until the container ID is known from the `containerContext`. I'm sure all of that can be refactored. Also one more thing is that register does a fetch on the consumer, if there is a consumer, so I'm not sure doing IO in the constructor is a good idea.

samza-core/src/main/java/org/apache/samza/storage/ChangelogStreamManager.java
dnishimura: Agreed, will change that.

dnishimura:
@prateekm added changes regarding your comments.

dnishimura:
Thanks for helping review!

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -725,7 +726,7 @@ class SamzaContainer(
       shutdownDiskSpaceMonitor
       shutdownHostStatisticsMonitor
       shutdownProducers
-      shutdownLocalityManager
+      shutdownCoordinatorStreamManager
```
nickpan47: The fact that the start/shutdown sequence is asymmetric let me think that it would be better to fold this two steps.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -64,79 +61,31 @@ object JobModelManager extends Logging {
   /**
    * Does the following actions for a job.
    * a) Reads the jobModel from coordinator stream using the job's configuration.
```
nickpan47: Thanks! Way much better now!

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamSystemConsumer.java
```diff
@@ -203,8 +200,7 @@ public void bootstrap() {
 
   public Set<CoordinatorStreamMessage> getBoostrappedStream() {
     log.info("Returning the bootstrapped data from the stream");
-    if (!isBootstrapped)
-      bootstrap();
+    bootstrap();
```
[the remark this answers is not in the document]
dnishimura: @prateekm mentioned that we should be able to bootstrap, or a more appropriately, "refresh" the current configuration state at any time. Looks like the name is a bit of a misnomer. Was this added as a stopgap on a performance issue? If so, I can add it back in.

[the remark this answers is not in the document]
dnishimura: @nickpan47 @prateekm The isBoostrap flag was added back in.

samza-core/src/main/java/org/apache/samza/config/TaskConfigJava.java
```diff
@@ -43,13 +47,42 @@
...
+    return get(CHECKPOINT_MANAGER_FACTORY, null);
+  }
+
+  /**
+   * Create the checkpoint manager
+   *
+   * @param metricsRegistry Registry of metrics to use. Can be null if not using metrics.
+   * @return CheckpointManager object if checkpoint manager factory is configured, otherwise null.
+   */
+  public CheckpointManager getCheckpointManager(MetricsRegistry metricsRegistry) {
```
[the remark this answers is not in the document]
dnishimura: Seems a bit overkill to do a factory since all it needs is the config to instantiate a CheckpointManager. We can revisit in a future PR. Thanks!

samza-core/src/test/java/org/apache/samza/operators/impl/TestOperatorImplGraph.java
```diff
@@ -320,7 +320,7 @@ public void testOperatorGraphInitAndClose() {
   }
 
   /**
-   * Creates an identity map function that appends to the provided lists when init/close is invoked.
+   * Creates an identity map function that appends to the provided lists when createStream/close is invoked.
```
nickpan47: Why changing this file? Seems irrelevant?

dnishimura: Inadvertent refactor from IntelliJ. Fixed. Thanks!

dnishimura:
@nickpan47 PR ready for commit. Thanks!

PR #415, opened 2018-01-26
SAMZA-1578: Fix watermark bug found by BEAM tests

xinyuiscool:
The problem is getOutputWatermark() does not return the real outputWatermark. This caused problem in user override watermark function.


vjagadish1989 (approved):
thanks for the contribution, approved!

PR #417, opened 2018-01-30
SAMZA-1489; TaskInstance should commit offset before it closes() if auto commit is enabled

lindong28:


lindong28:
@vjagadish1989 Thanks for the review. Later we will have another patch to delete records from intermediate stream after taskInstance.commit(). In order to test this, we need the taskInstance.commit() to be invoked after RepartitionJoinWindowApp is closed. Not sure if end-of-stream will be signaled if we stop repartitionJoinWindowApp is closed. Does this make sense?

samza-core/src/main/scala/org/apache/samza/config/TaskConfig.scala
```diff
@@ -43,6 +44,11 @@ object TaskConfig {
   val CALLBACK_TIMEOUT_MS = "task.callback.timeout.ms"  // timeout period for triggering a callback
   val ASYNC_COMMIT = "task.async.commit" // to enable async commit in a AsyncStreamTask
 
+  val DEFAULT_WINDOW_MS: Long = -1L
+  val DEFAULT_COMMIT_MS = 60000L
+  val DEFAULT_CALLBACK_TIMEOUT_MS: Long = -1L
```
vjagadish1989: Thanks Dong for this cleanup.

vjagadish1989 (approved):
approved, looks good. thank you Dong for the contribution

PR #418, opened 2018-01-30
SAMZA-1579; RemoteApplicationRunner should allow user to kill app or query app status

lindong28:


samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -78,41 +89,47 @@ public void run(StreamApplication app) {
...
     } catch (Throwable t) {
       throw new SamzaException("Failed to run application", t);
     }
   }
 
   @Override
   public void kill(StreamApplication app) {
-    try {
-      ExecutionPlan plan = getExecutionPlan(app);
+    if (!jobRunnersByApp.containsKey(app)) {
```
jmakes: I don't think this will work. 

Take a look at run-app.sh and ApplicationRunnerMain.java. The current pattern is to invoke run-app.sh with an "operation" parameter to specify which command to run. Since this pattern starts a new JVM each time, this "jobRunnersByApp" will not survive between invocations. 

cc @nickpan47 and @xinyuiscool because I think they plan to change this in some way. 

But in the meantime, I think it's safer to replan the StreamApplication for each operation to get the JobRunner instances. Planning the same app should yield the same configs and therefore the same JobRunner behavior.

samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
```diff
@@ -161,10 +167,8 @@ class JobRunner(config: Config) extends Logging {
   }
 
   def status(): ApplicationStatus = {
-    val jobFactory: StreamJobFactory = getJobFactory
-
     // Create the actual job, and get its status.
-    jobFactory.getJob(config).getStatus
+    streamJob.getStatus
```
jmakes: Same issue here. "streamJob" may not survive between commands, so this isn't safe. 

e.g. this will definitely break the deployment at LinkedIn

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
lindong28: Thanks for the review @jmakes 

I understand that it will not work if user uses run-app.sh to kill a job because it will be in a different/new process. But it seems like an existing issue, right? So I am not sure whether this patch will break anything.

This patch will improve will work if the remoteApplicationRunner.kill() is called in the same process after remoteApplicationRunner.run(). So it seems like an improvement. I can update the patch if there is better way to do it.

Not sure if I understand your last comment regarding replan". Maybe I can ask you in person about it.

samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
lindong28: Good point. I realized that this will break the existing deployment if the streamJob is YarnJob. I have updated the patch to fix it. Can you take another look?

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
jmakes: No, it's not an existing issue because before the patch, this kill method would always regenerate the ExecutionPlan. 

After the patch, it will fail because there is nothing in the map and no executionPlan from which we can recreate the JobRunner instances.

lindong28: @jmakes Thanks for the review. I found another way to trigger commit without having to explicitly stopping the application. I will close this PR and the JIRA ticket.

PR #419, opened 2018-02-07
SAMZA-1498: Support arbitrary system clock timer in operators

xinyuiscool:
This patch adds the capability to register arbitrary timers for both high-level and low-level api.  
For high-level, InitableFunction will pass the TimerRegistry to user through the new OpContext, and user will implement the TimerFunction to get timer notifications.
For low-level api, user can register timer in the TaskContext, and then implements the TimerCallback for specific timer actions.


prateekm:
@xinyuiscool Can you provide some motivation for this new feature? My original intention was to let users get callbacks on a fixed container wide timer interval, similar to the behavior of WindowableTask. This feature seems a lot more general. Also, I'd prefer to not introduce a new context object/mechanism right now since @nickpan47 is working on cleaning up that part of the API.

xinyuiscool:
@prateekm : there are two main motivations behind:
- For BEAM to support processing time timer/trigger, this is a MUST. I need the capability of registering arbitrary timers anytime during processing. For example, for BEAM's trigger:
AfterProcessingTime
        .pastFirstElementInPane()
        .plusDelayOf(Duration.standardMinutes(5)
This requires registering the trigger after first element shows up in the window pane. On the BEAM side, I am waiting eagerly for this feature to be added so I can push the code.

- For Samza, this allows the better windowing and trigger. It allows each window to register a different timer, instead of using the GCD hack we currently do (this fix is also requested by @cpettitt-linkedin ). It also allows the time trigger to use a real system timer, instead of approximating it by using the time of incoming message. I believe with this patch it will be straightforward to improve both existing features.

I think adding the context on the Operator level is inevitable to support timer on the operator level. This should also be needed if we need more context object on the operator level. I can talk to @nickpan47 about this, but I don't think the future refactoring should block current feature development.

prateekm:
@xinyuiscool Makes sense to implement if this is required for BEAM. If we're adding it to the Samza public API, that's fine too - its certainly a more flexible behavior than the current window implementation. Good point about timer triggers as well. If we do that, we should update and clarify the existing documentation, both javadoc and the introduction/architecture pages.

Re: adding the OpContext, not sure I understand why that is required if we can register the timers on the TaskContext directly. Let me take a look at the code again. Agree that refactoring shouldn't block feature development, but we also shouldn't add/change public APIs that may get refactored again. Let's check with @nickpan47 what the timeline looks like.

samza-api/src/main/java/org/apache/samza/operators/functions/TimerFunction.java
```diff
@@ -0,0 +1,63 @@
...
+ *      }
+ *    }
+ * }</pre>
+ * @param <K> type of the key
+ * @param <OM> type of the output
+ */
+public interface TimerFunction<K, OM> {
+
+  /**
+   * Registers any system timers using the registry
```
prateekm: Remove "system"

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -992,6 +997,18 @@ class SamzaContainer(
...
 
+    if (timerExecutor != null) {
+      info("Shutting down timer executor")
+      try {
+        timerExecutor.shutdown()
+        if(timerExecutor.awaitTermination(shutdownMs, TimeUnit.MILLISECONDS)) {
+          timerExecutor.shutdownNow()
+        }
+      } catch {
+        case e: Exception => error(e.getMessage, e)
```
prateekm: Log meaningful context in the error message instead of repeating e.getMessage. E.g., "Ignoring exception shutting down timer executor."

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -992,6 +997,18 @@ class SamzaContainer(
       }
     }
 
+    if (timerExecutor != null) {
+      info("Shutting down timer executor")
+      try {
+        timerExecutor.shutdown()
+        if(timerExecutor.awaitTermination(shutdownMs, TimeUnit.MILLISECONDS)) {
```
prateekm: Nitpick: space after if and before (

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -651,7 +655,8 @@ class SamzaContainer(
   securityManager: SecurityManager = null,
   reporters: Map[String, MetricsReporter] = Map(),
   jvm: JvmMetrics = null,
-  taskThreadPool: ExecutorService = null) extends Runnable with Logging {
+  taskThreadPool: ExecutorService = null,
+  timerExecutor: ScheduledExecutorService = null) extends Runnable with Logging {
```
prateekm: Why default to null? Should this be required?

samza-core/src/main/java/org/apache/samza/task/SystemTimerSchedulerFactory.java
```diff
@@ -0,0 +1,179 @@
...
+    timerListener = listener;
+  }
+
+  public Map<TimerKey<?>, TimerCallback> removeReadyTimers() {
+    final Map<TimerKey<?>, TimerCallback> timers = new TreeMap<>(readyTimers);
+    readyTimers.keySet().removeAll(timers.keySet());
+    return timers;
+  }
+
+  public class SystemTimerScheduler<K> {
```
prateekm: This is not really a scheduler, right? It's just holding the key and the future for the scheduled task. Also, since this is not a Scheduler, we don't need a SchedulerFactory, or a per key "scheduler". Can we clean up these interfaces/naming?

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
```diff
@@ -166,6 +166,11 @@ OperatorImpl createAndRegisterOperatorImpl(OperatorSpec prevOperatorSpec, Operat
       operatorImpl.init(config, context);
       operatorImpl.registerInputStream(inputStream);
 
+      if (operatorSpec.getTimerFn() != null) {
+        final TimerRegistry timerRegistry = operatorImpl.createTimerRegistry();
+        operatorSpec.getTimerFn().registerTimer(timerRegistry);
```
prateekm: Same question as before, can we do this in OperatorImpl.init?

Also, it's confusing that operatorImpl has a createTimerRegistry, since the TimerRegistry has already been created by the container and passed to the StreamOperatorTask. Maybe rename to createOperatorTimerRegistry to disambiguate.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -415,6 +419,42 @@ final long getOutputWatermark() {
...
   }
 
+  /**
+   * Returns a registry which allows registering arbitrary system-clock timer with K-typed key.
+   * The user-defined function in the operator spec needs to implement {@link TimerFunction#onTimer(Object, long)}
+   * for timer notifications.
+   * @param <K> key type for the timer.
+   * @return an instance of {@link TimerRegistry}
+   */
+  <K> TimerRegistry<K> createTimerRegistry() {
```
prateekm: Rename to createOperatorTimerRegistry to avoid confusion with TimerRegistry with for the Task.

samza-core/src/main/java/org/apache/samza/operators/impl/TimerRegistryFactory.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.apache.samza.operators.impl;
+
+import org.apache.samza.operators.TimerRegistry;
+
+/**
+ * Factory to create instance of {@link TimerRegistry}.
+ */
+public interface TimerRegistryFactory {
```
prateekm: Looks like it's unused. Can we remove this?

samza-core/src/main/java/org/apache/samza/task/SystemTimerSchedulerFactory.java
prateekm: Agree that the API should be typed, but the container doesn't currently care about the type of the key. Even for High Level API/BEAM, I'd think that the type would only matter downstream of StreamOperatorTask (e.g. in WindowImpl for HLA). 

Unless there's some type-checking reasons it's required, maybe we can go with the simpler implementation first, and make it typed when the framework actually needs to know about the type of the keys? If we do that, a simple per task Object -> Future map would probably suffice.

samza-core/src/main/java/org/apache/samza/task/SystemTimerSchedulerFactory.java
```diff
@@ -0,0 +1,183 @@
...
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+
+      TimerKey<?> timerKey = (TimerKey<?>) o;
+      if (time != ((TimerKey<?>) o).time) {
+        return false;
+      }
+      return key != null ? key.equals(timerKey.key) : timerKey.key == null;
```
prateekm: key should never be null in this class. Same for hashcode/compareTo etc.
Should add a null and uniqueness check when the key is registered.

samza-core/src/main/java/org/apache/samza/task/SystemTimerSchedulerFactory.java
xinyuiscool: Sure, I can rename this to be a timer holder, and move the schedule function outside.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
xinyuiscool: The base OperatorImpl doesn't have operatorSpec, so I am not able to move the logic there.

Fixed the naming of the method.

samza-core/src/main/java/org/apache/samza/operators/impl/TimerRegistryFactory.java
xinyuiscool: Good catch. I removed this class.

samza-core/src/main/java/org/apache/samza/task/SystemTimerSchedulerFactory.java
xinyuiscool: Fixed by changing the map to be just Map<Object, SchedulerFuture>.

samza-core/src/main/java/org/apache/samza/task/SystemTimerSchedulerFactory.java
xinyuiscool: right, makes sense. Removed the null check.

PR #420, opened 2018-02-07
SAMZA-1572: Add fixed retries on failure in KafkaCheckpointManager.

shanthoosh:
KafkaCheckpointManager.writeCheckpoint currently goes into a infinite loop when an irrecoverable failure happens, this indefinitely blocks the commit phase (there by preventing processing). Added finite retries (50), which would retry for fixed time in case of failure before giving up.


shanthoosh:
@xinyuiscool @sborya @prateekm 

prateekm:
QQ: How much time is 50 retries here?

shanthoosh:
@prateekm 
50 retries is approximately equal to 83 mins retries in time(initialDelay=`100 ms`, backoffMultiplier=`2`).

For some critical jobs, it’s required that unavailability of broker for an hour or so should not bring their job down(which will require fail-over to a backup colo).

If required we can reduce this retry duration.

prateekm:
@shanthoosh That's fine, just add a comment next to the constant 50 that this is ~ 83 min.
Otherwise LGTM.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -157,7 +159,13 @@ class KafkaCheckpointManager(checkpointSpec: KafkaStreamSpec,
       },
 
       (exception, loop) => {
-        warn(s"Retrying failed checkpoint write to key: $key, checkpoint: $checkpoint for task: $taskName", exception)
+        if (loop.sleepCount >= MAX_RETRIES_ON_FAILURE) {
+          error(s"Exhausted failure retries: $MAX_RETRIES_ON_FAILURE when writing checkpoint: $checkpoint for task: $taskName.")
+          loop.done
```
prateekm: Do you need both done and throw?

"Exhausted $MAX_RETRIES retries when ..."

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -53,6 +53,8 @@ class KafkaCheckpointManager(checkpointSpec: KafkaStreamSpec,
                              checkpointMsgSerde: Serde[Checkpoint] = new CheckpointSerde,
                              checkpointKeySerde: Serde[KafkaCheckpointLogKey] = new KafkaCheckpointLogKeySerde) extends CheckpointManager with Logging {
 
+  var MAX_RETRIES_ON_FAILURE = 50
```
prateekm: Make this a val. Move to line 152, where the retry loop is.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
shanthoosh: Updated the patch in accordance to above comment.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
shanthoosh: I want to override retries in test(set to 1)(since it's relevant only to one api, i don't think it's logical to inject it through the constructor).

prateekm: It's fine to add it to the constructor. That's what they're for - setting up the class' state.

If you really want to keep this as a field, Scala doesn't use the ALL_CAPS convention for naming constants. Change to AllCaps instead.

shanthoosh: Changed to AllCaps.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -53,6 +53,8 @@ class KafkaCheckpointManager(checkpointSpec: KafkaStreamSpec,
                              checkpointMsgSerde: Serde[Checkpoint] = new CheckpointSerde,
                              checkpointKeySerde: Serde[KafkaCheckpointLogKey] = new KafkaCheckpointLogKeySerde) extends CheckpointManager with Logging {
 
+  var maxRetriesOnFailure = 50
```
prateekm: 'MaxRetriesOnFailure'. Also, add comment saying that this is ~83 minutes. Otherwise LGTM.

shanthoosh: Done.

PR #421, opened 2018-02-10
Support source types where the last part of the source is not the streamName

srinipunuru:
Contains following fixes

1. Right now Samza SQL framework assumes that the last part of the source is the stream Name, removed the assumption
2. Made consoleLoggingSystemFactory to log formatted json so that it's easily readable.
3. Added support in SamzaSqlRelMessage where the key may not be present.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
@@ -21,7 +21,6 @@ package org.apache.samza.sql.data;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
 import java.util.Optional;
 import org.apache.commons.lang.Validate;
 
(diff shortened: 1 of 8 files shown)


samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -42,6 +42,14 @@
   private final List<String> names = new ArrayList<>();
   private final Object key;
 
+  public SamzaSqlRelMessage(List<String> names, List<Object> values) {
```
misanc: nit: javadoc por public constructor

question: Looking at static method "createRelMessage(List<>, List<>)" seems to me confusing to have two methods with the same signature with different behavior on how the object is created.
      Consider making this constructor private, and converting "createRelMessage" to a constructor.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -42,6 +42,14 @@
...
+    Validate.isTrue(names.size() == values.size(), "Field Names and values are not of same length.");
```
misanc: Can you just call the other constructor passing the key as null?
That way we remove some duplicate code.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -56,9 +64,7 @@ public SamzaSqlRelMessage(Object key, List<String> names, List<Object> values) {
     this.key = key;
     this.value.addAll(values);
     this.names.addAll(names);
-    if (key != null) {
-      this.relFieldValues.add(key);
-    }
+    this.relFieldValues.add(key);
```
misanc: question: it is ok to add a null key to a message?

nitpick: can you add comment to the code (maybe at the getter level) of what is  the difference between FieldValues() and RelFieldValues(). Seems like the latter is just  values array prefixed by the key value ([key] + values). Why we need both?

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -113,13 +119,13 @@ public static SamzaSqlRelMessage createRelMessage(List<Object> fieldValues, List
     int keyIndex = fieldNames.indexOf(KEY_NAME);
```
misanc: nitpick: order of parameters: In all the other methods you have fieldNames first, why here the order is different?

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
srinipunuru: Good feedback, The SamzaSqlRelMessage class was getting bit confusing, I did some refactoring to make it simple. Hopefully the newer version is easier to understand.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -37,49 +36,50 @@
...
-   * Create the SamzaSqlRelMessage, Each rel message represents a row in the table.
-   * So it can contain a key and a list of fields in the row.
-   * @param key Represents the key in the row, Key is optional, in which case it can be null.
-   * @param names Ordered list of field names in the row.
-   * @param values Ordered list of all the values in the row. Since the samzaSqlRelMessage can represent
-   *               the row in a change capture event stream, It can contain delete messages in which case
-   *               all the fields in the row can be null.
+   * Creates a {@link SamzaSqlRelMessage} from the list of relational fields and values.
+   * If the field list contains KEY, then it extracts the key out of the fields to creates a
+   * RelMessage with key and values otherwise creates a Relmessage without the key.
```
atoomula: Could you add a comment saying that key could be null and fieldValues could be null ? This would be the case for outer joins.

PR #422, opened 2018-02-15
SAMZA-1588: Add random jitter to monitor’s scheduling interval.

shanthoosh:
We’ve observed in LinkedIn execution environments that, all the monitors running on the YARN node-manager machines hitting an external service at the same time based upon the configured monitor scheduling interval.

To eliminate unnecessary monitor execution spike and congestion caused to an external service at the same time, it’s essential to add a random jitter to the monitor scheduling interval.

Random jitter will be added to monitor scheduling interval based upon a boolean configuration.

### DIFF
--- a/samza-rest/src/main/java/org/apache/samza/monitor/MonitorConfig.java
+++ b/samza-rest/src/main/java/org/apache/samza/monitor/MonitorConfig.java
@@ -42,6 +42,10 @@ public class MonitorConfig extends MapConfig {
 
   private static final String MONITOR_PREFIX = String.format("monitor%s", MONITOR_CONFIG_KEY_SEPARATOR);
 
+  private static final String CONFIG_SCHEDULING_JITTER_PERCENT = "scheduling.jitter.percent";
+
+  private static final int DEFAULT_SCHEDULING_JITTER_PERCENT = 0;
+
   public MonitorConfig(Config config) {
     super(config);
   }
@@ -95,5 +99,9 @@ public class MonitorConfig extends MapConfig {
    */
   public int getSchedulingIntervalInMs() {
     return getInt(CONFIG_SCHEDULING_INTERVAL, DEFAULT_SCHEDULING_INTERVAL_IN_MS);
+  }
+
+  public double getSchedulingJitterPercent() {
+    return getDouble(CONFIG_SCHEDULING_JITTER_PERCENT, DEFAULT_SCHEDULING_JITTER_PERCENT);
   }
 }
--- a/samza-rest/src/main/java/org/apache/samza/monitor/SamzaMonitorService.java
+++ b/samza-rest/src/main/java/org/apache/samza/monitor/SamzaMonitorService.java
@@ -19,6 +19,7 @@
 package org.apache.samza.monitor;
 
 import com.google.common.base.Strings;
+import java.security.SecureRandom;
 import java.util.Map;
 import org.apache.samza.SamzaException;
 import org.apache.samza.metrics.MetricsRegistry;
@@ -40,6 +41,7 @@ import static org.apache.samza.monitor.MonitorLoader.instantiateMonitor;
 public class SamzaMonitorService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(SamzaMonitorService.class);
+    private static final SecureRandom RANDOM = new SecureRandom();
 
     private final SchedulingProvider scheduler;
     private final SamzaRestConfig config;
@@ -62,9 +64,9 @@ public class SamzaMonitorService {
 
                 if (!Strings.isNullOrEmpty(monitorConfig.getMonitorFactoryClass())) {
                     int schedulingIntervalInMs = monitorConfig.getSchedulingIntervalInMs();
-                    LOGGER.info("Scheduling monitor {} to run every {} ms", monitorName, schedulingIntervalInMs);
-                    // MetricsRegistry has been added in the Monitor interface, since it's required in the eventual future to record metrics.
-                    // We have plans to record metrics, hence adding this as a placeholder. We just aren't doing it yet.
+                    int monitorSchedulingJitterInMs = (int) (RANDOM.nextInt(schedulingIntervalInMs + 1) * (monitorConfig.getSchedulingJitterPercent() / 100.0));
+                    schedulingIntervalInMs += monitorSchedulingJitterInMs;
+                    LOGGER.info("Scheduling the monitor: {} to run every {} ms.", monitorName, schedulingIntervalInMs);
                     scheduler.schedule(getRunnable(instantiateMonitor(monitorName, monitorConfig, metricsRegistry)),
                                                    schedulingIntervalInMs);
                 } else {
--- a/docs/learn/documentation/versioned/rest/monitors.md
+++ b/docs/learn/documentation/versioned/rest/monitors.md
@@ -47,6 +47,14 @@ The following configurations are required for each of the monitors.
             not defined, it is defaulted to 60 seconds.</td>
           </tr>
           <tr>
+            <td>monitor.monitorName.scheduling.jitter.percent</td>
+            <td></td>
+            <td>
+            Defines the random jitter percentage that should be added to the monitor
+            scheduling interval for a monitor named monitorName. If undefined,
+            it is defaulted to zero.</td>
+          </tr>
+          <tr>
             <td>monitor.monitorName.factory.class</td>
             <td></td>
             <td>


shanthoosh:
@jmakes @xinyuiscool 

samza-rest/src/main/java/org/apache/samza/monitor/SamzaMonitorService.java
```diff
@@ -65,6 +66,15 @@ public void start() {
                     LOGGER.info("Scheduling monitor {} to run every {} ms", monitorName, schedulingIntervalInMs);
```
prateekm: This log message is incorrect now.

samza-rest/src/main/java/org/apache/samza/monitor/MonitorConfig.java
```diff
@@ -42,6 +42,8 @@
 
   private static final String MONITOR_PREFIX = String.format("monitor%s", MONITOR_CONFIG_KEY_SEPARATOR);
 
+  private static final String CONFIG_JITTER_ENABLED = "jitter.enabled";
```
prateekm: Make this config a percent b/w 0 to 100, default to 0, and add that amount of jitter to the scheduling interval.

Also, namespace the config correctly. This is a scheduling related config, so should be scheduling.jitter.percent or something.

samza-rest/src/main/java/org/apache/samza/monitor/SamzaMonitorService.java
```diff
@@ -65,6 +66,15 @@ public void start() {
...
                     // MetricsRegistry has been added in the Monitor interface, since it's required in the eventual future to record metrics.
                     // We have plans to record metrics, hence adding this as a placeholder. We just aren't doing it yet.
+
+                    /**
+                     * Add a random jitter from the range [0, schedulingIntervalInMs] to the monitor
+                     * scheduling interval.
+                     */
+                    if (monitorConfig.isJitterEnabled()) {
```
prateekm: Not sure why you need SecureRandom, just use Random. Also, make it a private static final.

If nextInt returns an int, it can be negative. With 100% jitter range, you can end up scheduling tasks that run every 1 ms, even though they were meant to run every x minutes/hours/days. Clearly not what we want. Use a jitter.percent config instead.

prateekm: Correction: in the second part, nextInt it's b/w 0 and upper bound. So it can be up to 2x the original amount. Still better to use %.

samza-rest/src/main/java/org/apache/samza/monitor/MonitorConfig.java
shanthoosh: Done. Added a jitter percent which gets multiplied with random number.

samza-rest/src/main/java/org/apache/samza/monitor/SamzaMonitorService.java
shanthoosh: Fixed.

samza-rest/src/main/java/org/apache/samza/monitor/SamzaMonitorService.java
shanthoosh: - Added a jitter percent, which controls the randomness that gets added to scheduling time.
- SecureRandom is a strong random number generator where as Random is pseudorandom number generator. It's better and proven to provide a more uniform distribution of a random numbers in a interval. 
- Made it as a static constant.

samza-rest/src/main/java/org/apache/samza/monitor/MonitorConfig.java
```diff
@@ -96,4 +98,11 @@ public String getMonitorFactoryClass() {
   public int getSchedulingIntervalInMs() {
     return getInt(CONFIG_SCHEDULING_INTERVAL, DEFAULT_SCHEDULING_INTERVAL_IN_MS);
   }
+
+  public double getSchedulingJitterPercent() {
+    if (containsKey(CONFIG_SCHEDULING_JITTER_PERCENT)) {
+      return getDouble(CONFIG_SCHEDULING_JITTER_PERCENT) / 100.0;
+    }
+    return 0;
```
prateekm: Should be a constant like DEFAULT_SCHEDULING_INTERVAL_IN_MS.

samza-rest/src/main/java/org/apache/samza/monitor/MonitorConfig.java
```diff
@@ -96,4 +98,11 @@ public String getMonitorFactoryClass() {
   public int getSchedulingIntervalInMs() {
     return getInt(CONFIG_SCHEDULING_INTERVAL, DEFAULT_SCHEDULING_INTERVAL_IN_MS);
   }
+
+  public double getSchedulingJitterPercent() {
+    if (containsKey(CONFIG_SCHEDULING_JITTER_PERCENT)) {
+      return getDouble(CONFIG_SCHEDULING_JITTER_PERCENT) / 100.0;
```
prateekm: Config is already a percent, if you divide by 100 and return that's not the jitter percent anymore. Let's do the division at the call site.

samza-rest/src/main/java/org/apache/samza/monitor/SamzaMonitorService.java
```diff
@@ -62,9 +64,16 @@ public void start() {
 
                 if (!Strings.isNullOrEmpty(monitorConfig.getMonitorFactoryClass())) {
                     int schedulingIntervalInMs = monitorConfig.getSchedulingIntervalInMs();
-                    LOGGER.info("Scheduling monitor {} to run every {} ms", monitorName, schedulingIntervalInMs);
                     // MetricsRegistry has been added in the Monitor interface, since it's required in the eventual future to record metrics.
                     // We have plans to record metrics, hence adding this as a placeholder. We just aren't doing it yet.
+
+                    /**
+                     *  Jitter is generated by multiplying a random number from the range([0, schedulingIntervalInMs]) with
```
prateekm: Can remove comment, does not add anything to what the next line already describes.

samza-rest/src/main/java/org/apache/samza/monitor/SamzaMonitorService.java
```diff
@@ -62,9 +64,16 @@ public void start() {
 
                 if (!Strings.isNullOrEmpty(monitorConfig.getMonitorFactoryClass())) {
                     int schedulingIntervalInMs = monitorConfig.getSchedulingIntervalInMs();
-                    LOGGER.info("Scheduling monitor {} to run every {} ms", monitorName, schedulingIntervalInMs);
                     // MetricsRegistry has been added in the Monitor interface, since it's required in the eventual future to record metrics.
```
prateekm: Can remove this comment too.

samza-rest/src/main/java/org/apache/samza/monitor/SamzaMonitorService.java
```diff
@@ -40,6 +41,7 @@
 public class SamzaMonitorService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(SamzaMonitorService.class);
+    private static final SecureRandom RAND = new SecureRandom();
```
prateekm: Nitpick: RANDOM

samza-rest/src/main/java/org/apache/samza/monitor/SamzaMonitorService.java
shanthoosh: Removed.

samza-rest/src/main/java/org/apache/samza/monitor/SamzaMonitorService.java
shanthoosh: Changed the constant from RAND to RANDOM.

samza-rest/src/main/java/org/apache/samza/monitor/SamzaMonitorService.java
shanthoosh: Removed the comment.

samza-rest/src/main/java/org/apache/samza/monitor/MonitorConfig.java
shanthoosh: Fixed.

samza-rest/src/main/java/org/apache/samza/monitor/MonitorConfig.java
shanthoosh: Fixed.

samza-rest/src/main/java/org/apache/samza/monitor/MonitorConfig.java
```diff
@@ -96,4 +100,11 @@ public String getMonitorFactoryClass() {
   public int getSchedulingIntervalInMs() {
     return getInt(CONFIG_SCHEDULING_INTERVAL, DEFAULT_SCHEDULING_INTERVAL_IN_MS);
   }
+
+  public double getSchedulingJitterPercent() {
+    if (containsKey(CONFIG_SCHEDULING_JITTER_PERCENT)) {
+      return getDouble(CONFIG_SCHEDULING_JITTER_PERCENT);
```
prateekm: There's no getDouble(key, default) like the getInt version above?

shanthoosh: Switched to `getDouble(key, default) ` version.

PR #423, opened 2018-02-17
SAMZA-1460: StreamAppender does not explicitly create logging topic

dnishimura:
Creates the StreamAppender stream explicitly instead of relying on auto stream creation.

### DIFF
--- a/docs/learn/documentation/versioned/jobs/logging.md
+++ b/docs/learn/documentation/versioned/jobs/logging.md
@@ -116,12 +116,14 @@ And then updating your log4j.xml to include the appender:
 
 #### Stream Log4j Appender
 
-Samza provides a StreamAppender to publish the logs into a specific system. You can specify the system name using "task.log4j.system" and change name of log stream with param 'StreamName'. Also, we have the [MDC](http://logback.qos.ch/manual/mdc.html) keys "containerName", "jobName" and "jobId", which help identify the source of the log. In order to use this appender, simply add:
+Samza provides a StreamAppender to publish the logs into a specific system. You can specify the system name using "task.log4j.system" and change name of log stream with param 'StreamName'. You can also specify the number of partitions for the log stream with param 'PartitionCount'; otherwise, the number of partitions will equal the number of containers configured for the job. Also, we have the [MDC](http://logback.qos.ch/manual/mdc.html) keys "containerName", "jobName" and "jobId", which help identify the source of the log. In order to use this appender, simply add:
 
 {% highlight xml %}
 <appender name="StreamAppender" class="org.apache.samza.logging.log4j.StreamAppender">
    <!-- optional -->
    <param name="StreamName" value="EpicStreamName"/>
+   <!-- optional -->
+   <param name="PartitionCount" value="8"/>
    <layout class="org.apache.log4j.PatternLayout">
      <param name="ConversionPattern" value="%X{containerName} %X{jobName} %X{jobId} %d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %c{1} [%p] %m%n" />
    </layout>
--- a/docs/learn/documentation/versioned/jobs/logging.md
+++ b/docs/learn/documentation/versioned/jobs/logging.md
@@ -116,14 +116,14 @@ And then updating your log4j.xml to include the appender:
 
 #### Stream Log4j Appender
 
-Samza provides a StreamAppender to publish the logs into a specific system. You can specify the system name using "task.log4j.system" and change name of log stream with param 'StreamName'. You can also specify the number of partitions for the log stream with param 'PartitionCount'; otherwise, the number of partitions will equal the number of containers configured for the job. Also, we have the [MDC](http://logback.qos.ch/manual/mdc.html) keys "containerName", "jobName" and "jobId", which help identify the source of the log. In order to use this appender, simply add:
+Samza provides a StreamAppender to publish the logs into a specific system. You can specify the system name using "task.log4j.system" and change name of log stream with param 'StreamName'. You can also specify the number of partitions for the log stream with param 'DefaultPartitionCount'; otherwise, the number of partitions will equal the number of containers configured for the job. The partition count is set upon the creation of the logging stream and changing the partition count requires manual intervention with the system stream. The [MDC](http://logback.qos.ch/manual/mdc.html) contains the keys "containerName", "jobName" and "jobId", which help identify the source of the log. In order to use this appender, add:
 
 {% highlight xml %}
 <appender name="StreamAppender" class="org.apache.samza.logging.log4j.StreamAppender">
    <!-- optional -->
    <param name="StreamName" value="EpicStreamName"/>
    <!-- optional -->
-   <param name="PartitionCount" value="8"/>
+   <param name="DefaultPartitionCount" value="8"/>
    <layout class="org.apache.log4j.PatternLayout">
      <param name="ConversionPattern" value="%X{containerName} %X{jobName} %X{jobId} %d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %c{1} [%p] %m%n" />
    </layout>
--- a/docs/learn/documentation/versioned/jobs/logging.md
+++ b/docs/learn/documentation/versioned/jobs/logging.md
@@ -116,14 +116,14 @@ And then updating your log4j.xml to include the appender:
 
 #### Stream Log4j Appender
 
-Samza provides a StreamAppender to publish the logs into a specific system. You can specify the system name using "task.log4j.system" and change name of log stream with param 'StreamName'. You can also specify the number of partitions for the log stream with param 'DefaultPartitionCount'; otherwise, the number of partitions will equal the number of containers configured for the job. The partition count is set upon the creation of the logging stream and changing the partition count requires manual intervention with the system stream. The [MDC](http://logback.qos.ch/manual/mdc.html) contains the keys "containerName", "jobName" and "jobId", which help identify the source of the log. In order to use this appender, add:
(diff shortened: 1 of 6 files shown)


samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -289,4 +292,8 @@ public static StreamSpec createCoordinatorStreamSpec(String physicalName, String
   public static StreamSpec createCheckpointStreamSpec(String physicalName, String systemName) {
     return new StreamSpec(CHECKPOINT_STREAM_ID, physicalName, systemName, 1);
   }
+
+  public static StreamSpec createStreamAppenderSpec(String physicalName, String systemName, int partitionCount) {
```
jmakes: nit: for consistency "createStreamAppenderStreamSpec"

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
```diff
@@ -260,6 +263,12 @@ protected void setupSystem() {
 
     setSerde(log4jSystemConfig, systemName, streamName);
 
+    // Explicitly create stream appender stream with the partition count the same as the number of containers.
+    JobConfig jobConfig = new JobConfig(config);
+    StreamSpec streamSpec = StreamSpec.createStreamAppenderSpec(streamName, systemName, jobConfig.getContainerCount());
```
jmakes: What's the behavior for standalone mode, which doesn't use a predefined container count?
Perhaps use the number of tasks in that case? Or require a config. Or use some reasonable default.

samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
dnishimura: Good catch. My intention was your suggestion. I'll edit that.

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
dnishimura: Good point. I looked into using the number of tasks, but it involves fetching metadata for each input stream's system admin to figure out the max task count. Is there an easier way to get task count? If not, I think I'll opt to require a config with a default of 8. That sound good?

dnishimura: In digging further, the stream appender log stream is keyed by container name, so it looks like the number of partitions should be determined by container count. For standalone, since the number of containers cannot be determined during start up, requiring a config seems to be the only option.

On another note, I'm not sure how StreamAppender even works in standalone because the container names are only set by the `run-container.sh` and `run-jc.sh` scripts: `export JAVA_OPTS="$JAVA_OPTS -Dsamza.container.name=samza-container-$SAMZA_CONTAINER_ID"`. StreamAppender throws an exception if there is no container name set. Standalone mode probably needs more investigation.

jmakes: Thanks for looking into it. Perhaps rather than adding a new config for the container count, it would be more appropriate to add another log4j parameter, analogous to `StreamName` and make both mandatory for standalone.

dnishimura: The `KafkaSystemAdmin.createStream` seems to hang on the ZK connection when called from StreamAppender when running it in `hello-samza`. Need to look into why.

dnishimura: Issue was that the `master` branch of `samza` matches the `latest` branch of `hello-samza`. I was using the `master` branch of `hello-samza`. Seems to work now. Thanks @bharathkk for helping look into this.

docs/learn/documentation/versioned/jobs/logging.md
```diff
@@ -116,12 +116,14 @@ And then updating your log4j.xml to include the appender:
 
 #### Stream Log4j Appender
 
-Samza provides a StreamAppender to publish the logs into a specific system. You can specify the system name using "task.log4j.system" and change name of log stream with param 'StreamName'. Also, we have the [MDC](http://logback.qos.ch/manual/mdc.html) keys "containerName", "jobName" and "jobId", which help identify the source of the log. In order to use this appender, simply add:
+Samza provides a StreamAppender to publish the logs into a specific system. You can specify the system name using "task.log4j.system" and change name of log stream with param 'StreamName'. You can also specify the number of partitions for the log stream with param 'PartitionCount'; otherwise, the number of partitions will equal the number of containers configured for the job. Also, we have the [MDC](http://logback.qos.ch/manual/mdc.html) keys "containerName", "jobName" and "jobId", which help identify the source of the log. In order to use this appender, simply add:
```
prateekm: What happens when the number of containers changes (fairly common)?

PS: Unrelated but lets reword the "Also, we have the" to "The MDC contains the keys ... which may help identify ...". Also remove "simply"

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
```diff
@@ -70,6 +73,7 @@
   private SystemProducer systemProducer = null;
   private String key = null;
   private String streamName = null;
+  private int partitionCount = 0;
```
prateekm: Why is this a field on the StreamAppender?

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
```diff
@@ -93,6 +97,17 @@ public void setStreamName(String streamName) {
...
   }
 
+  public int getPartitionCount() {
+    if (partitionCount > 0) {
+      return partitionCount;
+    }
+    return new JobConfig(getConfig()).getContainerCount();
+  }
+
+  public void setPartitionCount(int partitionCount) {
```
prateekm: Can this be passed in as part of the constructor or the setupSystem method? Not a big fan of order dependent state mutations.

Also, where/how is this set to the user specified value? Does this happen automagically if the field name matches the xml property name?

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
```diff
@@ -260,6 +275,11 @@ protected void setupSystem() {
 
     setSerde(log4jSystemConfig, systemName, streamName);
 
+    // Explicitly create stream appender stream with the partition count the same as the number of containers.
+    StreamSpec streamSpec = StreamSpec.createStreamAppenderStreamSpec(streamName, systemName, getPartitionCount());
```
prateekm: Log at info that we're creating the log stream x on system y with num partitions n.

docs/learn/documentation/versioned/jobs/logging.md
dnishimura: Nothing changes. Similar to how StreamAppender topics get auto-created. If the logging system is Kafka, the # of partitions is the default partition count configured on Kafka. I'll add that additional description and reword the other parts as well.

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
dnishimura: For log4j appenders, when you set a property in the log4j.xml file, it uses the set<property name> and get<property name> as getter/setter automagically.

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
dnishimura: Sounds good will do.

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
dnishimura: Lets the user set the partition count on the StreamAppender logging topic. In yarn, we could use the container count only, but in standalone deployments, we can't deduce the container count on startup. It was a suggestion from @jmakes and I think it's the only way to support both yarn and standalone deployments for now.

dnishimura:
@prateekm Added updates w.r.t. your feedback.

docs/learn/documentation/versioned/jobs/logging.md
```diff
@@ -116,12 +116,14 @@ And then updating your log4j.xml to include the appender:
...
+Samza provides a StreamAppender to publish the logs into a specific system. You can specify the system name using "task.log4j.system" and change name of log stream with param 'StreamName'. You can also specify the number of partitions for the log stream with param 'DefaultPartitionCount'; otherwise, the number of partitions will equal the number of containers configured for the job. The partition count is set upon the creation of the logging stream and changing the partition count requires manual intervention with the system stream. The [MDC](http://logback.qos.ch/manual/mdc.html) contains the keys "containerName", "jobName" and "jobId", which help identify the source of the log. In order to use this appender, add:
 
 {% highlight xml %}
 <appender name="StreamAppender" class="org.apache.samza.logging.log4j.StreamAppender">
    <!-- optional -->
    <param name="StreamName" value="EpicStreamName"/>
+   <!-- optional -->
+   <param name="DefaultPartitionCount" value="8"/>
```
prateekm: I think PartitionCount would be fine since you described the behavior above. DefaultPartitionCount isn't much more descriptive.

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
```diff
@@ -93,6 +97,17 @@ public void setStreamName(String streamName) {
...
   }
 
+  public int getDefaultPartitionCount() {
+    if (defaultPartitionCount > 0) {
+      return defaultPartitionCount;
+    }
+    return new JobConfig(getConfig()).getContainerCount();
+  }
+
+  public void setDefaultPartitionCount(int defaultPartitionCount) {
```
prateekm: Can you add javadoc to this method to explain up to what point in the container lifecycle may this be called? In other words, when is activateOptions called?

dnishimura:
@prateekm minor edits done. Thanks for approving this PR.

vjagadish1989:
merged and submitted!

PR #425, opened 2018-02-22
Add stream-table join support for samza sql

atoomula:


samza-sql/src/main/java/org/apache/samza/sql/serializers/SamzaSqlCompositeKeySerdeFactory.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.samza.serializers.SerdeFactory;
+import org.apache.samza.sql.data.SamzaSqlCompositeKey;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.codehaus.jackson.type.TypeReference;
+
+
+/**
+ * A serializer for SamzaSqlCompositeKey.
+ */
+public final class SamzaSqlCompositeKeySerdeFactory implements SerdeFactory<SamzaSqlCompositeKey> {
```
misanc: Consider adding a Generic JsonSerdeFactory<T> that implements the common function from all the json serders we could have.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlCompositeKey.java
```diff
@@ -0,0 +1,76 @@
...
+
+
+/**
+ * A serializable class that holds different key parts.
+ */
+public class SamzaSqlCompositeKey implements Serializable {
+
+  @JsonProperty("keyParts")
+  private ArrayList<Object> keyParts;
+  int hashCode;
```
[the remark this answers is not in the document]
vjagadish1989: @misanc : Usually, if there's a cpu-time trade-off, it's worthwhile to cache hash-codes.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -32,12 +34,15 @@
  * SamzaSqlRelationalMessage, In future if we find a need, we could add additional column ddl metadata around
  * primary Key, nullability, etc.
  */
-public class SamzaSqlRelMessage {
+public class SamzaSqlRelMessage implements Serializable {
```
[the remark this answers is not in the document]
atoomula: This is for adding the rel msg to table.

samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedSourceResolverFactory.java
```diff
@@ -44,6 +44,7 @@ public SourceResolver create(Config config) {
   }
```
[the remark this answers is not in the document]
atoomula: Sure

samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedSourceResolverFactory.java
```diff
@@ -53,17 +54,35 @@ public ConfigBasedSourceResolver(Config config) {
...
+          String msg = String.format("Source %s is not of the format [table.]{systemName}.{streamName{", source);
+          LOG.error(msg);
+          throw new SamzaException(msg);
+        }
       }
-      String systemName = sourceComponents[0];
-      String streamName = sourceComponents[1];
 
-      return new SqlSystemStreamConfig(systemName, streamName, fetchSystemConfigs(systemName));
+      if (sourceComponents[endIdx].toLowerCase().equals(SAMZA_SQL_QUERY_TABLE_KEYWORD)) {
```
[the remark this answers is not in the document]
atoomula: This was the result of hastily changing the format from the beginning to the end after design review. Sorry about that! It should be fixed in the latest version.

samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedSourceResolverFactory.java
```diff
@@ -53,17 +54,35 @@ public ConfigBasedSourceResolver(Config config) {
...
+        isTable = true;
+        streamIdx = endIdx - 1;
+      }
+
+      String systemName = sourceComponents[systemIdx];
```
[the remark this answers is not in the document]
atoomula: If we have multiple levels, then the streamName is either sourceParts.length - 1 or sourceParts.length - 2. We cannot represent it as a constant.

samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedSourceResolverFactory.java
```diff
@@ -53,17 +54,35 @@ public ConfigBasedSourceResolver(Config config) {
...
+      String streamName = sourceComponents[streamIdx];
+
+      return new SqlSystemStreamConfig(systemName, streamName, fetchSystemConfigs(systemName), isTable);
+    }
+
+    @Override
+    public boolean isTable(String sourceName) {
+      String[] sourceComponents = sourceName.split("\\.");
+      return sourceComponents[sourceComponents.length - 1].toLowerCase().equals(SAMZA_SQL_QUERY_TABLE_KEYWORD);
```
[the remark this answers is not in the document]
atoomula: Fixed. It should be at the end.

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlSystemStreamConfig.java
```diff
@@ -36,6 +36,8 @@
```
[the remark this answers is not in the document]
atoomula: Sure

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlSystemStreamConfig.java
```diff
@@ -52,11 +54,15 @@
   private List<String> sourceParts;
 
   public SqlSystemStreamConfig(String systemName, String streamName, Config systemConfig) {
-    this(systemName, streamName, Arrays.asList(systemName, streamName), systemConfig);
+    this(systemName, streamName, Arrays.asList(systemName, streamName), systemConfig, false);
+  }
+
+  public SqlSystemStreamConfig(String systemName, String streamName, Config systemConfig, boolean isTable) {
```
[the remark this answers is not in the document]
atoomula: This is a moot point. Samza configs still represent a table as a stream and the whole concept is around system streams including the configs. I'm ok either ways but to make it consistent with Samza core, I'm for now leaving it as system stream.

samza-sql/src/main/java/org/apache/samza/sql/serializers/SamzaSqlCompositeKeySerdeFactory.java
atoomula: Let me talk to you on this.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -49,9 +54,13 @@
    *               delete change capture event in the stream or because of the result of the outer join or the fields
    *               themselves are null in the original stream.
    */
-  public SamzaSqlRelMessage(List<String> fieldNames, List<Object> fieldValues) {
+  public SamzaSqlRelMessage(@JsonProperty("fieldNames") List<String> fieldNames,
+      @JsonProperty("fieldValues") List<Object> fieldValues) {
     Validate.isTrue(fieldNames.size() == fieldValues.size(), "Field Names and values are not of same length.");
 
+    this.fieldNames = new ArrayList<>();
+    this.fieldValues = new ArrayList<>();
```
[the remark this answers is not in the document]
atoomula: Sure

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlCompositeKey.java
```diff
@@ -0,0 +1,76 @@
...
+  public int hashCode() {
+    return hashCode;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    return this == o || o != null && getClass() == o.getClass() && keyParts.equals(((SamzaSqlCompositeKey) o).keyParts);
+  }
+
+  public static SamzaSqlCompositeKey createSamzaSqlCompositeKey(SamzaSqlRelMessage message, List<Integer> relIdx) {
```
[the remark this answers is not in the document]
atoomula: Sure

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+    SamzaSqlCompositeKeySerde keySerde =
+        (SamzaSqlCompositeKeySerde) new SamzaSqlCompositeKeySerdeFactory().getSerde(null, null);
+    SamzaSqlRelMessageSerde relMsgSerde =
+        (SamzaSqlRelMessageSerde) new SamzaSqlRelMessageSerdeFactory().getSerde(null, null);
+
+    // Create a table backed by RocksDb store with the fields in the join condition as composite key and relational
+    // message as the value. Send the messages from the input stream denoted as 'table' to the created table store.
+    Table<KV<SamzaSqlCompositeKey, SamzaSqlRelMessage>> table =
+        context.getStreamGraph()
+            .getTable(new RocksDbTableDescriptor("table_" + joinId)
```
misanc: Can you add in the class comment, that only RocksDb tables are supported at the moment (external tables is a TODO)

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+    validateJoinQuery(join);
+
+    boolean isTablePosOnRight = isTable(join.getRight());
+    List<Integer> streamIds = new LinkedList<>();
+    List<Integer> tableIds = new LinkedList<>();
+
+    // Get the stream and table indices corresponding to the fields given in the join condition.
+    getStreamAndTableKeyIds(((RexCall) join.getCondition()).getOperands(), join, isTablePosOnRight, streamIds, tableIds);
+
+    MessageStream<SamzaSqlRelMessage> inputTable =
```
misanc: consider moving  lines 94 to 112 to its own method.

This is the logic for loading the table to RocksDb; but that can change depending of the type of store we use in the future. Could be nice to keep this part of the code cleaner of that detail.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+ * Translator to translate the LogicalJoin node in the relational graph to the corresponding StreamGraph
+ * implementation.
+ * Join is supported with the following caveats:
+ *   1. Only stream-table joins are supported. No stream-stream joins.
+ *   2. Only Equi-joins are supported. No theta-joins.
+ *   3. Inner joins, Left and Right outer joins are supported. No cross joins, full outer joins or natural joins.
+ *   4. Compound join condition with only AND operator is supported. AND operator with a literal is not supported. No
+ *      support for OR operator or any other operator in the join condition.
+ *
+ * It is assumed that the stream denoted as 'table' is already partitioned by the key(s) specified in the join
```
misanc: This assumption could be problematic. Can you detect this is not the case and throw an error if that is the case?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+ * condition. We do not repartition the table. But we always repartition the stream by the key(s) specified
+ * in the join condition.
+ */
```
misanc: What about the use case of broadcast table?. (e.g. I want to load in all nodes some city information, and add details of the city to each incoming message).

Seems like we are limiting the type of use case this code can be use for. 

I suggest to create a document, explain what is cover what is not cover, and when are we planning to add the non covered use cases.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+    // Create a table backed by RocksDb store with the fields in the join condition as composite key and relational
+    // message as the value. Send the messages from the input stream denoted as 'table' to the created table store.
+    Table<KV<SamzaSqlCompositeKey, SamzaSqlRelMessage>> table =
+        context.getStreamGraph()
+            .getTable(new RocksDbTableDescriptor("table_" + joinId)
+                .withSerde(KVSerde.of(keySerde, relMsgSerde)));
+
+    inputTable
+        .map(m -> new KV(createSamzaSqlCompositeKey(m, tableIds), m))
+        .sendTo(table);
```
misanc: can you throw a error message, or increase a counter. If the hash of the composite key does not match the hash assigned to this task (breaking one of the assumption in the class comment)

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+
+    // Always re-partition the messages from the input stream by the composite key and then join the messages
+    // with the table.
+    MessageStream<SamzaSqlRelMessage> outputStream =
+        inputStream
+            .partitionBy(m -> createSamzaSqlCompositeKey(m, streamIds),
+                m -> m,
+                KVSerde.of(keySerde, relMsgSerde),
+                "stream_" + joinId)
+            .map(KV::getValue)
```
misanc: Seems wasteful to me to drop the Key here, and recompute it in the next step (inside  StreamTableJoinFunction::getMessageKey())

But I do not see an easy solution. Can you think how to avoid this duplication?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+    this.joinId = joinId;
+    this.sourceResolver = sourceResolver;
+  }
+
+  public void translate(final LogicalJoin join, final TranslatorContext context) {
+
+    // Do the validation of join query
+    validateJoinQuery(join);
+
+    boolean isTablePosOnRight = isTable(join.getRight());
```
misanc: Assumption: can you just make the assumption that the table is on the right, and if that is not the case then just create a new LogicalJoin with the elements swap? This could simplify the code

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+          joinRelType.compareTo(JoinRelType.INNER) == 0);
+      this.streamFieldIds = streamFieldIds;
+      this.tableFieldNames = tableFieldNames;
+    }
+
+    @Override
+    public SamzaSqlRelMessage apply(SamzaSqlRelMessage message, KV<SamzaSqlCompositeKey, SamzaSqlRelMessage> record) {
+
+      if (joinRelType.compareTo(JoinRelType.INNER) == 0 && record == null) {
+        log.debug("Record not found for the message with key: " + getMessageKey(message));
```
misanc: nit: can you include in the debug message that is for an "inner join"

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+        return null;
+      }
+
+      // The resulting join output should be a SamzaSqlRelMessage containing the fields from both the stream message and
+      // table record. The order of stream message fields and table record fields are dictated by the sql query. The
+      // output should also include the keys from both the stream message and the table record.
+      List<String> outFieldNames = new ArrayList<>();
```
misanc: Is not clear to me, but if  "outFieldNames" is the same for all result messages, consider moving this computation outside apply().

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    if (condition.getKind() != SqlKind.EQUALS && condition.getKind() != SqlKind.AND) {
+      throw new SamzaException("Only equi-joins and AND operator is supported in join condition.");
+    }
+  }
+
+  // Get the stream and table indices corresponding to the fields given in the join condition by parsing through
+  // the condition.
+  private void getStreamAndTableKeyIds(List<RexNode> operands, final LogicalJoin join, boolean isTablePosOnRight,
```
srinipunuru: Can you add details around where the output is populated to? And what the format of the output is?

srinipunuru: Are the streamids and tableIds size same. Do we support comparison to a constant? If so how does this work?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+      List<Integer> streamIds, List<Integer> tableIds) {
```
srinipunuru: Are these streamKeyIds and tableKeyIds? Can you rename if that is the case, Streamid and tableId have a different meaning in Samza so it might be confusing.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+    // a.key = "constant" is not yet supported.
+    if (!(operands.get(0) instanceof RexInputRef) || !(operands.get(1) instanceof RexInputRef)) {
+      throw new SamzaException("SQL query is not supported. Join condition " + join.getCondition() + " should have "
+          + "reference operands but the types are " + operands.get(0).getClass() + " and " + operands.get(1).getClass());
+    }
+
+    // Join condition is commutative, meaning, a.key = b.key is equivalent to b.key = a.key.
+    // Calcite assigns the indices to the fields based on the order a and b are specified in
+    // the sql 'from' clause. Let's put the operand with smaller index in leftRef and larger
+    // index in rightRef.
```
srinipunuru: How does this work when the query contains join condition like select * from a join b on a.key=b.key and b.field1=a.field1?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+    RexInputRef leftRef = (RexInputRef) operands.get(0);
+    RexInputRef rightRef = (RexInputRef) operands.get(1);
+    if (leftRef.getIndex() > rightRef.getIndex()) {
+      RexInputRef tmpRef = leftRef;
+      leftRef = rightRef;
+      rightRef = tmpRef;
+    }
+
+    // Get the table index and stream index
```
srinipunuru: Please rename this table key index and stream key index.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+
+  private boolean isTable(RelNode relNode) {
+    // NOTE: Any intermediate form of a join is always a stream. Eg: For the second level join of
+    // stream-table-table join, the left side of the join is join output, which we always
+    // assume to be a stream. The intermediate stream won't be an instance of EnumerableTableScan.
+    return relNode instanceof EnumerableTableScan &&
+        sourceResolver.isTable(String.join(".", relNode.getTable().getQualifiedName()));
+  }
+
+  private final class SamzaSqlRelMessageJoinFunction implements StreamTableJoinFunction<SamzaSqlCompositeKey,
```
srinipunuru: Consider separating this into a separate class, So that we can add unit tests targeting specifically the logic in this class.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,319 @@
...
+      SamzaSqlRelMessage, KV<SamzaSqlCompositeKey, SamzaSqlRelMessage>, SamzaSqlRelMessage> {
+
+    JoinRelType joinRelType;
+    boolean isTablePosOnRight;
+    List<Integer> streamFieldIds;
+    // Table field names are used in the outer join when the table record is not found.
+    List<String> tableFieldNames;
+
+    SamzaSqlRelMessageJoinFunction(JoinRelType joinRelType, boolean isTablePosOnRight, List<Integer> streamFieldIds,
```
srinipunuru: Sorry this comment is not related to the code here, But a general comment We dont support left outer join if the left is a table and right outer join if the right is the table. Are we doing the validation for this?

atoomula: Yes - this validation is done in validateJoinQuery().

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: Sure

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: The comment is an attempt to explain that :) Operand with smaller index are collected in leftRef while the other in rightRef. In the example that you gave, a.key and a.field1 are collected in leftRef while b.key and b.field1 are collected in rightRef. Let me see if I can make this clear in the comment.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: Sure

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: Comparison to constant is not supported in the join condition. It is documented at the class level.

atoomula: I will add documentation around the output.

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlSystemStreamConfig.java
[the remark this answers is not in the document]
pdu-mn1: I actually agree with Miguel. In Samza, either low-level or high-level, table is a top-level concept and it's not subclassed from stream. Even though system might make sense in table, "stream" and "systemstream" do not. Besides, for Table, SamzaSQL needs to get TableDescriptor to instantiate tables for both join and sendTo, most likely from SourceResolver. I do think we should create a base class, eg. SqlSourceConfig which only has the common parts. Then make SqlStreamConfig and SqlTableConfig subclasses of SqlSourceConfig with the deltas. 

We can sync up offline about this.

samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedSourceResolverFactory.java
```diff
@@ -53,17 +54,35 @@ public ConfigBasedSourceResolver(Config config) {
...
     public SqlSystemStreamConfig fetchSourceInfo(String source) {
       String[] sourceComponents = source.split("\\.");
+      boolean isTable = false;
+      int systemIdx = 0;
+      int endIdx = sourceComponents.length - 1;
+      int streamIdx = endIdx;
 
-      // This source resolver expects sources of format {systemName}.{streamName}
+      // This source resolver expects sources of format [table.]{systemName}.{streamName}
       if (sourceComponents.length != 2) {
```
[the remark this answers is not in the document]
atoomula: Fixed this.

samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedSourceResolverFactory.java
```diff
@@ -44,26 +46,57 @@ public SourceResolver create(Config config) {
...
-      String streamName = sourceComponents[1];
 
-      return new SqlSystemStreamConfig(systemName, streamName, fetchSystemConfigs(systemName));
+      if (sourceComponents[endIdx].equalsIgnoreCase(SAMZA_SQL_QUERY_TABLE_KEYWORD)) {
+        isTable = true;
+        streamIdx = endIdx - 1;
+      }
+
+      String systemName = sourceComponents[systemIdx];
+      String streamName = sourceComponents[streamIdx];
```
pdu-mn1: nit: seems the above code can be made simpler/cleaner by using regex and group captures?

samza-sql/src/main/java/org/apache/samza/sql/translator/SamzaSqlRelMessageJoinFunction.java
```diff
@@ -0,0 +1,115 @@
...
+import org.apache.samza.operators.functions.StreamTableJoinFunction;
+import org.apache.samza.sql.data.SamzaSqlCompositeKey;
+import org.apache.samza.sql.data.SamzaSqlRelMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.samza.sql.data.SamzaSqlCompositeKey.*;
+
+
+public class SamzaSqlRelMessageJoinFunction implements StreamTableJoinFunction<SamzaSqlCompositeKey, SamzaSqlRelMessage, KV<SamzaSqlCompositeKey, SamzaSqlRelMessage>, SamzaSqlRelMessage> {
```
srinipunuru: We probably need some test cases for this class. 

1.  positive join case.
2. Case where there is no entry in the table and inner join is asked for.
3. Negative cases where
     a. Some of the join keys are null
     b. Payload fields are null

samza-sql/src/main/java/org/apache/samza/sql/translator/SamzaSqlRelMessageJoinFunction.java
```diff
@@ -0,0 +1,115 @@
...
+      outFieldNames.addAll(streamFieldNames);
+    }
+  }
+
+  @Override
+  public SamzaSqlRelMessage apply(SamzaSqlRelMessage message, KV<SamzaSqlCompositeKey, SamzaSqlRelMessage> record) {
+
+    if (joinRelType.compareTo(JoinRelType.INNER) == 0 && record == null) {
+      log.debug("Inner Join: Record not found for the message with key: " + getMessageKey(message));
+      return null;
```
srinipunuru: On innerJoins when you return null, Does this get filtered out? Does the Fluent join function take care of filtering these values out? Can you add a comment if that's the case?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,269 @@
...
+import org.slf4j.LoggerFactory;
+
+import static org.apache.samza.sql.data.SamzaSqlCompositeKey.*;
+
+
+/**
+ * Translator to translate the LogicalJoin node in the relational graph to the corresponding StreamGraph
+ * implementation.
+ * Join is supported with the following caveats:
+ *   0. Only local tables are supported. Remote/composite tables are not yet supported.
```
srinipunuru: Do we support joins 

1. where udf1(a.key) == udf2(b.key) 
2. udf(a.key, b.key) == true
Can you add workitems to support these scenarios.

srinipunuru: Can you also add test case for the above type of joins

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -32,14 +34,17 @@
  * SamzaSqlRelationalMessage, In future if we find a need, we could add additional column ddl metadata around
  * primary Key, nullability, etc.
  */
-public class SamzaSqlRelMessage {
+public class SamzaSqlRelMessage implements Serializable {
```
srinipunuru: Can you add test cases for the serialization and deserialization of this class.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -32,14 +34,17 @@
...
   public static final String KEY_NAME = "__key__";
 
-  private final List<Object> fieldValues = new ArrayList<>();
-  private final List<String> fieldNames = new ArrayList<>();
   private final Object key;
 
+  @JsonProperty("fieldNames")
+  private final List<String> fieldNames;
+  @JsonProperty("fieldValues")
+  private final List<Object> fieldValues;
```
srinipunuru: Can you add a TODO for the workitem to support serialization for the nested SamzaSQLRelMessages?

samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedSourceResolverFactory.java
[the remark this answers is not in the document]
misanc: Seems like you can only have two levels (and optional third part, with the constant $table)
Are you sure that is the case for all system? I thought that for kafka there are more than 2 levels to specify the stream. See: https://iwww.corp.linkedin.com/wiki/cf/display/ENGS/Samza+SQL+Input+and+Output+naming+convention

nitpick: The code works the way you have it. But if you can find a way to remove this extra variables will be easier to read: systemIdx, streamIdx.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
misanc: That is ok, but the current convention for representing a table join will change when all this new feature are introduced.  Can we have a document that explain how Stream Table join will work from the user point of view for the planned use cases we are going to support. I just a little worry that a lot of refactoring will occur to this code as we start adding more cases, and we will have problem migrating previous sql in production.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
misanc: consider to change the signature of SamzaSqlRelMessageJoinFunction to accept KV<...> objects directly as input stream.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: UDFs are not yet supported in join condition. I added a test case for this.

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlSystemSourceConfig.java
```diff
@@ -30,9 +30,13 @@
...
 
 /**
- * Configs associated with a system stream.
+ * Configs associated with a system source. Both streams and table sources are supported.
+ * For now, only local tables are supported.
  */
-public class SqlSystemStreamConfig {
+public class SqlSystemSourceConfig {
+
+  private static final String CFG_SAMZA_STREAM_BOOTSTRAP_STREAM = "streams.%s.samza.bootstrap";
```
nickpan47: Please refer to the constants that are defined in samza-core/src/main/scala/org/apache/samza/config/StreamConfig.scala, instead of creating new static constant strings here.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  void translate(final LogicalJoin join, final TranslatorContext context) {
+
+    // Do the validation of join query
+    validateJoinQuery(join);
+
+    boolean isTablePosOnRight = isTable(join.getRight());
+    List<Integer> streamKeyIds = new LinkedList<>();
+    List<Integer> tableKeyIds = new LinkedList<>();
```
nickpan47: Question: is this under the assumption that translate() is invoked once for each task in init()? Such that the join operator here will be referring to a new instance of List objects under each task context? The upcoming serialization PR is going to change that and only invoke translate() once for each container and I wonder that this instantiation of List objects should be included in join function's init() method instead? Could you confirm the use case of these List objects? Are they supposed to be only used within the join operator in a single task context? Thanks!

samza-sql/src/main/java/org/apache/samza/sql/translator/SamzaSqlRelMessageJoinFunction.java
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * This class joins incoming {@link SamzaSqlRelMessage} from a stream with the records in a table with the join key
+ * being {@link SamzaSqlCompositeKey}
+ */
+public class SamzaSqlRelMessageJoinFunction
+    implements StreamTableJoinFunction<SamzaSqlCompositeKey, SamzaSqlRelMessage, KV<SamzaSqlCompositeKey, SamzaSqlRelMessage>, SamzaSqlRelMessage> {
+
+  private static final Logger log = LoggerFactory.getLogger(SamzaSqlRelMessageJoinFunction.class);
+
+  private final JoinRelType joinRelType;
```
nickpan47: A request here: I don't think that JoinRelType will be serializable. Can we keep in mind that all operator functions that we implement should be serializable?

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlSystemSourceConfig.java
atoomula: Sure. I will reuse here.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: Today, SamzaSqlApplicationRunner's init calls the translate() once for each query on QueryTranslator. Within QueryTranslator translate(), we shuttle through all the joins and call translate on each of these joins. I can talk to you in person about it.

PR #426, opened 2018-02-22
SAMZA-1593: Upgrade gradle nexus plugin.

shanthoosh:
Current nexus plugin which samza repository is using is not compatible with the gradle versions greater than 3.0.

For doing the ligradle migration at linkedin, we need to update the nexus gradle plugin to the latest version.


shanthoosh:

PR #427, opened 2018-02-22
SAMZA-1594: Remove ScalaCompileOptions to make samza codebase build with gradle version > 3.0.

shanthoosh:
When samza repository is built with the gradle version greater than 3.0, we notice the following build failure.

No such property: useAnt for class: org.gradle.api.tasks.scala.ScalaCompileOptions  

This needs to be fixed to build samza with  gradle version >= 3.0.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -118,8 +118,6 @@ subprojects {
   tasks.withType(ScalaCompile) {
     // show compile errors in console output
     logging.setLevel LogLevel.WARN
-    scalaCompileOptions.fork = false
-    scalaCompileOptions.useAnt = true
   }
 
   tasks.withType(Test) {


shanthoosh:
@xinyuiscool Please take a look when you get a chance.

prateekm:
@shanthoosh That's because gradle version  > 3.0 uses zinc instead of ant to compile scala. Does removing useAnt and fork configs work?

shanthoosh:
@prateekm
Here's the suggested solution in the gradle forums for the problem. It is followed in the patch.

https://discuss.gradle.org/t/idea-integration-with-scala-plugin-broken-since-gradle-3-0-no-such-property-daemonserver/19159

Removing those two configurations did not help.

PR #430, opened 2018-02-23
SAMZA-1595: Fix scalacCompileOptions format to build with zinc scala compiler.

shanthoosh:
Zinc scala compiler(part of gradle version >= 3.0) expects the scala compilation arguments as a list(where each compilation argument is an element of the list).

In samza, the compilation arguments are concatenated into a single string and passed to the compiler.

This causes build failures when samza is built with Zinc scala compiler.

Existing ant scala compiler used to build samza in open source accepts the compilation arguments both as list and string.


shanthoosh:

PR #431, opened 2018-02-23
SAMZA-1596: Staging directory name has to be formatted in config

Zedmor:
When we instantiate a HDFS config staging directory we missing a formatter for getStagingDirectory so systems.hdfs-system-name.stagingDirectory does not parse from config, but only systems.%s.stagingDirectory only parses instead.
 
Solution is to add formatter to getStagingDirectory method.


vjagadish1989 (approved):
looks good to me, approved. Thanks much @Zedmor for this contribution.

@lhaiesp  Can you please take another look at this ?

PR #436, opened 2018-03-06
SD-1599: Improve the efficiency of the AsynRunLoop when some partitio…

jwlent55:
…ns are empty.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
+++ b/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
@@ -169,7 +169,6 @@ public class AsyncRunLoop implements Runnable, Throttleable {
         long activeNs = currentNs - blockNs;
         long totalNs = currentNs - prevNs;
         prevNs = currentNs;
-        containerMetrics.blockNs().update(currentNs - blockNs);
 
         if (totalNs != 0) {
           // totalNs is not 0 if timer metrics are enabled
--- a/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
+++ b/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
@@ -67,6 +67,7 @@ public class AsyncRunLoop implements Runnable, Throttleable {
   private final long windowMs;
   private final long commitMs;
   private final long callbackTimeoutMs;
+  private final long maxNoWorkWaitMs;
   private final SamzaContainerMetrics containerMetrics;
   private final ScheduledExecutorService workerTimer;
   private final ScheduledExecutorService callbackTimer;
@@ -75,6 +76,7 @@ public class AsyncRunLoop implements Runnable, Throttleable {
   private volatile Throwable throwable = null;
   private final HighResolutionClock clock;
   private final boolean isAsyncCommitEnabled;
+  private boolean runLoopResumedSinceLastChecked;
 
   public AsyncRunLoop(Map<TaskName, TaskInstance> taskInstances,
       ExecutorService threadPool,
@@ -84,6 +86,7 @@ public class AsyncRunLoop implements Runnable, Throttleable {
       long commitMs,
       long callbackTimeoutMs,
       long maxThrottlingDelayMs,
+      long maxNoWorkWaitMs,
       SamzaContainerMetrics containerMetrics,
       HighResolutionClock clock,
       boolean isAsyncCommitEnabled) {
@@ -95,6 +98,7 @@ public class AsyncRunLoop implements Runnable, Throttleable {
     this.commitMs = commitMs;
     this.maxConcurrency = maxConcurrency;
     this.callbackTimeoutMs = callbackTimeoutMs;
+    this.maxNoWorkWaitMs = maxNoWorkWaitMs;
     this.callbackTimer = (callbackTimeoutMs > 0) ? Executors.newSingleThreadScheduledExecutor() : null;
     this.callbackExecutor = new ThrottlingScheduler(maxThrottlingDelayMs);
     this.coordinatorRequests = new CoordinatorRequests(taskInstances.keySet());
@@ -153,6 +157,8 @@ public class AsyncRunLoop implements Runnable, Throttleable {
         long chooseNs = clock.nanoTime();
 
         containerMetrics.chooseNs().update(chooseNs - startNs);
+
+        blockIfNoWork(envelope);
 
         runTasks(envelope);
 
@@ -256,6 +262,31 @@ public class AsyncRunLoop implements Runnable, Throttleable {
   }
 
   /**
+   * When there are no new messages to process and the run loop has not been resumed since the last time this code was
+   * run, block the AsyncRunLoop thread for a short while. This will prevent this task from spinning when it has no work
+   * to do. If a task worker finishes or window/commit completes before the timeout it will resume the AsyncRunLoop
+   * thread immediately. This performs a similar function to what the poll timeout in the SystemConsumers.choose method
+   * provides for the single threaded RunLoop code, but, in the async case we need the ability to resume any time a task
+   * worker completes processing a message. In the Async case, that event will allow the task worker to start processing
+   * a message that may have already been chosen.
+   */
+  private void blockIfNoWork(IncomingMessageEnvelope envelope) {
+    synchronized (latch) {
+      if ((envelope != null) || runLoopResumedSinceLastChecked) {
+        runLoopResumedSinceLastChecked = false;
+        return;
+      }
+      try {
+        log.trace("Start no work wait");
+        latch.wait(maxNoWorkWaitMs);
+        log.trace("End no work wait");
+      } catch (InterruptedException e) {
+        throw new SamzaException("Run loop is interrupted", e);
+      }
+    }
+  }
+
+  /**
    * Resume the runloop thread. It is triggered once a task becomes ready again or has failure.
    */
   private void resume() {
@@ -265,6 +296,7 @@ public class AsyncRunLoop implements Runnable, Throttleable {
     }
     synchronized (latch) {
       latch.notifyAll();
+      runLoopResumedSinceLastChecked = true;
     }
   }
 
--- a/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
+++ b/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
@@ -76,7 +76,7 @@ public class AsyncRunLoop implements Runnable, Throttleable {
   private volatile Throwable throwable = null;
   private final HighResolutionClock clock;
   private final boolean isAsyncCommitEnabled;
-  private boolean runLoopResumedSinceLastChecked;
+  private volatile boolean runLoopResumedSinceLastChecked;
 
   public AsyncRunLoop(Map<TaskName, TaskInstance> taskInstances,
       ExecutorService threadPool,
--- a/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
+++ b/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
@@ -154,20 +154,19 @@ public class AsyncRunLoop implements Runnable, Throttleable {
         long startNs = clock.nanoTime();
 
         IncomingMessageEnvelope envelope = chooseEnvelope();
-        long chooseNs = clock.nanoTime();
 
+        long chooseNs = clock.nanoTime();
         containerMetrics.chooseNs().update(chooseNs - startNs);
 
-        blockIfNoWork(envelope);
+        blockIfBusyOrNoNewWork(envelope);
+
+        long blockNs = clock.nanoTime();
+        containerMetrics.blockNs().update(blockNs - chooseNs);
 
         runTasks(envelope);
 
-        long blockNs = clock.nanoTime();
-
-        blockIfBusy(envelope);
-
         long currentNs = clock.nanoTime();
-        long activeNs = blockNs - chooseNs;
+        long activeNs = currentNs - blockNs;
         long totalNs = currentNs - prevNs;
         prevNs = currentNs;
         containerMetrics.blockNs().update(currentNs - blockNs);
@@ -239,14 +238,35 @@ public class AsyncRunLoop implements Runnable, Throttleable {
   /**
    * Block the runloop thread if all tasks are busy. When a task worker finishes or window/commit completes,
    * it will resume the runloop.
+   *
+   * In addition, delay the AsyncRunLoop thread for a short time if there are no new messages to process and the run loop
+   * has not been resumed since the last time this code was run. This will prevent the main thread from spinning when it
+   * has no work to distribute. If a task worker finishes or window/commit completes before the timeout then resume
+   * the AsyncRunLoop thread immediately. That event may allow a task worker to start processing a message that has already
+   * been chosen.  In any event it should only delay for a short time.  It needs to periodically check for new messages.
    */
-  private void blockIfBusy(IncomingMessageEnvelope envelope) {
+  private void blockIfBusyOrNoNewWork(IncomingMessageEnvelope envelope) {
     synchronized (latch) {
+
+      // First check to see if we should delay the run loop for a short time.  The runLoopResumedSinceLastChecked boolean
+      // is used to ensure we don't delay if there may already be a task ready to dequeue a previously chosen/pending
+      // message. It is better to occasionally make one additional loop when there is no work to do then delay the
+      // runloop when there is work that could be started immediately.
+      if ((envelope == null) && !runLoopResumedSinceLastChecked) {
+        try {
+          log.trace("Start no work wait");
+          latch.wait(maxIdleMs);
+          log.trace("End no work wait");
+        } catch (InterruptedException e) {
+          throw new SamzaException("Run loop is interrupted", e);
+        }
+      }
+      runLoopResumedSinceLastChecked = false;
+
+      // Next check to see if we should block if all the tasks are busy.
       while (!shutdownNow && throwable == null) {
         for (AsyncTaskWorker worker : taskWorkers) {
           if (worker.state.isReady()) {
-            // should continue running if any worker state is ready
-            // consumerMultiplexer will block on polling for empty partitions so it won't cause busy loop
             return;
           }
         }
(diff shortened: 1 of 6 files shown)


xinyuiscool:
@jwlent55 : thanks a lot for this patch. I need some time to think through about this chance. Please bear with me for a few days. Thanks.

jwlent55:
I appreciate your interest in this issue and realize it may take time to evaluate.  There may well be other and better ways to address this issue.  I took a minimalist approach which I hoped would reduce the chance of actually making things worse.  

I briefly considered a more aggressive approach. My thought was to make each thread have its own main loop (and get rid of ThreadJob) and permanently assign each thread one or more partitions to work with.  Basically simulate the individual JVM approach with threads.  Each thread would round robin among the assigned partitions, but, there would be no notion of fairness between threads.  The JVM would do the load sharing between threads much as the OS controlled load sharing between multiple JVMs (each with there own set of partitions) for the single threaded approach.

I also considered relaxing the notion of fairness between threads by keeping the Chooser full (calling tryUpdate as soon as the previous message was inserted).

Both those ideas sounded more like features than bug fixes and would require extensive design, review, and testing. I therefore did not pursue them, but, it would not surprise me if you suggested something equally different from what I have coded up.

samza-core/src/main/scala/org/apache/samza/config/TaskConfig.scala
```diff
@@ -44,11 +44,13 @@ object TaskConfig {
   val MAX_CONCURRENCY = "task.max.concurrency" // max number of concurrent process for a AsyncStreamTask
   val CALLBACK_TIMEOUT_MS = "task.callback.timeout.ms"  // timeout period for triggering a callback
   val ASYNC_COMMIT = "task.async.commit" // to enable async commit in a AsyncStreamTask
+  val MAX_NO_WORK_WAIT_MS = "task.max.no.work.wait.ms"  // maximum time to wait for a task worker to complete when there are no new messages to handle
```
nickpan47: nit: can we change it to task.max.idle.ms?

xinyuiscool:
Sorry that it took so long. Thanks a lot for this patch. As you already analyzed very thoroughly, this is a flaw in current AsyncRunLoop that relies on the poll timeout to avoid busy loop, which caused long delays for tasks which have already completed processing. The way of your fix makes sense to me: AsyncRunLoop should use the latch to wait instead of relying on poll timeout.This will allow the other tasks to proceed. Nice work!

For the fix implementation, I notice two things:

1) Did you also configure task.poll.interval.ms to be 0 in order to get rid of the poll waiting? Seems to me it's a must since otherwise the run loop will still poll quite often in case of empty partitions. To this point, looks like we don't need the poll timeout (task.poll.inverval.ms) anymore after this change, since runloop will control the poll interval, and we should immediately return after systemconsumers.poll() in any case. If this is the case, would you mind also make SystemConsumers.poll() always return without waiting?

2) seems the logic of blockIfNoWork() can be moved into blockIfBusy(), basically doing the same check of null envelope and wait if no worker is available. Is that something feasible?

Thanks!

jwlent55:
Certainly.  I waffle over naming variables all the time and am happy to
take other people's suggestions.

Somewhat related to a question from @xinyuiscool we could question why I
even made this timeout configurable.  It did not expect anyone to set it to
0 (although I did try that just to see the main loop peg one of the CPUs on
my dev box), but, I suppose I could image a case where you had a box with a
lot of CPUs and were willing to have one spin faster to minimize latency.
The real reason I made in configurable was so it could play with it during
testing without rebuilding my application.

jwlent55:
Regarding your questions:

1) I agree the timeout should never be set to 0.  Not sure it should even
be configurable.  I did so really for testing purposes (so I could modify
it without rebuilding my application).  I suppose someone might want to
tweak it to trade CPU for latency, but, I would be OK with not even making
it configurable.  The 10 ms value I based it on was not.

2) I do like the idea of having only one check.  Originally I thought that
the blockIfNoWork needed to come before the runTasks call, but, that was
most likely before I found some race conditions and decided to add the
(admittedly somewhat ugly) runLoopResumedSinceLastChecked boolean.  Now I
think it could be after and therefore the two checks could be combined.
That will require a bit of testing (and thought).  One other option I
originally considered, but, dismissed as a bit risky was to remove the
blockIfBusy check.  The idea being that eventually if all the task were
busy the Chooser would stop finding work and the new check would force a
pause.  I could not convince myself I understood the code well enough to
guarantee that in all cases.

jwlent55:
Regarding your questions: 1) I agree the timeout should never be set to 0. Not sure it should even be configurable. I did so really for testing purposes (so I could modify it without rebuilding my application). I suppose someone might want to tweak it to trade CPU for latency, but, I would be OK with not even making it configurable. The 10 ms value I based it on was not. 2) I do like the idea of having only one check. Originally I thought that the blockIfNoWork needed to come before the runTasks call, but, that was most likely before I found some race conditions and decided to add the (admittedly somewhat ugly) runLoopResumedSinceLastChecked boolean. Now I think it could be after and therefore the two checks could be combined. That will require a bit of testing (and thought). One other option I originally considered, but, dismissed as a bit risky was to remove the blockIfBusy check. The idea being that eventually if all the task were busy the Chooser would stop finding work and the new check would force a pause. I could not convince myself I understood the code well enough to guarantee that in all cases.

samza-core/src/main/scala/org/apache/samza/config/TaskConfig.scala
jwlent55: Certainly.  I waffle over naming variables all the time and am happy to
take other people's suggestions.

Somewhat related to a question from @xinyuiscool we could question why I
even made this timeout configurable.  It did not expect anyone to set it to
0 (although I did try that just to see the main loop peg one of the CPUs on
my dev box), but, I suppose I could image a case where you had a box with a
lot of CPUs and were willing to have one spin faster to minimize latency.
The real reason I made in configurable was so it could play with it during
testing without rebuilding my application.

xinyuiscool:
@jwlent55 : I think you might misunderstood my first question. Now we have two waits here: one for the poll (task.poll.interval.ms), and the other for no work (task.max.no.work.wait.ms). My question is that since we can now use the second wait to control the polling (if there is no work, don't poll until task.max.no.work.wait.ms), seems we don't need the first wait anymore (equal to set task.poll.interval.ms=0). The AsyncRunLoop was relying on the first poll timer to avoid busy looping, but now it should be fine to return immediately after polling any SystemConsumer. This seems to be the goal we want to achieve here: AsyncRunLoop will control the polling by knowing whether there is worker available to process the message. And this should be much more efficient as you experiments demonstrated. What do you think?

jwlent55:
You are correct.  I completely misread your first question.  Without looking at the code I mistook it for a question I already had in my mind and went off into the weeds.  Let me try again.

I did not set the task.poll.interval.ms to 0 (which is used to set "pollIntervalMs").  I left it at its default 50 ms.  I think it is still necessary to ensure that if some partitions have exhausted the their "unprocessedMessagesBySSP", but, others are keeping up with main loop Chooser (so that it always finds new work to process) that the main loop will still occasionally (every 50 ms) check to see if the currently empty partitions have new messages to "drain" from the bufferedMessages queue.  This call should not block if the "timeout" variable is set to zero (when it is 0 this "poll" is really just a "drainTo").  My proposed update includes logic to ensure that timeout value is always set to 0 if we are not planning to update the Chooser during the call to choose (i.e., in the AsyncRunLoop case).  I think this same logic stops the blocking/polling both in the "no work" case and in the "long time (50 ms) since last poll" case.

On a related note when I first started working on this issue I thought the problem might be in the 
BlockingEnvelopeMap.poll method.  I noted that it would block on the first SSP that did not have any messages (if the timeout was greater than zero) even if other empty partitions had messages ready to "drain".  I reorganized the logic so it first "drained" all the queues and only then if it could not find at least one message to process start "polling" the SSPs one by one (and most likely blocking on the first one). That did not address my issue so I did not include that logic in my change.

jwlent55:
While investigating your question I noted that I had not included documentation for my new parameter in the configuration-table.html file.  I assume I should do so.  Let me know if you want this new parameter left out.

jwlent55:
I have gone ahead and made a first pass at addressing your code inspection comments.  Happy to make more changes.

jwlent55:
Not sure about this failure.  Looks like a networking issue.  The testClearStream passes for me locally.

```
$ ./gradlew clean check
...
:samza-kafka_2.11:testClasses
:samza-kafka_2.11:test
16301 [SessionTracker] INFO org.apache.zookeeper.server.SessionTrackerImpl - SessionTrackerImpl exited loop!
16335 [kafka-producer-network-thread | producer-1] WARN org.apache.kafka.clients.NetworkClient - Connection to node 0 could not be established. Broker may not be available.
16335 [kafka-producer-network-thread | producer-1] WARN org.apache.kafka.clients.NetworkClient - Connection to node 1 could not be established. Broker may not be available.
16335 [kafka-producer-network-thread | producer-1] WARN org.apache.kafka.clients.NetworkClient - Connection to node 2 could not be established. Broker may not be available.
24447 [kafka-producer-network-thread | samza_producer-some_job_name-i001] WARN org.apache.kafka.clients.NetworkClient - Connection to node 1 could not be established. Broker may not be available.
24497 [kafka-producer-network-thread | samza_producer-some_job_name-i001] WARN org.apache.kafka.clients.NetworkClient - Connection to node 2 could not be established. Broker may not be available.
24497 [kafka-producer-network-thread | samza_producer-some_job_name-i001] WARN org.apache.kafka.clients.NetworkClient - Connection to node 1 could not be established. Broker may not be available.
24499 [kafka-producer-network-thread | samza_producer-some_job_name-i001] WARN org.apache.kafka.clients.NetworkClient - Connection to node 1 could not be established. Broker may not be available.
24500 [kafka-producer-network-thread | samza_producer-some_job_name-i001] WARN org.apache.kafka.clients.NetworkClient - Connection to node 2 could not be established. Broker may not be available.
:samza-kafka_2.11:check
...
BUILD SUCCESSFUL

Total time: 6 mins 21.453 secs
```
Is there a way for me to retry the test?

jwlent55:
As for the suggestion:

  "would you mind also make SystemConsumers.poll() always return without waiting"

I think I have made it never block when it is called (indirectly) from AsyncRunLoop (by ensuring timeout = 0 when it calls ), but, I think it still needs the ability to block when it is called (indirectly) from RunLoop (where timeout may be equal to 10).  Maybe I misunderstand this comment.

Now that I take a second look at the timeout setting code I think I need to make setting it to 0 more explicit.  I will make another update.  On the plus side this change should trigger another build.  Maybe that will clear the error.

I am a bit busy for the next couple of days and may no have time much time to spend on this issue.  Also I have not yet had time to work on combining the blockIfNoWork() and blockIfBusy() code.

xinyuiscool:
@jwlent55 : your update makes a lot of sense to me (and clears the error too :) )! Runloop is actually deprecated (I didn't get a chance to finally remove that class, we have a ticket to track it), so this poll will only be made from AsyncRunLoop. Seems to me we don't need the poll timer anymore after your change. If that's out of scope for this patch, I can put it under the ticket of removing RunLoop so it will be cleaned up.

No hurry on the combining the block() methods. I believe it will work pretty well afterwards (need your help to test with your test jobs). Thanks a lot for being very thorough on this issue and make our runloop much more efficient.

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -255,6 +261,31 @@ private void blockIfBusy(IncomingMessageEnvelope envelope) {
...
+   * run, block the AsyncRunLoop thread for a short while. This will prevent this task from spinning when it has no work
+   * to do. If a task worker finishes or window/commit completes before the timeout it will resume the AsyncRunLoop
+   * thread immediately. This performs a similar function to what the poll timeout in the SystemConsumers.choose method
+   * provides for the single threaded RunLoop code, but, in the async case we need the ability to resume any time a task
+   * worker completes processing a message. In the Async case, that event will allow the task worker to start processing
+   * a message that may have already been chosen.
+   */
+  private void blockIfNoWork(IncomingMessageEnvelope envelope) {
+    synchronized (latch) {
+      if ((envelope != null) || runLoopResumedSinceLastChecked) {
```
nickpan47: I think that @xinyuiscool mentioned it. I also just realized that we can merge the two block functions into one, if we do not wait on the consumer.poll:
1) make sure the exit/return condition in the synchronized block is: (there is a pending messages in the queue or there is an new envelope returned from poll) && (the corresponding task worker is available: i.e. not busy);
2) otherwise, wait on latch with task.max.idle.ms. 

Does that sound good?

jwlent55: I am also not sure how to implement the exact exit/return logic you suggest above.  It sounds like that would (at least) require checking all the pendingEnvelopeQueues for messages and matching them to the idle workers.  It seems like I might also need to check for pending windows and/or commits.  It seems easier/safer to just call run() on all the task workers if at least one task worker is idle (the current logic).

I am also not sure what I should do if the "combined trigger" wait you suggest times out?  The current blockIfBusy can continue to wait for an unlimited amount of time and can immediately wait again if after the notify all the tasks are still busy.  Are you suggesting changing this logic to block at most once for a fixed amount of time (the blockIfNoWork condition should block at most once)?

What I can do is check the two existing conditions used in blockIfNoWork and blockIfBusy serially inside a common latch (most likely called between the choose and run steps).  I might be able to combine them into a single "if" check it the blockIfBusy logic is relaxed to block at most once and for a fixed amount of time.  That might cause the run loop to waste cycles in a very busy application.

Comments before I proceed?

jwlent55: I did not make "runLoopResumedSinceLastChecked" volatile because I only set and check that variable inside code blocks synchronized on "latch", but, it could not hurt to make it volatile (and if my assumption is wrong it is required).  My assumption was that the synchronized/wait/notify semantics would ensure each thread saw the latest value.  Change is on the way.

xinyuiscool: @jwlent55 : it will be super nice if you can help move your logic in blockInNoWork into blockIfBusy which do the things you mentioned: check the two existing conditions used in blockIfNoWork and blockIfBusy serially inside a common latch. That will make the code easier to follow and reason since the latch happens in one place. The other part looks good to me. Thanks.

jwlent55: Done.

jwlent55:
My apologies for the bogus call to containerMetrics that I accidentally left in the code when I combined the block and delay code and moved them both to before run.  The reason I chose that location was because in the delay case it makes sense to call run next.  Choose will most likely not return anything again until run has been called (unless the 50 ms poll interval has been reached).  I think the code will work in either location so if you prefer the previous order better (choose/run/block-delay) let me know.

xinyuiscool:
@jwlent55 : The changes look great! Thanks a lot for this fix. 

Btw, did you get a chance to run a few cases in your benchmark after the code refactoring? It will be great to confirm the change didn't break anything. Thanks.

jwlent55:
While working on these changes I developed a new test case:

- 6 Partitions
- 200/400/600/800/1000/2000 messages per partition
- Test Samza job logs every 100 messages per partition.  Does nothing else.
- Always starts at the beginning of the topic.
- Calculated run time as difference time between the last "100th message processed" log and the "Entering run loop." log.

I tested:

- master code as it was earlier this week.
- Version of code with separate delay/block methods.
- Latest version with one delay/block method.

I tested 4 variations of the parameters and ran each scenario 3 times.  Again these are on my dev box so the background load is not well controlled.  There will be some variations run to run.

Here are the results (which look good to me):

  |   |   | Got Thread | Got Single | Process | Run Loop |   |   | Current Code |   |   | First Fix: a7c0ac4 |   |   | Current Fix: ef0e53b |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
pool.size | thread.mode |   | Pool Size | Thread Mode | Thread | Mode |   | Run 1 | Run 2 | Run 3 | Run 1 | Run 2 | Run 3 | Run 1 | Run 2 | Run 3
not specified | true |   | 0 | true | ThreadJob | single thread |   | 126 | 126 | 113 | 103 | 98 | 95 | 110 | 130 | 102
not specified | not specified |   | 0 | false | ThreadJob | asynchronous |   | 156 | 141 | 149 | 116 | 137 | 135 | 111 | 135 | 105
1 | not specified |   | 1 | false | Container Thread | asynchronous |   | 17,560 | 17,642 | 18,272 | 179 | 165 | 211 | 182 | 180 | 216
6 | not specified |   | 6 | false | Container Thread | asynchronous |   | 17,097 | 16,921 | 17,526 | 229 | 210 | 162 | 215 | 184 | 167

jwlent55:
Just FYI, this is how long it takes to get from "Entering run loop." log to the last "200th message processed" log (all 6 partitions have at least 200 messages).  Up to tis point the current code generally runs about as fast as the new code.  It supports the conclusion that the problem only occurs once the partitions have started to exhaust.

  |   |   |   |   |   |   |   |   |   |   |   | Time (ms) |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
  |   |   | Got Thread | Got Single | Process | Run Loop |   |   | Current Code |   |   | First Fix: a7c0ac4 |   |   | Current Fix: ef0e53b |  
pool.size | thread.mode |   | Pool Size | Thread Mode | Thread | Mode |   | Run 1 | Run 2 | Run 3 | Run 1 | Run 2 | Run 3 | Run 1 | Run 2 | Run 3
not specified | true |   | 0 | true | ThreadJob | single thread |   | 106 | 102 | 85 | 98 | 68 | 64 | 71 | 95 | 68
not specified | not specified |   | 0 | false | ThreadJob | asynchronous |   | 109 | 80 | 107 | 91 | 114 | 103 | 70 | 103 | 68
1 | not specified |   | 1 | false | Container Thread | asynchronous |   | 164 | 282 | 98 | 112 | 107 | 110 | 94 | 103 | 123
6 | not specified |   | 6 | false | Container Thread | asynchronous |   | 181 | 115 | 152 | 125 | 70 | 94 | 145 | 115 | 100

nickpan47:
@jwlent55 awesome results! Thanks a lot!

jwlent55:
Happy to try to help.  Hope the code works well for you.  A 0.12 version of the code will be going to production here pretty soon.  I will alert you if we have any issues with it.

xinyuiscool:
Sounds great! We are looking forward to seeing performance enhancements with this patch too for our prod jobs. Nice work!

PR #438, opened 2018-03-07
SAMZA-1589: Reduce failure retry duration in KafkaCheckpointManager.writeCheckpoint

shanthoosh:


shanthoosh:
@prateekm 
Please take a look.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -53,8 +53,8 @@ class KafkaCheckpointManager(checkpointSpec: KafkaStreamSpec,
                              checkpointMsgSerde: Serde[Checkpoint] = new CheckpointSerde,
                              checkpointKeySerde: Serde[KafkaCheckpointLogKey] = new KafkaCheckpointLogKeySerde) extends CheckpointManager with Logging {
 
-  // Retry duration is approximately 83 minutes.
-  var MaxRetriesOnFailure = 50
+  // Maximum failure wait time when writing checkpoint is 15 minutes.
+  var MaxFailureWaitTimeInMs = 900000
```
vjagadish1989: 1. Should name this as MaxRetryDurationMs to reflect what the property is. Also update comment
2. Use `TimeUnit.MINUTES.toMillis(15)` to make it more obvious

vjagadish1989:
Shanthoosh, I don’t see the suggested changes yet. Did you forget to commit? I was proposing to rename the variable.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
vjagadish1989: @shanthoosh 

Not sure why Github missed your earlier comment. 

>> I think the existing name makes sense. By this change we’re essentially introducing a maximum timeOut upon failure for the writeCheckPoint api. IMHO, this constant value is essentially a retry timeout and the naming is fine. maxRetryDuration name doesn’t indicate that.

Of course, I disagree.

- I would argue that the "retryDuration" is more semantically correct: We will not issue another retry-request if the duration elapses. It imposes no requirements on when the call to "writeCheckpoint" is expected to return during a failure.

- The timeout upon failure is additive on the per-request timeout. If the per-request timeout takes more than the retry timeout, then what's the whole point of calling it a "failure-timeout" for the "writeCheckpoint" API?

I would recommend that we have a clear definition of what parameters mean when specifying APIs and stick to the earlier proposal.

vjagadish1989: - Had a discussion offline and agreed to switch to the earlier proposal to keep the API clear. This is not a "maximum timeOut upon failure for the writeCheckPoint api". 
- Let's not do force commits accidentally so that we loose history.

vjagadish1989 (approved):
Looks good, thanks for the contributions.

PR #439, opened 2018-03-07
SAMZA-1602 : Moving class StreamAssert from src/test to src/main

Sanil15:
Tested with ./gradlew clean build, which is passing

### DIFF
--- a/samza-test/src/test/java/org/apache/samza/test/operator/BroadcastAssertApp.java
+++ b/samza-test/src/test/java/org/apache/samza/test/operator/BroadcastAssertApp.java
@@ -25,7 +25,7 @@ import org.apache.samza.operators.MessageStream;
 import org.apache.samza.operators.StreamGraph;
 import org.apache.samza.serializers.JsonSerdeV2;
 import org.apache.samza.test.operator.data.PageView;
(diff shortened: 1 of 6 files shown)


samza-test/src/test/java/org/apache/samza/test/operator/BroadcastAssertApp.java
```diff
@@ -25,7 +25,7 @@
 import org.apache.samza.operators.StreamGraph;
 import org.apache.samza.serializers.JsonSerdeV2;
 import org.apache.samza.test.operator.data.PageView;
-import org.apache.samza.test.util.StreamAssert;
+import com.apache.samza.test.framework.StreamAssert;
```
prateekm: Why is this com.apache.samza?

Also, can you explain why this change is necessary? Is this expected to be used somewhere else?

Sanil15: Updated com.apache.samza to org.apache.samza, my bad with the naming confusion.

Necessity for this change:

The StreamAsset class will be part of bigger framework that helps in Continuous integration testing that can be used by Open Source users as well as Samza-Li. Hence we move this out of test folder

PR #441, opened 2018-03-08
SAMZA-1584: Improve logging in StreamProcessor. 

shanthoosh:
Add the processorID in the log lines wherever necessary(since we support running multiple stream applications in a JVM) and improving logging in general in StreamProcessor.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -56,7 +56,7 @@ import org.slf4j.LoggerFactory;
 @InterfaceStability.Evolving
 public class StreamProcessor {
   private static final Logger LOGGER = LoggerFactory.getLogger(StreamProcessor.class);
-  private static final String CONTAINER_THREAD_NAME_FORMAT = "Samza StreamProcessor-%s-container-thread";
+  private static final String CONTAINER_THREAD_NAME_FORMAT = "Samza StreamProcessor Container Thread-%d";
 
   private final JobCoordinator jobCoordinator;
   private final StreamProcessorLifecycleListener processorListener;
@@ -186,7 +186,7 @@ public class StreamProcessor {
         LOGGER.info("Waiting {} milliseconds for the container: {} to shutdown.", taskShutdownMs, container);
         containerShutdownInvoked = true;
       } catch (Exception exception) {
-        LOGGER.error(String.format("Shutdown of container: %s threw an exception.", container), exception);
+        LOGGER.error(String.format("Ignoring the exception during the shutdown of container: %s.", container), exception);
       }
     }
 
@@ -218,21 +218,21 @@ public class StreamProcessor {
               LOGGER.info("Job model expired. Shutting down the container: {} of stream processor: {}.", container, processorId);
               container.pause();
               shutdownComplete = jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
-              LOGGER.info("Container shutdown completion status is: {} for stream processor: {}.", shutdownComplete, processorId);
+              LOGGER.info("Shutdown status of container: {} for stream processor: {} is: {}.", container, processorId, shutdownComplete);
             } catch (IllegalContainerStateException icse) {
               // Ignored since container is not running
-              LOGGER.info(String.format("Container: %s of stream processor: %s is not running.", container, processorId), icse);
+              LOGGER.info(String.format("Cannot shutdown container: %s for stream processor: %s. Container is not running.", container, processorId), icse);
               shutdownComplete = true;
             } catch (InterruptedException e) {
               Thread.currentThread().interrupt();
-              LOGGER.warn(String.format("Shutdown of the container: %s was interrupted", container), e);
+              LOGGER.warn(String.format("Shutdown of container: %s for stream processor: %s was interrupted", container, processorId), e);
             }
             if (!shutdownComplete) {
               LOGGER.warn("Container: {} shutdown was unsuccessful. Stopping the stream processor: {}.", container, processorId);
               container = null;
               stop();
             } else {
-              LOGGER.info("Container shutdown completed for stream processor: {}.", processorId);
+              LOGGER.info("Container: {} shutdown completed for stream processor: {}.", container, processorId);
             }
           } else {
             LOGGER.info("Container: {} of the stream processor: {} is not running.", container, processorId);
@@ -257,7 +257,7 @@ public class StreamProcessor {
                 processorListener.onStart();
               }
             } else {
-              LOGGER.debug("Received duplicate container start notification for container: {} in stream processor: {}.", container, processorId);
+              LOGGER.warn("Received duplicate container start notification for container: {} in stream processor: {}.", container, processorId);
             }
           }
 
@@ -283,7 +283,7 @@ public class StreamProcessor {
               LOGGER.warn("JobCoordinatorLatch was null. It is possible for some component to be waiting.");
             }
             containerException = t;
-            LOGGER.error(String.format("Container: %s failed. Stopping the stream processor: %s.", container, processorId), containerException);
+            LOGGER.error(String.format("Container: %s failed with an exception. Stopping the stream processor: %s. Original exception:", container, processorId), containerException);
             container = null;
             stop();
           }
@@ -292,7 +292,7 @@ public class StreamProcessor {
         container = createSamzaContainer(processorId, jobModel);
         container.setContainerListener(containerListener);
         LOGGER.info("Starting the container: {} for the stream processor: {}.", container, processorId);
-        ThreadFactory threadFactory = new ThreadFactoryBuilder().setNameFormat(String.format(CONTAINER_THREAD_NAME_FORMAT, processorId)).build();
+        ThreadFactory threadFactory = new ThreadFactoryBuilder().setNameFormat(CONTAINER_THREAD_NAME_FORMAT).build();
         executorService = Executors.newSingleThreadExecutor(threadFactory);
         executorService.submit(container::run);
       }
@@ -313,7 +313,7 @@ public class StreamProcessor {
 
       @Override
       public void onCoordinatorFailure(Throwable throwable) {
-        LOGGER.info(String.format("Coordinator failed with an exception. Stopping the stream processor: %s.", processorId), throwable);
+        LOGGER.info(String.format("Coordinator: %s failed with an exception. Stopping the stream processor: %s. Original exception:", coordinator, processorId), throwable);
         stop();
         if (processorListener != null) {
           processorListener.onFailure(throwable);
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -24,7 +24,6 @@ import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import org.apache.samza.SamzaContainerStatus;
@@ -72,7 +71,6 @@ public class StreamProcessor {
   @VisibleForTesting
   volatile SamzaContainer container = null;
   private volatile Throwable containerException = null;
-  private volatile Future<?> scheduledSamzaContainerFuture = null;
 
   // Latch used to synchronize between the JobCoordinator thread and the container thread, when the container is
   // stopped due to re-balancing
@@ -187,14 +185,6 @@ public class StreamProcessor {
         LOGGER.info("Shutting down the container: {} of stream processor: {}.", container, processorId);
         container.shutdown();
         LOGGER.info("Waiting {} milliseconds for the container: {} to shutdown.", taskShutdownMs, container);
-        if (executorService != null) {
-          executorService.shutdown();
-          if (!executorService.awaitTermination(taskShutdownMs, TimeUnit.MILLISECONDS)) {
-            LOGGER.info("Forcefully shutting down the executor service of the processor: {}.", processorId);
-            executorService.shutdownNow();
-            executorService.awaitTermination(taskShutdownMs, TimeUnit.MILLISECONDS);
-          }
-        }
         containerShutdownInvoked = true;
       } catch (Exception exception) {
         LOGGER.error(String.format("Shutdown of container: %s threw an exception.", container), exception);
@@ -230,11 +220,6 @@ public class StreamProcessor {
               container.pause();
               shutdownComplete = jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
               LOGGER.info("Container shutdown completion status is: {} for stream processor: {}.", shutdownComplete, processorId);
-              if (scheduledSamzaContainerFuture != null && !scheduledSamzaContainerFuture.isDone()) {
-                LOGGER.info(String.format("Container: %s did not shutdown after %s milliseconds. Interrupting the container thread of stream processor: %s.", container, taskShutdownMs, processorId));
-                scheduledSamzaContainerFuture.cancel(true);
-                shutdownComplete = jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
-              }
             } catch (IllegalContainerStateException icse) {
               // Ignored since container is not running
               LOGGER.info(String.format("Container: %s of stream processor: %s is not running.", container, processorId), icse);
@@ -310,7 +295,7 @@ public class StreamProcessor {
         LOGGER.info("Starting the container: {} for the stream processor: {}.", container, processorId);
         ThreadFactory threadFactory = new ThreadFactoryBuilder().setNameFormat(String.format(CONTAINER_THREAD_NAME_FORMAT, processorId)).setDaemon(true).build();
         executorService = Executors.newSingleThreadExecutor(threadFactory);
-        scheduledSamzaContainerFuture = executorService.submit(container::run);
+        executorService.submit(container::run);
       }
 
       @Override
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -218,7 +218,7 @@ public class StreamProcessor {
               LOGGER.info("Job model expired. Shutting down the container: {} of stream processor: {}.", container, processorId);
               container.pause();
               shutdownComplete = jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
-              LOGGER.info("Shutdown status of container: {} for stream processor: {} is: {}.", container, processorId, shutdownComplete);
+              LOGGER.info(String.format("Shutdown status of container: %s for stream processor: %s is: %s.", container, processorId, shutdownComplete));
             } catch (IllegalContainerStateException icse) {
               // Ignored since container is not running
               LOGGER.info(String.format("Cannot shutdown container: %s for stream processor: %s. Container is not running.", container, processorId), icse);
@@ -313,7 +313,7 @@ public class StreamProcessor {
 
       @Override
       public void onCoordinatorFailure(Throwable throwable) {
-        LOGGER.info(String.format("Coordinator: %s failed with an exception. Stopping the stream processor: %s. Original exception:", coordinator, processorId), throwable);
+        LOGGER.info(String.format("Coordinator: %s failed with an exception. Stopping the stream processor: %s. Original exception:", jobCoordinator, processorId), throwable);
         stop();
         if (processorListener != null) {
           processorListener.onFailure(throwable);
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -293,7 +293,7 @@ public class StreamProcessor {
         container = createSamzaContainer(processorId, jobModel);
         container.setContainerListener(containerListener);
         LOGGER.info("Starting the container: {} for the stream processor: {}.", container, processorId);
-        ThreadFactory threadFactory = new ThreadFactoryBuilder().setNameFormat(String.format(CONTAINER_THREAD_NAME_FORMAT, processorId)).setDaemon(true).build();
+        ThreadFactory threadFactory = new ThreadFactoryBuilder().setNameFormat(String.format(CONTAINER_THREAD_NAME_FORMAT, processorId)).build();
         executorService = Executors.newSingleThreadExecutor(threadFactory);
         executorService.submit(container::run);
       }
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -68,8 +68,7 @@ public class StreamProcessor {
 
   private ExecutorService executorService;
 
-  @VisibleForTesting
-  volatile SamzaContainer container = null;
+  private volatile SamzaContainer container = null;
   private volatile Throwable containerException = null;
 
   // Latch used to synchronize between the JobCoordinator thread and the container thread, when the container is
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -24,6 +24,8 @@ import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import org.apache.samza.SamzaContainerStatus;
 import org.apache.samza.annotation.InterfaceStability;
@@ -55,6 +57,7 @@ import org.slf4j.LoggerFactory;
 @InterfaceStability.Evolving
 public class StreamProcessor {
   private static final Logger LOGGER = LoggerFactory.getLogger(StreamProcessor.class);
+  private static final String CONTAINER_THREAD_NAME_FORMAT = "Samza StreamProcessor-%s-container-thread";
 
   private final JobCoordinator jobCoordinator;
   private final StreamProcessorLifecycleListener processorListener;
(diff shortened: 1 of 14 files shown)


shanthoosh:
@xinyuiscool @sborya 
Please take a look.

shanthoosh:
@xinyuiscool 
If you agree, we can have a follow up PR to this, where we will have  bounded wait time in LocalApplicationRunner.waitForFinish API as well.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -137,19 +138,22 @@ public void start() {
...
-    }
-    //Setting the isLeader metric to false when the stream processor shuts down because it does not remain the leader anymore
-    metrics.isLeader.set(false);
-    debounceTimer.stopScheduler();
-    zkController.stop();
-
-    shutdownMetrics();
-    if (coordinatorListener != null) {
-      coordinatorListener.onCoordinatorStop();
+    if (!isStopped) {
```
sborya: Please add log line. Having the thread name here will help debug issues.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -213,32 +226,33 @@ public void onJobModelExpired() {
...
-            LOGGER.info("Shutting down container done for pid=" + processorId + "; complete =" + shutdownComplete);
             if (!shutdownComplete) {
-              LOGGER.warn("Container " + container.toString() + " may not have shutdown successfully. " +
-                  "Stopping the processor.");
-              container = null;
-              stop();
+              LOGGER.warn("Container: {} shutdown was unsuccessful. Stopping the stream processor: {}.", container, processorId);
+              if (!shutdownComplete && scheduledSamzaContainerFuture != null && !scheduledSamzaContainerFuture.isDone()) {
+                LOGGER.info(String.format("Container: %s did not shutdown after %s milliseconds. Interrupting the container thread of stream processor: %s.", container, taskShutdownMs, processorId));
+                scheduledSamzaContainerFuture.cancel(true);
```
sborya: Should wait for results (if any) of the interrupt.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -176,30 +181,38 @@ public void start() {
...
+          if (!executorService.isShutdown()) {
+            LOGGER.info("Forcefully shutting down the executor service of the container: {}.", container);
+            executorService.shutdownNow();
+          }
+        }
+      } catch (Exception exception) {
+        LOGGER.info(String.format("Shutdown of container: %s threw an exception.", container), exception);
       }
     }
+    stopJobCoordinator();
```
sborya: this logic of shutting down container and jobCoordinator in different order becoming very complicated. Is there a way to simplify it?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -55,6 +57,7 @@
 @InterfaceStability.Evolving
 public class StreamProcessor {
   private static final Logger LOGGER = LoggerFactory.getLogger(StreamProcessor.class);
+  private static final String CONTAINER_THREAD_NAME_FORMAT = "processorId-%s-container-thread";
```
prateekm: Please use "Samza StreamProcessor % Thread-%d" for consistency with rest of the thread names. See https://github.com/apache/samza/commit/1dfc5cecdfa358851f01b4ef105ccb0f62e9891c

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -176,30 +181,38 @@ public void start() {
...
-        containerShutdownInvoked = true;
-      } catch (IllegalContainerStateException icse) {
-        LOGGER.info("Container was not running", icse);
+        LOGGER.info("Waiting {} milliseconds for the container: {} to shutdown.", taskShutdownMs, container);
+        jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
+        if (executorService != null) {
+          executorService.shutdown();
+          executorService.awaitTermination(taskShutdownMs, TimeUnit.MILLISECONDS);
+          if (!executorService.isShutdown()) {
+            LOGGER.info("Forcefully shutting down the executor service of the container: {}.", container);
```
prateekm: Of the stream processor, not container.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -176,30 +181,38 @@ public void start() {
...
+            executorService.shutdownNow();
+          }
+        }
+      } catch (Exception exception) {
+        LOGGER.info(String.format("Shutdown of container: %s threw an exception.", container), exception);
```
prateekm: Should this be error? 
Also, "Exception shutting down container ...". This implies that container.shutdown is what threw the exception.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -176,30 +181,38 @@ public void start() {
...
     }
+    stopJobCoordinator();
+  }
 
-    if (!containerShutdownInvoked) {
-      LOGGER.info("Shutting down JobCoordinator from StreamProcessor");
-      jobCoordinator.stop();
-    }
-
+  private synchronized void stopJobCoordinator() {
```
prateekm: Why is this synchronized, and why does this need to be a separate method?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -213,32 +226,33 @@ public void onJobModelExpired() {
...
-              LOGGER.info("Shutting down container in onJobModelExpired for processor:" + processorId);
+              LOGGER.info("Shutting down the container: {} of stream processor: {} in onJobModelExpired.", container, processorId);
               container.pause();
               shutdownComplete = jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
-              LOGGER.info("ShutdownComplete=" + shutdownComplete);
+              LOGGER.info("Container shutdown completion status is: {} for stream processor: {}.", shutdownComplete, processorId);
             } catch (IllegalContainerStateException icse) {
               // Ignored since container is not running
-              LOGGER.info("Container was not running.", icse);
+              LOGGER.info(String.format("Container: %s of stream processor: %s is dead. Shutdown is NO-OP.", container, processorId), icse);
```
prateekm: This is not a no-op, since we set shutdownComplete = true. Also don't guess that its dead.
"Container ... for processsor ... is in an illegal state. Marking container shutdown as complete."

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -213,32 +226,33 @@ public void onJobModelExpired() {
...
+              LOGGER.warn(String.format("Shutdown of the container: %s was interrupted", container), e);
             }
-            LOGGER.info("Shutting down container done for pid=" + processorId + "; complete =" + shutdownComplete);
             if (!shutdownComplete) {
-              LOGGER.warn("Container " + container.toString() + " may not have shutdown successfully. " +
-                  "Stopping the processor.");
-              container = null;
-              stop();
+              LOGGER.warn("Container: {} shutdown was unsuccessful. Stopping the stream processor: {}.", container, processorId);
+              if (!shutdownComplete && scheduledSamzaContainerFuture != null && !scheduledSamzaContainerFuture.isDone()) {
```
prateekm: Why check shutdownComplete again?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -213,32 +226,33 @@ public void onJobModelExpired() {
...
+                scheduledSamzaContainerFuture.cancel(true);
+              }
+              stopJobCoordinator();
             } else {
-              LOGGER.debug("Container " + container.toString() + " shutdown successfully");
+              LOGGER.debug("Container shutdown completed for stream processor: {}.", processorId);
             }
           } else {
-            LOGGER.debug("Container " + container.toString() + " is not running.");
+            LOGGER.debug("Container: {} of the stream processor: {} is dead.", container, processorId);
```
prateekm: "Dead" is not a state for the container. Fix messages.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -257,21 +271,20 @@ public void onContainerStart() {
                 processorListener.onStart();
               }
             } else {
-              LOGGER.debug("StreamProcessorListener was notified of container start previously. Hence, skipping this time.");
+              LOGGER.debug("Received duplicate container start notification in stream processor: {}.", processorId);
```
prateekm: Log container ID if possible.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -257,21 +271,20 @@ public void onContainerStart() {
...
             }
           }
 
           @Override
           public void onContainerStop(boolean pauseByJm) {
             if (pauseByJm) {
-              LOGGER.info("Container " + container.toString() + " stopped due to a request from JobCoordinator.");
+              LOGGER.info("Container: {} of the stream processor: {} is stopped by the JobCoordinator.", container, processorId);
```
prateekm: "was stopped"

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -283,24 +296,25 @@ public void onContainerFailed(Throwable t) {
...
         };
 
         container = createSamzaContainer(processorId, jobModel);
         container.setContainerListener(containerListener);
-        LOGGER.info("Starting container " + container.toString());
-        executorService = Executors.newSingleThreadExecutor(new ThreadFactoryBuilder()
-            .setNameFormat("p-" + processorId + "-container-thread-%d").build());
-        executorService.submit(container::run);
+        LOGGER.info("Starting the container: {} for the stream processor: {}.", container, processorId);
+        ThreadFactory threadFactory = new ThreadFactoryBuilder().setNameFormat(String.format(CONTAINER_THREAD_NAME_FORMAT, processorId))
```
prateekm: Fix indentation.

prateekm:
Please ignore the outdated comments. Agree with @sborya that the previous version was making the shutdown sequence very complex.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -179,27 +184,36 @@ public synchronized void stop() {
     boolean containerShutdownInvoked = false;
     if (container != null) {
       try {
-        LOGGER.info("Shutting down container " + container.toString() + " from StreamProcessor");
+        LOGGER.info("Shutting down the container: {} of stream processor: {}.", container, processorId);
         container.shutdown();
+        LOGGER.info("Waiting {} milliseconds for the container: {} to shutdown.", taskShutdownMs, container);
+        jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
```
prateekm: The jcContainerShutdownLatch is for the JC to wait on container shutdown. Are you sure we can reuse it here? If so, let's rename it.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -179,27 +184,36 @@ public synchronized void stop() {
...
+        if (executorService != null) {
```
prateekm: Would this ever be null?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -179,27 +184,36 @@ public synchronized void stop() {
...
+          executorService.shutdown();
+          if (!executorService.isShutdown()) {
+            LOGGER.info("Forcefully shutting down the executor service of the container: {}.", container);
```
prateekm: For the stream processor, not the container.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -213,32 +227,35 @@ public void onJobModelExpired() {
           if (SamzaContainerStatus.NOT_STARTED.equals(status) || SamzaContainerStatus.STARTED.equals(status)) {
             boolean shutdownComplete = false;
             try {
-              LOGGER.info("Shutting down container in onJobModelExpired for processor:" + processorId);
+              LOGGER.info("Shutting down the container: {} of stream processor: {} in onJobModelExpired.", container, processorId);
```
prateekm: "Job model expired. Shutting down..."

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -213,32 +227,35 @@ public void onJobModelExpired() {
...
               container.pause();
               shutdownComplete = jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
-              LOGGER.info("ShutdownComplete=" + shutdownComplete);
+              LOGGER.info("Container shutdown completion status is: {} for stream processor: {}.", shutdownComplete, processorId);
+              if (scheduledSamzaContainerFuture != null && !scheduledSamzaContainerFuture.isDone()) {
```
prateekm: There are states where this could be null during execution. Please check if ignoring that is the right option here. If not, let it throw an NPE/IllegalStateException.

prateekm: Where is this future field updated? I only see it set once in the beginning. What happens when a container is restarted?
Also, it's only volatile, so these operations aren't atomic. Does that matter here?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -213,32 +227,35 @@ public void onJobModelExpired() {
...
-            LOGGER.info("Shutting down container done for pid=" + processorId + "; complete =" + shutdownComplete);
             if (!shutdownComplete) {
-              LOGGER.warn("Container " + container.toString() + " may not have shutdown successfully. " +
-                  "Stopping the processor.");
+              LOGGER.warn("Container: {} shutdown was unsuccessful. Stopping the stream processor: {}.", container, processorId);
               container = null;
               stop();
             } else {
-              LOGGER.debug("Container " + container.toString() + " shutdown successfully");
+              LOGGER.debug("Container shutdown completed for stream processor: {}.", processorId);
```
prateekm: You can make these info. Will help debug later.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
shanthoosh: This field is updated whenever SamzaContainer is started.
SamzaContainer is started only in the debounce thread (in response to handling the onNewJobModel zookeeper event).  
All runnable executions on the debounce thread are serialized(via DebounceScheduler task queue) and there're no expected concurrency problems.

volatile is used to address variable visibility problem.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
shanthoosh: Fixed.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -1086,6 +1086,10 @@ class SamzaContainer(
       hostStatisticsMonitor.stop()
     }
   }
+
+  override def toString: String = {
```
shanthoosh: Adding to fix the log messages of following form in StreamProcessor:
```
Container org.apache.samza.container.SamzaContainer@3668218b may not have shutdown successfully. Stopping the processor.
```

prateekm:
@shanthoosh please update ticket and PR title.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
prateekm: 1. The %d in the end is the thread id, which is useful to know if the thread got recreated by the ES. If you think that's not possible here, its OK to leave it out.
2. Even then, your update does not contain what you've suggested above. Please fix.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -66,12 +69,12 @@
...
-  private volatile SamzaContainer container = null;
+  @VisibleForTesting
+  volatile SamzaContainer container = null;
   private volatile Throwable containerException = null;
+  private volatile Future<?> scheduledSamzaContainerFuture = null;
 
-  // Latch used to synchronize between the JobCoordinator thread and the container thread, when the container is
-  // stopped due to re-balancing
-  /* package private */volatile CountDownLatch jcContainerShutdownLatch;
+  /* package private */volatile CountDownLatch containerShutdownLatch;
```
prateekm: Don't need to add package private, it's implied. Also, why is this package private?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -179,27 +182,36 @@ public synchronized void stop() {
...
     if (container != null) {
       try {
-        LOGGER.info("Shutting down container " + container.toString() + " from StreamProcessor");
+        LOGGER.info("Shutting down the container: {} of stream processor: {}.", container, processorId);
         container.shutdown();
+        LOGGER.info("Waiting {} milliseconds for the container: {} to shutdown.", taskShutdownMs, container);
+        containerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
+        if (executorService != null) {
+          executorService.shutdown();
+          if (!executorService.isShutdown()) {
```
prateekm: Issue: This will always be false, since isShutdown returns whether shutdown or shutdownNow has been invoked by the user, not whether any threads are currently executing.

shanthoosh: The right method to invoke there is isTerminated, which returns true if all the tasks in that `executorService` had completed. However we have to invoke shutdown or shutdownNow before that(from isTerminated javadoc).
Switched to that.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
shanthoosh: Just to clarify this was not introduced as a part of this change. 
- Removed the comment as per your suggestion.
- It is set explicitly in StreamProcessor tests.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
shanthoosh: Switched to use isTerminated method, which returns true if all the tasks in a `executorService` had completed.

However we should invoke  executorService.shutdown or executorService.shutdownNow prior to that(from isTerminated javadoc).

When the initial container shutdown was successful, `isTerminated` check prevents an additional `task.shutdown.ms` wait.

shanthoosh:
@prateekm
Discussed offline. 
Removing all the changes in StreamProcessor/ZkJobCoordinator shutdown sequence and making this as a logging improvement patch.
Created a JIRA for shutdown sequence improvement in StreamProcessor/ZkJobCoordinator.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -55,6 +56,7 @@
 @InterfaceStability.Evolving
 public class StreamProcessor {
   private static final Logger LOGGER = LoggerFactory.getLogger(StreamProcessor.class);
+  private static final String CONTAINER_THREAD_NAME_FORMAT = "Samza StreamProcessor-%s-container-thread";
```
prateekm: Please fix thread name. "Samza StreamProcessor Container Thread-%d"

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -179,27 +181,27 @@ public synchronized void stop() {
...
       try {
-        LOGGER.info("Shutting down container " + container.toString() + " from StreamProcessor");
+        LOGGER.info("Shutting down the container: {} of stream processor: {}.", container, processorId);
         container.shutdown();
+        LOGGER.info("Waiting {} milliseconds for the container: {} to shutdown.", taskShutdownMs, container);
         containerShutdownInvoked = true;
-      } catch (IllegalContainerStateException icse) {
-        LOGGER.info("Container was not running", icse);
+      } catch (Exception exception) {
+        LOGGER.error(String.format("Shutdown of container: %s threw an exception.", container), exception);
```
prateekm: Mention that you're ignoring this.

"Ignoring error during shutdown of container."

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -179,27 +181,27 @@ public synchronized void stop() {
...
-
   }
 
   SamzaContainer createSamzaContainer(String processorId, JobModel jobModel) {
     return SamzaContainer.apply(
         processorId,
         jobModel,
         config,
-        Util.<String, MetricsReporter>javaMapAsScalaMap(customMetricsReporter),
+        Util.javaMapAsScalaMap(customMetricsReporter),
```
prateekm: Thanks.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -213,32 +215,30 @@ public void onJobModelExpired() {
           if (SamzaContainerStatus.NOT_STARTED.equals(status) || SamzaContainerStatus.STARTED.equals(status)) {
             boolean shutdownComplete = false;
             try {
-              LOGGER.info("Shutting down container in onJobModelExpired for processor:" + processorId);
+              LOGGER.info("Job model expired. Shutting down the container: {} of stream processor: {}.", container, processorId);
```
prateekm: Possible to log model version number?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -213,32 +215,30 @@ public void onJobModelExpired() {
...
               container.pause();
               shutdownComplete = jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
-              LOGGER.info("ShutdownComplete=" + shutdownComplete);
+              LOGGER.info("Container shutdown completion status is: {} for stream processor: {}.", shutdownComplete, processorId);
             } catch (IllegalContainerStateException icse) {
               // Ignored since container is not running
-              LOGGER.info("Container was not running.", icse);
+              LOGGER.info(String.format("Container: %s of stream processor: %s is not running.", container, processorId), icse);
```
prateekm: "Cannot shutdown container: %s for processor: %s. Container is not running."

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -213,32 +215,30 @@ public void onJobModelExpired() {
           if (SamzaContainerStatus.NOT_STARTED.equals(status) || SamzaContainerStatus.STARTED.equals(status)) {
             boolean shutdownComplete = false;
             try {
-              LOGGER.info("Shutting down container in onJobModelExpired for processor:" + processorId);
+              LOGGER.info("Job model expired. Shutting down the container: {} of stream processor: {}.", container, processorId);
               container.pause();
               shutdownComplete = jcContainerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
-              LOGGER.info("ShutdownComplete=" + shutdownComplete);
+              LOGGER.info("Container shutdown completion status is: {} for stream processor: {}.", shutdownComplete, processorId);
```
prateekm: Log container ID.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -213,32 +215,30 @@ public void onJobModelExpired() {
...
             } catch (IllegalContainerStateException icse) {
               // Ignored since container is not running
-              LOGGER.info("Container was not running.", icse);
+              LOGGER.info(String.format("Container: %s of stream processor: %s is not running.", container, processorId), icse);
               shutdownComplete = true;
             } catch (InterruptedException e) {
               Thread.currentThread().interrupt();
-              LOGGER.warn("Container shutdown was interrupted!" + container.toString(), e);
+              LOGGER.warn(String.format("Shutdown of the container: %s was interrupted", container), e);
```
prateekm: Log processor ID.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -213,32 +215,30 @@ public void onJobModelExpired() {
...
-            LOGGER.info("Shutting down container done for pid=" + processorId + "; complete =" + shutdownComplete);
             if (!shutdownComplete) {
-              LOGGER.warn("Container " + container.toString() + " may not have shutdown successfully. " +
-                  "Stopping the processor.");
+              LOGGER.warn("Container: {} shutdown was unsuccessful. Stopping the stream processor: {}.", container, processorId);
               container = null;
               stop();
             } else {
-              LOGGER.debug("Container " + container.toString() + " shutdown successfully");
+              LOGGER.info("Container shutdown completed for stream processor: {}.", processorId);
```
prateekm: Log container ID? Or is it null already?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -257,19 +257,19 @@ public void onContainerStart() {
                 processorListener.onStart();
               }
             } else {
-              LOGGER.debug("StreamProcessorListener was notified of container start previously. Hence, skipping this time.");
+              LOGGER.debug("Received duplicate container start notification for container: {} in stream processor: {}.", container, processorId);
```
prateekm: Warn.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -283,24 +283,24 @@ public void onContainerFailed(Throwable t) {
               LOGGER.warn("JobCoordinatorLatch was null. It is possible for some component to be waiting.");
             }
             containerException = t;
-            LOGGER.error("Container failed. Stopping the processor.", containerException);
+            LOGGER.error(String.format("Container: %s failed. Stopping the stream processor: %s.", container, processorId), containerException);
```
prateekm: "... failed with an exception. Stopping the stream processor: %s. Original Exception:"

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -312,11 +312,11 @@ public void onCoordinatorStop() {
       }
 
       @Override
-      public void onCoordinatorFailure(Throwable e) {
-        LOGGER.info("Coordinator Failed. Stopping the processor.");
+      public void onCoordinatorFailure(Throwable throwable) {
+        LOGGER.info(String.format("Coordinator failed with an exception. Stopping the stream processor: %s.", processorId), throwable);
```
prateekm: "stream processor: %s. Original exception:"

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
shanthoosh: It will be null already.

PR #443, opened 2018-03-09
SAMZA-1611 : BootstrappingChooser should use systemAdmin offsetComparator API to compare the offsets

atoomula:


vjagadish1989:
Is there a JIRA that tracks this? If not, consider creating one and linking it here? It will be helpful for others who don't have as much context.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -266,9 +266,17 @@ class BootstrappingChooser(
...
+      -1
+    } else {
+      val systemAdmin = systemAdmins.getSystemAdmin(systemStreamPartition.getSystem)
+      systemAdmin.offsetComparator(offset, offsetToCheck)
+    }
+
+    // The SSP is no longer lagging if the envelope's offset is greater than or equal to the
     // latest offset.
-    if (offset != null && offset.equals(offsetToCheck)) {
+    if (comparatorResult != null && comparatorResult >= 0) {
```
vjagadish1989: In most Java compilers, `comparatorResult == k` , will get compiled to `comparatorResult.compareTo(k) == 0` due to auto-boxing of types.

can make it explicit on need the for the null check - by using `comparatorResult.intValue() >= 0` instead.

vjagadish1989 (approved):
approved, thanks for the contributions. ship it!

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
prateekm: Just curious, can you explain what the intValue achieves? It would be an Integer comparision either way, right? Trying to avoid auto-boxing?

PR #444, opened 2018-03-12
SAMZA-1615: Fix a couple of issues in ControlMessageSender

xinyuiscool:
Two issues I found during testing: 1) medaDataCache.getSystemStreamMetadata(): if we pass in partitionOnly to be true, it will actually not store the metadata into the cache, resulting every call being another query to kafka. I turned off the partitionOnly in order to make it in the cache. 2) change the log for info to debug.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
+++ b/samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
@@ -43,9 +43,9 @@ class ControlMessageSender {
   }
 
   void send(ControlMessage message, SystemStream systemStream, MessageCollector collector) {
-    SystemStreamMetadata metadata = metadataCache.getSystemStreamMetadata(systemStream, true);
+    SystemStreamMetadata metadata = metadataCache.getSystemStreamMetadata(systemStream, false);
     int partitionCount = metadata.getSystemStreamPartitionMetadata().size();
-    LOG.info(String.format("Broadcast %s message from task %s to %s with %s partition",
+    LOG.debug(String.format("Broadcast %s message from task %s to %s with %s partition",
         MessageType.of(message).name(), message.getTaskName(), systemStream, partitionCount));
 
     for (int i = 0; i < partitionCount; i++) {
--- a/samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
+++ b/samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
@@ -29,12 +29,16 @@ import org.apache.samza.task.MessageCollector;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
 
 /**
  * This is a helper class to broadcast control messages to each partition of an intermediate stream
  */
 class ControlMessageSender {
   private static final Logger LOG = LoggerFactory.getLogger(ControlMessageSender.class);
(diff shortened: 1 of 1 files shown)


samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
```diff
@@ -43,9 +47,14 @@
   }
 
   void send(ControlMessage message, SystemStream systemStream, MessageCollector collector) {
-    SystemStreamMetadata metadata = metadataCache.getSystemStreamMetadata(systemStream, true);
-    int partitionCount = metadata.getSystemStreamPartitionMetadata().size();
-    LOG.info(String.format("Broadcast %s message from task %s to %s with %s partition",
+    Integer partitionCount = partitions.get(systemStream);
```
vjagadish1989: Simplify this by using .putIfAbsent / computeIfAbsent?

samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
```diff
@@ -29,12 +29,16 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
 
 /**
  * This is a helper class to broadcast control messages to each partition of an intermediate stream
  */
 class ControlMessageSender {
   private static final Logger LOG = LoggerFactory.getLogger(ControlMessageSender.class);
+  private static final Map<SystemStream, Integer> partitions = new ConcurrentHashMap<>();
```
vjagadish1989: s/partitions/partitionCounts maybe?

vjagadish1989 (approved):
approved, looks good to me. thank you for the contribution

samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
sborya: +1

samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
sborya: or even partitionCountCache

vjagadish1989: +1

xinyuiscool: Fixed. Changed it to PARTITION_COUNT_CACHE.

samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
xinyuiscool: Good suggestion. computeIfAbsent works pretty nicely here.

samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
```diff
@@ -43,9 +48,15 @@
   }
 
   void send(ControlMessage message, SystemStream systemStream, MessageCollector collector) {
-    SystemStreamMetadata metadata = metadataCache.getSystemStreamMetadata(systemStream, true);
-    int partitionCount = metadata.getSystemStreamPartitionMetadata().size();
-    LOG.info(String.format("Broadcast %s message from task %s to %s with %s partition",
+    Integer partitionCount = PARTITION_COUNT_CACHE.computeIfAbsent(systemStream, ss -> {
+        SystemStreamMetadata metadata = metadataCache.getSystemStreamMetadata(ss, true);
+        if (metadata == null) {
```
sborya: thanks for adding the check.

PR #446, opened 2018-03-12
sql: initial insert support for table destination

pdu-mn1:
Table is another main type of IO abstraction in Samza which supports
both read and write (optional). For the table that do support writes, we
should be able to allow SamzaSQL users to write a query to do that. One
example is to insert into a database. The current code only supports
inserting into a stream. This change adds the initial support for table
insert operation.


prateekm:
@pdu-mn1 Is this PR still relevant? If not, can you please close it? Thanks!

PR #450, opened 2018-03-20
ScheduleAfterDebounceTime should catch all Throwable to avoid losing unhandled Errors

thunderstumpges:
If an Action executed in the scheduler throws an Error (or other Throwable?) besides Exception, it is silently lost since the Action/Runnable wrapper only catches Exception, not Throwable. This made my troubleshooting of an issue very difficult. Made it seem like the code was "hung" when really it had thrown a "NoSuchMethodError" (Error instead of Exception) due to a simple dependency issue on my side.

Catching Throwable instead ensures this is handled and propagated properly.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
@@ -93,6 +93,8 @@ public class ScheduleAfterDebounceTime {
    * and all pending enqueued tasks will be cancelled.
    */
   public synchronized void stopScheduler() {
+    LOG.info("Stopping Scheduler");
+
     scheduledExecutorService.shutdownNow();
 
     // Clear the existing future handles.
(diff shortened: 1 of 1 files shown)


samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -93,6 +93,8 @@ public synchronized void scheduleAfterDebounceTime(String actionName, long delay
    * and all pending enqueued tasks will be cancelled.
    */
   public synchronized void stopScheduler() {
+    LOG.info("Stopping Scheduler");
```
prateekm: 👍

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -159,7 +161,7 @@ private Runnable getScheduleableAction(String actionName, Runnable runnable) {
    *
    * @param exception the exception happened during task execution.
    */
-  private void doCleanUpOnTaskException(Exception exception) {
+  private void doCleanUpOnTaskException(Throwable exception) {
```
prateekm: Nitpick: Let's rename param to 'throwable' for consistency.

PR #451, opened 2018-03-20
SAMZA-1621: Delete the ephemeral processor node in StreamProcessor shutdown phase.

shanthoosh:


shanthoosh:

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -152,6 +152,19 @@ public synchronized String registerProcessorAndGetId(final ProcessorData data) {
     return ephemeralPath;
   }
 
+  /**
+   * Deletes the ephemeral node of a processor in zookeeper.
+   * @param processorId uniqueId identifying the stream processor to delete.
+   */
+  public void deleteProcessorNode(String processorId) {
+    for (ProcessorNode processorNode : getAllProcessorNodes()) {
+      if (processorNode.getProcessorData().getProcessorId().equals(processorId)) {
```
vjagadish: do u need to guard against NPE's anywhere here?

samza-core/src/main/java/org/apache/samza/zk/ZkControllerImpl.java
```diff
@@ -87,6 +87,8 @@ public void stop() {
       zkLeaderElector.resignLeadership();
     }
 
+    zkUtils.deleteProcessorNode(processorId);
```
vjagadish: Should this be done during finally so that it is always executed? it looks like any exception during `resignLeadership` will not trigger the delete?

shanthoosh: resignLeadership simply overrides a boolean variable's value.

Agree that it will be safer to move this to finally block. Moved it to finally block.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
shanthoosh: `getAllProcessorNodes()` has sufficient guards for NPE and will not return null values for any fields . Adding the same checks here would be unnecessary.

samza-core/src/main/java/org/apache/samza/zk/ZkControllerImpl.java
vjagadish: `resignLeadership` is an interface - It's just that the Zk implementation is implementing it in a certain way by returning a boolean value.

vjagadish1989 (approved):
approved. looks good. please also fix all shutdown messages to be `INFO` level in `ZKJC`

samza-core/src/main/java/org/apache/samza/zk/ZkControllerImpl.java
```diff
@@ -83,12 +83,12 @@ public boolean isLeader() {
...
-    }
-
-    // close zk connection
-    if (zkUtils != null) {
+    try {
+      if (isLeader()) {
+        zkLeaderElector.resignLeadership();
+      }
+    } finally {
+      zkUtils.deleteProcessorNode(processorId);
```
vjagadish1989: `zkUtils.deleteProcessorNode()` should internally catch all exceptions since it's on the finally block.

vjagadish1989 (approved):
approved. Ship it!

shanthoosh:
@vjagadish1989 
Added a retry based on sessionTimeOut in ZkUtils.registration if there're duplicate processors. Can you take a look.

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -132,24 +134,60 @@ public ZkKeyBuilder getKeyBuilder() {
...
-        metrics.deletions.inc();
-        throw new SamzaException(String.format("Processor: %s is duplicate in the group. Registration failed.", processorId));
+      while (true) {
+        ProcessorNode processorNode = new ProcessorNode(data, ephemeralPath);
+        // Determine if there are duplicate processors with this.processorId after registration.
+        if (!isValidRegisteredProcessor(processorNode)) {
+          long currentTime = System.currentTimeMillis();
+          if ((currentTime - startTime) < retryTimeOutMs) {
+            LOG.info("Processor: {} is duplicate. Retrying registration again.", processorId);
+            timeDelay(1000);
```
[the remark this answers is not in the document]
shanthoosh: Done.

shanthoosh:
@vjagadish1989 
Rebased with master and fixed the merge conflicts.

PR #454, opened 2018-03-23
Skipping large messages in the EventHub system producer

srinipunuru:
EventHubs have restriction on maximum message sizes that can be allowed. Adding a `systems.%s.eventhubs.skipMessagesLargerThanBytes` that can be used in the event hub system to make it skip messages that are larger than specific bytes so that we don't even try to send those large messages because EventHubs will reject them anyways.

### DIFF
--- a/samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
@@ -213,9 +213,11 @@ public class EventHubSystemProducer implements SystemProducer {
 
     EventData eventData = createEventData(streamId, envelope);
     int eventDataLength = eventData.getBytes() == null ? 0 : eventData.getBytes().length;
-    if (eventDataLength > this.maxMessageSize) {
-      LOG.info("Received a message with size {} > maxMessageSize configured {(}), Skipping it", eventDataLength,
-          this.maxMessageSize);
+
+    // If the maxMessageSize is lesser than zero, then it means there is no message size restriction.
+    if (this.maxMessageSize > 0 && eventDataLength > this.maxMessageSize) {
+      LOG.info("Received a message (Key:{}) with size {} > maxMessageSize configured {(}), Skipping it",
+          envelope.getKey() == null ? "null" : String.valueOf(envelope.getKey()), eventDataLength, this.maxMessageSize);
       eventSkipRate.get(streamId).inc();
       aggEventSkipRate.inc();
       return;
--- a/samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
@@ -216,8 +216,8 @@ public class EventHubSystemProducer implements SystemProducer {
 
     // If the maxMessageSize is lesser than zero, then it means there is no message size restriction.
     if (this.maxMessageSize > 0 && eventDataLength > this.maxMessageSize) {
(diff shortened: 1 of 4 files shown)


samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -204,7 +212,15 @@ public synchronized void send(String source, OutgoingMessageEnvelope envelope) {
     checkCallbackThrowable("Received exception on message send");
 
     EventData eventData = createEventData(streamId, envelope);
-    int eventDataLength =  eventData.getBytes() == null ? 0 : eventData.getBytes().length;
+    int eventDataLength = eventData.getBytes() == null ? 0 : eventData.getBytes().length;
+    if (eventDataLength > this.maxMessageSize) {
+      LOG.info("Received a message with size {} > maxMessageSize configured {(}), Skipping it", eventDataLength,
```
lhaiesp: I have a little bit concern about this. We never know whether one day we will be streaming a pipeline that has many many large messages. Some naive restriction on the per event logging would be nice.

lhaiesp:
Wei can probably weigh in. But I remember skipBadMessage is not just about large message. It's about some schema issue with Gobblin as well. So what is our design here? Will that be addressed separated in LI code?

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
weisong44: Can we also log some identifier for the message, such as the message key?

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -204,7 +212,15 @@ public synchronized void send(String source, OutgoingMessageEnvelope envelope) {
     checkCallbackThrowable("Received exception on message send");
 
     EventData eventData = createEventData(streamId, envelope);
-    int eventDataLength =  eventData.getBytes() == null ? 0 : eventData.getBytes().length;
+    int eventDataLength = eventData.getBytes() == null ? 0 : eventData.getBytes().length;
+    if (eventDataLength > this.maxMessageSize) {
```
weisong44: Should we not also check the skipBadMessage flag is turned on?

weisong44: We should also disable the handling of badMessage flag in EventProducer

srinipunuru: skipBadMessage config is datastream config which gets translated into skipLargeMessageConfig for now. 

Added a way to disable the skipLargeMessage. If `skipMessagesLargerThanBytes` is < 0 then we skip the check.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
srinipunuru: As we discussed offline, key may contain the sensitive information. So we cannot log the key or the payload.

srinipunuru: Logging the skipped message is a requirement from the customer, So keeping it as is.

srinipunuru:
@lhaiesp Right now the only bad message that we skip is large message so the skipBadMessage datastream config will be converted into skipLargeMesage right now. In future skipBadMesage datastream config can be converted into any other config that we may add in future. Discussed the above behavior with @weisong44 and he is fine with it.

PR #456, opened 2018-03-26
SAMZA-1627: Watermark broadcast enhancements

xinyuiscool:
Currently each upstream task needs to broadcast to every single partition of intermediate streams in order to aggregate watermarks in the consumers. A better way to do this is to have only one downstream consumer doing the aggregation, and then broadcast to all the partitions. This is safe as we can prove the broadcast watermark message is after all the upstream tasks finished producing the events whose event time are before the watermark. This reduced the full message count from O(n^2) to O(n).

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
+++ b/samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
@@ -26,6 +26,7 @@ import org.apache.samza.system.OutgoingMessageEnvelope;
 import org.apache.samza.system.StreamMetadataCache;
 import org.apache.samza.system.SystemStream;
 import org.apache.samza.system.SystemStreamMetadata;
+import org.apache.samza.system.SystemStreamPartition;
 import org.apache.samza.task.MessageCollector;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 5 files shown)


samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
```diff
@@ -48,20 +49,35 @@
   }
 
   void send(ControlMessage message, SystemStream systemStream, MessageCollector collector) {
-    Integer partitionCount = PARTITION_COUNT_CACHE.computeIfAbsent(systemStream, ss -> {
+    int partitionCount = getPartitionCount(systemStream);
+    int aggregatePartition = systemStream.getStream().hashCode() % partitionCount;
```
sborya: please add a comment explaining what aggregatePartition is.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -265,6 +267,11 @@ public final void aggregateEndOfStream(EndOfStreamMessage eos, SystemStreamParti
     SystemStream stream = ssp.getSystemStream();
     if (eosStates.isEndOfStream(stream)) {
       LOG.info("Input {} reaches the end for task {}", stream.toString(), taskName.getTaskName());
+      if (eos.getTaskName() != null) {
```
sborya: I don't understand this condition? In what case task name is null? And why we send broadcast in case it is not NULL?

sborya: please add comments for this.

samza-core/src/main/java/org/apache/samza/operators/impl/ControlMessageSender.java
xinyuiscool: Added the comment.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
xinyuiscool: Chatted offline. This check verifies this is the aggregation task and needs to broadcast the result to other partitions. Added the comment.

xinyuiscool: Right, already added.

samza-core/src/main/java/org/apache/samza/operators/impl/WatermarkStates.java
```diff
@@ -71,6 +71,8 @@ synchronized void update(long timestamp, String taskName) {
       if (timestamps.size() == expectedTotal) {
         Optional<Long> min = timestamps.values().stream().min(Long::compare);
         watermarkTime = min.orElse(timestamp);
+      } else if (taskName == null) {
```
sborya: Please clarify the execution path here + comments.

xinyuiscool: Fixed! Thanks for pointing this out.

PR #457, opened 2018-03-27
SAMZA-1632: KinesisConfig: Making getProxyHost and getProxyPort APIs protected

atoomula:


samza-aws/src/main/java/org/apache/samza/system/kinesis/KinesisConfig.java
```diff
@@ -169,10 +157,22 @@ String getProxyHost(String system) {
...
    * @return proxy port number or 0 if not defined
    */
-  int getProxyPort(String system) {
+  protected int getProxyPort(String system) {
     return getInt(String.format(CONFIG_PROXY_PORT, system), DEFAULT_CONFIG_PROXY_PORT);
   }
 
+  /**
+   * @param system name of the system
+   * @return {@link ClientConfiguration} which has options controlling how the client connects to kinesis
```
vjagadish: Thanks for organizing it better!

PR #458, opened 2018-03-27
Refactoring the EventHub System Producer into reusable AsyncSystemProducer and NoFlushAsyncSystemProducer

srinipunuru:
Refactoring eventhub system producer into common reusable components
1. AsyncSystemProducer : All the system producers that have real async send API with call backs can use this
2. NoFlushAsyncSystemProducer: All the system producers that have real async call back based send API and doesn't provide flush semantics can use this.

The system producers that implement AsyncSystemproducers can be used across Samza and Brooklin.

TODO: AsyncSystemProducer needs to be moved to the api layer so that it can be used across different system producers (i.e. eventhub and kinesis)


samza-azure/src/main/java/org/apache/samza/NoFlushAsyncSystemProducer.java
```diff
@@ -0,0 +1,68 @@
...
+      // Block until all the pending sends are complete or timeout.
+      future.get(DEFAULT_FLUSH_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      incompleteSends = pendingFutures.stream().filter(x -> !x.isDone()).count();
+      String msg = String.format("Flush failed with error. Total pending sends %d", incompleteSends);
+      LOG.error(msg, e);
+      throw new SamzaException(msg, e);
+    }
+
+    pendingFutures.clear();
```
misanc: here you are clearing the pendingFutures, but other message could have been added to pendingFutures by the send method. This could be a race condition, where the message is removed from the pendingFutures before the callback is completed.

Suggestion, remove the message from the pendingFutures after the callback is done (in the closure in the send method), or make a copy of the pendingFutures and use set::removeAll(<copy of pendingFutures>) to allow other send to take place while flushing.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -85,16 +79,12 @@
   private final HashMap<String, Counter> eventWriteRate = new HashMap<>();
   private final HashMap<String, Counter> eventByteWriteRate = new HashMap<>();
   private final HashMap<String, SamzaHistogram> sendLatency = new HashMap<>();
```
misanc: one question: can you move the sendLatency logic to the AsyncSystemProducer class?

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
```diff
@@ -0,0 +1,169 @@
...
+    checkCallbackThrowable("Received exception on message send");
+    CompletableFuture<Void> sendResult = sendAsync(source, envelope);
+    pendingFutures.add(sendResult);
+
+    String streamName = envelope.getSystemStream().getStream();
+    String streamId = physicalToStreamIds.getOrDefault(streamName, streamName);
+    long afterSendTimeMs = System.currentTimeMillis();
+
+    // Auto update the metrics and possible throwable when futures are complete.
+    sendResult.handle((aVoid, throwable) -> {
```
lhaiesp: is this a blocking call?

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
```diff
@@ -0,0 +1,169 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * SystemProducer Helper that can be used to implement SystemProducer that provide async send APIs
+ * e.g. Kinesis, Event Hubs, etc..
+ */
+public abstract class AsyncSystemProducer implements SystemProducer {
+
+  private static final Logger LOG = LoggerFactory.getLogger(NoFlushAsyncSystemProducer.class.getName());
```
lhaiesp: wrong class name

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
```diff
@@ -0,0 +1,169 @@
...
+import org.apache.samza.system.eventhub.metrics.SamzaHistogram;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * SystemProducer Helper that can be used to implement SystemProducer that provide async send APIs
+ * e.g. Kinesis, Event Hubs, etc..
+ */
+public abstract class AsyncSystemProducer implements SystemProducer {
```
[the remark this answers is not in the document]
srinipunuru: Good point i was thinking about the same. I think we need some really deep extensive coverage on this file when we move this into Samza-api. But for this refactoring, these files are covered by the unit tests on the TestEventHubSystemProducer. I will open a work item to move this to api and add tests.

samza-azure/src/main/java/org/apache/samza/NoFlushAsyncSystemProducer.java
srinipunuru: I think removing the specific future would involve having a mechanism to lookup that future base on the message id or someother id. I think the initial eventHubSystemProducer had that logic. I believe Daniel removed it and made it simpler. I like the current approach we have where we clear the pendingFutures on flush. 

flush is synchronized with send, So i don't think there should be race between them.

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
```diff
@@ -0,0 +1,169 @@
...
+      throw new SamzaException(msg, sendThrowable);
+    }
+  }
+
+  /**
+   * Gets pending futures.
+   *
+   * @return the pending futures
+   */
+  public Collection<CompletableFuture<Void>> getPendingFutures() {
```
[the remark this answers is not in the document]
srinipunuru: It's actually used by the test TestEventHubSystemProducer. I made it package private and moved it to EventHubSystemProducer.

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
```diff
@@ -0,0 +1,169 @@
...
+  public abstract CompletableFuture<Void> sendAsync(String source, OutgoingMessageEnvelope envelope);
+
+  /**
+   * Check callback throwable.
+   *
+   * @param msg the msg
+   */
+  protected void checkCallbackThrowable(String msg) {
+    // Check for send errors from EventHub side
+    Throwable sendThrowable = sendExceptionOnCallback.get();
```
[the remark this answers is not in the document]
srinipunuru: When there is an exception on callback, Only way to clear it is to recreate the system producer. Because it is possible that the connection is broken.

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
```diff
@@ -0,0 +1,169 @@
...
+    // Auto update the metrics and possible throwable when futures are complete.
+    sendResult.handle((aVoid, throwable) -> {
+        long callbackLatencyMs = System.currentTimeMillis() - afterSendTimeMs;
+        sendCallbackLatency.get(streamId).update(callbackLatencyMs);
+        aggSendCallbackLatency.update(callbackLatencyMs);
+        if (throwable != null) {
+          sendErrors.get(streamId).inc();
+          aggSendErrors.inc();
+          LOG.error("Send message to event hub: {} failed with exception: ", streamId, throwable);
+          sendExceptionOnCallback.compareAndSet(null, throwable);
```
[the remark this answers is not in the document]
srinipunuru: We are logging all the exceptions. The exception that will be thrown to the sender is the last exception thrown.

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
```diff
@@ -0,0 +1,169 @@
...
+    this.registry = registry;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public void send(String source, OutgoingMessageEnvelope envelope) {
+    checkCallbackThrowable("Received exception on message send");
+    CompletableFuture<Void> sendResult = sendAsync(source, envelope);
+    pendingFutures.add(sendResult);
```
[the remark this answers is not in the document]
srinipunuru: I think removing the specific future would involve having a mechanism to lookup that future base on the message id or someother id. I think the initial eventHubSystemProducer had that logic. I believe Daniel removed it and made it simpler. I like the current approach we have where we clear the pendingFutures on flush.

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
srinipunuru: Not sure i understood your questions. Lambda inside sendResult.handle gets called after the send is complete,  It is a callback.  sendResult.handle just hooks up the callback so it is not blocking on the handle logic.

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
```diff
@@ -0,0 +1,169 @@
...
+
+  private final AtomicReference<Throwable> sendExceptionOnCallback = new AtomicReference<>(null);
+
+  private static Counter aggSendErrors = null;
+  private static SamzaHistogram aggSendCallbackLatency = null;
+  private final HashMap<String, SamzaHistogram> sendCallbackLatency = new HashMap<>();
+  private final HashMap<String, Counter> sendErrors = new HashMap<>();
+
+  /**
+   * Instantiates a new Async system producer.
```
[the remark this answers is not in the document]
srinipunuru: I added the comments, But i am not sure whether we need to be adding such comments if there is an abstract method.  compiler or IDE will give hints on what needs to be overridden.

srinipunuru:
@lhaiesp Any system that provides callbacks as well as flush (e.g. kinesis and kafka) should extend AsyncSystemProducer. Event hubs is the one of the few systems that doesn't provide flush semantics.

samza-azure/src/main/java/org/apache/samza/NoFlushAsyncSystemProducer.java
```diff
@@ -0,0 +1,68 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(NoFlushAsyncSystemProducer.class.getName());
+  private static final long DEFAULT_FLUSH_TIMEOUT_MILLIS = Duration.ofMinutes(1L).toMillis();
+
+  public NoFlushAsyncSystemProducer(String systemName, Config config, MetricsRegistry registry) {
+    super(systemName, config, registry);
+  }
+
+  @Override
+  public synchronized void flush(String source) {
```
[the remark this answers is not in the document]
misanc: send does not have syncronized

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
```diff
@@ -0,0 +1,182 @@
...
+  protected final Map<String, String> physicalToStreamIds;
+  /**
+   * The Registry.
+   */
+  protected final MetricsRegistry registry;
+
+  /**
+   * The Pending futures.
+   */
+  protected final Set<CompletableFuture<Void>> pendingFutures = ConcurrentHashMap.newKeySet();
```
misanc: comment: If you decide to  synchronize access to this Set, you can use a simple HashSet here.

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
```diff
@@ -0,0 +1,182 @@
...
+    physicalToStreamIds =
+        streamIds.stream().collect(Collectors.toMap(sconfig::getPhysicalName, Function.identity()));
+    this.registry = registry;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void send(String source, OutgoingMessageEnvelope envelope) {
```
misanc: comment: If you decide to synchronize access to this pendingFutures, I suggest you add a synchronized modified to this method.

samza-azure/src/main/java/org/apache/samza/NoFlushAsyncSystemProducer.java
misanc: You are using a set, you can simple use pendingFutures::remove as part of the handle (in method send). 

 sendResult.handle((aVoid, throwable) -> {
           pendingFutures.remove(sendResult);
       ....
}


===================
Another comment:
If a SystemProducer is not using the class NoFlushAsyncSystemProducer  (e.g. inherit directly from AsyncSystemProducer; I do not see how entries are removed from  pendingFutures.

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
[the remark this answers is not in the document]
misanc: you are using a set using the default Object::equals and Object:hash contract (the reference id).
So you can very easily just call pendingFutures.remove(sendResult) as part of the handle.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -186,17 +169,19 @@ public synchronized void start() {
...
-        aggSendErrors = registry.newCounter(AGGREGATE, SEND_ERRORS);
       }
     }
 
     isStarted = true;
   }
 
   @Override
-  public synchronized void send(String source, OutgoingMessageEnvelope envelope) {
+  public synchronized void flush(String source) {
```
misanc: I suggest to call super.flush() here.

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
srinipunuru: As discussed, We are removing the pendingFutures on callback. As an added protection for EventHubSystemProducer we will clear the pendingFutures once the flush is called.

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
srinipunuru: As discussed, We are removing the pendingFutures on callback. As an added protection for EventHubSystemProducer we will clear the pendingFutures once the flush is called.

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
```diff
@@ -0,0 +1,181 @@
...
+   */
+  protected final Set<CompletableFuture<Void>> pendingFutures = ConcurrentHashMap.newKeySet();
+
+  private final AtomicReference<Throwable> sendExceptionOnCallback = new AtomicReference<>(null);
+
+  private static Counter aggSendErrors = null;
+  private static SamzaHistogram aggSendLatency = null;
+  private static SamzaHistogram aggSendCallbackLatency = null;
+
+  private final HashMap<String, SamzaHistogram> sendLatency = new HashMap<>();
```
sborya: add a comment on what is the key.

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
```diff
@@ -0,0 +1,181 @@
...
+  /**
+   * Sends a specified message envelope from a specified Samza source.
+   * @param source String representing the source of the message.
+   * @param envelope Aggregate object representing the serialized message to send from the source.
+   * @return the completable future for the async send call
+   */
+  public abstract CompletableFuture<Void> sendAsync(String source, OutgoingMessageEnvelope envelope);
+
+  /**
+   * Check callback throwable.
```
sborya: Please explain what this method do here.

samza-azure/src/main/java/org/apache/samza/NoFlushAsyncSystemProducer.java
```diff
@@ -0,0 +1,68 @@
...
+      // Block until all the pending sends are complete or timeout.
+      future.get(DEFAULT_FLUSH_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      incompleteSends = pendingFutures.stream().filter(x -> !x.isDone()).count();
+      String msg = String.format("Flush failed with error. Total pending sends %d", incompleteSends);
+      LOG.error(msg, e);
+      throw new SamzaException(msg, e);
+    }
+
+    pendingFutures.clear();
```
sborya: This should be synchronized.

samza-azure/src/main/java/org/apache/samza/AsyncSystemProducer.java
```diff
@@ -0,0 +1,181 @@
...
+*
+* Unless required by applicable law or agreed to in writing,
+* software distributed under the License is distributed on an
+* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+* KIND, either express or implied.  See the License for the
+* specific language governing permissions and limitations
+* under the License.
+*/
+
+package org.apache.samza;
```
sborya: Until it is moved to samza-core please use a different package name.

samza-azure/src/main/java/org/apache/samza/NoFlushAsyncSystemProducer.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.samza.metrics.MetricsRegistry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Helper System producer that can be used to implement system producers whose APIs don't provide flush semantics
+ * but instead provides Async callbacks e.g. Event Hubs
+ */
+public abstract class NoFlushAsyncSystemProducer extends AsyncSystemProducer {
```
sborya: Do we really need a separate class here? Can we use the Async one with a default flush implementation?

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/AsyncSystemProducer.java
```diff
@@ -0,0 +1,215 @@
...
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public synchronized void send(String source, OutgoingMessageEnvelope envelope) {
+    checkForSendCallbackErrors("Received exception on message send");
+
+    String streamName = envelope.getSystemStream().getStream();
+    String streamId = physicalToStreamIds.getOrDefault(streamName, streamName);
```
sborya: using stream name if stream id is not available?

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/AsyncSystemProducer.java
```diff
@@ -0,0 +1,215 @@
...
+
+    long latencyMs = afterSendTimeMs - beforeSendTimeMs;
+    sendLatency.get(streamId).update(latencyMs);
+    aggSendLatency.update(latencyMs);
+
+    pendingFutures.add(sendResult);
+
+    // Auto update the metrics and possible throwable when futures are complete.
+    sendResult.handle((aVoid, throwable) -> {
+        pendingFutures.remove(sendResult);
```
sborya: this will be accessed in a different thread, correct?

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/AsyncSystemProducer.java
```diff
@@ -0,0 +1,215 @@
...
+    if (aggSendLatency == null) {
+      aggSendLatency = new SamzaHistogram(registry, AGGREGATE, SEND_LATENCY);
+      aggSendCallbackLatency = new SamzaHistogram(registry, AGGREGATE, SEND_CALLBACK_LATENCY);
+      aggSendErrors = registry.newCounter(AGGREGATE, SEND_ERRORS);
+    }
+  }
+
+  /**
+   * Default implementation of the flush that just waits for all the pendingFutures to be complete.
+   * SystemProducer should override this, If the underlying system provides flush semantics.
```
sborya: s/If/if/

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/AsyncSystemProducer.java
srinipunuru: I think this was added to support old style configs which doesn't use streamId, If there are no virtual streamIds used, we use the streamName as streamId in which case the expectation is `systems.%s.stream.list` should list the streamNames not the streamIds.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/AsyncSystemProducer.java
```diff
@@ -0,0 +1,215 @@
...
+  protected final Map<String, String> physicalToStreamIds;
+  /**
+   * The Registry.
+   */
+  protected final MetricsRegistry registry;
+
+  /**
+   * The Pending futures.
+   */
+  protected final Set<CompletableFuture<Void>> pendingFutures = ConcurrentHashMap.newKeySet();
```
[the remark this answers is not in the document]
srinipunuru: The reason why this was made concurrent is because this is accessed across send/flush thread and callback thread. I updated the comment to make it clear.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/AsyncSystemProducer.java
srinipunuru: Yes that is correct.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/AsyncSystemProducer.java
```diff
@@ -0,0 +1,215 @@
...
+  /**
+   * Instantiates a new Async system producer.
+   *
+   * @param systemName the system name
+   * @param config the config
+   * @param registry the registry
+   */
+  public AsyncSystemProducer(String systemName, Config config, MetricsRegistry registry) {
+    StreamConfig sconfig = new StreamConfig(config);
+    streamIds = config.getList(String.format(CONFIG_STREAM_LIST, systemName));
```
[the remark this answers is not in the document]
srinipunuru: Right now if the streamIds/streamNames are not present sends will throw, which i think is fine. I believe why this was made that way is to optimize on the startup time, which seems like a reasonable tradeoff.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/AsyncSystemProducer.java
```diff
@@ -0,0 +1,215 @@
...
+  /**
+   * Sends a specified message envelope from a specified Samza source.
+   * @param source String representing the source of the message.
+   * @param envelope Aggregate object representing the serialized message to send from the source.
+   * @return the completable future for the async send call
+   */
+  public abstract CompletableFuture<Void> sendAsync(String source, OutgoingMessageEnvelope envelope);
+
+  /**
+   * This method is used to check whether there were any previous exceptions in the send call backs.
```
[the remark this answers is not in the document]
srinipunuru: Thank you.

PR #459, opened 2018-03-28
Initial version for in memory system.

bharathkk:
Getting the PR out to unblock sanil.
Pending tasks
 1. Add documentation
 2. Add more tests
 3. Currently initialization of the stream happens using createStream on admin. We need to find a way to expose the same functionality to low level users.
 4. To populate the initial stream, clients need to get a handle on producer and produce the messages. Alternatively, we can support serialization of source data and pass it as config to the system to initialize. We have a hook in place for this but not implemented it completely.
  5. Clean up consumed data on the buffer based on the lowest offsets of the consumers.

I will create JIRAs for all the pending tasks and fix them iteratively.

@xinyuiscool ^^


Sanil15:
Thanks a lot for this Bharat, 

In regards to the pending task points above mentioned, here is my update:
3. This can be avoided by just spinning up a producer and producing to input stream, I will have a prototype for low level (Async and Sync) out soon, you can refer to that 
4. This too can be avoided since spinning up a producer in start initializes the stream, I have a working prototype for it. The idea of serializing it and again reading from config is redundant

One more thing to be considered in the system is the support of bounded streams, i.e we need to put an EndOfStreamMessage in a stream partition, this can only be done in consumer poll() as of current design.

Thanks

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
```diff
@@ -0,0 +1,173 @@
...
+
+  private List<IncomingMessageEnvelope> poll(SystemStreamPartition ssp, String offset) {
+    int startingOffset = Integer.parseInt(offset);
+    List<IncomingMessageEnvelope> messageEnvelopesForSSP = bufferedMessages.getOrDefault(ssp, new LinkedList<>());
+
+    if (startingOffset >= messageEnvelopesForSSP.size()) {
+      return new ArrayList<>();
+    }
+
+    return messageEnvelopesForSSP.subList(startingOffset, messageEnvelopesForSSP.size());
```
xinyuiscool: We do not remove the elements polled from the ssp?

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemAdmin.java
```diff
@@ -0,0 +1,154 @@
...
+   * offset1 == offset2 and offset1 &gt; offset2 respectively. Return
+   * null if those two offsets are not comparable
+   *
+   * @param offset1 First offset for comparison.
+   * @param offset2 Second offset for comparison.
+   * @return -1 if offset1 &lt; offset2; 0 if offset1 == offset2; 1 if offset1 &gt; offset2. Null if not comparable
+   */
+  @Override
+  public Integer offsetComparator(String offset1, String offset2) {
+    int o1 = Optional.ofNullable(offset1)
```
xinyuiscool: this logic is too hard to understand. I would refer a simple something like:

if(offset1 == null || offset2 == null) {
   return null;
} else {
   return Integer.valueOf(offset1).compare(Interger.valueOf(offset2));
}

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemAdmin.java
```diff
@@ -0,0 +1,154 @@
...
+   * Create a stream described by the spec.
+   *
+   * @param streamSpec  The spec, or blueprint from which the physical stream will be created on the system.
+   * @return            {@code true} if the stream was actually created and not pre-existing.
+   *                    {@code false} if the stream was pre-existing.
+   *                    A RuntimeException will be thrown if creation fails.
+   */
+  @Override
+  public boolean createStream(StreamSpec streamSpec) {
+    return Optional.ofNullable(inMemorySystemConfig.getSerializedDataSet(streamSpec.getId()))
```
xinyuiscool: I think you should just do a inMemoryManager.intializeStream(streamSpec, serializedData). Do a null check inside the function.

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  /**
+   * Populate the metadata for the {@link SystemStream} and initialize the buffer for {@link SystemStreamPartition}.
+   *
+   * @param streamSpec stream spec for the stream to be initialized
+   *
+   * @return true if successful, false otherwise
+   */
+  public boolean initializeStream(StreamSpec streamSpec) {
```
xinyuiscool: No need for this function. Use the following function with a null check.

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
```diff
@@ -0,0 +1,173 @@
...
+  public InMemoryManager() {
+    bufferedMessages = new ConcurrentHashMap<>();
+    systemStreamToPartitions = new ConcurrentHashMap<>();
+  }
+
+  private List<IncomingMessageEnvelope> newSynchronizedLinkedList() {
+    return  Collections.synchronizedList(new LinkedList<IncomingMessageEnvelope>());
+  }
+
+  public void put(SystemStreamPartition ssp, Object message) {
```
xinyuiscool: no need for this function. Use the following one with a null check or something.

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
```diff
@@ -0,0 +1,173 @@
...
+import org.apache.samza.system.SystemStreamMetadata;
+import org.apache.samza.system.SystemStreamPartition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Initial draft of in-memory manager. It is test only and not meant for production use right now.
+ */
+public class InMemoryManager {
```
xinyuiscool: This class should be package private and all the methods inside shouldn't be visible to public.

prateekm:
+1 on Xinyu's point about Java8 streams API making things harder to read. Would also prefer standard constructs (if-else/ for-each).

PR #465, opened 2018-04-05
sql: initial insert support for table destination

pdu-mn1:
Table is another main type of IO abstraction in Samza which supports
both read and write (optional). For the tables that do support writes, we
should be able to allow Samza SQL users to write a query to do that. One
example is to insert into a database. The current code only supports
inserting into a stream. This change adds the initial support for table
insert operation.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/interfaces/SourceResolver.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/interfaces/SourceResolver.java
@@ -19,8 +19,13 @@
 
 package org.apache.samza.sql.interfaces;
 
+import org.apache.samza.operators.TableDescriptor;
+
+
 /**
- * Source Resolvers are used by Samza Sql application to fetch the {@link SqlSystemSourceConfig} corresponding to the source.
+ * Source Resolvers are used by Samza Sql application to fetch the {@link SqlSystemSourceConfig} corresponding
+ * to the source. Additionally, it provides capability to resolve Samza table sources and creating the matching
+ * table descriptors.
  */
 public interface SourceResolver {
   /**
@@ -41,4 +46,12 @@ public interface SourceResolver {
    *  true if the source is a table, else false.
    */
   boolean isTable(String sourceName);
+
+  /**
+   * Return a Samza table descriptor corresponding to the {@code sourceName} if the underlying source is a table,
+   * ie. {@link SourceResolver#isTable(String)} should return true for the same {@code sourceName}.
+   * @param sourceName source for which the table descriptor is requested.
+   * @return table descriptor; null if the source is not a table.
+   */
+  TableDescriptor getTableDescriptor(String sourceName);
 }
--- a/samza-sql/src/test/java/org/apache/samza/sql/testutil/TestSourceResolverFactory.java
+++ b/samza-sql/src/test/java/org/apache/samza/sql/testutil/TestSourceResolverFactory.java
@@ -20,16 +20,123 @@
 package org.apache.samza.sql.testutil;
 
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.lang.NotImplementedException;
 import org.apache.samza.config.Config;
+import org.apache.samza.container.SamzaContainerContext;
+import org.apache.samza.operators.BaseTableDescriptor;
+import org.apache.samza.operators.TableDescriptor;
+import org.apache.samza.serializers.KVSerde;
+import org.apache.samza.serializers.NoOpSerde;
 import org.apache.samza.sql.interfaces.SourceResolver;
 import org.apache.samza.sql.interfaces.SourceResolverFactory;
 import org.apache.samza.sql.interfaces.SqlSystemSourceConfig;
+import org.apache.samza.table.ReadWriteTable;
+import org.apache.samza.table.Table;
+import org.apache.samza.table.TableProvider;
+import org.apache.samza.table.TableProviderFactory;
+import org.apache.samza.table.TableSpec;
+import org.apache.samza.task.TaskContext;
 
 
 public class TestSourceResolverFactory implements SourceResolverFactory {
+  public static final String TEST_DB_SYSTEM = "testDb";
+  public static final String TEST_TABLE_ID = "testDbId";
+
   @Override
   public SourceResolver create(Config config) {
     return new TestSourceResolver(config);
+  }
+
+  static class TestTableDescriptor extends BaseTableDescriptor {
+    protected TestTableDescriptor(String tableId) {
+      super(tableId);
+    }
+
+    @Override
+    public String getTableId() {
+      return tableId;
+    }
+
+    @Override
+    public TableSpec getTableSpec() {
+      return new TableSpec(tableId, KVSerde.of(new NoOpSerde(), new NoOpSerde()), TestTableProviderFactory.class.getName(), new HashMap<>());
+    }
+  }
+
+  public static class TestTable implements ReadWriteTable {
+    public static Map<Object, Object> records = new HashMap<>();
+    @Override
+    public Object get(Object key) {
+      throw new NotImplementedException();
+    }
+
+    @Override
+    public Map getAll(List keys) {
+      throw new NotImplementedException();
+    }
+
+    @Override
+    public void close() {
+    }
+
+    @Override
+    public void put(Object key, Object value) {
+      if (key == null) {
+        records.put(System.nanoTime(), value);
+      } else {
+        records.put(key, value);
+      }
+    }
+
+    @Override
+    public void delete(Object key) {
+      records.remove(key);
+    }
+
+    @Override
+    public void deleteAll(List keys) {
+      records.clear();
+    }
+
+    @Override
+    public void flush() {
+    }
+
+    @Override
+    public void putAll(List entries) {
+      throw new NotImplementedException();
+    }
+  }
+
+  public static class TestTableProviderFactory implements TableProviderFactory {
+    @Override
+    public TableProvider getTableProvider(TableSpec tableSpec) {
+      return new TestTableProvider();
+    }
+  }
+
+  static class TestTableProvider implements TableProvider {
+    @Override
+    public void init(SamzaContainerContext containerContext, TaskContext taskContext) {
+    }
+
+    @Override
+    public Table getTable() {
+      return new TestTable();
+    }
+
+    @Override
+    public Map<String, String> generateConfig(Map<String, String> config) {
+      return new HashMap<>();
+    }
+
+    @Override
+    public void close() {
+    }
   }
 
   private class TestSourceResolver implements SourceResolver {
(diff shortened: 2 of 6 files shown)


samza-sql/src/main/java/org/apache/samza/sql/interfaces/SourceResolver.java
```diff
@@ -41,4 +46,12 @@
...
    */
   boolean isTable(String sourceName);
+
+  /**
+   * Return a Samza table descriptor corresponding to the {@code sourceName} if the underlying source is a table,
+   * ie. {@link SourceResolver#isTable(String)} should return true for the same {@code sourceName}.
+   * @param sourceName source for which the table descriptor is requested.
+   * @return table descriptor; null if the source is not a table.
+   */
+  TableDescriptor getTableDescriptor(String sourceName);
```
srinipunuru: Can you merge getTableDescriptor and isTable with SqlSystemSourceConfig. Probably we could call SqlSystemSourceConfig as SamzaSqlIOConfig?

samza-sql/src/test/java/org/apache/samza/sql/testutil/TestSourceResolverFactory.java
```diff
@@ -52,15 +159,29 @@ public SqlSystemSourceConfig fetchSourceInfo(String sourceName) {
...
           Arrays.asList(sourceComponents), systemConfigs, isTable);
     }
 
     @Override
-    public boolean isTable(String sourceName) {
+    public boolean isTable(String sourceName){
       String[] sourceComponents = sourceName.split("\\.");
-      return sourceComponents[sourceComponents.length - 1].equalsIgnoreCase(SAMZA_SQL_QUERY_TABLE_KEYWORD);
+      return sourceComponents[sourceComponents.length - 1].equalsIgnoreCase(SAMZA_SQL_QUERY_TABLE_KEYWORD) ||
+          sourceName.equalsIgnoreCase("testDb.testTable");
```
atoomula: Why are we not using table keyword for output table ? If we do not use the table keyword, how do we plan to identify a table for a non-test source resolver ? IMO, we should not let test source resolver identify a table by <db>.<table>.

atoomula: Sorry, I did not think through this all the way when I commented earlier. Here is some context:
An input system in a given source resolver could be consumed either via a stream or a table. Hence, we added the table keyword to differentiate that. But for an output system, I think we made an explicit decision that an output system in a given source resolver would be only one of stream (SystemProducer) or table (TableProvider). So, what you are representing is correct. Could we explain the reasoning for this dichotomy between input and output systems in comments ? Of course, one could have their own source resolver to add table keyword even for output system.

Also, in this test resolver should we identify the table just by looking at the outputSystem component (could we rename testDb to testOutputSystem or some such ? testDb sounds like a db name) ?

pdu-mn1: Thanks for the context. Regarding the naming though, here we're testing the ability to insert into a table so naming it testDb and testTable are clearer than calling it outputSystem, right?

srinipunuru: @pdu-mn1 @atoomula Since we have table keyword in the name as a convention in the input. I think it might be good idea to support the same convention in the output as well. It is ok if we don't need it. But it might be good idea to support it as well. What do you folks think?

i.e. we will support both

1. insert into <testDb.Table> 
2. insert into table.<testDb.table>

Does it make sense?

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SourceResolver.java
srinipunuru: >To do this, we would have to move the local table code in join translator into all source resolvers. 
@pdu-mn1 @atoomula That makes sense. Can we do quick offline review to see how this code looks? If it is becoming too big of a change. I am ok with merging this PR and doing a fresh PR for that refactor.

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SourceResolver.java
```diff
@@ -19,26 +19,21 @@
...
-  /**
-   * Returns if a given source is a table. Different source resolvers could have different notations in the source
-   * name for denoting a table. Eg: system.stream.$table
-   * @param sourceName
-   *  source that needs to be checked if it is a table.
-   * @return
-   *  true if the source is a table, else false.
-   */
-  boolean isTable(String sourceName);
+  SqlSystemSourceConfig fetchSourceInfo(String sourceName, boolean isSink);
```
srinipunuru: Do you think it is better to have two methods here fetchSourceInfo and fetchSinkInfo?

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SourceResolver.java
```diff
@@ -19,26 +19,21 @@
...
 package org.apache.samza.sql.interfaces;
 
+
 /**
- * Source Resolvers are used by Samza Sql application to fetch the {@link SqlSystemSourceConfig} corresponding to the source.
+ * Source Resolvers are used by Samza Sql application to fetch the {@link SqlSystemSourceConfig} corresponding
+ * to the source. Additionally, it provides capability to resolve Samza table sources and creating the matching
+ * table descriptors.
  */
 public interface SourceResolver {
```
srinipunuru: I think we probably need to rename this to IOResolver right?

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlSystemSourceConfig.java
```diff
@@ -47,30 +48,34 @@
...
+  private final Config config;
 
-  private List<String> sourceParts;
+  private final List<String> sourceParts;
+
+  private final boolean isTable;
+
+  private final TableDescriptor tableDescriptor;
 
   public SqlSystemSourceConfig(String systemName, String streamName, Config systemConfig) {
```
srinipunuru: I think we probably might need to rename this as SQLIOConfig

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlSystemSourceConfig.java
```diff
@@ -126,4 +131,12 @@ public Config getConfig() {
   public String getSource() {
     return source;
   }
+
+  public boolean isTable() {
```
srinipunuru: Do we need a separate isTable API here or can we use getTableDescriptor to return Optional.empty() if it is not a table?

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SourceResolver.java
pdu-mn1: I think that makes sense.

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SourceResolver.java
pdu-mn1: Was thinking renaming could touch too many files. But given this PR is adding the write-path to SQL so it makes sense to do it in one go. I'll do the renaming.

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlSystemSourceConfig.java
pdu-mn1: will do

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlSystemSourceConfig.java
pdu-mn1: I can get rid of isTable

samza-sql/src/main/java/org/apache/samza/sql/impl/SqlTableJoinUtils.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.samza.sql.data.SamzaSqlCompositeKey;
+import org.apache.samza.sql.data.SamzaSqlRelMessage;
+import org.apache.samza.storage.kv.RocksDbTableDescriptor;
+
+
+/**
+ * Helper class for Samza SQL table join with RocksDb local table implementation. This can be used
+ * by Samza SQL IO resolvers to resolve the IO config for an input table in the SQL statement.
+ */
+public class SqlTableJoinUtils {
```
vjagadish: Looks like a cache of TableDescriptor objects - Are these objects expensive to create? why do we require this class?

samza-sql/src/main/java/org/apache/samza/sql/impl/SqlTableJoinUtils.java
```diff
@@ -0,0 +1,65 @@
...
+          .withSerde(KVSerde.of(keySerde, valueSerde));
+
+      tableDescMap.put(sourceName, tableDesc);
+    }
+
+    return tableDesc;
+  }
+
+  public static Serde<SamzaSqlCompositeKey> getKeySerde() {
+    return new JsonSerdeV2<>(SamzaSqlCompositeKey.class);
```
vjagadish: Would prefer instantiations of JsonSerdeV2 directly at the call-site

samza-sql/src/main/java/org/apache/samza/sql/impl/SqlTableJoinUtils.java
pdu-mn1: This is not about caching. So the issue is that this part of the code used to be in JoinTranslator directly, which means all Samza SQL table joins are assumed to be local table only. In reality, there is no reason we can't support join with remote tables. In samza SQL, IO-related information is all stored in  SqlIOConfig (used to be SqlSourceConfig) hence we decided to store TableDescriptor there too. SqlIOConfig is fully managed by SqlIOResolver, so it does not make sense to have JoinTranslator to create its own RocksDb table descriptor anymore. The downside with this is that all IOResolvers need to be able to create them for table joins. We currently have three IOResolvers in Samza SQL, Config-based, Test, and SamzaLi IOResolvers. So without duplicating the code in all three places, I created this helper class so that all three resolvers can reuse the code. 

For Samza Table, we should always return the same descriptor for the same tableId so there is a cache in this helper. 

Hope this explains. 

@vjagadish

samza-sql/src/main/java/org/apache/samza/sql/impl/SqlTableJoinUtils.java
pdu-mn1: Sure.

samza-sql/src/main/java/org/apache/samza/sql/impl/SqlTableJoinUtils.java
vjagadish: Peng, 
the only method in this class is 
```
new RocksDbTableDescriptor("JoinTable-" + sourceName)
+          .withSerde(KVSerde.of(keySerde, valueSerde));
```
If there are no caching requirements for table-descriptors, can we not invoke the constructor for `new RocksDbTableDescriptor` at the call-site?

vjagadish: Let me chat with you offline on this.

vjagadish:
@pdu-mn1 please create a JIRA for this PR and update the title

vjagadish1989 (approved):
approved, looks good to me.

samza-sql/src/main/java/org/apache/samza/sql/impl/SqlTableJoinUtils.java
pdu-mn1: I just tried not caching the descriptor and it works so you're right that it's just one liner of constructor call without the caching. I've dropped the helper class and added the one liner in the IO resolvers directly. The descriptor is a light weight object so it is okay to not cache it.

PR #466, opened 2018-04-06
SAMZA-1640: JobModel Json deserialization error in ZkJobCoordinator.

shanthoosh:


shanthoosh:
@xinyuiscool @dnishimura 

shanthoosh:
Serialized JobModel after the change: 

```json
{
  "config" : {
  },
  "containers" : {
    "0000000000" : {
      "tasks" : {
        "SystemStreamPartition [TestSystemName, test-input-single-partition-topic-8d1667fc-23f6-42a6-8a6b-673d131ee8dc, 0]" : {
          "task-name" : "SystemStreamPartition [TestSystemName, test-input-single-partition-topic-8d1667fc-23f6-42a6-8a6b-673d131ee8dc, 0]",
          "system-stream-partitions" : [ {
            "system" : "TestSystemName",
            "partition" : 0,
            "stream" : "test-input-single-partition-topic-8d1667fc-23f6-42a6-8a6b-673d131ee8dc"
          } ],
          "changelog-partition" : 0
        }
      },
      "container-id" : -1,
      "processor-id" : "0000000000"
    }
  },
  "max-change-log-stream-partitions" : 1,
  "all-container-locality" : {
    "0000000000" : null
  }
}
```

samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
```diff
@@ -143,6 +144,7 @@ private void populateContainerLocalityMappings() {
     return containers;
   }
 
+  @JsonIgnore
```
prateekm: How does adding this fix existing versions? If it doesn't, why ignore it instead of serde-ing it correctly?

shanthoosh: Discussed offline. 

Currently public fields(maxChangeLogStreamPartitions) and public getter methods (getTaskProcessorLocality, getAllContainerLocality) in JobModel is included in the serialization. Exception was due to that.

Agreed short term fix is to inline the getTaskPartitionMappings method invocation (rather than using JsonIgnore). 

From the JsonJobModelMixin class, it looks like we wanted to serialize only jobModel.config and  jobModel.containers. Will create a JIRA for the long term fix(We can follow similar pattern used in SSP, TaskName json serde and choose a subset of fields to serialize/deserialize).

dnishimura: Just to clarify, in addition to getTaskPartitionMappings, getTaskProcessorLocality, getAllContainerLocality and maxChangeLogStreamPartitions are causing the serialization exception. Is that a correct statement?

shanthoosh: I dont think so. getAllContainerLocality and maxChangeLogStreamPartitions  were present in JobModel from really earlier samza versions(>=0.7). Hence they were not causing problem.

dnishimura:
👍 Thanks for looking into this. The unit test you fixed should hopefully catch future issues like this. Maybe not in this PR, but it'd be good to document the classes that are serializable. It wasn't obvious here because JobModel is a plain Java object. Good work!

PR #467, opened 2018-04-10
Upgrade to latest event hub version (1.0.1)

srinipunuru:
* Upgrade to the latest event hub version 1.0.1
* Adding configs for prefetchCount and maxEventPerPoll
* Fix the high cpu usage issue in SamzaHistogram
* Fixing a race condition in event hub system producer where the future was getting removed while it was being checked for completion resulting in NPE.

### DIFF
--- a/samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubClientManagerFactory.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubClientManagerFactory.java
@@ -26,7 +26,8 @@ public class EventHubClientManagerFactory {
     String entityPath = config.getStreamEntityPath(systemName, streamName);
     String sasKeyName = config.getStreamSasKeyName(systemName, streamName);
     String sasToken = config.getStreamSasToken(systemName, streamName);
+    int numClientThreads = config.getNumClientThreads(systemName);
 
-    return new SamzaEventHubClientManager(eventHubNamespace, entityPath, sasKeyName, sasToken, config.getNumClientThreads(systemName));
+    return new SamzaEventHubClientManager(eventHubNamespace, entityPath, sasKeyName, sasToken, numClientThreads);
   }
 }
--- a/samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubClientManagerFactory.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubClientManagerFactory.java
@@ -27,6 +27,6 @@ public class EventHubClientManagerFactory {
     String sasKeyName = config.getStreamSasKeyName(systemName, streamName);
     String sasToken = config.getStreamSasToken(systemName, streamName);
 
-    return new SamzaEventHubClientManager(eventHubNamespace, entityPath, sasKeyName, sasToken);
+    return new SamzaEventHubClientManager(eventHubNamespace, entityPath, sasKeyName, sasToken, config.getNumClientThreads(systemName));
   }
 }
--- a/samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubConfig.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubConfig.java
@@ -154,7 +154,8 @@ public class EventHubConfig extends MapConfig {
   }
 
   /**
-   * Get the number of client threads
+   * Get the number of client threads, This is used to create the ThreadPool executor that is passed to the
+   * {@link EventHubClient#create}
    * @param systemName Name of the system.
    * @return Num of client threads to use.
    */
--- a/samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubConfig.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubConfig.java
@@ -20,6 +20,7 @@
 package org.apache.samza.system.eventhub;
 
 import com.microsoft.azure.eventhubs.EventHubClient;
+import com.microsoft.azure.eventhubs.PartitionReceiver;
 import org.apache.samza.SamzaException;
 import org.apache.samza.config.Config;
 import org.apache.samza.config.MapConfig;
@@ -48,6 +49,15 @@ public class EventHubConfig extends MapConfig {
 
   public static final String CONFIG_STREAM_CONSUMER_GROUP = "streams.%s.eventhubs.consumer.group";
   public static final String DEFAULT_CONFIG_STREAM_CONSUMER_GROUP = EventHubClient.DEFAULT_CONSUMER_GROUP_NAME;
+
+  public static final String CONFIG_SYSTEM_NUM_CLIENT_THREADS = "streams.%s.eventhubs.numClientThreads";
+  public static final int DEFAULT_CONFIG_SYSTEM_NUM_CLIENT_THREADS = 10;
+
+  public static final String CONFIG_PREFETCH_COUNT = "systems.%s.eventhubs.prefetchCount";
+  public static final int DEFAULT_CONFIG_PREFETCH_COUNT = PartitionReceiver.DEFAULT_PREFETCH_COUNT;
+
+  public static final String CONFIG_MAX_EVENT_COUNT_PER_POLL = "systems.%s.eventhubs.maxEventCountPerPoll";
+  public static final int DEFAULT_CONFIG_MAX_EVENT_COUNT_PER_POLL = 50;
 
   public static final String CONFIG_PRODUCER_PARTITION_METHOD = "systems.%s.eventhubs.partition.method";
   public static final String DEFAULT_CONFIG_PRODUCER_PARTITION_METHOD = EventHubSystemProducer
@@ -141,6 +151,33 @@ public class EventHubConfig extends MapConfig {
   public String getStreamEntityPath(String systemName, String streamName) {
     return validateRequiredConfig(getFromStreamIdOrName(CONFIG_STREAM_ENTITYPATH, streamName),
             "EntityPath", systemName, streamName);
+  }
+
+  /**
+   * Get the number of client threads
+   * @param systemName Name of the system.
+   * @return Num of client threads to use.
+   */
+  public Integer getNumClientThreads(String systemName) {
+    return getInt(String.format(CONFIG_SYSTEM_NUM_CLIENT_THREADS, systemName), DEFAULT_CONFIG_SYSTEM_NUM_CLIENT_THREADS);
+  }
+
+  /**
+   * Get the max event count returned per poll
+   * @param systemName Name of the system
+   * @return Max number of events returned per poll
+   */
+  public Integer getMaxEventCountPerPoll(String systemName) {
+    return getInt(String.format(CONFIG_MAX_EVENT_COUNT_PER_POLL, systemName), DEFAULT_CONFIG_MAX_EVENT_COUNT_PER_POLL);
+  }
+
+  /**
+   * Get the per partition prefetch count for the event hub client
+   * @param systemName Name of the system.
+   * @return Per partition Prefetch count for the event hub client.
+   */
+  public Integer getPrefetchCount(String systemName) {
+    return getInt(String.format(CONFIG_PREFETCH_COUNT, systemName), DEFAULT_CONFIG_PREFETCH_COUNT);
   }
 
   /**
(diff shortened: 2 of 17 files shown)


samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubClientManagerFactory.java
```diff
@@ -27,6 +27,6 @@ public EventHubClientManager getEventHubClientManager(String systemName, String
     String sasKeyName = config.getStreamSasKeyName(systemName, streamName);
     String sasToken = config.getStreamSasToken(systemName, streamName);
 
-    return new SamzaEventHubClientManager(eventHubNamespace, entityPath, sasKeyName, sasToken);
+    return new SamzaEventHubClientManager(eventHubNamespace, entityPath, sasKeyName, sasToken, config.getNumClientThreads(systemName));
```
prateekm: Minor: extract field for consistency.

samza-azure/src/main/java/org/apache/samza/system/eventhub/SamzaEventHubClientManager.java
```diff
@@ -52,29 +54,36 @@
...
+    this.numClientThreads = numClientThreads;
   }
 
   @Override
   public void init() {
     String remoteHost = String.format(EVENTHUB_REMOTE_HOST_FORMAT, eventHubNamespace);
     try {
       ConnectionStringBuilder connectionStringBuilder =
-              new ConnectionStringBuilder(eventHubNamespace, entityPath, sasKeyName, sasKey);
+              new ConnectionStringBuilder()
```
prateekm: Minor: Fix indentation.

samza-azure/src/main/java/org/apache/samza/system/eventhub/SamzaEventHubClientManager.java
```diff
@@ -52,29 +54,36 @@
...
+          .setNamespaceName(eventHubNamespace)
+          .setEventHubName(entityPath)
+          .setSasKeyName(sasKeyName)
+          .setSasKey(sasKey);
 
-      eventHubClient = EventHubClient.createFromConnectionStringSync(connectionStringBuilder.toString(), retryPolicy);
-    } catch (IOException | ServiceBusException e) {
+      eventHubClient = EventHubClient.createSync(connectionStringBuilder.toString(), retryPolicy,
+          Executors.newFixedThreadPool(numClientThreads));
```
prateekm: Does this need to be shut down when this client is stopped? Do the threads need to be daemon? Should also name the threads in this pool. Convention: "Samza EventHubClient Thread-%d"

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -183,7 +187,9 @@ public void register(SystemStreamPartition systemStreamPartition, String offset)
 
     if (streamPartitionOffsets.containsKey(systemStreamPartition)) {
       // Only update if new offset is lower than previous offset
-      if (END_OF_STREAM.equals(offset)) return;
+      if (END_OF_STREAM.equals(offset)) {
```
prateekm: Thanks for the formatting cleanup.

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -348,24 +356,30 @@ private boolean isErrorTransient(Throwable throwable) {
...
-                                 SamzaHistogram readLatency, Counter readErrors, Interceptor interceptor) {
-      super(MAX_EVENT_COUNT_PER_PARTITION_POLL);
+        SamzaHistogram readLatency, Counter readErrors, Interceptor interceptor) {
       this.ssp = ssp;
       this.eventReadRate = eventReadRate;
       this.eventByteReadRate = eventByteReadRate;
       this.readLatency = readLatency;
       this.errorRate = readErrors;
       this.interceptor = interceptor;
+      this.maxEventCount = config.getMaxEventCountPerPoll(systemName);
```
prateekm: Minor: Prefer passing as an explicit constructor param. For consistency, as well as avoiding access to parent class state.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/AsyncSystemProducer.java
```diff
@@ -139,7 +139,6 @@ public synchronized void send(String source, OutgoingMessageEnvelope envelope) {
 
     // Auto update the metrics and possible throwable when futures are complete.
     sendResult.handle((aVoid, throwable) -> {
-        pendingFutures.remove(sendResult);
```
prateekm: This isn't necessary anymore?

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubConfig.java
```diff
@@ -143,6 +153,33 @@ public String getStreamEntityPath(String systemName, String streamName) {
             "EntityPath", systemName, streamName);
   }
 
+  /**
+   * Get the number of client threads
```
prateekm: Doc should explain what this is for.

samza-azure/src/main/java/org/apache/samza/system/eventhub/SamzaEventHubClientManager.java
```diff
@@ -52,29 +54,36 @@
   private final String sasKey;
   private final RetryPolicy retryPolicy;
 
-  public SamzaEventHubClientManager(String eventHubNamespace, String entityPath, String sasKeyName, String sasKey) {
+  public SamzaEventHubClientManager(String eventHubNamespace, String entityPath, String sasKeyName, String sasKey,
+      Integer numClientThreads) {
     this(eventHubNamespace, entityPath, sasKeyName, sasKey,
-            new RetryExponential(MIN_RETRY_BACKOFF, MAX_RETRY_BACKOFF, MAX_RETRY_COUNT, SAMZA_EVENTHUB_RETRY));
+            new RetryExponential(MIN_RETRY_BACKOFF, MAX_RETRY_BACKOFF, MAX_RETRY_COUNT, SAMZA_EVENTHUB_RETRY), numClientThreads);
```
prateekm: Minor: Maybe document what the client threads are used for?

samza-azure/src/main/java/org/apache/samza/system/eventhub/SamzaEventHubClientManager.java
```diff
@@ -52,29 +54,36 @@
   private final String sasKey;
   private final RetryPolicy retryPolicy;
 
-  public SamzaEventHubClientManager(String eventHubNamespace, String entityPath, String sasKeyName, String sasKey) {
+  public SamzaEventHubClientManager(String eventHubNamespace, String entityPath, String sasKeyName, String sasKey,
+      Integer numClientThreads) {
```
prateekm: Minor: Inconsistent indentation for constructor params.

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/AsyncSystemProducer.java
srinipunuru: This was not necessary event before because we clear up the pendingFutures on flush. But long term we need to fix this properly where we remove the futures on callback and flush doesn't need to clear. This is to support brooklin pipelines that don't call flush.

samza-azure/src/main/java/org/apache/samza/system/eventhub/SamzaEventHubClientManager.java
srinipunuru: Added the documentation in the header.

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubConfig.java
```diff
@@ -143,6 +153,34 @@ public String getStreamEntityPath(String systemName, String streamName) {
             "EntityPath", systemName, streamName);
   }
 
+  /**
+   * Get the number of client threads, This is used to create the ThreadPool executor that is passed to the
```
prateekm: I meant that this should explain what this is used _for_ in the client. I.e., why would I want to override this?
Also, we should update the configuration reference with these new values.

samza-azure/src/main/java/org/apache/samza/system/eventhub/metrics/SamzaHistogram.java
```diff
@@ -1,62 +1,81 @@
...
     this.histogram = new Histogram(new ExponentiallyDecayingReservoir());
     this.percentiles = percentiles;
     this.gauges = this.percentiles.stream()
-            .filter(x -> x > 0 && x <= 100)
-            .collect(
-                    Collectors.toMap(Function.identity(), x -> registry.newGauge(group, name + "_" + String.valueOf(0), 0D)));
+        .filter(x -> x > 0 && x <= 100)
+        .collect(
+            Collectors.toMap(Function.identity(), x -> registry.newGauge(group, name + "_" + String.valueOf(0), 0D)));
+    executorService.schedule(this, UPDATE_FREQUENCY_MS, TimeUnit.MILLISECONDS);
```
prateekm: Why switch to an async update? Is this taking too long in the sync path?

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/AsyncSystemProducer.java
```diff
@@ -113,6 +117,8 @@ public AsyncSystemProducer(String systemName, Config config, MetricsRegistry met
     physicalToStreamIds =
         streamIds.stream().collect(Collectors.toMap(sconfig::getPhysicalName, Function.identity()));
     this.metricsRegistry = metricsRegistry;
+    ThreadFactoryBuilder threadFactoryBuilder = new ThreadFactoryBuilder().setNameFormat("SamzaHistogram_%d");
```
prateekm: "Samza Histogram Poll Thread-%d".
Should be daemon.

samza-azure/src/main/java/org/apache/samza/system/eventhub/metrics/SamzaHistogram.java
srinipunuru: Previously the calculation was happening for every update which was pretty expensive, We are now doing it only at regular intervals (1 minute) which is what we need.

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubConfig.java
srinipunuru: This is event hubs configuration. Update the configuration reference which points to the event hubs documentation.

samza-azure/src/main/java/org/apache/samza/system/eventhub/metrics/SamzaHistogram.java
lhaiesp: This is wrong. I think you meant to use scheduelAtFixedRate

samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/AsyncSystemProducer.java
```diff
@@ -113,6 +117,8 @@ public AsyncSystemProducer(String systemName, Config config, MetricsRegistry met
...
+    executorService = Executors.newScheduledThreadPool(1, threadFactoryBuilder.build());
```
lhaiesp: This is equivalent to newSingleThreadScheduledExecutor

samza-azure/src/main/java/org/apache/samza/system/eventhub/metrics/SamzaHistogram.java
prateekm: If getSnapshot() isn't expensive, you can also register a custom Gauge (using MetricsRegistry#newGauge(String, Gauge)) that overrides its getValue to do this calculation when polled. This way the calculation will be on the MetricsReporter thread.

srinipunuru: Good catch!

srinipunuru: I considered this, if you see there can be multiple gauges per Histogram, I thought it is wasteful to do multiple updates for the same Histogram, But i do like that it is simple. Do you know if the visit calls are all in sequence? Let's chat about this.

PR #468, opened 2018-04-11
SAMZA-1643: StreamPartitionCountMonitor should only restart/shut down the job if partition count increases

prateekm:
As an aside, also update the gauge to report current number of partitions instead of the change, since that's what its name indicates.


prateekm:
@vjagadish1989 @nickpan47

vjagadish:
merged and submitted. thanks Prateek!

PR #471, opened 2018-04-12
SAMZA-1649: Improve host-aware allocation to account for strict locality

vjagadish:
Working on a doc for the behavior of the CapacityScheduler and further testing is needed on an actual cluster - but here's a summary: 
 - Node-local requests are honored only when relax-locality = false
 - With relax-locality = true, the scheduler biases interactivity over data-locality for requests that ask for few resources relative to the size of the cluster. 

In addition to the above change, this PR also modifies the allocator algorithm to fallback to "ANY_HOST" requests so that we make progress when the node is unavailable.


samza-core/src/main/java/org/apache/samza/clustermanager/ResourceRequestState.java
```diff
@@ -69,6 +69,18 @@ public ResourceRequestState(boolean hostAffinityEnabled, ClusterResourceManager
     this.manager = manager;
   }
 
+  public void cancelResourceRequest(SamzaResourceRequest request) {
```
prateekm: Minor: Move lower (to after updateStateAfterAssignment) and maybe add documentation.

samza-core/src/main/java/org/apache/samza/clustermanager/ResourceRequestState.java
```diff
@@ -69,6 +69,18 @@ public ResourceRequestState(boolean hostAffinityEnabled, ClusterResourceManager
...
+    synchronized (lock) {
```
prateekm: Log a message describing what is being cancelled.

PR #472, opened 2018-04-12
SAMZA-1650: Fix for firing trigger at the end of trigger interval for tumbling window

atoomula:


vjagadish:
merged and submitted, thanks Aditya!

PR #475, opened 2018-04-16
SAMZA-1659: Serializable OperatorSpec

nickpan47:
This change is to make the user supplied functions serializable. Hence, making the full user defined DAG serializable.


nickpan47:
@xinyuiscool @prateekm and @srinipunuru can you help to review this patch? Thanks!

prateekm:
@nickpan47 Thanks for this. Can you add some description of the major code and design changes in this PR?

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -52,8 +52,11 @@
   private final StreamManager streamManager;
   private final SystemAdmins systemAdmins;
 
+  protected final StreamGraphImpl streamGraph;
+
   public AbstractApplicationRunner(Config config) {
     super(config);
+    this.streamGraph = new StreamGraphImpl(this, config);
```
vjagadish: I'd love to see some docs on the scope of the AppRunner - currently, appRunner.run(app) allows for multiple applications to be run with the same instance of the runner (atleast the api does not disallow it). 

With this change of instantiating the `StreamGraph` in the constructor, are we restricting the scope to running a single application? If not, how do we distinguish `StreamGraph`s of 2 different apps on the same runner

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -253,15 +254,15 @@ public void waitForFinish() {
...
    * @return {@link StreamProcessor]}
    */
   /* package private */
   StreamProcessor createStreamProcessor(
       Config config,
-      StreamApplication app,
+      StreamGraphImpl streamGraph,
       StreamProcessorLifecycleListener listener) {
-    Object taskFactory = TaskFactoryUtil.createTaskFactory(config, app, new LocalApplicationRunner(config));
+    Object taskFactory = TaskFactoryUtil.createTaskFactory(config, streamGraph);
```
vjagadish: I wonder if we can initialize the `StreamApplication` directly in the runner instead of initializing it inside an Util method? It's not obvious to observe that `app#init` is being invoked.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -39,60 +37,58 @@
...
-    this.streamApplication = streamApplication;
-    this.runner = runner;
+  public StreamOperatorTask(SerializedStreamGraph serializedGraph, Clock clock) {
+    this.serializedGraph = serializedGraph;
     this.clock = clock;
   }
 
-  public StreamOperatorTask(StreamApplication application, ApplicationRunner runner) {
-    this(application, runner, SystemClock.instance());
+  public StreamOperatorTask(SerializedStreamGraph streamGraph) {
```
vjagadish: Love this!! This change is really the core of this PR 👍 Do you mind adding a brief description on the JIRA/PR comment? That way, other reviewers have a bit more context.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -39,60 +37,58 @@
...
+    this(streamGraph, SystemClock.instance());
   }
 
   /**
    * Initializes this task during startup.
    * <p>
-   * Implementation: Initializes the user-implemented {@link StreamApplication}. The {@link StreamApplication} sets
-   * the input and output streams and the task-wide context manager using the {@link StreamGraphImpl} APIs,
+   * Implementation: Initializes the user-implemented {@link org.apache.samza.operators.StreamGraphImpl}.
```
vjagadish: Docs don't seem accurate - Are we actually initializing the `StreamGraphImpl` here?

samza-api/src/main/java/org/apache/samza/operators/functions/SupplierFunction.java
```diff
@@ -0,0 +1,38 @@
...
+ * under the License.
+ */
+package org.apache.samza.operators.functions;
+
+import java.io.Serializable;
+import org.apache.samza.annotation.InterfaceStability;
+
+
+/**
+ * A forced function interface as {@link java.util.function.Supplier}, but added mandatory interface methods in Samza
```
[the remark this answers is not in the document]
prateekm: +1

samza-api/src/main/java/org/apache/samza/operators/functions/ClosableFunction.java
```diff
@@ -33,5 +33,8 @@
  */
 @InterfaceStability.Unstable
 public interface ClosableFunction {
+  /**
+   * Method to free any resource initialized/acquired by the operators in {@link InitableFunction}
```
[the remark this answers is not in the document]
prateekm: +1

samza-api/src/main/java/org/apache/samza/operators/functions/SinkFunction.java
```diff
@@ -30,7 +31,7 @@
  */
 @InterfaceStability.Unstable
 @FunctionalInterface
-public interface SinkFunction<M>  extends InitableFunction, ClosableFunction {
+public interface SinkFunction<M>  extends InitableFunction, ClosableFunction, Serializable {
```
prateekm: We should document the serializability requirement for all these functions more explicitly.

samza-api/src/main/java/org/apache/samza/operators/functions/SupplierFunction.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+@InterfaceStability.Unstable
+@FunctionalInterface
+public interface SupplierFunction<T> extends InitableFunction, ClosableFunction, Serializable {
+
+  /**
+   * Method to offer the initial value of type T
```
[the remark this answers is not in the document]
prateekm: +1

samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
```diff
@@ -125,16 +125,23 @@ private Windows() { }
...
-      Supplier<? extends WV> initialValue, FoldLeftFunction<? super M, WV> aggregator, Serde<K> keySerde,
+  public static <M, K, WV> Window<M, K, WV> keyedTumblingWindow(MapFunction<? super M, ? extends K> keyFn, Duration interval,
+      SupplierFunction<? extends WV> initialValue, FoldLeftFunction<? super M, WV> aggregator, Serde<K> keySerde,
       Serde<WV> windowValueSerde) {
 
     Trigger<M> defaultTrigger = new TimeTrigger<>(interval);
-    return new WindowInternal<>(defaultTrigger, (Supplier<WV>) initialValue, (FoldLeftFunction<M, WV>) aggregator,
-        (Function<M, K>) keyFn, null, WindowType.TUMBLING, keySerde, windowValueSerde, null);
+    return new WindowInternal<>(
+        defaultTrigger,
```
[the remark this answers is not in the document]
prateekm: +1 for keeping them on the same line everywhere.

samza-api/src/main/java/org/apache/samza/operators/windows/internal/WindowInternal.java
```diff
@@ -55,28 +55,32 @@
...
-
   private Trigger<M> earlyTrigger;
   private Trigger<M> lateTrigger;
   private AccumulationMode mode;
 
-  public WindowInternal(Trigger<M> defaultTrigger, Supplier<WV> initializer, FoldLeftFunction<M, WV> foldLeftFunction,
-      Function<M, WK> keyExtractor, Function<M, Long> eventTimeExtractor, WindowType windowType, Serde<WK> keySerde,
+  /**
+   * The following serdes should only be used to generate configs for store config.
+   * No need to create per-task copy of those {@link Serde} objects
```
prateekm: Maybe (everywhere):
"The following Serde fields are serialized by the ExecutionPlanner when generating the store configs, and deserialized once during startup in SamzaContainer. They don't need to be deserialized here on a per-task basis."

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -71,57 +70,52 @@
   /**
    * The {@link OperatorSpec} associated with this {@link MessageStreamImpl}
    */
-  private final OperatorSpec operatorSpec;
+  private final OperatorSpec source;
```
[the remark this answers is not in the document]
prateekm: +1 for keeping as is. I also personally prefer moving the getOpId call to its own line since that has side effects.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
```diff
@@ -77,24 +79,28 @@
...
+   * A mapping from {@link JoinOperatorSpec} IDs to the single copy of {@link JoinOperatorSpec} per each task, and their
+   * two {@link PartialJoinFunction}s.
+   * Used to associate the two {@link PartialJoinOperatorImpl}s for a {@link JoinOperatorSpec} with each other since they're
    * reached from different {@link OperatorSpec} during DAG traversals.
    */
-  private final Map<String, KV<PartialJoinFunction, PartialJoinFunction>> joinFunctions = new HashMap<>();
+  private final Map<String, Tuple3<JoinOperatorSpec, PartialJoinFunction, PartialJoinFunction>> joinSpecs = new HashMap<>();
 
+  /**
+   * Method to get the current time.
```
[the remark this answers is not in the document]
prateekm: Prefer just removing this. Functionality is clear from the class/field name.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
```diff
@@ -77,24 +79,28 @@
...
    * in the {@code streamGraph}.
    *
-   * @param streamGraph  the {@link StreamGraphImpl} containing the logical {@link OperatorSpec} DAG
+   * @param streamGraph  the {@link SerializedStreamGraph} containing the logical {@link OperatorSpec} DAG
    * @param config  the {@link Config} required to instantiate operators
    * @param context  the {@link TaskContext} required to instantiate operators
    * @param clock  the {@link Clock} to get current time
    */
-  public OperatorImplGraph(StreamGraphImpl streamGraph, Config config, TaskContext context, Clock clock) {
+  public OperatorImplGraph(final SerializedStreamGraph streamGraph, final Config config, final TaskContext context, final Clock clock) {
```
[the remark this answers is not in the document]
prateekm: +1, SerializedStreamGraph feels odd since it's not actually serialized anymore, and the name seems to describe the impl than the purpose.

Another alternative for naming/structuring this might be to have a StreamGraphSpec and StreamGraphImpl similar to the OperatorSpec and OperatorImpl duality. I.e., the existing StreamGraphImpl would then become the StreamGraphSpec that's serialized, and the deserialized version used here would be StreamGraphImpl.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
```diff
@@ -204,51 +224,63 @@ OperatorImpl createAndRegisterOperatorImpl(OperatorSpec prevOperatorSpec, Operat
...
    * @return  the {@link OperatorImpl} implementation instance
    */
-  OperatorImpl createOperatorImpl(OperatorSpec prevOperatorSpec, OperatorSpec operatorSpec,
-      Config config, TaskContext context) {
+  OperatorImpl createOperatorImpl(final OperatorSpec prevOperatorSpec, final OperatorSpec operatorSpec,
+      final Config config, final TaskContext context, final SerializedStreamGraph streamGraph)
+      throws IOException, ClassNotFoundException {
     if (operatorSpec instanceof InputOperatorSpec) {
-      return new InputOperatorImpl((InputOperatorSpec) operatorSpec);
+      return new InputOperatorImpl((InputOperatorSpec) streamGraph.getOpSpec(operatorSpec.getOpId()));
```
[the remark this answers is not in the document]
prateekm: +1, it's not clear what the preOperatorSpec and operatorSpec params are for. Maybe just do the getOpSpec call at the call site and pass that in directly?

samza-core/src/main/java/org/apache/samza/operators/impl/SerializedStreamGraph.java
```diff
@@ -0,0 +1,77 @@
...
+  private final StreamGraphImpl originalGraph;
+
+  public SerializedStreamGraph(StreamGraphImpl streamGraph) {
+    this.originalGraph = streamGraph;
+    streamGraph.getAllOperatorSpecs().stream().forEach(opSpec -> this.operatorSpecMap.put(opSpec.getOpId(), opSpec));
+  }
+
+  public OperatorSpec getOpSpec(String opId) throws IOException, ClassNotFoundException {
+
+    return OperatorSpec.fromByte(this.serializedOpSpecs.computeIfAbsent(opId, operatorId -> {
```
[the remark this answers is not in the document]
prateekm: 1. Should be `fromBytes` and `toBytes` (plural)
2. Would be cleaner to have a separate Serde for OperatorSpec instead of static methods in the class itself.
3. Maybe include version info in the serialized representation.

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpec.java
```diff
@@ -62,8 +69,17 @@
    * The set of operators that consume the messages produced from this operator.
    * <p>
    * We use a LinkedHashSet since we need deterministic ordering in initializing/closing operators.
+   *
+   * Note: this is also made transient since the deserialized copies of {@code nextOperatorSpecs} are made externally when
```
prateekm: Comment is somewhat unclear. Where is the DAG _structure_ serialized? Are you saying that this field is set in OperatorImplGraph?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
prateekm: +1

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -39,60 +37,58 @@
...
   public final void init(Config config, TaskContext context) throws Exception {
-    StreamGraphImpl streamGraph = new StreamGraphImpl(runner, config);
-    // initialize the user-implemented stream application.
-    this.streamApplication.init(streamGraph, config);
 
     // get the user-implemented context manager and initialize it
-    this.contextManager = streamGraph.getContextManager();
-    if (this.contextManager != null) {
-      this.contextManager.init(config, context);
+    if (serializedGraph.getContextManager() != null) {
```
prateekm: IIUC this is the same as before?

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpec.java
nickpan47: Updated the doc.

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpec.java
```diff
@@ -79,10 +95,30 @@ public void registerNextOperatorSpec(OperatorSpec<OM, ?> nextOperatorSpec) {
...
+  /**
+   * Get the collection of chained {@link OperatorSpec}s that are consuming the output of this node
+   *
+   * @return the collection of chained {@link OperatorSpec}s
+   */
   public Collection<OperatorSpec<OM, ?>> getRegisteredOperatorSpecs() {
     return nextOperatorSpecs;
   }
 
+  public static byte[] toByte(OperatorSpec opSpec) throws IOException {
```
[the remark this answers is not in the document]
nickpan47: Taking the suggestion to create a OperatorSpecSerde class.

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpecs.java
```diff
@@ -73,29 +69,7 @@ private OperatorSpecs() {}
...
-      public void init(Config config, TaskContext context) {
-        mapFn.init(config, context);
-      }
-
-      @Override
-      public void close() {
-        mapFn.close();
-      }
-    }, mapFn, OperatorSpec.OpCode.MAP, opId);
+    return StreamOperatorSpec.<M, OM>createStreamOperatorSpec((MapFunction<M, OM>) mapFn, OperatorSpec.OpCode.MAP, opId);
```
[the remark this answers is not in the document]
nickpan47: Agreed. Working on the change.

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpecs.java
```diff
@@ -236,13 +188,12 @@ public void close() {
    * @return  the {@link StreamOperatorSpec} for the merge
    */
   public static <M> StreamOperatorSpec<M, M> createMergeOperatorSpec(String opId) {
-    return new StreamOperatorSpec<>(message ->
+    return StreamOperatorSpec.<M, M>createStreamOperatorSpec((M message) ->
```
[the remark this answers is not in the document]
nickpan47: Unfortunately, this can't be a separate PR since it is part of the merging w/ Xinyu's change of TimerFunction and WatermarkFunction.

samza-core/src/main/java/org/apache/samza/operators/spec/StreamOperatorSpec.java
```diff
@@ -32,20 +39,74 @@
...
-   * @param originalFn the original user function before wrapping to transformFn
    * @param opCode  the {@link OpCode} for this {@link StreamOperatorSpec}
    * @param opId  the unique ID for this {@link StreamOperatorSpec}
    */
-  StreamOperatorSpec(FlatMapFunction<M, OM> transformFn, Object originalFn, OperatorSpec.OpCode opCode, String opId) {
+  private StreamOperatorSpec(FlatMapFunction<M, OM> transformFn, OperatorSpec.OpCode opCode, String opId) {
     super(opCode, opId);
     this.transformFn = transformFn;
-    this.originalFn = originalFn;
+    this.userFn = transformFn;
```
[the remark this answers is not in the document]
nickpan47: Part of merging effort from Xinyu's patch. Will redo it as @prateekm suggest to have different OperatorSpec class for each flatMap/filter/map function now.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
nickpan47: @vjagadish agreed. Documentation will be added. However, I never thought that a single instance of AppRunner can be used for multiple different applications. The current implementation of LocalApplicationRunner actually has a bad combination of run with highlevel + runTask() with low level, which should not be the supported case at all. In addition, the internal state of LocalApplicationRunner is currently assumed to have a single application (i.e. processors are not separated by application instance names, counter down latch is not per application instance either). I would love to move to the direction that we enforce one application instance per AppRunner (i.e. essentially, we just create one application instance and runner is internally configured and attached for this application). I would prefer to move the doc effort of ApplicationRunner related change to a separate PR. This is the minimum required to get the current patch working.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
nickpan47: This is also part of the ApplicationRunner changes that I would love to address in a separate PR. Unfortunately, to remove the multiple invocation of init() per task context, this is the minimum that we will need to get this PR compiled and run.

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -108,6 +107,14 @@ public void onContainerFailed(Throwable t) {
     }
   }
 
+  private Object getTaskFactory(StreamApplication streamApp) {
+    if (streamApp != null) {
+      streamApp.init(streamGraph, config);
+      return TaskFactoryUtil.createTaskFactory(config, streamGraph);
+    }
+    return TaskFactoryUtil.createTaskFactory(config, null);
```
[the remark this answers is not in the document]
nickpan47: Done.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
nickpan47: Thanks! I have created a separate design doc to illustrate this change.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
nickpan47: Missed the doc change here! Thanks for pointing out. Fixed.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
nickpan47: Yep, shuffled during multiple rounds of merge. Revert to original.

samza-core/src/main/scala/org/apache/samza/job/local/ThreadJobFactory.scala
```diff
@@ -61,7 +63,16 @@ class ThreadJobFactory extends StreamJobFactory with Logging {
     val jmxServer = new JmxServer
     val streamApp = TaskFactoryUtil.createStreamApplication(config)
     val appRunner = new LocalContainerRunner(jobModel, "0")
-    val taskFactory = TaskFactoryUtil.createTaskFactory(config, streamApp, appRunner)
+    val streamGraph = Option(streamApp) match {
```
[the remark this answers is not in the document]
nickpan47: Done.

nickpan47:
Added overall design doc to SAMZA-1659: https://issues.apache.org/jira/secure/attachment/12920237/Serialization%20of%20OperatorSpec.pdf

Working on all fixes and merge conflicts.

nickpan47:
Updated to fix some of the review comments. Merge conflicts and more fixes WIP.

samza-core/src/main/java/org/apache/samza/operators/impl/PartitionByOperatorImpl.java
```diff
@@ -66,6 +64,8 @@
 
   @Override
   protected void handleInit(Config config, TaskContext context) {
+    this.keyFunction.init(config, context);
```
prateekm: What happens if someone implements these MapFunctions with a TimerFunction? Do we need to tie in the timer callbacks too?

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -112,11 +112,18 @@ public WindowOperatorImpl(WindowOperatorSpec<M, K, Object> windowOpSpec, Clock c
 
   @Override
   protected void handleInit(Config config, TaskContext context) {
-    WindowInternal<M, K, Object> window = windowOpSpec.getWindow();
 
     KeyValueStore<TimeSeriesKey<K>, Object> store =
         (KeyValueStore<TimeSeriesKey<K>, Object>) context.getStore(windowOpSpec.getOpId());
 
+    if (initializer != null) {
+      initializer.init(config, context);
```
prateekm: Same question as the keyFunction and valueFunction above.

samza-core/src/main/java/org/apache/samza/operators/spec/InputOperatorSpec.java
```diff
@@ -33,11 +33,16 @@
  */
 public class InputOperatorSpec<K, V> extends OperatorSpec<KV<K, V>, Object> { // Object == KV<K, V> | V
 
-  private final StreamSpec streamSpec;
-  private final Serde<K> keySerde;
-  private final Serde<V> valueSerde;
   private final boolean isKeyed;
 
+  /**
+   * The following {@link Serde}s are serialized by the ExecutionPlanner when generating the configs for a stream, and deserialized
```
prateekm: "The following fields ..." (includes streamSpec which is not a Serde)

Also, we don't need to serialize the input stream information in the StreamSpec for this operator?

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpec.java
```diff
@@ -79,10 +97,30 @@ public void registerNextOperatorSpec(OperatorSpec<OM, ?> nextOperatorSpec) {
     nextOperatorSpecs.add(nextOperatorSpec);
   }
 
+  /**
+   * Get the collection of chained {@link OperatorSpec}s that are consuming the output of this node
```
prateekm: "of this operator"

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorSpecGraph.java
```diff
@@ -0,0 +1,77 @@
...
+ * initialized in the {@link StreamGraphImpl} and constructsthe corresponding serialized instances of {@link OperatorSpec}.
+ * The {@link StreamGraphImpl} and {@link OperatorSpec} instances included in this class are considered as immutable and read-only.
+ * The instance of {@link OperatorSpecGraph} should only be used in runtime to construct {@link org.apache.samza.task.StreamOperatorTask}.
+ */
+public class OperatorSpecGraph {
+  private final Map<String, OperatorSpec> operatorSpecMap = new HashMap<>();
+  private final Map<String, byte[]> serializedOpSpecs = new ConcurrentHashMap<>();
+  private final StreamGraphImpl originalGraph;
+
+  public OperatorSpecGraph(StreamGraphImpl streamGraph) {
```
prateekm: Discussed offline: 
If we serialize this graph (both the operator specs vertices and the edges) in a planner, we can then avoid having to re-init the StreamApplication at runtime in the container to get the StreamGraphImpl (which is actually OperatorSpecGraph). This will also makes it easier to understand the OperatorImplGraph - it should take the serialized representation of the graph of operator specs, and create a corresponding graph of OperatorImpls.

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpec.java
```diff
@@ -79,10 +97,30 @@ public void registerNextOperatorSpec(OperatorSpec<OM, ?> nextOperatorSpec) {
...
+   *
+   * @return the collection of chained {@link OperatorSpec}s
+   */
   public Collection<OperatorSpec<OM, ?>> getRegisteredOperatorSpecs() {
     return nextOperatorSpecs;
   }
 
+  public static byte[] toBytes(OperatorSpec opSpec) throws IOException {
```
prateekm: This is the same logic as the `SerializableSerde`. We can use that instead of these methods.

samza-api/src/main/java/org/apache/samza/operators/functions/ClosableFunction.java
```diff
@@ -33,5 +33,8 @@
  */
 @InterfaceStability.Unstable
 public interface ClosableFunction {
+  /**
+   * Frees any resource initialized/acquired by the operators in {@link InitableFunction}
```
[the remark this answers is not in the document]
nickpan47: Done.

samza-api/src/main/java/org/apache/samza/operators/functions/SupplierFunction.java
```diff
@@ -0,0 +1,38 @@
...
+ * under the License.
+ */
+package org.apache.samza.operators.functions;
+
+import java.io.Serializable;
+import org.apache.samza.annotation.InterfaceStability;
+
+
+/**
+ * A supplier of values with life-cycle hooks for initialization and close. It is expected to return a new value at each invocation
```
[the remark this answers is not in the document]
nickpan47: Done.

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -80,47 +79,42 @@ public MessageStreamImpl(StreamGraphImpl graph, OperatorSpec<?, M> operatorSpec)
...
+    StreamOperatorSpec<M, M> op = OperatorSpecs.createFilterOperatorSpec(filterFn, this.graph.getNextOpId(OpCode.FILTER));
     this.operatorSpec.registerNextOperatorSpec(op);
     return new MessageStreamImpl<>(this.graph, op);
   }
 
   @Override
   public <TM> MessageStream<TM> flatMap(FlatMapFunction<? super M, ? extends TM> flatMapFn) {
-    String opId = this.graph.getNextOpId(OpCode.FLAT_MAP);
-    OperatorSpec<M, TM> op = OperatorSpecs.createFlatMapOperatorSpec(flatMapFn, opId);
+    StreamOperatorSpec<M, TM> op = OperatorSpecs.createFlatMapOperatorSpec(flatMapFn, this.graph.getNextOpId(OpCode.FLAT_MAP));
```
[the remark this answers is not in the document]
nickpan47: Done. Made the change throughout the file.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -328,8 +328,9 @@ public final void aggregateWatermark(WatermarkMessage watermarkMessage, SystemSt
     LOG.debug("Received watermark {} from {}", watermarkMessage.getTimestamp(), ssp);
     watermarkStates.update(watermarkMessage, ssp);
     long watermark = watermarkStates.getWatermark(ssp.getSystemStream());
-    if (watermark != WatermarkStates.WATERMARK_NOT_EXIST) {
+    if (currentWatermark < watermark) {
```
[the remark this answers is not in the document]
nickpan47: This is from the merge. I will double check.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorSpecGraph.java
```diff
@@ -0,0 +1,77 @@
...
+
+
+/**
+ * Defines the serialized format of {@link StreamGraphImpl}. This class encapsulates all getter methods to get the {@link OperatorSpec}
+ * initialized in the {@link StreamGraphImpl} and constructsthe corresponding serialized instances of {@link OperatorSpec}.
+ * The {@link StreamGraphImpl} and {@link OperatorSpec} instances included in this class are considered as immutable and read-only.
+ * The instance of {@link OperatorSpecGraph} should only be used in runtime to construct {@link org.apache.samza.task.StreamOperatorTask}.
+ */
+public class OperatorSpecGraph {
+  private final Map<String, OperatorSpec> operatorSpecMap = new HashMap<>();
```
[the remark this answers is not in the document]
nickpan47: Actually, made the change to the class s.t. allOperatorSpecs are only accessible from this class now. Hence, there should be no more confusion.

samza-core/src/main/java/org/apache/samza/operators/impl/PartitionByOperatorImpl.java
nickpan47: PartitionBy should always be triggered by events, not timer. Hence, I don't think there is a use case for TimerFunction as keyFunction. Even there is a use case later, we can always change the implementation of partitionBy() to allow the keyFunction to be a TimerFunction w/o the need to change the API.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
nickpan47: My answer would be same as well. We don't see a use case as of today. Adding the support later would only require implementation change, not API change.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -52,8 +52,11 @@
   private final StreamManager streamManager;
   private final SystemAdmins systemAdmins;
 
+  protected final StreamGraphImpl streamGraph;
```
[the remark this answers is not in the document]
nickpan47: Not just for testing any more. This is used to directly instantiate TaskFactory. I will work on separating the TaskFactory/AppRunner related changes outside this PR as much as possible, to avoid confusion.

samza-core/src/main/java/org/apache/samza/operators/impl/PartitionByOperatorImpl.java
prateekm: So the issue is that the keyFunction is just a MapFunction, and users can generally add `implements TimerFunction` to any of their custom functions. So there might be an expectation that this should work, but it will silently do nothing. So either support it, or maybe explicitly disallow this with instanceOf checks at runtime?

nickpan47:
@vjagadish1989 and @prateekm could you please review the updated revision? Merged with current master as well.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -39,60 +37,60 @@
 
 /**
  * A {@link StreamTask} implementation that brings all the operator API implementation components together and
- * feeds the input messages into the user-defined transformation chains in {@link StreamApplication}.
+ * feeds the input messages into the user-defined transformation chains in {@link org.apache.samza.operators.OperatorSpecGraph}.
```
prateekm: Minor: {@link OperatorSpecGraph} everywhere instead of fully qualified name.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -125,24 +132,24 @@ public ExecutionPlan getExecutionPlan(StreamApplication app) throws Exception {
 
   /* package private */
   ExecutionPlan getExecutionPlan(StreamApplication app, String runId) throws Exception {
-    // build stream graph
-    StreamGraphImpl streamGraph = new StreamGraphImpl(this, config);
-    app.init(streamGraph, config);
+    // getOperatorSpecGraph stream graph
```
prateekm: s/getOperatorSpecGraph/build

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -265,8 +265,8 @@ object SamzaContainer extends Logging {
     val serdes = serdesFromFactories ++ serdesFromSerializedInstances
 
     /*
-     * A Helper function to build a Map[String, Serde] (systemName -> Serde) for systems defined
-     * in the config. This is useful to build both key and message serde maps.
+     * A Helper function to getOperatorSpecGraph a Map[String, Serde] (systemName -> Serde) for systems defined
```
prateekm: s/getOperatorSpecGraph/build in this class.

samza-aws/src/main/java/org/apache/samza/system/kinesis/KinesisConfig.java
```diff
@@ -79,7 +79,7 @@ public KinesisConfig(Config config) {
    * @return a set of streams
    */
   public Set<String> getKinesisStreams(String system) {
-    // build stream-level configs
+    // getOperatorSpecGraph stream-level configs
```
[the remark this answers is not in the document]
nickpan47: Good catch! It might be changed by the refactor function in intelliji. Will fix them.

samza-core/src/main/java/org/apache/samza/operators/OperatorSpecGraph.java
```diff
@@ -0,0 +1,126 @@
...
+  /**
+   * Returns <tt>true</tt> iff this {@link StreamGraphSpec} contains a join or a window operator
+   *
+   * @return  <tt>true</tt> iff this {@link StreamGraphSpec} contains a join or a window operator
+   */
+  public boolean hasWindowOrJoins() {
+    return hasWindowOrJoins;
+  }
+
+  public OperatorSpecGraph clone() {
```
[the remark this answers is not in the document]
nickpan47: Sure. Will add the javadoc to explain the usage of this method.

samza-core/src/main/java/org/apache/samza/operators/spec/FilterOperatorSpec.java
```diff
@@ -0,0 +1,76 @@
...
+
+
+/**
+ * The spec for an operator that filters input messages based on some conditions. This is a package-private class since
+ * the constructor of the class is only accessed via {@link OperatorSpecs} and the runtime reference to the object
+ * is always upcast to {@link StreamOperatorSpec},
+ *
+ * @param <M> type of input message
+ */
+class FilterOperatorSpec<M> extends StreamOperatorSpec<M, M> {
```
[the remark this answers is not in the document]
nickpan47: Actually, you are right that the comment is a bit confusing and only talking about implementation details instead of public API. I will remove the explanation on why this is package-private.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
nickpan47: Good catch!

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpec.java
```diff
@@ -61,9 +62,15 @@
...
    * The set of operators that consume the messages produced from this operator.
    * <p>
-   * We use a LinkedHashSet since we need deterministic ordering in initializing/closing operators.
+   * We use a LinkedHashSet since we need both deterministic ordering in initializing/closing operators and serializability.
    */
-  private final Set<OperatorSpec<OM, ?>> nextOperatorSpecs = new LinkedHashSet<>();
+  private final LinkedHashSet<OperatorSpec<OM, ?>> nextOperatorSpecs = new LinkedHashSet<>();
+
+  @VisibleForTesting
+  final boolean isClone(OperatorSpec other) {
```
prateekm: Is this still required?

vjagadish1989 (approved):
looks good to me, approved. thanks!

samza-core/src/main/java/org/apache/samza/operators/spec/OperatorSpec.java
nickpan47: As annotation indicates, this method is only for unit test usage now.

prateekm: Ah ok. Can we add a comment that this is only for testing instead of the annotation? VisibleForTesting usually indicates that the method has a looser access restriction (e.g. protected or public instead of private) for testing purposes, but should otherwise be considered private.

PR #476, opened 2018-04-17
SAMZA-1664: ZkJobCoordinator stability fixes.

shanthoosh:
Issues fixed:
* Handle job coordinator shutdown gracefully in case of unclean container shutdowns.
* Fix the zookeeper session handling logic.
* Fix the forever retry timeout in ZkClient re-connect.


shanthoosh:
@xinyuiscool 

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -144,27 +157,27 @@ private Runnable getScheduleableAction(String actionName, Runnable runnable) {
         } else {
           LOG.debug("Action: {} completed successfully.", actionName);
         }
-      } catch (Throwable t) {
-        LOG.error("Execution of action: {} failed.", actionName, t);
-        doCleanUpOnTaskException(t);
+      } catch (Exception exception) {
```
bharathkk: I don't see these changes in hot-fix branch.
Can you provide context on why are we changing this to `exception` from `throwable`?

shanthoosh: Reverted this change.

PR #478, opened 2018-04-17
SAMZA-1651: Samza-sql - Implement GROUP BY SQL operator

atoomula:


samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
```diff
@@ -92,6 +95,9 @@
 
   private final List<QueryInfo> queryInfo;
 
+  private final long windowDurationMs;
+  private static final long DEFAULT_WINDOW_DURATION_MS = 300000; // default window duration is 5 mins.
```
pdu-mn1: nit: move constant to top

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
```diff
@@ -77,6 +77,9 @@
 
   public static final String CFG_UDF_RESOLVER = "samza.sql.udfResolver";
   public static final String CFG_FMT_UDF_RESOLVER_DOMAIN = "samza.sql.udfResolver.%s.";
+
+  public static final String CFG_WINDOW_DURATION_MS = "samza.sql.window.ms";
```
pdu-mn1: nit: should we prefix this with groupby, ie. "samza.sql.groupby.window.ms", since this is only used by groupBy?

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
```diff
@@ -96,6 +98,14 @@ public RelNode visit(LogicalJoin join) {
         new JoinTranslator(joinId[0], sourceResolver).translate(join, context);
         return node;
       }
+
+      @Override
+      public RelNode visit(LogicalAggregate aggregate) {
+        RelNode node = super.visit(aggregate);
+        windowId[0] += 2;
```
pdu-mn1: nit: can you add a comment on why this is incremented by 2 for each node?

samza-sql/src/main/java/org/apache/samza/sql/translator/LogicalAggregateTranslator.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  void translate(final LogicalAggregate aggregate, final TranslatorContext context) {
+    MessageStream<SamzaSqlRelMessage> inputStream = context.getMessageStream(aggregate.getInput().getId());
+
+    Supplier<Long> initialValue = () -> (long) 0;
+    FoldLeftFunction<SamzaSqlRelMessage, Long> foldLeftFn = (m, c) -> c + 1;
+
+    MessageStream<SamzaSqlRelMessage> outputStream =
+        inputStream
```
pdu-mn1: curious: do we need to expose early triggers for groupby?

samza-sql/src/test/java/org/apache/samza/sql/system/TestAvroSystemFactory.java.orig
```diff
@@ -0,0 +1,299 @@
+/*
```
pdu-mn1: nit: This should not be part of the PR right?

samza-sql/src/main/java/org/apache/samza/sql/translator/LogicalAggregateTranslator.java
```diff
@@ -0,0 +1,71 @@
...
+
+  LogicalAggregateTranslator(int windowId) {
+    this.windowId = windowId;
+  }
+
+  void translate(final LogicalAggregate aggregate, final TranslatorContext context) {
+    MessageStream<SamzaSqlRelMessage> inputStream = context.getMessageStream(aggregate.getInput().getId());
+
+    Supplier<Long> initialValue = () -> (long) 0;
+    FoldLeftFunction<SamzaSqlRelMessage, Long> foldLeftFn = (m, c) -> c + 1;
```
srinipunuru: This foldLeftFun is specifically for count right? Can you call it as countFn?

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
```diff
@@ -65,6 +66,7 @@ public void translate(SamzaSqlQueryParser.QueryInfo queryInfo, StreamGraph strea
     final TranslatorContext context = new TranslatorContext(streamGraph, relRoot, executionContext);
     final RelNode node = relRoot.project();
     final int[] joinId = new int[1];
+    final int[] windowId = new int[1];
```
srinipunuru: Looks like both windowId and joinId is used only within in the anonymous fn. I think you can move them  inside the anonymous class RelShuttleImpl, that ways you could convert them into a regular integer rather than a single valued array?

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
srinipunuru: Can you add a comment on why this needs to be increased by 2?

samza-sql/src/test/java/org/apache/samza/sql/system/TestAvroSystemFactory.java.orig
```diff
@@ -0,0 +1,299 @@
...
+* Licensed to the Apache Software Foundation (ASF) under one
+* or more contributor license agreements.  See the NOTICE file
+* distributed with this work for additional information
```
srinipunuru: Is this file checked in by mistake? it's called .java.orig

PR #479, opened 2018-04-18
SAMZA-1667: Skip storing configuration as a part of JobModel in zookeeper data nodes.

shanthoosh:
In general, jobModel configuration contains service access tokens and certs. It's a common practice to run zookeeper in a non-ACL environment. Hence for security purposes, it's essential not to store configuration as a part of JobModel in zookeeper.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
@@ -33,6 +33,7 @@ import org.apache.samza.config.ApplicationConfig;
 import org.apache.samza.config.Config;
 import org.apache.samza.config.ConfigException;
 import org.apache.samza.config.JobConfig;
+import org.apache.samza.config.MapConfig;
 import org.apache.samza.config.MetricsConfig;
 import org.apache.samza.config.TaskConfigJava;
 import org.apache.samza.config.ZkConfig;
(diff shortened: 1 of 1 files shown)


shanthoosh:
@xinyuiscool @bharathkk 

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -348,7 +349,9 @@ private JobModel generateNewJobModel(List<String> processors) {
      * Host affinity is not supported in standalone. Hence, LocalityManager(which is responsible for container
      * to host mapping) is passed in as null when building the jobModel.
      */
-    return JobModelManager.readJobModel(this.config, changeLogPartitionMap, null, streamMetadataCache, processors);
+    JobModel model = JobModelManager.readJobModel(this.config, changeLogPartitionMap, null, streamMetadataCache, processors);
+    // Nuke the configuration in JobModel.
+    return new JobModel(new MapConfig(), model.getContainers());
```
vjagadish: Assert it in an existing unit-test maybe?

shanthoosh: Done.

prateekm:
+1 for Bharath's comment. There's no indication anywhere that the caller isn't supposed to use the configs from the JobModel passed to it. What's the plan for the config field in the JobModel?

shanthoosh:
@prateekm @bharathkk 
IMHO, JobModel is meant to store the task to partitions assignment and logically config does not belong there. 
We have used it so far to propagate config to SamzaContainer running on a NodeManager in yarn.
I think in long term, we should move it out of the JobModel.

For now, I will add a comment for it.

PR #483, opened 2018-04-23
SAMZA-1670 : When fetching a newest offset for a partition, also prefetch and cache the newest offsets for other partitions on the container

cameronlee314:
Testing done:
./gradlew clean check
./bin/integration-tests.sh
Implemented getNewestOffset with a custom admin and verified that newest offsets get correctly written and used when restoring stores


prateekm:
@cameronlee314 Thanks for this. Can you create a JIRA at issues.apache.org and add it to the PR title.

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -94,17 +112,120 @@ class StreamMetadataCache (
...
-        if (now - lastRefresh > cacheTTLms) None else Some(stream -> metadata)
+  /**
+    * Using a MapLike argument in order to work with both immutable.Map and mutable.Map to get an entry.
+    */
+  private def getFromCache[K, V, M <: MapLike[K, CacheEntry[V], M] with scala.collection.Map[K, CacheEntry[V]]]
+    (cache: MapLike[K, CacheEntry[V], M],
+     key: K,
+     now: Long): Option[(K, V)] = {
+    cache.get(key) match {
+      case Some(CacheEntry(data, lastRefresh)) => if (isStale(now, lastRefresh)) None else Some(key -> data)
```
prateekm: Minor: Maybe remove stale entries from Map?

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -94,17 +112,120 @@ class StreamMetadataCache (
...
+     * the metadata fetch request now if it is stale. Also make sure it has the same system, since we will only use the
+     * admin for the system of the requestedSystemStreamPartition.
+     */
+    var systemStreamPartitionsToFetch = newestOffsetCache
+      .filter({case (ssp, cacheEntry) => systemToFetch.equals(ssp.getSystem) && isStale(now, cacheEntry.lastRefreshMs)})
+      .keySet.toSet
+    if (!systemStreamPartitionsToFetch.contains(requestedSystemStreamPartition)) {
+      systemStreamPartitionsToFetch += requestedSystemStreamPartition
+    }
+    val newestOffsets: Map[SystemStreamPartition, String] = systemAdmins.getSystemAdmin(systemToFetch) match {
```
prateekm: I think it'll be cleaner to have a single method 'getNewestOffsets(ssps, maxRetries)` that combines lines 179 to 186, and inlines fetchNewestoffsetsForSystemAdmin.

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -94,17 +112,120 @@ class StreamMetadataCache (
...
+      case extendedSystemAdmin: ExtendedSystemAdmin =>
+        /*
+         * ExtendedSystemAdmin has a way to fetch the newest offsets for specific SSPs instead of fetching the
+         * newest and oldest offsets for all SSPs. Use it if we can.
+         */
+        extendedSystemAdmin.getNewestOffsets(systemStreamPartitionsToFetch.asJava, maxRetries).asScala.toMap
+      case systemAdmin => fetchNewestOffsetsForSystemAdmin(systemStreamPartitionsToFetch, systemAdmin)
+    }
+    addNewestOffsetsToCache(systemStreamPartitionsToFetch, newestOffsets, now)
+    newestOffsets.get(requestedSystemStreamPartition) match {
```
prateekm: Minor: I think this is equivalent to newestOffsets.get(requestedSSP).orNull

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -94,17 +112,120 @@ class StreamMetadataCache (
...
+    systemStreamPartitionsToFetch.foreach(systemStreamPartition => {
+      // want to put an entry in regardless of if we got data from newestOffsets
+      newestOffsetCache.update(systemStreamPartition, CacheEntry(newestOffsets.get(systemStreamPartition), time))
+    })
+  }
+
+  def fetchNewestOffsetsForSystemAdmin(systemStreamPartitionsToFetch: Set[SystemStreamPartition],
+                                       systemAdmin: SystemAdmin): Map[SystemStreamPartition, String] = {
+    val streamsToFetch = systemStreamPartitionsToFetch
+      .map(systemStreamPartitionToFetch => systemStreamPartitionToFetch.getStream)
```
prateekm: Minor: `sspToFetch.map(_.getStream)`

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -94,17 +112,120 @@ class StreamMetadataCache (
...
+    val streamToMetadata = systemAdmin.getSystemStreamMetadata(streamsToFetch.asJava).asScala.toMap
```
prateekm: Minor: Might be cleaner to use the returned Java Map and use if/else instead of Scala Map + Options.

prateekm: Maybe update the full metadata cache as well if we already have the information?

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -78,7 +96,7 @@ class StreamMetadataCache (
       throw new SamzaException("Cannot get metadata for unknown streams: " + missing.mkString(", "))
     }
     if (!partitionsMetadataOnly) {
-      cacheMisses.foreach { case (stream, metadata) => addToCache(stream, metadata, time) }
+      cacheMisses.foreach { case (stream, metadata) => addToFullMetadataCache(stream, metadata, time) }
```
prateekm: Minor: prefer inlining this single use method.

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -39,10 +41,26 @@ class StreamMetadataCache (
...
+  private var fullMetadataCache = Map[SystemStream, CacheEntry[SystemStreamMetadata]]()
   private val lock = new Object
 
+  /**
+    * Caches newest offset data.
+    * If there is a CacheEntry with Some data, and the TTL has not passed, then use the data and do not fetch.
+    * If there is a CacheEntry with None data, and the TTL has not passed, then use the empty entry and do not fetch.
+    * If there is no CacheEntry, then fetch the metadata and fill in the cache.
+    *
+    * Using a mutable.Map here instead of immutable.Map because this will get updated relatively often, so we do not
```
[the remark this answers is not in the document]
prateekm: Also, would it be reasonable to replace the lock + Map mutation with a ConcurrentHashMap instead?

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -94,17 +112,120 @@ class StreamMetadataCache (
...
+                                               now: Long): String = {
+    val systemToFetch = requestedSystemStreamPartition.getSystem
+    /*
+     * If a systemStreamPartition is in the newestOffsetCache, then we have fetched it before, so include it in
+     * the metadata fetch request now if it is stale. Also make sure it has the same system, since we will only use the
+     * admin for the system of the requestedSystemStreamPartition.
+     */
+    var systemStreamPartitionsToFetch = newestOffsetCache
+      .filter({case (ssp, cacheEntry) => systemToFetch.equals(ssp.getSystem) && isStale(now, cacheEntry.lastRefreshMs)})
+      .keySet.toSet
```
prateekm: Probably don't need the final .toSet

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -94,17 +112,120 @@ class StreamMetadataCache (
...
+                                               maxRetries: Integer,
+                                               now: Long): String = {
+    val systemToFetch = requestedSystemStreamPartition.getSystem
+    /*
+     * If a systemStreamPartition is in the newestOffsetCache, then we have fetched it before, so include it in
+     * the metadata fetch request now if it is stale. Also make sure it has the same system, since we will only use the
+     * admin for the system of the requestedSystemStreamPartition.
+     */
+    var systemStreamPartitionsToFetch = newestOffsetCache
+      .filter({case (ssp, cacheEntry) => systemToFetch.equals(ssp.getSystem) && isStale(now, cacheEntry.lastRefreshMs)})
```
prateekm: Minor: Don't need () around {}. I.e. `filter { case (s, e) => ... }` should work

samza-api/src/main/java/org/apache/samza/system/ExtendedSystemAdmin.java
```diff
@@ -18,16 +18,43 @@
...
   String getNewestOffset(SystemStreamPartition ssp, Integer maxRetries);
+
+  /**
+   * Batch call for getting newest offsets for multiple SystemStreamPartitions.
+   * By default, this will delegate to the individual getNewestOffset (also gives backwards compatibility), but it can
+   * be overridden by implementors as an optimization.
+   * There will be no entry for an SSP if no newest offset was found (e.g. topic is empty).
+   * @param maxRetriesPerSSP Max number of attempts to fetch the newest offset for an individual SSP
+   */
+  default Map<SystemStreamPartition, String> getNewestOffsets(Set<SystemStreamPartition> ssps,
```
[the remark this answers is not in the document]
cameronlee314: Yeah, the default impl max retries was a bit weird to interpret. Is it ok to change the interface for ExtendedSystemAdmin.getNewestOffset to remove the retries parameter? Is there a backwards compatibility concern?

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -39,10 +41,26 @@ class StreamMetadataCache (
...
   private val lock = new Object
 
+  /**
+    * Caches newest offset data.
+    * If there is a CacheEntry with Some data, and the TTL has not passed, then use the data and do not fetch.
+    * If there is a CacheEntry with None data, and the TTL has not passed, then use the empty entry and do not fetch.
+    * If there is no CacheEntry, then fetch the metadata and fill in the cache.
+    *
+    * Using a mutable.Map here instead of immutable.Map because this will get updated relatively often, so we do not
+    * need to create new maps for each update. This does need to be protected by a lock for thread safety.
```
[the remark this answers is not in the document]
prateekm: Just noticed that we're actually synchronizing on the newestOffsetsCache field. Would recommend making it a val in that case.

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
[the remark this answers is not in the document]
cameronlee314: I initially was using mutable.Map for both, but then I realized that this is not threadsafe given the way that the fullMetadataCache is currently being read/written. I think immutable.Map gives thread safety since it creates a new Map on every update, so you can't catch it in an intermediate update state. I could try to add some thread safety protection around the fullMetadataCache if you would prefer that.
I don't think ConcurrentHashMap helps because if two threads both read a stale entry, then they will both try to fetch the metadata, and that defeats the caching benefit. I don't think ConcurrentHashMap can support atomically doing that refill. There is probably some actual threadsafe caching library we could use which provides TTL and a computeIfAbsent. I could look into that if you think it is worth it.

samza-api/src/main/java/org/apache/samza/system/ExtendedSystemAdmin.java
```diff
@@ -18,16 +18,47 @@
...
+   * Returns null if no newest offset was found (e.g. topic is empty).
+   *
+   * Deprecated: Use/implement getNewestOffsets instead
+   */
+  @Deprecated
   String getNewestOffset(SystemStreamPartition ssp, Integer maxRetries);
+
+  /**
+   * Batch call for getting newest offsets for multiple SystemStreamPartitions.
+   * Override the default implementation if a more efficient batch get exists. The default implementation exists for
```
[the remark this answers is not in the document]
prateekm: Btw, do we need a try/catch in fetchNewestOffsetsWithSystemAdmin? Either we should handle exceptions, or document it in StreamMetadataCache#getNewestOffset

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -94,17 +106,111 @@ class StreamMetadataCache (
...
+  private def fetchNewestOffsetWithPrefetching(requestedSSP: SystemStreamPartition, now: Long): String = {
+    val systemToFetch = requestedSSP.getSystem
+    /*
+     * If an ssp is in the newestOffsetCache, then we have fetched it before, so include it in the metadata fetch
+     * request now if it is stale. Also make sure it has the same system, since we will only use the admin for the
+     * system of the requestedSSP.
+     */
+    val sspsToFetchBuilder = new ImmutableSet.Builder[SystemStreamPartition]()
+    sspsToFetchBuilder.add(requestedSSP)
+    for (entry <- newestOffsetCache.entrySet.asScala) {
```
vjagadish: Would if it be more conventional to iterate this as a java collection + filter?

samza-api/src/main/java/org/apache/samza/system/ExtendedSystemAdmin.java
```diff
@@ -18,16 +18,53 @@
...
 /**
  * Interface extends the more generic SystemAdmin interface
  * TODO: Merge this interface method with SystemAdmin when we upgrade to JDK 1.8
  */
 public interface ExtendedSystemAdmin extends SystemAdmin {
   Map<String, SystemStreamMetadata> getSystemStreamPartitionCounts(Set<String> streamNames, long cacheTTL);
 
-  // Makes fewer offset requests than getSystemStreamMetadata
+  /**
+   * Deprecated: Use/implement getNewestOffsets instead
```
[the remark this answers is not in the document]
prateekm: +1

samza-api/src/main/java/org/apache/samza/system/ExtendedSystemAdmin.java
```diff
@@ -18,16 +18,53 @@
...
+   * than the "no newest offset exists" case, because in the "no newest offset exists" case, the newest offset info was
+   * accessible, but none existed.
+   */
+  @Deprecated
   String getNewestOffset(SystemStreamPartition ssp, Integer maxRetries);
+
+  /**
+   * Gets the newest offsets for {@code ssps}.
+   * <p>
+   *   Implementation notes: Override the default implementation if a more efficient batch get exists. It is up to the
```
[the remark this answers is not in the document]
prateekm: +1 for both, with s/merely// in 1

samza-api/src/main/java/org/apache/samza/system/ExtendedSystemAdmin.java
```diff
@@ -18,16 +18,53 @@
...
+   * corresponding value is the newest offset. An ssp which does not have a newest offset (e.g. topic is empty) does not
+   * have an entry in the map.
+   * @throws RuntimeException if the newest offset information could not be accessed. This exception case is different
+   * than the "no newest offset exists" case, because in the "no newest offset exists" case, the newest offset info was
+   * accessible, but none existed.
+   */
+  default Map<SystemStreamPartition, String> getNewestOffsets(Set<SystemStreamPartition> ssps) {
+    // default implementation exists for backwards compatibility
+    // some fallback retry count since getNewestOffset needs it; remove this when getNewestOffset is removed
+    final int defaultRetriesPerSSP = 3;
```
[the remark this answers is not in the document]
prateekm: +1

samza-core/src/main/scala/org/apache/samza/system/StreamMetadataCache.scala
```diff
@@ -94,17 +106,111 @@ class StreamMetadataCache (
...
+    */
+  def getNewestOffset(systemStreamPartition: SystemStreamPartition): String = {
+    // needs synchronization so that other threads don't come in and try to also do a fetch
+    newestOffsetLock synchronized {
+      val now = clock.currentTimeMillis
+      getFromCache(newestOffsetCache, systemStreamPartition, now) match {
+        // has a fresh enough entry; even if it does not have data, don't do another fetch yet
+        case Some((key, newestOffset)) => newestOffset.orElse(null)
+        // no entry or stale entry; do some fetching
+        case None => fetchNewestOffsetWithPrefetching(systemStreamPartition, now)
```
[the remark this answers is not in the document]
prateekm: Good call, +1.

prateekm:
Summarizing offline discussion with @vjagadish1989, @jmakes and @cameronlee314:

While this PR looks pretty good and will work as-is, it would be cleaner to take a different approach to the API.

The suggestion is to deprecate the current `ExtendedSystemAdmin#getNewestOffset(SystemStreamPartition ssp, Integer maxRetries)` API and add a new `SystemAdmin#getSystemStreamMetadata(Set<SystemStreamPartition> ssps)` API instead. This new API is an optimization of the existing `SystemAdmin#getSystemStreamMetadata(Set<String> streams)` API, in that it allows fetching metadata for specific partitions instead of the entire stream. The default implementation of this new API can fetch metadata for all streams in the requested SSP set using the current API, and select the SSPs that were requested. Systems can override this if a better batch API is available.

Then, we can add a cache on top of this new API that has similar prefetch logic as in this PR. For example, it could be a use-case specific `ChangelogMetadataCache` that maintains metadata for the changelog partitions for the container, and is used by the TaskStorageManager. Alternatively, it could be a more general implementation of the StreamMetadataCache, (or a separate SSPMetadataCache) that maintains metadata for all SSPs used in the container. We could also tune the cache TTL based on the commit interval such that we reduce the frequency of metadata fetches.

cameronlee314:
Closing: replaced by https://github.com/apache/samza/pull/520

vjagadish1989:
@cameronlee314 : If the remote branch is not deleted, I wonder if this PR can be resurrected? We should strive to complete review cycles in the same PR if possible.

PR #484, opened 2018-04-23
SAMZA-1676 miscellaneous fix and improvement for eventhubs system

lhaiesp:
Including these changes:
- Log the metadata that we are fetching from the event hubs
- Rename readLatency to consumptionLagMs
- fix the issue that readLatency metric returns negative value

### DIFF
--- a/samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
@@ -19,6 +19,7 @@
 
 package org.apache.samza.system.eventhub.admin;
 
+import com.microsoft.azure.eventhubs.EventHubClient;
 import com.microsoft.azure.eventhubs.EventHubRuntimeInformation;
 import com.microsoft.azure.eventhubs.PartitionRuntimeInformation;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
```diff
@@ -90,7 +96,8 @@ private String getNextOffset(String currentOffset) {
 
           long timeoutMs = eventHubConfig.getRuntimeInfoWaitTimeMS(systemName);
           EventHubRuntimeInformation ehInfo = runtimeInfo.get(timeoutMs, TimeUnit.MILLISECONDS);
-          LOG.info(String.format("Adding partition ids=%s for stream=%s", Arrays.toString(ehInfo.getPartitionIds()), streamName));
+          LOG.info(String.format("Adding partition ids=%s for stream=%s. EHRuntimetInfo=%s",
+              Arrays.toString(ehInfo.getPartitionIds()), streamName, printEventHubRuntimeInfo(ehInfo)));
```
srinipunuru: Can you log the partition level metadata that we are returning including the number of messages in the partition (endSequenceNumber-startSequenceNumber)

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -172,7 +172,7 @@ public EventHubSystemConsumer(EventHubConfig config, String systemName,
       if (aggEventReadRate == null) {
         aggEventReadRate = registry.newCounter(AGGREGATE, EVENT_READ_RATE);
         aggEventByteReadRate = registry.newCounter(AGGREGATE, EVENT_BYTE_READ_RATE);
-        aggReadLatency = new SamzaHistogram(registry, AGGREGATE, READ_LATENCY);
+        aggReadLatency = new SamzaHistogram(registry, AGGREGATE, CONSUMPTION_LATENCY_MS);
```
vjagadish1989: consumptionLagMs? Personally, I'd prefer to not use "latency"

samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
vjagadish1989: @lhaiesp : Is this a one-time logging operation or a periodic one?

lhaiesp: Hi, @vjagadish1989  we log this once before we add these partitionId into streamPartitions. After that we won't go into the same execution branch.

vjagadish: makes sense, thanks for explaining.

PR #486, opened 2018-04-24
SAMZA-1582 : removed the yajl-ruby dependency for security reasons

cameronlee314:
The original ticket suggested updating the yajl-ruby dependency for security reasons.
Many of the ruby dependencies for building the docs site are years old. After upgrading the dependencies, the yajl-ruby dependency just went away.
The dependencies were upgraded to the highest version that was compatible with ruby 2.0.0 (jekyll was the only dependency which has newer versions that are only compatible with ruby versions greater than 2.0.0).

Testing done:
bundle exec jekyll serve --watch --baseurl ""
_docs/local-site-test.sh
sanity checked a couple of links


vjagadish1989:
LGTM. Were some of the dependencies unsupported/EOL'ed? 
nit: update the PR title.

vjagadish:
Merged and submitted! Thanks @cameronlee314. Please resolve/close the JIRA with the fix version

PR #487, opened 2018-04-24
SAMZA-1686: Set finite operation timeout when creating zkClient.

shanthoosh:
Currently zkClient is created with operationRetryTimeOut of -1. This causes zkClient to retry indefinitely in case of irrecoverable exceptions thereby delaying the StreamProcessor shutdown.


vjagadish1989 (approved):
looks good to me. approved. thank you Shanthoosh for the contribution

PR #489, opened 2018-04-25
SAMZA-1688 use per partition eventhubs client

lhaiesp:
Use per partition eventhubs client to improve throughput. As each eventhub client maintains only one single TCP connection.
Also reduce the time spent on getPartitionRuntimeInfo on starting up, by making the calls run in parallels in system admin and also completely removing it from system consumer.

Benchmark:
See 8x improvement on consumption throughput from ~3Mb/s (or 1.5k QPS) to ~21Mb/s (or 10.5k QPS). Memory footprint doesn't seem to get affected. CPU usage also increases by 8x. Essentially, with per partition client, we are able to saturate the CPU usages. The benchmark I did was on a machine with 8 cores. If the machine has say 16 CPU cores, I expect the the throughput to improve by around 16x.

### DIFF
--- a/samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
@@ -361,6 +361,7 @@ public class EventHubSystemConsumer extends BlockingEnvelopeMap {
       LOG.warn("Failed to close receivers", e);
     }
     perPartitionEventHubManagers.values()
+        .parallelStream()
         .forEach(ehClientManager -> ehClientManager.close(DEFAULT_SHUTDOWN_TIMEOUT_MILLIS));
     perPartitionEventHubManagers.clear();
     perStreamEventHubManagers.clear();
--- a/samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
@@ -19,6 +19,7 @@
 
 package org.apache.samza.system.eventhub.consumer;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.microsoft.azure.eventhubs.EventData;
 import com.microsoft.azure.eventhubs.EventHubException;
 import com.microsoft.azure.eventhubs.EventPosition;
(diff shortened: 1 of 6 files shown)


lhaiesp:
Working on benchmarking

vjagadish:
1. Doesn't the current implementation already uses a single consumption thread per-partition of the client? It'd be really interesting if the per-partition client performs better. 
2. Currently the shut-down of the container is dependent on the perf-impact of ehClient.close(). Should also ensure that adding multiple clients does not increase the shut-down timeout.

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -306,7 +343,7 @@ public void start() {
 
   private void renewPartitionReceiver(SystemStreamPartition ssp) {
     String streamId = config.getStreamId(ssp.getStream());
-    EventHubClientManager eventHubClientManager = streamEventHubManagers.get(streamId);
+    EventHubClientManager eventHubClientManager = perPartitionEventHubManagers.get(ssp);
```
srinipunuru: What happens when the connection is lost? Don't we need to renew the EventHubClient to create a new connection?

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -219,13 +220,49 @@ private String getNewestEventHubOffset(EventHubClientManager eventHubClientManag
     }
   }
 
+  // Based on the config PerPartitionConnection, create or get EventHubClientManager for the SSP
+  // Note: this should be used only when starting up. After initialization, directly use perPartitionEventHubManagers
+  // to obtain the corresponding EventHubClientManager
+  private EventHubClientManager createOrGetEventHubClientManagerForSSP(String streamId, SystemStreamPartition ssp) {
```
srinipunuru: Do we have data on what is the overhead in terms or memory? Can you add this to description if you do?

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
lhaiesp: Temporary connection lost doesn't need a new EventHubClient. I have tested the behavior.

PR #490, opened 2018-04-27
SAMZA-1689: Add validations before state transitions in ZkBarrierForVersionUpgrade.

shanthoosh:
Prevent invalid state updations on barrier.
* Introduced a additional barrier state NEW.
* Add state validations before updating the barrier.
* Fix existing TestZkBarrier tests that are disabled and add new tests
  to verify the intended behavior.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
@@ -42,6 +42,25 @@ import java.util.Optional;
  * with value {@link org.apache.samza.zk.ZkBarrierForVersionUpgrade.State#TIMED_OUT} and indicates to everyone that it
  * is no longer valid.
  *
+ *
+ * Describes the lifecycle of a barrier.
+ * <pre>
+ *
+ *                               When expected participants join
+ *      Leader    ---&lt; NEW ---------------------------------------- &lt; DONE
+ *                         |      barrier within barrierTimeOut.
+ *                         |
+ *                         |
+ *                         |
+ *                         |
+ *                         |
+ *                         |    When expected participants doesn't
+ *                         | ----------------------------------------- &lt; TIMED_OUT
+ *                              join barrier within barrierTimeOut.
+ *
+ * </pre>
+ *
+ *
  * The caller can listen to events associated with the barrier by registering a {@link ZkBarrierListener}.
  *
  * Zk Tree Reference:
@@ -214,7 +233,7 @@ public class ZkBarrierForVersionUpgrade {
         zkUtils.unsubscribeDataChanges(barrierStatePath, this);
         barrierListenerOptional.ifPresent(zkBarrierListener -> zkBarrierListener.onBarrierStateChanged(barrierVersion, (State) data));
       } else {
-        LOG.debug("Barrier version: {} has state: {}. Ignoring the barrierState change notification.", barrierVersion, barrierState);
+        LOG.debug("Barrier version: {} is at state: {}. Ignoring the barrierState change notification.", barrierVersion, barrierState);
       }
     }
 
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
@@ -19,11 +19,12 @@
 
 package org.apache.samza.zk;
 
+import com.google.common.collect.ImmutableList;
+import java.util.Objects;
 import org.apache.commons.collections4.CollectionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 
@@ -48,7 +49,7 @@ import java.util.Optional;
  *  |
  *  |- barrier_{version1}/
  *  |   |- barrier_state/
- *  |   |  ([DONE|TIMED_OUT])
+ *  |   |  ([NEW|DONE|TIMED_OUT])
  *  |   |- barrier_participants/
  *  |   |   |- {id1}
  *  |   |   |- {id2}
@@ -61,9 +62,19 @@ public class ZkBarrierForVersionUpgrade {
   private final Optional<ZkBarrierListener> barrierListenerOptional;
 
   public enum State {
-    TIMED_OUT, DONE
-  }
+    NEW("NEW"), TIMED_OUT("TIMED_OUT"), DONE("DONE");
 
+    private String str;
+
+    State(String str) {
+      this.str = str;
+    }
+
+    @Override
+    public String toString() {
+      return str;
+    }
+  }
 
   public ZkBarrierForVersionUpgrade(String barrierRoot, ZkUtils zkUtils, ZkBarrierListener barrierListener) {
     if (zkUtils == null) {
@@ -81,16 +92,19 @@ public class ZkBarrierForVersionUpgrade {
    * @param participants List of expected participated for this barrier to complete
    */
   public void create(final String version, List<String> participants) {
+    LOG.info(String.format("Creating barrier with version: %s, participants: %s.", version, participants));
     String barrierRoot = keyBuilder.getBarrierRoot();
     String barrierParticipantsPath = keyBuilder.getBarrierParticipantsPath(version);
+    String barrierStatePath = keyBuilder.getBarrierStatePath(version);
     zkUtils.validatePaths(new String[]{
         barrierRoot,
         keyBuilder.getBarrierPath(version),
         barrierParticipantsPath,
-        keyBuilder.getBarrierStatePath(version)});
+        barrierStatePath});
+    LOG.info("Marking the barrier state: {} as {}.", barrierStatePath, State.NEW);
+    zkUtils.writeData(barrierStatePath, State.NEW);
 
-    // subscribe for participant's list changes
-    LOG.info("Subscribing for child changes at " + barrierParticipantsPath);
+    LOG.info("Subscribing child changes on the path: {} for barrier version: {}.", barrierParticipantsPath, version);
     zkUtils.subscribeChildChanges(barrierParticipantsPath, new ZkBarrierChangeHandler(version, participants, zkUtils));
 
     barrierListenerOptional.ifPresent(zkBarrierListener -> zkBarrierListener.onBarrierCreated(version));
@@ -103,8 +117,10 @@ public class ZkBarrierForVersionUpgrade {
    * @param participantId Identifier of the participant
    */
   public void join(String version, String participantId) {
-    String barrierDonePath = keyBuilder.getBarrierStatePath(version);
-    zkUtils.subscribeDataChanges(barrierDonePath, new ZkBarrierReachedHandler(barrierDonePath, version, zkUtils));
+    LOG.info("Joining the barrier version: {} as participant: {}.", version, participantId);
+    String barrierStatePath = keyBuilder.getBarrierStatePath(version);
+    LOG.info("Subscribing data changes on the path: {} for barrier version: {}.", barrierStatePath, version);
+    zkUtils.subscribeDataChanges(barrierStatePath, new ZkBarrierReachedHandler(barrierStatePath, version, zkUtils));
 
     // TODO: Handle ZkNodeExistsException - SAMZA-1304
     zkUtils.getZkClient().createPersistent(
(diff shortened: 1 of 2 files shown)


shanthoosh:
@prateekm @vjagadish1989 @xinyuiscool 

Verification: 

`
i=0
while [ $i -lt 100 ]; do
                i=expr $i + 1
                ./gradlew clean :samza-core:test -Dtest.single="TestZkBarrierForVersionUpgrade"  >> ~/test-logs_32
done;
`

Result: 

 samza: grep -i 'BUILD SUCCESS' ~/test-logs_32 | wc -l
     100
 samza:  grep -i 'BUILD FAIL' ~/test-logs_32 | wc -l
       0

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -81,16 +91,19 @@ public ZkBarrierForVersionUpgrade(String barrierRoot, ZkUtils zkUtils, ZkBarrier
...
     String barrierRoot = keyBuilder.getBarrierRoot();
     String barrierParticipantsPath = keyBuilder.getBarrierParticipantsPath(version);
+    String barrierStatePath = keyBuilder.getBarrierStatePath(version);
     zkUtils.validatePaths(new String[]{
         barrierRoot,
         keyBuilder.getBarrierPath(version),
         barrierParticipantsPath,
-        keyBuilder.getBarrierStatePath(version)});
+        barrierStatePath});
+    LOG.info(String.format("Marking the barrier state: %s as %s.", barrierStatePath, State.NEW));
```
vjagadish: Can remove the String.format boiler plate and use log4j formatters instead

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -117,41 +132,50 @@ public void join(String version, String participantId) {
...
    */
   public void expire(String version) {
-    zkUtils.writeData(keyBuilder.getBarrierStatePath(version), State.TIMED_OUT);
-
+    String barrierStatePath = keyBuilder.getBarrierStatePath(version);
+    State barrierState = zkUtils.getZkClient().readData(barrierStatePath);
+    if (Objects.equals(barrierState, State.NEW)) {
+      LOG.info(String.format("Expiring the barrier version: %s. Marking the barrier state: %s as %s.", version, barrierStatePath, State.TIMED_OUT));
+      zkUtils.writeData(keyBuilder.getBarrierStatePath(version), State.TIMED_OUT);
+    }
```
vjagadish: Add logging to debug the else code-path.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -178,14 +202,15 @@ public void handleDataChange(String dataPath, Object data) {
       if (notAValidEvent())
         return;
 
-      zkUtils.unsubscribeDataChanges(barrierStatePath, this);
-      barrierListenerOptional.ifPresent(
-          zkBarrierListener -> zkBarrierListener.onBarrierStateChanged(barrierVersion, (State) data));
+      State barrierState = (State) data;
+      if (!Objects.equals(barrierState, State.NEW)) {
```
vjagadish: Should be explicit in the state-checks here. The states that we are looking for are terminal states: DONE, TIMED_OUT. This will also make code immune to refactors later - for instance, if someone adds another state to the barrier state machine we don't want this code-path to be triggered and prior subscriptions being cancelled.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -117,41 +132,50 @@ public void join(String version, String participantId) {
...
-        zkUtils.writeData(barrierDonePath, State.DONE); // this will trigger notifications
-        zkUtils.unsubscribeChildChanges(barrierDonePath, this);
+      if (participantIds.size() == expectedParticipantIds.size() && CollectionUtils.containsAll(participantIds, expectedParticipantIds)) {
+        String barrierStatePath = keyBuilder.getBarrierStatePath(barrierVersion);
+        State barrierState = zkUtils.getZkClient().readData(barrierStatePath);
+        if (Objects.equals(barrierState, State.NEW)) {
+          LOG.info(String.format("Expected participants has joined the barrier version: %s. Marking the barrier state: %s as %s.", barrierVersion, barrierStatePath, State.DONE));
+          zkUtils.writeData(barrierStatePath, State.DONE); // this will trigger notifications
+        }
+        LOG.info("Unsubscribing child changes for the path: {}.", barrierParticipantPath);
```
vjagadish: LOG the barrier version wherever applicable. It'd be great for debuggability if we do a 
`grep "barrier version: xxx"` and we can view all state transitions in the barrier.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -117,41 +132,50 @@ public void join(String version, String participantId) {
    * @param version Version associated with the Barrier
    */
   public void expire(String version) {
-    zkUtils.writeData(keyBuilder.getBarrierStatePath(version), State.TIMED_OUT);
-
+    String barrierStatePath = keyBuilder.getBarrierStatePath(version);
+    State barrierState = zkUtils.getZkClient().readData(barrierStatePath);
+    if (Objects.equals(barrierState, State.NEW)) {
+      LOG.info(String.format("Expiring the barrier version: %s. Marking the barrier state: %s as %s.", version, barrierStatePath, State.TIMED_OUT));
+      zkUtils.writeData(keyBuilder.getBarrierStatePath(version), State.TIMED_OUT);
```
vjagadish: Not necessarily in this PR: For debugability, it will be helpful to write strings to zookeeper instead of objects. Often, open-source tools like _zooinspector_ and internal UIs like _reflection_ work better with UTF-8 encoded strings.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -61,9 +61,19 @@
   private final Optional<ZkBarrierListener> barrierListenerOptional;
 
   public enum State {
```
vjagadish: Document all valid state-transitions allowed in the state-machine and the events that trigger them. 
Call out that currently, the only allowed transitions are from NEW->DONE and NEW->TIMED_OUT.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -178,14 +202,15 @@ public void handleDataChange(String dataPath, Object data) {
...
+      if (!Objects.equals(barrierState, State.NEW)) {
+        zkUtils.unsubscribeDataChanges(barrierStatePath, this);
+        barrierListenerOptional.ifPresent(zkBarrierListener -> zkBarrierListener.onBarrierStateChanged(barrierVersion, (State) data));
+      }
     }
 
     @Override
-    public void handleDataDeleted(String dataPath)
-        throws Exception {
+    public void handleDataDeleted(String dataPath) {
```
vjagadish: Add a `getState()` API on the barrier that tests can assert when validating expected transitions. This should also clean-up tests a bit since it's un-necessary to expose the underlying ZkClient.

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
```diff
@@ -18,205 +18,221 @@
...
 
-      @Override
-      public void onBarrierStateChanged(String version, ZkBarrierForVersionUpgrade.State state) {
-        if (state.equals(ZkBarrierForVersionUpgrade.State.DONE)) {
-          latch.countDown();
-          stateChangedCalled.incrementAndGet();
-        }
-      }
+    private final CountDownLatch stateChangedLatch;
+    private final AtomicInteger stateChangedCounter;
```
vjagadish: Do we need the stateChangedCounter in-addition to the stateChangedLatch? It looks like stateChangedLatch.countDown is exactly equivalent to "stateChangedCounter == 1"

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
```diff
@@ -18,205 +18,221 @@
...
-    Assert.assertEquals(2, stateChangedCalled.get());
+    assertTrue("Barrier Timeout test failed to complete within test timeout.", result);
+
+    List<String> children = zkUtils.getZkClient().getChildren(barrierId + "/barrier_1/barrier_participants");
+    State barrierState = zkUtils.getZkClient().readData(barrierId + "/barrier_1/barrier_state");
+    assertEquals(State.TIMED_OUT, barrierState);
+    assertNotNull(children);
+    assertEquals("Unexpected barrier state. Didn't find two processors.", 2, children.size());
+    assertEquals("Unexpected barrier state. Didn't find the expected members.", ImmutableList.of("p1", "p2"), children);
+    assertEquals(2, timeoutStateChangeCalled.get());
```
vjagadish: Could this be flaky in this scenario?

- thread-1
await on the latch
check counter==2

- zk-client thread
listener.onEvent() is called for processor-1
processor1 increments counter
processor1 counts down the latch
listener.onEvent() is called for processor-2
processor2 increments the counter

Could the counter-check in thread-1 succeed/fail depending on whether the ZkClient thread got to deliver the event for processor2?

Would be simpler to use a single latch with count = 2 or use separate listener instances

samza-core/src/test/java/org/apache/samza/zk/TestZkBarrierForVersionUpgrade.java
```diff
@@ -18,205 +18,221 @@
...
-    Assert.assertTrue("Barrier Timeout test failed to complete within test timeout.", result);
+    assertTrue("Barrier Timeout test failed to complete within test timeout.", result);
+
+    processor1Barrier.expire(BARRIER_VERSION);
+
+    State barrierState = zkUtils.getZkClient().readData(barrierId + "/barrier_1/barrier_state");
+    assertEquals(State.DONE, barrierState);
+  }
+
+  @Test
```
vjagadish: Should we add tests to ensure that DONE->TIMED_OUT and TIMED_OUT -> DONE have no effect?

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -117,41 +132,50 @@ public void join(String version, String participantId) {
...
-        String barrierDonePath = keyBuilder.getBarrierStatePath(barrierVersion);
-        LOG.info("Writing BARRIER DONE to " + barrierDonePath);
-        zkUtils.writeData(barrierDonePath, State.DONE); // this will trigger notifications
-        zkUtils.unsubscribeChildChanges(barrierDonePath, this);
+      if (participantIds.size() == expectedParticipantIds.size() && CollectionUtils.containsAll(participantIds, expectedParticipantIds)) {
+        String barrierStatePath = keyBuilder.getBarrierStatePath(barrierVersion);
+        State barrierState = zkUtils.getZkClient().readData(barrierStatePath);
+        if (Objects.equals(barrierState, State.NEW)) {
+          LOG.info(String.format("Expected participants has joined the barrier version: %s. Marking the barrier state: %s as %s.", barrierVersion, barrierStatePath, State.DONE));
+          zkUtils.writeData(barrierStatePath, State.DONE); // this will trigger notifications
```
vjagadish: Not necessarily in this PR but would be good to think about what would happen if the barrier state changed in the time we do this check and act on it.

vjagadish: exactly what I meant, thanks!

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
shanthoosh: I agree, we can eliminate redundancy in `getBarrierState()` calls in tests. I will add a method to test class itself,  we don't need to add it to `ZkBarrierForVersionUpgrade` class(other than tests it won't be used elsewhere).

PR #491, opened 2018-04-27
Samza-1563: Make RocksDB store base directory configurable

bharathkk:
@xinyuiscool ^^

### DIFF
--- a/docs/learn/documentation/versioned/jobs/configuration-table.html
+++ b/docs/learn/documentation/versioned/jobs/configuration-table.html
@@ -476,11 +476,11 @@
                 </tr>
 
                 <tr>
-                    <td class="property" id="job.logged.store.base.dir">job.logged.store.base.dir</td>
+                    <td class="property" id="job.store.path">job.store.path</td>
                     <td class="default"></td>
                     <td class="description">
-                        The base directory for state stores used by Samza application. Other ways to configure the base directory is by setting environment variable <i>LOGGED_STORE_BASE_DIR</i>.
-                        <br><b>Note:</b> Configuration through environment variable takes higher precedence compared to <i>job.logged.store.base.dire</i>.
+                        The base directory for state stores used by Samza application. Another way to configure the base directory is by setting environment variable <i>LOGGED_STORE_BASE_DIR</i>.
+                        <br><b>Note:</b> Configuration through environment variable takes precedence over <i>job.store.path</i>.
                     </td>
                 </tr>
 
--- a/docs/learn/documentation/versioned/jobs/configuration-table.html
+++ b/docs/learn/documentation/versioned/jobs/configuration-table.html
@@ -474,6 +474,16 @@
                         These coordination utils are currently used for intermediate stream creation.
                     </td>
                 </tr>
+
+                <tr>
+                    <td class="property" id="job.logged.store.base.dir">job.logged.store.base.dir</td>
+                    <td class="default"></td>
+                    <td class="description">
+                        The base directory for state stores used by Samza application. Other ways to configure the base directory is by setting environment variable <i>LOGGED_STORE_BASE_DIR</i>.
+                        <br><b>Note:</b> Configuration through environment variable takes higher precedence compared to <i>job.logged.store.base.dire</i>.
+                    </td>
+                </tr>
+
                 <tr>
                                               <!-- change link to StandAlone design/tutorial doc. SAMZA-1299 -->
                 <th colspan="3" class="section" id="ZkBasedJobCoordination"><a href="../index.html">Zookeeper-based job configuration</a></th>
--- a/docs/learn/documentation/versioned/jobs/configuration-table.html
+++ b/docs/learn/documentation/versioned/jobs/configuration-table.html
@@ -472,6 +472,7 @@
                             <dt><code>org.apache.samza.coordinator.AzureCoordinationUtilsFactory</code></dt>
                             <dd>Azure based coordination utils.</dd>
                         These coordination utils are currently used for intermediate stream creation.
+                        </dl>
                     </td>
                 </tr>
 
@@ -484,11 +485,10 @@
                         </dl>
                     </td>
                     <td class="description">
-                        The base directory for changelog stores used by Samza application. Another way to configure the base directory is by setting environment variable <i>LOGGED_STORE_BASE_DIR</i>.
-                        <br>Samza does not have any default clean up policy for changelog store directories. Typically, changelog stores are persisted across application attempts for host affinity feature to work.
-                        Although, application may still need some clean up policy in the event of store schema changes, etc to prevent stale data in clusters.
+                        The base directory for changelog stores used by Samza application. It allows users to configure the base directory at a job level. Another way to configure the base directory is by setting environment variable <i>LOGGED_STORE_BASE_DIR</i>.
+                        <b>Note:</b> The environment variable takes precedence over <i>job.logged.store.path</i>.
 
-                        <br><b>Note:</b> Environment variable <i>LOGGED_STORE_BASE_DIR</i> takes precedence over <i>job.logged.store.path</i>
+                        <br>By opting in, users are responsible for cleaning up the store directories if necessary. For jobs with host-affinity, make sure the changelog stores are persisted across application restarts to minimize restoration time.
                     </td>
                 </tr>
 
@@ -501,13 +501,12 @@
                         </dl>
                     </td>
                     <td class="description">
-                        The base directory for non-changelog stores used by Samza application. Typically, non-changelog stores get cleaned up across application restarts.
+                        The base directory for non-changelog stores used by Samza application. It allows users to configure the base directory at a job level.
 
-                        Samza does not have any default clean up policy for non changelog stores.
-                        For Yarn, the current behaviour without the configuration is to create non-changelog store directories within the yarn container directory. The data gets cleaned up
-                        as part of yarn container clean up across application attempts. With this configuration, application needs to be mindful about cleaning up store data upon completion.
+                        <br>For YARN, the current behaviour without the configuration is to create non-changelog store directories in CWD which happens to be the yarn container directory. As part of NodeManager's deletion service, the data gets cleaned up
+                        periodically, which is controlled by the YARN config <i>yarn.nodemanager.delete.debug-delay-sec</i>.
 
-                        <br><b>Note:</b>Lack of clean up policy might affect application behaviour across restarts
+                        <br>By opting in, users are responsible for cleaning up the store directories if necessary.
                     </td>
                 </tr>
 
--- a/docs/learn/documentation/versioned/jobs/configuration-table.html
+++ b/docs/learn/documentation/versioned/jobs/configuration-table.html
@@ -477,36 +477,31 @@
                 </tr>
 
                 <tr>
-                    <td class="property" id="job.logged.store.path">job.logged.store.path</td>
+                    <td class="property" id="job.logged.store.base.dir">job.logged.store.base.dir</td>
                     <td class="default">
-                        <dl>
-                            <dt><i>user.dir</i> environment property</dt>
-                            <dt>If <i>user.dir</i> is not defined, <i>CWD</i> is used</dt>
-                        </dl>
+                        <i>user.dir</i> environment property if set, else current working directory of the process
                     </td>
                     <td class="description">
-                        The base directory for changelog stores used by Samza application. It allows users to configure the base directory at a job level. Another way to configure the base directory is by setting environment variable <i>LOGGED_STORE_BASE_DIR</i>.
-                        <b>Note:</b> The environment variable takes precedence over <i>job.logged.store.path</i>.
+                        The base directory for changelog stores used by Samza application. Another way to configure the base directory is by setting environment variable <i>LOGGED_STORE_BASE_DIR</i>.
+                        <b>Note:</b> The environment variable takes precedence over <i>job.logged.store.base.dir</i>.
 
-                        <br>By opting in, users are responsible for cleaning up the store directories if necessary. For jobs with host-affinity, make sure the changelog stores are persisted across application restarts to minimize restoration time.
+                        <br>By opting in, users are responsible for cleaning up the store directories if necessary. Jobs using host affinity should ensure that the stores are persisted across application/container restarts.
+                        This means that the location and cleanup of this directory should be separate from the container lifecycle and resource cleanup.
                     </td>
                 </tr>
 
                 <tr>
-                    <td class="property" id="job.non-logged.store.path">job.non-logged.store.path</td>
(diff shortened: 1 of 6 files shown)


docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -474,6 +474,16 @@ <h1>Samza Configuration Reference</h1>
                         These coordination utils are currently used for intermediate stream creation.
                     </td>
                 </tr>
+
+                <tr>
+                    <td class="property" id="job.logged.store.base.dir">job.logged.store.base.dir</td>
```
prateekm: Should clarify why this is *logged.*store.base.dir, since non-changelogged stores are stored in YARN container directory.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -474,6 +474,16 @@ <h1>Samza Configuration Reference</h1>
...
+                    <td class="default"></td>
+                    <td class="description">
+                        The base directory for state stores used by Samza application. Other ways to configure the base directory is by setting environment variable <i>LOGGED_STORE_BASE_DIR</i>.
+                        <br><b>Note:</b> Configuration through environment variable takes higher precedence compared to <i>job.logged.store.base.dire</i>.
```
prateekm: "takes precedence over"

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -474,6 +474,16 @@ <h1>Samza Configuration Reference</h1>
                         These coordination utils are currently used for intermediate stream creation.
                     </td>
                 </tr>
+
+                <tr>
+                    <td class="property" id="job.logged.store.base.dir">job.logged.store.base.dir</td>
+                    <td class="default"></td>
+                    <td class="description">
+                        The base directory for state stores used by Samza application. Other ways to configure the base directory is by setting environment variable <i>LOGGED_STORE_BASE_DIR</i>.
```
prateekm: "Another way"

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -474,6 +474,16 @@ <h1>Samza Configuration Reference</h1>
                         These coordination utils are currently used for intermediate stream creation.
                     </td>
                 </tr>
+
+                <tr>
+                    <td class="property" id="job.logged.store.base.dir">job.logged.store.base.dir</td>
+                    <td class="default"></td>
```
prateekm: What's the default value?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -469,18 +469,25 @@ object SamzaContainer extends Logging {
...
         loggedStorageBaseDir = new File(System.getenv(ShellCommandConfig.ENV_LOGGED_STORE_BASE_DIR)
           + File.separator + jobNameAndId._1 + "-" + jobNameAndId._2)
       } else {
-        warn("No override was provided for logged store base directory. This disables local state re-use on " +
-          "application restart. If you want to enable this feature, set LOGGED_STORE_BASE_DIR as an environment " +
-          "variable in all machines running the Samza container")
-        loggedStorageBaseDir = defaultStoreBaseDir
+        config.getLoggedStoreBaseDir match {
+          case Some(storeBaseDir) =>
+            info("Using user provided value for job state store base directory")
```
prateekm: Log the path. Remove "user provided".

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -469,18 +469,25 @@ object SamzaContainer extends Logging {
...
+            loggedStorageBaseDir = new File(storeBaseDir + File.separator + jobNameAndId._1 + "-" + jobNameAndId._2)
+          case None =>
+            warn("No override was provided for logged store base directory. This disables local state re-use on " +
```
prateekm: Update comment.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -81,6 +81,9 @@ object JobConfig {
   val PROCESSOR_ID = "processor.id"
   val PROCESSOR_LIST = "processor.list"
 
+  // job level configuration for base directory for state stores
```
prateekm: Comment unnecessary, config name should be self-explanatory.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -469,18 +469,25 @@ object SamzaContainer extends Logging {
...
-          config.getJobId.getOrElse("1")
-        )
         loggedStorageBaseDir = new File(System.getenv(ShellCommandConfig.ENV_LOGGED_STORE_BASE_DIR)
           + File.separator + jobNameAndId._1 + "-" + jobNameAndId._2)
       } else {
-        warn("No override was provided for logged store base directory. This disables local state re-use on " +
-          "application restart. If you want to enable this feature, set LOGGED_STORE_BASE_DIR as an environment " +
-          "variable in all machines running the Samza container")
-        loggedStorageBaseDir = defaultStoreBaseDir
+        config.getLoggedStoreBaseDir match {
```
xinyuiscool: let's do a "else if .. else" instead of else { .. match case..}. The flow is simpler to understand. Thanks.

docs/learn/documentation/versioned/jobs/configuration-table.html
xinyuiscool: job.store.path

docs/learn/documentation/versioned/jobs/configuration-table.html
bharathkk: fixed

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
bharathkk: fixed

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
bharathkk: fixed

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
bharathkk: fixed

docs/learn/documentation/versioned/jobs/configuration-table.html
bharathkk: fixed

docs/learn/documentation/versioned/jobs/configuration-table.html
bharathkk: fixed

docs/learn/documentation/versioned/jobs/configuration-table.html
bharathkk: modified the configuration name

docs/learn/documentation/versioned/jobs/configuration-table.html
bharathkk: there is no default value.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -474,6 +474,16 @@ <h1>Samza Configuration Reference</h1>
                         These coordination utils are currently used for intermediate stream creation.
                     </td>
                 </tr>
+
+                <tr>
+                    <td class="property" id="job.store.path">job.store.path</td>
```
prateekm: Non-changelogged stores are stored in YARN container's workspace so that they can be clean up when the container is unused. Changelogged stores are kept separately and need something like state store GC to be cleaned up. Should this config be "job.logged.store.base.dir" or is the intention to move them all in the same place? If the latter, why?

shanthoosh: There's existing configuration `job.local.store.dir`, which is used by LocalStoreMonitor as a base directory for cleaning up state store.
I think we should unify those two configurations and not have duplicate configurations. Either delete the existing one or use that configuration instead. Update the relevant documentation .

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -506,7 +515,11 @@ object SamzaContainer extends Logging {
               case _ => null
             }
 
-            val storeDir = if (changeLogSystemStreamPartition != null) {
+            // We want to use the job store path from the config for standalone deployment model for both changelog
+            // enabled & disabled stores. We preserve the existing behaviour in yarn where non-changelog stores
+            // use "user.dir" for store base directory even if the environment variable LOGGED_STORE_BASE_DIR is set.
+            // TODO: Identify the impact of using logged store directories for non-changelog stores.
```
shanthoosh: > Identify the impact of using logged store directories for non-changelog stores.

Existing behavior for non changelogged stores is that whenever the samza container restarts, we will begin with the clean slate. That will be violated if non-changelog stores are put in change-log directory and SamzaContainer starts again in the same host after a restart. 

Typical scenario I've seen is to use rocksDb as a cache in front of remote store.

xinyuiscool:
@bharathkk : the updated patch looks much cleaner and now we have a way to deprecate the env var in the future. The changes look good to me.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -474,6 +474,43 @@ <h1>Samza Configuration Reference</h1>
...
+                    <td class="default">
+                        <dl>
+                            <dt><i>user.dir</i> environment property</dt>
+                            <dt>If <i>user.dir</i> is not defined, <i>CWD</i> is used</dt>
+                        </dl>
+                    </td>
+                    <td class="description">
+                        The base directory for non-changelog stores used by Samza application. Typically, non-changelog stores get cleaned up across application restarts.
+
+                        Samza does not have any default clean up policy for non changelog stores.
```
prateekm: Conflicts with the previous statement. Can you clarify?

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -474,6 +474,43 @@ <h1>Samza Configuration Reference</h1>
...
+                        For Yarn, the current behaviour without the configuration is to create non-changelog store directories within the yarn container directory. The data gets cleaned up
```
prateekm: Can you clarify what YARN (also, uppercase) configuration controls this?

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -474,6 +474,43 @@ <h1>Samza Configuration Reference</h1>
...
+                        as part of yarn container clean up across application attempts. With this configuration, application needs to be mindful about cleaning up store data upon completion.
+
+                        <br><b>Note:</b>Lack of clean up policy might affect application behaviour across restarts
```
prateekm: This sounds pretty vague. Clarify or remove.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -81,6 +81,14 @@ object JobConfig {
   val PROCESSOR_ID = "processor.id"
   val PROCESSOR_LIST = "processor.list"
 
+  // Represents the store path for non-changelog stores which are persisted in the disk and potentially cleaned up
```
prateekm: Remove "persisted in the disk", it's unrelated to changelogging. Also, what does potentially mean here?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -76,6 +76,49 @@ object SamzaContainer extends Logging {
         classOf[JobModel])
   }
 
+  // Ideally, we want this to reside in TaskStorageManager since container should not be in the business of
```
prateekm: If this is not an actionable, lets remove this comment. Otherwise let's create a ticket and make a TODO.

samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
```diff
@@ -45,18 +45,18 @@ object TaskStorageManager {
...
-  changeLogSystemStreams: Map[String, SystemStream] = Map(),
-  changeLogStreamPartitions: Int,
-  streamMetadataCache: StreamMetadataCache,
-  storeBaseDir: File = new File(System.getProperty("user.dir"), "state"),
-  loggedStoreBaseDir: File = new File(System.getProperty("user.dir"), "state"),
-  partition: Partition,
-  systemAdmins: SystemAdmins,
-  changeLogDeleteRetentionsInMs: Map[String, Long],
-  clock: Clock) extends Logging {
+                          taskName: TaskName,
```
prateekm: Revert formatting changes.

bharathkk:
@prateekm 
Please take a look at the latest changes. I have addressed the documentation from the following standpoint.
1. What is the configuration for?
2. Why do we need it?
3. Other ways of configuring and precedence, if applicable?
4. What is the current behavior?
5. What users need to do as part of opting in?

Hopefully, that captures most of what we want to convey to the users.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -472,8 +472,44 @@ <h1>Samza Configuration Reference</h1>
...
+                <tr>
+                    <td class="property" id="job.non-logged.store.path">job.non-logged.store.path</td>
+                    <td class="default">
+                        <dl>
+                            <dt><i>user.dir</i> environment property</dt>
+                            <dt>If <i>user.dir</i> is not defined, <i>CWD</i> is used</dt>
+                        </dl>
+                    </td>
+                    <td class="description">
+                        The base directory for non-changelog stores used by Samza application. It allows users to configure the base directory at a job level.
```
prateekm: Second sentence seems redundant.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -472,8 +472,44 @@ <h1>Samza Configuration Reference</h1>
...
+
+                        <br>For YARN, the current behaviour without the configuration is to create non-changelog store directories in CWD which happens to be the yarn container directory. As part of NodeManager's deletion service, the data gets cleaned up
+                        periodically, which is controlled by the YARN config <i>yarn.nodemanager.delete.debug-delay-sec</i>.
+
+                        <br>By opting in, users are responsible for cleaning up the store directories if necessary.
```
prateekm: Maybe: "When set to something other than the YARN container directory, or running in a different environment, these stores should be cleaned up periodically."

PR #492, opened 2018-04-27
SAMZA-1691: Support get iterable from KeyValueStore

xinyuiscool:
Right now for KeyValueStore we have a range query to return an iterator. For usage in BEAM, we need a iterable which will 1) create the snapshot when called, and 2) create an iterator when needed. Add the iterate() function in KeyValueStore to support it. It's implemented as follows:

1) for rocksDb, it will create the iterator when it's called, which will has a snapshot of the elements. Then every time when the iterator is needed, we will seek the iterator from beginning;

2) for inMemoryDb, it will create the snapshot submap when iterate() is called. The submap is an iterable and it can return a new iterator when needed.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/storage/kv/KeyValueIterable.java
+++ b/samza-api/src/main/java/org/apache/samza/storage/kv/KeyValueIterable.java
@@ -1,0 +1,24 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.storage.kv;
+
+public interface KeyValueIterable<K, V> extends Iterable<Entry<K, V>> {
+  KeyValueIterator<K, V> iterator();
+}
--- a/samza-api/src/main/java/org/apache/samza/storage/kv/KeyValueStore.java
+++ b/samza-api/src/main/java/org/apache/samza/storage/kv/KeyValueStore.java
@@ -112,6 +112,8 @@ public interface KeyValueStore<K, V> {
 
   /**
    * Returns an iterable for a sorted range of entries specified by [{@code from}, {@code to}).
+   * Note that we snapshot the iterator when the iterable is created from this function, and
+   * the iteration results is guaranteed to reflect the snapshot if only one iterator is in use at a time.
    *
    * @param from the key specifying the low endpoint (inclusive) of the keys in the returned range.
    * @param to the key specifying the high endpoint (exclusive) of the keys in the returned range.
--- a/samza-api/src/main/java/org/apache/samza/storage/kv/KeyValueStore.java
+++ b/samza-api/src/main/java/org/apache/samza/storage/kv/KeyValueStore.java
@@ -111,6 +111,18 @@ public interface KeyValueStore<K, V> {
   KeyValueIterator<K, V> range(K from, K to);
 
   /**
+   * Returns an iterable for a sorted range of entries specified by [{@code from}, {@code to}).
+   *
+   * @param from the key specifying the low endpoint (inclusive) of the keys in the returned range.
+   * @param to the key specifying the high endpoint (exclusive) of the keys in the returned range.
+   * @return an iterable for the specified key range.
+   * @throws NullPointerException if null is used for {@code from} or {@code to}.
+   */
+  default KeyValueIterable<K, V> iterate(K from, K to) {
+    return () -> range(from, to);
+  }
+
+  /**
    * Returns an iterator for all entries in this key-value store.
    *
    * <p><b>API Note:</b> The returned iterator MUST be closed after use.</p>
--- a/samza-kv-inmemory/src/test/java/org/apache/samza/storage/kv/inmemory/TestInMemoryKeyValueStore.java
+++ b/samza-kv-inmemory/src/test/java/org/apache/samza/storage/kv/inmemory/TestInMemoryKeyValueStore.java
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.storage.kv.inmemory;
+
+import com.google.common.collect.Iterators;
+import com.google.common.primitives.Ints;
+import org.apache.samza.metrics.MetricsRegistryMap;
+import org.apache.samza.storage.kv.Entry;
+import org.apache.samza.storage.kv.KeyValueIterable;
+import org.apache.samza.storage.kv.KeyValueStoreMetrics;
+import org.junit.Test;
+
+import java.io.ByteArrayOutputStream;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class TestInMemoryKeyValueStore {
+  @Test
+  public void testIterate() throws Exception {
+    InMemoryKeyValueStore store = new InMemoryKeyValueStore(
+        new KeyValueStoreMetrics("testInMemory", new MetricsRegistryMap()));
+    ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+    String prefix = "prefix";
+    for(int i = 0; i < 100; i++) {
+      store.put(genKey(outputStream, prefix, i), genValue());
+    }
+
+    byte[] firstKey = genKey(outputStream, prefix, 0);
+    byte[] lastKey = genKey(outputStream, prefix, 100);
+    KeyValueIterable<byte[], byte[]> iterable = store.iterate(firstKey, lastKey);
+    // Make sure the cached Iterable won't change when new elements are added
+    store.put(genKey(outputStream, prefix, 200), genValue());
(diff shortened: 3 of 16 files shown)


samza-api/src/main/java/org/apache/samza/storage/kv/KeyValueIterable.java
```diff
@@ -0,0 +1,24 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.storage.kv;
+
+public interface KeyValueIterable<K, V> extends Iterable<Entry<K, V>> {
```
cameronlee314: Does this need a "close" method so that you can close the internal iterator (e.g. tell internal RocksDB iterator to clean up snapshot)?

xinyuiscool: Good question. I was thinking about this when I did the change. The reason that I didn't add the close method in KeyValueIterable is that the KeyValueIterator itself already has a close() method in the API, so it should be the responsibility of KeyValueIterator to close it (might even be confusing if we also have another close() here). Think of this as a factory class that creates the iterator, but close() should be on the iterator itself.

cameronlee314: I agree that KeyValueIterator should have the logic for closing the RocksDB snapshot. However, someone needs to call KeyValueIterator.close when it is done being used. If the KeyValueIterable uses the KeyValueIterator, then KeyValueIterable needs to provide a way for someone to call KeyValueIterator.close, or KeyValueIterable needs to call KeyValueIterator.close automatically.

xinyuiscool: Oh, I believe when the users will get hold of KeyValueIterable instances when they call getIterator() from the iterable. So they can close the iterators by themselves.

samza-kv-inmemory/src/test/java/org/apache/samza/storage/kv/inmemory/TestInMemoryKeyValueStore.java
```diff
@@ -0,0 +1,81 @@
...
+      int key = Ints.fromByteArray(Arrays.copyOfRange(entry.getKey(), prefix.getBytes().length, entry.getKey().length));
+      keys.add(key);
+    }
+    assertEquals(keys, IntStream.rangeClosed(0, 99).boxed().collect(Collectors.toList()));
+
+    outputStream.close();
+    store.close();
+  }
+
+  private byte[] genKey(ByteArrayOutputStream outputStream, String prefix, int i) throws Exception {
```
sborya: why do you need to pass outputStream to this method. It is just used for generation of byte arr, isn't it?

samza-kv-rocksdb/src/test/java/org/apache/samza/storage/kv/TestRocksDbKeyValueStoreJava.java
```diff
@@ -0,0 +1,90 @@
...
+      int key = Ints.fromByteArray(Arrays.copyOfRange(entry.getKey(), prefix.getBytes().length, entry.getKey().length));
+      keys.add(key);
+    }
+    assertEquals(keys, IntStream.rangeClosed(0, 99).boxed().collect(Collectors.toList()));
+
+    outputStream.close();
+    store.close();
+  }
+
+  private byte[] genKey(ByteArrayOutputStream outputStream, String prefix, int i) throws Exception {
```
sborya: same as in the other test.

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -91,6 +92,13 @@ class AccessLoggedStore[K, V](
     store.all()
   }
 
+  def iterate(from: K, to: K): KeyValueIterable[K, V] = {
+    val list : util.ArrayList[K] = new util.ArrayList[K]()
+    list.add(from)
+    list.add(to)
+    logAccess(DBOperation.ITERATE, serializeKeys(list), store.iterate(from, to))
```
sborya: why do you need this list? is it the name of the iterator?

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
```diff
@@ -160,4 +160,11 @@ class KeyValueStorageEngine[K, V](
   }
 
   override def getStoreProperties: StoreProperties = storeProperties
+
+  override def iterate(from: K, to: K): KeyValueIterable[K, V] = {
+    updateTimer(metrics.iterateNs) {
```
sborya: not sure how this works..

samza-kv/src/main/scala/org/apache/samza/storage/kv/LoggedStore.scala
```diff
@@ -114,4 +114,7 @@ class LoggedStore[K, V](
     store.close
   }
 
+  override def iterate(from: K, to: K): KeyValueIterable[K, V] = {
+    store.iterate(from, to)
```
sborya: where this iterate comes from?

samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
```diff
@@ -203,6 +204,18 @@ class RocksDbKeyValueStore(
     new RocksDbIterator(iter)
   }
 
+  def iterate(from: Array[Byte], to: Array[Byte]): KeyValueIterable[Array[Byte], Array[Byte]] = {
+    //snapshot the iterator
+    val iter : RocksDbRangeIterator = range(from, to).asInstanceOf[RocksDbRangeIterator]
+    new KeyValueIterable[Array[Byte], Array[Byte]] {
+      def iterator(): KeyValueIterator[Array[Byte], Array[Byte]] = {
+        // reset to the beginning
+        iter.seek(from)
```
vjagadish: I do have a concern with re-purposing the Iterator returned by `range()` and using it to mint an `Iterable` here. For example, Would thread-2 calling `getIterator` end-up resetting the cursor for thread-1? Please let me know.

KeyValueIterable iterable = store.iterate();
**thread-1:** 
 Iterator i1 = iterable.getIterator();
 iterate using i1; 
**thread-2:**
 Iterator i2 = iterable.getIterator();
 iterate using i2;

Alternately, RocksDb has a top-level notion of a 'snapshot()'. Does the JNI binding for RocksDb expose it? IIUC, it seems to be achieving exactly what we want.

xinyuiscool: NO multithreading in a task in this use case.

samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
```diff
@@ -203,6 +204,18 @@ class RocksDbKeyValueStore(
     new RocksDbIterator(iter)
   }
 
+  def iterate(from: Array[Byte], to: Array[Byte]): KeyValueIterable[Array[Byte], Array[Byte]] = {
+    //snapshot the iterator
+    val iter : RocksDbRangeIterator = range(from, to).asInstanceOf[RocksDbRangeIterator]
```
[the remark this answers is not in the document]
xinyuiscool: sorry, I don't know what you mean. An iterable is a way to create iterator as Java doc here: https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html. I believe it's the responsibility of the holder of the iterator who should do the close(). Iterable is the factory, why is it concerned about the compaction and clearnup?

samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
xinyuiscool: I didn't know about snapshot. Right now when I created the iterator, it does the snapshot at that point and it fits the purpose (single thread rocksDb access). To use snapshot, I think I need to do some benchmarking to understand the performance aspect of it. I am not comfortable using it directly.

samza-kv/src/main/scala/org/apache/samza/storage/kv/LoggedStore.scala
xinyuiscool: I added it to the general interface of KeyValueIterable.

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
xinyuiscool: Scala magic :). It's called currying, I finally remembered the name of it :)

samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
xinyuiscool: Oh, I believe this is for logging every access of the store to a stream.

samza-kv-inmemory/src/test/java/org/apache/samza/storage/kv/inmemory/TestInMemoryKeyValueStore.java
xinyuiscool: it's easier for me to reuse the outputStream (and the buffer with it), right :)?

samza-api/src/main/java/org/apache/samza/storage/kv/KeyValueStore.java
```diff
@@ -110,6 +110,18 @@ default void deleteAll(List<K> keys) {
...
 
+  /**
+   * Returns an iterable for a sorted range of entries specified by [{@code from}, {@code to}).
+   *
+   * @param from the key specifying the low endpoint (inclusive) of the keys in the returned range.
+   * @param to the key specifying the high endpoint (exclusive) of the keys in the returned range.
+   * @return an iterable for the specified key range.
+   * @throws NullPointerException if null is used for {@code from} or {@code to}.
+   */
+  default KeyValueIterable<K, V> iterate(K from, K to) {
```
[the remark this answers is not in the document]
xinyuiscool: Iterable is a factory class which creates iterator, that's a java standard so I don't want to mess with it. I don't think it's a good idea at all to call it snapshot which implements a java iterable interface. Since it's not returning the snapshot, I don't want to call this function snapshot either. I am willing to take other names if iterate() is not clear enough.

samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
xinyuiscool: Looks like we should be able to make this logic work for normal use cases, basically the logic should be:

if (!iter.isClosed() || iter.next() == false) {
    // if the cached iterator already goes to the end, we can reset it for reuse
   iter.seek(from)
   iter
} else {
   // in case the cached iter is not done or already closed, create a new one.
   range(from, to)
}

Looks like this solution should be okey. As a side note, Flink does similar things for their entries() iterable: https://github.com/apache/flink/blob/master/flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/RocksDBMapState.java, line 161.

vjagadish:
To close the loop on this PR. Had an offline discussion with @xinyuiscool, @nickpan47 and @prateekm and we agreed on the following:
- The need for the BEAM runner is to return an `Iterable` whose iterators are not impacted by mutations to the underlying store.
- We will rename the API to `snapshot()` to actually indicate what its semantics is.
- We will return an immutable, point-in-time snapshot of the store. We will leverage the RocksDb's `db.snapshot()` instead of re-purposing `db.iterator()` since it seems to be achieving our needs. This will ensure that the behavior is consistent across in-memory and rocks-db stores. The follow-up PR is in https://github.com/apache/samza/pull/510/files .

PR #493, opened 2018-04-30
SAMZA-1647: Fix NPE in onJobModelExpired handler in StreamProcessor.

shanthoosh:
**Changes:**
* Switching to using explicit lock in StreamProcessor to make things simpler on state updation.
* Switch from using synchronized in ZkJobCoordinator to prevent any potential deadlocks
between two threads (where one thread holds the StreamProcessor and other thread has ZkJobCoordinator lock).
* Misc cleanups in StreamProcessor: Remove volatile qualifiers from state variables in StreamProcessor. Remove reinstantiating the
executorService in onNewJobModel.
* ZkJobCoordinator cleanups: Make some state variables as immutable.

**NOTE**: The classes in which these changes were made were aynonymous inner classes,
so to add proper unit tests we need to do big haul of refactor.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -183,7 +183,6 @@ public class StreamProcessor {
         try {
           LOGGER.info("Shutting down the container: {} of stream processor: {}.", container, processorId);
           container.shutdown();
-          LOGGER.info("Waiting {} milliseconds for the container: {} to shutdown.", taskShutdownMs, container);
           containerShutdownInvoked = true;
         } catch (Exception exception) {
           LOGGER.error(String.format("Ignoring the exception during the shutdown of container: %s.", container), exception);
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -75,7 +75,7 @@ public class StreamProcessor {
 
   // Latch used to synchronize between the JobCoordinator thread and the container thread, when the container is
   // stopped due to re-balancing
-  CountDownLatch jcContainerShutdownLatch;
+  volatile CountDownLatch jcContainerShutdownLatch;
 
   @VisibleForTesting
   JobCoordinatorListener jobCoordinatorListener = null;
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -66,16 +66,16 @@ public class StreamProcessor {
   private final Config config;
   private final long taskShutdownMs;
   private final String processorId;
+  private final ExecutorService executorService;
+  private final Object lock = new Object();
 
-  private ExecutorService executorService;
-
-  private volatile SamzaContainer container = null;
-  private volatile Throwable containerException = null;
+  private SamzaContainer container = null;
+  private Throwable containerException = null;
+  private boolean processorOnStartCalled = false;
 
   // Latch used to synchronize between the JobCoordinator thread and the container thread, when the container is
   // stopped due to re-balancing
-  volatile CountDownLatch jcContainerShutdownLatch;
-  private volatile boolean processorOnStartCalled = false;
+  CountDownLatch jcContainerShutdownLatch;
 
   @VisibleForTesting
   JobCoordinatorListener jobCoordinatorListener = null;
@@ -134,8 +134,9 @@ public class StreamProcessor {
     this.jobCoordinator = (jobCoordinator != null) ? jobCoordinator : getJobCoordinator();
     this.jobCoordinatorListener = createJobCoordinatorListener();
     this.jobCoordinator.setListener(jobCoordinatorListener);
-
-    processorId = this.jobCoordinator.getProcessorId();
+    ThreadFactory threadFactory = new ThreadFactoryBuilder().setNameFormat(CONTAINER_THREAD_NAME_FORMAT).build();
+    this.executorService = Executors.newSingleThreadExecutor(threadFactory);
+    this.processorId = this.jobCoordinator.getProcessorId();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -208,7 +210,8 @@ JobCoordinatorListener createJobCoordinatorListener() {
 
       @Override
       public void onJobModelExpired() {
-        if (container != null) {
+        SamzaContainer container = containerRef.get();
+        if (container != null && containerRef.compareAndSet(container, null)) {
```
dnishimura: The `else` block assumes if `containerRef.compareAndSet(container, null)` is false, the containerRef is null. Can we guarantee containerRef doesn't hold another `container` reference?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -224,10 +227,11 @@ public void onJobModelExpired() {
             } catch (InterruptedException e) {
               Thread.currentThread().interrupt();
               LOGGER.warn(String.format("Shutdown of container: %s for stream processor: %s was interrupted", container, processorId), e);
+            } catch (Exception e) {
```
dnishimura: Why is `Exception` being swallowed?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -255,20 +259,25 @@ public void onContainerStart() {
...
+              }
               if (jcContainerShutdownLatch != null) {
                 jcContainerShutdownLatch.countDown();
               }
             } else {  // sp.stop was called or container stopped by itself
-              LOGGER.info("Container: {} stopped. Stopping the stream processor: {}.", container, processorId);
-              container = null; // this guarantees that stop() doesn't try to stop container again
+              SamzaContainer container = containerRef.getAndSet(null); // this guarantees that stop() doesn't try to stop container again
+              if (container != null) {
+                LOGGER.info("Container: {} stopped. Stopping the stream processor: {}.", container, processorId);
```
dnishimura: Would it be useful to log the `processorId` even though `container` is null? Same goes for the if block above.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -281,14 +290,22 @@ public void onContainerFailed(Throwable t) {
               LOGGER.warn("JobCoordinatorLatch was null. It is possible for some component to be waiting.");
             }
             containerException = t;
-            LOGGER.error(String.format("Container: %s failed with an exception. Stopping the stream processor: %s. Original exception:", container, processorId), containerException);
-            container = null;
+            SamzaContainer container = containerRef.getAndSet(null);
+            if (container != null) {
+              LOGGER.error(String.format("Container: %s failed with an exception. Stopping the stream processor: %s. Original exception:", container, processorId), containerException);
```
dnishimura: I think it'd be useful to log the `processorId` and `containerException` even if container is null.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
shanthoosh: When `SamzaContainerRef` is assigned to new value(other than null), it happens only in `onNewJobModel` in StreamProcessor(Other updations sets it to null: `onContainerStop` & `onContainerFailure`). 

`onJobModelExpired` and `onNewJobModel` will not be executed concurrently, since their executions are serialized through debounceThread. 

I think it's safe to assume that it'll be null in else case(if the compareAndSet fails).

Does that makes sense?

dnishimura: Yup makes sense. Thanks for the explanation. I think it'd be useful to document this info in comments since it's not immediately obvious. Might be useful to log for the non-null else case if we come across a future bug through regression or new code.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
dnishimura: 👍

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -69,7 +70,7 @@
 
   private ExecutorService executorService;
 
-  private volatile SamzaContainer container = null;
+  private final AtomicReference<SamzaContainer> containerRef = new AtomicReference<>(null);
```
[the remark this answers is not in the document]
dnishimura: 👍

vjagadish:
Discussed with @shanthoosh offline. Currently, the synchronization model of this class is convoluted to understand. We currently maintain state in volatile variables and atomic-references being set/reset.  We should consider simplifying and updating it via a single lock for all the state changes. This will ensure that there's a clear ordering among updates and will also achieve atomicity when updating state-variables.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -281,14 +286,20 @@ public void onContainerFailed(Throwable t) {
...
             stop();
           }
         };
 
-        container = createSamzaContainer(processorId, jobModel);
+        SamzaContainer container = createSamzaContainer(processorId, jobModel);
         container.setContainerListener(containerListener);
+
+        SamzaContainer previousContainer = containerRef.getAndSet(container);
+        if (previousContainer != null && previousContainer.getStatus().equals(SamzaContainerStatus.STARTED)) {
```
vjagadish: Don't think this should be necessary if you update state atomically. It looks like you would not have the state for `previousContainer` in the first place?

shanthoosh: @vjagadish1989 
Agreed to this suggestion and switched to using explicit locks.

vjagadish1989:
LGTM, merged and submitted thanks!

PR #495, opened 2018-04-30
SAMZA-1693: Samza-sql - Adding Serde for rel record and few other minor fixes for Avro and Rel conversion.

atoomula:
Adding Serde for rel record, as calcite expects the keys to be in string format. Rel converters are always expected to provide keys as strings. If key is an avro record, it is expected that the rel converter changes the avro record to rel record and serializes it and deserializes it when conerting rel message to samza message.

### DIFF
--- a/samza-sql/src/test/java/org/apache/samza/sql/TestSamzaSqlRelRecordSerde.java
+++ b/samza-sql/src/test/java/org/apache/samza/sql/TestSamzaSqlRelRecordSerde.java
@@ -20,12 +20,24 @@
 package org.apache.samza.sql;
 
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericData;
+import org.apache.samza.config.MapConfig;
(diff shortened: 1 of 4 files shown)


samza-sql/src/test/java/org/apache/samza/sql/TestSamzaSqlRelRecordSerde.java
```diff
@@ -0,0 +1,48 @@
...
+import static org.apache.samza.sql.data.SamzaSqlRelMessage.SamzaSqlRelRecord;
+
+
+public class TestSamzaSqlRelRecordSerde {
+
+  private List<Object> values = Arrays.asList("value1", 1, null);
+  private List<String> names = Arrays.asList("field1", "field2", "field3");
+
+  @Test
+  public void testWithDifferentFields() {
```
srinipunuru: I think it might be good idea to add non primitive values (List and Maps) to the test so that those scenarios are protected if there are any regressions in future.

atoomula: Sure - added it.

PR #496, opened 2018-04-30
SAMZA-1692: Standalone stability fixes.

shanthoosh:
- Currently, on session expiration processorListener with incorrect generationId is registered with zookeeper(ZkUtils generationId is incremented on reconnect but the generationId in processorListener is zero all the time). When a session reconnect happens to a processor successive to leader, leader expiration event will be skipped. This will prevent leader re-election on a current leader death and will stall the processors group. Fix is to reinstantiate and then register processorChangeListener on session expiration.
- Add processorId to debounce thread name (this can aid debugging when multiple processors are running within a jvm).
- After ScheduleAfterDebounceTime queue is shutdown, don't accept new schedule requests. Current ZkJobCoordinator shutdown sequence comprise of the following steps
     - Shutdown the ScheduleAfterDebounceTime queue.
     - Stop the zkClient  and relinquish it's resources.

After we shutdown ScheduleAfterDebounceTime and before zkclient is stopped, any new operations can be scheduled in ScheduleAfterDebounceTime queue. This will result in RejectedExecutionException, since executorService is stopped.

```
Caused by: java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@23f962a8 rejected from java.util.concurrent.ScheduledThreadPoolExecutor@43408be8
```

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
@@ -50,7 +50,7 @@ public class ScheduleAfterDebounceTime {
   private Optional<ScheduledTaskCallback> scheduledTaskCallback;
 
   // Responsible for scheduling delayed actions.
-  private ScheduledExecutorService scheduledExecutorService;
+  private final ScheduledExecutorService scheduledExecutorService;
 
   /**
    * A map from actionName to {@link ScheduledFuture} of task scheduled for execution.
@@ -169,7 +169,7 @@ public class ScheduleAfterDebounceTime {
             LOG.warn("Action: {} is interrupted.", actionName);
             doCleanUpOnTaskException(new InterruptedException());
           } else {
-            LOG.debug("Action: {} completed successfully.", actionName);
+            LOG.info("Action: {} completed successfully.", actionName);
           }
         }
       } catch (Throwable throwable) {
--- a/samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
@@ -20,13 +20,13 @@
 package org.apache.samza.zk;
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


shanthoosh:
@vjagadish1989 @xinyuiscool 

tdavis:
Would be great to get this into 0.14.1 too ;)

samza-core/src/main/java/org/apache/samza/zk/ZkLeaderElector.java
```diff
@@ -130,6 +130,7 @@ public void tryBecomeLeader() {
         LOG.debug(zLog("Unsubscribing data change for " + currentSubscription));
         zkUtils.unsubscribeDataChanges(keyBuilder.getProcessorsPath() + "/" + currentSubscription,
             previousProcessorChangeListener);
+        previousProcessorChangeListener = new PreviousProcessorChangeListener(zkUtils);
```
vjagadish: fyi, I did a pass at the other `IZKDataListener`s for barrier-joins and job-model upgrade as well and how they are instantiated, it lgtm.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -50,17 +50,20 @@
   private Optional<ScheduledTaskCallback> scheduledTaskCallback;
 
   // Responsible for scheduling delayed actions.
-  private final ScheduledExecutorService scheduledExecutorService;
+  private ScheduledExecutorService scheduledExecutorService;
```
vjagadish: nit: revert this for this particular PR since the executor is initialized during construction

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
```diff
@@ -146,16 +159,18 @@ private void tryCancelScheduledAction(String actionName) {
...
+          runnable.run();
+          /*
+           * Expects all run() implementations <b>not to swallow the interrupts.</b>
+           * This thread is interrupted from an external source(mostly executor service) to die.
+           */
+          if (Thread.currentThread().isInterrupted()) {
+            LOG.warn("Action: {} is interrupted.", actionName);
+            doCleanUpOnTaskException(new InterruptedException());
+          } else {
+            LOG.debug("Action: {} completed successfully.", actionName);
```
vjagadish: Make this "info". there's relatively fewer operations scheduled on the executor, and would be good to log their completion status

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkLeaderElector.java
shanthoosh: Yes, this is the only place where the reregistration was done incorrectly.

PR #497, opened 2018-05-01
SAMZA-1243: Fix flaky tests in TestLocalStoreMonitor

ahmedahamid:
Having attempted to reproduce this issue without success, careful analysis of the code indicates the underlying issue is most likely incomplete cleanup of file system directories after test execution possibly due to incomplete or manual test runs. To address this issue, this fix generates a different local store directory for every test to isolate file system side-effects across different tests and test runs.


prateekm:
Thanks @ahmedahamid! @shanthoosh, can you review this?

PR #500, opened 2018-05-02
SAMZA-1696: Delete TestKeyValueStores flaky test

ahmedahamid:
`testParallelReadWriteSameKey` is a flaky test that was failing due to a race condition between the main test thread and the thread manipulating the key-value store under test. Specifically, the main test thread could assert the store has received a value before another thread gets to set it. The hard-coded wait durations did not guarantee the main test thread would wait for all other threads to complete, causing it to assert prematurely.

This change deletes this test since the underlying `RocksDB` store wrapped within `RocksDbKeyValueStore` is thread-safe by design, which renders this test redundant and unnecessary.

### DIFF
--- a/samza-test/src/test/scala/org/apache/samza/storage/kv/TestKeyValueStores.scala
+++ b/samza-test/src/test/scala/org/apache/samza/storage/kv/TestKeyValueStores.scala
@@ -409,40 +409,6 @@ class TestKeyValueStores(typeOfStore: String, storeConfig: String) {
   }
 
   @Test
-  def testParallelReadWriteSameKey(): Unit = {
-    // Make test deterministic by seeding the random number generator.
-    val key = b("key")
-    val oldVal = "val1"
-    val newVal = "val2"
-
-    val runner1 = new Thread(new Runnable {
-      override def run(): Unit = {
-        store.put(key, b(oldVal))
-      }
-    })
-
-    val runner2 = new Thread(new Runnable {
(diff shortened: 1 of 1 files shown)


ahmedahamid:
@xinyuiscool Could you, please, review this change when you get a chance?

samza-test/src/test/scala/org/apache/samza/storage/kv/TestKeyValueStores.scala
```diff
@@ -388,32 +412,31 @@ class TestKeyValueStores(typeOfStore: String, storeConfig: String) {
   def testParallelReadWriteSameKey(): Unit = {
```
shanthoosh: The test tries to test if `RocksDBKeyValueStore` is threadsafe by reading/writing a key from two different threads. 

RocksDB JNI class inherently is thread-safe. Here's the relevant documentation: 

`
  A RocksDB is a persistent ordered map from keys to values.  It is safe for
  concurrent access from multiple threads without any external synchronization
` 

Samza's `RocksDBKeyValueStore` implementation appears to a wrapper over `RocksDB` JNI class and doesn't have any additional state. I think this doesn't test doesn't add much value and can be deleted.

ahmedahamid: Fair. I deleted the test as agreed.

PR #502, opened 2018-05-04
SAMZA-1698: Update appStatus on failures in localApplication.run(streamApp).

shanthoosh:


shanthoosh:
@dnishimura @vjagadish1989 @xinyuiscool 

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -175,8 +175,10 @@ public void run(StreamApplication app) {
 
       // 4. start the StreamProcessors
       processors.forEach(StreamProcessor::start);
-    } catch (Exception e) {
```
vjagadish: thanks for the fix! This currently works for the single StreamProcessor case and is sufficient. For the multi-stream processor case, we may have to teardown the stream-processor that was already started.

PR #509, opened 2018-05-08
SAMZA-1700 : Clean up SystemAdmins instance creation flows

cameronlee314:
The SystemAdmins class has a special "test" constructor for building SystemAdmins instances for unit tests. However, that test constructor has leaked into non-test code.
This makes it harder to manage all flows which use SystemAdmins. An upcoming change is to fix lifecycle management for SystemAdmins in the ApplicationRunner classes, so doing some clean up will help that future change.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
+++ b/samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
@@ -20,26 +20,30 @@
 package org.apache.samza.system;
 
 import java.util.Map;
+import java.util.Set;
 import org.apache.samza.SamzaException;
 import org.apache.samza.config.Config;
 import org.apache.samza.config.JavaSystemConfig;
+import org.apache.samza.config.MapConfig;
 
 
+/**
+ * Provides a mapping from system name to a {@link SystemAdmin}. Remember to start/stop this.
+ */
 public class SystemAdmins {
   private final Map<String, SystemAdmin> systemAdminMap;
-
-  public Map<String, SystemAdmin> getSystemAdminsMap() {
-    return systemAdminMap;
-  }
 
   public SystemAdmins(Config config) {
     JavaSystemConfig systemConfig = new JavaSystemConfig(config);
     this.systemAdminMap = systemConfig.getSystemAdmins();
   }
 
-  // Used only for test
-  public SystemAdmins(Map<String, SystemAdmin> systemAdminMap) {
-    this.systemAdminMap = systemAdminMap;
+  /**
+   * @return New instance of {@link SystemAdmins} with an empty admin mapping. Since it has no admins, it does not need
+   * to be started/stopped.
+   */
+  public static SystemAdmins empty() {
+    return new SystemAdmins(new MapConfig());
   }
 
   public void start() {
@@ -59,5 +63,9 @@ public class SystemAdmins {
       throw new SamzaException("Cannot get systemAdmin for system " + systemName);
     }
     return systemAdminMap.get(systemName);
+  }
+
+  public Set<String> getAdminKeys() {
+    return systemAdminMap.keySet();
   }
 }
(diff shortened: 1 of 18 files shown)


samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
```diff
@@ -60,4 +64,8 @@ public SystemAdmin getSystemAdmin(String systemName) {
     }
     return systemAdminMap.get(systemName);
   }
+
+  public Set<String> getAdminKeys() {
```
prateekm: Maybe name the method to what the keys represent? E.g. getSystemNames?

samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
```diff
@@ -20,26 +20,30 @@
...
   public SystemAdmins(Config config) {
     JavaSystemConfig systemConfig = new JavaSystemConfig(config);
     this.systemAdminMap = systemConfig.getSystemAdmins();
   }
 
-  // Used only for test
-  public SystemAdmins(Map<String, SystemAdmin> systemAdminMap) {
-    this.systemAdminMap = systemAdminMap;
+  /**
+   * @return New instance of {@link SystemAdmins} with an empty admin mapping. Since it has no admins, it does not need
```
prateekm: Javadoc format: description, param, return.

Can remove comment about not needing start/stop, since they should work (no-op) even if called.

samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
```diff
@@ -20,26 +20,30 @@
...
 import java.util.Map;
+import java.util.Set;
 import org.apache.samza.SamzaException;
 import org.apache.samza.config.Config;
 import org.apache.samza.config.JavaSystemConfig;
+import org.apache.samza.config.MapConfig;
 
 
+/**
+ * Provides a mapping from system name to a {@link SystemAdmin}. Remember to start/stop this.
```
prateekm: Maybe: Needs to be started/stopped before/after use.

samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
cameronlee314: I had the comment in there for a few cases where it is a bit difficult to actually add the start/stop. For example, there are a few cases where an empty instance is used as the default argument value for a Scala method, and I wanted to clarify that the Scala method does not need to actually manage the lifecycle of that SystemAdmins, even though it is creating a new instance of the SystemAdmins.
Do you think the clarification is unnecessary?

prateekm: > there are a few cases where an empty instance is used as the default argument value for a Scala method
Is it simple to remove the default value? I can't imagine why a method would have an empty SystemAdmins.

cameronlee314: The non-test use cases ignore this default value, since they all pass an actual SystemAdmins. However, there are a significant number of tests which depend on this default value (I'm guessing those particular tests just don't need the admins). I was hoping to avoid refactoring all of those tests right now. Also, we would also possibly want to reorder argument lists to put the now-required SystemAdmins argument before other optional arguments, which is another bit of refactoring.
My hope with this current change is to give some clarification to where "real" SystemAdmins are created and to prevent future usage of this "test constructor" which is a bit too flexible.

samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
cameronlee314: Resolved

samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
cameronlee314: Resolved

samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
cameronlee314: Synced offline about the start/stop comment. Will remove it in order to reduce confusion. In general, users should do start/stop anyways, but for certain cases like defaults for testing, it can be ok to leave out the start/stop.
Resolved

PR #511, opened 2018-05-08
SAMZA-1706 lazy initialization for eventhub system producer

lhaiesp:
We are seeing slow shutdown issue for eventhub system producers for users who only use eventhub consumer (but then Samza system creates both consumer and producer together no matter what). As a workaround, add lazy initialization to the producer to avoid the slow shutdown

### DIFF
--- a/samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
@@ -97,7 +97,7 @@ public class EventHubSystemProducer extends AsyncSystemProducer {
 
   // We implement lazy initialization for producer as a workaround for
   // slow shutdown issue.
-  private volatile boolean isInitialized = false;
+  private boolean isInitialized = false;
 
   /**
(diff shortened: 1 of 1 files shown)


samza-azure/src/main/java/org/apache/samza/system/eventhub/producer/EventHubSystemProducer.java
```diff
@@ -245,6 +254,10 @@ public synchronized void flush(String source) {
     if (!isStarted) {
       throw new SamzaException("Trying to call send before the producer is started.");
     }
+    if (!isInitialized) {
+      // lazy initialization on the first send
+      init();
```
vjagadish: @lhaiesp thanks for the workaround. it will also be good to understand why the `PartitionSender` does not shut-down cleanly - maybe there's a bug in the client or we use the client is incorrectly? Can you create a ticket for it and follow-up with the EventHubs team (if need be)?

lhaiesp: Created the ticket:

https://issues.apache.org/jira/browse/SAMZA-1708

lhaiesp:
@vjagadish  can you merge the commit if it looks good to you?

PR #512, opened 2018-05-09
SAMZA-1709 : Use lazy creation for SystemAdmins in ApplicationRunners

cameronlee314:
An instance of SystemAdmins is created when instantiating any AbstractApplicationRunner, but the SystemAdmins is only actually needed for some of the methods for some of the runners. For example, LocalApplicationRunner.kill does not need SystemAdmins, and LocalContainerRunner does not need SystemAdmins for anything.
Doing lazy instantiation allows us to more easily manage the SystemAdmins lifecycle, since it removes the need to add lifecycle hooks for the ApplicationRunner.
This also fixes the lifecycle management for SystemAdmins in ApplicationRunners.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
@@ -41,10 +41,12 @@ import org.apache.samza.config.TaskConfig;
 import org.apache.samza.coordinator.CoordinationUtils;
 import org.apache.samza.coordinator.DistributedLockWithState;
 import org.apache.samza.execution.ExecutionPlan;
+import org.apache.samza.execution.StreamManager;
 import org.apache.samza.job.ApplicationStatus;
 import org.apache.samza.processor.StreamProcessor;
 import org.apache.samza.processor.StreamProcessorLifecycleListener;
 import org.apache.samza.system.StreamSpec;
+import org.apache.samza.system.SystemAdmins;
 import org.apache.samza.task.AsyncStreamTaskFactory;
 import org.apache.samza.task.StreamTaskFactory;
 import org.apache.samza.task.TaskFactoryUtil;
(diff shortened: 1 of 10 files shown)


prateekm:
@xinyuiscool can you take a look at this?

prateekm:
@cameronlee314 can you rebase from master and update this PR. @xinyuiscool can you please take a look at this.

xinyuiscool:
Thanks for the patch. The idea is awesome. I actually think that we can be even being lazier: no start() function in the SystemAdmins(), and only do systemAdmin.start() when the getSysteAdmin() is called. That way we don't need to start the admin it if the system is not being used. This is actually the case in Beam: I configured the kafka system as default system for intermediate topics, but it's not being used in beam tests since we only have 1 task (no repartition needed). Now after upgrading to 0.14.1, I saw the kafka admin getting created and constantly logging warnings about broker not available. I would suggest keeping the old behavior and only create the admin when used.

cameronlee314:
SystemAdmins.getSystemAdmin() gets called throughout the lifecycle of the container (e.g. startup, commit). We are reusing SystemAdmin instances, so do we want to be calling SystemAdmin.start() every time someone wants a SystemAdmin? That seems a bit inconsistent with the normal start/stop usage pattern.
For the problem you are seeing in Beam: By the time Samza needs to instantiate the SystemAdmin instances, does Samza know if the default system will actually get used? Or is there some "runtime" usage of the default system that can't be detected at start time? If we can detect that the default system will never actually be used, we could avoid instantiating that system in the first place.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -149,10 +151,13 @@ public void runTask() {
 
   @Override
   public void run(StreamApplication app) {
+    SystemAdmins systemAdmins = null;
     try {
-      super.run(app);
+      systemAdmins = buildAndStartSystemAdmins();
```
xinyuiscool: seems this three lines are in quite a few places. I suggest hiding the systemAdmins inside the streamManager, add add start(), stop() in streamManager so we can invoke life cycle methods of systemAdmins. With this, we can still make streamManager a member in the abstract runner to avoid constructing it everywhere.

cameronlee314: Updated to put SystemAdmins inside StreamManager, and then manage the StreamManager lifecycle inside the runners.

PR #513, opened 2018-05-10
Fixed test failure for TestRocksDbKeyValueStoreJava#testIterate

prateekm:
RocksDB Snapshots and any iterators obtained from them need to be closed before the store is closed. 

Otherwise the process aborts with the following message (at least on OSX):
`Assertion failed: (is_last_reference), function ~ColumnFamilyData, file db/column_family.cc, line 457.`


xinyuiscool:
Thanks for the fix! I am going to pull this in to 0.14.1 release.

PR #517, opened 2018-05-10
SAMZA-1559 moving hello-samza repo to samza repo and put in samza-samples directory

fredji97:
Tested and jobs can be deployed in the new place. 
After this is committed, we will do the following two in order: 
1. adjust the documentation to use the sample jobs from the new place;
2. remove the old hello-samza repo.


nickpan47:
There is an earlier ticket that covers this issue as well. Please check https://issues.apache.org/jira/browse/SAMZA-205 to see how much this PR covers the issues discussed in that thread.

prateekm:
@fredji97 Doesn't seem like the deploy directory be a part of this commit. Are you moving a clean copy or a local copy?

prateekm:
@fredji97 also, have you tried moving the files in the existing hello-samza repo to this repo such that it preserves git history? If not, we should try that first before copying files over.

fredji97:
@prateekm Thanks for the comments. I have pulled in the hello-samza git log history as well.

shanthoosh:
@fredji97 
I think after yi pan's serializable operator changes, `samza-hello-samza` repository does not build with `samza` open source head. I'm fixing the compilation problems in `samza-hello-samza`. Please update the PR with my fixes and then merge it.

PR #519, opened 2018-05-11
SAMZA-1712: Tear down ZookeeperServer connections in ZkClient on interrupts.

shanthoosh:
**Problem:** 

If a thread executing zkClient.close is interrupted, currently we swallow the ZkInterruptedException and proceed without closing the zookeeper connection. 

This leads to ephemeral nodes of StreamProcessor lurking around in zookeeper after StreamProcessor shutdown.

Users had to wait till zookeeper server session timeout for the ephemeral nodes to get deleted. 

**Change:**

Retry once on InterruptedException when closing the zkClient.

Misc changes:
* Remove unnecessary null checks.
* Remove unnecessary typecasts.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
@@ -78,7 +78,6 @@ public class ZkUtils {
   private static final Logger LOG = LoggerFactory.getLogger(ZkUtils.class);
   /* package private */static final String ZK_PROTOCOL_VERSION = "1.0";
 
-
   private final ZkClient zkClient;
   private volatile String ephemeralPath = null;
   private final ZkKeyBuilder keyBuilder;
@@ -105,9 +104,7 @@ public class ZkUtils {
   public void connect() throws ZkInterruptedException {
     boolean isConnected = zkClient.waitUntilConnected(connectionTimeoutMs, TimeUnit.MILLISECONDS);
     if (!isConnected) {
-      if (metrics != null) {
-        metrics.zkConnectionError.inc();
-      }
+      metrics.zkConnectionError.inc();
       throw new RuntimeException("Unable to connect to Zookeeper within connectionTimeout " + connectionTimeoutMs + "ms. Shutting down!");
     }
   }
@@ -144,6 +141,7 @@ public class ZkUtils {
       if (!isValidRegisteredProcessor(processorNode)) {
         LOG.info("Processor: {} is duplicate. Deleting zookeeper node at path: {}.", processorId, ephemeralPath);
         zkClient.delete(ephemeralPath);
+        metrics.deletions.inc();
         throw new SamzaException(String.format("Processor: %s is duplicate in the group. Registration failed.", processorId));
       }
     } else {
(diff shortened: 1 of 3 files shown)


shanthoosh:
@xinyuiscool @sborya 

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -290,23 +286,25 @@ public void unsubscribeChildChanges(String path, IZkChildListener childListener)
...
-      Thread.interrupted();
-      LOG.warn("Ignoring the exception when closing the zookeeper client.", e);
-    }
+    int retryCount = 0;
+    do {
+      try {
+        zkClient.close();
+        return;
+      } catch (ZkInterruptedException e) {
+        Thread.interrupted();
```
vjagadish: We shouldn't be clearing the interrupt status of the thread. Not restoring interrupt can often be dangerous since it masks the existence of an interrupt upstream in the current-thread's call-stack.  If the current thread is interrupted, we should be restoring its interruption status back right?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -290,23 +286,25 @@ public void unsubscribeChildChanges(String path, IZkChildListener childListener)
...
 
   public void close() throws ZkInterruptedException {
-    try {
-      zkClient.close();
-    } catch (ZkInterruptedException e) {
-      // Swallowing due to occurrence in the last stage of lifecycle (Not actionable) and clear the interrupted status.
-      Thread.interrupted();
-      LOG.warn("Ignoring the exception when closing the zookeeper client.", e);
-    }
+    int retryCount = 0;
```
vjagadish: Not sure if the retry is necessary here.

vjagadish:
@shanthoosh : Please add an integration test (in this PR or in a follow-up PR) that verifies that calling zkUtils.close() actually nullifies the Zk client connection

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
```diff
@@ -463,4 +459,44 @@ public void testgetNextJobModelVersion() {
...
+        } catch (InterruptedException e) {
+          Thread.currentThread().interrupt();
+        }
+        zkUtils.close();
+      });
+
+    threadToInterrupt.start();
+    threadToInterrupt.interrupt();
+    threadToInterrupt.join();
+
```
vjagadish: We assert that the client connection is closed .. Should we assert that it was "connected" prior to invoking interrupt()?

shanthoosh: @vjagadish1989 
`ZkClient`  setups connection with the zookeeper server during instantiation. Here's the ZkClient constructor:
```
    public ZkClient(final IZkConnection zkConnection, final int connectionTimeout, final ZkSerializer zkSerializer, final long operationRetryTimeout) {
        if (zkConnection == null) {
            throw new NullPointerException("Zookeeper connection is null!");
        }
        _connection = zkConnection;
        _zkSerializer = zkSerializer;
        _operationRetryTimeoutInMillis = operationRetryTimeout;
        _isZkSaslEnabled = isZkSaslEnabled();
        connect(connectionTimeout, this);
    }
```
`ZkClient` constructor used in the test invoke this through a call chain. 

Added a assertion before interruption that the connection wasn't closed. There're no other state exposed from `ZkConnection` interface, hence this is only possible check we can do.

PR #520, opened 2018-05-12
SAMZA-1670 : When fetching a newest offset for a partition, also prefetch and cache the newest offsets for other partitions on the container

cameronlee314:


samza-core/src/main/scala/org/apache/samza/system/SSPMetadataCache.java
```diff
@@ -0,0 +1,124 @@
...
+   * @throws RuntimeException if there was an error in fetching metadata
+   */
+  public SystemStreamMetadata.SystemStreamPartitionMetadata getMetadata(SystemStreamPartition ssp) {
+    refreshMetadataIfNecessary(ssp);
+    CacheEntry cacheEntry = cache.get(ssp);
+    // shouldn't get null cacheEntry once refresh is done, but check anyways to be safe
+    return cacheEntry == null ? null : cacheEntry.getMetadata();
+  }
+
+  private void refreshMetadataIfNecessary(SystemStreamPartition requestedSSP) {
```
vjagadish1989: s/refreshMetadataIfNecessary/maybeRefreshMetadata so that it's consistent with the rest of the Samza codebase?

samza-core/src/main/scala/org/apache/samza/system/SSPMetadataCache.java
```diff
@@ -0,0 +1,124 @@
...
+    synchronized (this.metadataRefreshLock) {
+      Instant refreshRequestedAt = Instant.ofEpochMilli(this.clock.currentTimeMillis());
+      if (needToFetch(requestedSSP, refreshRequestedAt)) {
```
vjagadish1989: s/needToFetch/isStale or shouldRefresh?

vjagadish1989 (approved):
looks good except some minor f/b for consistency and naming. Thanks Cameron for the contribution. Nice work!

samza-core/src/main/scala/org/apache/samza/system/SSPMetadataCache.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import org.apache.samza.util.Clock;
+
+
+/**
+ * Fetches and caches metadata about a set of SSPs. This can prefetch the metadata for a defined set of SSPs when it is
+ * fetching metadata for some arbitrary SSP.
+ */
+public class SSPMetadataCache {
```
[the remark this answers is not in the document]
cameronlee314: Although there are some similarities in functionality, it will be difficult to have a shared class at this point:
1. StreamMetadataCache has a "partitionsMetadataOnly" parameter which is specific to stream metadata.
2. SSPMetadataCache uses prefetching, but StreamMetadataCache does not, so we still need some extra logic.
3. The code structure of SSPMetadataCache is a bit different from StreamMetadataCache, due to the prefetching. It might be possible to do some refactoring to get them more similar, but given the two other reasons, it might not be worth it.

samza-core/src/main/scala/org/apache/samza/system/SSPMetadataCache.java
```diff
@@ -0,0 +1,124 @@
...
+    this.sspsToPrefetch = sspsToPrefetch;
+
+    this.metadataRefreshLock = new Object();
+    this.cache = new ConcurrentHashMap<>();
+  }
+
+  /**
+   * Gets the metadata for an SSP. This will return a cached value if it is fresh enough. Otherwise, it will fetch the
+   * metadata from a source-of-truth.
+   * If the metadata for the SSP needs to be fetched, then this will also prefetch and cache the metadata for any
```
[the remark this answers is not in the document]
cameronlee314: Good suggestion. Updated comment.

samza-core/src/main/scala/org/apache/samza/system/SSPMetadataCache.java
```diff
@@ -0,0 +1,124 @@
...
+   *
+   * @param ssp SSP for which to get metadata
+   * @return metadata for the SSP; null if the source-of-truth returned no metadata
+   * @throws RuntimeException if there was an error in fetching metadata
+   */
+  public SystemStreamMetadata.SystemStreamPartitionMetadata getMetadata(SystemStreamPartition ssp) {
+    refreshMetadataIfNecessary(ssp);
+    CacheEntry cacheEntry = cache.get(ssp);
+    // shouldn't get null cacheEntry once refresh is done, but check anyways to be safe
+    return cacheEntry == null ? null : cacheEntry.getMetadata();
```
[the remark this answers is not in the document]
cameronlee314: The null return value in the javadoc is more for the case when cacheEntry is not null, but cacheEntry.getMetadata is null. I updated the comment to reflect that.

samza-core/src/main/scala/org/apache/samza/system/SSPMetadataCache.java
cameronlee314: Resolved

samza-core/src/main/scala/org/apache/samza/system/SSPMetadataCache.java
cameronlee314: Changed to "shouldRefresh"

PR #522, opened 2018-05-15
SAMZA-1540:SystemProducer instance for StreamAppender should have task.drop.producer.errors==true

PawasChhokra:
@vjagadish1989 Kindly take a look, thanks.

### DIFF
--- a/samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
+++ b/samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
@@ -19,13 +19,11 @@
 
 package org.apache.samza.logging.log4j;
 
-import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.URL;
 import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Map;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
@@ -40,6 +38,7 @@ import org.apache.samza.config.Log4jSystemConfig;
 import org.apache.samza.config.MapConfig;
 import org.apache.samza.config.SerializerConfig;
 import org.apache.samza.config.ShellCommandConfig;
+import org.apache.samza.config.TaskConfig;
 import org.apache.samza.coordinator.JobModelManager;
 import org.apache.samza.job.model.JobModel;
 import org.apache.samza.logging.log4j.serializers.LoggingEventJsonSerdeFactory;
@@ -72,7 +71,7 @@ public class StreamAppender extends AppenderSkeleton {
   // Hidden config for now. Will move to appropriate Config class when ready to.
   private static final String CREATE_STREAM_ENABLED = "task.log4j.create.stream.enabled";
 
-  private static final String DROP_PRODUCER_ERRORS = "task.drop.producer.errors";
+  private static final String DROP_PRODUCER_ERROR = TaskConfig.DROP_PRODUCER_ERROR();
 
   protected static final int DEFAULT_QUEUE_SIZE = 100;
   private static final long DEFAULT_QUEUE_TIMEOUT_S = 2; // Abitrary choice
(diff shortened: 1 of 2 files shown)


samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
```diff
@@ -279,8 +285,17 @@ protected Config getConfig() {
     return config;
   }
 
+  private void addDropProducerErrorsConfig() {
+    Map<String, String> newConfig = new HashMap<>();
+    if (config.get(DROP_PRODUCER_ERRORS) == null) {
```
vjagadish: What do you think about using `config.containsKey()` here?

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
```diff
@@ -68,6 +72,8 @@
   // Hidden config for now. Will move to appropriate Config class when ready to.
   private static final String CREATE_STREAM_ENABLED = "task.log4j.create.stream.enabled";
 
+  private static final String DROP_PRODUCER_ERRORS = "task.drop.producer.errors";
```
vjagadish: Are you able to use the constant defined in `TaskConfig.DROP_PRODUCER_ERROR`?

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
```diff
@@ -279,8 +285,17 @@ protected Config getConfig() {
     return config;
   }
 
+  private void addDropProducerErrorsConfig() {
+    Map<String, String> newConfig = new HashMap<>();
+    if (config.get(DROP_PRODUCER_ERRORS) == null) {
+      newConfig.put(DROP_PRODUCER_ERRORS, "true");
+      config = new MapConfig(ImmutableList.of(config, newConfig));
```
vjagadish: remove this method and inline this inside `getConfig` so that all config wirings happen in a single place

vjagadish: Cleaner to use `new MapConfig(config, ImmutableMap.of("drop.xyz.errors", "baz"))` here? A nice thing is that you don;t need this boiler-plate write to the single-element hashmap.

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
shanthoosh: Does it makes sense to turn this on by default regardless of whatever value user has defined for `task.drop.producer.errors`?

PawasChhokra: Makes sense. Done.

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
PawasChhokra: Yes.

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
PawasChhokra: Got it. Done.

PawasChhokra: Got it. Thank you :)

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
prateekm: @PawasChhokra This was the intention. It shouldn't check if they've opted in for this, it should be enabled by default.

PawasChhokra: Got it. That means I have to let go of the if statement and turn it on by default.

samza-log4j/src/main/java/org/apache/samza/logging/log4j/StreamAppender.java
```diff
@@ -275,6 +278,8 @@ protected Config getConfig() {
     } catch (IOException e) {
       throw new SamzaException("can not read the config", e);
     }
+    // Make system producer drop producer errors for StreamAppender
+    config = new MapConfig(config, ImmutableMap.of(TaskConfig.DROP_PRODUCER_ERROR(), "true"));
```
shanthoosh: @PawasChhokra 
I think it will be better to add a assertion in existing  `StreamAppender` tests to check if this config property is turned on.

PawasChhokra: Yes. Got it.

PR #523, opened 2018-05-15
SAMZA-1715: Unit test for Kafka admin deletedMessagesCalled() fail

nickpan47:


nickpan47:
@vjagadish1989 and @prateekm could you take a quick look? Thanks!

PR #526, opened 2018-05-18
Schedule ZkBarrierChangeHandler into debounce queue for execution. 

shanthoosh:
In existing implementation, `ZkBarrierChangeHandler` is executed from the `ZkEventThread` and has following drawbacks:
* `ZkWatch` events are buffered into a in-memory queue(maintained by ZkClient) and delivered one at a time to ZkClient listener implementations. If the exeuction of a delivered `ZkWatch` event is in progress, then no other `ZkWatch` event will be delivered to the listeners. If `ZkBarrierChangeHandler` is executed from `ZkEventThread`,  any increase in processing latency will delay the delivery of other `ZkWatch` events(buffered in in-memory queue of ZkClient).
* During session expiration(zkConnection error scenario), buffering all events into `ScheduleAfterDebounceTime` helps us to garbage collect older generation events(to ensure correctness and not execute older generation `ZkWatch` events).

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ScheduleAfterDebounceTime.java
@@ -50,7 +50,7 @@ public class ScheduleAfterDebounceTime {
   private Optional<ScheduledTaskCallback> scheduledTaskCallback;
 
   // Responsible for scheduling delayed actions.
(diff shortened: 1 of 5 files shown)


shanthoosh:
@vjagadish1989 @xinyuiscool @dnishimura 

vjagadish1989 (approved):
looks good, thanks for the contribution, approved!

vjagadish1989:
edit the title with the created JIRA: SAMZA-1723

PR #527, opened 2018-05-18
SAMZA-1720: Remove javafx.util dependency from samza-sql tests.

shanthoosh:
In samza-sql module, currently few test classes(`TestSamzaSqlRelMessageSerde` and `TestSamzaSqlRelRecordSerde`) are dependent upon `javafx.util.Pair` class(coming from `javafx` module). `javafx.util.Pair` is not supported by default in all JDK builds(example; open-jdk java-8 doesn't support `javafx` module) and it belongs to `javafx` package which is primarily used for developing GUI applications. This dependency is removed and replaced with `Pair` class from `apache-commons`.


shanthoosh:
@atoomula  @prateekm @xinyuiscool 

PR #528, opened 2018-05-18
Implementing the fetchSinkInfo in ConfigBasedIOResolver

srinipunuru:
1. I think we missed implementing the fetchSinkInfo method in the ConfigBasedResolver when the API was introduced which is breaking the samza sql console tool. This fixes it.
2. latest release of mac removed realpath so the command line tools are broken. Removed the usage of realpath to fix these tools.

Thanks to @nickpan47 for identifying these problems.


srinipunuru:
@xinyuiscool Can you take a look at this one?

xinyuiscool:
Thanks. I will merge it to both master and 0.14.1 branch and verified the sql tutorial.

PR #532, opened 2018-05-23
SAMZA-1726: Isolate InMemorySystemFactory to run separately per job

Sanil15:
Tested by running the corresponding integration and unit tests

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
+++ b/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
@@ -19,7 +19,11 @@
 
 package org.apache.samza.system.inmemory;
 
+import com.google.common.base.Preconditions;
+import java.util.HashMap;
+import java.util.Map;
 import org.apache.samza.config.Config;
+import org.apache.samza.config.JobConfig;
 import org.apache.samza.metrics.MetricsRegistry;
 import org.apache.samza.system.SystemAdmin;
 import org.apache.samza.system.SystemConsumer;
@@ -31,20 +35,28 @@ import org.apache.samza.system.SystemProducer;
  * Initial draft of in-memory {@link SystemFactory}. It is test only and not meant for production use right now.
  */
 public class InMemorySystemFactory implements SystemFactory {
-  private static final InMemoryManager MEMORY_MANAGER = new InMemoryManager();
+  private static Map<Integer, InMemoryManager> jobToManager = new HashMap<>();
 
   @Override
   public SystemConsumer getConsumer(String systemName, Config config, MetricsRegistry registry) {
-    return new InMemorySystemConsumer(MEMORY_MANAGER);
+    jobToManager.putIfAbsent(parseJobId(config), new InMemoryManager());
+    return new InMemorySystemConsumer(jobToManager.get(parseJobId(config)));
   }
 
   @Override
   public SystemProducer getProducer(String systemName, Config config, MetricsRegistry registry) {
-    return new InMemorySystemProducer(systemName, MEMORY_MANAGER);
+    jobToManager.putIfAbsent(parseJobId(config), new InMemoryManager());
+    return new InMemorySystemProducer(systemName, jobToManager.get(parseJobId(config)));
   }
 
   @Override
   public SystemAdmin getAdmin(String systemName, Config config) {
-    return new InMemorySystemAdmin(MEMORY_MANAGER);
+    jobToManager.putIfAbsent(parseJobId(config), new InMemoryManager());
+    return new InMemorySystemAdmin(jobToManager.get(parseJobId(config)));
+  }
+
+  private Integer parseJobId(Config config) {
+    Preconditions.checkNotNull(config.get(JobConfig.JOB_ID()));
+    return Integer.parseInt(config.get(JobConfig.JOB_ID()));
   }
 }
--- a/samza-core/src/test/java/org/apache/samza/system/inmemory/TestInMemorySystem.java
+++ b/samza-core/src/test/java/org/apache/samza/system/inmemory/TestInMemorySystem.java
@@ -20,6 +20,7 @@
 package org.apache.samza.system.inmemory;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -28,6 +29,7 @@ import java.util.stream.IntStream;
 import java.util.stream.Stream;
 import org.apache.samza.Partition;
 import org.apache.samza.config.Config;
+import org.apache.samza.config.JobConfig;
 import org.apache.samza.config.MapConfig;
 import org.apache.samza.metrics.MetricsRegistry;
 import org.apache.samza.system.EndOfStreamMessage;
@@ -46,6 +48,7 @@ import static org.mockito.Mockito.*;
 
 
 public class TestInMemorySystem {
+  private static final String JOB_ID = "1";
   private static final String SYSTEM_NAME = "in-memory";
   private static final String STREAM_NAME = "test-stream";
   private static final String SOURCE = "test-in-memory-source";
@@ -61,14 +64,15 @@ public class TestInMemorySystem {
   private static final SystemStream SYSTEM_STREAM = new SystemStream(SYSTEM_NAME, STREAM_NAME);
 
   private MetricsRegistry mockRegistry = mock(MetricsRegistry.class);
-  private Config config = new MapConfig();
+  private Config config;
 
   private InMemorySystemFactory systemFactory;
   private SystemAdmin systemAdmin;
 
   public TestInMemorySystem() {
-    Config config = new MapConfig();
-
+    HashMap<String, String> configs = new HashMap<>();
+    configs.put(JobConfig.JOB_ID(), JOB_ID);
+    config = new MapConfig(configs);
     systemFactory = new InMemorySystemFactory();
     systemAdmin = systemFactory.getAdmin(SYSTEM_NAME, config);
     systemAdmin.createStream(new StreamSpec(STREAM_NAME, STREAM_NAME, SYSTEM_NAME, PARTITION_COUNT));


samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
```diff
@@ -31,20 +35,28 @@
  * Initial draft of in-memory {@link SystemFactory}. It is test only and not meant for production use right now.
  */
 public class InMemorySystemFactory implements SystemFactory {
-  private static final InMemoryManager MEMORY_MANAGER = new InMemoryManager();
+  private static Map<Integer, InMemoryManager> jobToManager = new HashMap<>();
```
shanthoosh: Does the access to `jobToManager` map should be thread-safe?

I think it would be better to synchronize all these factory methods.

Sanil15: I suppose changing Map -> ConcurrentHashMap makes it thread safe, unless you have a better approach to recommend.

shanthoosh: Yes, that would work.

Do you know if `InMemoryManager` (and associated dependencies) are thread safe(Was unable to find it from documentation)?

shanthoosh: * Static final variables should be of camel case notation. 
* Would prefer to rename `jobToManager` to `IN_MEMORY_MANAGERS`.

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
```diff
@@ -31,20 +35,28 @@
...
-    return new InMemorySystemConsumer(MEMORY_MANAGER);
+    jobToManager.putIfAbsent(parseJobId(config), new InMemoryManager());
+    return new InMemorySystemConsumer(jobToManager.get(parseJobId(config)));
   }
 
   @Override
   public SystemProducer getProducer(String systemName, Config config, MetricsRegistry registry) {
-    return new InMemorySystemProducer(systemName, MEMORY_MANAGER);
+    jobToManager.putIfAbsent(parseJobId(config), new InMemoryManager());
+    return new InMemorySystemProducer(systemName, jobToManager.get(parseJobId(config)));
```
shanthoosh: Just curious.

Would using `jobId` here work for high level API job? 

Besides functionality it will be better to not use JobId here. It's  looks odd to use JobId  in a factory for producer/consumer(In memory producer and consumer could be used outside of a samza job and doesn't have to be a part of a samza job).

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
```diff
@@ -31,20 +35,28 @@
  * Initial draft of in-memory {@link SystemFactory}. It is test only and not meant for production use right now.
  */
 public class InMemorySystemFactory implements SystemFactory {
-  private static final InMemoryManager MEMORY_MANAGER = new InMemoryManager();
+  private static Map<Integer, InMemoryManager> jobToManager = new HashMap<>();
 
   @Override
   public SystemConsumer getConsumer(String systemName, Config config, MetricsRegistry registry) {
-    return new InMemorySystemConsumer(MEMORY_MANAGER);
+    jobToManager.putIfAbsent(parseJobId(config), new InMemoryManager());
```
bharathkk: Can we roll this into the `parseJobId` method and maybe rename that to reflect its purpose of creating a manager from config along with the attribute that its keyed on?
E.g. `getManagerFor...` or `getManagerFrom...`

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
Sanil15: added final

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
Sanil15: since I have configured a new parameter I now call this method getOrDefaultInMemoryManagerByTestId

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
Sanil15: Replaced `jobId` with `test.id`. The guarantees of this param is, if configured it will provide an independent instance of InMemorySystem to run, if not the application uses a default static InMemorySystem which is shared across all the classes

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
```diff
@@ -31,20 +35,28 @@
...
 
   @Override
   public SystemAdmin getAdmin(String systemName, Config config) {
-    return new InMemorySystemAdmin(MEMORY_MANAGER);
+    jobToManager.putIfAbsent(parseJobId(config), new InMemoryManager());
+    return new InMemorySystemAdmin(jobToManager.get(parseJobId(config)));
+  }
+
+  private Integer parseJobId(Config config) {
+    Preconditions.checkNotNull(config.get(JobConfig.JOB_ID()));
```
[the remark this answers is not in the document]
Sanil15: From Above:

Replaced jobId with test.id. The guarantees of this param is, if configured it will provide an independent instance of InMemorySystem to run, if not the application uses a default static InMemorySystem which is shared across all the classes

Additionally:
- Test Framework will generate an random TestId for an application, we don't expose this to users

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
```diff
@@ -31,20 +35,28 @@
...
+    return Integer.parseInt(config.get(JobConfig.JOB_ID()));
```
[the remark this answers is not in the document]
Sanil15: Not required now, please refer to comments above!

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
```diff
@@ -31,20 +32,30 @@
  * Initial draft of in-memory {@link SystemFactory}. It is test only and not meant for production use right now.
  */
 public class InMemorySystemFactory implements SystemFactory {
-  private static final InMemoryManager MEMORY_MANAGER = new InMemoryManager();
+  private static final String TEST_ID = "test.id";
+  private static final ConcurrentHashMap<Integer, InMemoryManager> IN_MEMORY_MANAGERS = new ConcurrentHashMap<>();
```
shanthoosh: * Does the test.id have to be of `integer` type?
* Can you document the semantics of `test.id` here?  
* How does it work when multiple threads in a test share the same inmemory manager(Will each thread  have different in-memory-manager or all threads within the same test will have same inmemory manager)?

Sanil15: - Is there any advantage to use data types other than integer?
- Sure will document semantics of `test.id` inmemory manager
- We are planning to create inmemory manager per test

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
```diff
@@ -29,22 +30,43 @@
...
+ * </p>
+ * <p>
+ * If {@code test.id} param is not configured for an application, it shares a default InMemorySystem for all the streams
+ * This system is shared between all the applications missing {@code test.id} in their configs running in the same JVM
+ * using InMemorySystem
+ * </p>
  */
 public class InMemorySystemFactory implements SystemFactory {
-  private static final InMemoryManager MEMORY_MANAGER = new InMemoryManager();
+  private static final String TEST_ID = "test.id";
```
bharathkk: Can we rename this to a much more general configuration so that its not specific to tests? like `inmemory.scope`

bharathkk: I would suggest we move this configuration to `InMemorySystemConfig` class similar to our existing pattern for isolation.

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
```diff
@@ -29,22 +30,43 @@
...
   }
 
   @Override
   public SystemAdmin getAdmin(String systemName, Config config) {
-    return new InMemorySystemAdmin(MEMORY_MANAGER);
+    return new InMemorySystemAdmin(getOrDefaultInMemoryManagerByTestId(config));
+  }
+
+  private InMemoryManager getOrDefaultInMemoryManagerByTestId(Config config) {
+    if (!config.containsKey(TEST_ID) || config.get(TEST_ID) == null) {
```
bharathkk: Maybe something like
```
String inMemoryScope = config.get(...);
if (inMemoryScope != null) {
   return IN_MEMORY_MANAGER.computeIfAbsent(inMemoryScope, new InMemoryManager());
}

return DEFAULT_MANAGER;
```

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
```diff
@@ -29,22 +30,43 @@
...
+ * <p>
+ * If {@code test.id} param is not configured for an application, it shares a default InMemorySystem for all the streams
+ * This system is shared between all the applications missing {@code test.id} in their configs running in the same JVM
+ * using InMemorySystem
+ * </p>
  */
 public class InMemorySystemFactory implements SystemFactory {
-  private static final InMemoryManager MEMORY_MANAGER = new InMemoryManager();
+  private static final String TEST_ID = "test.id";
+  private static final ConcurrentHashMap<Integer, InMemoryManager> IN_MEMORY_MANAGERS = new ConcurrentHashMap<>();
```
bharathkk: You may want to change this accordingly based on the new `inmemory.scope` configuration.
I will leave it upto you to decide on the type of the configuration

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
Sanil15: agreed!

Sanil15: used `inmemory.scope` in InMemorySystemConfig

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
Sanil15: I am also adding DEFAULT_MANAGER in ConcurrentHashMap to make it thread safe (just in case)

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemFactory.java
```diff
@@ -31,20 +33,28 @@
...
   public SystemAdmin getAdmin(String systemName, Config config) {
-    return new InMemorySystemAdmin(MEMORY_MANAGER);
+    return new InMemorySystemAdmin(getOrDefaultInMemoryManagerByTestId(config));
+  }
+
+  private InMemoryManager getOrDefaultInMemoryManagerByTestId(Config config) {
+    String inMemoryScope = config.get(InMemorySystemConfig.INMEMORY_SCOPE);
+    if (inMemoryScope == null) {
+      inMemoryScope = InMemorySystemConfig.DEFAULT_INMEMORY_SCOPE;
+    }
```
bharathkk: We can move this logic to `InMemorySystemConfig` where if the config is present we return the value if not we return the default scope. By doing this, the SystemFactory will just need to get the scope and use that directly for fetch the value from map.

This will offer consistency in terms of what we default to in the absence of scope plus it eliminates the duplication of defaulting logic in multiple places.

Sanil15: Agreed!

PR #533, opened 2018-05-24
SAMZA-1728: BootstrappingChooser - Call checkOffset only for a lagging partition while choosing.

atoomula:


atoomula:
Moved checkOffset into the if block.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -196,12 +196,12 @@ class BootstrappingChooser(
           val systemStream = systemStreamPartition.getSystemStream
 
           updatedSystemStreams += systemStream -> (updatedSystemStreams.getOrElse(systemStream, 0) - 1)
-        }
 
-        // If the offset we just read is the same as the offset for the last
-        // message (newest) in this system stream partition, then we have read
-        // all messages, and can mark this SSP as bootstrapped.
-        checkOffset(systemStreamPartition, offset, OffsetType.NEWEST)
+          // If the offset we just read is the same as the offset for the last
```
shanthoosh: `NEWEST` offsetType denotes the offset at the tail-end of the system. 

If we receive a offset equal to NEWEST, then it will be sufficient to mark a bootstrap SSP as caught-up(the check doesn't have to be greater than or equal to). 

Instead of the following check in `checkOffset`: 
```java
    val comparatorResult = systemAdmin.offsetComparator(offset, offsetToCheck)
    if (comparatorResult != null && comparatorResult.intValue() >= 0) {
      ...
      ...
    }
```

I think it should be like this:
```java
    val comparatorResult = systemAdmin.offsetComparator(offset, offsetToCheck)
    if (comparatorResult != null && comparatorResult.intValue() == 0) {
      ...
      ...
    }
```

I think this should solve the problem reported in `SAMZA-1728`.

Do you see any issues in this?

atoomula: There could be systems where there could be two different topics: one for bootstrap and the other for change capture. When systemAdmin advertises it's newest offset, it would be that of change capture's latest offset, when the job started. There is this corner case: By the time we end up consuming all of bootstrap topic and we switch to change capture, the initially advertised offset could be out of retention. In that case, greater than would be required.

shanthoosh: I understand why we need this check. Makes sense.

samza-core/src/main/scala/org/apache/samza/system/chooser/BootstrappingChooser.scala
```diff
@@ -196,12 +196,12 @@ class BootstrappingChooser(
...
+          // message (newest) in this system stream partition, then we have read
+          // all messages, and can mark this SSP as bootstrapped.
+          checkOffset(systemStreamPartition, offset, OffsetType.NEWEST)
```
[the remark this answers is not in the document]
atoomula: Added the test. Thanks!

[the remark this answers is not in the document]
shanthoosh: Build fails in jenkins with compilation errors.

```
[ant:scalac] Element '/home/jenkins/jenkins-slave/workspace/samza-pr-checks/samza-core/build/resources/main' does not exist.
[ant:scalac] Element '/home/jenkins/jenkins-slave/workspace/samza-pr-checks/samza-api/build/resources/test' does not exist.
:samza-core_2.11:compileTestScala FAILED

FAILURE: Build failed with an exception.
```

Can we please fix that.

PR #534, opened 2018-05-24
SAMZA-1677 : Make httpcore and httpclient dependencies consistent

cameronlee314:
Samza currently depends on httpclient 4.5.2 and httpcore 4.4.5. However, httpclient 4.5.2 also has a direct dependency on httpcore 4.4.4, which is not backwards compatible with httpcore 4.4.5 since some classes were removed (e.g. ThreadSafe/NotThreadSafe annotation classes).

Although this does not currently cause any direct build problems, there may be cases where this conflict introduces transitive dependency conflicts. In addition, this inconsistency can cause confusion in future development if those libraries need to be used.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -236,7 +236,6 @@ project(':samza-aws') {
     compile project(":samza-core_$scalaVersion")
     compile "org.slf4j:slf4j-api:$slf4jVersion"
     runtime "org.apache.httpcomponents:httpclient:4.5.2"
-    runtime "org.apache.httpcomponents:httpcore:4.4.5"
     testCompile "junit:junit:$junitVersion"
     testCompile "org.mockito:mockito-all:$mockitoVersion"
   }


vjagadish1989 (approved):
looks good to me. thanks Cameron for the contribution!

build.gradle
```diff
@@ -236,7 +236,6 @@ project(':samza-aws') {
     compile project(":samza-core_$scalaVersion")
     compile "org.slf4j:slf4j-api:$slf4jVersion"
     runtime "org.apache.httpcomponents:httpclient:4.5.2"
```
shanthoosh: Modules `samza-yarn`, `samza-autoscaling` are using httpClient version `4.4.1` through `httpClientVersion` variable defined in dependency-versions.gradle.

Since httpclient `4.5.x` version is backwards compatible with `4.4.x`, I think it will be better to use one consistent version across all samza modules(Either update other modules to use 4.5.x or downgrade this one to 4.4.x).

Please change if this makes sense.

cameronlee314: My observation was that the version backwards incompatibility problem is on httpcore (between 4.4.4 and 4.4.5), not httpclient. The httpclient:4.5.2 and httpclient:4.4.1 both depend on versions of httpcore which are 4.4.4 or lower, so there doesn't seem to be a backwards incompatibility problem there.
I do agree that it would be nice to have consistent versions, but I feel that is out of scope here and can be done in a separate PR which pulls in new versions for a bunch of other dependencies as well.

prateekm:
@cameronlee314 @shanthoosh What was the conclusion? Should this be merged?

cameronlee314:
This can be merged. Sorry for not following up on it earlier.

PR #535, opened 2018-05-25
SAMZA-1730: Adding state  valiations in StreamProcessor before any lifecycle operation and group coordination.

shanthoosh:


prateekm:
@shanthoosh Minor: Please create a SAMZA ticket for this and move the details there. Lengthy descriptions aren't appropriate for commit messages. Thanks!

shanthoosh:
@prateekm 
I created a ticket before creating the PR. Forgot to link to PR. Thanks for pointing it out. Fixed.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -178,20 +248,25 @@ public void start() {
    */
   public void stop() {
     synchronized (lock) {
-      boolean containerShutdownInvoked = false;
-      if (container != null) {
+      List<State> expectedStates = ImmutableList.of(State.NEW, State.RUNNING, State.STARTED, State.IN_REBALANCE);
```
sborya: Stop can be called in any state. So this check just needs to see if you are already STOPPING.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -200,80 +275,94 @@ SamzaContainer createSamzaContainer(String processorId, JobModel jobModel) {
...
+              LOGGER.warn("Container: {} shutdown was unsuccessful. Stopping the stream processor: {}.", container, processorId);
+              state = State.STOPPING;
+              jobCoordinator.stop();
             } else {
-              LOGGER.info("Container: {} of the stream processor: {} is not running.", container, processorId);
+              LOGGER.info("Container: {} shutdown completed for stream processor: {}.", container, processorId);
             }
           } else {
-            LOGGER.info("Container is not instantiated for stream processor: {}.", processorId);
+            LOGGER.info("Container: {} of the stream processor: {} is not running.", container, processorId);
```
sborya: Please add what command we are processing. Like "Cannot rebalance, because..."

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -200,80 +275,94 @@ SamzaContainer createSamzaContainer(String processorId, JobModel jobModel) {
...
       @Override
       public void onCoordinatorStop() {
-        if (executorService != null) {
+        synchronized (lock) {
           LOGGER.info("Shutting down the executor service of the stream processor: {}.", processorId);
+          stopSamzaContainer();
           executorService.shutdownNow();
+          state = State.STOPPED;
         }
-        if (processorListener != null) {
```
sborya: if you don't want to check processListener for null here, then add this check in constructor (or assert)

samza-core/src/main/java/org/apache/samza/standalone/PassthroughJobCoordinator.java
```diff
@@ -89,6 +89,7 @@ public void start() {
     }
     if (jobModel != null && jobModel.getContainers().containsKey(processorId)) {
       if (coordinatorListener != null) {
+        coordinatorListener.onJobModelExpired();
```
sborya: What do we need this event for?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
shanthoosh: Makes sense. Made the change.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkLocalApplicationRunner.java
```diff
@@ -566,12 +558,85 @@ public void testRollingUpgradeOfStreamApplicationsShouldGenerateSameJobModel() t
...
+
+    processedMessagesLatch3.await();
+
+    /**
+     * If the processing has started in the third stream processor, then other two stream processors should be stopped.
+     */
+    // TODO: This is a bug! Status should be unsuccessful finish.
+    assertEquals(applicationRunner1.status(streamApp1), ApplicationStatus.SuccessfulFinish);
+    assertEquals(applicationRunner2.status(streamApp2), ApplicationStatus.SuccessfulFinish);
+
```
sborya: do we need to check that processor 3 consumed all the events?

shanthoosh:
@vjagadish1989 
Can you please take a look at this PR. This has been linger around here for a while.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkLocalApplicationRunner.java
shanthoosh: I think we already check if processors consume messages in other tests after a re-balance. IMHO. Adding that here will be unnecessary.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -287,46 +376,36 @@ SamzaContainer getContainer() {
...
       if (!processorOnStartCalled) {
-        // processorListener is called on start only the first time the container starts.
-        // It is not called after every re-balance of partitions among the processors
+        processorListener.onStart();
         processorOnStartCalled = true;
-        if (processorListener != null) {
-          processorListener.onStart();
-        }
-      } else {
-        LOGGER.warn("Received duplicate container start notification for container: {} in stream processor: {}.", container, processorId);
```
vjagadish1989: is this log line not needed? It seems like it'll be useful to log duplicate notifications?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -178,20 +249,24 @@ public void start() {
...
-          container.shutdown();
-          containerShutdownInvoked = true;
-        } catch (Exception exception) {
-          LOGGER.error(String.format("Ignoring the exception during the shutdown of container: %s.", container), exception);
+          boolean hasContainerShutdown = stopSamzaContainer();
+          if (!hasContainerShutdown) {
+            LOGGER.info("Interrupting the container: {} thread to die.", container);
+            executorService.shutdownNow();
+            LOGGER.info("Waiting {} ms for the container:{} thread to die.", taskShutdownMs, container);
+            executorService.awaitTermination(taskShutdownMs, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
shanthoosh: Makes sense. Will change it.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
shanthoosh: `SamzaContainer` triggers `containerListener.onContainerStart` after starting all its components.

When `SamzaContainer` is stopped/started multiple times by `StreamProcessor`, each `containerStart` callback from different `SamzaContainer` instances  are not duplicate. 

The boolean variable guard is to prevent firing multiple `onProcessorStart` callbacks to LocalApplicationRunner layer and that deleted log line was a bug(Existing SamzaContainer implementation doesn't trigger duplicate `onContainerStart` callback within its single lifecycle).

vjagadish1989:
merged and submitted thanks!

PR #536, opened 2018-05-26
SAMZA-1731: Daemonize threads spawned from SamzaEventHubClientManager.

shanthoosh:
**Problem:**
Existing SamzaEventHubClientManager implementation spawns non-daemon threads when instantiating EventHubClient. These non-daemon threads stalls shutdown of samza processes even when the main thread has exited.

**Fix:** 
Daemonize threads that are spawned from SamzaEventHubClientManager.


shanthoosh:
@vjagadish1989 Can you please take a look when you get a chance.

vjagadish1989 (approved):
looks good, approved; thanks for the contribution!

PR #539, opened 2018-05-30
SAMZA-1736: Add counters and timers for batch get/put/delete operations in KeyValueStorageEngine

prateekm:


prateekm:
cc @vjagadish1989 @xinyuiscool @sborya for review.

samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
```diff
@@ -188,7 +189,7 @@ class RocksDbKeyValueStore(
 
   def delete(key: Array[Byte]): Unit = ifOpen {
     metrics.deletes.inc
-    put(key, null)
```
prateekm: This prevents deletes from incrementing put metrics.

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
```diff
@@ -134,7 +147,11 @@ class KeyValueStorageEngine[K, V](
     info(count + " total entries restored.")
 
     if (batch.size > 0) {
-      rawStore.putAll(batch)
+      updateTimer(metrics.putAllNs) {
```
cameronlee314: There are multiple places where the putAll metrics are being updated in the same way. Could you consolidate into a "rawPutAll"? Or even have a doPutAll(keyValueStore, batch) which covers the regular putAll also.

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngineMetrics.scala
```diff
@@ -28,27 +28,32 @@ class KeyValueStorageEngineMetrics(
   val registry: MetricsRegistry = new MetricsRegistryMap) extends MetricsHelper {
 
   val gets = newCounter("gets")
-  val ranges = newCounter("ranges")
-  val alls = newCounter("alls")
+  val getAlls = newCounter("getAlls")
```
cameronlee314: Minor: Should this be "get-alls" for consistency?

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
```diff
@@ -110,17 +119,21 @@ class KeyValueStorageEngine[K, V](
       batch.add(new Entry(keyBytes, valBytes))
 
       if (batch.size >= batchSize) {
-        rawStore.putAll(batch)
+        updateTimer(metrics.putAllNs) {
```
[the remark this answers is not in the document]
prateekm: Yeah, only happens during bootstrap. We can think of it as counting putAll operations from the KeyValueStorageEngine. Raw store currently does not have any timers, and this class is missing a putAll timer, so I updated it here.

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngineMetrics.scala
prateekm: Good catch, thanks.

samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
prateekm: Either way is fine. One way to interpret these metrics is as call counts to put/delete methods (this PR). Another is to interpret them as DB operations (your suggestion). I can update them to be db operations to be consistent with putAll (it increments puts and deletes as well).

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
prateekm: Sure, will do.

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
```diff
@@ -52,8 +52,11 @@ class KeyValueStorageEngine[K, V](
   }
 
   override def getAll(keys: java.util.List[K]): java.util.Map[K, V] = {
-    metrics.gets.inc(keys.size)
-    wrapperStore.getAll(keys)
+    updateTimer(metrics.getAllNs) {
```
shanthoosh: Does it makes sense to add metric assertions for new counters to existing `TestKeyValueStorageEngine` test

samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
```diff
@@ -154,18 +154,20 @@ class RocksDbKeyValueStore(
...
       db.put(writeOptions, key, value)
     }
   }
 
   // Write batch from RocksDB API is not used currently because of: https://github.com/facebook/rocksdb/issues/262
   def putAll(entries: java.util.List[Entry[Array[Byte], Array[Byte]]]): Unit = ifOpen {
+    metrics.putAlls.inc()
     val iter = entries.iterator
     var wrote = 0
     var deletes = 0
```
shanthoosh: Does it make sense to update `wrote` counter only in the else part where we update keys(similar to your change in put api).

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
prateekm: There are no existing tests for getAll/putAll etc. Since this change doesn't change the existing logic, and IMHO metrics update isn't something worth testing by itself, I'd prefer to skip those for now.

samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
prateekm: I was trying to avoid changing existing metrics in this PR, but sure, will change to make it consistent.

vjagadish1989 (approved):
approved

PR #542, opened 2018-05-31
JIRA: SAMZA-1733 Specify MetricsSnapshotStream partitionKey in config

rayman7718:
1. Adding config to specify MetricsSnapshotStream partitionKey, defaults to hostname (to maintain backword compat)


samza-core/src/main/scala/org/apache/samza/config/MetricsConfig.scala
```diff
@@ -20,16 +20,19 @@
...
   // metrics config constants
   val METRICS_REPORTERS = "metrics.reporters"
   val METRICS_REPORTER_FACTORY = "metrics.reporter.%s.class"
   val METRICS_SNAPSHOT_REPORTER_STREAM = "metrics.reporter.%s.stream"
-  val METRICS_SNAPSHOT_REPORTER_INTERVAL= "metrics.reporter.%s.interval"
-  val METRICS_TIMER_ENABLED= "metrics.timer.enabled"
+  val METRICS_SNAPSHOT_REPORTER_INTERVAL = "metrics.reporter.%s.interval"
+  val METRICS_TIMER_ENABLED = "metrics.timer.enabled"
+  val METRICS_SNAPSHOT_REPORTER_PARTITION_KEY = "metrics.reporter.%s.partitionkey"
+  val METRICS_SNAPSHOT_REPORTER_PARTITION_KEY_JOBNAME = "jobname"
```
vjagadish1989: s/partition_key_jobname/partition_key_default

samza-core/src/main/scala/org/apache/samza/config/MetricsConfig.scala
```diff
@@ -44,9 +47,36 @@ class MetricsConfig(config: Config) extends ScalaMapConfig(config) {
...
+    * Returns the actual value of the partition key, based on the keyName.
+    * If specified as "jobname" returns the jobname.
+    * If none is specified or is specified as "hostname", returns the hostname,
+    * else returns the actual value specified.
+    *
+    * @param keyName  the desired key name (hostname, jobname, none, or another custom name)
+    * @param jobName  the current job name
+    * @param hostname the current host name
+    */
+  def getMetricsReporterStreamPartitionKeyValue(keyName: Option[String], jobName: String, hostname: String) = keyName match {
```
vjagadish1989: prefer moving this method out of the MetricReporterConfig and instead adding it to `SnapshotReporter` instead.

samza-core/src/main/scala/org/apache/samza/config/MetricsConfig.scala
```diff
@@ -20,16 +20,18 @@
...
 object MetricsConfig {
   // metrics config constants
   val METRICS_REPORTERS = "metrics.reporters"
   val METRICS_REPORTER_FACTORY = "metrics.reporter.%s.class"
   val METRICS_SNAPSHOT_REPORTER_STREAM = "metrics.reporter.%s.stream"
-  val METRICS_SNAPSHOT_REPORTER_INTERVAL= "metrics.reporter.%s.interval"
-  val METRICS_TIMER_ENABLED= "metrics.timer.enabled"
+  val METRICS_SNAPSHOT_REPORTER_INTERVAL = "metrics.reporter.%s.interval"
+  val METRICS_TIMER_ENABLED = "metrics.timer.enabled"
+  val METRICS_SNAPSHOT_REPORTER_PARTITION_KEY = "metrics.reporter.%s.partitionkey"
```
nickpan47: Just curious, shouldn't we have a default value for this configuration variable?

samza-core/src/main/scala/org/apache/samza/metrics/reporter/MetricsSnapshotReporter.scala
```diff
@@ -25,15 +25,18 @@ import org.apache.samza.serializers.Serializer
...
+import org.apache.samza.util.{Logging, Util}
 import java.util.HashMap
 import java.util.Map
 import java.util.concurrent.Executors
 import java.util.concurrent.TimeUnit
 
 import scala.collection.JavaConverters._
 
+object MetricsSnapshotReporter {
+  val METRICS_SNAPSHOT_REPORTER_PARTITION_KEY_DEFAULT = "jobname"
```
nickpan47: We use a pattern where the default value is set in config class's getXXXConfig() methods, instead of the static defaults in the implementation class.

samza-core/src/main/scala/org/apache/samza/metrics/reporter/MetricsSnapshotReporter.scala
```diff
@@ -70,6 +71,25 @@ class MetricsSnapshotReporter(
...
+    * else returns the actual value specified.
+    *
+    * @param keyName  the desired key name (hostname, jobname, none, or another custom name)
+    * @param jobName  the current job name
+    * @param hostname the current host name
+    */
+  def getMetricsReporterStreamPartitionKeyValue(keyName: String, jobName: String, hostname: String) = keyName match {
+    case "jobname" => jobName
+    case "hostname" => hostname
+    case _ => keyName
```
nickpan47: Is this keyName or keyValue? The javadoc says "else returns the actual value specified", but the input argument only have key name. In addition, if "jobname" and "hostname" are special key names, I would prefer to define those as constant in MetricsConfig.

PR #543, opened 2018-06-02
JIRA: SAMZA-1733 Populating ListGauge metric using DiagnosticsAppender for exceptions

rayman7718:
This PR shows how the ListGauge can be used to emit exceptions using a DiagnosticsAppender.
1. DiagnosticsAppender is enabled using a config (diagnostics.appender.enable)
2. DiagnosticsAppender adds exception-events to a listgauge which is a samza container metric
2. This ListGauge uses a time-and-count based eviction policy, so that exception-events are not emitted to Kafka(SnapshotReporter) forever.


samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -869,6 +872,19 @@ class SamzaContainer(
     })
   }
 
+  def startDiagnostics {
+    // TODO: where should this reside, MetricConfig? Log4jSystemConfig? or a new separate config?
+    val DIAGNOSTICS_APPENDER_ENABLE = "diagnostics.appender.enable"
```
vjagadish1989: s/diagnostics.appender.enable/samza.diagnostics.enabled

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsExceptionEvent.java
```diff
@@ -0,0 +1,86 @@
...
+    this.threadName = threadName;
+    this.compactStackTrace = compactStackTrace;
+    this.stackTraceIdentifier = stackTraceIdentifier;
+  }
+
+  public long getTimestamp() {
+    return timestamp;
+  }
+
+  public void setTimestamp(long timestamp) {
```
vjagadish1989: probably unnecessary to expose setters

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsAppender.java
```diff
@@ -0,0 +1,81 @@
...
+ * Provides an in-memory appender that parses LoggingEvents to filter events relevant to diagnostics.
+ * Currently, filters exception related events and update an exception metric ({@link ListGauge}) in
+ * {@link SamzaContainerMetrics}.
+ *
+ * When used inconjunction with {@link org.apache.samza.metrics.reporter.MetricsSnapshotReporter} provides a
+ * stream of diagnostics-related events.
+ */
+public class DiagnosticsAppender extends AppenderSkeleton {
+
+  private final Logger logger = LoggerFactory.getLogger(this.getClass().getName());
```
nickpan47: We have been using the code convention like: logger = LoggerFactory.getLog(DiagnosticsAppender.class.getName()).

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsAppender.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Override
+  protected void append(LoggingEvent loggingEvent) {
+
+    try {
+
+      // if an event with a non-null throwable is received => exception event
+      if (loggingEvent.getThrowableInformation() != null) {
+        DiagnosticsExceptionEvent diagnosticsExceptionEvent =
+            new DiagnosticsExceptionEvent(loggingEvent.timeStamp, loggingEvent.getThrowableInformation());
```
nickpan47: So, how do we get the context info regarding to the exception? Through the stack trace? How do we add the logic component context like sub-query, task instance, container, and thread context?

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsAppender.java
```diff
@@ -0,0 +1,81 @@
...
+ * Currently, filters exception related events and update an exception metric ({@link ListGauge}) in
+ * {@link SamzaContainerMetrics}.
+ *
+ * When used inconjunction with {@link org.apache.samza.metrics.reporter.MetricsSnapshotReporter} provides a
+ * stream of diagnostics-related events.
+ */
+public class DiagnosticsAppender extends AppenderSkeleton {
+
+  private final Logger logger = LoggerFactory.getLogger(this.getClass().getName());
+  private final ListGauge<DiagnosticsExceptionEvent> samzaContainerExceptionMetric;
```
nickpan47: Question: how is this metric serialized and sent to diagnostic topic? If yes, shouldn't DiagnosticsExceptionEvent be serializable?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -904,6 +906,19 @@ class SamzaContainer(
...
+  def startDiagnostics {
+    // TODO: where should this reside, MetricConfig? Log4jSystemConfig? or a new separate config?
+    val DIAGNOSTICS_APPENDER_ENABLE = "samza.diagnostics.enabled"
+    if (containerContext.config.getBoolean(DIAGNOSTICS_APPENDER_ENABLE, false)) {
+      import org.apache.log4j.Logger
+      val rootLogger = Logger.getRootLogger
+
+      info("Starting Diagnostics Appender.")
+      val diagnosticsAppender = new DiagnosticsAppender(this.metrics)
+      rootLogger.addAppender(diagnosticsAppender)
```
shanthoosh: In standalone, there're plans to run more than one `StreamProcessor` in a jvm(which translates to more than one `SamzaContainer` thread in jvm). 
So it will be better to add diagnosticsAppender to rootLogger if it is not already added.

rayman7718:
Made the following changes: 
 - Depend directly on using java objects rather than log4j objects in appender (done)
 - Check if logger already attached, only than attach appender to root logger (done)
 - JobConfig for samza.diagnostics.enabled (done)

 - Obtain/store an exception's context in appender 
1. Container-level context is already present in MetricsHeaders. 
2. Adding TaskContext to SamzaException didnt quite work because TaskContext doesn't apply to many components (like KafkaSysCons).
3. Instead, we use the MDC map associated with the exception. Context can then be added by the components to MDC map, e.g., topicName, espressoTableName, etc. LOg4j's LoggingEvent takes care of multi-threading and async-appender.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsAppender.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Provides an in-memory appender that parses LoggingEvents to filter events relevant to diagnostics.
+ * Currently, filters exception related events and update an exception metric ({@link ListGauge}) in
+ * {@link SamzaContainerMetrics}.
+ *
+ * When used inconjunction with {@link org.apache.samza.metrics.reporter.MetricsSnapshotReporter} provides a
+ * stream of diagnostics-related events.
+ */
+public class DiagnosticsAppender extends AppenderSkeleton {
```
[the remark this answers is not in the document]
rayman7718: the same implementation can be adopted, as of log4j2.4.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -56,6 +56,8 @@ import org.apache.samza.task._
 import org.apache.samza.util.Util
 import org.apache.samza.util._
 import org.apache.samza.{SamzaContainerStatus, SamzaException}
+import org.apache.samza.diagnostics.DiagnosticsAppender;
+import org.apache.log4j.Logger
```
nickpan47: We are not using log4j directly in samza-core. We always use slf4j libraries in all our source code, except for the samza-log4j module.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -904,6 +907,17 @@ class SamzaContainer(
     })
   }
 
+  def startDiagnostics {
+    if (containerContext.config.getDiagnosticsEnabled) {
+      val rootLogger = Logger.getRootLogger
+
+      if (rootLogger.getAppender(classOf[DiagnosticsAppender].getName) == null) {
```
nickpan47: It seems that in log4j2, the API to get appender dynamically also changed: https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/config/Configuration.html#getAppender(java.lang.String). Please check w/ @PawasChhokra to see what's the best way to make sure the code here is log4j version agnostic.

nickpan47: Shouldn't this condition be (rootLogger.getAppender(classOf[DiagnosticsAppender].getName) *!=* null)?

samza-core/src/main/scala/org/apache/samza/metrics/reporter/Metrics.scala
```diff
@@ -19,9 +19,11 @@
 
 package org.apache.samza.metrics.reporter
 
+import java.util
 import java.util.Collections
 import java.util.HashMap
 import java.util.Map
+
```
nickpan47: nit: whitespace change should be avoided.

samza-core/src/main/scala/org/apache/samza/metrics/reporter/Metrics.scala
```diff
@@ -52,4 +54,8 @@ class Metrics(metrics: Map[String, Map[String, Object]]) {
   def get(group: String) = immutableMetrics.get(group)
 
   def getAsMap(): Map[String, Map[String, Object]] = Collections.unmodifiableMap(immutableMetrics)
+
+  def this() {
```
nickpan47: I did a code search and couldn't find any reference to this default constructor. What's the reason to add it?

samza-core/src/main/scala/org/apache/samza/metrics/reporter/Metrics.scala
```diff
@@ -19,9 +19,11 @@
 
 package org.apache.samza.metrics.reporter
 
+import java.util
```
nickpan47: This seems to be redundant, if we remove the prefix to HashMap in line 59.

rayman7718:
- Addressed the log4j dependency issue by using a config for the Appender class, this removes the use of log4j Logger from SamzaContainer
- For log4j2, verified with Pawas that the getappender and addAppender apis are consistent. 

- Addressed the other Metrics comments.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -186,4 +192,8 @@ class JobConfig(config: Config) extends ScalaMapConfig(config) with Logging {
   def getNonLoggedStorePath = getOption(JobConfig.JOB_NON_LOGGED_STORE_BASE_DIR)
 
   def getLoggedStorePath = getOption(JobConfig.JOB_LOGGED_STORE_BASE_DIR)
+
+  def getDiagnosticsEnabled = { getBoolean(JobConfig.JOB_DIAGNOSTICS_ENABLED, false) }
+
+  def getDiagnosticsAppenderClass = { getOption(JobConfig.DIAGNOSTICS_APPENDER_CLASS) }
```
nickpan47: I think that it would make sense to provide a default class name here as well, for log4j 1.2.x (since in open source, we still depend on log4j 1.2.x).

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -904,6 +903,14 @@ class SamzaContainer(
     })
   }
 
+  def startDiagnostics {
+    if (containerContext.config.getDiagnosticsEnabled) {
+      info("Starting diagnostics.")
+      val diagnosticsAppender = Class.forName(containerContext.config.getDiagnosticsAppenderClass.get).
```
nickpan47: It would be nice to capture the class loading exceptions and log a line here, then throw them as ConfigException(msg, e).

samza-log4j/src/main/java/org/apache/samza/logging/log4j/SimpleDiagnosticsAppender.java
```diff
@@ -0,0 +1,89 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(SimpleDiagnosticsAppender.class);
+  protected final ListGauge<DiagnosticsExceptionEvent> samzaContainerExceptionMetric;
+
+  public SimpleDiagnosticsAppender(SamzaContainerMetrics samzaContainerMetrics) {
+    this.samzaContainerExceptionMetric = samzaContainerMetrics.exceptions();
+    this.setName(SimpleDiagnosticsAppender.class.getName());
+
+    // ensure appender is attached only once per JVM (regardless of #containers)
+    if (org.apache.log4j.Logger.getRootLogger().getAppender(SimpleDiagnosticsAppender.class.getName()) == null) {
```
nickpan47: Just realize: is this code section safe for multi-threaded environment? Ideally, we should make it threadsafe. If not, we better make 100% sure that the loading of this appender only happens in one single thread and make the usage documentation here *very clear*.

rayman7718:
- Added simple synchronization to make appender attachment thread safe (from constructor). 
- Added default value for appender class path in job config. 
- Adding error logging in samza-container for appender instantiation.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -186,4 +192,10 @@ class JobConfig(config: Config) extends ScalaMapConfig(config) with Logging {
   def getNonLoggedStorePath = getOption(JobConfig.JOB_NON_LOGGED_STORE_BASE_DIR)
 
   def getLoggedStorePath = getOption(JobConfig.JOB_LOGGED_STORE_BASE_DIR)
+
+  def getDiagnosticsEnabled = { getBoolean(JobConfig.JOB_DIAGNOSTICS_ENABLED, false) }
+
+  def getDiagnosticsAppenderClass = {
+    getOrDefault(JobConfig.DIAGNOSTICS_APPENDER_CLASS, "org.apache.samza.logging.log4j.SimpleDiagnosticsAppender")
```
nickpan47: nit: define a static constant for the default value in the class.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -904,6 +904,23 @@ class SamzaContainer(
...
+      info("Starting diagnostics.")
+
+      try {
+        val diagnosticsAppender = Class.forName(containerContext.config.getDiagnosticsAppenderClass).
+          getDeclaredConstructor(classOf[SamzaContainerMetrics]).newInstance(this.metrics);
+      }
+      catch {
+        case e@(_: ClassNotFoundException | _: InstantiationException | _: InvocationTargetException) => {
+          error("Failed to instantiate diagnostic appender", e)
+          throw new ConfigException("Failed to instantiate diagnostic appender class specified in config", e)
```
nickpan47: nit: add the configured DiagnosticsAppenderClass name in the log and exception would make it easier for debugging.

PR #544, opened 2018-06-05
SAMZA-929 : Set initialDelay in tokenRenewExecutor schedule to 0

apoorva121:
Changed initialDelay in tokenRenewExecutor scheduler to 0 so that it can re-login using the keytab as soon as the application master container starts.  This way even if application master restarts after the delegation token in launcher context has expired, it will be able to use the new token to launch other containers.


vjagadish1989 (approved):
approved. Thanks for the fix!

vjagadish1989:
merged and submitted, thanks much @apoorva121 for your contribution!

PR #545, opened 2018-06-05
SAMZA-1740: Moving SamzaSqlRelRecord to samza-api as it is needed to be used in UDFs

atoomula:
Please see description in the ticket. Also, implementing equals and hashCode methods for SamzaSqlRelRecord and SamzaSqlRelMessage.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/sql/SamzaSqlRelRecord.java
+++ b/samza-api/src/main/java/org/apache/samza/sql/SamzaSqlRelRecord.java
@@ -22,6 +22,7 @@ package org.apache.samza.sql;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import org.codehaus.jackson.annotate.JsonProperty;
 
@@ -93,11 +94,7 @@ public class SamzaSqlRelRecord implements Serializable {
 
   @Override
   public int hashCode() {
-    final int prime = 31;
-    int result = 1;
-    result = prime * result + ((fieldNames == null) ? 0 : fieldNames.hashCode());
-    result = prime * result + ((fieldValues == null) ? 0 : fieldValues.hashCode());
-    return result;
+    return Objects.hash(fieldNames, fieldValues);
   }
 
   @Override
@@ -109,16 +106,6 @@ public class SamzaSqlRelRecord implements Serializable {
     if (getClass() != obj.getClass())
       return false;
     SamzaSqlRelRecord other = (SamzaSqlRelRecord) obj;
-    if (fieldNames == null) {
-      if (other.fieldNames != null)
-        return false;
-    } else if (!fieldNames.equals(other.fieldNames))
-      return false;
-    if (fieldValues == null) {
-      if (other.fieldValues != null)
-        return false;
-    } else if (!fieldValues.equals(other.fieldValues))
-      return false;
-    return true;
+    return Objects.equals(fieldNames, other.fieldNames) && Objects.equals(fieldValues, other.fieldValues);
   }
 }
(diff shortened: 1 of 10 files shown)


samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -110,67 +105,34 @@ public Object getKey() {
...
-    public List<Object> getFieldValues() {
-      return this.fieldValues;
-    }
+  @Override
+  public int hashCode() {
+    final int prime = 31;
+    int result = 1;
+    result = prime * result + ((key == null) ? 0 : key.hashCode());
+    result = prime * result + ((samzaSqlRelRecord == null) ? 0 : samzaSqlRelRecord.hashCode());
+    return result;
```
pdu-mn1: would Objects.hashCode or HashCodeBuilder work?

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -110,67 +105,34 @@ public Object getKey() {
...
-        if (fieldNames.get(index).equals(name)) {
-          return Optional.ofNullable(fieldValues.get(index));
-        }
-      }
-
-      return Optional.empty();
-    }
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj)
```
pdu-mn1: Objects.equals() ?

atoomula: Good suggestion. Changed to that.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
atoomula: Good suggestion. Changed to Objects.hashCode.

samza-api/src/main/java/org/apache/samza/sql/SamzaSqlRelRecord.java
```diff
@@ -0,0 +1,111 @@
...
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import org.codehaus.jackson.annotate.JsonProperty;
+
+
+/**
```
nickpan47: One suggestion: if this is going to be a public API for UDF implementation, please add interface stability annotations if we are not fully settled with this class (i.e. something like @InterfaceStability.Unstable).

samza-api/src/main/java/org/apache/samza/sql/SamzaSqlRelRecord.java
```diff
@@ -0,0 +1,111 @@
...
+ * Samza sql relational record. A record consists of list of column values and the associated column names.
+ * A column value could be nested, meaning, it could be another SamzaSqlRelRecord.
+ * Right now we do not store any metadata (like nullability, etc) other than the column name in the SamzaSqlRelRecord.
+ */
+public class SamzaSqlRelRecord implements Serializable {
+
+  @JsonProperty("fieldNames")
+  private final List<String> fieldNames;
```
nickpan47: Since this class is implementing Serializable, it is better to force the type of List here to a concrete serializable implementation of List (e.g. ArrayList or LinkedList).

samza-api/src/main/java/org/apache/samza/sql/SamzaSqlRelRecord.java
atoomula: Sure

samza-api/src/main/java/org/apache/samza/sql/SamzaSqlRelRecord.java
atoomula: Makes sense.

PR #546, opened 2018-06-06
Minor: KafkaConfig should treat empty changelog name as no changelog.

prateekm:
If a store changelog stream name is empty, treat is as a non-changelogged store instead of throwing an exception.


prateekm:
cc. @vjagadish1989 for review.

PR #547, opened 2018-06-06
Convert a put to to a delete operation in ReadWriteTable and TableWriteFunction when input value is null

weisong44:
Currently, the behavior of putting a null value is inconsistent: it is a delete for RocksDB, and not supported in in-memory store, and on a case-by-case basis for remote tables. It is desirable to unify the behavior. Furthermore, it eases the writing of a change captured stream to a table. A change captured stream contains typically 3 types of events: INSERT, UPDATE and DELETE, and they need to be applied properly when written to a table to produce a correct snapshot. In a change captured stream the payload of a DELETE event is typically is null, and this would result in a delete operation to a table in sendTo() operator.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/table/ReadWriteTable.java
+++ b/samza-api/src/main/java/org/apache/samza/table/ReadWriteTable.java
@@ -19,7 +19,6 @@
 package org.apache.samza.table;
 
 import java.util.List;
-
 import org.apache.samza.annotation.InterfaceStability;
 import org.apache.samza.storage.kv.Entry;
 
@@ -36,17 +35,21 @@ public interface ReadWriteTable<K, V> extends ReadableTable<K, V> {
   /**
    * Updates the mapping of the specified key-value pair; Associates the specified {@code key} with the specified {@code value}.
    *
+   * The key is deleted from the table if value is {@code null}.
+   *
    * @param key the key with which the specified {@code value} is to be associated.
    * @param value the value with which the specified {@code key} is to be associated.
-   * @throws NullPointerException if the specified {@code key} or {@code value} is {@code null}.
+   * @throws NullPointerException if the specified {@code key} is {@code null}.
    */
   void put(K key, V value);
 
   /**
    * Updates the mappings of the specified key-value {@code entries}.
    *
+   * A key is deleted from the table if its corresponding value is {@code null}.
+   *
    * @param entries the updated mappings to put into this table.
-   * @throws NullPointerException if any of the specified {@code entries} has {@code null} as key or value.
+   * @throws NullPointerException if any of the specified {@code entries} has {@code null} as key.
    */
   void putAll(List<Entry<K, V>> entries);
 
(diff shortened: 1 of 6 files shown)


samza-core/src/main/java/org/apache/samza/table/caching/guava/GuavaCacheTable.java
```diff
@@ -19,19 +19,17 @@
 
 package org.apache.samza.table.caching.guava;
 
+import com.google.common.cache.Cache;
```
pdu-mn1: nit: unnecessary change

samza-core/src/main/java/org/apache/samza/table/remote/RemoteReadWriteTable.java
```diff
@@ -19,8 +19,8 @@
 
 package org.apache.samza.table.remote;
 
+import com.google.common.base.Preconditions;
```
pdu-mn1: nit: unnecessary change

samza-core/src/main/java/org/apache/samza/table/remote/TableWriteFunction.java
```diff
@@ -44,13 +43,19 @@
 public interface TableWriteFunction<K, V> extends Serializable, InitableFunction, ClosableFunction {
   /**
    * Store single table {@code record} with specified {@code key}. This method must be thread-safe.
+   *
+   * The key is deleted, if record is {@code null}.
```
pdu-mn1: nit: add "its corresponding" to match the previous doc

samza-kv/src/main/java/org/apache/samza/storage/kv/LocalStoreBackedReadWriteTable.java
```diff
@@ -19,7 +19,6 @@
 package org.apache.samza.storage.kv;
 
 import java.util.List;
-
```
pdu-mn1: nit: unnecessary change

samza-core/src/main/java/org/apache/samza/table/remote/RemoteReadWriteTable.java
```diff
@@ -30,9 +30,7 @@
 import org.apache.samza.task.TaskContext;
 import org.apache.samza.util.RateLimiter;
 
-import com.google.common.base.Preconditions;
-
-import static org.apache.samza.table.remote.RemoteTableDescriptor.RL_WRITE_TAG;
+import static org.apache.samza.table.remote.RemoteTableDescriptor.*;
```
pdu-mn1: nit: unnecessary change

samza-core/src/main/java/org/apache/samza/table/remote/TableWriteFunction.java
weisong44: There is only one key/value pair, hence no mentioning.

samza-api/src/main/java/org/apache/samza/table/ReadWriteTable.java
```diff
@@ -36,17 +35,21 @@
   /**
    * Updates the mapping of the specified key-value pair; Associates the specified {@code key} with the specified {@code value}.
    *
+   * The key is deleted from the table, if value is {@code null}.
```
prateekm: Nitpick: no comma before if everywhere.

weisong44: +1

weisong44:
I've updated the PR based on comments, can you kindly merge it?

PR #548, opened 2018-06-06
SAMZA-1741: fix issue that EH consumer taking too long to shutdown

lhaiesp:
1.  lower the shutdown timeout from 1 min to 15 seconds
2. make sure EventHubManagers are shutdown in parallel
3. print a thread dump when we do fail during shutdown

### DIFF
--- a/samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
@@ -36,6 +36,9 @@ import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -57,6 +60,7 @@ import org.apache.samza.system.eventhub.admin.EventHubSystemAdmin;
 import org.apache.samza.system.eventhub.metrics.SamzaHistogram;
 import org.apache.samza.system.eventhub.producer.EventHubSystemProducer;
 import org.apache.samza.util.BlockingEnvelopeMap;
+import org.apache.samza.util.Util;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -102,7 +106,7 @@ public class EventHubSystemConsumer extends BlockingEnvelopeMap {
 
   // Overall timeout for EventHubClient exponential backoff policy
   private static final Duration DEFAULT_EVENTHUB_RECEIVER_TIMEOUT = Duration.ofMinutes(10L);
-  private static final long DEFAULT_SHUTDOWN_TIMEOUT_MILLIS = Duration.ofMinutes(1L).toMillis();
+  private static final long DEFAULT_SHUTDOWN_TIMEOUT_MILLIS = Duration.ofSeconds(15).toMillis();
 
   public static final String START_OF_STREAM = ClientConstants.START_OF_STREAM; // -1
   public static final String END_OF_STREAM = "-2";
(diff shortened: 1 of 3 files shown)


samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -354,15 +358,29 @@ public void stop() {
...
     streamPartitionReceivers.values().forEach((receiver) -> futures.add(receiver.close()));
-    CompletableFuture<Void> future = CompletableFuture.allOf(futures.toArray(new CompletableFuture[futures.size()]));
+    CompletableFuture<Void> allFuture = CompletableFuture.allOf(futures.toArray(new CompletableFuture[futures.size()]));
     try {
-      future.get(DEFAULT_SHUTDOWN_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
-    } catch (ExecutionException | InterruptedException | TimeoutException e) {
+      allFuture.get(DEFAULT_SHUTDOWN_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
+    } catch (Exception e) {
       LOG.warn("Failed to close receivers", e);
+      Util.logThreadDump("Log thread dump as we failed during consumer shutdown");
```
vjagadish1989: the string is meant to be used as an id: maybe name it as - "Thread dump for EventHubSystemConsumer.PartitionReceiver#close failure"

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -354,15 +358,29 @@ public void stop() {
...
+    LOG.info("Receivers are shutdown. Start shutting down EventHubManagers");
+    // There could be potentially many EventHubManagers, so close the managers in parallel
+    ExecutorService shutdownExecutorService = Executors.newCachedThreadPool();
+    List<Future> futureList = new ArrayList<>();
     perPartitionEventHubManagers.values()
-        .parallelStream()
-        .forEach(ehClientManager -> ehClientManager.close(DEFAULT_SHUTDOWN_TIMEOUT_MILLIS));
+        .forEach(ehClientManager -> futureList.add(
+            shutdownExecutorService.submit(() -> ehClientManager.close(DEFAULT_SHUTDOWN_TIMEOUT_MILLIS))));
+    futureList.forEach(future -> {
```
vjagadish1989: Would it be possible to use  the same `CompletableFuture.allOf(DEFAULT_SHUTDOWN_TIMEOUT)` to aggregate the futures here?

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -354,15 +358,29 @@ public void stop() {
...
+      try {
+        future.get(); // ehClientManager.close() already has a timeout
```
vjagadish1989: Could actually extract this to an util method: 
```public static void waitForFutures(List<Future> futures, long timeout)``` that waits for the futures for a given timeout and logs the thread-dump in case of an exception?

Comment (lhaiesp) [reply samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java]: do you have some suggestion where to place this util method?

File: samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java (hunk 1 above)
Comment (lhaiesp) [reply samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java]: sure. fixing it

File: samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java (hunk 2 above)
Comment (lhaiesp) [reply samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java]: So the ehClientManager.close is already performing the timeout logic. That's sequently call "future.get()" shouldn't cause any issues. 
Plus, CompletableFuture.allOf only works for CompletableFuture<T> type

File: samza-core/src/main/java/org/apache/samza/util/ShutdownUtil.java [hunk 4]
```diff
@@ -0,0 +1,80 @@
...
+   *
+   * The shutdown is performed with best-effort. Depending on the implementation of the shutdown function, resource
+   * leak might be possible.
+   *
+   * @param shutdownFunction the shutdown function will be provided with an ExecutorService and is supposed to
+   *                         submit all the shutdown tasks via the given executor
+   * @param message message that will show in the thread name and the thread dump
+   * @param timeoutMs timeout in ms
+   */
+  public static void boundedShutdown(Function<ExecutorService, Void> shutdownFunction, String message, long timeoutMs) {
```
prateekm: 1. Does the caller need access to the executor service? Can they caller provide a `List<Runnable> tasks`.
2. [Optional] Can this be a general method unrelated to shutdown? Seems like it essentially takes a List<Runnable/Callable>, executes them concurrently and tries to cancel them and return after waiting for a bounded amount of time.

vjagadish1989: +1

vjagadish1989: additionally, can return the status as boolean if all `Runnable`s complete.

lhaiesp: Discussed with Prateek offline. Totally makes sense to change the signature to just take a list of Runnable. Will also return the status as boolean.

In terms of making it a general method for concurrent execution, that would probably go too far. I still think this is more specific for shutdown because of the fact that Samza requires the system consumer to shutdown within a timeout during rebalance. In general, people should simply use an ExecutorService for executing concurrent tasks in a normal way.  So I prefer keeping this as a shutdown util for now to discourage people to use it for general purpose.

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -352,17 +349,28 @@ private void renewPartitionReceiver(SystemStreamPartition ssp) {
...
+          LOG.error("Failed to shutdown receiver.", e);
+        }
+      }
+    }).collect(Collectors.toList()), "EventHubSystemConsumer.Receiver#close", DEFAULT_SHUTDOWN_TIMEOUT_MILLIS);
+
+    LOG.info("Start shutting down eventhubs managers");
+    ShutdownUtil.boundedShutdown(perPartitionEventHubManagers.values().stream().map(manager -> new Runnable() {
+      @Override
+      public void run() {
+        manager.close(DEFAULT_SHUTDOWN_TIMEOUT_MILLIS);
```
vjagadish1989: for safety wrap this in a try as well?

samza-core/src/main/java/org/apache/samza/util/ShutdownUtil.java
```diff
@@ -0,0 +1,82 @@
...
+   *
+   * @param shutdownTasks the list of shutdown tasks that need to be executed in parallel
+   * @param message message that will show in the thread name and the thread dump
+   * @param timeoutMs timeout in ms
+   * @return true if all tasks terminate in the end
+   */
+  public static boolean boundedShutdown(List<Runnable> shutdownTasks, String message, long timeoutMs) {
+    ExecutorService shutdownExecutorService = Executors.newCachedThreadPool(new ThreadFactory() {
+      private final AtomicInteger counter = new AtomicInteger(0);
+      @Override
```
vjagadish1989: cleaner to use gauva's `ThreadFactoryBuilder` instead of having to write our own. you could also remove the `AtomicInteger` then

PR #552, opened 2018-06-11
SAMZA-1745: Remove all usages of StreamSpec and ApplicationRunner from the operator spec and impl layers.

prateekm:
This PR is a pre-requisite for adding support for user-provided SystemDescriptors and StreamDescriptors to the High Level API.

It removes all usages of StreamSpec and ApplicationRunner from the OperatorSpec and OperatorImpl layers. DAG specification (StreamGraphSpec, OperatorSpecs) now only relies on logical streamIds (and in future, will use the user-provided StreamDescriptors). DAG execution (i.e., StreamOperatorTask, OperatorImpls) now only relies on logical streamIds and their corresponding SystemStreams, which are obtained using StreamConfig in OperatorImplGraph.

After this change, StreamSpec can be thought of as the API between StreamManager and SystemAdmins for creating and validating streams. Ideally ExecutionPlanner shouldn't rely on StreamSpec either, but it currently does so extensively, so I'll leave that refactor for later.

Additional changes:
1. ApplicationRunner is no longer responsible for creating/returning StreamSpec instances. Instances can be created directly using the StreamSpec constructors, or by using one of the util methods in the new StreamUtil class.

2. StreamSpec class no longer tracks the isBroadcast and isBounded status for streams. 
The former was being used for communicating broadcast status from the StreamGraphSpec to the planner so that it could write the broadcast input configurations. This is now done using a separate Set of broadcast streamIds in StreamGraphSpec.
The latter was being set by the ApplicationRunner based on a config, and then passed to the planner so that it could write the bounded input configs. This was redundant, so I removed it.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
+++ b/samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
@@ -34,6 +34,7 @@ import org.apache.samza.config.ApplicationConfig;
 import org.apache.samza.config.ClusterManagerConfig;
 import org.apache.samza.config.Config;
 import org.apache.samza.config.JobConfig;
+import org.apache.samza.config.StreamConfig;
 import org.apache.samza.operators.OperatorSpecGraph;
 import org.apache.samza.operators.spec.JoinOperatorSpec;
 import org.apache.samza.operators.spec.OperatorSpec;
@@ -42,6 +43,8 @@ import org.apache.samza.system.SystemStream;
 import org.apache.samza.table.TableSpec;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import static org.apache.samza.util.StreamUtil.*;
 
 
 /**
@@ -93,8 +96,9 @@ public class ExecutionPlanner {
    */
   /* package private */ JobGraph createJobGraph(OperatorSpecGraph specGraph) {
     JobGraph jobGraph = new JobGraph(config, specGraph);
-    Set<StreamSpec> sourceStreams = new HashSet<>(specGraph.getInputOperators().keySet());
-    Set<StreamSpec> sinkStreams = new HashSet<>(specGraph.getOutputStreams().keySet());
+    StreamConfig streamConfig = new StreamConfig(config);
+    Set<StreamSpec> sourceStreams = getStreamSpecs(specGraph.getInputOperators().keySet(), streamConfig);
+    Set<StreamSpec> sinkStreams = getStreamSpecs(specGraph.getOutputStreams().keySet(), streamConfig);
     Set<StreamSpec> intStreams = new HashSet<>(sourceStreams);
     Set<TableSpec> tables = new HashSet<>(specGraph.getTables().keySet());
     intStreams.retainAll(sinkStreams);
@@ -128,7 +132,7 @@ public class ExecutionPlanner {
    */
   /* package private */ void calculatePartitions(JobGraph jobGraph) {
     // calculate the partitions for the input streams of join operators
-    calculateJoinInputPartitions(jobGraph);
+    calculateJoinInputPartitions(jobGraph, config);
 
     // calculate the partitions for the rest of intermediate streams
     calculateIntStreamPartitions(jobGraph, config);
@@ -172,7 +176,7 @@ public class ExecutionPlanner {
   /**
    * Calculate the partitions for the input streams of join operators
    */
-  /* package private */ static void calculateJoinInputPartitions(JobGraph jobGraph) {
+  /* package private */ static void calculateJoinInputPartitions(JobGraph jobGraph, Config config) {
     // mapping from a source stream to all join specs reachable from it
     Multimap<OperatorSpec, StreamEdge> joinSpecToStreamEdges = HashMultimap.create();
     // reverse mapping of the above
@@ -183,10 +187,10 @@ public class ExecutionPlanner {
     Set<OperatorSpec> visited = new HashSet<>();
 
     jobGraph.getSpecGraph().getInputOperators().entrySet().forEach(entry -> {
-        StreamEdge streamEdge = jobGraph.getOrCreateStreamEdge(entry.getKey());
+        StreamConfig streamConfig = new StreamConfig(config);
+        StreamEdge streamEdge = jobGraph.getOrCreateStreamEdge(getStreamSpec(entry.getKey(), streamConfig));
         // Traverses the StreamGraph to find and update mappings for all Joins reachable from this input StreamEdge
-        findReachableJoins(entry.getValue(), streamEdge, joinSpecToStreamEdges, streamEdgeToJoinSpecs,
-            joinQ, visited);
+        findReachableJoins(entry.getValue(), streamEdge, joinSpecToStreamEdges, streamEdgeToJoinSpecs, joinQ, visited);
       });
 
     // At this point, joinQ contains joinSpecs where at least one of the input stream edge partitions is known.
@@ -203,7 +207,7 @@ public class ExecutionPlanner {
           } else if (partitions != edgePartitions) {
             throw  new SamzaException(String.format(
                 "Unable to resolve input partitions of stream %s for join. Expected: %d, Actual: %d",
-                edge.getFormattedSystemStream(), partitions, edgePartitions));
+                edge.getName(), partitions, edgePartitions));
           }
         }
       }
@@ -282,7 +286,7 @@ public class ExecutionPlanner {
   private static void validatePartitions(JobGraph jobGraph) {
     for (StreamEdge edge : jobGraph.getIntermediateStreamEdges()) {
       if (edge.getPartitionCount() <= 0) {
-        throw new SamzaException(String.format("Failure to assign the partitions to Stream %s", edge.getFormattedSystemStream()));
+        throw new SamzaException(String.format("Failure to assign the partitions to Stream %s", edge.getName()));
       }
     }
   }
--- a/samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
+++ b/samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
@@ -29,7 +29,7 @@ import org.apache.samza.config.MapConfig;
 import org.apache.samza.config.StreamConfig;
 import org.apache.samza.system.StreamSpec;
 import org.apache.samza.system.SystemStream;
-import org.apache.samza.util.Util;
+import org.apache.samza.util.StreamUtil;
 
 
 /**
@@ -41,23 +41,24 @@ public class StreamEdge {
   public static final int PARTITIONS_UNKNOWN = -1;
 
   private final StreamSpec streamSpec;
+  private final boolean isBroadcast;
+  private final boolean isIntermediate;
   private final List<JobNode> sourceNodes = new ArrayList<>();
   private final List<JobNode> targetNodes = new ArrayList<>();
   private final Config config;
+  private final String name;
 
-  private String name = "";
   private int partitions = PARTITIONS_UNKNOWN;
-  private final boolean isIntermediate;
 
-  StreamEdge(StreamSpec streamSpec, Config config) {
-    this(streamSpec, false, config);
-  }
-
-  StreamEdge(StreamSpec streamSpec, boolean isIntermediate, Config config) {
+  StreamEdge(StreamSpec streamSpec, boolean isIntermediate, boolean isBroadcast, Config config) {
     this.streamSpec = streamSpec;
-    this.name = Util.getNameFromSystemStream(getSystemStream());
     this.isIntermediate = isIntermediate;
+    this.isBroadcast = isBroadcast;
     this.config = config;
+    if (isBroadcast) {
+      partitions = 1;
+    }
+    this.name = StreamUtil.getNameFromSystemStream(getSystemStream());
   }
 
   void addSourceNode(JobNode sourceNode) {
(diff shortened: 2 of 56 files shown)


prateekm:
cc. @vjagadish1989 @nickpan47 @xinyuiscool @jmakes for review and feedback.

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
```diff
@@ -128,12 +121,13 @@ Config generateConfig() {
       config.put(String.format(StreamConfig.CONSUMER_OFFSET_DEFAULT_FOR_STREAM_ID(), spec.getId()), "oldest");
       config.put(String.format(StreamConfig.PRIORITY_FOR_STREAM_ID(), spec.getId()), String.valueOf(Integer.MAX_VALUE));
     }
-    if (spec.isBounded()) {
-      config.put(String.format(StreamConfig.IS_BOUNDED_FOR_STREAM_ID(), spec.getId()), "true");
```
cameronlee314: Did this get moved somewhere else?

samza-core/src/main/java/org/apache/samza/operators/StreamGraphSpec.java
```diff
@@ -55,13 +53,13 @@
...
+  private static final Pattern ID_PATTERN = Pattern.compile("[\\d\\w-_.]+");
 
   // We use a LHM for deterministic order in initializing and closing operators.
-  private final Map<StreamSpec, InputOperatorSpec> inputOperators = new LinkedHashMap<>();
-  private final Map<StreamSpec, OutputStreamImpl> outputStreams = new LinkedHashMap<>();
+  private final Map<String, InputOperatorSpec> inputOperators = new LinkedHashMap<>();
+  private final Map<String, OutputStreamImpl> outputStreams = new LinkedHashMap<>();
+  private final Set<String> broadcastStreams = new HashSet<>();
   private final Map<TableSpec, TableImpl> tables = new LinkedHashMap<>();
-  private final ApplicationRunner runner;
```
cameronlee314: Nice :)

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
```diff
@@ -323,70 +333,81 @@ public void init(Config config, TaskContext context) {
     };
   }
 
-  private boolean hasIntermediateStreams(OperatorSpecGraph specGraph) {
-    return !Collections.disjoint(specGraph.getInputOperators().keySet(), specGraph.getOutputStreams().keySet());
-  }
+  protected static class OperatorImplGraphUtil {
```
cameronlee314: Could you clarify the benefit of extracting this out?
One benefit I could see is if you wanted to mock this, but I don't see any mocking of this class.
It just looks a bit odd that you have a "util" class which you are using as an instance variable.

samza-core/src/main/java/org/apache/samza/util/StreamUtil.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  /**
+   * Converts the provided list of (streamId, system, physicalName) triplets to their corresponding
+   * stream.stream-id.* configurations.
+   *
+   * @param streams a list of (streamId, system, physicalName) triplets to get the stream configuration for.
+   * @return the configuration for the provided { @code streams}
+   */
+  public static Config toStreamConfigs(List<ImmutableTriple<String, String, String>> streams) {
```
cameronlee314: From what I can tell, this is just used in test code. Should we avoid mixing prod utils and test utils?
Also, would it be clearer to make some explicit container for this set of data instead of using a triple, especially because all of the types are String.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
```diff
@@ -217,9 +221,13 @@ OperatorImpl createOperatorImpl(OperatorSpec prevOperatorSpec, OperatorSpec oper
     } else if (operatorSpec instanceof SinkOperatorSpec) {
       return new SinkOperatorImpl((SinkOperatorSpec) operatorSpec, config, context);
     } else if (operatorSpec instanceof OutputOperatorSpec) {
-      return new OutputOperatorImpl((OutputOperatorSpec) operatorSpec);
+      String streamId = ((OutputOperatorSpec) operatorSpec).getOutputStream().getStreamId();
+      SystemStream systemStream = util.toSystemStream(streamId);
+      return new OutputOperatorImpl((OutputOperatorSpec) operatorSpec, systemStream);
     } else if (operatorSpec instanceof PartitionByOperatorSpec) {
-      return new PartitionByOperatorImpl((PartitionByOperatorSpec) operatorSpec, config, context);
+      String streamId = ((PartitionByOperatorSpec) operatorSpec).getOutputStream().getStreamId();
```
cameronlee314: Just to double check:
Does the planner rewrite the config to add intermediate stream info before this runs?

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
prateekm: These previously static util methods now need access to the `config` object to get the SystemStream for the streamId. I can either pass the StreamConfig through all of these explicitly and update all their tests, or make this a util class that can be tested separately. I chose the latter to minimize test refactors, but can do the former if that's preferable.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
prateekm: Yeah, it does so in JobNode#generateConfig():
`inEdges.stream().filter(StreamEdge::isIntermediate).forEach(edge -> configs.putAll(edge.generateConfig()));`

In StreamEdge#generateConfig:
```
if (isIntermediate()) {
   config.put(String.format(StreamConfig.IS_INTERMEDIATE_FOR_STREAM_ID(), spec.getId()), "true");
   ...
}
```

samza-core/src/main/java/org/apache/samza/util/StreamUtil.java
prateekm: Yeah, it's currently only used in tests. There's no general pattern / central place for test utils in this code base so I put it here for discoverability. Do you think it's worth extracting to a single method StreamTestUtils? Does anyone else know of a better place for this?

Agree that the String triple is confusing, but I didn't want to introduce a new holder type in the API just for this. StreamSpec could be used as that holder class but that would be undesirable since the intention here is to remove StreamSpec usage from these classes.

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
prateekm: No, I removed it. AFAICT, this is only set by AbstractApplicationRunner when creating a StreamSpec, and it does so by reading it from the config:
AbstractApplicationRunner#getStreamSpec(String, String, String):
```
boolean isBounded = streamConfig.getIsBounded(streamId);
```
StreamEdge is just reading it back from the spec and writing to the config for the stream.

@xinyuiscool can you confirm this is safe?

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
cameronlee314: If we were writing this from scratch, I think my preference would be to pass the StreamConfig through all of the static methods, so my vote would be to do that. Not a big deal if someone feels strongly about doing it the other way though.

samza-core/src/main/java/org/apache/samza/util/StreamUtil.java
cameronlee314: In general, I do think that test utils should be separate from prod utils. It helps prevent someone from accidentally using it in prod code in the future when it is not intended for prod code. I think it is reasonable to start a new StreamTestUtils for that if there isn't already a "test utils" class for it.
If you do move it into a test utils class, then it might be more reasonable to use StreamSpec as a holder, since it would just be using it in tests (although maybe your own new holder would be better if something changes in the future). Another way to do it would be to make this only work for a single (id, systemName, physicalName), and then have the caller call it multiple times if they have multiple streams.

dnishimura: Is it possible to move it to a separate util package in the `samza-core/src/test/` path?
Also, I wouldn't be against having a new holder type. The benefit of a new type is that it will be self documenting. Not a big concern for me though.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
prateekm: Sure, will pass the StreamConfig to the methods and update the tests.

samza-core/src/main/java/org/apache/samza/util/StreamUtil.java
prateekm: Will move to a new test util class. I'd prefer to not introduce StreamSpec back into this package, even for tests. 

How about changing this method to:
`void addStreamConfigs(Map<String, String> configs, String id, String system, String physicalName)`

Would prefer that over adding a tests-only holder class.

cameronlee314: Sounds good to me.

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
nickpan47: So, I think that it is a bit confusing here to me, regarding to where we specify a stream to be bounded as a single place for the source of truth: is it in the user config for a stream? is it in StreamSpec? And what's the relationship between user defined stream configuration and StreamSpec? I thought that the relationship is StreamSpec will be translating to stream configuration while submitting the job. So, it seems that config is the *derived* data set from *StreamSpec*. Hence, all stream-specific properties should be found in the SoT object (i.e. StreamSpec). Could you elaborate what do you consider as the SoT object for a stream in the coming change? StreamDescriptor?StreamSpec?Or StreamConfig? And what are the relationship between them? Thanks!

samza-core/src/main/java/org/apache/samza/operators/OperatorSpecGraph.java
```diff
@@ -54,20 +54,25 @@
...
+  public Map<String, InputOperatorSpec> getInputOperators() {
     return inputOperators;
   }
 
-  public Map<StreamSpec, OutputStreamImpl> getOutputStreams() {
+  public Map<String, OutputStreamImpl> getOutputStreams() {
     return outputStreams;
   }
 
+  public Set<String> getBroadcastStreams() {
```
nickpan47: Is the expectation here that StreamSpec has already been serialized to StreamConfig with the streamId as the key in the config object? Removing the StreamSpec object from OperatorSpecGraph would mean that the information included in StreamSpec (which is currently a serializable object) would need to be in some serialized format stored externally somewhere. Also, if we make this change, ideally, we should remove TableSpec reference from this class as well. I am wondering what's the specific reason for removing the reference to StreamSpec here?

samza-core/src/main/java/org/apache/samza/operators/StreamGraphSpec.java
```diff
@@ -91,15 +86,15 @@ public void setDefaultSerde(Serde<?> serde) {
 
   @Override
   public <M> MessageStream<M> getInputStream(String streamId, Serde<M> serde) {
-    StreamSpec streamSpec = runner.getStreamSpec(streamId);
```
nickpan47: Just one thought: if the reason to remove the reference to StreamSpec is simply because it is accessed via runner.getStreamSpec() method, can we replace it with StreamSpecManager.getStreamSpec() here? It seems that here the only reason that we are calling runner.getStreamSpec() is because we want to construct the StreamSpec from the configuration, which can be put into StreamSpecManager class.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -93,8 +96,9 @@ private void validateConfig() {
    */
   /* package private */ JobGraph createJobGraph(OperatorSpecGraph specGraph) {
     JobGraph jobGraph = new JobGraph(config, specGraph);
-    Set<StreamSpec> sourceStreams = new HashSet<>(specGraph.getInputOperators().keySet());
-    Set<StreamSpec> sinkStreams = new HashSet<>(specGraph.getOutputStreams().keySet());
+    StreamConfig streamConfig = new StreamConfig(config);
+    Set<StreamSpec> sourceStreams = getStreamSpecs(specGraph.getInputOperators().keySet(), streamConfig);
```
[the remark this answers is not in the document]
prateekm: Re: StreamSpecManager and static functions in the planner, I'd prefer to stop at this amount of refactor to time bound it and make it easily reviewable. Once StreamSpec usage is restricted enough, we can consider further refactors like consolidating StreamSpec creation, making them immutable, removing StreamSpec from the SystemAdmin API, making other Planner objects (JobNode, JobGraph) etc. immutable and so on. Does that sound reasonable?

prateekm:
@nickpan47 Are you OK with merging this PR? The test failures on travis seem to be unrelated. My local build passes consistently.

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
nickpan47: it would be more clear if we know in step 2 and 4, which config takes precedence in the merge. And since it is a complex logic here, it would be nice to document it somewhere. Maybe in ExecutionPlanner or JobNode.generateConfig()?

samza-core/src/main/java/org/apache/samza/operators/StreamGraphSpec.java
nickpan47: Make sense!

prateekm:
@nickpan47 Definitely needs to be documented. The documentation will be part of the System/StreamDescriptor SEP and implementation. I'll also add it inline to the ExecutionPlanner with the StreamDescriptors PR.

PR #553, opened 2018-06-14
SAMZA-1747: Add metric to measure effectiveness of host-affinity

vjagadish1989:
We require visibility into how effectively host-affinity performs. The goal is to help easily answer the following questions. 
- How effectively is YARN matching my preferred-host requests
- When does Samza fallback to abandoning locality and issuing any-host requests?

design doc: https://docs.google.com/document/d/1oeNKDnG4JIGT2846us-jpnGW_RUjMjPIKDeEUlE_-jg/edit#


vjagadish1989:
@prateekm for review

PR #556, opened 2018-06-15
SAMZA-1774: Support table API in low level

atoomula:
Code changes to support table in low level API.


samza-core/src/main/java/org/apache/samza/config/TableConfigRewriter.java
```diff
@@ -0,0 +1,60 @@
...
+      if (!TableDescriptorsFactory.class.isAssignableFrom(Class.forName(taskClassName))) {
+        // The configured task does not implement TableDescriptorsFactory.
+        return config;
+      }
+
+      TableDescriptorsFactory tableDescriptorsFactory = Util.getObj(taskClassName, TableDescriptorsFactory.class);
+      List<TableDescriptor> tableDescs = tableDescriptorsFactory.getTableDescriptors();
+      return new MapConfig(Arrays.asList(config, new TableConfigGenerator().generateConfigsForTableDescs(tableDescs)));
+    } catch (Throwable t) {
+      throw new ConfigException(String.format("Invalid configuration for Task class: %s", taskClassName), t);
```
prateekm: Should we be catching exception instead of Throwable here?

Minor: Don't guess exception cause, just log that this failed, with any relevan context (e.g. taskClassName), and log the original exception. For example, this could fail because generateConfigsForTableDescs throws a NPE.

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
```diff
@@ -255,21 +239,11 @@ void addSerdeConfigs(Map<String, String> configs) {
         }
       });
 
-    // collect all key and msg serde instances for tables
```
prateekm: Ideally all serialized objects should be written to the same OutputStream in one pass to maintain their relationships upon deserialization. Is there a way to refactor TableConfigGenerator so that this is still possible?

prateekm:
@atoomula Can you fix the checkstyle failures (see the travis build results).

samza-core/src/main/java/org/apache/samza/config/TableConfigRewriter.java
atoomula: Sure. Changed to catching exception.

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
weisong44: Perhaps the TableConfigGenerator take a list of already generated serdes?

samza-api/src/main/java/org/apache/samza/table/TableDescriptorsFactory.java
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.samza.annotation.InterfaceStability;
+import org.apache.samza.operators.TableDescriptor;
+
+
+/**
+ * Factory to create a list of {@link TableDescriptor} objects to describe one or more Samza tables. This is the
+ * mechanism for providing table support for Samza low level API.
+ *
+ * Developers writing Samza jobs using Samza table(s) should describe the table(s) by implementing
+ * TableDescriptorFactory in their task. Please note that the class that implements this factory should also implement
```
pdu-mn1: nit: typo, missing "s" in TableDescriptorFactory

samza-api/src/main/java/org/apache/samza/table/TableDescriptorsFactory.java
```diff
@@ -0,0 +1,105 @@
...
+ * <pre>
+ * {@code
+ * public class SampleTask implements TableDescriptorsFactory, InitableTask, StreamTask {
+ *   private ReadableTable<String, Long> remoteTable;
+ *   private ReadWriteTable<String, String> localTable;
+ *
+ *   {@code @Override}
+ *   public List<TableDescriptor> getTableDescriptors() {
+ *     List<TableDescriptor> tableDescriptors = new ArrayList<>();
+ *     final TableReadFunction readRemoteTable = (TableReadFunction) key -> null;
```
pdu-mn1: nit: might be clearer to use some concrete table read function for key -> null, eg. new MySQLTableReadFunction()

samza-api/src/main/java/org/apache/samza/table/TableDescriptorsFactory.java
```diff
@@ -0,0 +1,105 @@
...
+ *     tableDescriptors.add(new RocksDbTableDescriptor("local-table-1")
+ *       .withBlockSize(4096)
+ *       .withSerde(KVSerde.of(new LongSerde(), new StringSerde<>())));
+ *       .withConfig("some-key", "some-value");
+ *     return tableDescriptors;
+ *   }
+ *
+ *   {@code @Override}
+ *   public void init(Config config, TaskContext context) {
+ *     remoteTable = (ReadableTable<String, String>) context.getTable(“remote-table-1”);
```
pdu-mn1: nit: given the process() method below, maybe we should call this remote-profile-table, and the local-profile-table to match up. Otherwise, below we should just use generic names, eg. some-field, instead of memberId.

samza-core/src/main/java/org/apache/samza/config/TableConfigRewriter.java
```diff
@@ -0,0 +1,60 @@
...
+      if (!TableDescriptorsFactory.class.isAssignableFrom(Class.forName(taskClassName))) {
+        // The configured task does not implement TableDescriptorsFactory.
+        return config;
+      }
+
+      TableDescriptorsFactory tableDescriptorsFactory = Util.getObj(taskClassName, TableDescriptorsFactory.class);
+      List<TableDescriptor> tableDescs = tableDescriptorsFactory.getTableDescriptors();
+      return new MapConfig(Arrays.asList(config, new TableConfigGenerator().generateConfigsForTableDescs(tableDescs)));
+    } catch (Exception e) {
+      throw new ConfigException(String.format("Invalid configuration for Task class: %s", taskClassName), e);
```
pdu-mn1: nit: Invalid table configuration?

samza-api/src/main/java/org/apache/samza/table/TableDescriptorsProvider.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.samza.operators.TableDescriptor;
+
+
+/**
+ * Provider to create a list of {@link TableDescriptor} objects to describe one or more Samza tables. This is the
+ * mechanism for providing table support for Samza low level API.
+ *
+ * Developers writing Samza jobs using Samza table(s) should describe the table(s) by implementing
+ * TableDescriptorsProvider.
+ *
```
weisong44: Can you add a note here that the hook up is through config rewriter.

atoomula: Sure

samza-api/src/main/java/org/apache/samza/table/TableDescriptorsProvider.java
```diff
@@ -0,0 +1,100 @@
...
+ * <pre>
+ * {@code
+ * public class SampleTableDescriptorsProvider implements TableDescriptorsProvider {
+ *   private ReadableTable<String, Long> remoteTable;
+ *   private ReadWriteTable<String, String> localTable;
+ *
+ *   {@code @Override}
+ *   public List<TableDescriptor> getTableDescriptors() {
+ *     List<TableDescriptor> tableDescriptors = new ArrayList<>();
+ *     final TableReadFunction readRemoteTable = (TableReadFunction) key -> null;
```
pdu-mn1: nit: do you think using a imaginary table reader function is more clear than using the lambda which returns a null? Example: tableReadFn = new MyStoreReadFunction(); 

also: the variables should end with Function or Fn since they're not referencing tables.

samza-core/src/main/java/org/apache/samza/table/TableConfigGenerator.java
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.samza.serializers.SerializableSerde;
+import org.apache.samza.util.Util;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Helper class to generate table configs.
+ */
+public class TableConfigGenerator {
```
pdu-mn1: nit: seems this class can be a simple helper class with all static methods?

PR #557, opened 2018-06-15
Samza-1752: Pass full config to the IO resolver

srinipunuru:
SQL IO Resolver needs full configs so that it can filter out the configs specific to the source that the SQL application is interested in. This change provides the IO resolver with the full config.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedIOResolverFactory.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedIOResolverFactory.java
@@ -47,8 +47,8 @@ public class ConfigBasedIOResolverFactory implements SqlIOResolverFactory {
   public static final String CFG_FMT_SAMZA_PREFIX = "systems.%s.";
 
   @Override
-  public SqlIOResolver create(Config config, Config fullConfig) {
-    return new ConfigBasedIOResolver(config);
+  public SqlIOResolver create(Config resolverConfig, Config fullConfig) {
+    return new ConfigBasedIOResolver(resolverConfig);
   }
 
   private class ConfigBasedIOResolver implements SqlIOResolver {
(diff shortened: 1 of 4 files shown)


samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedIOResolverFactory.java
```diff
@@ -47,7 +47,7 @@
   public static final String CFG_FMT_SAMZA_PREFIX = "systems.%s.";
 
   @Override
-  public SqlIOResolver create(Config config) {
+  public SqlIOResolver create(Config config, Config fullConfig) {
```
vjagadish1989: Can the resolver derive its specific-config from the "fullConfig"?

vjagadish1989:
nit in PR title:  "SAMZA-1234: "

samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedIOResolverFactory.java
srinipunuru: Great question @vjagadish1989 
The pattern that is followed is each of these plugins (IOresolvers, schema resolvers and schema converters)  are given a name. The configs for these plugins will be under the sub domain determined by the name. i.e. Say if this ConfigBasedIOResolver is given a name called "config", configs will be using the prefix `samza.sql.ioResolver.config.`. 
The name of the plugin is defined at the runtime using config, the reason being the same plugin class can behave differently based on different configs passed to it and may need different names for e.g. Avro schema resolver can either use kafka schema registry or espresso schema registry to look up schema and return avro schema and may need different names.   
We could accomplish this using code as well by having these two schema resolvers implement two different factories and schema resolver factories But it just adds unnecessary bulk to the code. And more over it makes sense to have names controllable through configs for better operational flexibility. 

Sorry for the long response. I hope this makes sense :).

vjagadish1989: I like the idea of name-spacing configs to well-defined scopes - We've used that pattern with success in other parts of Samza. I wonder if we can achieve name-spacing by having `create()` interface take in the `resolverName` and a `Config`? 

It keeps the interface clean and we'll have less to explain on what the difference is between "fullConfig" vs "config".  Additionally, it's also consistent with Samza's other factory interfaces - eg: `SystemFactory` 

```
public SqlIOResolver create(Config config, String resolverName); 
```

>> We could accomplish this using code as well by having these two schema resolvers implement two different factories and schema resolver factories

Agreed we shouldn't be doing this.

PR #559, opened 2018-06-21
SAMZA-1753: Added timestamp to Incoming message envelope.

sborya:


samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -36,6 +36,7 @@
   private final Object key;
   private final Object message;
   private final int size;
+  private long timestamp = 0L;
```
prateekm: Can you add context (maybe in the JIRA) for why this is required?
Also, can we set this in the constructor for every IME instead of a setter called by specific consumers?

sborya: Added JIRA.
I didn't want neither change the existing constructor nor add another one to IME. So I thought it is easier just to add getter/setter.

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -66,6 +67,14 @@ public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, Stri
     this.size = size;
   }
 
+  public void setTimestamp(long timestamp) {
```
shanthoosh: Just wondering if it will be clearer to call this as `setProcessingTimestamp()` or `setArrivalTimestamp()`. 

Timestamp sounds abstract enough here and could be interpreted in multiple ways here.

Would be great if we can add annotation `InterfaceStability.Evolving`  to the signature and java doc.

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
prateekm: Apologies for the late reply. I was suggesting set it as System.currentTimeMillis() by default, either in the IME constructors or in the field declaration so that users get a reasonable value when they use it.

+1 for Javadocs. Since this is a public API, we should document what this timestamp represents.

prateekm: @sborya Is the task to add documentation + default timestamp for the IME tracked somewhere? We can try to finish it before the next major release.

PR #560, opened 2018-06-22
Fixed failed unit tests in TestCachingTable

weisong44:
This is due to the recent refactoring of table metrics, for some reason running build locally didn't catch these failed tests.


vjagadish1989 (approved):
approved, thanks Wei.

If you think this change is small enough that it doesn't require a jira, then add a "Minor: ..." to it.

PR #562, opened 2018-06-22
Minor: Changing tests which use resource files to use Class.getResource instead of File to get the path to the resource; without this change, these tests fail when run in the IDE

cameronlee314:


vjagadish1989 (approved):
approved. nice, thanks @cameronlee314 for the contribution!

PR #563, opened 2018-06-26
SAMZA-1758: Configuring a timeout for TestRunner to execute the SamzaJob

Sanil15:


samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -267,20 +273,41 @@ public TestRunner addOutputStream(CollectionStream stream) {
...
   /**
    * Utility to run a test configured using TestRunner
+   *
+   * @param timeout in ms for the Test to Run
+   * @throws InterruptedException if the current thread was interrupted
+   * @throws ExecutionException if the samza job threw an exception
+   * @throws TimeoutException if the wait timed out while running Samza job
    */
-  public void run() {
+  public void run(int timeout) throws InterruptedException, ExecutionException, TimeoutException {
```
shanthoosh: Integer timeout doesn't clearly define the unit of granularity(`seconds` or `milliseconds` or `minutes`) of the parameter and how to interpret it for the user.
Would be better to take `TimeUnit` as a parameter here along with the int timeout or simply `Duration` as a parameter.
Please change if it makes sense.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -267,20 +273,41 @@ public TestRunner addOutputStream(CollectionStream stream) {
...
+          runner.runTask();
+          runner.waitForFinish();
+        } else {
+          runner.run(app);
+          runner.waitForFinish();
+        }
+      }
+    };
+
+    final ExecutorService executor = Executors.newSingleThreadExecutor();
```
shanthoosh: Two points here: 
1. `LocalApplicationRunner` has a bounded `waitForFinish` api which already solves the timeout problem (that is solved by spawning a thread via `ExectorService`).
2. All the threads owned by LocalApplicationRunner api are of deamon type. Ideally any connector(kafka, brooklin etc) should not own/manage non-daemon thread.  So I think it will be unnecessary to forcefully shutdown the `LocalApplicationRunner` threads through `executorService.shutdownNow()`. 

So I think this executor thread block can be replaced with the following. 

```java
        if (app == null) {
          runner.runTask();
          status = runner.waitForFinish(timeout);
        } else {
          runner.run(app);
          status = runner.waitForFinish(timeout);
        }
        ApplicationStatus status = runner.status(app);

	// Add checks here for the application status and throw up to junit if necessary. 
        if (status.getStatusCode() == UnsuccessfulFinish) {
        	throw new SamzaException(status.getThrowable());
        }
```

Currently the `localApplicationRunner.status(app)` api is semantically broken and returns the ApplicationStatus for both the StreamApp, StreamTask.

To get status of StreamTask we have to use `runner.getStatus(app)`

I think we can use that here till we get in Yi's high level API change which will solve this problem(`StreamTaskApplication`, `StreamApplication` & `ApplicationRunnable`).

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -267,20 +273,41 @@ public TestRunner addOutputStream(CollectionStream stream) {
...
   }
 
+
   /**
    * Utility to run a test configured using TestRunner
+   *
+   * @param timeout in ms for the Test to Run
+   * @throws InterruptedException if the current thread was interrupted
+   * @throws ExecutionException if the samza job threw an exception
+   * @throws TimeoutException if the wait timed out while running Samza job
```
shanthoosh: I think rather than throwing all the `Exceptions` back to user, would be better to throw one custom-defined samza exception(`SamzaException`, `TestRunnerException` etc)

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -267,20 +273,41 @@ public TestRunner addOutputStream(CollectionStream stream) {
     return this;
   }
 
+
   /**
    * Utility to run a test configured using TestRunner
+   *
+   * @param timeout in ms for the Test to Run
```
shanthoosh: nit: Would be better to clearly define the behavior when `timeout` elapses.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -267,20 +273,41 @@ public TestRunner addOutputStream(CollectionStream stream) {
     return this;
   }
 
+
```
shanthoosh: Commenting here since github doesn't allow me to comment on the top(Line 86). 

Currently we use `PassThroughCoordinator` for coordination and `SingleContainerGrouper` as `TaskNameGrouper`. This means that `TestRunner` api  supports single containers alone. So it will be better to remove `MULTI_CONTAINER` from the supported modes. Either we can do it as a part of this patch or in a separate one.

Sanil15: will do this in another PR

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: using Samza's waitForFinish(timeout) should remove this, although will keep in mind

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: Thanks for this suggestion, will use this and remove ExecutorService, also add more failure tests

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: Will make this change!

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -267,22 +271,34 @@ public TestRunner addOutputStream(CollectionStream stream) {
...
       runner.run(app);
-      runner.waitForFinish();
+      timedOut = !runner.waitForFinish(timeout);
+    }
+    if (timedOut) {
+      throw new TimeoutException("Timed out waiting for application to finish");
+    }
+    ApplicationStatus status = runner.status(app);
+    if (status.getStatusCode() == ApplicationStatus.StatusCode.UnsuccessfulFinish) {
+      throw new SamzaException(status.getThrowable());
```
shanthoosh: We will not get the complete stacktrace by creating `SamzaException` from throwable object.

Resultant `SamzaException` will contain the head of exception message. We can do either of the following:

1. Just bubble up the throwable: `throw status.getThrowable()`.
2. Use `ExceptionUtils`(from `apache-commons`) to get complete stacktrace from the throwable: `throw new SamzaException(ExceptionUtils.getStackTrace(t))`

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -267,22 +271,34 @@ public TestRunner addOutputStream(CollectionStream stream) {
     return this;
   }
 
+
   /**
    * Utility to run a test configured using TestRunner
+   *
+   * @param timeout time to wait for the high level application or low level task to finish. This timeout does not
+   *                include input stream initialization time or the assertion time over output streams.
+   *                This timeout just accounts for time that samza job takes run
```
shanthoosh: When timeout elapses, we throw up. It's better to mention that in the java doc.

samza-test/src/test/java/org/apache/samza/test/framework/SamzaApplicationFailureTests.java
```diff
@@ -0,0 +1,128 @@
...
+import org.apache.samza.application.StreamApplication;
+import org.apache.samza.operators.KV;
+import org.apache.samza.system.OutgoingMessageEnvelope;
+import org.apache.samza.system.SystemStream;
+import org.apache.samza.test.controlmessages.TestData;
+import org.apache.samza.test.framework.stream.CollectionStream;
+import org.junit.Test;
+
+
+public class SamzaApplicationFailureTests {
```
shanthoosh: There're existing integration tests for test-runner api viz `StreamApplicationIntegrationTest` and `StreamTaskIntegrationTest` .  Since the new tests are testing error scenarios for streamApplication and streamTask, it will be better to move them to the existing test classes.

Sanil15: `StreamApplicationIntegrationTest` and  `StreamTaskIntegrationTest` will be serving as example to users, hence I wanted to keep tests for samza seprately(based on category). We might also move them (StreamApplicationIntegrationTest, StreamTaskIntegrationTest and AsyncStreamTaskIntegrationTest) to src for examples (if needed)

Thanks

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: Used ExceptionUtils from apache-commons to get the stack trace!

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: Added that

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -267,22 +272,34 @@ public TestRunner addOutputStream(CollectionStream stream) {
...
+
   /**
    * Utility to run a test configured using TestRunner
+   *
+   * @param timeout time to wait for the high level application or low level task to finish. This timeout does not include
+   *                input stream initialization time or the assertion time over output streams. This timeout just accounts
+   *                for time that samza job takes run. When job times out a TimeoutException is thrown to the callee
    */
-  public void run() {
+  public void run(Duration timeout) throws SamzaException, TimeoutException {
```
shanthoosh: `SamzaException`  is a sub-type of `RuntimeException`, we don't have to add it in throws clause.

samza-test/src/test/java/org/apache/samza/test/framework/SamzaApplicationFailureTests.java
shanthoosh: Ideally `test-runner` api samples should be part of https://github.com/apache/samza-hello-samza (which already houses samples for high-level, low-level API using `standalone` and `yarn` or samza-samples in main codebase).

IMHO, It should not be placed in samza-test module(contains integration tests for components) and shared with users.

Having duplicate test classes responsible for same thing will create confusion. 

It'll be better to combine these test classes.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -267,22 +272,34 @@ public TestRunner addOutputStream(CollectionStream stream) {
     return this;
   }
 
+
   /**
    * Utility to run a test configured using TestRunner
+   *
+   * @param timeout time to wait for the high level application or low level task to finish. This timeout does not include
+   *                input stream initialization time or the assertion time over output streams. This timeout just accounts
+   *                for time that samza job takes run. When job times out a TimeoutException is thrown to the callee
```
shanthoosh: Should be like the following

```
@param the maximum time to wait for the test execution to complete ... 

@throws TimeoutException when the test execution doesn't complete within the timeout. 
```

Sanil15: done

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: Removed SamzaException here

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -267,22 +272,38 @@ public TestRunner addOutputStream(CollectionStream stream) {
...
     Preconditions.checkState((app == null && taskClass != null) || (app != null && taskClass == null),
         "TestRunner should run for Low Level Task api or High Level Application Api");
+    Preconditions.checkState(!timeout.isZero() && !timeout.isNegative(),
+        "Timeouts should be positive");
     final LocalApplicationRunner runner = new LocalApplicationRunner(new MapConfig(configs));
+    boolean timedOut = false;
     if (app == null) {
       runner.runTask();
-      runner.waitForFinish();
+      timedOut = !runner.waitForFinish(timeout);
```
bharathkk: You can move out the `timedOut = !runner.waitForFinish(timeout)` outside the condition loop.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -267,22 +272,38 @@ public TestRunner addOutputStream(CollectionStream stream) {
...
     } else {
       runner.run(app);
-      runner.waitForFinish();
+      timedOut = !runner.waitForFinish(timeout);
+    }
+    if (timedOut) {
+      throw new TimeoutException("Timed out waiting for application to finish");
```
bharathkk: How about doing something like `Preconditions.checkState(timedOut == false, "Samza application timed out....")`? 
By this, you don't explicitly require the caller to handle `TimeoutException`. I don't see a valid code path beyond the job timeout except failing the test case. So expecting the user to handle timeout exception or bubble it up in their test by adding the exception to the method signature seems unnecessary.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: moved

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -267,22 +272,33 @@ public TestRunner addOutputStream(CollectionStream stream) {
...
    * Utility to run a test configured using TestRunner
+   *
+   * @param timeout time to wait for the high level application or low level task to finish. This timeout does not include
+   *                input stream initialization time or the assertion time over output streams. This timeout just accounts
+   *                for time that samza job takes run. Samza job won't be invoked with negative or zero timeout
+   * @throws SamzaException if Samza job fails with exception and returns UnsuccessfulFinish as the statuscode
    */
-  public void run() {
+  public void run(Duration timeout) {
     Preconditions.checkState((app == null && taskClass != null) || (app != null && taskClass == null),
```
bharathkk: I suggest, we split the validations into simpler ones so that you can precisely convey the error message to the user. The structure right now, forces you to come up with a generic error message for multiple conditions which can be hard to decipher and cryptic to address.

As discussed offline, if you are confident this is an invariant for the `run` method, feel free to remove the condition.

Sanil15: I need to see the use cases, if I plan to remove it I will anyways do it in a different PR (cleanup)

Thanks

PR #564, opened 2018-06-26
SAMZA-1760: SystemAdmin is not started properly in a few places

cameronlee314:


cameronlee314:
It looks like there are more places where the lifecycle is not managed correctly, so this PR is not yet complete. I will leave this open, but it does not need a review for now.

vjagadish1989 (approved):
approved. looks good to me. thank you Cameron for the contribution.

cameronlee314:
I haven't had the opportunity to finish this for a while, so I will close this for now. If this ticket is picked up again in the future, it would be best to just make a new PR. I think some of the code in this PR is still valid, but a lot has changed in a year.

PR #566, opened 2018-06-26
SAMZA-1762: Fix Memory link in the Timer Registry Map

xinyuiscool:
Found a memory leak in the SystemTimerScheduler which does not remove the timers from scheduledFutures after the timers are fired. This caused memory problem for Samza jobs using TimerFn feature. This patch fixes this issue.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/task/SystemTimerScheduler.java
+++ b/samza-core/src/main/java/org/apache/samza/task/SystemTimerScheduler.java
@@ -63,6 +63,7 @@ public class SystemTimerScheduler {
 
     final long delay = timestamp - System.currentTimeMillis();
     final ScheduledFuture<?> scheduledFuture = executor.schedule(() -> {
+        scheduledFutures.remove(key);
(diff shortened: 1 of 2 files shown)


samza-core/src/main/java/org/apache/samza/task/SystemTimerScheduler.java
```diff
@@ -63,6 +63,7 @@ private SystemTimerScheduler(ScheduledExecutorService executor) {
 
     final long delay = timestamp - System.currentTimeMillis();
     final ScheduledFuture<?> scheduledFuture = executor.schedule(() -> {
+        scheduledFutures.remove(key);
```
sborya: but , isn't there a check two lines above that checks that key is not in this map?

xinyuiscool: Right, the removal is in the callback.

sborya: ok, got it.

PR #568, opened 2018-07-11
Samza 1759: Stream Assert utilities for low level and high level api for TestFramework

Sanil15:
Adding utilities and corresponding test for low and high level api


Sanil15:
@shanthoosh @bharathkk for review

samza-test/src/test/java/org/apache/samza/test/framework/AsyncStreamTaskIntegrationTest.java
```diff
@@ -44,10 +49,82 @@ public void testAsyncTaskWithSinglePartition() throws Exception {
...
+    Map<Integer, List<Integer>> output = new HashMap<>();
+    List<Integer> partition = Arrays.asList(1, 2, 3, 4, 5);
+    List<Integer> outputPartition = partition.stream().map(x -> x * 10).collect(Collectors.toList());
+    for (int i = 0; i < 5; i++) {
+      List<KV> keyedPartition = new ArrayList<>();
+      for (Integer val : partition) {
+        keyedPartition.add(KV.of(i, val));
+      }
+      input.put(i, keyedPartition);
+      output.put(i, new ArrayList<Integer>(outputPartition));
```
cameronlee314: Since there is just one input/output message per partition, this doesn't really seem to be testing per-task multithreading. Could you add more inputs/outputs per partition?

samza-test/src/test/java/org/apache/samza/test/framework/StreamApplicationIntegrationTest.java
```diff
@@ -76,7 +82,7 @@ public void testHighLevelApi() throws Exception {
         .addOverrideConfig("job.default.system", "test")
         .run(Duration.ofMillis(1500));
 
-    Assert.assertEquals(TestRunner.consumeStream(output, 10000).get(random.nextInt(count)).size(), 1);
+    Assert.assertEquals(TestRunner.consumeStream(output, Duration.ofMillis(1000)).get(random.nextInt(count)).size(), 1);
```
cameronlee314: Just double checking: Was it intentional to go from 10K ms to 1K ms?

samza-test/src/test/java/org/apache/samza/test/framework/StreamTaskIntegrationTest.java
```diff
@@ -54,11 +60,79 @@ public void testSamzaJobFailureForSyncTask() {
...
+    CollectionStream outputStream = CollectionStream.empty("test", "output", 5);
+
+    TestRunner
+        .of(MyStreamTestTask.class)
+        .addInputStream(inputStream)
+        .addOutputStream(outputStream)
+        .addOverrideConfig("job.container.thread.pool.size", "4")
+        .run(Duration.ofSeconds(2));
+
+    StreamAssert.that(outputStream).contains(output, Duration.ofMillis(1000));
```
cameronlee314: Similar comment about using multiple inputs/outputs per partition. It looks like you want to verify that everything still comes in the same order, but you can't really verify ordering when using just a single element.

samza-test/src/main/java/org/apache/samza/test/framework/StreamAssert.java
```diff
@@ -178,4 +197,72 @@ private void check() {
...
+   * Util to assert  presence of messages in a stream with single partition in any order
+   *
+   * @param expected represents the expected stream of messages
+   * @param timeout maximum time to wait for consuming the stream
+   * @throws InterruptedException when {@code consumeStream} is interrupted by another thread during polling messages
+   *
+   */
+  public void containsInAnyOrder(final List<M> expected, Duration timeout) throws InterruptedException {
+    Preconditions.checkNotNull(collectionStream,
+        "This util is intended to use only on CollectionStream)");
```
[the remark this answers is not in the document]
Sanil15: done

samza-test/src/test/java/org/apache/samza/test/framework/StreamApplicationIntegrationTest.java
Sanil15: 10K was a typo, this should not take more than 1K

samza-test/src/test/java/org/apache/samza/test/framework/AsyncStreamTaskIntegrationTest.java
Sanil15: There are 5 messages per partition and 5 partitions to test, not sure what the confusion is.

samza-test/src/test/java/org/apache/samza/test/framework/StreamTaskIntegrationTest.java
Sanil15: same as above

samza-test/src/test/java/org/apache/samza/test/framework/AsyncStreamTaskIntegrationTest.java
```diff
@@ -44,10 +49,82 @@ public void testAsyncTaskWithSinglePartition() throws Exception {
...
```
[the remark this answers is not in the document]
Sanil15: do you mean more than one stream?

samza-test/src/main/java/org/apache/samza/test/framework/StreamAssert.java
```diff
@@ -178,4 +197,72 @@ private void check() {
...
+  }
+
+  /**
+   * Util to assert ordering of messages in a stream with single partition
+   *
+   * @param expected represents the expected stream of messages
+   * @param timeout maximum time to wait for consuming the stream
+   * @throws InterruptedException when {@code consumeStream} is interrupted by another thread during polling messages
+   */
+  public void contains(final List<M> expected, Duration timeout) throws InterruptedException {
```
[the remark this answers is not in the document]
Sanil15: I followed the Junit standards but since it makes sense to use "containsInOrder" to make it more explicit here

samza-test/src/main/java/org/apache/samza/test/framework/StreamAssert.java
```diff
@@ -64,17 +70,30 @@
 
   private final String id;
   private final MessageStream<M> messageStream;
+  private final CollectionStream<M> collectionStream;
```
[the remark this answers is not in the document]
Sanil15: The issue here is there is no common logic between assertions on MessageStreams an CollectionStream. To split them we have to provide something like CollectionStreamAssert and MessageStreamAssert classes

samza-test/src/test/java/org/apache/samza/test/framework/AsyncStreamTaskIntegrationTest.java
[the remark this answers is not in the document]
cameronlee314: Oops, I think I misread this initially. My comment was incorrect.
That said, could you please slightly rename your variables? Maybe "partition" -> "partitionData" and "outputPartition" -> "expectedOutputPartitionData", to help clarify between the partition ids and the partition data.

samza-test/src/test/java/org/apache/samza/test/framework/AsyncStreamTaskIntegrationTest.java
cameronlee314: You are right. I misread it before. I got mixed up where the data was coming from.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -334,7 +334,7 @@ public void run(Duration timeout) {
     long t = System.currentTimeMillis();
     Map<SystemStreamPartition, List<IncomingMessageEnvelope>> output = new HashMap<>();
     HashSet<SystemStreamPartition> didNotReachEndOfStream = new HashSet<>(ssps);
-    while (System.currentTimeMillis() < t + timeout) {
+    while (System.currentTimeMillis() < t + timeout.toMillis()) {
```
shanthoosh: Minor: please change if it makes sense. 

We can do either of the two things here: 
1. `ExponentialSleepStrategy` util class in samza can be used here to simplify this while loop based timeout logic and have just the consumer fetch operation alone. 
2. Guava [TimeLimiter](https://google.github.io/guava/releases/23.0/api/docs/com/google/common/util/concurrent/TimeLimiter.html)  could be used to replace this retry logic.

samza-test/src/test/java/org/apache/samza/test/framework/AsyncStreamTaskIntegrationTest.java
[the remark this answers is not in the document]
Sanil15: Changed input -> inputPartitionData and output -> expectedOutputPartitionData to make it more obvious

samza-test/src/main/java/org/apache/samza/test/framework/StreamAssert.java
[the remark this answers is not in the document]
cameronlee314: Oh right, instead of "not work", I should have said "not ideal".
A couple of considerations for the user point of view:
1. If you want nice-looking test code: You could have a MessageStreamAssert class and a CollectionStreamAssert class, and then have the StreamAssert class have two "that" methods (one creates MessageStreamAssert, other creates CollectionStreamAssert). Since users just chain the methods, they won't actually see the terms "MessageStreamAssert" or "CollectionStreamAssert", but they will still use the methods from those classes.
2. Regarding confusion of having two classes: In my opinion, it is less confusing to have two classes, because then there is not really a question about which of the methods are usable for each use case, since the API forces users to use the correct ones.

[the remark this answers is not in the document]
Sanil15: @shanthoosh pointed out that since this class StreamAssert does not maintain any state it will be better to just create static assertion utils, although this idea of StreamAssert being the builder for MessageStreamAssert and CollectionStreamAssert is great, we can use this in future once we have need for fluent assertions for ex

`StreamAssert.that(stream).function1().function1()..`

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: I have noted this down, will do in a cleanup patch

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
```diff
@@ -0,0 +1,181 @@
...
+        LATCHES.put(id, new CountDownLatch(count));
+        timer.schedule(timerTask, TIMEOUT);
+      }
+    }
+
+    @Override
+    public void apply(M message, MessageCollector messageCollector, TaskCoordinator taskCoordinator) {
+      actual.add(message);
+
+      if (actual.size() >= expected.size()) {
```
cameronlee314: What if the job first sends the expected output and then it sends some extra output that was not expected? Then it looks like the test would pass, even though the full output did not match the expected output.

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException {
+      in.defaultReadObject();
+      timer = new Timer();
+      actual = Collections.synchronizedList(new ArrayList<>());
+      timerTask = new TimerTask() {
+        @Override
+        public void run() {
+          check();
```
cameronlee314: If "check" fails, will the test fail? I think the timer task runs in a separate thread, so how will the main test thread know that the other thread had an error?

samza-test/src/main/java/org/apache/samza/test/framework/StreamAssert.java
```diff
@@ -20,195 +20,33 @@
...
+   * @param <M> represents the type of Message in the stream
    * @throws InterruptedException when {@code consumeStream} is interrupted by another thread during polling messages
-   *
    */
-  public void containsInAnyOrder(final List<M> expected, Duration timeout) throws InterruptedException {
-    Preconditions.checkNotNull(collectionStream,
-        "This util is intended to use only on CollectionStream)");
+  public static <M> void containsInAnyOrder(CollectionStream<M> collectionStream, final List<M> expected, Duration timeout)
+      throws InterruptedException {
+    Preconditions.checkNotNull(collectionStream, "This util is intended to use only on CollectionStream");
```
cameronlee314: Should this precondition message be updated? If you are going to validate the "collectionStream", should you check against the other arguments as well?

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
```diff
@@ -0,0 +1,181 @@
...
+            LATCHES.remove(id);
+          }
+        }
+      }
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static final class CheckAgainstExpected<M> implements SinkFunction<M> {
```
cameronlee314: Could you please clarify why you are using this strategy for high-level instead of doing something similar to TestRunner.consumeStream, which reads the messages out of the output stream directly?

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
```diff
@@ -0,0 +1,181 @@
...
+
+import static org.junit.Assert.assertThat;
+
+/**
+ * An assertion on the content of a {@link MessageStream}.
+ *
+ * <pre>Example: {@code
+ * MessageStream<String> stream = streamGraph.getInputStream("input", serde).map(some_function)...;
+ * ...
+ * StreamAssert.that(id, stream, stringSerde).containsInAnyOrder(Arrays.asList("a", "b", "c"));
```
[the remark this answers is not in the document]
Sanil15: yes will do

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
Sanil15: This was an existing class which I have renamed from StreamAssert to MessageStreamAssert, the utility of this class to grab messages from an operator and not from stream. StreamAssert using TestRunner.consumeStream() polls messages from an in memory stream, while MessageStreamAssert is used for high level api to grab messages from a node in DAG (ex filter operator)

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
```diff
@@ -0,0 +1,181 @@
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static final class CheckAgainstExpected<M> implements SinkFunction<M> {
+    private static final long TIMEOUT = 5000L;
+
+    private final String id;
+    private final boolean checkEachTask;
+    private final Collection<M> expected;
```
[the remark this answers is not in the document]
Sanil15: Not sure why serialization is important here, since operators will transform deserialized messages

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
```diff
@@ -0,0 +1,181 @@
...
+    CheckAgainstExpected(String id, Collection<M> expected, boolean checkEachTask) {
+      this.id = id;
+      this.expected = expected;
+      this.checkEachTask = checkEachTask;
+    }
+
+    @Override
+    public void init(Config config, TaskContext context) {
+      final SystemStreamPartition ssp = Iterables.getFirst(context.getSystemStreamPartitions(), null);
+      if (ssp == null ? false : ssp.getPartition().getPartitionId() == 0) {
```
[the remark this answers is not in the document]
Sanil15: I have not authored this code, but will do the change

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
Sanil15: The expected is what user passes to assert against actual, not sure about the use case when the expected collection will change after calling this function

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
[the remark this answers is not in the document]
shanthoosh: @Sanil15 
Cameron is refering to serialization of the operator functions in the DAG.
After serializable operator changes, either we need to make each operator explicitly implement `Serializable`  or mark the fields in the operatorFunctionImpl as transient(Synonymous to other member variables in CheckAgainstExpected class).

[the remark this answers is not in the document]
Sanil15: @shanthoosh explained the recent api changes that requires serializable objects getting passed with execution planner, for testing since this is single container mode we will mark it transient

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
shanthoosh: I think it's better to make an immutable copy of the collection in the constructor and use it for making assertions(So the expectedOutput will not change after constructor in the context of this CheckAgainstExpected util).

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
shanthoosh: If `readObject` is not used anywhere, the shall we delete the method(Since it won't cause any problem)?

Sanil15: thought of cleaning that in a new patch when we fix MessageStreamAssert but will do here since it doesn't hurt

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
Sanil15: will fix MessageStreamAssert in a new patch

Sanil15:
@sborya can you merge this, the travis failure is due to some ZK tests failing:

`testRollingUpgradeOfStreamApplicationsShouldGenerateSameJobModel FAILED`
`org.apache.samza.test.processor.TestZkLocalApplicationRunner.shouldReElectLeaderWhenLeaderDies`
`testFailStreamProcessor`

@shanthoosh will have a patch to fix this

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
```diff
@@ -0,0 +1,192 @@
...
+      this.id = id;
+      this.expected = expected;
+      this.checkEachTask = checkEachTask;
+    }
+
+    @Override
+    public void init(Config config, TaskContext context) {
+      final SystemStreamPartition ssp = Iterables.getFirst(context.getSystemStreamPartitions(), null);
+      if (ssp != null || ssp.getPartition().getPartitionId() == 0) {
+        final int count = checkEachTask ? context.getSamzaContainerContext().taskNames.size() : 1;
```
sborya: this condition doesn't make sense!

Sanil15: https://issues.apache.org/jira/browse/SAMZA-1798

Sanil15:
Jira for tracking fix for MessageStreamAssert https://issues.apache.org/jira/browse/SAMZA-1798

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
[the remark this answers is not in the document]
cameronlee314: My example code was indeed incorrect (per sborya's comment). Might need to be "&&" instead of "||".

samza-test/src/main/java/org/apache/samza/test/framework/MessageStreamAssert.java
cameronlee314: My comment was referring to the fact that a test could succeed if the "actual" is a superset of "expected".
For example, a job could send some extra events that should not have been sent. In that case, should the test fail?

PR #569, opened 2018-07-13
Initial implementation of side input stores

bharathkk:
Note: The patch is not for open review yet. It is one of the illustrations of how side input storage manager can be implemented in Samza. I have an alternate implementation that contains refactor of the current implementation. I will have a PR soon that is for public review with the refactored changes.


samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -272,27 +285,38 @@ class TaskInstance(
...
-          val system = envelope.getSystemStreamPartition.getSystem
+          var offset:Option[String] = None
+
+          if (sideInputSSPs.contains(incomingMessageSsp)) {
+            offset = storageManager.getStartingOffset(incomingMessageSsp)
+          } else {
+            offset = offsetManager.getStartingOffset(taskName, incomingMessageSsp)
+          }
+
+          val startingOffset: String = offset
```
cameronlee314: Same as above about storageManager.getStartingOffset returning None.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -77,97 +78,112 @@ class TaskInstance(
...
-      consumerMultiplexer.register(systemStreamPartition, offset)
-      metrics.addOffsetGauge(systemStreamPartition, () => {
-        offsetManager
-          .getLastProcessedOffset(taskName, systemStreamPartition)
-          .orNull
-      })
+      var metricsValueGetter: () => String = null
+      var offset: Option[String] = None
+
+      if (sideInputSSPs.contains(systemStreamPartition)) {
```
cameronlee314: It seems like it might be difficult to ensure that this pattern of:
"if (is side input) {
// use storageManager.getStartingOffset
} else {
// use offsetManager.getStartingOffset
}" is used correctly. It seems like offsetManager won't work on side input SSPs, but there is nothing stopping someone from using a side input SSP with offsetManager except if they know they need to check it is a side input SSP. The system stream partitions are even available through TaskContext, so application code has access to them (maybe even for custom checkpointing?). In addition, this if-else pattern (also used further below) seems a bit hard to maintain.

Is there a way to more strictly enforce that side input SSPs have to use storageManager.getStartingOffset?
Maybe if there was a way to more cleanly categorize an SSP as a side input or not, then you wouldn't need these checks.

bharathkk:
@cameronlee314 This PR is not for open review. Sorry about not making that explicit. We have couple of ways of implementing side input stores and this was one of the illustration of the prototype.

We decided to drop this prototype and go with a separate class for side input storage manager. I will take a look at the comments and apply whatever is applicable to the new prototype.

weisong44:
Question: will support for table come in a separate PR?

prateekm:
@bharathkk Let's close this PR.

PR #570, opened 2018-07-19
SAMZA-1773: Side inputs for local stores

bharathkk:
@prateekm @vjagadish 
Please take a look.

I will update the PR with the unit tests for SideInputStorageManager and the util functions.


samza-core/src/main/java/org/apache/samza/config/TaskConfigJava.java
```diff
@@ -51,6 +51,8 @@
   // class name to use when sending offset checkpoints
   public static final String CHECKPOINT_MANAGER_FACTORY = "task.checkpoint.factory";
 
+  public static final String SIDE_INPUT_PROCESSOR_FACTORY = "task.side.input.processor.factory";
```
prateekm: Unused. Revert changes here.

samza-api/src/main/java/org/apache/samza/storage/StoreProperties.java
```diff
@@ -50,9 +50,19 @@ public boolean isLoggedStore() {
     return loggedStore;
   }
 
+  /**
+   * Flag to indicate whether the store has a side input stream or not
+   *
+   * @return True, if store has side input. False, otherwise
+   */
+  public boolean hasSideInputs() {
```
prateekm: I think this is unused. Can revert changes in this class.

samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java
```diff
@@ -0,0 +1,374 @@
...
+
+
+/**
+ * A storage manager for all side input stores. It is associated with each {@link org.apache.samza.container.TaskInstance}
+ * and is responsible for handling directory management, offset tracking and offset file management for the side input stores.
+ */
+public class SideInputStorageManager {
+  private static final Logger LOG = LoggerFactory.getLogger(SideInputStorageManager.class);
+  private static final String OFFSET_FILE = "SIDE-INPUT-OFFSETS";
+  private static final long STORE_DELETE_RETENTION_MS = TimeUnit.DAYS.toMillis(1); // same as changelog delete retention
```
prateekm: Maybe move to StorageManagerUtil.

samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java
```diff
@@ -0,0 +1,374 @@
...
+  private final Map<SystemStreamPartition, String> sspsToStore;
+  private final StreamMetadataCache streamMetadataCache;
+  private final SystemAdmins systemAdmins;
+  private final TaskName taskName;
+  private final JavaStorageConfig storageConfig;
+  private final Map<SystemStreamPartition, String> lastProcessedOffsets = new HashMap<>();
+
+  private Map<SystemStreamPartition, String> startingOffsets;
+
+  public SideInputStorageManager(
```
prateekm: Minor:  arrange methods in this class in order of use.

samza-core/src/main/java/org/apache/samza/storage/StorageManagerHelper.java
```diff
@@ -0,0 +1,142 @@
...
+      long offsetFileLastModifiedTime = offsetFileRef.lastModified();
+      if ((currentTimeInMs - offsetFileLastModifiedTime) >= storeDeleteRetentionInMs) {
+        LOG.info(
+            String.format("Store: %s is stale since lastModifiedTime of offset file: %d, is older than store deleteRetentionMs: %d.",
+            storePath, offsetFileLastModifiedTime, storeDeleteRetentionInMs));
+        isStaleStore = true;
+      }
+    } else {
+      LOG.info("Storage partition directory: {} does not exist.", storePath);
+      LOG.info("{}", storePath, storePath);
```
prateekm: Remove?

samza-core/src/main/java/org/apache/samza/storage/StorageManagerHelper.java
```diff
@@ -0,0 +1,142 @@
...
+   * Read and return the contents of the offset file.
+   *
+   * @param storagePartitionDir the base directory of the store
+   * @param offsetFileName name of the offset file
+   *
+   * @return the content of the offset file if it exists for the store, null otherwise.
+   */
+  public String readOffsetFile(File storagePartitionDir, String offsetFileName) {
+    String offset = null;
+    File offsetFileRef = new File(storagePartitionDir, offsetFileName);
```
prateekm: s/offsetFileRef/offsetFile everywhere in this class.

samza-core/src/main/java/org/apache/samza/storage/StorageManagerHelper.java
```diff
@@ -0,0 +1,142 @@
...
+import com.google.common.collect.ImmutableMap;
+import java.io.File;
+import org.apache.samza.system.SystemAdmin;
+import org.apache.samza.system.SystemStreamPartition;
+import org.apache.samza.util.FileUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public final class StorageManagerHelper {
```
prateekm: Should delete this class.

samza-core/src/main/scala/org/apache/samza/config/TaskConfig.scala
```diff
@@ -109,6 +109,8 @@ class TaskConfig(config: Config) extends ScalaMapConfig(config) with Logging {
 
   def getCheckpointManagerFactory() = Option(javaTaskConfig.getCheckpointManagerFactoryName)
 
+  def getSideInputProcessorFactory() = Option(javaTaskConfig.getSideInputProcessorFactory)
```
prateekm: Remove

samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
```diff
@@ -42,22 +40,22 @@ object TaskStorageManager {
...
-  changeLogStreamPartitions: Int,
-  streamMetadataCache: StreamMetadataCache,
-  sspMetadataCache: SSPMetadataCache,
-  nonLoggedStoreBaseDir: File = new File(System.getProperty("user.dir"), "state"),
-  loggedStoreBaseDir: File = new File(System.getProperty("user.dir"), "state"),
-  partition: Partition,
-  systemAdmins: SystemAdmins,
-  changeLogDeleteRetentionsInMs: Map[String, Long],
-  clock: Clock) extends Logging {
+                          taskName: TaskName,
```
prateekm: Revert.

samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
```diff
@@ -360,8 +282,8 @@ class TaskStorageManager(
   }
 
   /**
-   * Builds a map from SystemStreamPartition to oldest offset for changelogs.
-   */
+    * Builds a map from SystemStreamPartition to oldest offset for changelogs.
```
prateekm: Revert.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -79,13 +79,22 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
...
+    val storeFactory = storageConfig.get("factory", null)
+    val sideInputs = storageConfig.get("side.inputs", null)
     var storePropertiesBuilder = new StoreProperties.StorePropertiesBuilder()
     val accessLog = storageConfig.getBoolean("accesslog.enabled", false)
 
     if (storeFactory == null) {
       throw new SamzaException("Store factory not defined. Cannot proceed with KV store creation!")
     }
+
+    if (sideInputs != null && !sideInputs.isEmpty) {
```
prateekm: Can revert changes in this class. This is validated elsewhere.

prateekm:
@bharathkk Also, build is failing because of checkstyle issues.

samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java
prateekm: If you're using intellij, you can do  this using Code -> Rearrange code.

samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java
```diff
@@ -0,0 +1,372 @@
...
+
+
+/**
+ * A storage manager for all side input stores. It is associated with each {@link org.apache.samza.container.TaskInstance}
+ * and is responsible for handling directory management, offset tracking and offset file management for the side input stores.
+ */
+public class SideInputStorageManager {
+  private static final Logger LOG = LoggerFactory.getLogger(SideInputStorageManager.class);
+  private static final String OFFSET_FILE = "SIDE-INPUT-OFFSETS";
+  private static final long STORE_DELETE_RETENTION_MS = TimeUnit.DAYS.toMillis(1); // same as changelog delete retention
```
cameronlee314: I thought changelog delete retention was configurable. Should this also be configurable?

samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java
```diff
@@ -0,0 +1,372 @@
...
+          storeLocation.mkdirs();
+        }
+      });
+  }
+
+  private boolean isValidSideInputStore(String storeName) {
+    File storeLocation = getStoreLocation(storeName);
+
+    return isPersistedStore(storeName)
+        && StorageManagerUtil.isStaleStore(storeLocation, OFFSET_FILE, STORE_DELETE_RETENTION_MS, clock.currentTimeMillis())
```
cameronlee314: It seems like this should only be valid if the store is NOT stale.

samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
```diff
@@ -0,0 +1,143 @@
...
+      long offsetFileLastModifiedTime = offsetFileRef.lastModified();
+      if ((currentTimeMs - offsetFileLastModifiedTime) >= storeDeleteRetentionInMs) {
+        LOG.info(
+            String.format("Store: %s is stale since lastModifiedTime of offset file: %d, is older than store deleteRetentionMs: %d.",
+            storePath, offsetFileLastModifiedTime, storeDeleteRetentionInMs));
+        isStaleStore = true;
+      }
+    } else {
+      LOG.info("Storage partition directory: {} does not exist.", storePath);
+      LOG.info("{}", storePath, storePath);
```
cameronlee314: Can this be removed?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -592,17 +614,25 @@ object SamzaContainer extends Logging {
         new StorageConfig(config).getChangeLogDeleteRetentionsInMs,
         new SystemClock)
 
+      var sideInputStorageManager: SideInputStorageManager = null
```
cameronlee314: It looks like the pattern for storageManager is to still be non-null even if there are no stores. Should we use a consistent pattern with this?

prateekm:
+1 to Cameron's comments on StorageManagerUtil and SideInputStorageManager. I think I already fixed some of them in my PR earlier. Let's check if they got merged correctly.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -128,46 +151,52 @@ class TaskInstance(
...
-          val callback = callbackFactory.createCallback()
-          task.asInstanceOf[AsyncStreamTask].processAsync(envelope, collector, coordinator, callback)
-        }
+      if (sideInputSSPs.contains(incomingMessageSsp)) {
+        sideInputStorageManager.process(envelope)
       } else {
-        exceptionHandler.maybeHandle {
-         task.asInstanceOf[StreamTask].process(envelope, collector, coordinator)
-        }
+        if (isAsyncTask) {
```
shanthoosh: Just curious.

* AsyncStreamTask numProcess calls: By existing AsynStreamTask semantics, `task.max.concurrency` denotes the maximum number of `TaskInstance.process` calls that can happen  at a time. Now after this change, we're combining SideInput.process and TaskInstance.process together and that will affect the overall processing throughput of  input streams. Since anyways after bootstrap, we want the store update happen in background in changelog, do we want SideInput.process to affect the number of task process calls? In worst case, sideInput message arrival rate could throttle the message processing rate of input streams of the job.  
* StreamTask commit and process: By existing semantics TaskInstance.commit and TaskInstance.process cannot happen at same time for StreamTask. Now after this change after bootstrap, SideInputProcess.process and TaskInstance.commit cannot occur at same time. Similarily sideInput.process and taskInstance.process cannot happen at same time.   Ideally after sideinput bootstrap, SideInput.process() should not affect the normal job's throughput (occurence of task.process() and task.commit()). 

Please share your thoughts.

Even if we really want to go down this route then atleast it would be better to document it .

samza-api/src/main/java/org/apache/samza/storage/SideInputProcessor.java
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * Process the incoming side input message for the {@code store}.
+   *
+   * @param message incoming message envelope
+   * @param store key value store associated with the incoming message envelope
+   *
+   * @return a {@link Collection} of {@link Entry}s that will be written to the {@code store}.
+   */
+  Collection<Entry<?, ?>> process(IncomingMessageEnvelope message, KeyValueStore store);
```
shanthoosh: We've had scenarios in standalone in past where heavy-weight processing scenarios in TaskInstance.process calls stalling the shutdown phase. 
* What are our semantics if a taskInstance.shutdown is triggered and SideInputManager.process is in progress. 
* Can you please document it in java doc(either here or in SideInputProcessManager).

samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java
```diff
@@ -0,0 +1,372 @@
...
+  public void flush() {
+    LOG.info("Flushing the side input stores.");
+    stores.values().forEach(StorageEngine::flush);
+    writeOffsetFiles();
+  }
+
+  /**
+   * Stops the storage engines for all the stores and writes the offset file to disk.
+   */
+  public void stop() {
```
shanthoosh: Just curious. Shouldn't we call `sideInputManager.stop()` at the end of taskInstance lifecycle? From the implementation, I see it's not called from anywhere.

samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java
```diff
@@ -0,0 +1,372 @@
...
+    SystemStreamPartition ssp = message.getSystemStreamPartition();
+    String storeName = sspsToStore.get(ssp);
+    SideInputProcessor sideInputProcessor = storeToProcessor.get(storeName);
+
+    KeyValueStore keyValueStore = (KeyValueStore) stores.get(storeName);
+    Collection<Entry<?, ?>> entriesToBeWritten = sideInputProcessor.process(message, keyValueStore);
+    keyValueStore.putAll(ImmutableList.copyOf(entriesToBeWritten));
+
+    // update the last processed offset
+    lastProcessedOffsets.put(ssp, message.getOffset());
```
shanthoosh: Do we need synchronization around this update? Since it can be called from different threads for AsyncStreamTask?

samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java
```diff
@@ -0,0 +1,372 @@
...
+
+  private boolean isValidSideInputStore(String storeName) {
+    File storeLocation = getStoreLocation(storeName);
+
+    return isPersistedStore(storeName)
+        && StorageManagerUtil.isStaleStore(storeLocation, OFFSET_FILE, STORE_DELETE_RETENTION_MS, clock.currentTimeMillis())
+        && StorageManagerUtil.isOffsetFileValid(storeLocation, OFFSET_FILE);
+  }
+
+  private boolean isPersistedStore(String storeName) {
```
shanthoosh: Just curious. Do we have any use-cases where side-input stores are not persisted on disk?

samza-api/src/main/java/org/apache/samza/storage/SideInputProcessor.java
```diff
@@ -0,0 +1,43 @@
...
+ * The processing logic for store side inputs. Accepts incoming messages from side input streams
+ * and the current store contents, and returns the new key-value entries to be written to the store.
+ */
+public interface SideInputProcessor {
+
+  /**
+   * Process the incoming side input message for the {@code store}.
+   *
+   * @param message incoming message envelope
+   * @param store key value store associated with the incoming message envelope
```
shanthoosh: 1. If the SideInputProcessor has created resources to external services(db, restli calls), do we want to provide a close() method here to allow garbage collection of them(Or In current world, what is the recommended approach to clean up resources). 
2. If SideInputProcessor relies makes a external RPC/web-service calls, then it would benefit from async api. What is our story for that?

samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java
```diff
@@ -0,0 +1,372 @@
...
+   */
+  private Map<SystemStreamPartition, String> getStartingOffsets() {
+    Map<SystemStreamPartition, String> fileOffsets = getFileOffsets();
+    Map<SystemStreamPartition, String> oldestOffsets = getOldestOffsets();
+
+    Map<SystemStreamPartition, String> startingOffsets = new HashMap<>();
+
+    sspsToStore.keySet().forEach(ssp -> {
+        String fileOffset = fileOffsets.get(ssp);
+        String oldestOffset = oldestOffsets.get(ssp);
```
vjagadish1989: maybe log what the oldest offsets and fileOffsets are if it's not logged elsewhere?

samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java
```diff
@@ -0,0 +1,372 @@
...
+  /**
+   * Gets the oldest offset for the {@link SystemStreamPartition}s associated with all the store side inputs.
+   *   1. Groups the list of the SSPs based on system stream
+   *   2. Fetches the {@link SystemStreamMetadata} from {@link StreamMetadataCache}
+   *   3. Fetches the partition metadata for each system stream and fetch the corresponding partition metadata
+   *      and populates the oldest offset for SSPs belonging to the system stream.
+   *
+   * @return a {@link Map} of {@link SystemStreamPartition} to their oldest offset.
+   */
+  private Map<SystemStreamPartition, String> getOldestOffsets() {
```
vjagadish1989: maybe make it static ?
```Map<SSP, String> getOldestOffsets(Set<SSP> ssps)```

File: samza-api/src/main/java/org/apache/samza/storage/SideInputProcessor.java (hunk 17 above)
Comment (prateekm) [reply samza-api/src/main/java/org/apache/samza/storage/SideInputProcessor.java]: This is no different than earlier, and shutdown timeout is container-wide, so does not matter what's blocking shutdown. Could be task process, side input process, producer flush, store flush etc. Doesn't make sense to document it everywhere.

File: samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java (hunk 13 above)
Comment (vjagadish1989) [reply samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java]: +1

File: samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala (hunk 16 above)
Comment (prateekm) [reply samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala]: These are good points, thanks for bringing them up.

For 1, we'll see how this new feature performs and tune accordingly. If task throughput is affected significantly, we have the option of taking side input consumption out of the run loop and into its own thread(s) with its own consumer(s), similar to changelog restores.

For 2, SideInput#process is currently a part of TaskInstance.process, they're not exclusive. It is only exclusive with Task.process for non-async processing. We can document these semantics in more detail in public documentation when we have a stable design.

Please consider this PR an alpha version for experimentation, not the final release. @bharathkk For this reason, I'd prefer to _not_ document the feature yet, including adding it to configuration reference. Also, all public APIs should have the appropriate InterfaceStability annotations.

File: samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java (hunk 19 above)
Comment (prateekm) [reply samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java]: +1. Don't need synchronization, but should use a CHM to avoid ConcurrentModificationException when invoked concurrently by different TaskInstances.

Comment (prateekm) [reply samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java]: Actually, this can't be called concurrently. There's one SideInputStorageManager per TaskInstance. Since this is called within SideInputManager#process, which is blocking (does not support task.max.concurrency), it will be called sequentially. Still, wouldn't hurt to make it a CHM.

**Review by vjagadish1989 — APPROVED** [review]
approved, few comments on naming. Looks good otherwise!

File: samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala [hunk 24]
```diff
@@ -354,6 +354,14 @@ object SamzaContainer extends Logging {
 
     info("Got intermediate streams: %s" format intermediateStreams)
 
+    val sideInputStoresToSystemStreams = config.getStoreNames
```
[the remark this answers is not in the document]
prateekm: Side inputs are configured and wired as regular task.inputs. They're configured automatically if using TableDescriptors (see changes in JobNode, BaseLocalStoreBackedTableProvider), and have to be configured manually if using them directly.

samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java
```diff
@@ -0,0 +1,372 @@
...
+  }
+
+  /**
+   * Processes the incoming side input message envelope and updates the last processed offset for its SSP.
+   *
+   * @param message incoming message to be processed
+   */
+  public void process(IncomingMessageEnvelope message) {
+    SystemStreamPartition ssp = message.getSystemStreamPartition();
+    String storeName = sspsToStore.get(ssp);
```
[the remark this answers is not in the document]
prateekm: No, since the message will only be delivered to SideInputStorageManager if it belongs to a side input store SSP.

samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java
```diff
@@ -0,0 +1,372 @@
...
+   */
+  @SuppressWarnings("unchecked")
+  private Map<SystemStreamPartition, String> getFileOffsets() {
+    LOG.info("Loading initial offsets from the file for side input stores.");
+    Map<SystemStreamPartition, String> fileOffsets = new HashMap<>();
+
+    stores.keySet().forEach(storeName -> {
+        LOG.debug("Reading local offsets for store {}", storeName);
+
+        File storeLocation = getStoreLocation(storeName);
```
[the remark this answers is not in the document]
prateekm: Creating File objects is fairly cheap (it's a regular object). Since it's also an immutable object and may not reflect changes in the underlying file, it's safer to create a new object when needed. Does make sense to share this logic across the two implementations (TaskStorageManager and TaskSideInputStorageManager). Added this to SAMZA-1780.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1700,6 +1700,28 @@ <h1>Samza Configuration Reference</h1>
                     </td>
                 </tr>
 
+                <tr>
+                    <td class="property" id="stores-side-inputs">stores.<span class="store">store-name</span>.side.inputs</td>
+                    <td class="default"></td>
+                    <td class="description">
+                        A comma-separated list of streams that are side inputs to the store. Each stream is of the format
+                        <span class="system">system-name</span>.<span class="stream">stream-name</span>. Applications are
+                        required to add the side inputs to <a href="#task-inputs" class="property">task.inputs</a> as well.
```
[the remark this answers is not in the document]
prateekm: Feature documentation (including configuration) tracked in SAMZA-1780.

samza-core/src/main/java/org/apache/samza/storage/SideInputStorageManager.java
prateekm: Tracked in SAMZA-1780.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
prateekm: Tracked in SAMZA-1780.

PR #571, opened 2018-07-19
SAMZA-1775: add some delay before renew under transient EH exception

lhaiesp:
There is no delay at all before we renew the partition. This sometimes lead to spam in the log for the following messages:

Received transient exception from EH client. Renew partition receiver for ssp ...

### DIFF
--- a/samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
@@ -479,8 +479,7 @@ public class EventHubSystemConsumer extends BlockingEnvelopeMap {
               String.format("Received transient exception from EH client. Renew partition receiver for ssp: %s", ssp),
               throwable);
           try {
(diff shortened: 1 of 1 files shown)


vjagadish1989 (approved):
approved, thanks for the contribution!

samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java
```diff
@@ -478,6 +478,13 @@ public void onError(Throwable throwable) {
           LOG.warn(
               String.format("Received transient exception from EH client. Renew partition receiver for ssp: %s", ssp),
               throwable);
+          try {
+            // Even though these are transient exceptions, we don't want to retry too frequently which sometimes
```
vjagadish1989: Minor: maybe rewrite this comment as
"Add a fixed delay so that we don't keep retrying when there are long-lasting failures"

vjagadish1989 (approved):
approved, one minor suggestion on rewording the doc.

PR #572, opened 2018-07-20
SAMZA-1733: Adding containerID to metric header

rmatharu:
Adding containerID to MetricsHeader (published by MetricsSnapshotReporter). 
It is populated using the value set for the env variable ShellCommandConfig.ENV_EXECUTION_ENV_CONTAINER_ID

### DIFF
--- a/samza-core/src/test/scala/org/apache/samza/serializers/TestMetricsSnapshotSerde.scala
+++ b/samza-core/src/test/scala/org/apache/samza/serializers/TestMetricsSnapshotSerde.scala
@@ -22,6 +22,7 @@ package org.apache.samza.serializers
 import java.util.HashMap
 import java.util.Map
 
+import org.apache.samza.config.ShellCommandConfig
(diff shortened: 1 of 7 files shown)


samza-core/src/test/scala/org/apache/samza/serializers/TestMetricsSnapshotSerde.scala
```diff
@@ -33,7 +34,7 @@ class TestMetricsSnapshotSerde {
   @Ignore
   @Test
   def testMetricsSerdeShouldSerializeAndDeserializeAMetric {
-    val header = new MetricsHeader("test", "testjobid", "task", "test", "version", "samzaversion", "host", 1L, 2L)
+    val header = new MetricsHeader("test", "testjobid", "task", "test", "test", "version", "samzaversion", "host", 1L, 2L)
```
nickpan47: Minor comment: if we want to verify the result in the MetricsHeader in the unit test, it would be better to have a different string here for the execEnvironmentContainerId.

PR #573, opened 2018-07-20
NO-JIRA: Change instances of "is catched up" to "is caught up"

tbutterwith:
Replaces any use of "catched" with "caught"


vjagadish1989 (approved):
approved.
Thanks for improving the quality of the code-base!

PR #574, opened 2018-07-23
SAMZA-1779: Fix SamzaSql AvroRelConverter to handle BYTES and FIXED avro schema types

atoomula:


samza-sql/src/test/java/org/apache/samza/sql/system/TestAvroSystemFactory.java
```diff
@@ -291,6 +292,7 @@ private Object createComplexRecord(int index) {
       GenericRecord record = new GenericData.Record(ComplexRecord.SCHEMA$);
       record.put("id", index);
       record.put("string_value", "Name" + index);
+      record.put("bytes_value", java.nio.ByteBuffer.wrap(("sample bytes").getBytes()));
```
vjagadish1989: s/java.nio.ByteBuffer.wrap/ByteBuffer.wrap 

Or, Are there naming conflicts with ByteBuffer class that require us to fully namespace it?

vjagadish1989 (approved):
approved. just a question, otherwise looks good

samza-sql/src/test/java/org/apache/samza/sql/system/TestAvroSystemFactory.java
atoomula: Actually, nothing. I removed the namespace prefix.

atoomula:
Another PR with https://github.com/apache/samza/pull/575/ has fixes for both byte and fixed types.

PR #580, opened 2018-07-25
Side inputs: Bug fixes

bharathkk:
- Use the correct regex in Util#getSystemStreamFromNameOrId
 - Handle end of stream during dispatch of message to side input
processor
 - Fix validation to check for presence of side input processor for a
given store instead of looking up the side input processor factory.


bharathkk:
@prateekm please take a look

PR #582, opened 2018-07-26
Handle end of stream in BootstrapChooser

bharathkk:
Handle end of stream envelopes in bootstrap chooser and don't invoke check offsets since offset of end of stream are not comparable.


shanthoosh:
@bharathkk 

Do we need to handle `EOS` envelopes in register, update methods as well? 

In register the comparator is invoked to see if  existingOffset for a SSP has to be replaced.

1. Here's a sample scenario. Let's say two tasks consume from the same SSP in a container. Let's assume one task is at EOS and other task is at beginning of the stream for that SSP. After a `SamzaContainer` bounce, they both could try to register with different offsets for the same SSP in `BootstrappingChooser`. So we could potentially run into not comparable scenario in `BootstrappingChooser.register` method as well.

2. It would be good to add unit tests to verify different EOS handling scenarios in BootstrappingChooser.

PR #583, opened 2018-07-26
SAMZA-1786: Introduce metadata store abstraction.

shanthoosh:
As a part of SEP-11, this patch adds MetadataStore interface to store task and container locality in both yarn and standalone deployment models. Please refer to SEP-11 for more details. 

Few important points to note:
1. As a part of this changes, LocalityManager/TaskAsssignmentManager alone will be updated to use this interface(subsequently in upcoming future RB's other util classes will be moved to use this interface as well).
2. In an immediate followup RB, ZkMetadataStore(storing metadata information in zookeeper) will be added. It will be used in standalone to read/write locality into zookeeper(through LocalityManager & other standard util classes).
3. In future, ExecutionPlan, streamGraph and other job related metadata can be stored in any custom store through the same abstraction.

Testing:
1. Added unit tests for new classes introduced in the patch(Fixed the existing unit tests in LocalityManager/TaskAssignmentManager).
2. All the changes in the patch were validated with test jobs in samza-hello-samza(https://github.com/apache/samza-hello-samza).
3. LinkedIn testing job(maes-tests-host-affinity) was verified with these changes to validate if things work end-to-end.


shanthoosh:
@prateekm @xinyuiscool @sborya 

samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
```diff
@@ -0,0 +1,78 @@
...
+public interface MetadataStore {
+
+  /**
+   * Initializes the metadata store, if applicable, setting up the underlying resources
+   * and connections to the store endpoints. Upon successful completion of this method,
+   * metadata store is considered available to accept the client operations.
+   *
+   * @param containerContext represents the Samza container context.
+   */
+  void init(SamzaContainerContext containerContext);
```
prateekm: This can be used outside of SamzaContainer, and as such may not have access to the container context. Should provide access to Config and MetricsRegistry instead.

prateekm:
cc @rayman7718, in case you have any feedback on the MetadataStore interface from the perspective of a diagnostics data store.

vjagadish1989:
@shanthoosh : Would be preferable to structure the metadata store API to have string keys and byte-array values. Strings offer you nice properties around equality and prefix-checks.

samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
rayman7718: +1 a CLI tool using this interface to read metadata wouldn't have this context.

<String, byte[]> or <String, Object> might be a simpler/better alternative for a starter impl.

It'd be nice if the metadatastore offered atomic updates across keys, e.g., for diagnostics a container-id and error-occurrence in that container should be recorded atomically. 

Does the implementation of this plan to internally use tableAPI implementation(s)?

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamKeySerde.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.samza.serializers.Serde;
+
+/**
+ * Serializer for keys written into coordinator stream(kafka topic). CoordinatorStreamMessage combines
+ * both key and value serde for coordinator stream messages. Since key is relevant to this serializer,
+ * coordinator stream value is nuked for different message types.
+ */
+public class CoordinatorStreamKeySerde implements Serde<String> {
+
+  private final Serde<List<?>> keySerde;
```
sborya: should use JsonSerde

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamValueSerde.java
```diff
@@ -0,0 +1,85 @@
...
+    this.config = config;
+    this.type = type;
+    messageSerde = new JsonSerde<>();
+  }
+
+  @Override
+  public String fromBytes(byte[] bytes) {
+    Map<String, Object> values = messageSerde.fromBytes(bytes);
+    CoordinatorStreamMessage message = new CoordinatorStreamMessage(new Object[]{}, values);
+    if (type.equalsIgnoreCase(SetContainerHostMapping.TYPE)) {
```
sborya: This is called for EACH message. I think we should optimize this code performance wise (may be use integer for type).

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -19,38 +19,75 @@
...
+   * @param metricsRegistry the registry for reporting metrics in metadata store.
+   * @param keySerde denotes the key serializer.
+   * @param valueSerde denotes the value serializer.
    */
-  public LocalityManager(CoordinatorStreamManager coordinatorStreamManager) {
-    this.coordinatorStreamManager = coordinatorStreamManager;
-    this.taskAssignmentManager = new TaskAssignmentManager(coordinatorStreamManager);
+  public LocalityManager(Config config, MetricsRegistry metricsRegistry, Serde<String> keySerde, Serde<String> valueSerde) {
+    this.config = config;
+    MetadataStoreFactory metadataStoreFactory = Util.getObj(new JobConfig(config).getMetadataStoreFactory(), MetadataStoreFactory.class);
```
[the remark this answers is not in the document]
shanthoosh: Discussed offline.

`Serde` is not part of the store, it's used for serialize/deserialize messages before writing/reading from the store  in `LocalityManager`/`TaskAssignmentManager`(util classes) and not passed to the store.

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
```diff
@@ -84,16 +117,16 @@ public TaskAssignmentManager(CoordinatorStreamManager coordinatorStreamManager)
...
-      log.info("Task \"{}\" moved from container {} to container {}", new Object[]{taskName, existingContainerId, containerId});
+      LOG.info("Task \"{}\" moved from container {} to container {}", new Object[]{taskName, existingContainerId, containerId});
     } else {
-      log.debug("Task \"{}\" assigned to container {}", taskName, containerId);
+      LOG.debug("Task \"{}\" assigned to container {}", taskName, containerId);
     }
 
     if (containerId == null) {
-      coordinatorStreamManager.send(new Delete(SOURCE, taskName, SetTaskContainerMapping.TYPE));
+      metadataStore.remove(keySerde.toBytes(taskName));
```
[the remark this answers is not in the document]
shanthoosh: We propagate it back to the caller(which was the existing behavior as well prior to this change).  

I think it's better to retain the behavior and not alter it as a part of this change.

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
```diff
@@ -105,7 +138,11 @@ public void writeTaskContainerMapping(String taskName, String containerId) {
    */
   public void deleteTaskContainerMappings(Iterable<String> taskNames) {
     for (String taskName : taskNames) {
-      writeTaskContainerMapping(taskName, null);
+      metadataStore.remove(keySerde.toBytes(taskName));
```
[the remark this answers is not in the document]
shanthoosh: Good catch. Fixed.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamKeySerde.java
shanthoosh: Yes. That's right. I instantiate the field to `JsonSerde` in the constructor.

Moved the assignment to declaration itself.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamValueSerde.java
shanthoosh: Agree. You're right with respect to the performance. 

But the number of messages in coordinator stream is supposed to small(due to log-compaction guarantees from kafka). 

Existing implementation uses the same comparator for each message. We have not seen any performance problems with that setup yet. Moreover, I don't want to alter the existing comparator behavior as a part of this change. If acceptable, will create a followup JIRA and see if we really want to do this.

samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
shanthoosh: Makes sense. Removed Context and switched to having `config` and `metricsRegistry` as parameters. 

I'm not sure if it's a good idea to fix string as key type. This will prevent us in the future from persisting non-string keys in the metadatastore(since majority of store API's offer byte[], byte[] contract, this will incur a unnecessary limitation) and reading/writing it across samza-job, custom tools etc. 

>> It'd be nice if the metadatastore offered atomic updates across keys.

Not sure if it's a good idea to add this to interface contract and mandate this for every store implementation. For majority of scenarios in samza, a store implementation that persists the (key,value) through a synchronous remote call should be sufficient. 

Atomic updates through write-batch(or similar batch accumulation) is not supported by every store implementation. Unless we have prevalent and concrete use-cases across different layers in samza, i think it's a good idea not to introduce it.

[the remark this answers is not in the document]
shanthoosh: Replaced with `config` and `metricsRegistry` as arguments.

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -60,29 +97,23 @@ public LocalityManager(CoordinatorStreamManager coordinatorStreamManager) {
...
-
+    metadataStore.all().forEach((keyBytes, valueBytes) -> {
+        if (valueBytes != null) {
+          String locationId = valueSerde.fromBytes(valueBytes);
+          allMappings.put(keySerde.fromBytes(keyBytes), ImmutableMap.of(SetContainerHostMapping.HOST_KEY, locationId,
+                                                                        SetContainerHostMapping.JMX_TUNNELING_URL_KEY, "",
+                                                                        SetContainerHostMapping.JMX_URL_KEY, ""));
+        }
+      });
+    containerToHostMapping = allMappings;
```
bharathkk: why is this no longer an `unmodifiableMap`? I don't see any places where you write a value to this map.

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -93,35 +124,30 @@ public LocalityManager(CoordinatorStreamManager coordinatorStreamManager) {
...
-        new SetContainerHostMapping(CONTAINER_PREFIX + containerId, String.valueOf(containerId), hostName, jmxAddress,
-            jmxTunnelingAddress));
+
+    metadataStore.put(keySerde.toBytes(containerId), valueSerde.toBytes(hostName));
+
     Map<String, String> mappings = new HashMap<>();
     mappings.put(SetContainerHostMapping.HOST_KEY, hostName);
     mappings.put(SetContainerHostMapping.JMX_URL_KEY, jmxAddress);
     mappings.put(SetContainerHostMapping.JMX_TUNNELING_URL_KEY, jmxTunnelingAddress);
     containerToHostMapping.put(containerId, mappings);
```
bharathkk: I don't see the purpose of this as well. It looks like `writeContainerToHostMapping` seems to be invoked only during the start of the container and this update is not even read subsequently. I am surprised as to how this is working prior to your change since the old implementation sets `containerToHostMapping` to `unmodifiableMap` inside `readContainerLocality`

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStore.java
```diff
@@ -0,0 +1,197 @@
...
+    bootstrapMessagesFromStream();
+    return bootstrappedMessages.get(key);
+  }
+
+  @Override
+  public void put(byte[] key, byte[] value) {
+    try {
+      OutgoingMessageEnvelope envelope = new OutgoingMessageEnvelope(coordinatorSystemStream, 0, key, value);
+      if (value != null) {
+        bootstrappedMessages.put(key, value);
```
bharathkk: why do we need this if all the read path `get` and `all` methods bootstrap for messages prior to reading?

samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
```diff
@@ -0,0 +1,78 @@
...
+   */
+  byte[] get(byte[] key);
+
+  /**
+   * Updates the mapping of the specified key-value pair; Associates the specified {@code key} with the specified {@code value}.
+   *
+   * @param key the key with which the specified {@code value} is to be associated.
+   * @param value the value with which the specified {@code key} is to be associated.
+   */
+  void put(byte[] key, byte[] value);
```
rmatharu: Thread-safety is the responsibility of the implementers or of the callers?

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -19,38 +19,75 @@
...
+import org.apache.samza.util.Util;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Locality Manager is used to persist and read the container-to-host
- * assignment information from the coordinator stream
+ * assignment information from the coordinator stream.
  * */
 public class LocalityManager {
```
rmatharu: This seems to be a wrapper over MetadataStore, specialized to store host-container mapping, so 
could we name it something specific like ContainerHostMappingStore? 
Or do you foresee extensions of this

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -19,38 +19,75 @@
...
-  private static final String CONTAINER_PREFIX = "SamzaContainer-";
   private static final Logger LOG = LoggerFactory.getLogger(LocalityManager.class);
 
-  private final CoordinatorStreamManager coordinatorStreamManager;
+  private final Config config;
+  private Map<String, Map<String, String>> containerToHostMapping = new HashMap<>();
```
rmatharu: // Map from containerID -> Map of per-container properties, e.g., {@link SetContainerHostMapping.HOST_KEY}

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
```diff
@@ -21,32 +21,72 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-
 /**
  * Task assignment Manager is used to persist and read the task-to-container
- * assignment information from the coordinator stream
+ * assignment information from the coordinator stream.
  * */
 public class TaskAssignmentManager {
```
rmatharu: Can we make "LocalityManager" and "TaskAssignmentManager" have a common base class that encapsulates 
config, serdes, Metadatastore?
This way any additions to Metadatastore can naturally propagate to all "*Managers", and will provide a way to enforce somethings across *Manager classes

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
bharathkk: I should have been more explicit. I do understand the need for it to be in coordinator stream. I was referring to the part on why we persist the information in a local map within `LocalityManager`.
`  Map<String, String> mappings = new HashMap<>();
    mappings.put(SetContainerHostMapping.HOST_KEY, hostName);
    mappings.put(SetContainerHostMapping.JMX_URL_KEY, jmxAddress);
    mappings.put(SetContainerHostMapping.JMX_TUNNELING_URL_KEY, jmxTunnelingAddress);
    containerToHostMapping.put(containerId, mappings);
`
After writing the locality information to the coordinator stream as part of the startup, the locality information is not accessed for the rest of the lifecycle of the running container. Subsequent reads happen only for container movements, restarts or deploy where it is guaranteed to see the update from coordinator stream.

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
bharathkk: It is, in general, a good practice to enforce immutability where ever possible especially, in concurrent programming. `containerToHostMapping` is only set in `readContainerLocality()` and not modified anywhere else. 

The other reason I feel strongly about this change is, your PR doesn't introduce any changes to the map as well. so why modify its properties? It makes it hard to backtrack and rule out changes during debugging.

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStore.java
bharathkk: Agree about the delta part. Are there possibilities of `get` or `all` being invoked concurrently during a `put`?  If no, then whatever new message that gets written (appended) to the coordinator stream should be read by subsequent `get` or `all` as part of `bootstrapMessagesFromStream()`, right?

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStore.java
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  @Override
+  public Map<byte[], byte[]> all() {
+    bootstrapMessagesFromStream();
+    return Collections.unmodifiableMap(bootstrappedMessages);
+  }
+
+  /**
+   * Read all messages from the earliest offset, all the way to the latest.
```
prateekm: Is this correct? Doesn't this only consume from the offset we last left at?

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStore.java
```diff
@@ -0,0 +1,197 @@
...
+   * Currently, this method only pays attention to config messages.
```
prateekm: Is this "only config messages" still correct?

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamValueSerde.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.samza.coordinator.stream.messages.CoordinatorStreamMessage;
+import org.apache.samza.coordinator.stream.messages.SetChangelogMapping;
+import org.apache.samza.coordinator.stream.messages.SetContainerHostMapping;
+import org.apache.samza.coordinator.stream.messages.SetTaskContainerMapping;
+import org.apache.samza.SamzaException;
+import org.apache.samza.serializers.JsonSerde;
+import org.apache.samza.serializers.Serde;
+
+/**
+ * Serializer for values written into the coordinator stream(kafka topic). CoordinatorStreamMessage combines both key
```
prateekm: s/kafka topic//

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamValueSerde.java
```diff
@@ -0,0 +1,85 @@
...
+ * and value serializer for coordinator stream messages. Since value is relevant to this serializer, coordinator stream
+ * key is nuked for different message types.
```
prateekm: "nuked for"? Also can you clarify what is this trying to say? Not sure I understand.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -186,4 +189,6 @@ class JobConfig(config: Config) extends ScalaMapConfig(config) with Logging {
   def getNonLoggedStorePath = getOption(JobConfig.JOB_NON_LOGGED_STORE_BASE_DIR)
 
   def getLoggedStorePath = getOption(JobConfig.JOB_LOGGED_STORE_BASE_DIR)
+
+  def getMetadataStoreFactory = getOption(JobConfig.METADATA_STORE_FACTORY).getOrElse(classOf[CoordinatorStreamStoreFactory].getCanonicalName)
```
prateekm: Should not be defaulted here. Let's do it at the call site where CoordinatorStreamStoreFactory is used.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamValueSerde.java
```diff
@@ -0,0 +1,85 @@
...
+      return String.valueOf(changelogMapping.getPartition());
+    } else {
+      throw new SamzaException(String.format("Unknown coordinator stream message type: %s", type));
+    }
+  }
+
+  @Override
+  public byte[] toBytes(String value) {
+    if (type.equalsIgnoreCase(SetContainerHostMapping.TYPE)) {
+      SetContainerHostMapping hostMapping = new SetContainerHostMapping(SOURCE, "", value, config.get("jmx.tunneling.url"), config.get("jmx.url"));
```
prateekm: Is the jmx url stuff still relevant? Maybe remove or update class javadoc to clarify that this will not be set anymore?

samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Deletes the mapping for the specified {@code key} from this metadata store (if such mapping exists).
+   *
+   * @param key the key for which the mapping is to be deleted.
+   */
+  void remove(byte[] key);
```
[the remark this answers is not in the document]
prateekm: Bump.

samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
```diff
@@ -0,0 +1,78 @@
...
+   *
+   * @return all entries in this metadata store.
+   */
+  Map<byte[], byte[]> all();
+
+  /**
+   * Closes the metadata store, if applicable, relinquishing all the underlying resources
+   * and connections.
+   */
+  void close();
```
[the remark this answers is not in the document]
prateekm: Minor: Move flush above close.

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStore.java
```diff
@@ -0,0 +1,198 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(CoordinatorStreamStore.class);
+  private static final String SOURCE = "SamzaContainer";
+
+  private final Config config;
+  private final SystemStream coordinatorSystemStream;
+  private final SystemStreamPartition coordinatorSystemStreamPartition;
+  private final SystemProducer systemProducer;
+  private final SystemConsumer systemConsumer;
+  private final SystemAdmin systemAdmin;
+  private final Object bootstrapLock = new Object();
```
prateekm: Move next to bootstrappedMessages to clarify what this is guarding.

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStore.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  @Override
+  public void put(byte[] key, byte[] value) {
+    try {
+      OutgoingMessageEnvelope envelope = new OutgoingMessageEnvelope(coordinatorSystemStream, 0, key, value);
+      systemProducer.send(SOURCE, envelope);
+      flush();
+    } catch (Exception e) {
+      throw new SamzaException(e);
```
prateekm: Don't need to wrap exceptions in SamzaException unless we have special handling for them at the call site. Here and everywhere else.

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStore.java
```diff
@@ -0,0 +1,198 @@
...
+      systemAdmin.stop();
+      systemProducer.stop();
+      systemConsumer.stop();
+    } catch (Exception e) {
+      LOG.error("Exception occurred when closing the metadata store:", e);
+    }
+  }
+
+  @Override
+  public void flush() {
```
prateekm: Minor: Move flush above close.

samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamValueSerde.java
prateekm: It's still setting them. Intentional?
If you remove this, you can remove the config from the serde too.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
prateekm: 1. CoordinatorStreamStore is not necessarily the long term default implementation, and it's not even the only current implementation. Wouldn't you need to call this with ZkStore for all standalone call sites anyway?

2. Use getName() for class names that are meant to be classloaded.

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStoreFactory.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.samza.config.Config;
+import org.apache.samza.metadatastore.MetadataStore;
+import org.apache.samza.metadatastore.MetadataStoreFactory;
+import org.apache.samza.metrics.MetricsRegistry;
+
+/**
+ * Builds the {@link CoordinatorStreamStore} based upon the provided {@link Config}
+ * and {@link MetricsRegistry}.
+ */
+public class CoordinatorStreamStoreFactory implements MetadataStoreFactory {
```
prateekm: Would prefer CoordinatorStreamMetadataStoreFactory etc.

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -19,109 +19,119 @@
...
+   * Key and value serializer are different for yarn (uses CoordinatorStreamMessage) and standalone (native ObjectOutputStream for serialization) modes.
+   * @param config the configuration required for setting up metadata store.
+   * @param metricsRegistry the registry for reporting metrics.
+   * @param keySerde the key serializer.
+   * @param valueSerde the value serializer.
    */
-  public LocalityManager(CoordinatorStreamManager coordinatorStreamManager) {
-    this.coordinatorStreamManager = coordinatorStreamManager;
-    this.taskAssignmentManager = new TaskAssignmentManager(coordinatorStreamManager);
+  public LocalityManager(Config config, MetricsRegistry metricsRegistry, Serde<String> keySerde, Serde<String> valueSerde) {
```
prateekm: Is this constructor used anywhere (other than tests)? If not, maybe make protected/package private.

samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.samza.annotation.InterfaceStability;
+import org.apache.samza.config.Config;
+import org.apache.samza.metrics.MetricsRegistry;
+import java.util.Map;
+
+/**
+ * Store abstraction responsible for managing the metadata of a Samza job.
+ */
+@InterfaceStability.Evolving
+public interface MetadataStore {
```
dnishimura: Should there also be a `create()` API? Without it, this assumes the underlying store is already created. In the case of the case of the `CoordinatorStreamStore` implementation, the underlying stream gets created directly from the SystemAdmin here:

https://github.com/apache/samza/blob/06488bf7a7dfdfa6d26ed91f79b6c316460960b1/samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala#L93

Should we replace that with a `MetadataStore#create()` call to fully abstract away the coordinator stream?

samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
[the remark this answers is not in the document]
shanthoosh: Done.

samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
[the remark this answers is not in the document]
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStore.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStore.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStore.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStoreFactory.java
shanthoosh: Done.

samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
shanthoosh: 1. I was thinking of introducing a new abstraction `MetadataStoreAdmin` with methods like `delete(store)`, `create(store)`, `validate`, `clear(store)` that will house all the admin related operations for a kind of metadata store. 
2. We will have two implementations viz `CoordinatorStreamMetadatastoreadmin` , `ZKMetadataStoreAdmin`  for yarn and standalone respectively. The existing `MetadataStore` contact assumes the presence of store and provides operations on it. IMHO, the admin operations should be kept separate abstraction. We should have `MetadataStoreFactory.getMetadataStoreAdmin(config, registry)` method and that should return admin object for a type of metadata store.

@prateekm @dnishimura  What do you guys think?

shanthoosh: 1. I was thinking of introducing a new abstraction `MetadataStoreAdmin` with methods like `delete(store)`, `create(store)`, `validate`, `clear(store)` that will house all the admin related operations for a kind of metadata store.
2. We will have two implementations viz `CoordinatorStreamMetadatastoreadmin`, `ZKMetadataStoreAdmin` for yarn and standalone respectively. The existing `MetadataStore` contact assumes the presence of store and provides operations on it. IMHO, the admin operations should be kept separate and should be part of separate abstraction. We should have `getMetadataStoreAdmin(config, registry)` method in `MetadataStoreFactory` and that should return admin object for a type of metadata store.

@prateekm @dnishimura What do you guys think?

dnishimura: It aligns with the `SystemFactory` pattern. Works for me.

PR #584, opened 2018-07-26
SAMZA-1787: Refactor TaskStorageManager to use SSPMetadataCache during init

cameronlee314:
Testing done:
Deployed test job with changelog with no existing file offset, verified through logging that oldest offset was registered with the changelog consumer in init stage for TaskStorageManager
Deployed test job with changelog with existing file offset, verified through logging that file offset was registered with the changelog consumer in init stage for TaskStorageManager

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
+++ b/samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
@@ -23,10 +23,10 @@ import java.io._
 import java.util
 
 import org.apache.samza.config.StorageConfig
-import org.apache.samza.{Partition, SamzaException}
 import org.apache.samza.container.TaskName
 import org.apache.samza.system._
 import org.apache.samza.util.{Clock, FileUtil, Logging}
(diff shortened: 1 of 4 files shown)


samza-core/src/main/scala/org/apache/samza/storage/TaskStorageManager.scala
```diff
@@ -152,10 +151,7 @@ class TaskStorageManager(
       systemAdmin.validateStream(changelogSpec)
```
rmatharu: This check appears to be redundant, since there is already a similar check at the AM. 
If we get rid of this check 
and modify getStartingOffset() method in this class to use sspMetadataCache.getMetadata(ssp).getOldestOffset

then we can remove this class's dependency on systemAdmins aswell (since it already has sspMetadataCache).

systemAdmins is not thread-safe, so this would help in parallelizing task-restores. 
Or I can send a separate PR with that change.

cameronlee314:
@rmatharu is taking this over through a separate PR

PR #585, opened 2018-07-27
SAMZA-1788: Add LocationIdProvider abstraction.

shanthoosh:
Currently in standalone, by default hostName of the standalone processor is used as LocationId. However, for containerized environments like azure cloud, kubernetes this defaulting does not work. Standalone processors can be launched from different kubernetes container on a physical machine(where each kubernetes container has different locatliyID than other kubernetes container within same machine). 

To solve this problem, we introduce locationID abstraction to allow users to plugin a uniqueId identifying the execution environment of the processor.

In containerized environments, LocationId is a composite key of multiple fields: (sliceId, containerId, hostname) By default hostname will be used as LocationId(if not configured by the user).

All the processors of an application registered from an locationID should be able to share(read/write) their local state stores. Any custom LocationIdProvider is expected to honor this contract when generating the locationID.

This patch is part of SEP-11. Please refer to it for more details.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/runtime/LocationIdProvider.java
+++ b/samza-api/src/main/java/org/apache/samza/runtime/LocationIdProvider.java
@@ -1,0 +1,30 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.runtime;
+
+import org.apache.samza.config.Config;
+
+/**
+ * Generates {@link LocationId} that uniquely identifies the
+ * execution environment of a stream processor.
+ */
+public interface LocationIdProvider {
+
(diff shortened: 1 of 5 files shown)


shanthoosh:
@xinyuiscool @sborya 

samza-core/src/main/java/org/apache/samza/runtime/DefaultLocationIdProviderFactory.java
```diff
@@ -0,0 +1,28 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.runtime;
+
+import org.apache.samza.util.Util;
+
+public class DefaultLocationIdProviderFactory implements LocationIdProviderFactory {
```
vjagadish1989: Consider adding java-docs here?

samza-api/src/main/java/org/apache/samza/runtime/LocationIdProvider.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.apache.samza.config.Config;
+
+/**
+ * Generates {@link LocationId} that uniquely identifies the
+ * execution environment of a stream processor.
+ */
+public interface LocationIdProvider {
+
+  LocationId getLocationId(Config config);
```
vjagadish1989: Prefer re-thinking these interfaces as follows:
```
public interface LocationIdProvider {
   LocationId getLocationId();
}
```
 
and the corresponding factory interface as:

```
public interface LocationIdProviderFactory {
  LocationIdProvider getLocationIdProvider(Config config);
}
```

shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/runtime/DefaultLocationIdProviderFactory.java
shanthoosh: Done.

vjagadish1989:
Looks like there are test failures. can you take a look? Otherwise, LGTM

PR #586, opened 2018-07-27
SAMZA-1790: LocalContainerRunner should not extend AbstractApplicationRunner.

prateekm:
LocalContainerRunner is the launcher for the process running SamzaContainer in YARN. It extends the AbstractApplicationRunner since the container was using ApplicationRunner#getStreamSpec to create StreamSpecs from config to create the High Level API DAG. It doesn't implement any of the other APIs from the ApplicationRunner.

With SAMZA-1659 and SAMZA-1745, SamzaContainer no longer needs access to StreamSpec to create and execute the High Level API DAG. We can now clean up the LocalContainerRunner implementation so that it doesn't need to implement the ApplicationRunner interface.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
@@ -31,9 +31,9 @@ import org.apache.samza.container.ContainerHeartbeatClient;
 import org.apache.samza.container.ContainerHeartbeatMonitor;
 import org.apache.samza.container.SamzaContainer;
 import org.apache.samza.container.SamzaContainer$;
+import org.apache.samza.operators.StreamGraphSpec;
 import org.apache.samza.util.SamzaUncaughtExceptionHandler;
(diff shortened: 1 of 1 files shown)


prateekm:
cc. @nickpan47 @vjagadish1989 @xinyuiscool for review.

vjagadish1989 (approved):
approved , Looks great!

nickpan47:
Just to confirm, this is not the intended end state of the container runner main, simply a step to rewind the hack we did before, right? SAMZA-1168 talks about launching StreamProcessor in container. In the end state, we want a proper implementation of application runner to launch the StreamProcessor in the container process, with start(); and waitForFinish(). Could you open a JIRA to track the end state of the container refactor effort?

Otherwise, lgtm. Thanks!

prateekm:
@nickpan47 Right, this is just to undo the "extends ApplicationRunner" and revert to a clean state, as a simple SamzaContainer runner process. We should still move towards using StreamProcessor in YARN as well. We can use "SAMZA-1168: Migrate yarn local runner to use StreamProcessor" to track that.

PR #588, opened 2018-07-27
SAMZA-1768: Handle corrupted OFFSET file

xinyuiscool:
This patch addresses the following tickets:

SAMZA-1778: SIGSEGV when reading properties (metrics) on a closed RocksDB store
SAMZA-1777: Logged store OFFSET file write during flush should be atomic
SAMZA-1768: Handle corrupted OFFSET file elegantly

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
+++ b/samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
@@ -128,7 +128,11 @@ public class StorageManagerUtil {
 
     if (offsetFileRef.exists()) {
       LOG.info("Found offset file in storage partition directory: {}", storePath);
-      offset = FileUtil.readWithChecksum(offsetFileRef);
+      try {
+        offset = FileUtil.readWithChecksum(offsetFileRef);
+      } catch (Exception e) {
+        LOG.warn("Fail to read offset file of " + storePath);
+      }
     } else {
       LOG.info("No offset file found in storage partition directory: {}", storePath);
     }
--- a/samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
+++ b/samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
@@ -131,7 +131,7 @@ public class StorageManagerUtil {
       try {
         offset = FileUtil.readWithChecksum(offsetFileRef);
       } catch (Exception e) {
-        LOG.warn("Fail to read offset file of " + storePath, e);
+        LOG.warn("Failed to read offset file in storage partition directory: {}", storePath, e);
       }
     } else {
       LOG.info("No offset file found in storage partition directory: {}", storePath);
--- a/samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
+++ b/samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
@@ -131,7 +131,7 @@ public class StorageManagerUtil {
       try {
         offset = FileUtil.readWithChecksum(offsetFileRef);
       } catch (Exception e) {
-        LOG.warn("Fail to read offset file of " + storePath);
+        LOG.warn("Fail to read offset file of " + storePath, e);
       }
     } else {
       LOG.info("No offset file found in storage partition directory: {}", storePath);
--- a/samza-core/src/main/scala/org/apache/samza/util/FileUtil.scala
+++ b/samza-core/src/main/scala/org/apache/samza/util/FileUtil.scala
@@ -36,7 +36,6 @@ object FileUtil {
     * */
   def writeWithChecksum(file: File, data: String): Unit = {
     val checksum = getChecksum(data)
-
     val tmpFilePath = file.getAbsolutePath + ".tmp"
     val tmpFile = new File(tmpFilePath)
     var oos: ObjectOutputStream = null
--- a/samza-core/src/main/scala/org/apache/samza/util/FileUtil.scala
+++ b/samza-core/src/main/scala/org/apache/samza/util/FileUtil.scala
@@ -45,17 +45,17 @@ object FileUtil {
       oos = new ObjectOutputStream(fos)
       oos.writeLong(checksum)
       oos.writeUTF(data)
-
-      //atomic swap of tmp and real offset file
-      try {
-        Files.move(tmpFile.toPath, file.toPath, StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING)
-      } catch {
-        case e: AtomicMoveNotSupportedException =>
-          Files.move(tmpFile.toPath, file.toPath, StandardCopyOption.REPLACE_EXISTING)
-      }
     } finally {
       oos.close()
       fos.close()
+    }
+
+    //atomic swap of tmp and real offset file
+    try {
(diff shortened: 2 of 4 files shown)


xinyuiscool:
@prateekm : could you please take a look?

samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
```diff
@@ -91,7 +91,13 @@ object RocksDbKeyValueStore extends Logging {
         .toSet
 
       (configuredMetrics ++ rocksDbMetrics)
-        .foreach(property => metrics.newGauge(property, () => rocksDb.getProperty(property)))
+        .foreach(property => metrics.newGauge(property, () =>
+          if (rocksDb.isOwningHandle) {
```
prateekm: Minor: Maybe add a comment explaining that isOwningHandle is false if the db is closed.

samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
```diff
@@ -128,7 +128,11 @@ public static String readOffsetFile(File storagePartitionDir, String offsetFileN
 
     if (offsetFileRef.exists()) {
       LOG.info("Found offset file in storage partition directory: {}", storePath);
-      offset = FileUtil.readWithChecksum(offsetFileRef);
+      try {
+        offset = FileUtil.readWithChecksum(offsetFileRef);
+      } catch (Exception e) {
+        LOG.warn("Fail to read offset file of " + storePath, e);
```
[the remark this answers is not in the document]
xinyuiscool: Fixed.

samza-core/src/main/scala/org/apache/samza/util/FileUtil.scala
```diff
@@ -35,17 +36,22 @@ object FileUtil {
...
       oos = new ObjectOutputStream(fos)
       oos.writeLong(checksum)
       oos.writeUTF(data)
     } finally {
       oos.close()
       fos.close()
     }
+
+    //atomic swap of tmp and real offset file
+    Files.move(tmpFile.toPath, file.toPath, StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING)
```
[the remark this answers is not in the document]
xinyuiscool: The caller already handles the exception by logging the error and throw a runtime exception.

samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
xinyuiscool: Makes sense. Added the comment.

samza-core/src/main/scala/org/apache/samza/util/FileUtil.scala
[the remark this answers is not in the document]
prateekm: Right, but would be good to move to try so that `oos` and `fos` are closed in finally.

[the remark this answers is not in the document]
xinyuiscool: ok, sure, let me do that.

samza-core/src/main/scala/org/apache/samza/util/FileUtil.scala
```diff
@@ -35,13 +36,18 @@ object FileUtil {
...
     var fos: FileOutputStream = null
     try {
-      fos = new FileOutputStream(file)
+      fos = new FileOutputStream(tmpFile)
       oos = new ObjectOutputStream(fos)
       oos.writeLong(checksum)
       oos.writeUTF(data)
+
+      //atomic swap of tmp and real offset file
+      Files.move(tmpFile.toPath, file.toPath, StandardCopyOption.ATOMIC_MOVE)
```
shanthoosh: 1. I think we need to add `StandardCopyOption.REPLACE_EXISTING` to the existing options to `Files.move` method. `Files.move` throws up `FileAlreadyExistsException` if target file already exists and we don't provide that Replace_existing option as an additional option. 

2. Atomic move is a file system level primtiive operation and not all file system support it by default. File.move throws up `AtomicMoveNotSupportedException` if underlying file system doesn't support it.  It would be better to handle it as well. 

Please change if it makes sense(Above details are taken from Files.move javadoc).

samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
```diff
@@ -128,7 +128,11 @@ public static String readOffsetFile(File storagePartitionDir, String offsetFileN
 
     if (offsetFileRef.exists()) {
       LOG.info("Found offset file in storage partition directory: {}", storePath);
-      offset = FileUtil.readWithChecksum(offsetFileRef);
+      try {
+        offset = FileUtil.readWithChecksum(offsetFileRef);
```
shanthoosh: I think it would be better to add a unit test for these changes.

samza-core/src/main/scala/org/apache/samza/util/FileUtil.scala
prateekm: I don't think we need both REPLACE_EXISTING and ATOMIC_MOVE, but we might need to catch IOException as well where we catch AtomicMoveNotSupportedException. From the javadoc for ATOMIC_MOVE:

```If the target file exists then it is implementation specific if the existing file is replaced or this method fails by throwing an IOException.```

File: samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java (hunk 5 above)
Comment (xinyuiscool) [reply samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java]: Added the test for the case of offset already exist.

**Comment by xinyuiscool** [comment]
@prateekm : I move the Files.move() out of the try{} block again since we should close the streams before we do the file movement. This is to enforce the write to the offset file before move. The IOExceptions are handled from the callers so we don't need to catch them in this function. The current handling of failing the container looks ok to me since it should happen very rarely, as we haven't seen it happening in production.

**Comment by prateekm** [comment]
@xinyuiscool Makes sense to move this to its own try-catch block. It might still be better to handle IOException explicitly and revert to a normal move + replace existing. The IOException can happen if the implementation doesn't support replacing existing files as a part of the move. My concern is that this might fail on a particular platform consistently and we'll always end up rebootstrapping. Alternatively, if you've tested that it works on OSX and Linux, that'll be good enough for now.

Otherwise LGTM, feel free to merge this.

**Comment by xinyuiscool** [comment]
Good point on testing on both OSX and linux. I only tested on linux. Let me try OSX. Thanks.

File: samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala [hunk 6]
```diff
@@ -91,7 +91,14 @@ object RocksDbKeyValueStore extends Logging {
         .toSet
 
       (configuredMetrics ++ rocksDbMetrics)
-        .foreach(property => metrics.newGauge(property, () => rocksDb.getProperty(property)))
+        .foreach(property => metrics.newGauge(property, () =>
```
sborya: what is this change do?

PR #589, opened 2018-07-27
Fix to test, for added field containerid in metrics-header 

rayman7718:


rayman7718:
Closing PR, subsequent PR with actual/better fix.


Root cause of the failing test (after  constructor param addition) is that java's StackTraceElement does not serialize/deserialize properly in all cases. 

Counter example: 

StackTraceElement s1 = new StackTraceElement("a", "b", null, 10);
byte[] b = new ObjectMapper().writeValueAsString(s1).getBytes("UTF-8");
StackTraceElement s2 = new ObjectMapper().readValue(b, StackTraceElement.class);
System.out.println(s1.equals(s2));// prints false

In reality, the null fileName for StackTraceElement is possible for java's internal method calls 

such as  org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:50), sun.reflect.GeneratedMethodAccessor4.invoke(Unknown Source)

PR #590, opened 2018-07-28
 SAMZA-1791: Fixing problem with Serde of StackTraceElement, was causing test failure

rmatharu:
Root cause of the failing test (after constructor param addition) is that java's StackTraceElement does not serialize/deserialize properly in all cases.

Counter example:

StackTraceElement s1 = new StackTraceElement("a", "b", null, 10);
byte[] b = new ObjectMapper().writeValueAsString(s1).getBytes("UTF-8");
StackTraceElement s2 = new ObjectMapper().readValue(b, StackTraceElement.class);
System.out.println(s1.equals(s2));// prints false

In reality, the null fileName for StackTraceElement **occurrs** for java's internal method calls
such as org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:50), sun.reflect.GeneratedMethodAccessor4.invoke(Unknown Source)

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsExceptionEvent.java
+++ b/samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsExceptionEvent.java
@@ -18,10 +18,10 @@
  */
 package org.apache.samza.diagnostics;
 
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


shanthoosh:
@rmatharu 
Can you update the PR with this JIRA: `SAMZA-1791`(created to fix the problem). It will be easier for the committers for fix with it.

PR #592, opened 2018-07-31
SAMZA-1794: setting application acl in launch context 

lhaiesp:
Currently we don't set application acl for container launch context. See https://hadoop.apache.org/docs/r2.6.4/api/org/apache/hadoop/yarn/api/records/ContainerLaunchContext.html#setApplicationACLs(java.util.Map)

This could potentially cause problem if samza job is running on a secured YARN cluster. Say user A submits the job, then by default only user A can view the log and the status of the job. Even worse case is that user A submits the job through some proxy account, then even user A herself/himself couldn't access to logs/status of the application.

We need to make some changes for the YARN application submission to set application acls in launch context as configured.

### DIFF
--- a/samza-yarn/src/main/java/org/apache/samza/config/YarnConfig.java
+++ b/samza-yarn/src/main/java/org/apache/samza/config/YarnConfig.java
@@ -115,6 +115,22 @@ public class YarnConfig extends MapConfig {
    */
   public static final String YARN_JOB_STAGING_DIRECTORY = "yarn.job.staging.directory";
 
+  /**
+   * For secured YARN cluster only.
+   * The 'viewing' acl of the YARN application. This controls who can view the application,
+   * for example, application status, logs.
+   * {@link org.apache.hadoop.yarn.api.records.ApplicationAccessType} for more details
+   */
+  public static final String YARN_APPLICATION_VIEW_ACL = "yarn.job.view.acl";
+
+  /**
+   * For secured YARN cluster only.
+   * The 'modify' acl of the YARN application. This controls who can modify the application,
+   * for example, killing the job.
+   * {@link org.apache.hadoop.yarn.api.records.ApplicationAccessType} for more details
+   */
+  public static final String YARN_APPLICATION_MODIFY_ACL = "yarn.job.modify.acl";
+
   public YarnConfig(Config config) {
     super(config);
   }
@@ -190,4 +206,13 @@ public class YarnConfig extends MapConfig {
   public String getYarnJobStagingDirectory() {
     return get(YARN_JOB_STAGING_DIRECTORY, null);
   }
+
+  public String getYarnApplicationViewAcl() {
+    return get(YARN_APPLICATION_VIEW_ACL, null);
+  }
+
+  public String getYarnApplicationModifyAcl() {
+    return get(YARN_APPLICATION_MODIFY_ACL, null);
+  }
+
 }
--- a/samza-yarn/src/main/scala/org/apache/samza/job/yarn/ClientHelper.scala
+++ b/samza-yarn/src/main/scala/org/apache/samza/job/yarn/ClientHelper.scala
@@ -52,11 +52,13 @@ import org.apache.samza.job.ApplicationStatus.New
 import org.apache.samza.job.ApplicationStatus.Running
 import org.apache.samza.job.ApplicationStatus.SuccessfulFinish
 import org.apache.samza.job.ApplicationStatus.UnsuccessfulFinish
-import org.apache.samza.util.Logging
+import org.apache.samza.util.{Logging, Util}
 import java.io.IOException
 import java.nio.ByteBuffer
 
 import org.apache.http.impl.client.HttpClientBuilder
+import org.apache.samza.container.SamzaContainer.info
+import org.apache.samza.container.SecurityManagerFactory
 import org.apache.samza.webapp.ApplicationMasterRestClient
 
 object ClientHelper {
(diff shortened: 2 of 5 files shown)


samza-yarn/src/main/scala/org/apache/samza/job/yarn/ClientHelper.scala
```diff
@@ -191,12 +193,15 @@ class ClientHelper(conf: Configuration) extends Logging {
...
 
-    if (UserGroupInformation.isSecurityEnabled()) {
+    if (UserGroupInformation.isSecurityEnabled) {
       validateJobConfig(config)
 
       setupSecurityToken(fs, containerCtx)
       info("set security token for %s" format appId.get)
 
+      val securityManager = new SamzaContainerSecurityManager(config, new YarnConfiguration())
+      securityManager.setApplicationAcl(containerCtx)
```
vjagadish1989: Would prefer to inline the `setApplicationAcl` here (or inside `setupSecurityTokens`) instead of exposing it as a method on the `SecurityManager`. A nice property is that all mutations to the "context" object happen in a single place. 

Alternately, If you want to move this to `SecurityManager`, we could move other aspects as well, including token setup and renewal. Currently the security-manager does only renewal.

lhaiesp: ClientHelper.scala and YarnClusterResourceManager.java share the setApplicationAcl logic here. If you prefer this small piece of duplicate codes instead of a common place, I guess that's fine. What do you think?

samza-yarn/src/main/scala/org/apache/samza/job/yarn/SamzaContainerSecurityManager.scala
```diff
@@ -44,6 +48,39 @@ class SamzaContainerSecurityManager(config: Config, hadoopConfig: Configuration)
...
+    if (modifyAcl != null) {
+      acls += ApplicationAccessType.MODIFY_APP -> modifyAcl
+    }
+    if (acls.nonEmpty) {
+      info("setting application acls in launch context: %s" format acls)
+      ctx.setApplicationACLs(acls.asJava)
+    }
+  }
+
+  def setupSecurityToken(fs: FileSystem, amContainer: ContainerLaunchContext): Unit = {
```
vjagadish1989: Now that I see how it fits together, I actually think that exposing a getApplicationAcl() API on the YARNConfig class is the right call here.
`Map<ApplicationAccessType, String> getApplicationAcl()`

Both, ClientHelper and YarnClusterResourceManager should be able to invoke it since they have access to the `YarnConfig`. We wouldn't need to expose these methods on the `SamzaContainerSecurityManager` and can move the `setApplicationAcls` close to where it's actually required.

vjagadish1989 (approved):
approved, one suggestion!

samza-yarn/src/main/scala/org/apache/samza/job/yarn/SamzaContainerSecurityManager.scala
lhaiesp: So the reason to place these logic in SamzaContainerSecurityManager is so that we have a single place to handle security related logic. Had a initial discussion with @nickpan47 before I implemented this and he suggest SamzaContainerSecurityManager, that's how I ended up with this approach. I think your point is also valid. To me I don't feel either approach is significantly better. I would rather stick with what it is now.

So having YARNConfig to directly return Map<ApplicationAccessType, String> is an interesting idea. However, I think the Map<ApplicationAccessType, String> data structure is tightly coupled with the signature of ContainerLaunchContext. setApplicationACLs(). In the hadoop codes, view acl and modify acl are used independently. I would think it's simpler that the YARNConfig also treat them differently (getViewAcl and getModifyAcl as two independent methods and configs)

vjagadish1989: Currently, the `SamzaSecurityManager` is scoped to the AMService and we shouldn't be instantiating multiple instances of it elsewhere. eg: With this, we're instantiating it in ClientHelper and YarnClusterResourceManager. 

Secondly, we should consider how the newly added methods fit in with the other methods on the SecurityManager. For eg: is it an expectation that we call `start` prior to invoking `setupSecurityToken`.  Until we have clear lifecycle, I'd prefer to avoid placing this logic inside the `SecurityManager` class - esp., if it entails something simple as setting some fields on the context.

@nickpan47 : What do you think?

lhaiesp: Yeah. the life cycle concern is a good point. I'm not familiar with scala and couldn't figure out how to do static method for a class. I have reverted my change on the security manager. And now do everything inside ClientHelper and YarnClusterResourceManager

samza-yarn/src/main/java/org/apache/samza/config/YarnConfig.java
```diff
@@ -190,4 +210,30 @@ public String getYarnCredentialsFile() {
   public String getYarnJobStagingDirectory() {
     return get(YARN_JOB_STAGING_DIRECTORY, null);
   }
+
+  public String getYarnApplicationViewAcl() {
```
vjagadish1989: Minor: can inline these one-line methods / make them private since they're only used inside `getApplicationAcls`

samza-yarn/src/main/scala/org/apache/samza/job/yarn/SamzaContainerSecurityManager.scala
```diff
@@ -35,7 +35,7 @@ import java.util.concurrent.TimeUnit
 
 class SamzaContainerSecurityManager(config: Config, hadoopConfig: Configuration) extends SecurityManager with Logging {
   private val InitialDelayInSeconds = 60
-  
+
```
vjagadish1989: trivial: revert changes to this file?

samza-yarn/src/main/java/org/apache/samza/config/YarnConfig.java
lhaiesp: This file seems to have a getter for every config even though a couple of them are not used anywhere, so I'm simply following the pattern. More importantly, as I said, the view and modify acl are independently used and should be treated as two different configs. The Map<ApplicationAccessType, String> is tightly coupled with the setApplicationAcls method. I'm simply providing a helper function here

PR #594, opened 2018-08-01
SideInput: Added tests and fixed serde issue during writing offsets

bharathkk:
- Fixed serde issues during persisted offsets locally
- Added unit tests for TaskSideInputStorageManager
- Added integration tests for table integration


samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
```diff
@@ -184,6 +188,14 @@ public String getLastProcessedOffset(SystemStreamPartition ssp) {
     return lastProcessedOffsets.get(ssp);
   }
 
+  /**
+   * For unit testing
+   */
+  @VisibleForTesting
+  void updatelastProcessedOffset(SystemStreamPartition ssp, String offset) {
```
prateekm: `updateLastProcessedOffset`
"For unit testing only."
Move to bottom of class, or after getLastProcessedOffset if it exists.

samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
```diff
@@ -242,7 +255,9 @@ private void writeOffsetFiles() {
               .collect(Collectors.toMap(Function.identity(), lastProcessedOffsets::get));
 
             try {
-              String fileContents = OBJECT_MAPPER.writeValueAsString(offsets);
+              String fileContents = OBJECT_MAPPER.writerWithType(
```
prateekm: 1. Is the writer required if we use SamzaObjectMapper? What happens if we use SamzaObjectMapper directly?
2. Can we save the ObjectWriter as a static field?

samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
```diff
@@ -268,7 +284,9 @@ private void writeOffsetFiles() {
         if (isValidSideInputStore(storeName, storeLocation)) {
           try {
             String fileContents = StorageManagerUtil.readOffsetFile(storeLocation, OFFSET_FILE);
-            Map<SystemStreamPartition, String> offsets = OBJECT_MAPPER.readValue(fileContents, Map.class);
+            TypeReference<HashMap<SystemStreamPartition, String>> ref =
```
prateekm: Save as final field?

samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
bharathkk: It is already under `getLastProcessedOffset`

samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
bharathkk: Done

samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
bharathkk: 1. It uses the default map key and value serializers since its unable to get the type of the key and value. During the deserialization part, it uses the type specific deserializer which expects the string format to be different from the `toString` representation of `SystemStreamPartition` and fails to load it correctly.
2. Done

samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
prateekm: Minor: Need to fix the method name too (L in Last should be capital).

samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
```diff
@@ -68,6 +68,9 @@
   private static final String OFFSET_FILE = "SIDE-INPUT-OFFSETS";
   private static final long STORE_DELETE_RETENTION_MS = TimeUnit.DAYS.toMillis(1); // same as changelog delete retention
   private static final ObjectMapper OBJECT_MAPPER = SamzaObjectMapper.getObjectMapper();
+  private static final TypeReference<HashMap<SystemStreamPartition, String>> OFFSET_TYPE_REFERENCE =
```
prateekm: Nitpick: OFFSETS_TYPE_REFERENCE, since this is for the entire map and not a single offset.

PR #598, opened 2018-08-04
SAMZA-1754: Improve validation error messages in StreamSpec

rgroothuijsen:
Currently, StreamSpec errors report only that an identifier has an invalid value, but they provide no context for the error. This PR includes the stream ID, physical name, system name and partition count in the error message, as requested in the ticket.

Code changes:
- Validation of the stream ID and system name is done after setting the fields, so that they can be read by the existing toString() method if an error occurs.
- The output of toString() was prettified to look more pleasant and readable for end users.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
+++ b/samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
@@ -150,9 +150,6 @@ public class StreamSpec implements Serializable {
    * @param config          A map of properties for the stream. These may be System-specfic.
    */
   public StreamSpec(String id, String physicalName, String systemName, int partitionCount, Map<String, String> config) {
-    validateLogicalIdentifier("streamId", id);
-    validateLogicalIdentifier("systemName", systemName);
-
     // partition count being 0 is a valid use case in Hadoop when the output stream is an empty folder
     if (partitionCount < 0) {
       throw new IllegalArgumentException("Parameter 'partitionCount' must be >= 0");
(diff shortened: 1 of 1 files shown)


samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
```diff
@@ -163,6 +160,9 @@ public StreamSpec(String id, String physicalName, String systemName, int partiti
     this.physicalName = physicalName;
     this.partitionCount = partitionCount;
 
+    validateLogicalIdentifier("streamId", id);
```
vjagadish1989: In general, it's a good idea to set class fields after validating them. Additionally, it'd be nice to retain doing all validations together.

What do you think about this?

- Remove validateLogicalIdentifier() all-together. The existing implementation does not have adequate context where it throws the exception. Hence, callers will have to catch and re-throw the exception wrapping additional context. While re-throwing exceptions is a common enough pattern, it feels like an over-kill for trivial validation checks. 

- Replace all validations with `Preconditions.check`
```
Preconditions.check(partitionCount > 0, "Partition count should be greater than zero.  for id = %s, physicalName = %s, systemName = %s, partitionCount = %s"); 
Preconditions.check(streamId != null, "Stream id should be non-null for id = %s, physicalName = %s, systemName = %s, partitionCount = %s"); 
Preconditions.check(..., "Stream id should start match [a-zA-z0-9]. Current values id = %s, physicalName = %s, systemName = %s, partitionCount = %s"); 
```

If you think it feels boiler-platey, you can wrap all validations in a helper static method : validateStreamSpec(String id, String physicalName, String systemName, String partitionCount) and move them there

rgroothuijsen: @vjagadish1989 Thanks for the input, I sort of had that idea when weighing implementations and wasn't sure which one to go for. Attempting to re-use toString() felt the most elegant at first, but I do see your point, so I'll go for that one.

PR #600, opened 2018-08-06
SAMZA-1802: Enable host affinity when RocksDB is present

weisong44:
We should enable host affinity when RocksDB table is present, this should be done in RocksDB table provider.


weisong44:
@bharathkk good point. I've updated the PR, can you please take a look again.

PR #602, opened 2018-08-06
SAMZA-1733: Create diagnostic topic if diagnostics enabled

rmatharu:
* Current SnapshotReporter semantics are to specify stream as <SYS-NAME>.<STREAM-NAME> 

* We create topic in JobRunner so that AM can also emit metrics (if desired).

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/config/MetricsConfig.scala
+++ b/samza-core/src/main/scala/org/apache/samza/config/MetricsConfig.scala
@@ -31,6 +31,7 @@ object MetricsConfig {
   val METRICS_SNAPSHOT_REPORTER_INTERVAL= "metrics.reporter.%s.interval"
   val METRICS_TIMER_ENABLED= "metrics.timer.enabled"
   val METRICS_SNAPSHOT_REPORTER_BLACKLIST = "metrics.reporter.%s.blacklist"
+  val METRICS_SNAPSHOT_REPORTER_NAME_FOR_DIAGNOSTICS = "diagnosticsreporter"
 
   implicit def Config2Metrics(config: Config) = new MetricsConfig(config)
 }
--- a/samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
+++ b/samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
@@ -23,7 +23,7 @@ package org.apache.samza.job
 import java.util.concurrent.TimeUnit
 
 import org.apache.samza.SamzaException
-import org.apache.samza.config.{Config, JobConfig, MetricsConfig}
+import org.apache.samza.config.{Config, JobConfig, MetricsConfig, StreamConfig}
 import org.apache.samza.config.JobConfig.Config2Job
 import org.apache.samza.coordinator.stream.{CoordinatorStreamSystemConsumer, CoordinatorStreamSystemProducer}
 import org.apache.samza.coordinator.stream.messages.{Delete, SetConfig}
@@ -31,7 +31,7 @@ import org.apache.samza.job.ApplicationStatus.{Running, SuccessfulFinish}
 import org.apache.samza.metrics.MetricsRegistryMap
 import org.apache.samza.runtime.ApplicationRunnerMain.ApplicationRunnerCommandLine
 import org.apache.samza.runtime.ApplicationRunnerOperation
-import org.apache.samza.system.StreamSpec
+import org.apache.samza.system.{StreamSpec, SystemAdmins}
 import org.apache.samza.util.{CoordinatorStreamUtil, Logging, StreamUtil, Util}
 
 import scala.collection.JavaConverters._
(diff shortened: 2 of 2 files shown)


samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
```diff
@@ -118,6 +118,21 @@ class JobRunner(config: Config) extends Logging {
...
     coordinatorSystemProducer.stop()
 
+
+    // if diagnostics is enabled, create diagnostics stream as specified in config
+    if (new JobConfig(config).getDiagnosticsEnabled) {
+      val systemStreamName = new MetricsConfig(config).
+        getMetricsReporterStream(MetricsConfig.METRICS_SNAPSHOT_REPORTER_NAME_FOR_DIAGNOSTICS).get
+      val systemStream = StreamUtil.getSystemStreamFromNames(systemStreamName)
+      val diagnosticStreamSpec = new StreamSpec("diagnostics-stream-id", systemStream.getStream, systemStream.getSystem)
+      val diagnosticSysAdmin = systemFactory.getAdmin(systemStream.getSystem, config)
```
cameronlee314: This "systemFactory" comes from the coordinator system. However, it does not look like it is guaranteed that systemStream.getSystem is the same as the coordinator system, so I don't think you can always use "systemFactory" here.

samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
```diff
@@ -118,6 +118,21 @@ class JobRunner(config: Config) extends Logging {
...
+      info("creating stream %s using  sysAdmin %s " format (systemStreamName, diagnosticSysAdmin))
```
cameronlee314: Nit: extra space between "using" and "sysAdmin"

cameronlee314: I don't think SystemAdmin objects in general have a toString, so it might not be meaningful to log it.

samza-core/src/main/scala/org/apache/samza/config/MetricsConfig.scala
```diff
@@ -31,6 +31,7 @@ object MetricsConfig {
   val METRICS_SNAPSHOT_REPORTER_INTERVAL= "metrics.reporter.%s.interval"
   val METRICS_TIMER_ENABLED= "metrics.timer.enabled"
   val METRICS_SNAPSHOT_REPORTER_BLACKLIST = "metrics.reporter.%s.blacklist"
+  val METRICS_SNAPSHOT_REPORTER_NAME_FOR_DIAGNOSTICS = "diagnosticsreporter"
```
cameronlee314: Maybe "metrics.diagnostics.reporter"? Or if you don't consider this as "metrics", then it might be better to put the constant somewhere else.

rmatharu:
* Added SystemAdmins to JobRunner for all stream creation to get the right systemAdmin. 
* Added/fixed log messages

samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
```diff
@@ -82,12 +82,12 @@ class JobRunner(config: Config) extends Logging {
...
     val coordinatorSystemConsumer = new CoordinatorStreamSystemConsumer(config, new MetricsRegistryMap)
     val coordinatorSystemProducer = new CoordinatorStreamSystemProducer(config, new MetricsRegistryMap)
+    val systemAdmins = new SystemAdmins(config)
 
     // Create the coordinator stream if it doesn't exist
     info("Creating coordinator stream")
     val coordinatorSystemStream = CoordinatorStreamUtil.getCoordinatorSystemStream(config)
-    val systemFactory = CoordinatorStreamUtil.getCoordinatorSystemFactory(config)
-    val systemAdmin = systemFactory.getAdmin(coordinatorSystemStream.getSystem, config)
+    val systemAdmin = systemAdmins.getSystemAdmin(coordinatorSystemStream.getSystem)
```
cameronlee314: Could you rename this var to coordinatorSystemAdmin? Might help prevent someone else from using the wrong systemAdmin in the future.

samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
```diff
@@ -118,6 +118,28 @@ class JobRunner(config: Config) extends Logging {
     }
     coordinatorSystemProducer.stop()
 
+
+    // if diagnostics is enabled, create diagnostics stream if it doesnt exist
+    if (new JobConfig(config).getDiagnosticsEnabled) {
+      val DIAGNOSTICS_STREAM_ID = "samza-diagnostics-stream-id"
+      val systemStreamName = new MetricsConfig(config).
+        getMetricsReporterStream(MetricsConfig.METRICS_SNAPSHOT_REPORTER_NAME_FOR_DIAGNOSTICS).get
```
cameronlee314: Will this throw an exception if the config is not specified?

samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
```diff
@@ -118,6 +118,28 @@ class JobRunner(config: Config) extends Logging {
     }
     coordinatorSystemProducer.stop()
 
+
+    // if diagnostics is enabled, create diagnostics stream if it doesnt exist
+    if (new JobConfig(config).getDiagnosticsEnabled) {
+      val DIAGNOSTICS_STREAM_ID = "samza-diagnostics-stream-id"
+      val systemStreamName = new MetricsConfig(config).
```
cameronlee314: Is this just a stream, not a systemStream? Also, could you please add a prefix like "diagnostics" to make it more specific to this block?

samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
```diff
@@ -118,6 +118,28 @@ class JobRunner(config: Config) extends Logging {
...
+        getMetricsReporterStream(MetricsConfig.METRICS_SNAPSHOT_REPORTER_NAME_FOR_DIAGNOSTICS).get
+      val systemStream = StreamUtil.getSystemStreamFromNames(systemStreamName)
```
cameronlee314: Same about using "diagnostics" in var naming

samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
rmatharu: Renamed variables and added a ConfigException throw.

rmatharu:
Addressed all comments.
Renamed variables and added a ConfigException throw.

PR #604, opened 2018-08-09
SAMZA-1806: Allow `task.broadcast.inputs` to be set to empty string

abhishekshivanna:
This fix addresses the issue that getBroadcastSystemStreamPartitions
in TaskConfigJava throws an IllegalArgumentException exception when
`task.broadcast.inputs` is set to an empty string.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/config/Config.java
+++ b/samza-api/src/main/java/org/apache/samza/config/Config.java
@@ -22,6 +22,7 @@ package org.apache.samza.config;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


samza-api/src/main/java/org/apache/samza/config/Config.java
```diff
@@ -154,6 +155,9 @@ public double getDouble(String k) {
       return defaultValue;
 
     String value = get(k);
+    if (value.isEmpty()) {
+      return Collections.emptyList();
```
bharathkk: Can we use `StringUtils.isNotBlank(value)` instead of `value.isEmpty()`?
The latter doesn't cover `"   "`.

bharathkk: This case is no different than the case where the key is absent. Unless we want to throw some exception or validation error, I recommend returning `defaultValue`

PR #605, opened 2018-08-11
Enabling auto-discovery of regex input topics

rmatharu:
Work on this PR is stalled until https://github.com/apache/samza/pull/662 is checked in, because 
of overlapping classes are being re-written from Scala to Java in 662. 
So some of this work will need to be manually redone.


rmatharu:
Super-seeded by  https://github.com/apache/samza/pull/796/files

PR #607, opened 2018-08-13
SAMZA-1809: Add more logging to intermediate topic partition count inference

bharathkk:


bharathkk:
@prateekm please take a look.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -211,9 +211,12 @@ private void validateConfig() {
           }
         }
       }
+
+      log.info("Inferred the partition count for the operator {} as {}.", join.getOpId(), partitions);
```
prateekm: "for inputs for the join operator"

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -211,9 +211,12 @@ private void validateConfig() {
...
       // assign the partition count for intermediate streams
       for (StreamEdge edge : joinSpecToStreamEdges.get(join)) {
         if (edge.getPartitionCount() <= 0) {
+          log.info("Initialize the partition count for the associated edge {} to {}.", edge.getName(), partitions);
```
prateekm: "Set partition count for join input {} to {}". Same below.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -278,6 +281,7 @@ private static void calculateIntStreamPartitions(JobGraph jobGraph, Config confi
     }
     for (StreamEdge edge : jobGraph.getIntermediateStreamEdges()) {
       if (edge.getPartitionCount() <= 0) {
+        log.info("Initialize the partition count for edge {} to {}.", edge.getName(), partitions);
```
prateekm: There should be another place where we set the default partition count for edges that we could not infer. Can you add logging there too?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
prateekm: Is there a place where we log all the inputs and their known partition counts before we start this traversal?

prateekm:
@ahmedahamid FYI, in case you have any other suggestions.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
bharathkk: We already log the partition count for all sources and sink in `updateExistingPartitions` which happens prior to traversal. Were you looking for something else or would that suffice?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
bharathkk: Are you referring to the place where we initialize the `partitions` to be `PARTITIONS_UNKNOWN` during field declaration? 
I have all other places where we set the partition count covered.

prateekm: I mean in calculateIntStreamPartitions. I see now that you already added it, thanks. Maybe distinguish this in the log message, and clarify that it is for partition count we could not infer.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -211,9 +211,12 @@ private void validateConfig() {
           }
         }
       }
+
+      log.info("Inferred the partition count for inputs to the join operator {} as {}.", join.getOpId(), partitions);
```
prateekm: Can we move this to line 204 and include the name of the edge that caused this count to be inferred?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -278,6 +282,7 @@ private static void calculateIntStreamPartitions(JobGraph jobGraph, Config confi
     }
     for (StreamEdge edge : jobGraph.getIntermediateStreamEdges()) {
       if (edge.getPartitionCount() <= 0) {
+        log.info("Set the partition count for join input {} to {}.", edge.getName(), partitions);
```
ahmedahamid: Regarding log message:

1. `edge` is not necessarily — and probably not likely to be — join input.
2. My preference if I'm looking at logs produced by `ExecutionPlanner` is to have this message be different from the message logged on line 220 so that it's easy to discern *why* a partition count value was set any given value w/o having to remember to look at the method name where each message was logged.

I suggest we change:
1. this log message to something like: `"Set partition count of intermediate stream edge <edge-name> to <value>."`
2. the log message on line 220 to something like: `"Set partition count of intermediate stream edge <edge-name> participating in join <join-op-id> to <value>."`

ahmedahamid: Regarding log message:

1. `edge` is not necessarily — and probably not likely to be — join input.
2. My preference if I'm looking at logs produced by `ExecutionPlanner` is to have this message be different from the message logged on line 220 so that it's easy to discern *why* a partition count value was set any given value w/o having to remember to look at the method name where each message was logged.

I suggest we change:
1. this log message to something like: `"Set partition count of intermediate stream edge <edge-name> to <value>."`
2. the log message on line 220 to something like: `"Set partition count of intermediate stream edge <edge-name> participating in join <join-op-id> to <value>."`

I think it would also be useful if we add a log message, maybe two, to indicate whether the value of `partitions` came from `config.getInt()` on line 267 or was decided by the code on lines 273~275.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -204,16 +204,20 @@ private void validateConfig() {
           if (partitions == StreamEdge.PARTITIONS_UNKNOWN) {
             //if the partition is not assigned
             partitions = edgePartitions;
+            log.info("Inferred the partition count {} for the join operator {} from {}."
+                , new Object[] {partitions, join.getOpId(), edge.getName()});
```
ahmedahamid: I don't think we need `new Object[]` here. `log.info()` has an overload that accepts variable-length arguments.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -204,16 +204,20 @@ private void validateConfig() {
...
           } else if (partitions != edgePartitions) {
             throw  new SamzaException(String.format(
                 "Unable to resolve input partitions of stream %s for join. Expected: %d, Actual: %d",
```
ahmedahamid: Could you, please, add join-op-id to the exception message over here?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: My bad. I wasn't looking at the right version of slf4j.

PR #608, opened 2018-08-15
SAMZA-1807 table-api: inline async helpers

pdu-mn1:
Currently we have a set of helper methods, execute() for various
table access methods. Some of them are only called by one place and some
are called in two places. This change inlines the async handling back to
the table methods and removed the helpers.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/table/remote/RemoteReadWriteTable.java
+++ b/samza-core/src/main/java/org/apache/samza/table/remote/RemoteReadWriteTable.java
@@ -93,10 +93,30 @@ public class RemoteReadWriteTable<K, V> extends RemoteReadableTable<K, V> implem
     }
 
     writeMetrics.numPuts.inc();
-    return execute(writeRateLimiter, key, value, writeFn::putAsync, writeMetrics.putNs)
-        .exceptionally(e -> {
-            throw new SamzaException("Failed to put a record with key=" + key, (Throwable) e);
-          });
+
+    final long startNs = System.nanoTime();
+    CompletableFuture<Void> ioFuture;
+    if (writeRateLimiter.isRateLimited()) {
+      ioFuture = CompletableFuture
+          .runAsync(() -> writeRateLimiter.throttle(key, value), tableExecutor)
+          .thenCompose((r) -> writeFn.putAsync(key, value));
+    } else {
+      ioFuture = writeFn.putAsync(key, value);
+    }
+    if (callbackExecutor != null) {
+      ioFuture.thenApplyAsync(r -> {
+          writeMetrics.putNs.update(System.nanoTime() - startNs);
+          return r;
+        }, callbackExecutor);
+    } else {
+      ioFuture.thenApply(r -> {
+          writeMetrics.putNs.update(System.nanoTime() - startNs);
+          return r;
+        });
+    }
+    return ioFuture.exceptionally(e -> {
+        throw new SamzaException("Failed to put a record with key=" + key, e);
+      });
   }
 
   /**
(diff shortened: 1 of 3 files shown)


pdu-mn1:
@vjagadish1989 @prateekm This change takes care of both SAMZA-1807 and the javadoc warnings.

vjagadish1989:
Hey @pdu-mn1 : 

I'm +1 to this PR. On a related note @prateekm proposed that we consider if we need both configs (I'm +1 to his idea) I'm summarizing his proposal.  

max.pending.requests: it seems like we might be able to default it to "task.max.concurrency" or at-least derive from it since tables scoped within a task.
--
async.callback.pool.size: the current usage of the thread-pool appears to be just doing timer/metric updates. Since, we return a `CompletableFuture`, the user of the class can run callbacks by passing in in their own executors to `CompletableFuture#runAsync`.
--

This is unrelated to this PR per-say but would be good to consider regardless, before we release it in 1.0

vjagadish1989 (approved):
approved, looks good. thank you Peng for the contribution

pdu-mn1:
Thanks for reviewing! 

Regarding the suggestion: 

- The checkin of the initial async API impl does not have max.pending.requests at all. We already rely on task.max.concurrency for traffic control.

- async.callback.pool.size does not only process metrics it also processes task callbacks which is the main reason we provided the option for job to specify the thread pool. By default, we don't use any thread pool for these, so it is opt-in feature.

Does this make sense? @prateekm @vjagadish1989

prateekm:
@pdu-mn1 re async.callback.pool.size, we're returning users a Future, and they can already provide a custom executor when chaining operations on a future. From a user's perspective, this seems more straightforward and natural than trying to supply a configuration to us.

In other words, if users understand why and how they should configure this pool size, wouldn't it be simpler for them to just provide an executor for their thenAsync etc operations instead?

pdu-mn1:
I can think of two benefits of the current approach:

- it frees users from creating an executor and do runAsync with the executor for them, whereas they can simply supply a callback to us with the current approach with one line of config code. So this
is still a convenient feature for simple jobs. 

- With such config, the user is prompted to consider the implication of their callback methods supplied to us.

Neither reason is a blocker so I'm open to removing the config as well both you guys think otherwise. 

@prateekm @vjagadish1989

prateekm:
@pdu-mn1 My argument is the opposite of yours: 

If users understand when and why they should be executing callbacks on separate threads, it's simpler for them to use the standard CompletableFuture APIs to do this instead of an obscure Samza configuration. In fact, since we can't prevent them from using the Future APIs, and since they might not know the configuration exists, they're likely to use the Future APIs anyway.

If users don't understand the need for a separate callback executor, and we want to take the responsibility for telling them, configuration reference would be the wrong place to do so - we should document this in the API javadocs (i.e., close to when the need to understand this) instead. 

Another way to think about this: Even if we assume users always read all configurations, to make the configuration useful you'd have to describe when to use it and what to set it to. Wouldn't it be simpler to just do this in the method Javadocs and tell them to use separate executors?

pdu-mn1:
@prateekm Very fair. I will add documentation for the expectations/semantics of the callback, along with removing the config.

samza-core/src/main/java/org/apache/samza/table/remote/RemoteReadWriteTable.java
```diff
@@ -93,10 +93,30 @@ public void put(K key, V value) {
...
+          writeMetrics.putNs.update(System.nanoTime() - startNs);
+          return r;
+        }, callbackExecutor);
+    } else {
+      ioFuture.thenApply(r -> {
+          writeMetrics.putNs.update(System.nanoTime() - startNs);
+          return r;
+        });
+    }
+    return ioFuture.exceptionally(e -> {
```
nickpan47: Quick question: how do we propagate this exception to the main runloop if the callback in ioFuture.exceptionally() was invoked by an IO thread?

samza-core/src/main/java/org/apache/samza/table/remote/RemoteReadWriteTable.java
```diff
@@ -132,10 +152,29 @@ public void putAll(List<Entry<K, V>> entries) {
     List<Entry<K, V>> putRecords = records.stream()
         .filter(e -> e.getValue() != null).collect(Collectors.toList());
 
+    final long startNs = System.nanoTime();
+    CompletableFuture<Void> putFuture;
+    if (writeRateLimiter.isRateLimited()) {
+      putFuture = CompletableFuture
+          .runAsync(() -> writeRateLimiter.throttleRecords(putRecords), tableExecutor)
```
nickpan47: two comments: a) not sure why we need two different APIs for writeRateLimiter.throttleRecords() and writeRateLimiter.throttle(key, value). If those two can be consolidated to a single common API, this putAll code section is almost the same as the write function you wrote above. b) if we can consolidate the writeRateLimiter function call, it seems that we can avoid duplicated code logic between single write function and the putAll function here, just passing in the IO method to be performed (i.e. writeFn.putAllAsync() or writeFn.putAsync()) and the metrics to record the latency (i.e. writeMetrics.putNs or writeMetrics.putAllNs).

samza-core/src/main/java/org/apache/samza/table/remote/RemoteReadWriteTable.java
```diff
@@ -161,10 +200,31 @@ public void delete(K key) {
   public CompletableFuture<Void> deleteAsync(K key) {
     Preconditions.checkNotNull(key);
     writeMetrics.numDeletes.inc();
-    return execute(writeRateLimiter, key, writeFn::deleteAsync, writeMetrics.deleteNs)
-        .exceptionally(e -> {
-            throw new SamzaException(String.format("Failed to delete the record for " + key), (Throwable) e);
-          });
+
+    final long startNs = System.nanoTime();
+    CompletableFuture<Void> ioFuture;
```
nickpan47: The same logic to compose the ioFuture w/ rate throttler and callbacks are repeated many times. Can we consolidate them?

samza-core/src/main/java/org/apache/samza/table/remote/RemoteReadableTable.java
```diff
@@ -158,134 +174,30 @@ public V get(K key) {
...
-
-  /**
-   * Execute an async request given a table key
-   * @param key key of the table record
-   * @param method method to be executed
-   * @param timer latency metric to be updated
-   * @param <T> return type
-   * @return CompletableFuture of the operation
-   */
-  protected <T> CompletableFuture<T> execute(TableRateLimiter<K, V> rateLimiter,
```
nickpan47: It seems that you have a common reuseable function here before. What's the reason to repeat it in many places now?

prateekm:
@pdu-mn1 Does this PR need to be merged? If not, can you please close this?

prateekm:
@pdu-mn1 Can you please close this PR? Thanks!

weisong44:
Later changes have resolved issues intended by this PR, closing as it's not relevant anymore.

PR #611, opened 2018-08-20
SAMZA-1812: Added configuration for changelog to local store backed tables

weisong44:
Currently changelog for tables can be supported implicitly by adding user defined configuration, to be more user friendly we want to expose it though table API. This is applicable to RocksDB and in memory table implementation.
 - By default, changelog is enabled with auto-generated name <job-name><job-id>-table<table-id>
 - Added ability to disable changelog
 - Added ability to set changelog stream name, by default
 - Added validation of changelog stream name, and modify to confirm to Kafka topic name spec
 - Added configuration for replication factor
 - Disable changelog when user enables side input


samza-api/src/main/java/org/apache/samza/table/TableProvider.java
```diff
@@ -31,6 +32,12 @@
  */
 @InterfaceStability.Unstable
 public interface TableProvider {
+  /**
+   * Initialize TableProvider with job configuration
```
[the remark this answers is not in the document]
weisong44: Config is contained in ContainerContext, but the problem is when we are generating config, the other method was not called. And we couldn't get hold of the context objects too.

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
```diff
@@ -36,11 +36,20 @@
  */
 abstract public class BaseLocalStoreBackedTableDescriptor<K, V, D extends BaseLocalStoreBackedTableDescriptor<K, V, D>>
     extends BaseTableDescriptor<K, V, D> {
+
+  static final public String INTERNAL_ENABLE_CHANGELOG = "internal.enable.changelog";
```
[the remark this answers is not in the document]
weisong44: The reason is that when we create the TableSpec, we don't know the job name and id, so the info has to be stashed for later use during planning phase.

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
```diff
@@ -72,6 +131,14 @@ protected void validate() {
           String.format("Invalid side input configuration for table: %s. " +
               "Both side inputs and the processor must be provided", tableId));
     }
+    if (!enableChangelog) {
+      Preconditions.checkState(changelogStream == null,
```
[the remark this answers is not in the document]
weisong44: Here they actually need to be null

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableProvider.java
```diff
@@ -47,26 +46,20 @@
  * the table provider will not manage the lifecycle of the backing
  * stores.
  */
-abstract public class BaseLocalStoreBackedTableProvider implements TableProvider {
-
-  protected final Logger logger = LoggerFactory.getLogger(getClass());
+abstract public class BaseLocalStoreBackedTableProvider extends BaseTableProvider {
```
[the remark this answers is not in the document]
weisong44: +1

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableProvider.java
```diff
@@ -47,26 +46,20 @@
...
 
-  protected final TableSpec tableSpec;
+  final static private Pattern INVALID_CHANGELOG_CHARS = Pattern.compile("[^a-zA-Z0-9_-]+");
```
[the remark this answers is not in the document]
weisong44: +1

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
```diff
@@ -36,11 +36,20 @@
...
 abstract public class BaseLocalStoreBackedTableDescriptor<K, V, D extends BaseLocalStoreBackedTableDescriptor<K, V, D>>
     extends BaseTableDescriptor<K, V, D> {
+
+  static final public String INTERNAL_ENABLE_CHANGELOG = "internal.enable.changelog";
+  static final public String INTERNAL_CHANGELOG_STREAM = "internal.changelog.stream";
+  static final public String INTERNAL_CHANGELOG_REPLICATION_FACTOR = "internal.changelog.replication.factor";
+
   protected List<String> sideInputs;
   protected SideInputsProcessor sideInputsProcessor;
+  protected boolean enableChangelog = true;
```
bharathkk: Are we taking a conscious decision of having changelog enabled by default for tables? The current behavior of user-defined stores in high-level or low level has changelog disabled by default.

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
```diff
@@ -49,6 +58,8 @@ public BaseLocalStoreBackedTableDescriptor(String tableId) {
 
   public D withSideInputs(List<String> sideInputs) {
     this.sideInputs = sideInputs;
+    this.enableChangelog = false;
+    this.changelogStream = null;
```
bharathkk: we need to set the `changelogReplicationFactor` to `null` as well.

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
```diff
@@ -57,9 +68,57 @@ public D withSideInputsProcessor(SideInputsProcessor sideInputsProcessor) {
...
+   */
+  public D withChangelogStream(String changelogStream) {
+    this.changelogStream = changelogStream;
+    return (D) this;
+  }
+
+  /**
+   * Refer to <code>stores.store-name.changelog.replication.factor</code> in Samza configuration guide
+   *
+   * @param replicationFactor replilcation factor
```
bharathkk: nit: s/replilcation/replication

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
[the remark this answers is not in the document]
bharathkk: +1.
Maybe only check for `changelogStream` to be non-null in the case of `enableChangelog` is `true`. `changelogReplicationFactor` is an optional configuration.

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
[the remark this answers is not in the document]
bharathkk: all the more reason to use the descriptor in the planning phase so that we can get rid of the internal placeholders for passing configurations around. Looking forward to the refactor of using descriptors in the planning phase :)

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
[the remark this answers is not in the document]
weisong44: In fact both are optional, the checks were added just to avoid any errors from user's side.

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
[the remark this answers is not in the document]
weisong44: :)

samza-api/src/main/java/org/apache/samza/table/TableProvider.java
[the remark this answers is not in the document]
weisong44: +1 will make the changes

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
weisong44: Well, this is a discrepancy already. But I think it is probably more advantageous to keep it enabled by default. We have an option to disable it and documented in the javadoc. What do you think?

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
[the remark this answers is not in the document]
bharathkk: Then, we are not truly enabling changelog by default right? We still require applications to set the `changelogStream`. Is that correct?

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
bharathkk: Sure, I don't have a strong opinion on that. However, what does it mean to have a changelog for in-memory stores or RocksDb that is in-memory? 
Also, refer to my comments below on both `changelogStream` and `changelogReplicationFactor` both being optional parameters.

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableProvider.java
```diff
@@ -114,6 +110,30 @@ public Table getTable() {
           SerdeUtils.serialize("Side Inputs Processor", tableSpec.getSideInputsProcessor()));
     }
 
+    // Changelog configuration
+    String enableChangelog = tableSpec.getConfig().get(BaseLocalStoreBackedTableDescriptor.INTERNAL_ENABLE_CHANGELOG);
+    if ("true".equalsIgnoreCase(enableChangelog)) {
```
bharathkk: Not sure if this is out of scope for this PR. But, I guess since `TableSpec` is immutable, you can have the config inside TableSpec as a `MapConfig` instead of a `Map<String, String>` so that you can get the type information for free :)

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
```diff
@@ -57,9 +70,57 @@ public D withSideInputsProcessor(SideInputsProcessor sideInputsProcessor) {
...
+   * generated in format [job-name]-[job-id]-table-[table-id]. For both auto-generated
+   * and user specified changelog stream, invalid characters will be automatically
+   * replaced with '-' to confirm with pattern { @literal [a-zA-Z0-9_-]+ }
+   *
+   * Refer to <code>stores.store-name.changelog</code> in Samza configuration guide
+   *
+   * @param changelogStream changelog stream name
+   * @return this table descriptor instance
+   */
+  public D withChangelogStream(String changelogStream) {
```
bharathkk: Do we have a use-case where it would be useful to override the changelogStream? 
Also, can we not rewrite the changelog stream names here but throw when the names don't conform to the pattern `@literal [a-zA-Z0-9_-]+`

bharathkk: Do we have a use-case for overriding the auto-generated `changelogStream`? 
Is it possible to throw an exception instead of rewriting the changelog stream if it doesn't conform with the pattern `[a-zA-Z0-9_-]+`?

weisong44: Since this is more of an internal feature, wouldn't it more user friendly to auto-repair it?

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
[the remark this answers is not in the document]
weisong44: No, the check is in effect only if user DISABLES changelog. Primarily to avoid user oversight. When enabled, both parameters are optional.

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableProvider.java
weisong44: Yeah, I'd prefer to take this to another PR.

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseLocalStoreBackedTableDescriptor.java
prateekm: @weisong44 I'd prefer to throw an exception here as well. If the auto-generated name is invalid, it would imply that the job name is invalid. Job name is used in other places to create streams (e.g., coordinator, checkpoint) so that'd still be an issue, and users should fix that. 

In general, I'd prefer to not automatically rewrite names - it causes confusion since the name you'd expect from code or convention doesn't match the actual name. For example, this is an issue with some metrics reporters that "normalize" dashes in the metric name to underscores - it's very confusing to figure out the actual metric name.

weisong44: Based on what we discussed

samza-core/src/main/java/org/apache/samza/operators/BaseTableDescriptor.java
```diff
@@ -45,7 +45,7 @@
 
   /**
    * Constructs a table descriptor instance
-   * @param tableId Id of the table
+   * @param tableId Id of the table, it must confirm to pattern { @literal [\d\w-_]+ }
```
prateekm: Maybe `\\d\\w` to match java regex conventions?

samza-core/src/main/java/org/apache/samza/operators/StreamGraphSpec.java
```diff
@@ -145,12 +146,11 @@ public void setDefaultSerde(Serde<?> serde) {
 
   @Override
   public <K, V> Table<KV<K, V>> getTable(TableDescriptor<K, V, ?> tableDesc) {
+    Preconditions.checkState(isValidTableId(tableDesc.getTableId()), String.format(
```
prateekm: Maybe validate changelog and side input names using the Stream ID pattern too?

PR #612, opened 2018-08-20
SAMZA-1813: ApplicationRunner should use Planner generated configs for StreamManager

prateekm:


prateekm:
cc @vjagadish1989 @nickpan47 @Sanil15 for review. This is required for SEP-14: System and Stream Descriptors.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -82,8 +82,14 @@ ExecutionPlan getExecutionPlan(StreamApplication app, String runId, StreamManage
     cfg.put(ApplicationConfig.APP_MODE, mode.name());
 
     // create the physical execution plan
-    ExecutionPlanner planner = new ExecutionPlanner(new MapConfig(cfg), streamManager);
-    return planner.plan(specGraph);
+    Config config = new MapConfig(cfg);
```
vjagadish1989: trivial: there's already a member-variable named "config" in this class and it might be a bit confusing to disambiguate it with the local variable. maybe name it to something specific to clarify the usage here.

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
```diff
@@ -108,8 +114,8 @@ final void writePlanJsonFile(String planJson) {
   }
 
   @VisibleForTesting
-  StreamManager buildAndStartStreamManager() {
-    StreamManager streamManager = new StreamManager(this.config);
+  StreamManager buildAndStartStreamManager(Config config) {
+    StreamManager streamManager = new StreamManager(config);
```
vjagadish1989: nit-pick: maybe out of scope for this PR, but consider inlining the 2 line helper.

vjagadish1989 (approved):
approved; couple of minor f/b. 

Thanks Prateek!

samza-core/src/main/java/org/apache/samza/runtime/AbstractApplicationRunner.java
prateekm: This is for testing, since we don't (can't) inject the StreamManager into the runner/planner.

PR #616, opened 2018-08-24
SAMZA-1821: fix the IllegalStateException complaining duplicate key when fetching systemStream configs

weiqingy:
This PR is to fix IllegalStateException complaining duplicate key when fetching resource configs in SamzaSqlApplicationConfig.

## How was this patch tested?
Pass the local build and current unit tests.
Added a new unit test.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
@@ -113,13 +113,13 @@ public class SamzaSqlApplicationConfig {
     inputSystemStreamConfigBySource = queryInfo.stream()
         .map(QueryInfo::getSources)
         .flatMap(Collection::stream)
-        .collect(Collectors.toMap(Function.identity(), src -> ioResolver.fetchSourceInfo(src)));
+        .collect(Collectors.toMap(Function.identity(), ioResolver::fetchSourceInfo, (src1, src2) -> src1));
 
     Set<SqlIOConfig> systemStreamConfigs = new HashSet<>(inputSystemStreamConfigBySource.values());
 
     outputSystemStreamConfigsBySource = queryInfo.stream()
         .map(QueryInfo::getSink)
-        .collect(Collectors.toMap(Function.identity(), x -> ioResolver.fetchSinkInfo(x)));
+        .collect(Collectors.toMap(Function.identity(), ioResolver::fetchSinkInfo, (src1, src2) -> src1));
     systemStreamConfigs.addAll(outputSystemStreamConfigsBySource.values());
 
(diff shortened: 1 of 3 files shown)


samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
```diff
@@ -113,13 +113,13 @@ public SamzaSqlApplicationConfig(Config staticConfig) {
     inputSystemStreamConfigBySource = queryInfo.stream()
         .map(QueryInfo::getSources)
         .flatMap(Collection::stream)
-        .collect(Collectors.toMap(Function.identity(), src -> ioResolver.fetchSourceInfo(src)));
+        .collect(Collectors.toMap(Function.identity(), ioResolver::fetchSourceInfo, (src1, src2) -> src1));
```
bharathkk: How about we use `distinct()` before `collect()` to save some resolver calls to fetch source information?

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
```diff
@@ -113,13 +113,13 @@ public SamzaSqlApplicationConfig(Config staticConfig) {
...
 
     Set<SqlIOConfig> systemStreamConfigs = new HashSet<>(inputSystemStreamConfigBySource.values());
 
     outputSystemStreamConfigsBySource = queryInfo.stream()
         .map(QueryInfo::getSink)
-        .collect(Collectors.toMap(Function.identity(), x -> ioResolver.fetchSinkInfo(x)));
+        .collect(Collectors.toMap(Function.identity(), ioResolver::fetchSinkInfo, (src1, src2) -> src1));
```
bharathkk: same comment as above.

samza-sql/src/test/java/org/apache/samza/sql/testutil/TestSamzaSqlFileParser.java
```diff
@@ -24,7 +24,6 @@
 import java.io.PrintWriter;
 import java.util.List;
 
-import org.apache.samza.sql.testutil.SqlFileParser;
```
bharathkk: nit: revert

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
weiqingy: right. thanks for pointing this out.

samza-sql/src/test/java/org/apache/samza/sql/testutil/TestSamzaSqlFileParser.java
weiqingy: Just cleaning up code when I see it.

PR #619, opened 2018-08-30
Samza 1817: Long classpath support for non-split deployments 

Sanil15:
https://jira.apache.org/jira/browse/SAMZA-1817

### DIFF
--- a/samza-shell/src/main/bash/run-class.sh
+++ b/samza-shell/src/main/bash/run-class.sh
@@ -54,19 +54,12 @@ if [ -d "$JOB_LIB_DIR" ] && [ "$JOB_LIB_DIR" != "$BASE_LIB_DIR" ]; then
   # in case of different version of the same lib - we pick the highest
 
   #all jars from the fwk
-  base_jars=`ls $BASE_LIB_DIR/*.[jw]ar`
+  CLASSPATH=$CLASSPATH:"$BASE_LIB_DIR/*";
   #all jars from the job
-  job_jars=`for file in $JOB_LIB_DIR/*.[jw]ar; do name=\`basename $file\`; if [[ $base_jars != *"$name"* ]]; then echo "$file"; fi; done`
-  # get all lib jars and reverse sort it by versions
-  all_jars=`for file in $base_jars $job_jars; do echo \`basename $file|sed 's/.*[-]\([0-9]\+\..*\)[jw]ar$/\1/'\` $file; done|sort -t. -k 1,1nr -k 2,2nr -k 3,3nr -k 4,4nr|awk '{print $2}'`
-  # generate the class path based on the sorted result
-  for jar in $all_jars; do CLASSPATH=$CLASSPATH:$jar; done
-
-  # for debug only
-  echo base_jars=$base_jars
-  echo job_jars=$job_jars
-  echo all_jars=$all_jars
+  CLASSPATH=$CLASSPATH:"$JOB_LIB_DIR/*";
+  # generate the class path
   echo generated combined CLASSPATH=$CLASSPATH
+
 else
   #default behavior
   CLASSPATH=$CLASSPATH:"$BASE_LIB_DIR/*";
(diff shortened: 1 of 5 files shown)


Sanil15:
@sborya @bringhurst for review Thanks!

samza-shell/src/main/bash/run-class.sh
```diff
@@ -69,10 +69,7 @@ if [ -d "$JOB_LIB_DIR" ] && [ "$JOB_LIB_DIR" != "$BASE_LIB_DIR" ]; then
   echo generated combined CLASSPATH=$CLASSPATH
 else
   #default behavior
-  for file in $BASE_LIB_DIR/*.[jw]ar;
-  do
-    CLASSPATH=$CLASSPATH:$file
-  done
+  CLASSPATH=$CLASSPATH:"$BASE_LIB_DIR/*";
```
sborya: Looks good.
One question. Now we are going to pickup everything including *.xml. Is it a problem?

Sanil15: From Java Docs: https://docs.oracle.com/javase/7/docs/technotes/tools/windows/classpath.html

Class path entries can contain the basename wildcard character *, which is considered equivalent to specifying a list of all the files in the directory with the extension .jar or .JAR. For example, the class path entry foo/* specifies all JAR files in the directory named foo. A classpath entry consisting simply of * expands to a list of all the jar files in the current directory.

This is based on assumption that we don't need any war files in classpath

bringhurst:
Although unlikely, it might be good to consider jar files with spaces in the filename (maybe in a followup ticket, since it looks like this script would break in a few places if that happens).

samza-shell/src/main/bash/run-class.sh
shanthoosh: Just curious.

1. The control script previously was adding both the `jar` and `war` files to the `classpath`.  After this change, we're just adding only jar files into the `classpath`.  There're existing samza jobs in linkedin which use war files in classpath.  Do we know if this change will break the existing jobs?

2. In addition to wild carding jars, can we iterate and add *only* the wars to the classpath just to maintain backward compatibility? 

3.  Link cited in above comment refers to classpath expansion semantics for windows. Since we use linux extensively, do we know if the same behavior applies to linux as well?

samza-shell/src/main/bash/run-class.sh
```diff
@@ -69,10 +69,7 @@ if [ -d "$JOB_LIB_DIR" ] && [ "$JOB_LIB_DIR" != "$BASE_LIB_DIR" ]; then
   echo generated combined CLASSPATH=$CLASSPATH
 else
   #default behavior
```
shanthoosh: Please add a comment and describe the behavior of the `.*` wildcard java argument( General notion of `dir/.*`  represents all the files in the directory(dir). Here it is different and it includes only .JAR files in the directory to class-path).

samza-shell/src/main/bash/run-class.sh
Sanil15: @shanthoosh thanks for pointing out, I updated with the link for linux and solaris

samza-shell/src/main/bash/run-class.sh
Sanil15: 1. Will test with a few cases, but really curious why do we need to add .war files since these are supposed to be not libraries and deployed using servlet/jsp
In my opinion, if there is some library level logic in a WAR file, a person should extract that and export that as a JAR

2. We can be backwards compatible for now but for a major release we should force people to only use libraries from JAR artifacts and not Wars

3. I updated the link (thanks for pointing out)

Will add some comments here

Thanks

samza-shell/src/main/bash/run-class.sh
```diff
@@ -44,10 +44,12 @@ DEFAULT_LOG4J_FILE=$base_dir/lib/log4j.xml
 BASE_LIB_DIR="$base_dir/lib"
 # JOB_LIB_DIR will be set for yarn container in ContainerUtil.java
 # for others we set it to home_dir/lib
-JOB_LIB_DIR="${JOB_LIB_DIR:-$home_dir/lib}"
+JOB_LIB_DIR="${JOB_LIB_DIR:-$base_dir/lib}"
```
shanthoosh: Not related to your changes, but line number 27, 28 are No-op and can be safely deleted. Please delete them and test it.

Sanil15: Will open a cleaning Jira

PR #620, opened 2018-08-31
Samza 1836: StreamManager created before ExecutionPlanner should also apply configuration overrides

nickpan47:
Our integration test framework uses configuration overrides (i.e. jobs.*) to override the user system configuration set in the code (e.g. KafkaSystemDescriptor) to test systems. However, the StreamManager we created before calling ExecutionPlanner.plan() does not apply the overrides and causes failure in tests since the system was not correctly set to in-memory systems by configuration overrides.

### DIFF
--- a/samza-test/src/test/java/org/apache/samza/test/table/TestLocalTableWithSideInputs.java
+++ b/samza-test/src/test/java/org/apache/samza/test/table/TestLocalTableWithSideInputs.java
@@ -35,14 +35,13 @@ import org.apache.samza.config.StreamConfig;
 import org.apache.samza.operators.KV;
 import org.apache.samza.operators.StreamGraph;
 import org.apache.samza.operators.TableDescriptor;
-import org.apache.samza.operators.descriptors.GenericSystemDescriptor;
 import org.apache.samza.serializers.IntegerSerde;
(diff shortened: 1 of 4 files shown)


nickpan47:
@Sanil15 @prateekm please take a look. Thanks!

samza-test/src/test/java/org/apache/samza/test/table/TestLocalTableWithSideInputs.java
```diff
@@ -67,7 +67,8 @@ public void testJoinWithSideInputsTable() {
         Arrays.asList(TestTableData.generateProfiles(10)));
   }
 
-  @Test
+  // @Test
+  // TODO: re-enable after fixing the coordinator stream issue in SAMZA-1786
```
prateekm: @nickpan47 Can you clarify what's the coordinator stream issue? The linked ticket is talking about the MetadataStore API.

nickpan47: After the metadata store change, it seems that this test will invoke the coordinator consumer to the coordinator stream, where the StreamManager actually does not create it in the test. Hence, results in test failure. @shanthoosh is actively working on that fix.

PR #622, opened 2018-09-05
[EXPERIMENTAL] : Handle Repartitioning of Streams

MayureshGharat:
Modified the Groupers to handle Repartitioning. This is an opt in feature and can be enabled by setting the config : "auto.handle.repartition" to true.
This patch assumes that we add a new interface in the groupers that provides access to the previous mapping of tasks to SSPs.

This assumes that the number of partitions are always increased by a factor of 2.
This also assumes that the partitioning algorithm used by the producer is "hash(key) % partitions"


shanthoosh:
@MayureshGharat 
Changes from this PR had been reviewed as a part of #803 and  merged to trunk.
Would be better to close this PR.

MayureshGharat:
Closing the PR as per the comments above.

PR #623, opened 2018-09-05
SAMZA-1838: Make some minor improvements to ExecutionPlanner

ahmedahamid:
@prateekm @vjagadish1989 whenever you get a chance. This is an initial set of mostly minor changes.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
+++ b/samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
@@ -21,6 +21,7 @@ package org.apache.samza.execution;
 
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
+import com.google.common.collect.Sets;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -38,7 +39,6 @@ import org.apache.samza.operators.OperatorSpecGraph;
 import org.apache.samza.operators.spec.JoinOperatorSpec;
 import org.apache.samza.operators.spec.OperatorSpec;
 import org.apache.samza.system.StreamSpec;
-import org.apache.samza.system.SystemStream;
 import org.apache.samza.table.TableSpec;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -54,14 +54,16 @@ import static org.apache.samza.util.StreamUtil.*;
 public class ExecutionPlanner {
   private static final Logger log = LoggerFactory.getLogger(ExecutionPlanner.class);
 
-  static final int MAX_INFERRED_PARTITIONS = 256;
+  /* package private */ static final int MAX_INFERRED_PARTITIONS = 256;
 
   private final Config config;
+  private final StreamConfig streamConfig;
   private final StreamManager streamManager;
 
   public ExecutionPlanner(Config config, StreamManager streamManager) {
     this.config = config;
     this.streamManager = streamManager;
+    this.streamConfig = new StreamConfig(config);
   }
 
   public ExecutionPlan plan(OperatorSpecGraph specGraph) {
@@ -71,12 +73,10 @@ public class ExecutionPlanner {
     JobGraph jobGraph = createJobGraph(specGraph);
 
     // fetch the external streams partition info
-    updateExistingPartitions(jobGraph, streamManager);
+    fetchInputAndOutputStreamPartitions(jobGraph, streamManager);
 
-    if (!jobGraph.getIntermediateStreamEdges().isEmpty()) {
-      // figure out the partitions for internal streams
-      calculatePartitions(jobGraph);
-    }
+    // figure out the partitions for internal streams
+    calculatePartitions(jobGraph);
 
     return jobGraph;
   }
@@ -85,9 +85,9 @@ public class ExecutionPlanner {
     ApplicationConfig appConfig = new ApplicationConfig(config);
     ClusterManagerConfig clusterConfig = new ClusterManagerConfig(config);
     // currently we don't support host-affinity in batch mode
-    if (appConfig.getAppMode() == ApplicationConfig.ApplicationMode.BATCH
-        && clusterConfig.getHostAffinityEnabled()) {
-      throw new SamzaException("Host affinity is not supported in batch mode. Please configure job.host-affinity.enabled=false.");
+    if (appConfig.getAppMode() == ApplicationConfig.ApplicationMode.BATCH && clusterConfig.getHostAffinityEnabled()) {
+      throw new SamzaException(String.format("Host affinity is not supported in batch mode. Please configure %s=false.",
+          ClusterManagerConfig.CLUSTER_MANAGER_HOST_AFFINITY_ENABLED));
     }
   }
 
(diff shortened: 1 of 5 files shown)


samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -132,71 +136,72 @@ private void validateConfig() {
...
-      // retrieve the partition counts for the streams in this system
-      Map<String, Integer> streamToPartitionCount = streamManager.getStreamPartitionCounts(systemName, streamToStreamEdge.keySet());
-      // set the partitions of a stream to its StreamEdge
-      streamToPartitionCount.forEach((stream, partitionCount) -> {
-          streamToStreamEdge.get(stream).setPartitionCount(partitionCount);
-          log.info("Partition count is {} for stream {}", partitionCount, stream);
-        });
+    existingStreams.addAll(jobGraph.getInputStreams());
+    existingStreams.addAll(jobGraph.getOutputStreams());
+
```
vjagadish1989: We are effectively un-nesting the lookup in this map by one-level. Previously, the map of [streamName -> streamEdge] was per-system. It appears that we have changed this to a single-map of [streamName -> streamEdge]..

Do you think it could present issues when we read the same streamName from two different systems? As a concrete example at LinkedIn: imagine reading the same stream - PageViewEvent from both the "tracking" and "aggregate-tracking" Kafka clusters. We would like the "system-level" scope to be considered and the streams treated as 2 distinct ones - each with their own StreamEdges.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -265,23 +272,34 @@ private static void findReachableJoins(OperatorSpec operatorSpec, StreamEdge sou
...
+      int maxOutPartitions = maxPartitions(jobGraph.getOutputStreams());
       partitions = Math.max(maxInPartitions, maxOutPartitions);
 
       if (partitions > MAX_INFERRED_PARTITIONS) {
         partitions = MAX_INFERRED_PARTITIONS;
         log.warn(String.format("Inferred intermediate stream partition count %d is greater than the max %d. Using the max.",
             partitions, MAX_INFERRED_PARTITIONS));
       }
+    } else {
+      // Reject any zero or other negative values explicitly specified in config.
```
vjagadish1989: For my own understanding, can you provide an example of when this code-path would be reached?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: I wasn't aware of that. Thanks for the catch. I have changed the logic accordingly.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: Someone mistakenly setting `job.intermediate.stream.partitions` to 0 or -2

vjagadish1989: thanks for adding the validation!

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -71,12 +73,10 @@ public ExecutionPlan plan(OperatorSpecGraph specGraph) {
...
     // fetch the external streams partition info
-    updateExistingPartitions(jobGraph, streamManager);
+    fetchInputAndOutputStreamPartitions(jobGraph, streamManager);
 
-    if (!jobGraph.getIntermediateStreamEdges().isEmpty()) {
-      // figure out the partitions for internal streams
-      calculatePartitions(jobGraph);
-    }
+    // figure out the partitions for internal streams
+    calculatePartitions(jobGraph);
```
vjagadish1989: Did this method already handle the empty case ie., was the existing check just superficial?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -71,12 +73,10 @@ public ExecutionPlan plan(OperatorSpecGraph specGraph) {
     JobGraph jobGraph = createJobGraph(specGraph);
 
     // fetch the external streams partition info
-    updateExistingPartitions(jobGraph, streamManager);
+    fetchInputAndOutputStreamPartitions(jobGraph, streamManager);
```
vjagadish1989: thanks for the rename! i like the distinction between "fetching" vs "computing" partition-counts

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
vjagadish1989: If you still prefer the top-level map style, you could change the map from system-stream -> streamEdge. I'll leave that entirely up-to you.

vjagadish1989 (approved):
approved, thanks for the contribution Ahmed! If need be, update SAMZA-1838 so that the JIRA and code are consistent

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: My pleasure (^_^;)

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: Yes, but the primary reason I removed this check is because we need join input validation even in the absence of any intermediate streams. This is why the ExecutionPlanner did not complain about the simple case of joining 2 input streams w/ different partition counts.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: I'm glad you like it :) You're most welcome.

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
```diff
@@ -248,10 +248,10 @@ void validate() {
   }
 
   /**
-   * Validate the sources should have indegree being 0 and outdegree greater than 0
+   * Validate the input streams should have indegree being 0 and outdegree greater than 0
    */
   private void validateSources() {
```
shanthoosh: s/validateSources/validateInputStreams. 
Just to keep it consistent with naming in rest of the classes.

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
```diff
@@ -270,10 +270,10 @@ private void validateSources() {
   }
 
   /**
-   * Validate the sinks should have outdegree being 0 and indegree greater than 0
+   * Validate the output streams should have outdegree being 0 and indegree greater than 0
    */
   private void validateSinks() {
```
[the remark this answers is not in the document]
ahmedahamid: Done. Thanks for the catch.

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
ahmedahamid: Absolutely. Thanks again.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: I did consider that but this would entail creating a system-stream object for every stream after calling `streamManager.getStreamPartitionCounts()` to index into the top-level map.

I think one of the primary root causes of the complexity present in this method is in fact due to `streamManager.getStreamPartitionCounts()``s signature.

PR #624, opened 2018-09-05
New consumer for kafka system

sborya:
Remove SimpleConsumer  and BrokerProxy from Samza's KafkaSystemConsumer implementation.
Instead use KafkaConsumerProxy with high-level kafka consumer.


samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -174,7 +174,7 @@ public ClusterBasedJobCoordinator(Config coordinatorSystemConfig) {
 
     // build a JobModelManager and ChangelogStreamManager and perform partition assignments.
     changelogStreamManager = new ChangelogStreamManager(coordinatorStreamManager);
-    jobModelManager = JobModelManager.apply(coordinatorStreamManager, changelogStreamManager.readPartitionMapping());
+    jobModelManager = JobModelManager.apply(coordinatorStreamManager.getConfig(), changelogStreamManager.readPartitionMapping());
```
vjagadish1989: any reason to use a different constructor? would be nice to keep this PR just focussed on the KafkaConsumer and revert other changes

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,180 @@
...
+    consumerProps.setProperty(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, maxPollRecords);
+
+    // put overrides
+    consumerProps.putAll(injectProps);
+
+    return new KafkaConsumerConfig(consumerProps);
+  }
+
+  // group id should be unique per job
+  static String getConsumerGroupId(Config config) {
```
vjagadish1989: Make all of these as per-instance methods on the KafkaConsumerConfig class?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,447 @@
...
+import org.apache.samza.system.SystemStreamPartition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Separate thread that reads messages from kafka and puts them into the BlockingEnvelopeMap.
+ * This class is not thread safe. There will be only one instance of this class per KafkaSystemConsumer object.
+ * We still need some synchronization around kafkaConsumer. See pollConsumer() method for details.
+ */
+public class KafkaConsumerProxy<K, V> {
```
vjagadish1989: This class is the "BrokerProxy". Maybe, rename it to reflect that accurately

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,447 @@
...
+  private final Map<SystemStreamPartition, Long> nextOffsets = new ConcurrentHashMap<>();
+  // lags behind the high water mark, as reported by the Kafka consumer.
+  private final Map<SystemStreamPartition, Long> latestLags = new HashMap<>();
+
+  private volatile boolean isRunning = false;
+  private volatile Throwable failureCause = null;
+  private final CountDownLatch consumerPollThreadStartLatch = new CountDownLatch(1);
+
+  public KafkaConsumerProxy(Consumer<K, V> kafkaConsumer, String systemName, String clientId,
+      NewKafkaSystemConsumer.KafkaConsumerMessageSink messageSink, KafkaSystemConsumerMetrics samzaConsumerMetrics,
```
vjagadish1989: Extract MessageSink as a separate java class instead of a static inner class

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,447 @@
...
+
+  /* package private */ final Thread consumerPollThread;
+  private final Consumer<K, V> kafkaConsumer;
+  private final NewKafkaSystemConsumer.KafkaConsumerMessageSink sink;
+  private final KafkaSystemConsumerMetrics kafkaConsumerMetrics;
+  private final String metricName;
+  private final String systemName;
+  private final String clientId;
+  private final Map<TopicPartition, SystemStreamPartition> topicPartitions2SSP = new HashMap<>();
+  private final Map<SystemStreamPartition, MetricName> ssp2MetricName = new HashMap<>();
```
vjagadish1989: s/sspToMetricName/perPartitionMetrics

samza-kafka/src/main/scala/org/apache/samza/system/kafka/NewKafkaSystemConsumer.java
```diff
@@ -0,0 +1,412 @@
...
+  private final AtomicBoolean stopped = new AtomicBoolean(false);
+  private final AtomicBoolean started = new AtomicBoolean(false);
+  private final Config config;
+  private final boolean fetchThresholdBytesEnabled;
+
+  // This sink is used to transfer the messages from the proxy/consumer to the BlockingEnvelopeMap.
+  /* package private */ KafkaConsumerMessageSink messageSink;
+
+  // proxy is doing the actual reading
+  private KafkaConsumerProxy proxy;
```
vjagadish1989: Can this be made immutable too? A nice thing with doing this is that we can make the lifecycle of the "proxy" mirror the lifecycle of the "SystemConsumer" instance. 
ie., register() -> delegates to proxy.register(tp, offset) likewise, start() -> invokes proxy.start()

samza-kafka/src/main/scala/org/apache/samza/system/kafka/NewKafkaSystemConsumer.java
```diff
@@ -0,0 +1,412 @@
...
+
+  /* package private */final Map<TopicPartition, String> topicPartitions2Offset = new HashMap<>();
+  /* package private */final Map<TopicPartition, SystemStreamPartition> topicPartitions2SSP = new HashMap<>();
+
+  /* package private */ long perPartitionFetchThreshold;
```
vjagadish1989: final here since they're set during construction anyways

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,447 @@
...
+      // but we still just rethrow, and log it up the stack.
+      LOG.error("Caught a Kafka exception in pollConsumer", e);
+      throw e;
+    }
+
+    return processResults(records);
+  }
+
+  private Map<SystemStreamPartition, List<IncomingMessageEnvelope>> processResults(ConsumerRecords<K, V> records) {
+    if (records == null) {
```
shanthoosh: Just curious, in what scenarios does the records parameter will be null. It appears to me that this check is unnecessary and can be deleted.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/NewKafkaSystemConsumer.java
```diff
@@ -0,0 +1,412 @@
...
+      KafkaSystemConsumerMetrics metrics, Clock clock) {
+
+    super(metrics.registry(), clock, metrics.getClass().getName());
+
+    this.kafkaConsumer = kafkaConsumer;
+    this.samzaConsumerMetrics = metrics;
+    this.clientId = clientId;
+    this.systemName = systemName;
+    this.config = config;
+    this.metricName = systemName + " " + clientId;
```
shanthoosh: Minor: Prefer using String.format to concatenate here.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -46,49 +45,22 @@ object KafkaSystemFactory extends Logging {
...
-      timeout = timeout,
-      bufferSize = bufferSize,
-      fetchSize = fetchSize,
-      consumerMinSize = consumerMinSize,
-      consumerMaxWait = consumerMaxWait,
-      fetchThreshold = fetchThreshold,
-      fetchThresholdBytes = fetchThresholdBytes,
-      fetchLimitByBytesEnabled = config.isConsumerFetchThresholdBytesEnabled(systemName),
-      offsetGetter = offsetGetter)
+    NewKafkaSystemConsumer.getNewKafkaSystemConsumer(
```
shanthoosh: Why do we use the static method in `NewKafkaSystmeConsumer` to create the object and use the factory to return it. It kind of looks odd to me. Can we inline this invocation here and delete that method?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/NewKafkaSystemConsumer.java
```diff
@@ -0,0 +1,412 @@
...
+import org.apache.samza.system.SystemConsumer;
+import org.apache.samza.system.SystemStreamPartition;
+import org.apache.samza.util.BlockingEnvelopeMap;
+import org.apache.samza.util.Clock;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scala.Option;
+
+
+public class NewKafkaSystemConsumer<K, V> extends BlockingEnvelopeMap implements SystemConsumer {
```
shanthoosh: May be better to rename this to `HighLevelKafkaConsumer` or simply `KafkaSystemConsumer` or choose some other better name. NewKafkaConsumer can mean multiple things here(could be a upgrade of kafka consumer version or a different samza internal implementation etc).

samza-kafka/src/main/scala/org/apache/samza/system/kafka/NewKafkaSystemConsumer.java
```diff
@@ -0,0 +1,412 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(NewKafkaSystemConsumer.class);
+
+  private static final long FETCH_THRESHOLD = 50000;
+  private static final long FETCH_THRESHOLD_BYTES = -1L;
+
+  private final Consumer<K, V> kafkaConsumer;
+  private final String systemName;
+  private final KafkaSystemConsumerMetrics samzaConsumerMetrics;
+  private final String clientId;
```
shanthoosh: Can we remove the individual fields like clientID , groupId and other kafka consumer related properties from here and `KafkaConsumerProxy` and expose individual instance methods in `KafkaConsumerConfig` (E.x: `kafkaConsumerConfig.getClientId`).

Rather than  the following done everywhere(in the form of static method):
```
String clientId = KafkaConsumerConfig.getProducerClientId(config)
```
Can we do this instead: 
```
String clientId = kafkaConsumerConfig.getClientId() 
```
Just by passing in the kafkaConsumerConfig object alone to `NewKafkaConsumer` and `KafkaConsumerProxy` we can get rid of unnecessary fields and assignments everywhere.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/NewKafkaSystemConsumer.java
```diff
@@ -0,0 +1,412 @@
...
+
+  /**
+   * Compare two String offsets.
+   * Note. There is a method in KafkaAdmin that does that, but that would require instantiation of systemadmin for each consumer.
+   * @param off1
+   * @param off2
+   * @return see {@link Long#compareTo(Long)}
+   */
+  public static int compareOffsets(String off1, String off2) {
+    return Long.valueOf(off1).compareTo(Long.valueOf(off2));
```
shanthoosh: Rather than parsing the offsets here and comparing, it will be better to use the comparator from `kafkaSystemAdmin` here(KafkaSystemAdmin.offsetComparator method).

samza-kafka/src/main/scala/org/apache/samza/system/kafka/NewKafkaSystemConsumer.java
```diff
@@ -0,0 +1,412 @@
...
+        LOG.warn("Failure cause is not populated for KafkaConsumerProxy");
+        throw new SamzaException("KafkaConsumerProxy has stopped");
+      }
+    }
+
+    Map<SystemStreamPartition, List<IncomingMessageEnvelope>> res = super.poll(systemStreamPartitions, timeout);
+    return res;
+  }
+
+  public static TopicAndPartition toTopicAndPartition(TopicPartition tp) {
```
shanthoosh: All of the following static methods(`toTopicAndPartition(topicParittion)`, `toTopicAndPartition(SystemStreamPartition`)  here seems to be named with similar names and have different parameter type.  Do we need a static method here just to create a object. Can we instead inline the method invocations?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/NewKafkaSystemConsumer.java
```diff
@@ -0,0 +1,412 @@
...
+   * return system name for this consumer
+   * @return system name
+   */
+  public String getSystemName() {
+    return systemName;
+  }
+
+  ////////////////////////////////////
+  // inner class for the message sink
+  ////////////////////////////////////
```
shanthoosh: Correct me if i'm wrong here. This appears to be used only within the `KafkaConsumerProxy`. Can we make it an internal static class of the `KafkaConsumerProxy` class instead?

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,180 @@
...
+
+    // Disable consumer auto-commit because Samza controls commits
+    consumerProps.setProperty(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "false");
+
+    // Translate samza config value to kafka config value
+    consumerProps.setProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG,
+        getAutoOffsetResetValue(consumerProps));
+
+    // make sure bootstrap configs are in ?? SHOULD WE FAIL IF THEY ARE NOT?
+    if (!subConf.containsKey(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG)) {
```
shanthoosh: Previously we were using SimpleConsumer in samza which had configurations (such as `zookeeper.connect`) which are no longer supported in the new kafkaConsumer.

New kafka consumer configs: http://kafka.apache.org/documentation/#newconsumerconfigs. 

We need to update our samza configuration table. 

1. Delete the old configuration references (systems.system-name.consumer.zookeeper.connect, etc) 
2. Stress some of the important features of new kafkaConsumer(`SSL`, `keySerializer`, `valueSerializer`, `max.poll.records`). Our implementation overrides the defaults with user-defined configurations and it will be better to document it.

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,180 @@
...
+   * Settings for auto.reset in samza are different from settings in Kafka (auto.offset.reset) - need to convert
+   * "largest" -> "latest"
+   * "smallest" -> "earliest"
+   * "none" -> "none"
+   * "none" - will fail the kafka consumer, if offset is out of range
+   * @param properties All consumer related {@link Properties} parsed from samza config
+   * @return String representing the config value for "auto.offset.reset" property
+   */
+  static String getAutoOffsetResetValue(Properties properties) {
+    String autoOffsetReset = properties.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, KAFKA_OFFSET_LATEST);
```
shanthoosh: 1. Can you please share the rationale why we want to allow users to define configuration `auto.reset.offset` with predefined kafka constants  directly? 
2. Previously, `samza`  did not allow users to specify value anything other than `largest`, `smallest` for the configuration `auto.reset.offset`. Now if we want to allow users to specify the predefined kafka  `auto.reset.offset` configurations, can we update the configuration table documentation in accordance with this implementation.

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,180 @@
...
+  private static final String SAMZA_OFFSET_SMALLEST = "smallest";
+  private static final String KAFKA_OFFSET_LATEST = "latest";
+  private static final String KAFKA_OFFSET_EARLIEST = "earliest";
+  private static final String KAFKA_OFFSET_NONE = "none";
+
+  /*
+   * By default, KafkaConsumer will fetch ALL available messages for all the partitions.
+   * This may cause memory issues. That's why we will limit the number of messages per partition we get on EACH poll().
+   */
+  private static final String KAFKA_CONSUMER_MAX_POLL_RECORDS_DEFAULT = "100";
```
shanthoosh: Minor: 
s/KAFKA_CONSUMER_MAX_POLL_RECORDS_DEFAULT/DEFAULT_KAFKA_CONSUMER_MAX_POLL_RECORDS

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,180 @@
...
+      return autoOffsetReset;
+    }
+
+    switch (autoOffsetReset) {
+      case SAMZA_OFFSET_LARGEST:
+        return KAFKA_OFFSET_LATEST;
+      case SAMZA_OFFSET_SMALLEST:
+        return KAFKA_OFFSET_EARLIEST;
+      default:
+        return KAFKA_OFFSET_LATEST;
```
dnishimura: Might be useful to document the default return value and also log the reset value if it isn't one of `latest` or `smallest`

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,180 @@
...
+   * This may cause memory issues. That's why we will limit the number of messages per partition we get on EACH poll().
+   */
+  private static final String KAFKA_CONSUMER_MAX_POLL_RECORDS_DEFAULT = "100";
+
+
+  public KafkaConsumerConfig(Properties props) {
+    super(props);
+  }
+
+  public static KafkaConsumerConfig getKafkaSystemConsumerConfig(Config config, String systemName, String clientId,
```
dnishimura: Good to document public factory methods.

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,180 @@
...
+  private static final String KAFKA_OFFSET_NONE = "none";
+
+  /*
+   * By default, KafkaConsumer will fetch ALL available messages for all the partitions.
+   * This may cause memory issues. That's why we will limit the number of messages per partition we get on EACH poll().
+   */
+  private static final String KAFKA_CONSUMER_MAX_POLL_RECORDS_DEFAULT = "100";
+
+
+  public KafkaConsumerConfig(Properties props) {
```
dnishimura: Can this be private since there's already a `getKafkaSystemConsumerConfig` factory method?

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,193 @@
...
+
+    return new KafkaConsumerConfig(consumerProps);
+  }
+
+  // group id should be unique per job
+  private static String getConsumerGroupId(Config config) {
+    JobConfig jobConfig = new JobConfig(config);
+    Option<String> jobIdOption = jobConfig.getJobId();
+    Option<String> jobNameOption = jobConfig.getName();
+    return (jobNameOption.isDefined() ? jobNameOption.get() : "undefined_job_name") + "-" + (jobIdOption.isDefined()
```
shanthoosh: Minor(Please change if makes sense): 
Can simplify the return statement with option.getOrElse().
```
String.format("%s-%s", jobNameOption.getOrElse("undefined_job_name"), jobIdOption.getOrElse("undefined_job_name")); 
```

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,447 @@
...
+    }
+
+    ConsumerRecords<K, V> records;
+    // make a call on the client
+    try {
+      // Currently, when doing checkpoint we are making a safeOffset request through this client, thus we need to synchronize
+      // them. In the future we may use this client for the actually checkpointing.
+      synchronized (kafkaConsumer) {
+        // Since we are not polling from ALL the subscribed topics, so we need to "change" the subscription temporarily
+        kafkaConsumer.pause(topicPartitionsToPause);
```
shanthoosh: Just curious on this block. 
We pause and resume the `KafkaConsumer` for the SSP's from this list `topicPartitionsToPause` within the synchronized block.  Why do we need to temporarily pause them if all we do is resume them right after poll? Can you please share the significance of this division and pausing, resuming a subset before/after poll?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,406 @@
...
+
+    if (!stopped.compareAndSet(false, true)) {
+      LOG.warn("attempting to stop stopped consumer.");
+      return;
+    }
+
+    // stop the proxy (with 5 minutes timeout)
+    if (proxy != null) {
+      LOG.info("Stopping proxy " + proxy);
+      proxy.stop(TimeUnit.MINUTES.toMillis(5));
```
shanthoosh: Why this has to wait for 5 mins to stop the proxy thread. Can we lower the stop timeout to reasonably low value?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,406 @@
...
+    // Create the proxy to do the actual message reading. It is a separate thread that reads the messages from the stream
+    // and puts them into the sink.
+    proxy = new KafkaConsumerProxy(kafkaConsumer, systemName, clientId, messageSink, samzaConsumerMetrics, metricName);
+    LOG.info("Created consumer proxy: " + proxy);
+
+    LOG.info("Created SamzaKafkaSystemConsumer for system={}, clientId={}, metricName={}, KafkaConsumer={}", systemName,
+        clientId, metricName, this.kafkaConsumer.toString());
+  }
+
+  public static <K, V> KafkaSystemConsumer getNewKafkaSystemConsumer(String systemName, Config config,
```
shanthoosh: Minor: s/getNewKafkaSystemConsumer/getKafkaSystemConsumer

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,447 @@
...
+
+  boolean isRunning() {
+    return isRunning;
+  }
+
+  Throwable getFailureCause() {
+    return failureCause;
+  }
+
+  public void stop(long timeout) {
```
shanthoosh: Would be great to take TimeUnit or duration as the second parameter here. From the arguments, it's not clear what is the timeunit of timeout parameter.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,447 @@
...
+    LOG.info("Shutting down KafkaConsumerProxy poll thread:" + consumerPollThread.getName());
+
+    isRunning = false;
+    try {
+      consumerPollThread.join(timeout);
```
shanthoosh: Do we need to interrupt the consumerPoll thread here after the timeout expires. Let's say If the thread did not die after this timeout (say got stuck in kafkaConsumer.fetch), what is the line of action?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
shanthoosh: > if Kafka decides throw an exception we don't handle.


In that scenario, it will be propogated back to the caller, control-flow won't even reach the processResults method here.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,447 @@
...
+
+    this.kafkaConsumerMetrics.registerClientProxy(metricName);
+
+    consumerPollThread = new Thread(createProxyThreadRunnable());
+    consumerPollThread.setDaemon(true);
+    consumerPollThread.setName(
+        "Samza KafkaConsumerProxy Poll " + consumerPollThread.getName() + " - " + systemName);
+  }
+
+  public void start() {
```
shanthoosh: Do we need to handle the interrupt in the consumer poll thread? Existing implementation doesn't seem to handle it and the consumer poll thread is interrupted during `KafkaSystemConsumer.stop()`

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,447 @@
...
+        results.put(ssp, listMsgs);
+      }
+
+      final K key = record.key();
+      final Object value = record.value();
+      final IncomingMessageEnvelope imEnvelope =
+          new IncomingMessageEnvelope(ssp, String.valueOf(record.offset()), key, value, getRecordSize(record));
+      listMsgs.add(imEnvelope);
+    }
+    if (LOG.isDebugEnabled()) {
```
shanthoosh: Why do we have  the log level guard here. In slf4j/log4j combination, event can be logged unconditionally and message formatting is applied only based upon the configured  log level of the application. To me this check introduces unnecessary complexity, what is the value add we get from this?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -46,49 +45,22 @@ object KafkaSystemFactory extends Logging {
...
-      timeout = timeout,
-      bufferSize = bufferSize,
-      fetchSize = fetchSize,
-      consumerMinSize = consumerMinSize,
-      consumerMaxWait = consumerMaxWait,
-      fetchThreshold = fetchThreshold,
-      fetchThresholdBytes = fetchThresholdBytes,
-      fetchLimitByBytesEnabled = config.isConsumerFetchThresholdBytesEnabled(systemName),
-      offsetGetter = offsetGetter)
+    KafkaSystemConsumer.getNewKafkaSystemConsumer(
```
vjagadish1989: Can we just do a new KafkaSystemConsumer()?

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
vjagadish1989: +1

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
sborya: This change came up during testing and seems like a good thing to keep.
CoordinatorStreamManager has its own life cycle, so unnecessary passing a reference to it into another class (with its own life cycle) is not desired.This way we can stop it here, without worrying the other reference.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
sborya: 1. Naming came from BrokerProxy, and later LiKafkaConsumerProxy. So it may be a good idea to keep the name to preserve the history. 
2. Do you mean keep the class or the constructor package private? I didn't see a lot of package private classes in our code.

samza-core/src/main/scala/org/apache/samza/job/local/ProcessJobFactory.scala
```diff
@@ -61,6 +61,7 @@ class ProcessJobFactory extends StreamJobFactory with Logging {
     }
 
     changelogStreamManager.writePartitionMapping(taskPartitionMappings)
+    coordinatorStreamManager.stop()
```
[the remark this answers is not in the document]
sborya: I'd rather not add even more stuff to this PR. It is too big already.

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,193 @@
...
+    final Properties consumerProps = new Properties();
+    consumerProps.putAll(subConf);
+
+    consumerProps.setProperty(ConsumerConfig.GROUP_ID_CONFIG, groupId);
+    consumerProps.setProperty(ConsumerConfig.CLIENT_ID_CONFIG, clientId);
+
+    //Kafka client configuration
+
+    // Disable consumer auto-commit because Samza controls commits
+    consumerProps.setProperty(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "false");
```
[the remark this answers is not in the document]
shanthoosh: Since we already manage offsets in samza through checkpoint topic, we should not allow users to turn auto-commit on to manage checkpoints in kafka(when using the kafkaSystemConsumer). Kafka-autocommit is unnecessary and could potentially create confusion given that we have our own mechanism to rewind/fast-forward the checkpointed offsets(with checkpoint tool). 

(Ex: What if the user turns on kafka auto-commit, stops the job and  updates offsets stored in kafka. After this user would expect samza to resume from the updated offset in kafka. But the job would resume with the checkpointed offset stored in samza checkpoint topic)

I think we should log the warning here if auto-commit is turned on and document in configuration table that it should not be turned on.

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,193 @@
...
+      String bootstrapServer =
+          config.get(String.format("systems.%s.producer.%s", systemName, ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG));
+      if (StringUtils.isEmpty(bootstrapServer)) {
+        throw new SamzaException("Missing " + ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG + " config  for " + systemName);
+      }
+      consumerProps.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServer);
+    }
+
+    // Always use default partition assignment strategy. Do not allow override.
+    consumerProps.setProperty(ConsumerConfig.PARTITION_ASSIGNMENT_STRATEGY_CONFIG, RangeAssignor.class.getName());
```
shanthoosh: Since we control the partition assignment strategy in `JobCoordinator` via generated `JobModel`,  setting this configuration is unnecessary and can be removed.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
sborya: Done.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
sborya: Done.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
sborya: In this case we need it (because there is a loop), but in other cases I've removed it.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
sborya: Done.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
sborya: I guess you are right, but this is just a good coding style - check function's arguments :)

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,446 @@
...
+      throw new SamzaException("Unknown/unregistered ssp in latestLags request: " + ssp);
+    }
+    return lag;
+  }
+
+  /*
+    Using the consumer to poll the messages from the stream.
+   */
+  private void fetchMessages() {
+    Set<SystemStreamPartition> SSPsToFetch = new HashSet<>();
```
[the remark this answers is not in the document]
sborya: but these are SSPs, not partitions!?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -46,49 +45,27 @@ object KafkaSystemFactory extends Logging {
...
-      timeout = timeout,
-      bufferSize = bufferSize,
-      fetchSize = fetchSize,
-      consumerMinSize = consumerMinSize,
-      consumerMaxWait = consumerMaxWait,
-      fetchThreshold = fetchThreshold,
-      fetchThresholdBytes = fetchThresholdBytes,
-      fetchLimitByBytesEnabled = config.isConsumerFetchThresholdBytesEnabled(systemName),
-      offsetGetter = offsetGetter)
+    val kc = new KafkaSystemConsumer(kafkaConsumer, systemName, config, clientId, metrics, new SystemClock)
```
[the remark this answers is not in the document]
sborya: Done.

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+  }
+
+  static class MockKafkaSystmeCosumer extends KafkaSystemConsumer {
+    public MockKafkaSystmeCosumer(Consumer kafkaConsumer, String systemName, Config config, String clientId,
+        KafkaSystemConsumerMetrics metrics, Clock clock) {
+      super(kafkaConsumer, systemName, config, clientId, metrics, clock);
+    }
+
+    //@Override
```
shanthoosh: Please delete these comments.

samza-kafka/src/test/java/org/apache/kafka/clients/consumer/TestKafkaConsumerConfig.java
```diff
@@ -0,0 +1,137 @@
...
+
+public class TestKafkaConsumerConfig {
+  private final Map<String, String> props = new HashMap<>();
+  public final static String SYSTEM_NAME = "testSystem";
+  public final static String KAFKA_PRODUCER_PROPERTY_PREFIX = "systems." + SYSTEM_NAME + ".producer.";
+  public final static String KAFKA_CONSUMER_PROPERTY_PREFIX = "systems." + SYSTEM_NAME + ".consumer.";
+  private final static String CLIENT_ID = "clientId";
+
+  @Before
+  public void setProps() {
```
shanthoosh: Can we please delete this empty method.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,391 @@
...
+    }
+  }
+
+  /*
+   record the ssp and the offset. Do not submit it to the consumer yet.
+   */
+  @Override
+  public void register(SystemStreamPartition systemStreamPartition, String offset) {
+    if (started.get()) {
+      String msg =
```
shanthoosh: Why do we log the message here and then throw up. The codeflow that triggers `register(ssp, offset)` already logs any exception. Please delete it if it's redundant.

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,194 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package org.apache.kafka.clients.consumer;
```
prateekm: This should be in a Samza package.

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
prateekm: I agree about the utility of allowing users to set auto.reset.offset, but doesn't make sense to allow using Samza's config values for Kafka's config keys. We should to remove this method and pass the user provided consumer config as-is.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -289,7 +289,10 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
     properties
   }
 
-  // kafka config
+  /**
+    * @deprecated Use KafkaConsumerConfig
+    */
+  @Deprecated
```
prateekm: Can we delete this method?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,391 @@
...
+
+  @Override
+  public Map<SystemStreamPartition, List<IncomingMessageEnvelope>> poll(
+      Set<SystemStreamPartition> systemStreamPartitions, long timeout) throws InterruptedException {
+
+    // check if the proxy is running
+    if (!proxy.isRunning()) {
+      stop();
+      if (proxy.getFailureCause() != null) {
+        String message = "KafkaConsumerProxy has stopped";
```
prateekm: Extract out of the if block and reuse in SamzaException for the else block.

samza-kafka/src/test/java/org/apache/kafka/clients/consumer/TestKafkaConsumerConfig.java
```diff
@@ -0,0 +1,137 @@
...
+        kafkaConsumerConfig.getString(ConsumerConfig.CLIENT_ID_CONFIG),
+        CLIENT_ID);
+
+    Assert.assertEquals(
+        kafkaConsumerConfig.getString(ConsumerConfig.GROUP_ID_CONFIG),
+        KafkaConsumerConfig.getConsumerGroupId(config));
+  }
+
+  @Test
+  // test stuff that should not be overridden
```
prateekm: Comment should be above @Test annotation.

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -0,0 +1,224 @@
...
+public class TestKafkaSystemConsumer {
+  public final String TEST_SYSTEM = "test-system";
+  public final String TEST_STREAM = "test-stream";
+  public final String TEST_CLIENT_ID = "testClientId";
+  public final String BOOTSTRAP_SERVER = "127.0.0.1:8888";
+  public final String FETCH_THRESHOLD_MSGS = "50000";
+  public final String FETCH_THRESHOLD_BYTES = "100000";
+
+  @Before
+  public void setUp() {
```
prateekm: Don't need empty setup method.

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -0,0 +1,224 @@
...
+
+  }
+
+  private KafkaSystemConsumer setupConsumer(String fetchMsg, String fetchBytes) {
```
prateekm: createConsumer?

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -0,0 +1,224 @@
...
+
+    return newKafkaSystemConsumer;
+  }
+
+  @Test
+  public void testConfigValidations() {
+
+    final KafkaSystemConsumer consumer = setupConsumer(FETCH_THRESHOLD_MSGS, FETCH_THRESHOLD_BYTES);
+
+    consumer.start();
```
prateekm: Should stop the consumer to prevent leaking Consumers during tests.

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  // mock kafkaConsumer and SystemConsumer
+  static class MockKafkaConsumer extends KafkaConsumer {
+    public MockKafkaConsumer(Map<String, Object> configs) {
+      super(configs);
+    }
+  }
+
+  static class MockKafkaSystmeCosumer extends KafkaSystemConsumer {
```
prateekm: Typo: MockKafkaSystemConsumer

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,456 @@
...
+      // we need to wait until the thread starts
+      while (!isRunning && failureCause == null) {
+        try {
+          consumerPollThreadStartLatch.await(3000, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+          LOG.info("Got InterruptedException", e);
+        }
+      }
+    } else {
+      LOG.debug("Tried to start an already started KafkaConsumerProxy (%s). Ignoring.", this.toString());
```
prateekm: Should be warn, since this is not a normal use case.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,391 @@
...
+    if (topicPartitions2Offset.size() <= 0) {
+      LOG.warn("Consumer {} is not subscribed to any SSPs", this);
+    }
+
+    topicPartitions2Offset.forEach((tp, startingOffsetString) -> {
+      long startingOffset = Long.valueOf(startingOffsetString);
+
+      try {
+        synchronized (kafkaConsumer) {
+          // TODO in the future we may need to add special handling here for BEGIN/END_OFFSET
```
prateekm: Not sure what this is talking about. Can you clarify what this means?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,391 @@
...
+
+      try {
+        put(ssp, envelope);
+      } catch (InterruptedException e) {
+        throw new SamzaException(
+            String.format("Interrupted while trying to add message with offset %s for ssp %s", envelope.getOffset(),
+                ssp));
+      }
+    }
+  }  // end of KafkaMessageSink class
```
prateekm: Can delete these comments.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -46,49 +45,27 @@ object KafkaSystemFactory extends Logging {
...
   def getConsumer(systemName: String, config: Config, registry: MetricsRegistry): SystemConsumer = {
-    val clientId = KafkaUtil.getClientId("samza-consumer", config)
+    val clientId = KafkaConsumerConfig.getConsumerClientId( config)
     val metrics = new KafkaSystemConsumerMetrics(systemName, registry)
 
-    // Kind of goofy to need a producer config for consumers, but we need metadata.
-    val producerConfig = config.getKafkaSystemProducerConfig(systemName, clientId)
-    val bootstrapServers = producerConfig.bootsrapServers
-    val consumerConfig = config.getKafkaSystemConsumerConfig(systemName, clientId)
+    val kafkaConsumer = KafkaSystemConsumer.getKafkaConsumerImpl(systemName, clientId, config)
```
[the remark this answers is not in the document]
sborya: I've used this static method to avoid awkward scala-java conversions. But this part may change in the other patch (changing SystemAdmin) and I will review it there.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -46,49 +45,27 @@ object KafkaSystemFactory extends Logging {
...
+    kafkaSystemConsumer
   }
 
   def getProducer(systemName: String, config: Config, registry: MetricsRegistry): SystemProducer = {
-    val clientId = KafkaUtil.getClientId("samza-producer", config)
+    val clientId = KafkaConsumerConfig.getProducerClientId(config)
     val injectedProps = KafkaSystemFactory.getInjectedProducerProperties(systemName, config)
     val producerConfig = config.getKafkaSystemProducerConfig(systemName, clientId, injectedProps)
-    val getProducer = () => { new KafkaProducer[Array[Byte], Array[Byte]](producerConfig.getProducerProperties) }
+    val getProducer = () => {
```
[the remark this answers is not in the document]
sborya: this is not related to this patch. Just formatting.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -119,13 +96,13 @@ class KafkaSystemFactory extends SystemFactory with Logging {
...
-       val changelogInfo = ChangelogInfo(replicationFactor, config.getChangelogKafkaProperties(storeName))
-       info("Creating topic meta information for topic: %s with replication factor: %s" format (topicName, replicationFactor))
-       (topicName, changelogInfo)
-    }}
+    val topicMetaInformation = storeToChangelog.map { case (storeName, topicName) => {
+      val replicationFactor = config.getChangelogStreamReplicationFactor(storeName).toInt
+      val changelogInfo = ChangelogInfo(replicationFactor, config.getChangelogKafkaProperties(storeName))
+      info("Creating topic meta information for topic: %s with replication factor: %s" format(topicName, replicationFactor))
+      (topicName, changelogInfo)
+    }
```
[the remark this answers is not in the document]
sborya: Scala formatting is causing many issues with checkStyle. So I prefer to delegate this task to Intelij. If checkstyle fails I will address it.

samza-kafka/src/test/java/org/apache/kafka/clients/consumer/TestKafkaConsumerConfig.java
sborya: sure.

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
sborya: Fixed the typo. Thanks.

samza-kafka/src/main/scala/org/apache/kafka/clients/consumer/KafkaConsumerConfig.java
sborya: Right. Fixed.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,444 @@
...
+  // Get the latest lag for a specific SSP.
+  private long getLatestLag(SystemStreamPartition ssp) {
+    Long lag = latestLags.get(ssp);
+    if (lag == null) {
+      throw new SamzaException("Unknown/unregistered ssp in latestLags request: " + ssp);
+    }
+    return lag;
+  }
+
+  // Using the consumer to poll the messages from the stream.
```
prateekm: This is a similar comment to pollConsumer. I'd prefer to clarify how they're related. Also, since pollConsumer is only called within this method, this method should be above pollConsumer in this class.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,444 @@
...
+      for (Map.Entry<SystemStreamPartition, List<IncomingMessageEnvelope>> e : response.entrySet()) {
+        List<IncomingMessageEnvelope> envelopes = e.getValue();
+        if (envelopes != null) {
+          moveMessagesToTheirQueue(e.getKey(), envelopes);
+        }
+      }
+
+      populateCurrentLags(sspsToFetch); // find current lags for for each SSP
+    } else { // nothing to read
+
```
prateekm: Minor: can remove extra newline.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,444 @@
...
+      LOG.debug("No topic/partitions need to be fetched for consumer {} right now. Sleeping {}ms.", kafkaConsumer,
```
prateekm: `consumer {}` should be systemName instead.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,444 @@
...
+      if (lag != null && offset != null && lag >= 0) {
+        long streamEndOffset = offset.longValue() + lag.longValue();
+        // update the metrics
+        kafkaConsumerMetrics.setHighWatermarkValue(tp, streamEndOffset);
+        kafkaConsumerMetrics.setLagValue(tp, lag.longValue());
+      }
+    }
+  }
+
+  boolean isRunning() {
```
prateekm: protected/package private methods should be above private methods, right after public methods.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
prateekm: I'd propose the following method order for readability:
```
package private Methods:
addTopicPartition
start
isRunning
stop
getFailureCause

private methods:
createProxyThreadRunnable
initializeLags
fetchMessages
pollConsumer
processResults
moveMessagesToTheirQueue
populateCurrentLags (inline populateMetricNames)
updateMetrics (inline getLatestLag)
getRecordSize
```

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,444 @@
...
+      LOG.warn("Join in KafkaConsumerProxy has failed", e);
+      consumerPollThread.interrupt();
+    }
+  }
+
+  /**
+   * Add new partition to the list of polled partitions.
+   * This method should be called only at the beginning, before the thread is started.
+   */
+  public void addTopicPartition(SystemStreamPartition ssp, long nextOffset) {
```
prateekm: I think these methods can all be package private (I mean remove `public`, not add the /* package private */ annotation) since the only user should be KafkaSystemConsumer

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,444 @@
...
+      // we need to wait until the thread starts
+      while (!isRunning && failureCause == null) {
+        try {
+          consumerPollThreadStartLatch.await(3000, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+          LOG.info("Got InterruptedException", e);
+        }
+      }
+    } else {
+      LOG.warn("Tried to start an already started KafkaConsumerProxy (%s). Ignoring.", this.toString());
```
prateekm: Prefer logging systemName explicitly instead of this. In this class this can refer to the thread or this class depending on where it's used, so would prefer to not use it in logs to avoid confusion.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+  private final AtomicBoolean started = new AtomicBoolean(false);
+  private final Config config;
+  private final boolean fetchThresholdBytesEnabled;
+  private final KafkaSystemConsumerMetrics metrics;
+
+  // This sink is used to transfer the messages from the proxy/consumer to the BlockingEnvelopeMap.
+  final KafkaConsumerMessageSink messageSink;
+
+  // This proxy contains a separate thread, which reads kafka messages (with consumer.poll()) and populates
+  // BlockeingEnvelopMap's buffers.
```
prateekm: Typo: "BlockingEnvelopeMap"

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+  final private KafkaConsumerProxy proxy;
+
+  // keep registration data until the start - mapping between registered SSPs and topicPartitions, and the offsets
```
prateekm: "and their offsets"

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+  final Map<TopicPartition, String> topicPartitionsToOffset = new HashMap<>();
+  final Map<TopicPartition, SystemStreamPartition> topicPartitionsToSSP = new HashMap<>();
+
+  long perPartitionFetchThreshold;
+  long perPartitionFetchThresholdBytes;
+
+  /**
+   * @param systemName system name for which we create the consumer
```
prateekm: Include description for javadocs. Maybe: "Create a KafkaSystemConsumer for the provided {@code systemName}"

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+   * @param config config passed into the the app
```
prateekm: "application config"

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+   * @param metrics metrics collecting object
```
prateekm: "metrics for this KafkaSystemConsumer"

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+   * @param clock - system clock, allows to override internal clock (System.currentTimeMillis())
```
prateekm: No '-' in param docs. ". Should be sufficient to say "clock system clock"

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+
+    // create a sink for passing the messages between the proxy and the consumer
+    messageSink = new KafkaConsumerMessageSink();
+
+    // Create the proxy to do the actual message reading.
+    String metricName = String.format("%s %s", systemName, clientId);
+    proxy = new KafkaConsumerProxy(kafkaConsumer, systemName, clientId, messageSink, metrics, metricName);
+    LOG.info("{}: Created KafkaConsumerProxy {} ", this, proxy);
+
+    LOG.info("{}: Created KafkaSystemConsumer {}", this, kafkaConsumer);
```
prateekm: This shouldn't be here since the consumer is being passed in.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+    this.config = config;
+    this.metrics = metrics;
+
+    fetchThresholdBytesEnabled = new KafkaConfig(config).isConsumerFetchThresholdBytesEnabled(systemName);
+
+    // create a sink for passing the messages between the proxy and the consumer
+    messageSink = new KafkaConsumerMessageSink();
+
+    // Create the proxy to do the actual message reading.
+    String metricName = String.format("%s %s", systemName, clientId);
```
prateekm: Is the space separator in metricName intentional? What is this used for? If this is prefix for metrics, do we need clientId in it - it's just the jobName + jobId.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+    proxy = new KafkaConsumerProxy(kafkaConsumer, systemName, clientId, messageSink, metrics, metricName);
+    LOG.info("{}: Created KafkaConsumerProxy {} ", this, proxy);
```
prateekm: Would prefer to log everything in this class as "Created KafkaConsumerProxy {} for systemName {}" for consistency with rest of the codebase.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+   * @param clientId client id to use int the kafka client
+   * @param config config
+   * @return kafka consumer object
+   */
+  public static KafkaConsumer<byte[], byte[]> getKafkaConsumerImpl(String systemName, String clientId, Config config) {
+
+    // extract kafka client configs
+    KafkaConsumerConfig consumerConfig = KafkaConsumerConfig.getKafkaSystemConsumerConfig(config, systemName, clientId);
+
+    LOG.info("{}:{} KafkaClient properties {}", systemName, clientId, consumerConfig);
```
prateekm: Don't need clientId. Also see comment about log message format above.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+    // initialize the subscriptions for all the registered TopicPartitions
+    startSubscription();
+    // needs to be called after all the registrations are completed
+    setFetchThresholds();
+
+    startConsumer();
+    LOG.info("{}: Consumer started", this);
+  }
+
+  private void startSubscription() {
```
prateekm: Can this be done in KafkaConsumerProxy#start instead? Would be better to consolidate interactions with the KafkaConsumer in one class.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+
+  @Override
+  public void start() {
+    if (!started.compareAndSet(false, true)) {
+      LOG.warn("{}: Attempting to start the consumer for the second (or more) time.", this);
+      return;
+    }
+    if (stopped.get()) {
+      LOG.warn("{}: Attempting to start a stopped consumer", this);
+      return;
```
prateekm: This should this be an exception, or at least an error log.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+  private void startSubscription() {
+    //subscribe to all the registered TopicPartitions
+    LOG.info("{}: Consumer subscribes to {}", this, topicPartitionsToSSP.keySet());
+    try {
+      synchronized (kafkaConsumer) {
+        // we are using assign (and not subscribe), so we need to specify both topic and partition
+        kafkaConsumer.assign(topicPartitionsToSSP.keySet());
+      }
+    } catch (Exception e) {
+      LOG.warn("{}: Start subscription failed", this);
```
prateekm: Should be error.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+      throw new SamzaException(e);
```
prateekm: Include context in exception message.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+
+  /**
+   * Set the offsets to start from.
+   * Register the TopicPartitions with the proxy.
+   * Start the proxy.
+   */
+  void startConsumer() {
+    // set the offset for each TopicPartition
+    if (topicPartitionsToOffset.size() <= 0) {
+      LOG.warn("{}: Consumer is not subscribed to any SSPs", this);
```
prateekm: Should this be an error?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    topicPartitionsToOffset.forEach((tp, startingOffsetString) -> {
+      long startingOffset = Long.valueOf(startingOffsetString);
+
+      try {
+        synchronized (kafkaConsumer) {
+          kafkaConsumer.seek(tp, startingOffset); // this value should already be the 'upcoming' value
```
prateekm: Can this entire method be in KafkaConsumerProxy to consolidate consumer management in one class (as the name KafkaConsumerProxy implies).

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+      fetchThresholdBytes = Long.valueOf(fetchThresholdBytesOption.get());
+      LOG.info("{}: fetchThresholdBytesOption is configured. fetchThresholdBytes={}", this, fetchThresholdBytes);
+    }
+
+    int numTPs = topicPartitionsToSSP.size();
+    if (numTPs != topicPartitionsToOffset.size()) {
+      throw new SamzaException("topicPartitionsToSSP.size() doesn't match topicPartitionsToOffset.size()");
+    }
+
+    LOG.info("{}: fetchThresholdBytes = {}; fetchThreshold={}; partitions num={}", this, fetchThresholdBytes,
```
prateekm: "numPartitions"

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+        fetchThreshold, numTPs);
+
+    if (numTPs > 0) {
+      perPartitionFetchThreshold = fetchThreshold / numTPs;
+      LOG.info("{}: perPartitionFetchThreshold={}", this, perPartitionFetchThreshold);
```
prateekm: Would prefer to combine these log lines in one message at the end of this method if possible. Currently this adds log noise every time a new consumer is created, which is often.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+        LOG.info("{}: Closing kafkaSystemConsumer {}", this, kafkaConsumer);
+        kafkaConsumer.close();
+      }
+    } catch (Exception e) {
+      LOG.warn("{}: Failed to stop KafkaSystemConsumer.", this, e);
+    }
+  }
+
+  /*
+   record the ssp and the offset. Do not submit it to the consumer yet.
```
prateekm: Use javadoc format.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+    return Long.valueOf(offset1).compareTo(Long.valueOf(offset2));
+  }
+
+  @Override
+  public String toString() {
+    return String.format("%s:%s", systemName, clientId);
+  }
+
+  @Override
+  public Map<SystemStreamPartition, List<IncomingMessageEnvelope>> poll(
```
prateekm: Same comment as other class: prefer ordering methods logically by how they're used. E.g., it makes more sense to order them as 'register, start, poll, stop' etc than the current order of 'start, stop, register, poll'. Also, prefer ordering methods by visibility: public, protected then private methods.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,193 @@
...
+import org.apache.samza.SamzaException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scala.Option;
+
+
+/**
+ * The configuration class for KafkaConsumer
+ */
+public class KafkaConsumerConfig extends HashMap<String, Object> {
```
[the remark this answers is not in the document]
sborya: this class is used to initialize kafka's KafkaConsumer, so which expects Map<String, Object).
We cannot use kafka's ConsumerConfig because it is packaged private for kafka packages.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,193 @@
...
+import org.apache.kafka.clients.consumer.RangeAssignor;
+import org.apache.kafka.common.serialization.ByteArrayDeserializer;
+import org.apache.samza.SamzaException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scala.Option;
+
+
+/**
+ * The configuration class for KafkaConsumer
```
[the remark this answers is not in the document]
sborya: As mentioned above,  since we cannot use ConsumerConfig directly, we need a class to pass to KafkaConsumer.

vjagadish1989:
@sborya : instead of deleting the "old" Kafka consumer's source-code, it would be good to re-name them under a new package if users require the ability to rollback

vjagadish1989:
Additionally, @prateekm and @nickpan47's feedback was to document the list of supported broker versions with this release.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,193 @@
...
+
+    return new KafkaConsumerConfig(consumerProps);
+  }
+
+  // group id should be unique per job
+  static String getConsumerGroupId(Config config) {
+    JobConfig jobConfig = new JobConfig(config);
+    Option<String> jobIdOption = jobConfig.getJobId();
+    Option<String> jobNameOption = jobConfig.getName();
+    return (jobNameOption.isDefined() ? jobNameOption.get() : "undefined_job_name") + "-" + (jobIdOption.isDefined()
```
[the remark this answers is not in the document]
sborya: In general I agree, but this matches current functionality. Not sure if it safe to change this behavior.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConsumerConfig.java
```diff
@@ -0,0 +1,193 @@
...
+   * If settings for auto.reset in samza are different from settings in Kafka (auto.offset.reset),
+   * then need to convert them (see kafka.apache.org/documentation):
+   * "largest" -> "latest"
+   * "smallest" -> "earliest"
+   *
+   * If no setting specified we return "latest" (same as Kafka).
+   * @param autoOffsetReset value from the app provided config
+   * @return String representing the config value for "auto.offset.reset" property
+   */
+  static String getAutoOffsetResetValue(final String autoOffsetReset) {
```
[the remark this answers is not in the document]
sborya: These are the correct 'samza' values. We cannot remove them if we want to be backward compatible (https://samza.apache.org/learn/documentation/0.14/jobs/configuration-table.html)

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -0,0 +1,444 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return String.format("consumerProxy-%s-%s", systemName, clientId);
+  }
+
+  public void start() {
+    if (!consumerPollThread.isAlive()) {
+      LOG.info("Starting KafkaConsumerProxy polling thread for system " + systemName + " " + this.toString());
```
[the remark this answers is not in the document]
sborya: this is the only differentiator between different proxies.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
sborya: I will remove client id here.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
sborya: will put this change for later.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
sborya: removed.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
sborya: made it error.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
sborya: May be. I don't think we should do it in this patch. I will add a jira for it.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemConsumer.java
sborya: will do in later patch.

PR #626, opened 2018-09-07
SAMZA-1858: Public APIs for shared context

cameronlee314:
This is just the API classes and their implementations, in an attempt to keep individual PRs more manageable. An upcoming PR will actually integrate them and remove the classes that these are intended to replace.
SEP: https://cwiki.apache.org/confluence/display/SAMZA/SEP-15%3A+New+Runtime+Context+API


cameronlee314:
@prateekm @nickpan47 @vjagadish1989 Could you please take a look?

samza-api/src/main/java/org/apache/samza/context/ApplicationDefinedContainerContext.java
```diff
@@ -0,0 +1,22 @@
+package org.apache.samza.context;
```
prateekm: Minor: You'll need the license header in the new files.

samza-api/src/main/java/org/apache/samza/context/ContainerContext.java
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * @return processor id for this container
+   */
+  String getProcessorId();
+
+  /**
+   * Returns {@link TaskName}s for all tasks running on this container. There is one {@link TaskName} for each task.
+   * @return {@link TaskName}s for all tasks running on this container
+   */
+  Collection<TaskName> getTaskNames();
```
prateekm: Should this be here in the new API? I think this (and more) could be obtained using JobModel with this processor's ID as the key.

samza-api/src/main/java/org/apache/samza/context/Context.java
```diff
@@ -0,0 +1,42 @@
+package org.apache.samza.context;
+
+/**
+ * Container object for all context provided to instantiate an application at runtime.
+ */
+public interface Context {
+  /**
+   * @return framework-provided context for the overall job that is being run
```
prateekm: Minor: Need description for all javadocs.

samza-api/src/main/java/org/apache/samza/context/Context.java
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Returns the application-defined container context object specified by the
+   * {@link ApplicationDefinedContainerContextFactory}. This is shared across all tasks in the container, but not across
+   * containers.
+   * @param clazz concrete type of the {@link ApplicationDefinedContainerContext} to return
+   * @return application-defined container context
+   * @throws IllegalStateException if no factory was provided
+   * @throws ClassCastException if the type of the concrete context does not match clazz
+   */
+  <T extends ApplicationDefinedContainerContext> T getApplicationDefinedContainerContext(Class<T> clazz);
```
prateekm: Is the class parameter just for knowing the concrete type? If so, I'm not sure this is better then requiring them to cast it themselves.

samza-api/src/main/java/org/apache/samza/context/SamzaContainerContext.java
```diff
@@ -0,0 +1,17 @@
+package org.apache.samza.context;
+
+/**
+ * Container object for job and container context provided by the Samza framework to instantiate an application at
+ * runtime.
+ */
+public interface SamzaContainerContext {
```
prateekm: Why do we need this wrapper context? Looks like `Context` already provides JobContext and ContainerContext.

samza-api/src/main/java/org/apache/samza/context/SamzaContext.java
```diff
@@ -0,0 +1,21 @@
+package org.apache.samza.context;
+
+/**
+ * Container object for all context provided by the Samza framework to instantiate an application at runtime.
+ */
+public interface SamzaContext {
```
prateekm: How is this different than `Context`?

samza-api/src/main/java/org/apache/samza/context/TaskContext.java
```diff
@@ -0,0 +1,70 @@
...
+   * @return {@link TaskName} for this task
+   */
+  TaskName getTaskName();
+
+  /**
+   * Returns all of the input {@link SystemStreamPartition}s for this task. This does not include side input
+   * {@link SystemStreamPartition}s.
+   * @return all of the input {@link SystemStreamPartition}s for this task
+   */
+  Set<SystemStreamPartition> getSystemStreamPartitions();
```
prateekm: Do we want this in the new API? Should be able to get this from the JobModel.

samza-api/src/main/java/org/apache/samza/context/JobContext.java
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Contains information at job granularity, provided by the Samza framework, to be used to instantiate an application at
+ * runtime.
+ */
+public interface JobContext {
+  /**
+   * @return final configuration for this job
+   */
+  Config getConfig();
```
prateekm: Maybe expose JobName, JobId as top level params for consistency with other contexts.

Also, don't we want to expose JobModel here?

samza-api/src/main/java/org/apache/samza/scheduling/Scheduler.java
```diff
@@ -0,0 +1,28 @@
+package org.apache.samza.scheduling;
+
+import org.apache.samza.task.TimerCallback;
+
+
+/**
+ * Provides a way for applications to register some logic to be executed at a future time.
+ */
+public interface Scheduler {
```
prateekm: I'd prefer to just use Timer/TimerRegistry for this. It'll be confusing to explain three separate time-based-callback concepts - Scheduler, Timer, Window.

samza-api/src/main/java/org/apache/samza/context/ContainerContext.java
prateekm: Looking at the models, there are a couple of things in JobModel that don't make sense in the public API. One is that it contains a LocalityManager, and the other is the maxChangeLogStreamPartitions field. These are known issues and we'll eventually clean this up. We can consider making them public APIs at that point.

I think it's fine to expose this information in the appropriate context too. If we do that, I think for this particular case it'll be better to expose a ContainerModel equivalent (i.e., Map<TaskName, List<SSPs for TaskName>>) rather than just the task names.

samza-api/src/main/java/org/apache/samza/context/Context.java
prateekm: Slightly prefer casting for simplicity and consistency with getTable and getStore APIs.

samza-api/src/main/java/org/apache/samza/context/SamzaContainerContext.java
prateekm: Do you mean we need to pass this (and SamzaContext) around to users, or internally within the framework? Also, do we need both this and SamzaContext, or could we do with just one? Just one with both container and task level context might not be that bad, esp. if it's only used to create application defined context.

PS: Let's create a SEP that documents the class hierarchy. Will be useful for reference in this PR and for public documentation.

samza-api/src/main/java/org/apache/samza/context/TaskContext.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.samza.table.Table;
+
+
+/**
+ * Contains information at task granularity, provided by the Samza framework, to be used to instantiate an application
+ * at runtime.
+ * <p>
+ * Note that application-defined task-level context is accessible through {@link ApplicationDefinedTaskContext}.
+ */
+public interface TaskContext {
```
shanthoosh: Just curious.  This interface already exists in samza : `org.apache.samza.task.TaskContext`. Are we going to retain both the interfaces or merge both of them into one.

samza-api/src/main/java/org/apache/samza/context/ApplicationDefinedTaskContextFactory.java
```diff
@@ -0,0 +1,20 @@
...
+
+
+/**
+ * An application should implement this if it has a {@link ApplicationDefinedTaskContext} that is needed for
+ * initialization. This will be used to create instance(s) of that {@link ApplicationDefinedTaskContext}.
+ * <p>
+ * This is {@link Serializable} because it is specified in {@link org.apache.samza.application.ApplicationDescriptor}.
+ * @param <T> concrete type of {@link ApplicationDefinedTaskContext} returned by this factory
+ */
+public interface ApplicationDefinedTaskContextFactory<T extends ApplicationDefinedTaskContext> extends Serializable {
```
shanthoosh: Just curious. If the user has a custom `ApplicationDefinedTaskContextFactory` or `ApplicationDefinedContainerContextFactory`, how will he provide that to the ApplicationRunner api? Are we planning to take it as a config or a separate high level method in `ApplciationRuntime`(like `register(AppDefinedTaskFactory`))?

samza-api/src/main/java/org/apache/samza/context/ApplicationDefinedTaskContext.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * An application should implement this to contain any runtime objects required to initialize the processing logic.
+ * <p>
+ * An instance of this will be created for each task.
+ * <p>
+ * If it is necessary to have a separate instance per task, then use {@link ApplicationDefinedContainerContext} instead.
+ * An {@link ApplicationDefinedTaskContextFactory} needs to be implemented to build instances of this context.
+ */
+public interface ApplicationDefinedTaskContext {
```
shanthoosh: Does it makes sense to do the following renames. Rename `SamzaContainerContext` to `PredefinedContainerContext` and `SamzaTaskContext` to `PredefinedTaskContext`. Rename `ApplicationDefinedTaskContext` to  `UserDefinedTaskContext` or just `TaskContext` .

samza-api/src/main/java/org/apache/samza/context/ApplicationDefinedTaskContext.java
```diff
@@ -0,0 +1,21 @@
...
+  /**
+   * Initialization which will run before processing begins.
+   */
+  void start();
+
+  /**
+   * Shutdown logic which will run after processing ends.
+   */
+  void stop();
```
shanthoosh: 1. Do we want to provide the configuration, metricsRegistry objects here to the user implementation or how will they get it?
2. We expect users to setup all the custom objects in the start implementation? Do the start/stop methods have to be idempotent by the high-level API? Would be better to document the contract here.
3. If the users have to use offspring factories internally here, are we going to offer any other alternative  other than static helper methods?

samza-api/src/main/java/org/apache/samza/context/SamzaContext.java
```diff
@@ -0,0 +1,21 @@
...
+  /**
+   * @return framework-provided context for the overall job that is being run
+   */
+  JobContext getJobContext();
+
+  /**
+   * @return framework-provided context for the container that this is in
+   */
+  ContainerContext getContainerContext();
```
shanthoosh: Can you please share the rationale behind having three different wrapper classes : `SamzaContainerContext`, `Context`, `ContainerContext` each one of them housing some of the context objects. In what scenarios or API each one of these individual classes will be used .

samza-api/src/main/java/org/apache/samza/context/ContainerContext.java
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * Returns the {@link MetricsRegistry} for this container. Metrics built using this registry will be associated with
+   * the container.
+   * @return {@link MetricsRegistry} for this container
+   */
+  MetricsRegistry getContainerMetricsRegistry();
```
shanthoosh: Minor: Since we expose processorId in one other API above, for consistency, do you think it's better to rename this to `getProcessorMetricsRegistry()`?

samza-api/src/main/java/org/apache/samza/context/ApplicationDefinedContainerContext.java
```diff
@@ -0,0 +1,22 @@
...
+ * single container.
+ * <p>
+ * If it is necessary to have a separate instance per task, then use {@link ApplicationDefinedTaskContext} instead.
+ * An {@link ApplicationDefinedContainerContextFactory} needs to be implemented to build instances of this context.
+ */
+public interface ApplicationDefinedContainerContext {
+  /**
+   * Initialization which will run before processing begins.
+   */
+  void start();
```
shanthoosh: It will be better to document the requirements of objects that are created by the start method. Do they have to `serializable`?  Are they going to be persisted in some external storage across runs?

samza-api/src/main/java/org/apache/samza/context/ApplicationDefinedContainerContext.java
```diff
@@ -0,0 +1,22 @@
...
+
+  /**
+   * Shutdown logic which will run after processing ends.
+   */
+  void stop();
+}
```
shanthoosh: Would be better to document the behavior when the `stop()` (or `start()`) method throws up?

samza-api/src/main/java/org/apache/samza/context/Context.java
```diff
@@ -0,0 +1,42 @@
+package org.apache.samza.context;
+
+/**
+ * Container object for all context provided to instantiate an application at runtime.
+ */
+public interface Context {
+  /**
+   * @return framework-provided context for the overall job that is being run
+   */
+  JobContext getJobContext();
```
shanthoosh: How will the user get hold of `JobContext` & `Context` objects? Will it be added to `ContextManager` API

samza-api/src/main/java/org/apache/samza/context/ContainerContext.java
cameronlee314: Synced offline. We will move the current ContainerModel and TaskModel to the API layer. We will not expose JobModel for now because it needs to be cleaned up (e.g. remove localityManager), and Samza currently does not expose the full job model now anyways.

samza-api/src/main/java/org/apache/samza/context/JobContext.java
cameronlee314: Added getJobName and getJobId

samza-api/src/main/java/org/apache/samza/context/SamzaContainerContext.java
cameronlee314: Discussed offline. Instead of using SamzaContext and SamzaContainerContext, the individual context objects will be passed around when needed. This simplifies the object hierarchy/wiring. It is unlikely we will need to add a new type of context object, so the signatures should be able to stay the same.

samza-api/src/main/java/org/apache/samza/context/SamzaContext.java
cameronlee314: SamzaContext has been removed (see above comment).

samza-api/src/main/java/org/apache/samza/context/SamzaContext.java
cameronlee314: SamzaContainerContext has been removed.
Context holds all of the context objects (e.g. job/container/task context, app-defined context). ContainerContext is just for framework-provided context at the container level (e.g. container metrics, container id).

samza-api/src/main/java/org/apache/samza/context/TaskContext.java
cameronlee314: Good question. This will be removed in the near future by a different PR. I added a comment to the javadoc to clarify.

samza-api/src/main/java/org/apache/samza/context/TaskContext.java
cameronlee314: Will replace this with TaskModel.

samza-api/src/main/java/org/apache/samza/context/ApplicationDefinedContainerContext.java
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * An application should implement this to contain any runtime objects required to initialize the processing logic.
+ * <p>
+ * A single instance of this will be created in each container, so it will be shared between all of the tasks on a
+ * single container.
+ * <p>
+ * If it is necessary to have a separate instance per task, then use {@link ApplicationDefinedTaskContext} instead.
+ * An {@link ApplicationDefinedContainerContextFactory} needs to be implemented to build instances of this context.
+ */
+public interface ApplicationDefinedContainerContext {
```
shanthoosh: One additional followup question with respect to standalone. 

Are we planning to invoke the `ApplicationDefinedContainerContext` start/stop once for a standalone stream processor or everytime before/after a rebalancing phase(when we kill/create a SamzaContainer as a part of rebalancing phase).

1. Existing `SamzaContainer` setup/teardown is expensive and setting the right coordination timeouts is a known problem for the standalone users. Just wondering if this will compound that problem even further.
2. In standalone, let's say if a user wants to instantiate some heavy-weight, expensive objects only once(using samza context) per java process, then what is the recommended approach after this change?

cameronlee314: The plan was to invoke start/stop whenever we create/kill a SamzaContainer.
You are right that this could possibly impact standalone during a rebalance. In general, I don't think start and stop takes a long time, but it would take some time, and there may be cases where it takes a long time. It is a good point that this would also change the existing behavior of when some objects are created in standalone. The difficulty is that some of the app-defined context does actually need to be recreated (e.g. what if the context depends on the container-partition mapping?). Do you think it is worth it to add additional complexity to differentiate between context that needs to be recreated and context that does not need to be recreated?

samza-api/src/main/java/org/apache/samza/context/TaskContext.java
nickpan47: I am hoping removing the old one is also part of 1.0, since that will involve some changes to the current ContextManager interface in the ApplicationDescriptor API. I assume that ContextManager will be removed or replaced by this PR.

samza-api/src/main/java/org/apache/samza/scheduling/Scheduler.java
```diff
@@ -0,0 +1,28 @@
...
+  /**
+   * Register a keyed callback of {@link TimerCallback} in this task.
+   * The callback will be invoked exclusively with any other operations for this task, e.g. processing, windowing, and
```
nickpan47: nit: be invoked periodically, exclusive to any other operations for this task,

samza-core/src/main/java/org/apache/samza/context/ContextImpl.java
```diff
@@ -0,0 +1,57 @@
...
+   * @param jobContext non-null job context
+   * @param containerContext non-null framework container context
+   * @param taskContext non-null framework task context
+   * @param applicationDefinedContainerContext nullable application-defined container context
+   * @param applicationDefinedTaskContext nullable application-defined task context
+   */
+  public ContextImpl(JobContext jobContext, ContainerContext containerContext, TaskContext taskContext,
+      ApplicationDefinedContainerContext applicationDefinedContainerContext,
+      ApplicationDefinedTaskContext applicationDefinedTaskContext) {
+    this.jobContext = jobContext;
```
nickpan47: nit: since the javadoc states "non-null", can we also add Preconditions.checkNotNull() here?

samza-core/src/main/java/org/apache/samza/context/SamzaContainerContextImpl.java
```diff
@@ -0,0 +1,21 @@
+package org.apache.samza.context;
+
+public class SamzaContainerContextImpl implements SamzaContainerContext {
```
nickpan47: I assume that this will be removed together with SamzaContainerContext?

samza-core/src/main/java/org/apache/samza/context/SamzaContextImpl.java
```diff
@@ -0,0 +1,28 @@
+package org.apache.samza.context;
+
+public class SamzaContextImpl implements SamzaContext {
```
nickpan47: Same here?

samza-core/src/test/java/org/apache/samza/context/TestContextImpl.java
```diff
@@ -0,0 +1,107 @@
+package org.apache.samza.context;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
nickpan47: nit: don't use *

samza-core/src/test/java/org/apache/samza/context/TestTaskContextImpl.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.samza.scheduling.Scheduler;
+import org.apache.samza.storage.kv.KeyValueStore;
+import org.apache.samza.system.SystemStreamPartition;
+import org.apache.samza.table.TableManager;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import static org.junit.Assert.*;
```
nickpan47: Same here. We do not use import *, as a coding convention.

samza-core/src/test/java/org/apache/samza/context/TestTaskContextImpl.java
```diff
@@ -0,0 +1,76 @@
...
+import static org.mockito.Mockito.*;
+
+
+public class TestTaskContextImpl {
+  private static final TaskName TASK_NAME = new TaskName("myTaskName");
+
+  @Mock
+  private Set<SystemStreamPartition> _systemStreamPartitions;
```
nickpan47: Coding convention: in Samza, we don't use '_' prefix for private members.

samza-core/src/test/java/org/apache/samza/scheduling/TestSchedulerImpl.java
```diff
@@ -0,0 +1,54 @@
+package org.apache.samza.scheduling;
+
+import org.apache.samza.task.SystemTimerScheduler;
+import org.apache.samza.task.TimerCallback;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import static org.mockito.Mockito.*;
```
nickpan47: Same here.

samza-core/src/test/java/org/apache/samza/scheduling/TestSchedulerImpl.java
```diff
@@ -0,0 +1,54 @@
...
+
+
+public class TestSchedulerImpl {
+  @Mock
+  private SystemTimerScheduler _systemTimerScheduler;
```
nickpan47: Same here.

samza-api/src/main/java/org/apache/samza/context/ApplicationDefinedContainerContext.java
```diff
@@ -0,0 +1,22 @@
+package org.apache.samza.context;
+
+/**
+ * An application should implement this to contain any runtime objects required to initialize the processing logic.
+ * <p>
+ * A single instance of this will be created in each container, so it will be shared between all of the tasks on a
+ * single container.
```
cameronlee314: Add a comment here that the context will be recreated if the container model changes (e.g. during rebalancing) (suggestion by @shanthoosh).

samza-api/src/main/java/org/apache/samza/context/ApplicationDefinedContainerContext.java
cameronlee314: Had some discussions in person about this:
1) Since YARN deployments still do not use StreamProcessor, it will be difficult to achieve consistency between YARN and standalone with a processor-level context.
2) The concrete use cases we came up with for application-defined processor context can currently go in the container-level context (it seems ok to recreate them).
3) Use cases like job coordinator lifecycle are not used by applications and they do not depend on application code, so it is currently unnecessary to add something to the public API for that.
4) It is still possible to extend this API to have a processor-level context in the future.

samza-api/src/main/java/org/apache/samza/context/ContainerContext.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Collection;
+import org.apache.samza.container.TaskName;
+import org.apache.samza.metrics.MetricsRegistry;
+
+
+/**
+ * Contains information at container granularity, provided by the Samza framework, to be used to instantiate an
+ * application at runtime.
+ * <p>
+ * Note that application-defined container-level context is accessible through
```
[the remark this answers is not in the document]
cameronlee314: Good suggestion. Also added a similar note about the task context.

samza-api/src/main/java/org/apache/samza/scheduling/Scheduler.java
cameronlee314: A single call to scheduleCallback will only schedule a callback to run at one future time and not multiple times, so in my opinion, we don't want to use the term "periodically" here. Would that be ok to leave it out?

samza-core/src/main/java/org/apache/samza/context/SamzaContainerContextImpl.java
cameronlee314: Yep

samza-core/src/main/java/org/apache/samza/context/SamzaContextImpl.java
cameronlee314: Yep, will be removed.

samza-core/src/test/java/org/apache/samza/context/TestContextImpl.java
cameronlee314: I agree with not using * and static imports for production code, but for these test static imports (e.g. Assert, Mockito) that are used widely in unit tests to reduce verbosity, how do you feel about allowing it? IDEs automatically consolidate static imports to use *, so it is inconvenient to remember to re-expand them for all tests (although I guess there probably is some IDE setting to avoid using *). Also, there are already a lot of other tests in Samza which use * for Assert/Mockito.

samza-core/src/test/java/org/apache/samza/context/TestTaskContextImpl.java
cameronlee314: Ah yes, I think I had caught that, but hadn't updated the PR yet. Thanks for pointing it out anyways.

cameronlee314:
@nickpan47 Yes, there will be a separate PR to update the ApplicationDescriptor APIs.

samza-core/src/test/java/org/apache/samza/context/TestContextImpl.java
cameronlee314: Spoke in person. It should be possible to configure the IDE to not consolidate to a wildcard, and it would be good to stick to the non-wildcard pattern for any new tests.
I'll update these tests to not use the wildcard.

samza-api/src/main/java/org/apache/samza/context/ApplicationContainerContextFactory.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public interface ApplicationContainerContextFactory<T extends ApplicationContainerContext> extends Serializable {
+  /**
+   * Create an instance of the application-defined {@link ApplicationContainerContext}.
+   *
+   * @param jobContext framework-provided job context used for building {@link ApplicationContainerContext}
+   * @param containerContext framework-provided container context used for building {@link ApplicationContainerContext}
+   * @return new instance of the application-defined {@link ApplicationContainerContext}
+   */
+  T create(JobContext jobContext, ContainerContext containerContext);
```
nickpan47: Is the idea to pass in all available / existing context objects that ApplicationContainerContext may depend on in the input parameters list? Would that be harder to extend if we need to add a ProcessorContext later?

samza-api/src/main/java/org/apache/samza/context/ApplicationTaskContextFactory.java
```diff
@@ -0,0 +1,44 @@
...
+   * Create an instance of the application-defined {@link ApplicationTaskContext}.
+   *
+   * @param jobContext framework-provided job context used for building {@link ApplicationTaskContext}
+   * @param containerContext framework-provided container context used for building {@link ApplicationTaskContext}
+   * @param taskContext framework-provided task context used for building {@link ApplicationTaskContext}
+   * @param applicationContainerContext application-provided container context used for building
+   * {@link ApplicationTaskContext}
+   * @return new instance of the application-defined {@link ApplicationContainerContext}
+   */
+  T create(JobContext jobContext, ContainerContext containerContext, TaskContext taskContext,
```
nickpan47: Same comment here. It seems that it would be nicer in terms of encapsulation if we have Context that provides access to the full set of Context objects, and ContainerContext that provides a subset of Context objects that ApplicationContainerContext will be dependent on, and TaskContext that provides a subset of Context objects that ApplicationTaskContext will be dependent on. I know that it seems to be a backward comments, but seems that grouping them together would make it easier for extension of dependent context needed for the create() method signature in the context factories.

samza-api/src/main/java/org/apache/samza/context/ApplicationContainerContextFactory.java
cameronlee314: Talked in-person about this:
It would be harder to extend, but the alternative of having a wrapper object which contains these arguments complicates the public API.
With this proposed API, the strategy for extending would be to add a new method which includes a new object and to deprecate the old method.

samza-api/src/main/java/org/apache/samza/context/ApplicationTaskContextFactory.java
cameronlee314: See above for conclusion about this.

PR #627, opened 2018-09-07
SAMZA-1841: Added support for marking an input stream as broadcast using InputDescriptors

prateekm:


prateekm:
@nickpan47 @vjagadish1989

samza-api/src/main/java/org/apache/samza/operators/descriptors/base/stream/InputDescriptor.java
```diff
@@ -166,22 +183,31 @@ public SubClass withDeleteCommittedMessages(boolean deleteCommittedMessages) {
...
+        generatedConfig.put(String.format(BOOTSTRAP_CONFIG_KEY, streamId), Boolean.toString(bootstrap)));
     this.isBoundedOptional.ifPresent(bounded ->
-        configs.put(String.format(BOUNDED_CONFIG_KEY, streamId), Boolean.toString(bounded)));
+        generatedConfig.put(String.format(BOUNDED_CONFIG_KEY, streamId), Boolean.toString(bounded)));
     this.deleteCommittedMessagesOptional.ifPresent(deleteCommittedMessages ->
-        configs.put(String.format(DELETE_COMMITTED_MESSAGES_CONFIG_KEY, streamId),
+        generatedConfig.put(String.format(DELETE_COMMITTED_MESSAGES_CONFIG_KEY, streamId),
             Boolean.toString(deleteCommittedMessages)));
-    return Collections.unmodifiableMap(configs);
+    this.broadcastPartitionsOptional.ifPresent(broadcastPartitions -> {
```
cameronlee314: It seems like it might get confusing for toConfig to return a "global" value, and it would be simpler if toConfig was independent of existing configs.
Would it be possible to manage this through the graph and then generate the full "task.broadcast.inputs" like how "task.inputs" gets generated (I think it happens in JobNode)? I'm not really sure what it would look like, when initially seeing this, it seemed like "task.broadcast.inputs" should be treated more similarly to "task.inputs".

prateekm: Agree that this pattern is confusing and error prone. However there's currently no single class and pattern for all config generation. It's done by a combination ExectionPlanner, ApplicationDescriptorImpl and JobNode, StreamDescriptors and TableSpec. We do want to consolidate and simplify this logic. 

We can either check this in now and fix the internal APIs as part of one of SAMZA-1811, 1814 or 1842, or I can try to pass this around through the ApplicationDescriptorImpl. I'd prefer the former, since I'm not sure the latter is going to be the right way after the eventual refactor. WDYT?

samza-api/src/main/java/org/apache/samza/operators/descriptors/base/stream/InputDescriptor.java
```diff
@@ -149,6 +155,17 @@ public SubClass withBounded(boolean isBounded) {
     return (SubClass) this;
   }
 
+  /**
+   * Specifies partitions of this stream that all tasks should consume.
+   *
+   * @param broadcastPartitions partitionId or [startingPartitionId-endingPartitionId]. E.g., "0" or "[0-2]"
+   * @return this input descriptor
+   */
+  public SubClass withBroadcastPartitions(String broadcastPartitions) {
```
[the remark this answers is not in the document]
prateekm: I'm not sure using one of the existing Range classes (apache commons, guava) would be better, since their API is a lot more general. It'll probably be better to write our own simple implementation. Either way, the final validation still happens at runtime (in JavaTaskConfig), since we can't move validation logic from Config classes to Descriptors. Since there's not much validation benefit, and it's fairly clear from documentation what the expected value is, it feels slightly unnecessary to add a new class.

I don't feel strongly about this. Do you prefer one way over the other?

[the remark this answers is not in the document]
cameronlee314: I prefer having an explicit class. I agree that it would be better to write our own simple implementation of the class.
An explicit class is a bit more self-documenting and it better prevents accidental mistakes. It also moves validation from run-time to compile-time.
(Less importantly) If we ever want to extend this functionality or remove the config for this in the future, it might be easier to manage with a class.

samza-api/src/main/java/org/apache/samza/operators/descriptors/base/stream/InputDescriptor.java
cameronlee314: Would it be feasible to pass it through the StreamDescriptor through StreamGraphSpec until it reaches JobNode, and then create the config there (kind of like how "task.inputs" is generated)?
Do you think having the currentConfig argument in the toConfig signature is a necessary long-term pattern? To me, it seems like it might not be necessary. If the toConfig signature makes it into a release with the currentConfig argument, and then it will be backwards incompatible to remove it (since it seems to be part of the public API for stream descriptor implementors).

samza-api/src/main/java/org/apache/samza/operators/descriptors/base/stream/InputDescriptor.java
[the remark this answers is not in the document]
prateekm: FWIW, it still won't move all validation to compile time. E.g., we can't assert at compile time that lower bound is indeed lower than the upper bound, that none of the bounds are negative, etc. The only thing it will help with is preventing formatting mistakes for the range syntax. 

We could achieve that by changing the method signature to `withBootstrapPartitions(int startIndex, int endIndex)` and document that startIndex <= endIndex. Would that be sufficient?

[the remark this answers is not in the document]
cameronlee314: Ah, you are right about that validation. I was thinking about validation for the formatting cases.
For your proposed method, you meant "withBroadcastPartitions", right?
Do you think there will be some other way in the future to specify the broadcast partitions other than with a start/end? For example, one possible pain point is that if an app wants to broadcast all partitions, then the number of partitions needs to be known, and if the number of partitions for a stream changes, then the code needs to change. If we could provide a way to say "all partitions" regardless of partition count, then it might be nice. Another example (less likely) is if someone was a non-continuous list of partitions (e.g. 0,2,4). If you think we might want to add something like these in the future, then it might be better to have a class/interface.

samza-api/src/main/java/org/apache/samza/operators/descriptors/base/stream/InputDescriptor.java
prateekm: These need to be generated in ExecutionPlanner, before the JobNode. I'll pass it around through StreamGraphSpec/ApplicationDescriptorImpl for now.

nickpan47: I agree with Cameron on this point to move the generation of task.broadcast.inputs to JobNode, similar to task.inputs. We do have a fragmentation of configuration generation code across multiple stages of the planner: JobPlanner creating the system and stream configuration (and potential override of system and streams) to create the StreamManager, and then the ExecutionPlanner. The ExecutionPlanner creates JobGraph assuming the system stream configuration are already in the configure and use it to generate StreamSpec to create StreamEdge and JobNode. Then, finally JobNode generates the configuration for tables, stores, sideInputs, and serdes. It would be much cleaner if we can just use system, stream, store, and table descriptors in the planners and only use the descriptors to generate configuration and apply overrides in JobNode. This would require the following refactoring:
1. make sure that we don't rely on configuration to create StreamManager, ExecutionPlanner, JobGraph, StreamEdge, and JobNode
2. configuration is only generated from descriptors in JobNodeConfigureGenerator (in my PR, I separated the logic to generate configuration for JobNode to JobNodeConfigureGenerator).

In this PR, I would still prefer to move the generation of task.broadcast.inputs to JobNode. I don't see a need to use the configuration task.broadcast.inputs in ExecutionPlanner and JobGraph.

samza-core/src/main/java/org/apache/samza/application/StreamApplicationDescriptorImpl.java
```diff
@@ -112,7 +112,6 @@ public StreamApplicationDescriptor withDefaultSystem(SystemDescriptor<?> default
       return expander.get().apply(this, inputDescriptor);
     }
 
-    // TODO: SAMZA-1841: need to add to the broadcast streams if inputDescriptor is for a broadcast stream
     Preconditions.checkState(!inputDescriptors.containsKey(inputDescriptor.getStreamId()),
```
nickpan47: We need to add this input to broadcastStreams if the inputDescriptor.isBroadcast() is true.

samza-api/src/main/java/org/apache/samza/operators/descriptors/base/stream/InputDescriptor.java
```diff
@@ -52,6 +57,7 @@
   private Optional<Boolean> isBootstrapOptional = Optional.empty();
   private Optional<Boolean> isBoundedOptional = Optional.empty();
   private Optional<Boolean> deleteCommittedMessagesOptional = Optional.empty();
+  private Optional<String> broadcastPartitionsOptional = Optional.empty();
```
nickpan47: It would be nice to have a method isBroadcast() to indicate whether this InputDescriptor is broadcast stream or not.

samza-core/src/main/java/org/apache/samza/application/StreamApplicationDescriptorImpl.java
```diff
@@ -112,7 +112,6 @@ public StreamApplicationDescriptor withDefaultSystem(SystemDescriptor<?> default
       return expander.get().apply(this, inputDescriptor);
```
nickpan47: We probably should throw exception for ExpandingInputDescriptor calling withBroadcastPartitions().

samza-core/src/main/java/org/apache/samza/application/TaskApplicationDescriptorImpl.java
```diff
@@ -62,7 +62,6 @@ public void setTaskFactory(TaskFactory factory) {
 
   @Override
   public void addInputStream(InputDescriptor inputDescriptor) {
-    // TODO: SAMZA-1841: need to add to the broadcast streams if inputDescriptor is for a broadcast stream
```
nickpan47: Same here, we need to add the inputDescriptor to broadcastStreams set if inputDescriptor.isBroadcast() == true

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
```diff
@@ -148,8 +148,6 @@ public JobConfig generateConfig(String executionPlanJson) {
     }
 
     if (!broadcasts.isEmpty()) {
-      // TODO: remove this once we support defining broadcast input stream in high-level
```
nickpan47: I see the reason that you are appending the task.broadcast.inputs in inputDescriptor.toConfigure() now, since this line below assume that the descriptors don't tell us whether the stream is broadcast or not and only info is from the configure. Another way is that you can skip this part and just make changes to InputDescriptor to add the broadcast option. As long as my earlier comment on populating the broadcastStreams set in the ApplicationDescriptorImpl is done, we can get the broadcast streams from the ApplicationDescriptorImpl. In my PR, I have already changed the JobNode to include ApplicationDescriptorImpl as a parameter to the constructor.

samza-core/src/main/java/org/apache/samza/execution/JobPlanner.java
```diff
@@ -169,10 +170,13 @@ private JobConfig prepareTaskJob(TaskApplicationDescriptorImpl taskAppDesc) {
   }
 
   private Map<String, String> expandSystemStreamConfigs(ApplicationDescriptorImpl<? extends ApplicationDescriptor> appDesc) {
+    // merge user-provided configuration with input/output descriptor generated configuration
+    // descriptor generated configuration has higher priority
     Map<String, String> systemStreamConfigs = new HashMap<>();
-    appDesc.getInputDescriptors().forEach((key, value) -> systemStreamConfigs.putAll(value.toConfig()));
-    appDesc.getOutputDescriptors().forEach((key, value) -> systemStreamConfigs.putAll(value.toConfig()));
-    appDesc.getSystemDescriptors().forEach(sd -> systemStreamConfigs.putAll(sd.toConfig()));
+    systemStreamConfigs.put(TaskConfigJava.BROADCAST_INPUT_STREAMS, config.get(TaskConfigJava.BROADCAST_INPUT_STREAMS));
```
nickpan47: One more question: I have exclude the task.inputs from high-level applications (and for low-level TaskApplication), only exception is the LegacyTaskApplication. Shouldn't we do the same for task.broadcast.inputs when we allow InputDescriptor to specify broadcast option?

PR #628, opened 2018-09-07
SAMZA-1845: Require providing serde for default system in ApplicationDescriptor#setDefaultSystem

prateekm:


prateekm:
@cameronlee314 @nickpan47

samza-api/src/main/java/org/apache/samza/application/StreamApplicationDescriptor.java
```diff
@@ -41,16 +42,19 @@
...
+   * A {@code defaultSystemSerde} may be provided for intermediate streams created on the default system. If the
+   * intermediate stream is explicitly created with a stream-level Serde, it will be used. Else, if a
+   * non-null {@code defaultSystemSerde} has been provided, it will be used. Else the key and msg serde specified
+   * in configuration for the {@code job.default.system} will be used.
    * <p>
    * Providing an incompatible message type for the intermediate streams that use the default serde will result in
    * {@link ClassCastException}s at runtime.
    *
    * @param defaultSystemDescriptor the default system descriptor to use
+   * @param defaultSystemSerde the {@link Serde} to use for streams on the default system
```
cameronlee314: The commit description states "Require providing serde", but it looks like it is also ok to pass null here. Just wanted to double check that it is ok to pass null. If it is ok to pass null, could you please note in the param doc that this param is nullable?

samza-test/src/main/java/org/apache/samza/example/KeyValueStoreExample.java
```diff
@@ -68,7 +68,7 @@ public void describe(StreamApplicationDescriptor appDesc) {
         trackingSystem.getOutputDescriptor("pageViewEventPerMember",
             KVSerde.of(new StringSerde(), new JsonSerdeV2<>(StatsOutput.class)));
 
-    appDesc.withDefaultSystem(trackingSystem);
+    appDesc.withDefaultSystem(trackingSystem, null); // serde must be specified explicitly for intermediate streams
```
cameronlee314: When I initially read this comment, it seemed to conflict with the capability that the default serde could be set for intermediate streams. Does this comment refer to the fact that in this example, the serde is already specified explicitly for intermediate streams? If so, since this is an example that people might be looking at, do you think it might be good to clarify this?

samza-test/src/main/java/org/apache/samza/example/RepartitionExample.java
```diff
@@ -64,7 +64,7 @@ public void describe(StreamApplicationDescriptor appDesc) {
         trackingSystem.getOutputDescriptor("pageViewEventPerMember",
             KVSerde.of(new StringSerde(), new JsonSerdeV2<>(MyStreamOutput.class)));
 
-    appDesc.withDefaultSystem(trackingSystem);
+    appDesc.withDefaultSystem(trackingSystem, null); // serde must be specified explicitly for intermediate streams
```
cameronlee314: same as above about clarifying the comment

samza-core/src/test/java/org/apache/samza/application/TestStreamApplicationDescriptorImpl.java
```diff
@@ -425,6 +426,30 @@ public void testGetIntermediateStreamWithNoSerde() {
...
+  public void testGetIntermediateStreamWithDefaultSystemSerde() {
+    Config mockConfig = mock(Config.class);
+    String streamId = "streamId";
+    SystemDescriptor mockSystemDescriptor = mock(SystemDescriptor.class);
+    Serde mockKeySerde = mock(Serde.class);
+    Serde mockValueSerde = mock(Serde.class);
+    when(mockSystemDescriptor.getTransformer()).thenReturn(Optional.empty());
+    StreamApplicationDescriptorImpl streamAppDesc = new StreamApplicationDescriptorImpl(appDesc -> {
+
+      appDesc.withDefaultSystem(mockSystemDescriptor, KVSerde.of(mockKeySerde, mockValueSerde));
```
nickpan47: Can we also add a test withDefaultSystem(mockSystemDescriptor, null)?

samza-test/src/main/java/org/apache/samza/example/KeyValueStoreExample.java
nickpan47: Alternatively, we can just say there is no default serde set for all intermediate streams.

prateekm:
Thanks for the review @nickpan47 and @cameronlee314. I'm discarding these changes for now. I think a better direction would be to removed the no-serde variants of operators that create intermediate streams. Will submit another PR.

PR #629, opened 2018-09-07
SAMZA-1859: Zookeeper implementation of MetadataStore.

shanthoosh:


shanthoosh:
@xinyuiscool @sborya @vjagadish1989 

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -0,0 +1,117 @@
...
+  public ZkMetadataStore(Config config, MetricsRegistry metricsRegistry, String zkBaseDir) {
+    this.zkConfig = new ZkConfig(config);
+    this.zkClient = new ZkClient(zkConfig.getZkConnect(), zkConfig.getZkSessionTimeoutMs(), zkConfig.getZkConnectionTimeoutMs(), new BytesPushThroughSerializer());
+    this.zkBaseDir = zkBaseDir;
+    createIfNotExists(zkBaseDir);
+  }
+
+  @Override
+  public void init() {
+    zkClient.waitUntilConnected(zkConfig.getZkConnectionTimeoutMs(), TimeUnit.MILLISECONDS);
```
dnishimura: Would `zkClient.retryUntilConnected` be better? Looks like `zkClient.readData/writeData/delete` have retries under the hood, but `zkClient.waitUntilConnected` does not.

samza-core/src/main/java/org/apache/samza/config/ZkConfig.java
```diff
@@ -25,6 +25,7 @@
   public static final String ZK_SESSION_TIMEOUT_MS = "job.coordinator.zk.session.timeout.ms";
   public static final String ZK_CONNECTION_TIMEOUT_MS = "job.coordinator.zk.connection.timeout.ms";
   public static final String ZK_CONSENSUS_TIMEOUT_MS = "job.coordinator.zk.consensus.timeout.ms";
+  public static final String CONFIG_COORDINATOR_ZK_BASE_PATH = "job.coordinator.zk.base.path";
```
prateekm: Make naming convention consistent (no CONFIG_ prefix)

samza-core/src/main/java/org/apache/samza/config/ZkConfig.java
```diff
@@ -52,4 +53,8 @@ public int getZkConnectionTimeoutMs() {
   public int getZkBarrierTimeoutMs() {
     return getInt(ZK_CONSENSUS_TIMEOUT_MS, DEFAULT_CONSENSUS_TIMEOUT_MS);
   }
+
+  public String getCoordinatorBasePath() {
```
prateekm: getZKBasePath?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorFactory.java
```diff
@@ -40,22 +41,25 @@
...
+   * Instantiates an {@link ZkJobCoordinator} using the {@link Config}.
    *
-   * @param config - configs relevant for the JobCoordinator TODO: Separate JC related configs into a "JobCoordinatorConfig"
-   * @return An instance of IJobCoordinator
+   * @param config zookeeper configurations required for instantiating {@link ZkJobCoordinator}
+   * @return An instance of {@link ZkJobCoordinator}
    */
   @Override
   public JobCoordinator getJobCoordinator(Config config) {
+    // TODO: Separate JC related configs into a "ZkJobCoordinatorConfig"
```
prateekm: Let's create a ticket and add it to the comment.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorFactory.java
```diff
@@ -40,22 +41,25 @@
...
     MetricsRegistry metricsRegistry = new MetricsRegistryMap();
-    ZkUtils zkUtils = getZkUtils(config, metricsRegistry);
-    LOG.debug("Creating ZkJobCoordinator instance with config: {}.", config);
-    return new ZkJobCoordinator(config, metricsRegistry, zkUtils);
+    String jobCoordinatorZkBasePath = getJobCoordinationZkPath(config);
+    ZkUtils zkUtils = getZkUtils(config, metricsRegistry, jobCoordinatorZkBasePath);
+    MapConfig mapConfig = new MapConfig(config, ImmutableMap.of(ZkConfig.CONFIG_COORDINATOR_ZK_BASE_PATH, jobCoordinatorZkBasePath));
```
prateekm: I'm assuming this constructor is combining the two maps? If so, didn't we just get the jobCoordinatorZkBasePath from config? Do we need to add it back again?

samza-core/src/main/java/org/apache/samza/zk/ZkKeyBuilder.java
```diff
@@ -96,6 +98,10 @@ public String getJobModelPath(String jobModelVersion) {
   }
 
   public String getJobModelVersionBarrierPrefix() {
-    return String.format("%s/%s/%s/versionBarriers", getRootPath(), JOBMODEL_GENERATION_PATH, JOB_MODEL_UPGRADE_BARRIER);
+    return String.format("%s/%s/%s/versionBarriers", getRootPath(), JOBMODEL_GENERATION_PATH, JOB_MODEL_UPGRADE_BARRIER_PATH);
+  }
+
+  String getTaskLocalityPath() {
```
prateekm: Doesn't need to be public like methods above?

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStoreFactory.java
```diff
@@ -0,0 +1,54 @@
...
+public class ZkMetadataStoreFactory implements MetadataStoreFactory {
+
+  private static final String TASK_LOCALITY_NAMESPACE = "TaskLocalityNamespace";
+
+  @Override
+  public MetadataStore getMetadataStore(String namespace, Config config, MetricsRegistry metricsRegistry) {
+    return new ZkMetadataStore(config, metricsRegistry, getZookeeperNamespace(namespace, config));
+  }
+
+  private static String getZookeeperNamespace(String namespace, Config config) {
```
prateekm: Namespace selection should be done by the caller, not implicitly by the factory.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStoreFactory.java
```diff
@@ -0,0 +1,54 @@
...
+ * Builds the {@link ZkMetadataStore} based upon the provided {@link Config}
+ * and {@link MetricsRegistry}.
+ */
+public class ZkMetadataStoreFactory implements MetadataStoreFactory {
+
+  private static final String TASK_LOCALITY_NAMESPACE = "TaskLocalityNamespace";
+
+  @Override
+  public MetadataStore getMetadataStore(String namespace, Config config, MetricsRegistry metricsRegistry) {
+    return new ZkMetadataStore(config, metricsRegistry, getZookeeperNamespace(namespace, config));
```
prateekm: Minor: Make order of constructor parameters consistent with the factory method params (i.e., namespace, config, metrics)

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStoreFactory.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.samza.metrics.MetricsRegistry;
+import org.apache.samza.SamzaException;
+
+/**
+ * Builds the {@link ZkMetadataStore} based upon the provided {@link Config}
+ * and {@link MetricsRegistry}.
+ */
+public class ZkMetadataStoreFactory implements MetadataStoreFactory {
+
+  private static final String TASK_LOCALITY_NAMESPACE = "TaskLocalityNamespace";
```
dnishimura: Where is this used other than in the unit test? 
https://github.com/apache/samza/blob/029f652cb0b0bb597c469ec8d748a4120294f3dc/samza-core/src/test/java/org/apache/samza/zk/TestZkMetadataStore.java#L39

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @Override
+  public void close() {
+    zkClient.close();
+  }
+
+  private String getZkPathForKey(byte[] key) {
+    try {
+      return String.format("%s/%s", zkBaseDir, new String(key, "UTF-8"));
+    } catch (UnsupportedEncodingException e) {
```
prateekm: Can use the String constructor with explicit Charset instead (Charset.forName("UTF-8") and remove the try-catch.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -0,0 +1,117 @@
...
+      throw new SamzaException(e);
+    }
+  }
+
+  private void createIfNotExists(String path) {
+    if (!zkClient.exists(path)) {
```
prateekm: Does this operation need to be atomic? Is it OK if multiple callers call createPersistent for the same path? If this is idempotent, do we need to check if it already exists?

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @Override
+  public Map<byte[], byte[]> all() {
+    try {
+      List<String> zkSubDirectories = zkClient.getChildren(zkBaseDir);
+      Map<byte[], byte[]> result = new HashMap<>();
+      for (String zkSubDir : zkSubDirectories) {
+        String completeZkPath = String.format("%s/%s", zkBaseDir, zkSubDir);
+        byte[] value = zkClient.readData(completeZkPath, true);
```
prateekm: Does this need to be done atomically for all data within the base dir?

samza-core/src/main/java/org/apache/samza/config/ZkConfig.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorFactory.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorFactory.java
shanthoosh: We construct the basePath(`getJobCoordinationZkPath`) by combining `appId`, `jobId`, `jobName` from the config. By combining the maps we add it to the configuration so that the layers beneath this doesn't have to construct the base path themselves(using the same logic).

samza-core/src/main/java/org/apache/samza/zk/ZkKeyBuilder.java
shanthoosh: No. It can be package private since the zk implementation using it lies within the same package.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStoreFactory.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
shanthoosh: `ZkClient` does not have a multiGet API to read the contents of all the childNodes of a `zkRootNode` atomically. The only plausible way is to getAllChildren, iterate, read and return data for each one of them individually.  

For the use-cases we're planning to use this in samza(LeaderJobCoordinator reading/writing locality & metadata in zookeeper), I think we don't need this to be atomic.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
shanthoosh: Zookeeper call `createPersistent` is not idempotent. If the zookeeper path exists, then this API will throw `ZkNodeExistsException`.  

I will make changes, so that this class is thread-safe(So we don't face this exception when multiple callers update the same key at the same time) .

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStoreFactory.java
shanthoosh: It will be used in the follow up patch where after the stream processors group agree on a JobModel, each one of them will update the taskLocality before launching the `SamzaContainer`.

samza-core/src/main/java/org/apache/samza/config/ZkConfig.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
shanthoosh: `zkClient.retryUntilConnected` API is meant for executing any zookeeper client operations with bounded retries(operation is defined by the `Callable` parameter). I think it might not apply here, since we're just trying to establish a zkSession.

But we have similar semantics here with waitUntilConnected() API, where we explicitly pass in the timeout for the initial zookeeper session establishment. I think this is recommended by the zkClient user-guide(and used elsewhere in samza).

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorFactory.java
prateekm: Not sure I understand what's happening here. Is the "derived" zkBasePath being written back to the config for the ZkJobCoordinator to use? Does that mean that `ZkConfig.CONFIG_COORDINATOR_ZK_BASE_PATH` is a derived config, and not user settable?

samza-core/src/main/java/org/apache/samza/zk/ZkKeyBuilder.java
prateekm: Minor: Does that apply to all the methods above? If so, I'd prefer to be consistent and update them all.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
prateekm: Current use case != all use cases, esp. for general purpose APIs like this. If this is not atomic, we should at least document the guarantee we provide in javadocs. Also, what happens if a path gets deleted while we're iterating? Does that throw an exception? Should document what this method does in that case (e.g., retries until success / ignores path in returned results / throws exception).

If we can detect missing paths, we can get a consistent result set by (bounded) retrying until we get a value for each children.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
prateekm: Thread safety isn't enough here. This could be called in different processes, so we need to handle ZkNodeExistsException. Can we do the following?
`try { createPersistent(path) } catch (ZkNodeExistsException e) { // ignore }`

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStoreFactory.java
prateekm: This doesn't belong in the Store Factory implementation. See comment about namespace selection below. Also, namespace shouldn't be specific to implementation, this should be the same whether for coordinator stream and ZK.

samza-core/src/main/java/org/apache/samza/zk/ZkKeyBuilder.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
shanthoosh: >> Also, what happens if a path gets deleted while we're iterating? Does that throw an exception? 

`ZkClient.readData` returns null if the zookeeper path does not exists(controlled by the second parameter). 

>> Should document what this method does in that case.
Sure, would add documentation about atomicity guarantees and the above scenarios.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStoreFactory.java
shanthoosh: Will change.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorFactory.java
prateekm: What I meant to say was that if it's not exposed to users, should this be a separate parameter instead of an "internal" config key? I'd prefer to not wire internal properties through Config classes. We can discuss offline if I'm misunderstanding something.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -0,0 +1,109 @@
...
+        if (value != null) {
+          result.put(completeZkPath.getBytes("UTF-8"), value);
+        }
+      }
+      return result;
+    } catch (Exception e) {
+      LOG.error("Exception occurred on reading the path: {} from zookeeper.", zkBaseDir, e);
```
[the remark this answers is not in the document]
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -0,0 +1,109 @@
...
+      throw new SamzaException(e);
```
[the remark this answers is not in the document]
shanthoosh: Done.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -978,6 +977,9 @@ class SamzaContainer(
         case unknownException: Throwable =>
           warn("Received an exception when persisting locality info for container %s: " +
             "%s" format (containerContext.id, unknownException.getMessage))
+      } finally {
+        info("Shutting down locality manager.")
+        localityManager.close()
```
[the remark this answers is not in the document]
shanthoosh: > Possible to move LocalityManager construction in this method as well?

Sounds like a good idea, done.

> Also, do you think it'll be useful if containers always reported their locality, regardless of whether host affinity is on or off?

Currently in YARN model, when host affinity is turned off,  non-sticky `ContainerAllocator` is used(Even if we report the host assignment, the `ContainerAllocator` won't use it for container-to-host assignment). In case of standalone, the processors register themselves to the group with their locality information. I think it might be unnecessary to report host assignment of containers by default.

PR #630, opened 2018-09-10
Samza SQL: Code re-org to accomodate Samza SQL engine to take Calcite IR as input

atoomula:
This PR has the following changes:
- Let QueryTranslator take Calcite IR as input
- Include 'INSERT INTO' sql statement for Calcite plan
- Basic DSLConverter Framework with SamzaSQL dialect as an example
- Some fixes to stream-table join wrt Serde

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/dsl/SamzaSqlDslConverter.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/dsl/SamzaSqlDslConverter.java
@@ -1,0 +1,46 @@
+/*
+* Licensed to the Apache Software Foundation (ASF) under one
+* or more contributor license agreements.  See the NOTICE file
+* distributed with this work for additional information
+* regarding copyright ownership.  The ASF licenses this file
+* to you under the Apache License, Version 2.0 (the
+* "License"); you may not use this file except in compliance
+* with the License.  You may obtain a copy of the License at
+*
+*   http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing,
+* software distributed under the License is distributed on an
+* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+* KIND, either express or implied.  See the License for the
+* specific language governing permissions and limitations
+* under the License.
+*/
+
+package org.apache.samza.sql.dsl;
+
+import java.util.Collection;
+import java.util.Collections;
+import org.apache.calcite.rel.RelRoot;
+import org.apache.samza.sql.interfaces.DslConverter;
+import org.apache.samza.sql.planner.QueryPlanner;
+import org.apache.samza.sql.runner.SamzaSqlApplicationConfig;
+
+
+public class SamzaSqlDslConverter implements DslConverter {
+
+  private SamzaSqlApplicationConfig sqlConfig;
+
+  SamzaSqlDslConverter(SamzaSqlApplicationConfig sqlConfig) {
+    this.sqlConfig = sqlConfig;
+  }
+
+  @Override
+  public Collection<RelRoot> convertDsl(String dsl) {
+    QueryPlanner planner =
+        new QueryPlanner(sqlConfig.getRelSchemaProviders(), sqlConfig.getSystemStreamConfigsBySource(),
+            sqlConfig.getUdfMetadata());
+    final RelRoot relRoot = planner.plan(dsl);
+    return Collections.singletonList(relRoot);
+  }
+}
--- a/samza-sql/src/main/java/org/apache/samza/sql/interfaces/SamzaSqlJavaTypeFactoryImpl.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/interfaces/SamzaSqlJavaTypeFactoryImpl.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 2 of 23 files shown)


samza-sql/src/main/java/org/apache/samza/sql/dsl/SamzaSqlDslConverter.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.samza.sql.testutil.SqlFileParser;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.samza.sql.runner.SamzaSqlApplicationConfig.*;
+
+
+public class SamzaSqlDslConverter implements DslConverter {
+
+  private static final Logger LOG = LoggerFactory.getLogger(SamzaSqlApplicationConfig.class);
```
weiqingy: "SamzaSqlApplicationConfig.class" should be "SamzaSqlDslConverter.class" ?

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SamzaSqlJavaTypeFactoryImpl.java
```diff
@@ -0,0 +1,71 @@
...
+  public static RelDataType toSql(final RelDataTypeFactory typeFactory,
+      RelDataType type) {
+    if (type instanceof RelRecordType) {
+      return typeFactory.createStructType(
+          Lists.transform(type.getFieldList(), a0 -> toSql(typeFactory, a0.getType())),
+          type.getFieldNames());
+    }
+    if (type instanceof JavaType) {
+      SqlTypeName typeName = JavaToSqlTypeConversionRules.instance().lookup(((JavaType) type).getJavaClass());
+      // For unknown sql type names, return
```
srinipunuru: Can you add more details on the scenarios where the existing type factory returns null?

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationRunner.java
```diff
@@ -63,25 +69,38 @@ public SamzaSqlApplicationRunner(Boolean localRunner, Config config) {
   public static Config computeSamzaConfigs(Boolean localRunner, Config config) {
     Map<String, String> newConfig = new HashMap<>();
 
+    // TODO: Get the converter factory based on the file type. Create abstraction around this.
+    DslConverterFactory dslConverterFactory = new SamzaSqlDslConverterFactory();
```
srinipunuru: Looks like we are creating the DslConverter in this two places right? SamzaSQLApplication and SamzaSQLApplicationRunner. Is there a way to refactor them in a way so that we create them only once and reuse them?

samza-sql/src/test/java/org/apache/samza/sql/e2e/TestSamzaSqlTable.java
```diff
@@ -58,7 +58,7 @@ public void testEndToEndWithKey() throws Exception {
     TestIOResolverFactory.TestTable.records.clear();
     Map<String, String> staticConfigs = SamzaSqlTestConfig.fetchStaticConfigsWithFactories(numMessages);
 
-    String sql1 = "Insert into testDb.testTable.`$table` select id __key__, name from testavro.SIMPLE1";
+    String sql1 = "Insert into testDb.testTable.`$table`(id,name) select id __key__, name from testavro.SIMPLE1";
```
srinipunuru: After this checkin, Do we expect users to always specify the field names in insert into?

atoomula: No, this typically happens only in tests where only a subset of the output schema fields are projected in select statement. In real world scenario, the output schema should reflect the fields projected in the select statement.

samza-sql/src/main/java/org/apache/samza/sql/dsl/SamzaSqlDslConverter.java
atoomula: Good catch. Fixed it!

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SamzaSqlJavaTypeFactoryImpl.java
atoomula: Added more comments on this to the class javadoc.

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationRunner.java
atoomula: Removed creation of DSLConverter from SamzaSqlApplicationRunner. It doesn't look like it is needed in the runner. SQL EndToEnd tests passed without this.

samza-sql/src/main/java/org/apache/samza/sql/translator/ModifyTranslator.java
```diff
@@ -0,0 +1,113 @@
...
+
+  private final Map<String, SamzaRelConverter> relMsgConverters;
+  private final Map<String, SqlIOConfig> systemStreamConfig;
+
+  ModifyTranslator(Map<String, SamzaRelConverter> converters, Map<String, SqlIOConfig> ssc) {
+    relMsgConverters = converters;
+    this.systemStreamConfig = ssc;
+  }
+
+  private static class OutputMapFunction implements MapFunction<SamzaSqlRelMessage, KV<Object, Object>> {
```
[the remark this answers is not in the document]
atoomula: Added

samza-sql/src/main/java/org/apache/samza/sql/translator/ModifyTranslator.java
```diff
@@ -0,0 +1,113 @@
...
+    private transient SamzaRelConverter samzaMsgConverter;
```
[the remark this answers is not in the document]
atoomula: Done

PR #632, opened 2018-09-10
SAMZA-1865: Increase coordination timeouts in zookeeper integration tests.

shanthoosh:


shanthoosh:
@xinyuiscool @sborya 

shanthoosh:
During the re-balancing phase, stream processors in the group have to stop the `SamzaContainer` within the shutdown-timeout(task.shutdown.ms) and join the barrier.

If the samza container did not stop within this configured timeout, then the stream processor is killed.

In the failed integration tests, `task.shutdown.ms` is set to 2 seconds. In some scenarios, this strict timeout has caused the container shutdown timeout took more than 2 seconds there by killing the stream processor.

Since the integration test expects the stream processor to be alive, the assertion fail. 

The change here is to increase the coordination timeouts. In longer run, if we want faster dev builds we can extract these integration tests out in a seperate module and run them nightly.

vjagadish1989 (approved):
approved, thanks Shanthoosh!

PR #637, opened 2018-09-12
SAMZA-1860: Modularize Join input validation in ExecutionPlanner

ahmedahamid:
This change breaks down the validation of partition counts of input and
intermediate streams participating in Join operations into 3 separate steps:
1. Grouping `InputOperatorSpec`s by the `JoinOperatorSpec`s of the Join operations they participate in
2. Replacing `InputOperatorSpec`s with their corresponding `StreamEdge`s
3. Verifying/Inferring partition counts of input/intermediate streams

This change covers stream-stream Joins only.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
+++ b/samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
@@ -22,12 +22,14 @@ package org.apache.samza.execution;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
-import java.util.Queue;
 import java.util.Set;
 import org.apache.samza.SamzaException;
 import org.apache.samza.config.ApplicationConfig;
@@ -36,13 +38,14 @@ import org.apache.samza.config.Config;
 import org.apache.samza.config.JobConfig;
 import org.apache.samza.config.StreamConfig;
 import org.apache.samza.operators.OperatorSpecGraph;
+import org.apache.samza.operators.spec.InputOperatorSpec;
 import org.apache.samza.operators.spec.JoinOperatorSpec;
-import org.apache.samza.operators.spec.OperatorSpec;
 import org.apache.samza.system.StreamSpec;
 import org.apache.samza.table.TableSpec;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.samza.execution.ExecutionPlanner.JoinedStreamsGroup.JoinedStreamsGroupCategory;
 import static org.apache.samza.util.StreamUtil.*;
 
 
@@ -69,14 +72,22 @@ public class ExecutionPlanner {
   public ExecutionPlan plan(OperatorSpecGraph specGraph) {
     validateConfig();
 
-    // create physical job graph based on stream graph
+    // Create physical job graph based on stream graph
     JobGraph jobGraph = createJobGraph(specGraph);
 
-    // fetch the external streams partition info
-    fetchInputAndOutputStreamPartitions(jobGraph, streamManager);
+    // Fetch the external streams partition info
+    fetchInputAndOutputStreamPartitions(jobGraph);
 
-    // figure out the partitions for internal streams
-    calculatePartitions(jobGraph);
+    // Verify agreement in partition count between all joined input/intermediate streams
+    validateJoinInputStreamPartitions(jobGraph);
+
+    if (!jobGraph.getIntermediateStreamEdges().isEmpty()) {
+      // Set partition count of intermediate streams not participating in joins
+      setIntermediateStreamPartitions(jobGraph);
+
+      // Validate partition counts were assigned for all intermediate streams
+      validateIntermediateStreamPartitions(jobGraph);
+    }
 
     return jobGraph;
   }
@@ -92,7 +103,7 @@ public class ExecutionPlanner {
   }
 
   /**
-   * Create the physical graph from {@link OperatorSpecGraph}
+   * Creates the physical graph from {@link OperatorSpecGraph}
    */
   /* package private */ JobGraph createJobGraph(OperatorSpecGraph specGraph) {
     JobGraph jobGraph = new JobGraph(config, specGraph);
@@ -131,25 +142,9 @@ public class ExecutionPlanner {
   }
 
   /**
-   * Figure out the number of partitions of all streams
+   * Fetches the partitions of input/output streams and update the corresponding StreamEdges.
    */
-  /* package private */ void calculatePartitions(JobGraph jobGraph) {
-    // calculate the partitions for the input streams of join operators
-    calculateJoinInputPartitions(jobGraph, streamConfig);
-
-    // calculate the partitions for the rest of intermediate streams
-    calculateIntermediateStreamPartitions(jobGraph, config);
-
-    // validate all the partitions are assigned
-    validateIntermediateStreamPartitions(jobGraph);
-  }
-
-  /**
-   * Fetch the partitions of input/output streams and update the corresponding StreamEdges.
-   * @param jobGraph {@link JobGraph}
-   * @param streamManager the {@link StreamManager} to interface with the streams.
-   */
-  /* package private */ static void fetchInputAndOutputStreamPartitions(JobGraph jobGraph, StreamManager streamManager) {
+  /* package private */ void fetchInputAndOutputStreamPartitions(JobGraph jobGraph) {
     Set<StreamEdge> existingStreams = new HashSet<>();
     existingStreams.addAll(jobGraph.getInputStreams());
     existingStreams.addAll(jobGraph.getOutputStreams());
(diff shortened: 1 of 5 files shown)


prateekm:
@bharathkk @vjagadish1989 Can you take a look at this?

prateekm:
cc @nickpan47, since this affects the ExecutionPlanner.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -188,98 +183,82 @@ private void validateConfig() {
...
+    // Determine category of stream group based on stream partition counts.
+    JoinedStreamsGroupCategory category;
+    if (countStreamEdgeWithSetPartitions == 0) {
+      category = JoinedStreamsGroupCategory.NO_PARTITION_COUNT_SET;
+    } else if (countStreamEdgeWithSetPartitions == streamEdges.size()) {
+      category = JoinedStreamsGroupCategory.ALL_PARTITION_COUNT_SET;
+    } else {
+      category = JoinedStreamsGroupCategory.SOME_PARTITION_COUNT_SET;
     }
+
```
vjagadish1989: The notion of a "category" looks like a detail of the `JoinedStreamsGroup`, which can be inferred from the other params. What do you think about moving the logic that determines "category" to the constructor of `JoinedStreamsGroup` ?

A nice property is that we could avoid inconsistent object states. For eg: with the current constructor: `JoinedStreamsGroup(groupId, streamEdges, category)`, one could create an instance of `JoinedStreamsGroup` such that its "streamEdges" and "category" contradict each other

samza-core/src/main/java/org/apache/samza/execution/OperatorSpecGraphAnalyzer.java
```diff
@@ -0,0 +1,89 @@
...
+    return joinedInputOperatorSpecVisitor.getJoinOpSpecToInputOpSpecs();
+  }
+
+  /**
+   * Traverses graph starting from {@code vertex}, invoking {@code visitor} with every encountered vertex,
+   * and using {@code getNextVertexes} to determine next set of vertexes to visit.
+   */
+  private static <T> void traverse(T vertex, Consumer<T> visitor, Function<T, Iterable<? extends T>> getNextVertexes) {
+    visitor.accept(vertex);
+    for (T nextVertex : getNextVertexes.apply(vertex)) {
```
vjagadish1989: Question: Could you have cycles anywhere in the traversal? If so, should this method guard against that? Instead, if the visitors are expected to track and avoid cycles, would be worth documenting it

samza-core/src/main/java/org/apache/samza/execution/OperatorSpecGraphAnalyzer.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    return joinedInputOperatorSpecVisitor.getJoinOpSpecToInputOpSpecs();
+  }
+
+  /**
+   * Traverses graph starting from {@code vertex}, invoking {@code visitor} with every encountered vertex,
+   * and using {@code getNextVertexes} to determine next set of vertexes to visit.
+   */
+  private static <T> void traverse(T vertex, Consumer<T> visitor, Function<T, Iterable<? extends T>> getNextVertexes) {
```
vjagadish1989: s/vertex/startVertex

samza-core/src/main/java/org/apache/samza/execution/OperatorSpecGraphAnalyzer.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  /**
+   * An {@link OperatorSpecGraph} visitor that records associations between {@link InputOperatorSpec}s
+   * and the {@link JoinOperatorSpec}s they participate in or lie on a path to.
+   */
+  private static class JoinedInputOperatorSpecVisitor implements Consumer<OperatorSpec> {
+    private final Multimap<JoinOperatorSpec, InputOperatorSpec> joinOpSpecToInputOpSpecs = HashMultimap.create();
+
+    private InputOperatorSpec currentInputOpSpec;
```
vjagadish1989: Any reason for the JoinedInputOperatorSpecVisitor to contain a mutable reference to the currentInputOpSpec ? If this is to track the first "InputOperatorSpec" encountered by the traversal, are there other ways to track it? 

For eg: you could have a visitor instance per InputOperatorSpec instead of sharing the same visitor instance all multiple traversals. You can poke each visitor for the set of reachable joined operator-specs and assemble the multi-map outside this class

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -188,98 +183,82 @@ private void validateConfig() {
...
-          // find other joins can be inferred by setting this edge
-          for (JoinOperatorSpec op : streamEdgeToJoinSpecs.get(edge)) {
-            if (!visited.contains(op)) {
-              joinQ.add(op);
-              visited.add(op);
-            }
-          }
-        }
-      }
+  private void validateJoinInputStreamPartitions(JobGraph jobGraph) {
```
bharathkk: I feel the essence of this method is already captured in the method name and having to suffix `joined` in declaration/method names/comments within seems redundant and in some cases confusing. 

E.g. `joinedStreamGroups` - it is not obvious or intuitive as to what this means.

What do you think?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -188,98 +183,82 @@ private void validateConfig() {
...
+     *   edges (partition counts in parentheses, question marks for intermediate streams):
+     *
+     *      a. e1 (16), e2 (16)
+     *      b. e2 (16), e3 (?)
+     *      c. e3 (?), e4 (?)
+     *
+     *   processing them in the above order (most constrained first) is guaranteed to
+     *   yield correct assignment of partition counts of e3 and e4 in a single scan.
+     */
+    Collections.sort(joinedStreamsGroups, Comparator.comparing(JoinedStreamsGroup::getCategory));
```
bharathkk: This is very loosely tied up with the ordering of the enums in the declaration. 
It is error-prone and not obvious if someone decides to change the ordering for e.g. to be alphabetical. 

I'd suggest we either have these enums as complex ones with explicit priority and have that used for comparison or introduce a comparator which specifically handles priorities that way it is much explicit and perceivable.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -188,98 +183,82 @@ private void validateConfig() {
...
+      streamEdges.add(streamEdge);
     }
 
-    Collection<OperatorSpec> registeredOperatorSpecs = operatorSpec.getRegisteredOperatorSpecs();
-    for (OperatorSpec registeredOpSpec : registeredOperatorSpecs) {
-      findReachableJoins(registeredOpSpec, sourceStreamEdge, joinSpecToStreamEdges, streamEdgeToJoinSpecs, joinQ,
-          visited);
+    // Determine category of stream group based on stream partition counts.
+    JoinedStreamsGroupCategory category;
+    if (countStreamEdgeWithSetPartitions == 0) {
```
bharathkk: How about moving this logic within the enum and exposing it as a static helper? 
E.g. ```JoinedStreamsGroupCategory.of(int numberOfKnownPartitions);```

It will be useful for the tests as well.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -323,7 +305,93 @@ private static void validateIntermediateStreamPartitions(JobGraph jobGraph) {
     }
   }
 
+  /**
+   * Ensures that all streams in supplied {@link JoinedStreamsGroup} agree in partition count.
+   * This includes setting partition counts of any intermediate streams that do not have their
+   * partition counts set.
+   */
+  private static void validateJoinedStreamsGroupPartitions(JoinedStreamsGroup joinedStreamsGroup) {
```
bharathkk: Is it the case where inference and validation need to coexist? 
If yes, can we rename this method to something reflective instead of `validateJoinedStreamsGroupPartitioms` which seems to suggest this is a non-mutative method?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -323,7 +305,93 @@ private static void validateIntermediateStreamPartitions(JobGraph jobGraph) {
     }
   }
 
+  /**
+   * Ensures that all streams in supplied {@link JoinedStreamsGroup} agree in partition count.
+   * This includes setting partition counts of any intermediate streams that do not have their
```
bharathkk: We are not setting the count for any intermediate streams here. It is only those streams that feed into joins that already have a source with known partition count.
Can we reflect that in the comment?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -323,7 +305,93 @@ private static void validateIntermediateStreamPartitions(JobGraph jobGraph) {
...
+  }
+
   /* package private */ static int maxPartitions(Collection<StreamEdge> edges) {
     return edges.stream().mapToInt(StreamEdge::getPartitionCount).max().orElse(StreamEdge.PARTITIONS_UNKNOWN);
   }
+
+  /**
+   * Represents a group of {@link StreamEdge}s participating in a Join operation.
+   */
+  /* package private */ static class JoinedStreamsGroup {
```
bharathkk: `Joined` is an overloaded term and can mean multiple things.
Can we rename this to something else?

I am open to any suggestions here.

samza-core/src/main/java/org/apache/samza/execution/OperatorSpecGraphAnalyzer.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public void accept(OperatorSpec operatorSpec) {
+      if (operatorSpec instanceof InputOperatorSpec) {
+        currentInputOpSpec = (InputOperatorSpec) operatorSpec;
+      } else if (operatorSpec instanceof JoinOperatorSpec) {
```
bharathkk: What about `StreamTableJoinOperatorSpec`?

samza-core/src/main/java/org/apache/samza/execution/OperatorSpecGraphAnalyzer.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.samza.operators.spec.InputOperatorSpec;
+import org.apache.samza.operators.spec.JoinOperatorSpec;
+import org.apache.samza.operators.spec.OperatorSpec;
+
+
+/**
+ * A utility class that encapsulates the logic for traversing an {@link OperatorSpecGraph} and building
+ * associations between related {@link OperatorSpec}s.
+ */
+/* package private */ class OperatorSpecGraphAnalyzer {
```
bharathkk: Here are my thoughts on this class. I feel we don't have a pressing need for it to be generified yet.
Can we start with a simple helper class that does traversal and returns a mapping in one go? 

By that, 
* we are still isolating the traversal logic.
* It simplifies the review
* It simplifies testing

We can always refactor to extract the traversal part if we plan to introduce more visitors.
Thoughts?

vjagadish1989: @bharathkk : We'll likely introduce more visitors for computing partition-counts for StreamTableJoin and side-inputs. Once we have those follow-up PRs, we can decide if the current implementation is overly general. If that is indeed the case, we can certainly revisit it.  

Until then, it's probably efficient to leave this PR in its current state. What do you think?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -188,98 +183,82 @@ private void validateConfig() {
...
-              joinQ.add(op);
-              visited.add(op);
-            }
-          }
-        }
-      }
+  private void validateJoinInputStreamPartitions(JobGraph jobGraph) {
+    // Group joined input operator specs (input/intermediate streams) by the join
+    // operator specs corresponding to the join operations they participate in.
+    Multimap<JoinOperatorSpec, InputOperatorSpec> joinOpSpecToInputOpSpecs =
```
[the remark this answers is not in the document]
ahmedahamid: This PR covers stream-stream joins only. Stream-table joins will be covered in a subsequent PR.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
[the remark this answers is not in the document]
ahmedahamid: I came to the same conclusion actually — making `StreamEdge` immutable can greatly improve the `ExecutionPlanner`. I didn't want to do it in this series of PRs to avoid a scoop creep though.

I think one idea we can explore is making partition count a ctor param and a readonly property of a `StreamEdge`. This would require deferring the creation of any `StreamEdge` until its partition count is known, which is probably not going to be difficult after this PR.

I'll take note of this and send a follow-up PR later on.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -188,98 +183,82 @@ private void validateConfig() {
...
+     *      c. e3 (?), e4 (?)
+     *
+     *   processing them in the above order (most constrained first) is guaranteed to
+     *   yield correct assignment of partition counts of e3 and e4 in a single scan.
+     */
+    Collections.sort(joinedStreamsGroups, Comparator.comparing(JoinedStreamsGroup::getCategory));
+
+    // Verify agreement between joined input/intermediate streams.
+    // This may involve setting partition counts of intermediate stream edges.
+    joinedStreamsGroups.forEach(ExecutionPlanner::validateJoinedStreamsGroupPartitions);
```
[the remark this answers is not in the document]
ahmedahamid: Yes, validate is also assigning partition counts. It's a little difficult to separate both operations at the moment. For groups containing stream edges with known partition counts only, we just need verification — this is the easy case. On the other hand, stream edge groups with a mix of set/unset partitions require a mix of verification and assignment.

I thought of deciding whether to do verification vs assignment depending on the `JoinedStreamsGroupCategory` of a `JoinedStreamsGroup`. Problem with that is:

1. For a group with a mix of set/unset partitions, I could very easily need to do both verification and assignment within the same group, e.g. {e1 (8), e2 (?), e3 (8)}.

2. `JoinedStreamsGroupCategory` is not really reliable once we start setting partition counts (which is another reason why `StreamEdge`s are better of being immutable). For instance, by the time we process group #2 of {e1 (8), e2 (?)} and {e2 (?), e3 (8)}, it will only require verification even though its (stale) category will be `SOME_PARTITION_COUNT_SET`.

I think we can just change the verb from `validate` to something that conveys possibility of mutation. I'll try to come up with something but I'm also open to suggestions.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: I like the idea of a static helper on the enum but let me ask a clarifying question: how is this static helper going to decide the category based on `numberOfKnownPartitions` only? Or did you mean it should receive 2 numbers, stream edge count & known partitions count?

If it's the latter then I'm not sure it's any better to write
`new JoinedStreamsGroup(groupId, streamEdges, JoinedStreamsGroupCategory.of(countStreamEdgeWithSetPartitions, streamEdges.size()))`.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: I did consider this and I would have really liked to do so. However,

1. Deciding the category in `JoinedStreamsGroup` requires one more (redundant) iteration over the `StreamEdge`s.

2. More importantly, there is no way to avoid inconsistent object states in `JoinedStreamsGroup` because `StreamEdge`s are mutable. In fact, we already throw every `JoinedStreamsGroup` with  intermediate streams into this inconsistent state once we start setting partition counts. This made me less inclined to incur the overhead of the extra iteration in #1.

Together, these 2 points made me prefer to keep `JoinedStreamsGroup` as a passive data object that contains zero logic. I thought this would help set the readers` expectations that this object is just a way of organizing data.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: We _are_ doing both verification and assignment in this method. Please, refer to my response to [this comment](https://github.com/apache/samza/pull/637/files#r219718076) for more details.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: I agree. I'll try to think of something and I'm also open to suggestions.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: I'm open to suggestions as well but let me ask you why the name felt confusing in this context. I thought it wasn't bad; a `JoinedStreamsGroup` is a group of `StreamEdge`s participating in a Join.

samza-core/src/main/java/org/apache/samza/execution/OperatorSpecGraphAnalyzer.java
ahmedahamid: If we're only doing stream-stream join then I agree. The reason I wrote it this way though is because I have another PR that will add one more visitor in order to support stream-table joins. I wanted to lay the grounds for the upcoming change since I have already finished coding it.

samza-core/src/main/java/org/apache/samza/execution/OperatorSpecGraphAnalyzer.java
ahmedahamid: I don't think there could be cycles in the `OperatorSpecGraph`, and if there could be then we've never handled them. 

A visitor's responsibility is strictly dictated by `getNextVertexes`, and since both are user-supplied, it's all up to the user. There are no general requirements on visitors.

samza-core/src/main/java/org/apache/samza/execution/OperatorSpecGraphAnalyzer.java
ahmedahamid: ["This change covers stream-stream joins only"](https://github.com/apache/samza/pull/637#issue-214839421) :)

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: Please, mark every such case individually. I can relate to what you're saying to some extent in the case of `JoinedStreamGroup` but I don't see many other cases.

As for `joinedStreamGroups`, would you prefer `joinedStreamEdges`? I don't have a strong preference but I don't think `streamEdges` would be sufficiently clear. Please, let me know what you think.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -188,98 +183,82 @@ private void validateConfig() {
...
-          for (JoinOperatorSpec op : streamEdgeToJoinSpecs.get(edge)) {
-            if (!visited.contains(op)) {
-              joinQ.add(op);
-              visited.add(op);
-            }
-          }
-        }
-      }
+  private void validateJoinInputStreamPartitions(JobGraph jobGraph) {
+    // Group joined input operator specs (input/intermediate streams) by the join
```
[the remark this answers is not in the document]
ahmedahamid: How about now?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
ahmedahamid: Changed to `StreamEdgeSet`.

samza-core/src/main/java/org/apache/samza/execution/OperatorSpecGraphAnalyzer.java
ahmedahamid: @bharathkk  @vjagadish1989 Would you prefer I send you the follow-up PRs now or is this PR good to go?

vjagadish1989 (approved):
approved

vjagadish1989:
merged and submitted!

samza-core/src/main/java/org/apache/samza/execution/OperatorSpecGraphAnalyzer.java
nickpan47: @ahmedahamid sorry to be late to the party. One comment for your future PR: OperatorSpecGraph is *only* available for high-level APIs. In the effort to unify the runtime support for both high-level and low-level APIs, ExecutionPlanner and the corresponding JobGraph/JobNode classes now have only access to ApplicationDescriptorImpl. Any need to traverse the graph should be starting from ApplicationDescriptorImpl.getInputOperators() now (see PR #642)

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -188,98 +183,81 @@ private void validateConfig() {
...
-        joinQ.add(joinOperatorSpec);
-        visited.add(joinOperatorSpec);
+  private StreamEdgeSet getStreamEdgeSet(String setId, Iterable<InputOperatorSpec> inputOpSpecs,
+      JobGraph jobGraph) {
+
+    int countStreamEdgeWithSetPartitions = 0;
+    Set<StreamEdge> streamEdges = new HashSet<>();
+
+    for (InputOperatorSpec inputOpSpec : inputOpSpecs) {
+      StreamEdge streamEdge = jobGraph.getOrCreateStreamEdge(getStreamSpec(inputOpSpec.getStreamId(), streamConfig));
```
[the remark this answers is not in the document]
nickpan47: Let's take this comment to a separate PR. I realized that this is an existing pattern in ExecutionPlanner/JobGraph. Ideally, get/create streamEdge from JobGraph should have separate methods, one should be read-only.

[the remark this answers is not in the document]
ahmedahamid: Certainly agree. This call site is using `getOrCreateStreamEdge` to retrieve existing `StreamEdge`s not create new edges.

PR #639, opened 2018-09-14
SAMZA-1879: Remove deprecated containerId from ContainerModel

cameronlee314:


samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
```diff
@@ -39,7 +38,6 @@
  * an id, partition information, etc.
  * </p>
  */
-@JsonIgnoreProperties(ignoreUnknown = true)
 public class JobModel {
```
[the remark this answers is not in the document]
cameronlee314: Yep, this line was moved to the mixin, which is used to add annotations to JobModel. So it should be the same behavior. This change decouples JobModel from Jackson.
There is a unit test for deserializing with unknown fields, so that verifies the functionality.

cameronlee314:
@shanthoosh can you please take a look?

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
```diff
@@ -301,7 +301,7 @@ private void assignTasksToContainers(int[] taskCountPerContainer, List<String> t
         containerTaskModels.put(model.getTaskName(), model);
       }
       containerModels.add(
-          new ContainerModel(container.containerId, Integer.valueOf(container.containerId), containerTaskModels));
```
shanthoosh: Since we're getting rid of containerId field from ContainerModel, just for the sake of consistency in other internal classes like `TaskGroup`, can we rename instance-fields, variables like `containerId` to `processorId`?  

Otherwise, after this change it will be hard for a newcomer without context to understand why `processorId` and `containerId` are used interchangeably in the grouper and it's internal classes.

 If you agree to this, then please change it.

samza-core/src/main/java/org/apache/samza/serializers/model/JsonContainerModelMixIn.java
```diff
@@ -19,25 +19,19 @@
...
-
 /**
  * A mix-in Jackson class to convert Samza's ContainerModel to/from JSON.
+ * Note: Constructor is not needed because this mixin is not used for deserialization. See {@link SamzaObjectMapper} for
+ * the deserialization code.
  */
+@JsonIgnoreProperties({"container-id"})
 public abstract class JsonContainerModelMixIn {
-  @JsonCreator
-  public JsonContainerModelMixIn(@JsonProperty("processor-id") String processorId, @JsonProperty("tasks") Map<TaskName, TaskModel> tasks) {
```
shanthoosh: Why did we delete the `Mixin` constructor only in this class and retain it in the other `Mixin` classes(`JsonTaskModelMixin`, `JsonJobModelMixin`).

samza-core/src/main/java/org/apache/samza/serializers/model/JsonJobModelMixIn.java
```diff
@@ -23,11 +23,13 @@
 import org.apache.samza.config.Config;
 import org.apache.samza.job.model.ContainerModel;
 import org.codehaus.jackson.annotate.JsonCreator;
+import org.codehaus.jackson.annotate.JsonIgnoreProperties;
 import org.codehaus.jackson.annotate.JsonProperty;
 
 /**
  * A mix-in Jackson class to convert Samza's JobModel to/from JSON.
  */
+@JsonIgnoreProperties(ignoreUnknown = true)
```
shanthoosh: Do we need to add this annotation to `JsonTaskModelMixin` as well?

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
prateekm: I'd prefer to not call fields/getters just "id". We have so many that it can be confusing to figure out which one is intended at call sites.

+1 for renaming fields and variables. We can do a general s/Container/Processor renaming later.

shanthoosh: > What do you think about naming the field id in ContainerModel and having a getId method.

Yes that sounds like a good idea to hide the complexity. At-least it would prevent us from exposing `getProcessorId` method in `ContainerModel`(`getProcessorId` seems to be misnomer to expose from `ContainerModel`).

samza-core/src/main/java/org/apache/samza/serializers/model/JsonJobModelMixIn.java
prateekm: @cameronlee314 Should we add them to other models as well? It makes migrations simpler if we add new properties in future. Specifically, allows rolling back to a version that does not have the newly added field.

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
cameronlee314: Discussed with @prateekm offline. Since callers will need to use containerModel.getId, it still is clear enough that it is the container id. Callers will need to ensure that they clearly name their variables so that they don't get mixed up by the multiple kinds of ids.

samza-core/src/main/java/org/apache/samza/serializers/model/JsonJobModelMixIn.java
cameronlee314: Good point. I'll add it to the others.

cameronlee314: It turns out that ContainerModel already has custom deserialization code which will ignore unknown properties (verified with unit test). So I only needed to add for TaskModel.

PR #642, opened 2018-09-17
Samza 1814: consolidate JobNode and JobGraph configuration generation for high and low-level API applications

nickpan47:
High-level changes:
- Move configuration generation to JobNodeConfigurationGenerator
- Move the intermediate partition calculation to IntermediationStreamPartitionPlanner
- Consolidate the code in JobPlanner and ExecutionPlanner for high and low-level API plan/configuration generation


nickpan47:
@prateekm @vjagadish and @cameronlee314 please take a look.

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
```diff
@@ -0,0 +1,340 @@
...
+
+  static final String CONFIG_JOB_PREFIX = "jobs.%s.";
+  static final String CONFIG_INTERNAL_EXECUTION_PLAN = "samza.internal.execution.plan";
+
+  JobNodeConfigureGenerator() {
+  }
+
+  static Config mergeJobConfig(Config originalConfig, Config generatedConfig) {
+    JobConfig jobConfig = new JobConfig(originalConfig);
+    String jobId = JobNode.createId(jobConfig.getName().get(), jobConfig.getJobId());
```
cameronlee314: It looks like this now uses job name and job id for the prefix, but before it looked like it was just job name. Is that an intentional change?

samza-core/src/main/java/org/apache/samza/execution/JobPlanner.java
```diff
@@ -149,25 +125,6 @@ final void writePlanJsonFile(String planJson) {
...
-  private JobConfig prepareTaskJob(TaskApplicationDescriptorImpl taskAppDesc) {
-    // copy original configure
-    Map<String, String> cfg = new HashMap<>();
-    // expand system and streams configure
-    Map<String, String> systemStreamConfigs = expandSystemStreamConfigs(taskAppDesc);
-    cfg.putAll(systemStreamConfigs);
-    // expand table configure
-    cfg.putAll(expandTableConfigs(cfg, taskAppDesc));
-    // adding app.class in the configuration
-    cfg.put(ApplicationConfig.APP_CLASS, appDesc.getAppClass().getName());
```
cameronlee314: Just to double check: Is this config no longer necessary?

samza-core/src/main/java/org/apache/samza/execution/IntermediateStreamPartitionPlanner.java
```diff
@@ -0,0 +1,207 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.samza.util.StreamUtil.getStreamSpec;
+
+
+/**
+ * {@link IntermediateStreamPartitionPlanner} calculates intermediate stream partitions based on the high-level application graph.
+ */
+class IntermediateStreamPartitionPlanner {
```
prateekm: Would prefer not calling this class a Planner, to avoid confusion with ExecutionPlanner. Maybe IntermediateStreamManager/IntermediateStreamPartitionCalculator?

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -125,27 +123,16 @@ private void validateConfig() {
...
-    // calculate the partitions for the rest of intermediate streams
-    calculateIntermediateStreamPartitions(jobGraph, config);
-
-    // validate all the partitions are assigned
-    validateIntermediateStreamPartitions(jobGraph);
-  }
-
-  /**
-   * Fetch the partitions of input/output streams and update the corresponding StreamEdges.
+   * Fetch the partitions of source/sink streams and update the StreamEdges.
```
prateekm: Maybe s/fetchInputAndOutputStreamPartitions/setInputAndOutputStreamPartitionCount for this method and javadoc to reflect that this method doesn't just fetch the count, it also sets it on the StreamSpec.

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
```diff
@@ -105,14 +109,11 @@ public String getPlanAsJson() throws Exception {
    * Returns the config for this application
    * @return {@link ApplicationConfig}
    */
+  @Override
   public ApplicationConfig getApplicationConfig() {
```
prateekm: Where is this method used? If this class doesn't mutate the config, would be nice to not expose the config from here.

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
```diff
@@ -183,11 +184,11 @@ StreamEdge getOrCreateStreamEdge(StreamSpec streamSpec) {
    * @param isIntermediate  boolean flag indicating whether it's an intermediate stream
    * @return stream edge
    */
-  StreamEdge getOrCreateStreamEdge(StreamSpec streamSpec, boolean isIntermediate) {
+  private StreamEdge getOrCreateStreamEdge(StreamSpec streamSpec, boolean isIntermediate) {
     String streamId = streamSpec.getId();
     StreamEdge edge = edges.get(streamId);
     if (edge == null) {
-      boolean isBroadcast = specGraph.getBroadcastStreams().contains(streamId);
```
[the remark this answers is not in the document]
prateekm: @cameronlee314 I'll address this in the broadcast stream PR. There are two sources of broadcast inputs: user defined inputs and those added by the high level API's broadcast method. This is tracking the latter.

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -117,6 +116,9 @@
     String applicationId;
   }
 
+  JobGraphJsonGenerator() {
```
prateekm: Minor: Can probably skip default constructor.

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
```diff
@@ -19,361 +19,150 @@
...
   }
 
-  public static Config mergeJobConfig(Config fullConfig, Config generatedConfig) {
-    return new JobConfig(Util.rewriteConfig(extractScopedConfig(
-        fullConfig, generatedConfig, String.format(JobConfig.CONFIG_JOB_PREFIX(), new JobConfig(fullConfig).getName().get()))));
-  }
-
-  public OperatorSpecGraph getSpecGraph() {
-    return this.specGraph;
+  static String createId(String jobName, String jobId) {
```
prateekm: Minor: Would prefer a more explicit method name reflecting what this ID is, since there are so many Ids.

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
```diff
@@ -19,361 +19,150 @@
...
-        .collect(Collectors.toList());
-
-    // Combine both the above lists
-    List<Long> candidateTimerIntervals = new ArrayList<>(joinTtlIntervals);
-    candidateTimerIntervals.addAll(windowTimerIntervals);
-
-    if (candidateTimerIntervals.isEmpty()) {
-      return -1;
+  // get all next operators consuming from the input {@code streamId}
+  Set<String> getNextOperatorIds(String streamId) {
```
prateekm: Minor: Prefer to order methods in class by visibility (constructors, public, protected, package private, private)

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
```diff
@@ -19,361 +19,150 @@
...
-    scopedConfig = new MapConfig(mergedConfig);
-    log.debug("Prefix '{}' has merged config {}", configPrefix, scopedConfig);
-
-    return scopedConfig;
+    return appDesc.getInputOperators().get(inputStreamId);
   }
 
-  static String createId(String jobName, String jobId) {
-    return String.format("%s-%s", jobName, jobId);
+  boolean isLegacyTaskApplication() {
```
prateekm: Minor: would prefer to not proxy a util method from this class. Can this be done by the caller directly?

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
```diff
@@ -0,0 +1,340 @@
...
+import org.apache.samza.util.StreamUtil;
+import org.apache.samza.util.Util;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * This class provides methods to generate configuration for a {@link JobNode}
+ */
+/* package private */ class JobNodeConfigureGenerator {
```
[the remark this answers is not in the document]
prateekm: +1

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
```diff
@@ -0,0 +1,340 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(JobNodeConfigureGenerator.class);
+
+  static final String CONFIG_JOB_PREFIX = "jobs.%s.";
```
prateekm: Minor: Prefer naming this something more explicit. Maybe OVERRIDE_CONFIG_PREFIX or something similar?

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  static Config mergeJobConfig(Config originalConfig, Config generatedConfig) {
+    JobConfig jobConfig = new JobConfig(originalConfig);
+    String jobId = JobNode.createId(jobConfig.getName().get(), jobConfig.getJobId());
+    return new JobConfig(Util.rewriteConfig(extractScopedConfig(originalConfig, generatedConfig,
+        String.format(CONFIG_JOB_PREFIX, jobId))));
+  }
+
+  JobConfig generateJobConfig(JobNode jobNode, String executionPlanJson) {
```
prateekm: Do you know why the executionPlanJson needs to be passed all the way through to here? Could this be added to the config by the runner/planner itself?

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
```diff
@@ -0,0 +1,340 @@
...
+    // finalize the task.inputs configuration
+    configs.put(TaskConfig.INPUT_STREAMS(), Joiner.on(',').join(inputs));
+
+    LOG.info("Job {} has generated configs {}", jobNode.getId(), configs);
+
+    // apply configure rewriters and user configure overrides
+    return applyConfigureRewritersAndOverrides(configs, config, jobNode);
+  }
+
+  private void configureBroadcastInputs(Map<String, String> configs, Config config, Set<String> broadcasts) {
```
prateekm: Minor: s/broadcasts/broadcastStreams

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
```diff
@@ -0,0 +1,340 @@
...
+        inputs.add(formattedSystemStream);
+      }
+    }
+
+    if (!broadcasts.isEmpty()) {
+      configureBroadcastInputs(configs, config, broadcasts);
+    }
+
+    // compute window and join operator intervals in this node
+    if (hasWindowOrJoin) {
```
prateekm: Minor: could the hasWindowOrJoin logic be encapsulated within the configureWindowAndJoinInterval?

Also, this method should be `configureWindowInterval`, since it's referring to configuring the task.window.ms.

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  private JobConfig applyConfigureRewritersAndOverrides(Map<String, String> configs, Config config, JobNode jobNode) {
+    String configPrefix = String.format(CONFIG_JOB_PREFIX, jobNode.getId());
+
+    // Disallow user specified job inputs/outputs. This info comes strictly from the user application.
+    Map<String, String> allowedConfigs = new HashMap<>(config);
+    if (!jobNode.isLegacyTaskApplication()) {
+      if (allowedConfigs.containsKey(TaskConfig.INPUT_STREAMS())) {
+        LOG.warn("Specifying task inputs in configuration is not allowed with Fluent API. " + "Ignoring configured value for " + TaskConfig.INPUT_STREAMS());
```
prateekm: Minor: s/with Fluent API/for SamzaApplications

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
```diff
@@ -0,0 +1,340 @@
...
+        allowedConfigs.remove(TaskConfig.INPUT_STREAMS());
+      }
+    }
+
+    LOG.debug("Job {} has allowed configs {}", jobNode.getId(), allowedConfigs);
+    return new JobConfig(Util.rewriteConfig(
```
prateekm: Can this call mergeJobConfig?

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
```diff
@@ -0,0 +1,340 @@
...
+        String serdeName = serdeUUIDs.computeIfAbsent(serde,
+            s -> serde.getClass().getSimpleName() + "-" + UUID.randomUUID().toString());
+        configs.putIfAbsent(String.format(SerializerConfig.SERDE_SERIALIZED_INSTANCE(), serdeName),
+            base64Encoder.encodeToString(serializableSerde.toBytes(serde)));
+      });
+
+    // set key and msg serdes for streams to the serde names generated above
+    streamKeySerdes.forEach((streamId, serde) -> {
+        String streamIdPrefix = String.format(StreamConfig.STREAM_ID_PREFIX(), streamId);
+        String keySerdeConfigKey = streamIdPrefix + StreamConfig.KEY_SERDE();
```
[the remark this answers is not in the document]
prateekm: The original goal was to serialize all serde instances in one pass to preserve the number of serde instances before and after deserialization. However Tables already break that guarantee since they serialize their Serde separately, so we can consider moving this logic to StreamDescriptor.

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
```diff
@@ -0,0 +1,340 @@
...
+    storeMsgSerdes.forEach((storeName, serde) -> {
+        String msgSerdeConfigKey = String.format(StorageConfig.MSG_SERDE(), storeName);
+        configs.put(msgSerdeConfigKey, serdeUUIDs.get(serde));
+      });
+  }
+
+  /**
+   * Computes the triggering interval to use during the execution of this {@link JobNode}
+   */
+  private long computeTriggerInterval(Collection<OperatorSpec> reachableOperators) {
```
prateekm: Minor: This should be after configureWindowInterval method since that one uses this.

samza-core/src/main/java/org/apache/samza/execution/JobPlanner.java
```diff
@@ -64,35 +55,20 @@
...
-      } catch (Exception e) {
-        throw new SamzaException("Failed to generate JobConfig for StreamApplication " + appId, e);
-      }
-    }
-    throw new IllegalArgumentException(String.format("ApplicationDescriptorImpl has to be either TaskApplicationDescriptorImpl or "
-        + "StreamApplicationDescriptorImpl. class %s is not supported", appDesc.getClass().getName()));
-  }
-
-  abstract List<JobConfig> prepareStreamJobs(StreamApplicationDescriptorImpl streamAppDesc) throws Exception;
+  public abstract List<JobConfig> prepareJobs() throws Exception;
```
prateekm: Minor: Can remove throws Exception?

samza-core/src/main/java/org/apache/samza/execution/LocalJobPlanner.java
```diff
@@ -53,25 +52,26 @@ public LocalJobPlanner(ApplicationDescriptorImpl<? extends ApplicationDescriptor
   }
 
   @Override
-  List<JobConfig> prepareStreamJobs(StreamApplicationDescriptorImpl streamAppDesc) throws Exception {
+  public List<JobConfig> prepareJobs() throws Exception {
```
prateekm: Nice, thanks for consolidating the method signature.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -164,7 +164,7 @@ class JobConfig(config: Config) extends ScalaMapConfig(config) with Logging {
 
   def getStreamJobFactoryClass = getOption(JobConfig.STREAM_JOB_FACTORY_CLASS)
 
-  def getJobId = getOption(JobConfig.JOB_ID)
+  def getJobId = getOption(JobConfig.JOB_ID).getOrElse("1")
```
prateekm: Thanks!

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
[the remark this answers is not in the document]
nickpan47: Sure.

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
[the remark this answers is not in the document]
nickpan47: @prateekm I like the idea, that would actually simplify the serde configuration generation a lot. When we also have Input/OutputDescriptor for intermediate streams, we don't even need this centralized serde generation method. I will work on that.

samza-core/src/main/java/org/apache/samza/application/ApplicationDescriptorImpl.java
```diff
@@ -176,4 +179,53 @@ public ProcessorLifecycleListenerFactory getProcessorLifecycleListenerFactory()
    */
   public abstract Set<SystemDescriptor> getSystemDescriptors();
 
+  /**
+   * Get all the unique input streamIds in this application
+   *
+   * @return an immutable set of input streamIds
+   */
+  public abstract Set<String> getInputStreamIds();
```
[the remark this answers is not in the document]
nickpan47: This is not necessary if we unify the input streams set between high and low-level applications. Currently, the full set of input stream IDs are stored differently in high and low- level APIs: a) in TaskApplicationDescriptorImpl, getInputDescriptor() will give you the full set of input streams, including the intermediate streams; b) in StreamApplicationDescriptorImpl, getInputOperators() give you the full set of input streams, including the intermediate streams. If we unify the representation of intermediate streams in TaskApplicationDescriptorImpl and StreamApplicationDescriptorImpl, we can remove this method. Before we do that, we need two implementations of the same method in the base class to get the consistent set of input streamIds. I would prefer to open another ticket for that, since this is also only in the Impl classes, not the public API interface class.

samza-core/src/main/java/org/apache/samza/application/ApplicationDescriptorImpl.java
```diff
@@ -176,4 +179,53 @@ public ProcessorLifecycleListenerFactory getProcessorLifecycleListenerFactory()
...
+
+  /**
+   * Get all the unique output streamIds in this application
+   *
+   * @return an immutable set of output streamIds
+   */
+  public abstract Set<String> getOutputStreamIds();
```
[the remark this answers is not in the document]
nickpan47: See my answers above as well. Differences in TaskApplicationDescriptorImpl and StreamApplicationDescriptorImpl: getOutputDescriptors() vs getOutputStreams()

samza-core/src/main/java/org/apache/samza/application/ApplicationDescriptorImpl.java
```diff
@@ -176,4 +179,53 @@ public ProcessorLifecycleListenerFactory getProcessorLifecycleListenerFactory()
...
+
+  /**
+   * Get the corresponding {@link Serde} for the input {@code inputStreamId}
+   *
+   * @param inputStreamId id of the input stream
+   * @return the {@link Serde} for the input stream. null if the serde is not defined or {@code inputStreamId}
+   *         does not exist
+   */
+  public abstract Serde getInputSerde(String inputStreamId);
```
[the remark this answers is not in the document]
nickpan47: Good point. Again, the main purpose to introduce this abstract method is still to bridge the gap between TaskApplicationDescriptorImpl and StreamApplicationDescriptorImpl (i.e. they store input streams in different format: inputDescriptors vs inputOperators).  I will change the name explicitly to getInputKVSerde().

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -125,27 +123,16 @@ private void validateConfig() {
     // Add tables
     tables.forEach(spec -> jobGraph.addTable(spec, node));
 
-    jobGraph.validate();
+    if (!ApplicationUtil.isLegacyTaskApplication(appDesc)) {
+      // skip the validation when input streamIds are empty. This is only possible for LegacyApplication
```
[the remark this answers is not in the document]
nickpan47: Thanks for catching that! Fixed.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -92,21 +90,21 @@ private void validateConfig() {
...
+      JobGraphJsonGenerator jobJsonGenerator, JobNodeConfigureGenerator jobConfigureGenerator) {
+    JobGraph jobGraph = new JobGraph(config, appDesc, jobJsonGenerator, jobConfigureGenerator);
+    StreamConfig streamConfig = new StreamConfig(config);
+    Set<StreamSpec> sourceStreams = getStreamSpecs(appDesc.getInputStreamIds(), streamConfig);
+    Set<StreamSpec> sinkStreams = getStreamSpecs(appDesc.getOutputStreamIds(), streamConfig);
     Set<StreamSpec> intermediateStreams = Sets.intersection(sourceStreams, sinkStreams);
     Set<StreamSpec> inputStreams = Sets.difference(sourceStreams, intermediateStreams);
     Set<StreamSpec> outputStreams = Sets.difference(sinkStreams, intermediateStreams);
 
-    Set<TableSpec> tables = specGraph.getTables().keySet();
```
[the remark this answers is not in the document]
nickpan47: It is not used for now. We might need this if we consolidate the ApplicationDescriptorImpl more and allow TaskApplicationDescriptorImpl to generate a single operator OperatorSpecGraph, w/ task class modeled as a single black box terminal TaskOperatorSpec. That would require some further refactoring, but fully internal to Samza, not to public APIs. We can remove it and revisit later when we come back for the refactor here.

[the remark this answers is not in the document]
nickpan47: Changed to VisibleForTesting now, since it is still useful in unit tests.

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
```diff
@@ -19,361 +19,150 @@
...
   private final String id;
-  private final OperatorSpecGraph specGraph;
-  private final List<StreamEdge> inEdges = new ArrayList<>();
-  private final List<StreamEdge> outEdges = new ArrayList<>();
-  private final List<TableSpec> tables = new ArrayList<>();
   private final Config config;
-
-  JobNode(String jobName, String jobId, OperatorSpecGraph specGraph, Config config) {
+  private final JobNodeConfigureGenerator configGenerator;
+  private final Map<String, StreamEdge> inEdges = new HashMap<>();
```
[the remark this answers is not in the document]
nickpan47: Sure. That's a good idea.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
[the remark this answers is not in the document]
cameronlee314: If it is not used though, then it seems like it should not be part of unit tests either.

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
nickpan47: Yes. Using jobName only is not a guaranteed globally unique name when we extend the current planner to create multi-stage jobs for an application. I took a stab to update this along with the code change to ensure that job configuration checks streams/ops belonging to the JobNode, instead of blindly generating and traversing the whole DAG to create configuration.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
nickpan47: Make sense. Thanks!

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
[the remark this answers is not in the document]
nickpan47: Actually, talked with Wei and confirmed that in runtime, TableSpec is re-constructed from configure. Hence, I am removing the whole internal variable tables in OperatorSpecGraph, together with the access method.

samza-core/src/main/java/org/apache/samza/execution/IntermediateStreamPartitionPlanner.java
```diff
@@ -0,0 +1,207 @@
...
+    calculateIntStreamPartitions(jobGraph);
+
+    // validate all the partitions are assigned
+    validatePartitions(jobGraph);
+  }
+
+  /**
+   * Calculate the partitions for the input streams of join operators
+   */
+  /* package private */ void calculateJoinInputPartitions(JobGraph jobGraph) {
```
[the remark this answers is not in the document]
nickpan47: Currently used in unit test. Change to VisibleForTesting now.

samza-core/src/main/java/org/apache/samza/execution/IntermediateStreamPartitionPlanner.java
```diff
@@ -0,0 +1,207 @@
...
+
+  private static void validatePartitions(JobGraph jobGraph) {
+    for (StreamEdge edge : jobGraph.getIntermediateStreamEdges()) {
+      if (edge.getPartitionCount() <= 0) {
+        throw new SamzaException(String.format("Failure to assign the partitions to Stream %s", edge.getName()));
+      }
+    }
+  }
+
+  /* package private */ static int maxPartition(Collection<StreamEdge> edges) {
```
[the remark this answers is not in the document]
nickpan47: Same here (i.e. VisibleForTesting)

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
nickpan47: This is used in JobGraphJsonGenerator and RemoteJobPlanner. Also, defined as the public interface method in ExecutionPlan. Besides, the ApplicationConfig object is immutable as it extends Config.

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
nickpan47: Make sense. I will change this to createJobNameAndId() together with getId() -> getJobNameAndId

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
nickpan47: Sure.

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
nickpan47: Sounds good.

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
nickpan47: Sounds good to me.

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
nickpan47: So, this json string can not be generated until we create the JobGraph in the planner. And the reason to pass it to each JobNodeConfigurationGenerator is to allow the DAG to be displaced by each job, not just once per application.

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
nickpan47: Not sure which word/variable name in the return line you are referring to? Is it referring to "extractScopedConfig"?

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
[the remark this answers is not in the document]
nickpan47: Actually, it might be nicer to keep a per JobNode serde map that are used by all serde configure generation for all streams/stores/tables. Let me work on that.

samza-core/src/main/java/org/apache/samza/execution/JobPlanner.java
nickpan47: Sure.

samza-core/src/main/java/org/apache/samza/execution/JobPlanner.java
nickpan47: After consolidation, this configuration is always added in getExecutionPlan() method.

samza-core/src/main/java/org/apache/samza/application/ApplicationDescriptorImpl.java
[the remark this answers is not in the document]
nickpan47: Actually changed it to getStreamSerdes() to return a KV<Serde, Serde> now to be explicit about what we are getting here: the exact KV pair of Serdes used for configuration. Will update the document here. Also included getTableSerdes() as well to allow centralized serde configuration generation.

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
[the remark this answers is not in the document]
nickpan47: I actually deleted it for now. @prateekm you can add it later if proper.

samza-core/src/main/java/org/apache/samza/application/StreamApplicationDescriptorImpl.java
```diff
@@ -239,6 +239,35 @@ public StreamApplicationDescriptor withDefaultSystem(SystemDescriptor<?> default
...
+    return Collections.unmodifiableSet(new HashSet<>(outputStreams.keySet()));
+  }
+
+  @Override
+  public Serde getInputSerde(String inputStreamId) {
+    if (!inputOperators.containsKey(inputStreamId)) {
+      // the corresponding inputStreamId does not exist
+      return null;
+    }
+    InputOperatorSpec inputOperator = inputOperators.get(inputStreamId);
```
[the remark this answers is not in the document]
nickpan47: Actually created two maps track the serdes for streams and tables: Map<String, KV<Serde, Serde>>.

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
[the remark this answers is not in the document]
prateekm: @nickpan47 Deleting this will break existing support for high level API broadcast streams. If we keep this, adding support for declaring input descriptors is an additive step that we can choose to do after the next release. If we remove this, it'll be a must have. So would prefer to keep this for now.

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
prateekm: Not sure why this is part of the API though. Since JobNode already exposes the generated config, callers could probably do `new ApplicationConfig(generatedConfig)` and use it. If so, would prefer to remove this, since it's confusing to have many different getters for config that don't specify how they're different from each other.

samza-core/src/main/java/org/apache/samza/execution/JobNode.java
prateekm: Another suggestion: Maybe jobUID since name + id uniquely identifies a job.

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigureGenerator.java
prateekm: Sorry, I meant can this entire line be replace by the new mergeJobConfig method? Seems to be doing a similar thing.

samza-core/src/main/java/org/apache/samza/application/ApplicationDescriptorImpl.java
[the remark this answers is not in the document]
prateekm: Minor: In the mean time, it might be helpful to change the internal API such that instead of getting the entire Map of input descriptors and input operators, we only expose a getInputDescriptor(String streamId) and getInputOperator(String streamId). This way it's clearer that the getInputStreamIds is the method that should be used for iteration.

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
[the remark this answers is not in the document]
nickpan47: Discussed with @prateekm and will leave it as it now for later complete fix of the broadcast stream configuration.

nickpan47:
@prateekm @cameronlee314 could you take a second pass? Thanks!

samza-core/src/main/java/org/apache/samza/application/ApplicationDescriptorImpl.java
```diff
@@ -176,4 +218,66 @@ public ProcessorLifecycleListenerFactory getProcessorLifecycleListenerFactory()
...
+  public abstract Set<String> getInputStreamIds();
+
+  /**
+   * Get all the unique output streamIds in this application
+   *
+   * @return an immutable set of output streamIds
+   */
+  public abstract Set<String> getOutputStreamIds();
+
+  KV<Serde, Serde> getOrCreateStreamSerdes(String streamId, Serde serde) {
```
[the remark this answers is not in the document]
nickpan47: Use "serdes" since the final result is a pair of serdes: KV&lt;Serde, Serde&gt;

samza-core/src/main/java/org/apache/samza/application/TaskApplicationDescriptorImpl.java
```diff
@@ -65,6 +66,7 @@ public void addInputStream(InputDescriptor inputDescriptor) {
     // TODO: SAMZA-1841: need to add to the broadcast streams if inputDescriptor is for a broadcast stream
     Preconditions.checkState(!inputDescriptors.containsKey(inputDescriptor.getStreamId()),
         String.format("add input descriptors multiple times with the same streamId: %s", inputDescriptor.getStreamId()));
+    getOrCreateStreamSerdes(inputDescriptor.getStreamId(), inputDescriptor.getSerde());
```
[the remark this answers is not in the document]
nickpan47: The serdes are included in the descriptor. However, a) the actual KV&lt;Serde, Serde&gt; to serialize/deserialize the incoming message key and msg may be different from the Serde in descriptor; b) there is no unified API between StreamApplicationDescriptorImpl and TaskApplicationDescriptorImpl to get the input/output serdes, since they are spread across in stream descriptors and OperatorSpec/OutputStreamImpl.

nickpan47:
Discussed offline with @prateekm and he agreed to merge this in. Thanks all!

PR #647, opened 2018-09-18
SAMZA-1868: New kafka consumer admin (using high level consumer)

sborya:
Rewrite of KafkaSystemAdmin in java, using new KafkaSystemConsumer (based on high level kafka consumer).


samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdminUtilsScala.scala
```diff
@@ -0,0 +1,187 @@
...
+
+  def createStream(kSpec: KafkaStreamSpec, connectZk: java.util.function.Supplier[ZkUtils]): Boolean = {
+    LOG.info("Create topic %s in system %s" format(kSpec.getPhysicalName, systemName))
+    var streamCreated = false
+
+    new ExponentialSleepStrategy(initialDelayMs = 500).run(
+      loop => {
+        val zkClient = connectZk.get()
+        try {
+          AdminUtils.createTopic(
```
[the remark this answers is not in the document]
sborya: We will. It is still missing some functionality, which is available in the next kafka client. When we move to the next kafka client we should change to use it.

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+  // The same default exponential sleep strategy values as in open source
+  private static final double DEFAULT_EXPONENTIAL_SLEEP_BACK_OFF_MULTIPLIER = 2.0;
+  private static final long DEFAULT_EXPONENTIAL_SLEEP_INITIAL_DELAY_MS = 500;
+  private static final long DEFAULT_EXPONENTIAL_SLEEP_MAX_DELAY_MS = 10000;
+  private static final int MAX_RETRIES_ON_EXCEPTION = 5;
+
+  @Override
+  public void start() {
+    // Plese note. There is slight inconsistency in the use of this class.
+    // Some of the functinality of this class may actually be used BEFORE start() is called.
```
srinipunuru: This seems bit confusing, Can we just deprecate start from the systemAdmin interface altogether? Any initialization that the systemAdmin implemetations needs to perform can be performed in the constructor right? This no argument start doesn't seem to add lot of value.

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+  }
+
+  /**
+   * Converts a StreamSpec into a KafakStreamSpec. Special handling for coordinator and changelog stream.
+   * @param spec a StreamSpec object
+   * @return KafkaStreamSpec object
+   */
+  KafkaStreamSpec toKafkaSpec(StreamSpec spec) {
+    KafkaStreamSpec kafkaSpec;
+    if (spec.isChangeLogStream()) {
```
shanthoosh: KafkaSystemAdmin should not concern itself with coordinator stream properties, changelog properties, intermediate stream properties. These streams are generally high level samza concepts and the underlying IO should not concern itself with constructing different properties bag for each stream. It's the responsibilty of the caller to pass in appropriate properties for each stream type before creating them.

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+              expectedPartitionCounter, actualPartitionCounter));
+    }
+  }
+
+  /**
+   * get partition info for topic
+   * @param topics set of topics to query
+   */
+  Map<String, List<PartitionInfo>> getTopicMetadata(Set<String> topics) {
+    // me
```
shanthoosh: Please delete this //me comment. I don't understand the significance of it.

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+    final Map<String, List<PartitionInfo>> listPartitionsInfo = new HashMap();
```
shanthoosh: s/listPartitionsInfo/streamToPartitionInfoMapping

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+    List<PartitionInfo> l;
+    for (String topic : topics) {
+      l = metadataConsumer.partitionsFor(topic);
```
shanthoosh: s/l/partitionsList

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+  private static final long DEFAULT_EXPONENTIAL_SLEEP_MAX_DELAY_MS = 10000;
+  private static final int MAX_RETRIES_ON_EXCEPTION = 5;
+
+  @Override
+  public void start() {
+    // Plese note. There is slight inconsistency in the use of this class.
+    // Some of the functinality of this class may actually be used BEFORE start() is called.
+    // The SamzaContainer gets metadata (using this class) in SamzaContainer.apply, but this "start" actually gets called in SamzaContainer.run.
+    // Also we assume that start is called only once.
+    if (metadataConsumer == null) {
```
shanthoosh: These if guards around `metadataConsumer`, `adminClient` are duplicated across in the ` start()`, `stop()` and individual API methods. Rather than duplicating these checks, we should add `Precondition.notNull` checks in the constructor once and delete it in the other individual API methods.

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+import scala.runtime.AbstractFunction1;
+import scala.runtime.AbstractFunction2;
+import scala.runtime.BoxedUnit;
+
+import static org.apache.kafka.clients.consumer.KafkaConsumerConfig.*;
+
+
+public class SamzaKafkaSystemAdmin<K, V> implements ExtendedSystemAdmin {
+  private final String systemName;
+  private Consumer<K, V> metadataConsumer = null;
```
shanthoosh: Why are the instance fields `metadataConsumer`, `adminClient` assigned to null value initially? They're set only in the constructor alone and their access is private. We should make them final and remove the initial null value assignment.

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+  private final int coordinatorStreamReplicationFactor;
+
+  // Replication factor for the Changelog topic in kafka
+  //Kafka properties to be used during the Changelog topic creation
+  private final Map<String, ChangelogInfo> changelogTopicMetaInformation;
+
+  // Kafka properties to be used during the intermediate topic creation
+  private final Map<String, Properties> intermediateStreamProperties;
+
+  private static final Logger LOG = LoggerFactory.getLogger(SamzaKafkaSystemAdmin.class);
```
shanthoosh: Please move static final fields`(LOG`, `DEFAULT_EXPONENTIAL_SLEEP_BACK_OFF_MULTIPLIER` and other constants) to the beginning of the class and then reorganize the private final instance fields after that.

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+  public static volatile boolean deleteMessageCalled = false;
+
+  // The same default exponential sleep strategy values as in open source
+  private static final double DEFAULT_EXPONENTIAL_SLEEP_BACK_OFF_MULTIPLIER = 2.0;
+  private static final long DEFAULT_EXPONENTIAL_SLEEP_INITIAL_DELAY_MS = 500;
+  private static final long DEFAULT_EXPONENTIAL_SLEEP_MAX_DELAY_MS = 10000;
+  private static final int MAX_RETRIES_ON_EXCEPTION = 5;
+
+  @Override
+  public void start() {
```
shanthoosh: Do we need to make `start()`, `stop()` idempotent?

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+  private final Map<String, ChangelogInfo> changelogTopicMetaInformation;
+
+  // Kafka properties to be used during the intermediate topic creation
+  private final Map<String, Properties> intermediateStreamProperties;
+
+  private static final Logger LOG = LoggerFactory.getLogger(SamzaKafkaSystemAdmin.class);
+
+  private final KafkaSystemAdminUtilsScala kafkaAdminUtils;
+
+  volatile private AdminClient adminClient = null;
```
shanthoosh: s/volatile private/private volatile/

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+
+      // Kafka's beginning Offset corresponds to the offset for the oldest message.
+      // Kafka's end offset corresponds to the offset for the upcoming message, and it is the newest offset + 1.
+      // When upcoming offset is <=0, the topic appears empty, we put oldest offset 0 and the newest offset null.
+      // When upcoming offset is >0, we subtract the upcoming offset by one for the newest offset.
+      // For normal case, the newest offset will correspond to the offset of the newest message in the stream;
+      // But for the big message, it is not the case. Seeking on the newest offset gives nothing for the newest big message.
+      // For now, we keep it as is for newest offsets the same as historical metadata structure.
+      if (offset <= 0) {
+        LOG.debug(
```
shanthoosh: Log level should be warn here.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdminUtilsScala.scala
```diff
@@ -0,0 +1,187 @@
...
+
+import scala.collection.JavaConverters._
+
+class KafkaSystemAdminUtilsScala(systemName: String) {
+  val CLEAR_STREAM_RETRIES = 3
+
+  val LOG: Logger = LoggerFactory.getLogger("KafkaSystemAdminUtils")
+
+  def createStream(kSpec: KafkaStreamSpec, connectZk: java.util.function.Supplier[ZkUtils]): Boolean = {
+    LOG.info("Create topic %s in system %s" format(kSpec.getPhysicalName, systemName))
```
shanthoosh: s/Create topic %s in system %s/ Creating topic %s for system %s

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdminUtilsScala.scala
```diff
@@ -0,0 +1,187 @@
...
+import org.apache.samza.config.ApplicationConfig.ApplicationMode
+import org.apache.samza.config.{ApplicationConfig, Config, KafkaConfig, StreamConfig}
+import org.apache.samza.system.SystemStreamMetadata.SystemStreamPartitionMetadata
+import org.apache.samza.system.{StreamSpec, SystemStreamMetadata, SystemStreamPartition}
+import org.apache.samza.util.ExponentialSleepStrategy
+import org.slf4j.{Logger, LoggerFactory}
+
+import scala.collection.JavaConverters._
+
+class KafkaSystemAdminUtilsScala(systemName: String) {
```
shanthoosh: Should name this just KafkaSystemAdminHelper or KafkaSystemAdminUtil(or some other better name), adding the scala suffix to the classname doesn't add much value.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdminUtilsScala.scala
```diff
@@ -0,0 +1,187 @@
...
+        streamCreated = true
+        loop.done
+      },
+
+      (exception, loop) => {
+        exception match {
+          case e: TopicExistsException =>
+            streamCreated = false
+            loop.done
+          case e: Exception =>
```
shanthoosh: When will this retry loop end in case of irrecoverable failures(Let's say that the broker is down or broker configuration is mis-configured by the user)?. In error scenarios,  looks like this retry loop will go on forever.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdminUtilsScala.scala
```diff
@@ -0,0 +1,187 @@
...
+          zkClient.close
+        }
+
+        streamCreated = true
+        loop.done
+      },
+
+      (exception, loop) => {
+        exception match {
+          case e: TopicExistsException =>
```
shanthoosh: We should already log the exception at the warn log level.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdminUtilsScala.scala
```diff
@@ -0,0 +1,187 @@
...
+            streamCreated = false
+            loop.done
+          case e: Exception =>
+            LOG.warn("Failed to create topic %s: %s. Retrying." format(kSpec.getPhysicalName, e))
```
shanthoosh: 1. Log level should be error here. 
2. We can combine the two log line into one line like the following: 
LOG.error("Exception occurred on creating topic: {}." , kSpec.getPhysicalName, e)

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdminUtilsScala.scala
```diff
@@ -0,0 +1,187 @@
...
+
+        loop.done
+      },
+
+      (exception, loop) => {
+        if (retries > 0) {
+          LOG.warn("Exception while trying to delete topic %s: %s. Retrying." format (spec.getPhysicalName, exception))
+          retries -= 1
+        } else {
+          LOG.warn("Fail to delete topic %s: %s" format (spec.getPhysicalName, exception))
```
shanthoosh: This log statement will not print out the entire stack trace. Should switch to the appender format to print it completely.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -71,6 +74,7 @@ class KafkaSystemFactory extends SystemFactory with Logging {
     // Unlike consumer, no need to use encoders here, since they come for free
     // inside the producer configs. Kafka's producer will handle all of this
     // for us.
+    info("Creating kafka producer for system %s, idPrefix %s" format(systemName, clientId))
```
shanthoosh: Would be better to rename idPrefix with `consumerClientId`. idPrefix doesn't share if it's part of consumerId  or producerId or some other uniqueID in samza.

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+
+  public static SamzaKafkaSystemAdmin getKafkaSystemAdmin(final String systemName, final Config config,
+      final String idPrefix) {
+
+    boolean zkSecure = false; // needs to be added to the argument if ever true is possible
+
+    // admin requires adminClient for deleteBeforeMessages
+    Properties props = new Properties();
+    String brokerListString =
+        config.get(String.format("systems.%s.consumer.%s", systemName, ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG));
```
shanthoosh: Should  use either `ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG` or `CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG` consistently here.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdminUtilsScala.scala
sborya: I've added the retries, and thus have both warn (while retrying) and error (when no more retries left).

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemAdminJava.java
```diff
@@ -172,20 +236,27 @@ public void testValidateStreamWrongName() {
...
 
-    assertTrue("createStream should return true if the stream does not exist and then is created.", systemAdmin().createStream(spec1));
+    assertTrue("createStream should return true if the stream does not exist and then is created.",
+        systemAdmin().createStream(spec1));
 
     systemAdmin().validateStream(spec2);
   }
 
-  @Test
+  //@Test TODO - see if it can be fixed
```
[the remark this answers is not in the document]
sborya: this test didn't work in the original code either (the failure was masked).  SAMZA-1887 was filed for it.

sborya:
This PR got garbled. Opened #662 in its stead.

PR #649, opened 2018-09-19
Samza 1849: Table Descriptors should take Serde at construction time so that descriptors can be typed

weisong44:
Changed table descriptor to take serde in constructor, and removed withSerde() from all table descriptors.


samza-kv-inmemory/src/main/java/org/apache/samza/storage/kv/inmemory/InMemoryTableDescriptor.java
```diff
@@ -40,6 +41,13 @@ public InMemoryTableDescriptor(String tableId) {
     super(tableId);
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  public InMemoryTableDescriptor(String tableId, KVSerde<K, V> serde) {
```
prateekm: I think inheritDoc doesn't work for constructors (since they're not overrides). Same comment for other classes as well.

vjagadish1989 (approved):
approved.

PR #651, opened 2018-09-20
SAMZA-1852 Adding default job system in TestRunner, disabling host affinity to support TableDescriptors and refining addConfig method for TestRunner API           

Sanil15:
- The default system is a required config for intermediate streams, and since no user will write assertions against them, defaulting it makes it easier for the user to write test
- To support stateful jobs using Table API descriptors we need to disable host affinity, which is enabled by table API by default 
- @vjagadish pointed out addConfigs vs addOverrideConfig to be a confusing user-facing API. We now support only addConfig with different signatures, this configs takes precedence over any descriptor or TestRunner generated configs


samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -102,11 +106,15 @@ private TestRunner() {
     configs.put(JobCoordinatorConfig.JOB_COORDINATION_UTILS_FACTORY, PassthroughCoordinationUtilsFactory.class.getName());
     configs.put(JobCoordinatorConfig.JOB_COORDINATOR_FACTORY, PassthroughJobCoordinatorFactory.class.getName());
     configs.put(TaskConfig.GROUPER_FACTORY(), SingleContainerGrouperFactory.class.getName());
+    configs.put(JobConfig.JOB_DEFAULT_SYSTEM(), JOB_DEFAULT_SYSTEM);
```
prateekm: Does this need to be addConfig as well? Users might be setting a default system descriptor in their application themselves.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -384,4 +395,15 @@ private TaskFactory createTaskFactory() {
           new EndOfStreamMessage(null)));
       });
   }
+
+  /**
+   * Only adds a config from {@code config} to samza job {@code configs} if they dont exist in it.
+   * @param config for the application
+   * @return this {@link TestRunner}
+   */
+  private TestRunner putIfAbsentConfig(Map<String, String> config) {
```
prateekm: Don't need a helper method for a one line method, can inline this.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -148,28 +156,31 @@ public static TestRunner of(StreamApplication app) {
   }
 
   /**
-   * Only adds a config from {@code config} to samza job {@code configs} if they dont exist in it.
-   * @param config configs for the application
+   * Adds a config to Samza application, this config takes precedence over default configs and descriptor generated configs
```
prateekm: Use period instead of comma for unrelated sentences. "Adds a ... application. This config ..."

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -148,28 +156,31 @@ public static TestRunner of(StreamApplication app) {
...
+   * Adds {@code config} to Samza application, these configs takes precedence over default configs
+   * and descriptor generated configs
+   *
+   * @param config configuration for samza application
    * @return this {@link TestRunner}
    */
-  public TestRunner addOverrideConfig(String key, String value) {
-    Preconditions.checkNotNull(key);
-    Preconditions.checkNotNull(value);
+  public TestRunner addConfig(Config config) {
```
prateekm: This should take a Map<String, String> instead of Config.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -148,28 +156,31 @@ public static TestRunner of(StreamApplication app) {
...
+    Preconditions.checkNotNull(config);
     String configKeyPrefix = String.format(JobConfig.CONFIG_JOB_PREFIX(), JOB_NAME);
-    configs.put(String.format("%s%s", configKeyPrefix, key), value);
+    config.forEach((key, value) -> configs.put(String.format("%s%s", configKeyPrefix, key), value));
```
prateekm: This can delegate to addConfig instead.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: if there is no usage of default system other than intermediate streams (when the user needs to set this), I can use addConfig

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
nickpan47: @Sanil15 this is related to the type-erasure of generic types. If you declare descriptor as InMemoryInputDescriptor&lt;?&gt;, the one liner would work. Otherwise, since the descriptor is of an erased type InMemoryInputDescriptor, Java will complain about type conversion failure in compile time.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -384,4 +396,15 @@ private TaskFactory createTaskFactory() {
           new EndOfStreamMessage(null)));
       });
   }
+
+  /**
+   * Only adds a config from {@code config} to samza job {@code configs} if they dont exist in it.
```
shanthoosh: This java doc comment is confusing. Might be good to rename  to the following(or something better):
```
Adds new configuration from {@param config} to samza application configuration {@code configs}
```

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -366,8 +378,8 @@ private TaskFactory createTaskFactory() {
     }
     InMemorySystemDescriptor imsd = (InMemorySystemDescriptor) descriptor.getSystemDescriptor();
     imsd.withInMemoryScope(this.inMemoryScope);
-    addConfigs(descriptor.toConfig());
-    addConfigs(descriptor.getSystemDescriptor().toConfig());
+    putIfAbsentConfig(descriptor.toConfig());
```
shanthoosh: `putIfAbsentConfig` API here is entirely unnecessary. You had to do this, since we didn't use the StreamMessageType as a type for the API signature(rather used a raw type). Rather than the following(at line 369): 

```java
 private <StreamMessageType> void initializeInMemoryInputStream(InMemoryInputDescriptor descriptor,
       Map<Integer, Iterable<StreamMessageType>> partitonData) {	       Map<Integer, Iterable<StreamMessageType>> partitonData) {
```

We can do the following :

```java
 private <StreamMessageType> void initializeInMemoryInputStream(InMemoryInputDescriptor<? extends StreeamMessageType> descriptor,
       Map<Integer, Iterable<StreamMessageType>> partitonData) {	       Map<Integer, Iterable<StreamMessageType>> partitonData) {
```

Since we haven't explicitly defined the type for genericly parameterized class, type-erasure kicks in and all the methods from `InMemoryInputDescriptor` return response of object type. If you do the above, then you don't have to do explicit typecast in all the places(For instance, at line 372 from `Object` to `String`).

After the above API signature change  `putifAbsentConfig` API will be unnecessary, and could be simply in-lined it with the following: 

```java
    descriptor.toConfig().forEach(this.configs::putIfAbsent);
```

If it does makes sense, then please change it.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -229,8 +241,8 @@ public TestRunner addOutputStream(InMemoryOutputDescriptor streamDescriptor, int
     factory
         .getAdmin(streamDescriptor.getSystemName(), config)
         .createStream(spec);
-    addConfigs(streamDescriptor.toConfig());
-    addConfigs(streamDescriptor.getSystemDescriptor().toConfig());
+    putIfAbsentConfig(streamDescriptor.toConfig());
```
nickpan47: Can we add to the Javadoc of this method to explain why we are calling putIfAbsent here? I assume the reason is to allow user supplied config to supersede descriptor generated config? Can we make it clear?

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
nickpan47: Same here.

samza-test/src/test/java/org/apache/samza/test/framework/StreamTaskIntegrationTest.java
```diff
@@ -102,7 +102,7 @@ public void testSyncTaskWithSinglePartitionMultithreaded() throws Exception {
         .of(MyStreamTestTask.class)
         .addInputStream(imid, inputList)
         .addOutputStream(imod, 1)
-        .addOverrideConfig("job.container.thread.pool.size", "4")
+        .addConfig("job.container.thread.pool.size", "4")
```
nickpan47: 👍 Nice!

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -148,28 +158,30 @@ public static TestRunner of(StreamApplication app) {
...
+   * @param value of the config
    * @return this {@link TestRunner}
    */
-  public TestRunner addConfigs(Map<String, String> config) {
-    Preconditions.checkNotNull(config);
-    config.forEach(this.configs::putIfAbsent);
+  public TestRunner addConfig(String key, String value) {
+    Preconditions.checkNotNull(key);
+    Preconditions.checkNotNull(value);
+    String configKeyPrefix = String.format(JobConfig.CONFIG_JOB_PREFIX(), JOB_NAME);
```
nickpan47: Please refer to PR #642 , we need to have the full jobNameAndId in the config job prefix.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -229,8 +243,8 @@ public TestRunner addOutputStream(InMemoryOutputDescriptor streamDescriptor, int
     factory
         .getAdmin(streamDescriptor.getSystemName(), config)
         .createStream(spec);
-    addConfigs(streamDescriptor.toConfig());
-    addConfigs(streamDescriptor.getSystemDescriptor().toConfig());
+    streamDescriptor.toConfig().forEach(this.configs::putIfAbsent);
```
nickpan47: minor: add a comment here to explain why we use putIfAbsent, instead of overwriting.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: Did that

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: Using the confix prefix, added a comment to docs

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -79,10 +81,13 @@
  *    <li>"task.name.grouper.factory" = {@link SingleContainerGrouperFactory}</li>
  *    <li>"job.name" = "test-samza"</li>
  *    <li>"processor.id" = "1"</li>
+ *    <li>"job.default.system" = {@code JOB_DEFAULT_SYSTEM}</li>
+ *    <li>"job.host-affinity.enabled" = "false"</li>
  *  </ol>
  *
  */
 public class TestRunner {
+  public static final String JOB_DEFAULT_SYSTEM = "default-samza-system";
```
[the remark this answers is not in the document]
Sanil15: Reverted the anti-pattern, made the constants private

PR #652, opened 2018-09-21
Samza-1818: Cleaning up TestRunner Apis to incorporate LegacyTaskApplication and StreamApplication

Sanil15:


samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -106,43 +99,30 @@ private TestRunner() {
...
+   * Creates an instance of {@link TestRunner} for Legacy Samza Api
+   * @param taskApp legacy samza task application
    * @return this {@link TestRunner}
    */
-  public static TestRunner of(Class taskClass) {
-    Preconditions.checkNotNull(taskClass);
-    Preconditions.checkState(
-        StreamTask.class.isAssignableFrom(taskClass) || AsyncStreamTask.class.isAssignableFrom(taskClass));
-    return new TestRunner(taskClass);
+  public static TestRunner of(LegacyTaskApplication taskApp) {
```
prateekm: Why do you need a separate API for LegacyTaskApplication? It should extend SamzaApplication too.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -106,43 +99,30 @@ private TestRunner() {
...
+    Preconditions.checkNotNull(taskApp);
+    return new TestRunner(taskApp);
   }
 
   /**
    * Creates an instance of {@link TestRunner} for High Level/Fluent Samza Api
```
prateekm: s/for High Level .../for a {@link SamzaApplication}/

samza-test/src/test/java/org/apache/samza/test/framework/StreamTaskIntegrationTest.java
```diff
@@ -123,7 +124,7 @@ public void testSyncTaskWithMultiplePartition() throws Exception {
         .getOutputDescriptor("output", new NoOpSerde<Integer>());
 
     TestRunner
-        .of(MyStreamTestTask.class)
+        .of(new LegacyTaskApplication(MyStreamTestTask.class.getName()))
```
nickpan47: FWIW, please take a look at PR #642 , most of the changes here are already incorporated in that PR. Question here: any specific reason that you want to explicitly construct a LegacyTaskApplication here, instead of retaining the same API, but internally convert the MyStreamTestTask.class to a LegacyTaskApplication instance in TestRunner?

Sanil15: We have three choices for Public Api

Option 1: `TestRunner.of(SamzaApplication)`, `TestRunner.of(AsyncStreamTask)`, `TestRunner.of(StreamTask)`

Option 2:  `TestRunner.of(SamzaApplication)`, `TestRunner.of(LegacyTaskApplication)`,

Option 3: `TestRunner.of(SamzaApplication)`, `TestRunner.of(Class taskClass)`

I prefer the option 2 makes it more straighforward for end user, option 3 is very generic and vague (which I want to remove)

I am ok with either Option 1 or Option 2. I will leave it up to the reviewers to suggest

Note: when i was read Samza-1818, i thought I am responsible for the refacotr, if #642 gets merged first I can rebase the changes

nickpan47: What's the specific concern on option 3? Type safety? We can add that in the TestRunner.of() method as well. If your concern is that what end user will see as a supported implementation of application, we already support the legacy case that end user only implements the task class anyways and I don't think that will add additional confusion. In fact, exposing the constructor of LegacyTaskApplication is worse, since we don't expect users construct the LegacyTaskApplication at all in the supported APIs.

Sanil15: Sure, will revert the changes

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -106,43 +99,30 @@ private TestRunner() {
...
-    this();
-    Preconditions.checkNotNull(taskClass);
-    configs.put(TaskConfig.TASK_CLASS(), taskClass.getName());
-    this.taskClass = taskClass;
-  }
-
-  /**
-   * Constructs a new {@link TestRunner} from following components
-   * @param app samza job implementing {@link StreamApplication}
+   * @param app samza job implementing {@link SamzaApplication}
```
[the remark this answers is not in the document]
prateekm: ^ Bump

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -102,7 +100,7 @@ private TestRunner() {
 
   /**
    * Constructs a new {@link TestRunner} from following components
-   * @param taskClass represent a class containing Samza job logic extending either {@link StreamTask} or {@link AsyncStreamTask}
+   * @param taskClass containing Samza job logic extending either {@link StreamTask} or {@link AsyncStreamTask}
    */
   private TestRunner(Class taskClass) {
```
prateekm: Is this method still required?

PR #655, opened 2018-09-24
SAMZA-1903: Samza-sql - Fix stream-table join to work with udfs

atoomula:


samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -250,7 +251,17 @@ private String dumpRelPlanForNode(RelNode relNode) {
         SqlExplainLevel.EXPPLAN_ATTRIBUTES);
   }
 
-  private SqlIOConfig resolveSourceConfig(RelNode relNode) {
+  private SqlIOConfig resolveSourceConfigForTable(RelNode relNode) {
+    if (relNode instanceof LogicalProject) {
+      return resolveSourceConfigForTable(((LogicalProject) relNode).getInput());
+    }
+
+    // For stream-table-table join, we will run into a join relNode. The output of stream-table join is a stream.
```
srinipunuru: This comment is not very clear.  Can you reword it a little bit. We are returning the sourceConfig for the table as null when the table is in the join rather than an output table, That's  because?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -263,8 +274,10 @@ private boolean isTable(RelNode relNode) {
     // NOTE: Any intermediate form of a join is always a stream. Eg: For the second level join of
     // stream-table-table join, the left side of the join is join output, which we always
     // assume to be a stream. The intermediate stream won't be an instance of EnumerableTableScan.
-    if (relNode instanceof EnumerableTableScan) {
-      return resolveSourceConfig(relNode).getTableDescriptor().isPresent();
+    // The join key(s) for the table could be an udf in which case the relNode would be LogicalProject.
+    if (relNode instanceof EnumerableTableScan || relNode instanceof LogicalProject) {
```
srinipunuru: If relNode is a enumerableTableScan, Shouldn't this result in infinite loop?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -250,7 +251,17 @@ private String dumpRelPlanForNode(RelNode relNode) {
...
+
+    // For stream-table-table join, we will run into a join relNode. The output of stream-table join is a stream.
+    // In that case, let's return null, as the source for relNode is not a table but a stream.
+    if (relNode.getInputs().size() > 1) {
+      return null;
+    }
+
     String sourceName = String.join(".", relNode.getTable().getQualifiedName());
     SqlIOConfig sourceConfig = ioResolver.fetchSourceInfo(sourceName);
     if (sourceConfig == null) {
```
srinipunuru: when is SourceConfig null?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: Done!

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: This is an unexpected error condition. We expect the source configs to be set.

PR #660, opened 2018-09-25
SAMZA-1832: Fix race condition between StreamProcessor and SamzaContainerListener

bharathkk:
The PR addresses the following issues

- We have few scenarios where there is a race condition between `SamzaContainerListener` and `StreamProcessor` that results in incorrect application status being propagated to client

> 1. Consider the scenario when the container runs into an exception in run loop and triggers shutdown sequence and at the same time the user triggers a stop on the `StreamProcessor`. The user request comes in and notices that the container is already shutting down and proceeds to shutdown the job coordinator which in turns shuts down successfully and the `processorListener.onStop()` is invoked. The container eventually invokes the `SamzaContainerListener` callback and updates the exception state after the `StreamProcessor` has finished shutting down.
> 2. Consider the scenario when the user invokes start on the stream processor followed by a stop. During the `StreamProcessor` shutdown sequence we can have different status propagated to the client based on when we receive the start notification from container. If we receive the notification prior to invoking `processListener.onStop()` as part of `StreamProcessor` shutdown sequence, we return `successfulFinish` while if the notification is received after, we return `unsuccessfulFinish`.

 
- `SamzaContainer` sets the status flag to failed as soon as it runs into an exception in the runloop. This is problematic because it is possible for `StreamProcessor` to assume the container has shutdown correctly even though container is still in the shutdown sequence. `StreamProcessor` uses `container.hasStopped()` to determine the shutdown status of the container. Internally, this method returns true if the status is either `FAILED` or `STOPPED`. 
- During rebalance phase, due to above mentioned bugs, it possible that we have a unclean shutdown but still proceed to wait for new job model in some cases and proceed to `StreamProcessor` shutdown in others. 
- Currently, we only propagate failures to `processorListener` when containerException is not null. It is possible for the samza container to take longer than `task.shutdown.ms` to shutdown in which case, we need to propagate a timeout exception to the `processorListener` as opposed assuming the shutdown was successful.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/util/StateTransitionUtil.java
+++ b/samza-core/src/main/java/org/apache/samza/util/StateTransitionUtil.java
@@ -44,7 +44,7 @@ public class StateTransitionUtil {
    * @return true if current state is one of the expected value and transition was successful; false otherwise
    */
   public static <T> boolean compareAndSet(AtomicReference<T> reference, Set<T> expectedValues, T update) {
-    while(true) {
+    while (true) {
       T currentValue = reference.get();
       if (expectedValues.contains(currentValue)) {
         if (reference.compareAndSet(currentValue, update)) {
@@ -67,13 +67,13 @@ public class StateTransitionUtil {
    * @return true if current state is not in the blacklisted values and transition was successful; false otherwise
    */
   public static <T> boolean compareNotInAndSet(AtomicReference<T> reference, Set<T> blacklistedValues, T update) {
-    while(true) {
+    while (true) {
       T currentValue = reference.get();
       if (blacklistedValues.contains(currentValue)) {
         return false;
       }
 
-      if(reference.compareAndSet(currentValue, update)) {
+      if (reference.compareAndSet(currentValue, update)) {
         return true;
       }
     }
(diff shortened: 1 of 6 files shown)


bharathkk:
@prateekm  @nickpan47 @shanthoosh please take a look.
I will update the JIRA and PR with an extensive summary. Meanwhile, it would be good to get early feedback on it.

samza-core/src/main/java/org/apache/samza/util/StateTransitionUtil.java
```diff
@@ -0,0 +1,121 @@
...
+   * Atomically sets the value to the given updated value if the black listed values does not contain current value.
+   *
+   * @param reference the atomic reference
+   * @param blacklistedValues set of blacklisted values
+   * @param update the new value
+   * @param <T> type of the atomic reference
+   *
+   * @return true if current state is not in the blacklisted values and transition was successful; false otherwise
+   */
+  public static <T> boolean compareNotInAndSet(AtomicReference<T> reference, Collection<T> blacklistedValues, T update) {
```
shanthoosh: Though it looks like we use atomic reference in StreamProcessor, we lock implicitly in StateTransitionUtil. From what I can understand, we've replaced synchronized lock with a busy spinning lock in this patch for performing state transition on a object.

My concern with the spinlock is that, while waiting to acquire a lock, it wastes the CPU cycles. Can you share the value-add and benefits that we get by replacing synchronized lock with the spinning lock. If we need to lock, why can't we use a explicit lock.

bharathkk:
Closing the PR since we have decided to pursue a simpler point fix w/ https://github.com/apache/samza/pull/673

PR #662, opened 2018-09-26
SAMZA-1868: Create new SamzaAmdmin for Kafka

sborya:
This Request is a copy of #647(got garbled). This PR  already addresses all the comments brought up in the other request.


samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -75,74 +76,59 @@
...
+  // adminClient is required for deleteBeforeMessages operation
+  private final AdminClient adminClient;
 
   private final boolean deleteCommittedMessages;
 
+  // used in TestRepartitionJoinWindowApp
   @VisibleForTesting
   public static volatile boolean deleteMessageCalled = false;
 
+  private final AtomicBoolean stoped = new AtomicBoolean(false);
```
dnishimura: Misspelled `stopped`

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdminUtilsScala.scala
```diff
@@ -0,0 +1,194 @@
...
+import org.apache.samza.config.ApplicationConfig.ApplicationMode
+import org.apache.samza.config.{ApplicationConfig, Config, KafkaConfig, StreamConfig}
+import org.apache.samza.system.SystemStreamMetadata.SystemStreamPartitionMetadata
+import org.apache.samza.system.{StreamSpec, SystemStreamMetadata, SystemStreamPartition}
+import org.apache.samza.util.ExponentialSleepStrategy
+import org.slf4j.{Logger, LoggerFactory}
+
+import scala.collection.JavaConverters._
+
+class KafkaSystemAdminUtilsScala(systemName: String) {
```
shanthoosh: Looks like systemName is used only for logging here(which can be obtained from the KafkaStreamSpec itself. `streamSpec.getSystemName`). I think it can be removed from constructor args (if theres not much value from it).

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,667 @@
...
+  private final KafkaSystemAdminUtilsScala kafkaAdminUtils;
+
+  // adminClient is required for deleteBeforeMessages operation
+  private final AdminClient adminClient;
+
+  private final boolean deleteCommittedMessages;
+
+  // used in TestRepartitionJoinWindowApp
+  @VisibleForTesting
+  public static volatile boolean deleteMessageCalled = false;
```
shanthoosh: Looks like this static constant is useful only for testing and not relevant with respect to the functionality of the API in itself.  Is it possible to replace this with a mock ?

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,667 @@
...
+            loop.done();
+            return allMetadata;
+          }
+        };
+
+    Map<String, SystemStreamMetadata> result = strategy.run(fetchMetadataOperation,
+        new AbstractFunction2<Exception, ExponentialSleepStrategy.RetryLoop, BoxedUnit>() {
+          @Override
+          public BoxedUnit apply(Exception exception, ExponentialSleepStrategy.RetryLoop loop) {
+            if (loop.sleepCount() < MAX_RETRIES_ON_EXCEPTION) {
```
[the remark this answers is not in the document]
shanthoosh: Then we need to add retries in createTopic call alone to make sure we return only after we know that a particular topic is created! Adding unnecessary retries everywhere makes it really hard to configure maximum timeouts for failure and makes the code  less readable.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -80,43 +82,11 @@ class KafkaSystemFactory extends SystemFactory with Logging {
...
-      coordinatorStreamProperties,
-      coordinatorStreamReplicationFactor,
-      timeout,
-      bufferSize,
-      clientId,
-      topicMetaInformation,
-      intermediateStreamProperties,
-      deleteCommittedMessages)
+      config,
+      KafkaConsumerConfig.ADMIN_CLIENT_ID_PREFIX);
```
shanthoosh: Would be better to inline this constant within the `SamzaKafkaSystemAdmin` implementation itself. Users cannot override this value through config currently and it's a fixed constant. I think having a `idPrefix` in the API signature is kind of unnecessary here with not clearly defined semantics. Please change if it makes sense.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -46,21 +43,25 @@ object KafkaSystemFactory extends Logging {
...
 
-    val kafkaSystemConsumer = new KafkaSystemConsumer(kafkaConsumer, systemName, config, clientId, metrics, new SystemClock)
-    info("Created samza system consumer %s" format  (kafkaSystemConsumer.toString))
+    val kafkaConsumer = KafkaSystemConsumer.getKafkaConsumerImpl(systemName, kafkaConsumerConfig)
+    info("Created kafka consumer for system %s, clientId %s: %s" format(systemName, clientIdPrefix, kafkaConsumer))
+
+    val clientId = kafkaConsumerConfig.getClientId
+    val kafkaSystemConsumer = new KafkaSystemConsumer(kafkaConsumer, systemName, config, clientId, metrics,
+      new SystemClock)
+    info("Created samza system consumer %s" format (kafkaSystemConsumer.toString))
```
shanthoosh: info("Created kafka system consumer for system: %s with config:  %s" format(system, config))

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -46,21 +43,25 @@ object KafkaSystemFactory extends Logging {
 class KafkaSystemFactory extends SystemFactory with Logging {
 
   def getConsumer(systemName: String, config: Config, registry: MetricsRegistry): SystemConsumer = {
-    val clientId = KafkaConsumerConfig.getConsumerClientId( config)
+    val clientIdPrefix = KafkaConsumerConfig.CONSUMER_CLIENT_ID_PREFIX;
     val metrics = new KafkaSystemConsumerMetrics(systemName, registry)
 
-    val kafkaConsumer = KafkaSystemConsumer.getKafkaConsumerImpl(systemName, clientId, config)
-    info("Created kafka consumer for system %s, clientId %s: %s" format (systemName, clientId, kafkaConsumer))
+    val kafkaConsumerConfig = KafkaConsumerConfig.getKafkaSystemConsumerConfig(config, systemName, clientIdPrefix);
```
shanthoosh: We pass in a constant  `KafkaConsumerConfig.CONSUMER_CLIENT_ID_PREFIX` defined in `KafkaConsumerConfig` to `KafkaConsumerConfig.getKafkaSystemConsumer` as `idPrefix`.  Why can't we inline the constant within the class itself?  Is there any scenarios that would need this?

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,666 @@
...
+import scala.runtime.AbstractFunction0;
+import scala.runtime.AbstractFunction1;
+import scala.runtime.AbstractFunction2;
+import scala.runtime.BoxedUnit;
+
+import static org.apache.samza.config.KafkaConsumerConfig.*;
+import static org.apache.samza.system.kafka.KafkaSystemDescriptor.*;
+
+
+public class SamzaKafkaSystemAdmin<K, V> implements ExtendedSystemAdmin {
```
shanthoosh: Would be better to name  this class simply as `KafkaSystemAdmin` itself (or some other meaningful name).

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -108,18 +107,15 @@ public KafkaSystemConsumer(Consumer<K, V> kafkaConsumer, String systemName, Conf
...
+   * @param kafkaConsumerConfig
+   * @return kafka consumer
    */
-  public static KafkaConsumer<byte[], byte[]> getKafkaConsumerImpl(String systemName, String clientId, Config config) {
+  public static KafkaConsumer<byte[], byte[]> getKafkaConsumerImpl(String systemName,
+      HashMap kafkaConsumerConfig) {
 
-    // extract kafka client configs
-    KafkaConsumerConfig consumerConfig = KafkaConsumerConfig.getKafkaSystemConsumerConfig(config, systemName, clientId);
+    LOG.info("KafkaClient properties for systemName {}: {}", systemName, kafkaConsumerConfig);
```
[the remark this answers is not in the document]
sborya: Changed the log message.
I think the log-level concern is valid. But in case of issues with the Kafka consumption, this will be an important part of the debugging. I would suggest to leave it as info level for now and change it to debug when the code gets some millage.

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,667 @@
...
+    LOG.info(String.format("Creating SamzaKafkaSystemAdmin for system %s, idPrefix %s", systemName, idPrefix));
+
+    return new SamzaKafkaSystemAdmin(systemName, metadataConsumerSupplier, zkConnectSupplier, adminClientSupplier,
+        topicMetaInformation, intermediateStreamProperties, coordinatorStreamProperties,
+        coordinatorStreamReplicationFactor, deleteCommittedMessages);
+  }
+
+  @Override
+  public Map<String, SystemStreamMetadata> getSystemStreamPartitionCounts(Set<String> streamNames, long cacheTTL) {
+    // This optimization omits actual metadata for performance. Instead, we inject a dummy for all partitions.
```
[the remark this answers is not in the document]
sborya: The problem is the API. It should've returned something like Map<String, Integer>.
But for now I can add a "non implemented exception" and javadoc.

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
```diff
@@ -0,0 +1,667 @@
...
+            Map<String, SystemStreamMetadata> allMetadata = new HashMap<>();
+
+            streamNames.forEach(streamName -> {
+              Map<Partition, SystemStreamMetadata.SystemStreamPartitionMetadata> partitionMetadata = new HashMap<>();
+
+              List<PartitionInfo> partitionInfos = metadataConsumer.partitionsFor(streamName);
+              LOG.debug("Stream {} has partitions {}", streamName, partitionInfos);
+
+              partitionInfos.forEach(partitionInfo -> {
+                partitionMetadata.put(new Partition(partitionInfo.partition()), dummySspm);
```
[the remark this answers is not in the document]
sborya: see above.

samza-kafka/src/main/java/org/apache/samza/system/kafka/SamzaKafkaSystemAdmin.java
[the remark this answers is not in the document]
sborya: It is a good suggestion. I will consider it as a part of the Samza ticket I've created for it.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
sborya: The idea is to provide ability to have a "custom" naming for the consumer, and not totally dependent on config.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
sborya: Again, the idea is to have an ability to customize the config for a specific kafka consumer instantiation.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+    OffsetsMaps topicPartitionsMetadata;
+    topicPartitionsMetadata = fetchTopicPartitionsMetadata(topicPartitions);
+
+    Map<SystemStreamPartition, SystemStreamMetadata.SystemStreamPartitionMetadata> sspToSSPMetadata = new HashMap<>();
+    for (SystemStreamPartition ssp : ssps) {
+      String oldestOffset = topicPartitionsMetadata.getOldestOffsets().get(ssp);
+      String newestOffset = topicPartitionsMetadata.getNewestOffsets().get(ssp);
+      String upcomingOffset = topicPartitionsMetadata.getUpcomingOffsets().get(ssp);
+
+      sspToSSPMetadata.put(ssp,
```
bharathkk: Are we guaranteed that one of these will be non-null?  I noticed we have that check internally.

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemAdmin.scala
```diff
@@ -284,30 +353,6 @@ class TestKafkaSystemAdmin {
     assertEquals(3, partitionMetadata.replicas.size)
   }
 
-  class KafkaSystemAdminWithTopicMetadataError extends KafkaSystemAdmin(SYSTEM, brokerList, () => ZkUtils(zkConnect, 6000, 6000, zkSecure)) {
```
nickpan47: It seems to me that the negative test cases are useful. Can we include similar negative test cases for the new KafkaSystemAdmin as well?

samza-test/src/test/java/org/apache/samza/test/operator/TestRepartitionJoinWindowApp.java
```diff
@@ -38,7 +38,7 @@
 import org.junit.Assert;
 import org.junit.Test;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
```
nickpan47: Please do not use *

samza-test/src/test/scala/org/apache/samza/test/harness/AbstractIntegrationTestHarness.scala
```diff
@@ -17,50 +17,71 @@
...
+import kafka.utils.TestUtils
+import org.apache.samza.config.{JobConfig, KafkaConsumerConfig, MapConfig}
 import org.apache.samza.system.kafka.KafkaSystemAdmin
 
 /**
- * LinkedIn integration test harness for Kafka
- * This is simply a copy of open source code. We do this because java does not support trait and we are making it an
- * abstract class so that user's java test class can extend it.
- */
+  * LinkedIn integration test harness for Kafka
```
nickpan47: Please remove LinkedIn here, since this is now open source code.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+  private static final double DEFAULT_EXPONENTIAL_SLEEP_BACK_OFF_MULTIPLIER = 2.0;
+  private static final long DEFAULT_EXPONENTIAL_SLEEP_INITIAL_DELAY_MS = 500;
+  private static final long DEFAULT_EXPONENTIAL_SLEEP_MAX_DELAY_MS = 10000;
+  private static final int MAX_RETRIES_ON_EXCEPTION = 5;
+
+  private final String systemName;
+  private final Consumer<K, V> metadataConsumer;
+
+  private final Supplier<ZkUtils> connectZk;
+  private final Supplier<AdminClient> connectAdminClient;
```
bharathkk: nit: looks like its not used anywhere outside constructor.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+      ChangelogInfo changelogInfo =
+          new ChangelogInfo(replicationFactor, kafkaConfig.getChangelogKafkaProperties(storeName));
+      LOG.info(String.format("Creating topic meta information for topic: %s with replication factor: %s", topicName,
+          replicationFactor));
+      topicMetaInformation.put(topicName, changelogInfo);
+    }
+
+    // special flag to allow/enforce deleting of committed messages
+    SystemConfig systemConfig = new SystemConfig(config);
+    Option<String> deleteCommittedMessagesOpt = systemConfig.deleteCommittedMessages(systemName);
```
bharathkk: nit: would be nicer to combine them and get rid of the intermediate variable

```
boolean deleteCommittedMessages = systemConfig.deleteCommittedMessages(systemName)
   .map(Boolean::valueOf)
   .orElse(false);
```

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+
+    // kafka.admin.AdminUtils requires zkConnect
+    // this will change after we move to the new org.apache..AdminClient
+    String zkConnectStr = config.get(String.format(CONSUMER_CONFIGS_CONFIG_KEY, systemName, ZOOKEEPER_CONNECT));
+    if (zkConnectStr == null) {
+      throw new SamzaException("Missing zookeeper.connect config for admin for system " + systemName);
+    }
+    props.put(ZOOKEEPER_CONNECT, zkConnectStr);
+
+    Supplier<AdminClient> adminClientSupplier = () -> {
```
bharathkk: nit: would prefer them to be consistent with line 200 by replacing it with expression lambda instead of statement.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+  }
+
+  /**
+   * Helper method to create KafkaSystemAdmin
+   * @param systemName system name
+   * @param config application config
+   * @param idPrefix will be used to generate clientId for the metadata consumer
+   * @return new KafkaSystemAdmin object
+   */
+  public static KafkaSystemAdmin getKafkaSystemAdmin(final String systemName, final Config config,
```
bharathkk: Can we update the signature to return KafkaSystemAdmin<byte[], byte[]> to get rid of unchecked warnings?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -0,0 +1,668 @@
...
+    // this will change after we move to the new org.apache..AdminClient
+    String zkConnectStr = config.get(String.format(CONSUMER_CONFIGS_CONFIG_KEY, systemName, ZOOKEEPER_CONNECT));
+    if (zkConnectStr == null) {
+      throw new SamzaException("Missing zookeeper.connect config for admin for system " + systemName);
+    }
+    props.put(ZOOKEEPER_CONNECT, zkConnectStr);
+
+    adminClient = AdminClient.create(props);
+
+    connectZk = () -> {
```
prateekm: Why is this a lambda? What does this do?

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemAdmin.scala
sborya: Well, if we use new kafka consumer it should take care of all recoverable errors. Do we still need these tests?

samza-test/src/test/java/org/apache/samza/test/operator/TestRepartitionJoinWindowApp.java
sborya: that's not me :)  Intelij reformat. Fixed.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemAdminJava.java
```diff
@@ -176,16 +300,21 @@ public void testValidateStreamWrongName() {
     systemAdmin().validateStream(spec2);
   }
 
-  @Test
+  //@Test TODO - see if it can be fixed
```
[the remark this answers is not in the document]
sborya: done.

samza-kafka/src/main/java/org/apache/samza/config/KafkaConsumerConfig.java
```diff
@@ -40,35 +42,36 @@
 
   public static final Logger LOG = LoggerFactory.getLogger(KafkaConsumerConfig.class);
 
-  static final String PRODUCER_CLIENT_ID_PREFIX = "kafka-producer";
-  static final String CONSUMER_CLIENT_ID_PREFIX = "kafka-consumer";
-  static final String ADMIN_CLIENT_ID_PREFIX = "samza-admin";
+  public static final String PRODUCER_CLIENT_ID_PREFIX = "kafka-producer";
```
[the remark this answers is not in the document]
sborya: I wanted to keep all of them in the same file, but I will move it out.

samza-kafka/src/main/java/org/apache/samza/config/KafkaConsumerConfig.java
```diff
@@ -113,48 +116,42 @@ public static KafkaConsumerConfig getKafkaSystemConsumerConfig(Config config, St
...
+    if (clientId == null) {
+      throw new SamzaException("client Id is not set for consumer for system=" + systemName);
     }
-    String jobName = (String) jobNameOption.get();
+    return clientId;
+  }
 
-    String jobId = jobConfig.getJobId();
+  // group id should be unique per job
+  static String createConsumerGroupId(Config config) {
```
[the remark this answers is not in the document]
sborya: Kafka consumer expects both.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -0,0 +1,668 @@
...
+    // Throw exception if start is called after stop
+    if (stopped.get()) {
+      throw new SamzaException("SamzaKafkaAdmin.start() is called after stop()");
+    }
+
+    // Plese note. There is slight inconsistency in the use of this class.
+    // Some of the functinality of this class may actually be used BEFORE start() is called.
+    // The SamzaContainer gets metadata (using this class) in SamzaContainer.apply, but this "start" actually gets called in SamzaContainer.run.
+    // Also we assume that start is called only once.
+    if (metadataConsumer == null) {
```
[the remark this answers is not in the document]
sborya: I've added a check in constructor. So I can remove it here.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdminUtilsScala.scala
```diff
@@ -0,0 +1,194 @@
...
+/**
+  * A helper class that is used to construct the changelog stream specific information
+  *
+  * @param replicationFactor The number of replicas for the changelog stream
+  * @param kafkaProps The kafka specific properties that need to be used for changelog stream creation
+  */
+case class ChangelogInfo(var replicationFactor: Int, var kafkaProps: Properties)
+
+
+// TODO move to org.apache.kafka.clients.admin.AdminClien from the kafka.admin.AdminClient
```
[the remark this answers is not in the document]
sborya: This class was created for Scala methods which are not converted to java yet. We may not need to convert them after we upgrade to the new JavaAdminClass.
As discussed elsewhere we will also consider creating a separate kafka util class for admin operations.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -46,21 +45,25 @@ object KafkaSystemFactory extends Logging {
...
+      new SystemClock)
+    info("Created samza system consumer %s with config %s" format (kafkaSystemConsumer, config))
 
     kafkaSystemConsumer
   }
 
   def getProducer(systemName: String, config: Config, registry: MetricsRegistry): SystemProducer = {
-    val clientId = KafkaConsumerConfig.getProducerClientId(config)
     val injectedProps = KafkaSystemFactory.getInjectedProducerProperties(systemName, config)
+    val clientId = KafkaConsumerConfig.createClientId(KafkaConsumerConfig.PRODUCER_CLIENT_ID_PREFIX, config);
```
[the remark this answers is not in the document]
sborya: seemed like a lot of people didn't like this idea, so I moved it out.

samza-kafka/src/main/java/org/apache/samza/config/KafkaConsumerConfig.java
```diff
@@ -40,35 +42,36 @@
...
    */
-  public static KafkaConsumerConfig getKafkaSystemConsumerConfig(Config config, String systemName, String clientId) {
+  public static KafkaConsumerConfig getKafkaSystemConsumerConfig(Config config, String systemName, String idPrefix) {
 
     Config subConf = config.subset(String.format("systems.%s.consumer.", systemName), true);
 
-    //Kafka client configuration
-    String groupId = getConsumerGroupId(config);
+    final String groupId = createConsumerGroupId(config);
+    final String clientId = createClientId(idPrefix, config);
```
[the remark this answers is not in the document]
sborya: Ok, seems like the idea of passing prefix instead of clientId is not very popular. I will change it to clientId created by the caller.

samza-kafka/src/main/java/org/apache/samza/config/KafkaConsumerConfig.java
```diff
@@ -40,35 +42,39 @@
...
    * @return KafkaConsumerConfig
    */
   public static KafkaConsumerConfig getKafkaSystemConsumerConfig(Config config, String systemName, String clientId) {
 
     Config subConf = config.subset(String.format("systems.%s.consumer.", systemName), true);
 
-    //Kafka client configuration
-    String groupId = getConsumerGroupId(config);
+    final String groupId = createConsumerGroupId(config);
+    //final String clientId = KafkaConsumerConfig.createClientId(clientIdPrefix, config);
```
prateekm: Remove unused code.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -0,0 +1,680 @@
...
+
+    // KafkaConsumer for metadata access
+    Supplier<Consumer<byte[], byte[]>> metadataConsumerSupplier =
+        () -> KafkaSystemConsumer.getKafkaConsumerImpl(systemName, consumerConfig);
+
+    KafkaConfig kafkaConfig = new KafkaConfig(config);
+    Properties coordinatorStreamProperties = KafkaSystemAdminUtilsScala.getCoordinatorTopicProperties(kafkaConfig);
+    int coordinatorStreamReplicationFactor = Integer.valueOf(kafkaConfig.getCoordinatorReplicationFactor());
+
+    Map<String, String> storeToChangelog =
```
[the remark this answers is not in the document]
prateekm: I think what @nickpan47 is asking is where was this being done in the existing admin implementation. Constructor for this class doesn't seem like the right place to do this.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -0,0 +1,662 @@
...
+  @Override
+  public void start() {
+    // Plese note. There is slight inconsistency in the use of this class.
+    // Some of the functionality of this class may actually be used BEFORE start() is called.
+    // The SamzaContainer gets metadata (using this class) in SamzaContainer.apply,
+    // but this "start" actually gets called in SamzaContainer.run.
+    // review this usage (SAMZA-1888)
+
+    // Throw exception if start is called after stop
+    if (stopped.get()) {
```
prateekm: Don't need to start the metadataConsumer/adminClient?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemAdmin.scala
```diff
@@ -1,608 +0,0 @@
...
-import org.apache.kafka.common.TopicPartition
-import org.apache.samza.system.SystemStreamMetadata.SystemStreamPartitionMetadata
-import org.apache.samza.system._
-import org.apache.samza.util.{ClientUtilTopicMetadataStore, ExponentialSleepStrategy, KafkaUtil, Logging}
-import org.apache.samza.{Partition, SamzaException}
-
-import scala.collection.JavaConverters._
-
-
-object KafkaSystemAdmin extends Logging {
```
prateekm: As mentioned in the other (KafkaConsumer) PR, we want to move the old implementations to a separate class (OldKafkaSystemAdmin/org.apache.samza.system.kafka.old.KafkaSystemAdmin) instead of deleting them. This gives users an option to fall back to the older implementation in case of any bugs. Any reason that won't work?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -21,46 +21,52 @@ package org.apache.samza.system.kafka
...
 }
 
 class KafkaSystemFactory extends SystemFactory with Logging {
 
   def getConsumer(systemName: String, config: Config, registry: MetricsRegistry): SystemConsumer = {
-    val clientId = KafkaConsumerConfig.getConsumerClientId( config)
     val metrics = new KafkaSystemConsumerMetrics(systemName, registry)
 
-    val kafkaConsumer = KafkaSystemConsumer.getKafkaConsumerImpl(systemName, clientId, config)
+    val clientId = KafkaConsumerConfig.createClientId("kafka-consumer", config);
```
prateekm: Use the constants defined above or delete them.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -21,46 +21,52 @@ package org.apache.samza.system.kafka
...
+    val kafkaConsumerConfig = KafkaConsumerConfig.getKafkaSystemConsumerConfig(config, systemName, clientId);
+
+    val kafkaConsumer = KafkaSystemConsumer.createKafkaConsumerImpl[Array[Byte], Array[Byte]](systemName, kafkaConsumerConfig)
     info("Created kafka consumer for system %s, clientId %s: %s" format (systemName, clientId, kafkaConsumer))
 
-    val kafkaSystemConsumer = new KafkaSystemConsumer(kafkaConsumer, systemName, config, clientId, metrics, new SystemClock)
-    info("Created samza system consumer %s" format  (kafkaSystemConsumer.toString))
+    val kafkaSystemConsumer = new KafkaSystemConsumer(kafkaConsumer, systemName, config, clientId, metrics,
+      new SystemClock)
+    info("Created samza system consumer %s with config %s" format(kafkaSystemConsumer, config))
```
prateekm: Use same format as the log message above.

samza-kafka/src/main/java/org/apache/samza/config/KafkaConsumerConfig.java
```diff
@@ -40,35 +42,39 @@
...
   public static final Logger LOG = LoggerFactory.getLogger(KafkaConsumerConfig.class);
 
-  static final String PRODUCER_CLIENT_ID_PREFIX = "kafka-producer";
-  static final String CONSUMER_CLIENT_ID_PREFIX = "kafka-consumer";
-  static final String ADMIN_CLIENT_ID_PREFIX = "samza-admin";
+  public static final String ZOOKEEPER_CONNECT = "zookeeper.connect";
 
+  public static final String CONSUMER_CONFIGS_CONFIG_KEY = "systems.%s.consumer.%s";
+  public static final String PRODUCER_BOOTSTRAP_SERVERS_CONFIG_KEY = "systems.%s.producer.bootstrap.servers";
+  public static final String PRODUCER_CONFIGS_CONFIG_KEY = "systems.%s.producer.%s";
```
prateekm: Should these be in KafkaConsumerConfig? Is there a KafkaConfig class/base class we can put these in?

samza-kafka/src/main/java/org/apache/samza/config/KafkaConsumerConfig.java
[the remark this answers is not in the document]
prateekm: Any reason we can't set the groupId == clientId?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -0,0 +1,662 @@
...
+
+public class KafkaSystemAdmin implements ExtendedSystemAdmin {
+  private static final Logger LOG = LoggerFactory.getLogger(KafkaSystemAdmin.class);
+
+  // Default exponential sleep strategy values
+  protected static final double DEFAULT_EXPONENTIAL_SLEEP_BACK_OFF_MULTIPLIER = 2.0;
+  protected static final long DEFAULT_EXPONENTIAL_SLEEP_INITIAL_DELAY_MS = 500;
+  protected static final long DEFAULT_EXPONENTIAL_SLEEP_MAX_DELAY_MS = 10000;
+  protected static final int MAX_RETRIES_ON_EXCEPTION = 5;
+  public static final int DEFAULT_REPL_FACTOR = 2;
```
prateekm: Why public? Also, "DEFAULT_REPLICATION_FACTOR".

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -0,0 +1,662 @@
...
+    // this will change after we move to the new org.apache..AdminClient
+    String zkConnect = config.get(String.format(CONSUMER_CONFIGS_CONFIG_KEY, systemName, ZOOKEEPER_CONNECT));
+    if (StringUtils.isBlank(zkConnect)) {
+      throw new SamzaException("Missing zookeeper.connect config for admin for system " + systemName);
+    }
+    props.put(ZOOKEEPER_CONNECT, zkConnect);
+
+    adminClient = AdminClient.create(props);
+
+    getZkConnection = () -> {
```
prateekm: I think i'll be cleaner to inline this or make this a private helper method (`createZkUtils`)

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -0,0 +1,662 @@
...
+
+    // special flag to allow/enforce deleting of committed messages
+    SystemConfig systemConfig = new SystemConfig(config);
+    this.deleteCommittedMessages = systemConfig.deleteCommittedMessages(systemName);
+
+    intermediateStreamProperties =
+        JavaConverters.mapAsJavaMapConverter(KafkaSystemAdminUtilsScala.getIntermediateStreamProperties(config))
+            .asJava();
+
+    LOG.info(String.format("Creating KafkaSystemAdmin for system %s", systemName));
```
prateekm: s/Creating/Created

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -100,26 +99,26 @@ public KafkaSystemConsumer(Consumer<K, V> kafkaConsumer, String systemName, Conf
     messageSink = new KafkaConsumerMessageSink();
 
     // Create the proxy to do the actual message reading.
-    String metricName = String.format("%s", systemName);
+    String metricName = generateMetricsName();
     proxy = new KafkaConsumerProxy(kafkaConsumer, systemName, clientId, messageSink, metrics, metricName);
     LOG.info("{}: Created KafkaConsumerProxy {} ", this, proxy);
   }
 
+  protected String generateMetricsName() {
```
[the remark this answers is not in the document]
sborya: I can make it completely deterministic, based on configs.

PR #664, opened 2018-09-26
SAMZA-1914: fix out of range starting offset in EH consumer

lhaiesp:


samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
```diff
@@ -65,19 +65,12 @@ public EventHubSystemAdmin(String systemName, EventHubConfig eventHubConfig,
...
-    return String.valueOf(Long.parseLong(currentOffset) + 1);
-  }
-
   @Override
   public Map<SystemStreamPartition, String> getOffsetsAfter(Map<SystemStreamPartition, String> offsets) {
-    Map<SystemStreamPartition, String> results = new HashMap<>();
-
-    offsets.forEach((partition, offset) -> results.put(partition, getNextOffset(offset)));
-    return results;
+    // In EventHubSystemConsumer#initializeEventHubsManagers, we exclude the offset that we specify. i.e.
```
dxichen: We only exclude the offset specified when the offset is set as the END_OF_STREAM constant. https://github.com/apache/samza/blob/032a16079db8cbb1caf00117b985f7575d2c823a/samza-azure/src/main/java/org/apache/samza/system/eventhub/consumer/EventHubSystemConsumer.java#L285

dxichen: Synced with @lhaiesp offline, since `getOffsetsAfter` is only used for getting start offsets on restart (`StorageManagerUtil`, `OffsetManager`), it will have the same behavior as `START_OF_STREAM` in all current uses `getOffsetAfter`.

PR #665, opened 2018-09-26
 SAMZA-1889: Extend ExecutionPlanner to support Stream-Table Joins

ahmedahamid:
Extend ExecutionPlanner to verify agreement in partition count among the
stream(s) behind Tables — including side-input streams — and other streams
participating in Stream-Table Joins.


ahmedahamid:
@vjagadish1989 @bharathkk Here is the 3rd and last PR of this series.
@nickpan47 Let's follow up on our discussion regarding verification vs assignment at your earliest convenience.

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
```diff
@@ -175,6 +184,14 @@ StreamEdge getOrCreateStreamEdge(StreamSpec streamSpec) {
     return getOrCreateStreamEdge(streamSpec, false);
   }
 
+  /**
+   * Returns the {@link ApplicationDescriptorImpl} of this graph.
```
ahmedahamid: nit: Return

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
```diff
@@ -202,6 +219,14 @@ StreamEdge getStreamEdge(String streamId) {
     return Collections.unmodifiableSet(inputStreams);
   }
 
+  /**
+   * Returns the side-input streams in the graph
```
ahmedahamid: nit: Return

samza-core/src/main/java/org/apache/samza/execution/OperatorSpecGraphAnalyzer.java
```diff
@@ -39,22 +43,27 @@
...
+    // — which are terminal OperatorSpecs — and all StreamTableJoinOperatorSpecs referencing the same TableSpec.
+    Function<OperatorSpec, Collection<OperatorSpec>> getNextOpSpecs = getCustomGetNextOpSpecs(inputOperatorSpecs);
 
     // Traverse graph starting from every input operator spec, observing connectivity between input operator specs
     // and Join operator specs.
     for (InputOperatorSpec inputOpSpec : inputOperatorSpecs) {
       // Observe all join operator specs reachable from this input operator spec.
-      JoinOperatorSpecVisitor joinOperatorSpecVisitor = new JoinOperatorSpecVisitor();
-      traverse(inputOpSpec, joinOperatorSpecVisitor, opSpec -> opSpec.getRegisteredOperatorSpecs());
+      JoinVisitor joinOperatorSpecVisitor = new JoinVisitor();
```
ahmedahamid: Update variable name

ahmedahamid:
@nickpan47 I separated the logic for assignment and verification of partition counts as suggested.

One concern I have now, however, is about the utility/usefulness of `IntermediateStreamManager` as an independent class.

1. With my change, `IntermediateStreamManager` cannot operate independently from the `ExecutionPlanner` since it relies on it for traversing the operator graph and grouping conjoined edges.

2. If we have plans to use `IntermediateStreamManager` independently, then our only alternative would be to have it perform the traversal/grouping work currently done in the `ExecutionPlanner` redundantly. In addition to the computational overhead, this would entail promoting the stream grouping class, `StreamSet`, into a package-level class — an option I would not prefer because this class has no way of maintaining its invariant, i.e. the correct categorization of a group of streams, e.g. all vs some vs no partitions set, since `StreamEdge`s are mutable.

3. I also dislike the idea of having `calculatePartitions()` expect both a `JobGraph` _and_ a list of conjoined stream edges since there seems to be quite a bit of overlap between the two. On the other hand, I wouldn't want to reconstruct these groups in the `IntermediateStreamManager` redundantly, unless it could be used independently from the `ExecutionPlanner`.

PR #667, opened 2018-09-27
 Updated doc for Kinesis Connector

atoomula:


docs/learn/documentation/versioned/index.html
```diff
@@ -53,7 +53,7 @@ <h4>Connectors</h4>
   <li><a href="jobs/configuration.html">Apache Kafka</a></li>
   <li><a href="jobs/packaging.html">Apache Hadoop</a></li>
   <li><a href="jobs/yarn-jobs.html">Azure EventHubs</a></li>
-  <li><a href="jobs/logging.html">AWS Kinesis</a></li>
+  <li><a href="aws/kinesis.html">AWS Kinesis</a></li>
```
dnishimura: @vjagadish1989 should we move the connectors to a `connectors` subdir? Not sure how it'll affect the versioning thought. Kafka doesn't have its own page yet, so that's why I'm asking.

vjagadish1989: yup, agreed! as long as the file is there, i can rename it myself in a separate pass and fix things

PR #668, opened 2018-09-27
SAMZA-1919; Add KafkaSystemDescriptor constructor to allow arbitrary factory class

lindong28:
Currently KafkaSystemDescriptor is hardcoded to use org.apache.samza.system.kafka.KafkaSystemFactory as the system factory for instantiating producer/consumer/admin. Users may want to use most methods in KafkaSystemDescriptor but with a different factory class. We can meet such use-case by adding a constructor that allows user to specify the factory class name.


lindong28:
@prateekm Would you have time to review this patch? Thanks!

PR #555, opened 2018-06-15
SAMZA-1751: Refactored metrics for table API

weisong44:
Refactored metrics for table API
 - Added TableMetricsUtil that encapsulates required parameters, maintains naming consistency and simplifies metrics creation API for tables.
 - Added metrics to local table
 - Maintained consistency between local, remote and caching table

PR #672, opened 2018-09-28
SAMZA-1714: Creating shared context factory for shared context objects

cameronlee314:
<s>This includes changes in https://github.com/apache/samza/pull/626.</s>
Update: PR #626 has been merged, so the diff here should no longer show those changes.


samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -133,9 +121,6 @@ public final void window(MessageCollector collector, TaskCoordinator coordinator
 
   @Override
   public void close() throws Exception {
-    if (this.contextManager != null) {
```
prateekm: Where is ApplicationTaskContext started and stopped? In TaskInstance?

samza-core/src/main/java/org/apache/samza/util/EmbeddedTaggedRateLimiter.java
```diff
@@ -106,16 +103,14 @@ public int acquire(int numberOfCredit, long timeout, TimeUnit unit) {
...
         .map(e -> {
             String tag = e.getKey();
-            int effectiveRate = e.getValue();
-            if (taskContext != null) {
-              effectiveRate /= taskContext.getSamzaContainerContext().taskNames.size();
-              LOGGER.info(String.format("Effective rate limit for task %s and tag %s is %d",
-                  taskContext.getTaskName(), tag, effectiveRate));
-            }
+            int effectiveRate =
+                e.getValue() / context.getContainerContext().getContainerModel().getTasks().keySet().size();
```
prateekm: Minor: Maybe extract local variable with the size. Will be more understandable.

prateekm: Same for taskName below.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -1004,6 +1021,7 @@ class SamzaContainer(
     info("Initializing stream tasks.")
 
     taskInstances.values.foreach(_.initTask)
+    taskInstances.values.foreach(_.startContext)
```
prateekm: Maybe move the startContext/stopContext inside initTask and shutdownTask instead of exposing as a new API?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -1128,6 +1145,7 @@ class SamzaContainer(
     }
 
     taskInstances.values.foreach(_.shutdownTask)
+    taskInstances.values.foreach(_.shutdownContext)
```
prateekm: Might be nicer if the relationship b/w init+close and context start+stop is symmetric. I.e., Either the context is started before and stopped after the task init and close respectively, or vice versa.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -357,4 +374,30 @@ class TaskInstance(
...
+        if (storageManager != null && storageManager.getStore(storeName).isDefined) {
+          storageManager.getStore(storeName).get.asInstanceOf[KeyValueStore[_, _]]
+        } else if (sideInputStorageManager != null && sideInputStorageManager.getStore(storeName) != null) {
+          sideInputStorageManager.getStore(storeName).asInstanceOf[KeyValueStore[_, _]]
+        } else {
+          null
+        }
+      })
+    val scheduler = new CallbackSchedulerImpl(epochTimeScheduler)
+    val taskContext = new TaskContextImpl(taskModel,
```
prateekm: Prefer multiple params per line instead of param per line.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -59,27 +58,29 @@ class TaskInstance(
...
-  val context = new TaskContextImpl(taskName, metrics, containerContext, systemStreamPartitions.asJava, offsetManager,
-                                    kvStoreSupplier, tableManager, jobModel, streamMetadataCache, timerExecutor)
+  val epochTimeScheduler: EpochTimeScheduler = EpochTimeScheduler.create(timerExecutor)
+  val (taskContext: org.apache.samza.context.TaskContextImpl,
+    applicationTaskContextOption: Option[ApplicationTaskContext]) =
+    buildTaskContexts(epochTimeScheduler)
+  val context: Context = new ContextImpl(jobContext,
+    containerContext,
+    taskContext,
+    applicationContainerContext.orNull,
```
prateekm: Minor: Maybe rename to applicationContainerContextOption

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -357,4 +374,30 @@ class TaskInstance(
 
     startingOffset
   }
+
+  private def buildTaskContexts(epochTimeScheduler: EpochTimeScheduler):
```
prateekm: Minor: Would it be possible to refactor this into separate methods instead of a multi-value return? Prefer the former if it's simple to do, but otherwise this is fine.

samza-core/src/test/java/org/apache/samza/context/TestContextImpl.java
```diff
@@ -63,11 +67,28 @@ public void testGetMissingApplicationTaskContext() {
...
+  }
+
+  private static Context buildWithApplicationContext(ApplicationContainerContext applicationContainerContext,
+      ApplicationTaskContext applicationTaskContext) {
+    return new ContextImpl(mock(JobContext.class),
+        mock(ContainerContext.class),
+        mock(TaskContext.class),
+        applicationContainerContext,
+        applicationTaskContext);
+=======
```
prateekm: Let's resolve this merge conflict.

samza-core/src/main/java/org/apache/samza/application/ApplicationDescriptorImpl.java
```diff
@@ -118,12 +125,21 @@ public S withMetricsReporterFactories(Map<String, MetricsReporterFactory> report
...
 
   /**
-   * Get the {@link ContextManager} associated with this application
-   *
-   * @return the {@link ContextManager} for this application
+   * @return {@link ApplicationContainerContextFactory} if application specified it; empty otherwise
+   */
+  public Optional<ApplicationContainerContextFactory<ApplicationContainerContext>> getApplicationContainerContextFactory() {
+    //noinspection unchecked; ok because all context types are at least ApplicationContainerContext
+    return Optional.ofNullable(
```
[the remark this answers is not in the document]
cameronlee314: Using Optional as a field type has always been something I've been unsure of. I like Optional as a field type since it forces the empty check, but a couple points against it is that it is still possible to set it to null and the original intention was for it to just be for return types.
Since you mention it though, I'm good with using it as a field type.

samza-core/src/main/java/org/apache/samza/context/JobContextImpl.java
```diff
@@ -18,27 +18,44 @@
...
+   */
+  private JobContextImpl(Config config, String jobName, String jobId) {
     this.config = config;
     this.jobName = jobName;
     this.jobId = jobId;
   }
 
+  public static JobContextImpl fromConfigWithDefaults(Config config) {
+    JobConfig jobConfig = new JobConfig(config);
+    String jobName = jobConfig.getName().isDefined() ? jobConfig.getName().get() : null;
```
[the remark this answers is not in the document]
cameronlee314: Good to hear. Since the JobConfig API returns Option, it seemed optional. I'll change it.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
```diff
@@ -405,9 +399,4 @@ private static void computeOutputToInput(SystemStream input, OperatorSpec opSpec
   private boolean hasIntermediateStreams(OperatorSpecGraph specGraph) {
     return !Collections.disjoint(specGraph.getInputOperators().keySet(), specGraph.getOutputStreams().keySet());
   }
-
-  private static MetricsRegistry getMetricsRegistry(TaskContext context) {
```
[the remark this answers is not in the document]
cameronlee314: I'm not sure how in the past how containerContext could be null (maybe for unit tests), but with the new design, containerContext is never null. Therefore, the fallback is unnecessary.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -173,27 +191,42 @@ public StreamProcessor(Config config, Map<String, MetricsReporter> customMetrics
...
       ProcessorLifecycleListener processorListener, JobCoordinator jobCoordinator) {
-    this(config, customMetricsReporters, taskFactory, sp -> processorListener, jobCoordinator);
+    this(config, customMetricsReporters, taskFactory, null, null, sp -> processorListener, jobCoordinator);
   }
 
   /**
-   * Same as {@link #StreamProcessor(Config, Map, TaskFactory, ProcessorLifecycleListener, JobCoordinator)}, except
-   * there is a {@link StreamProcessorLifecycleListenerFactory} as input instead of {@link ProcessorLifecycleListener}.
-   * This is useful to create a {@link ProcessorLifecycleListener} with a reference to this {@link StreamProcessor}
+   * Same as {@link #StreamProcessor(Config, Map, TaskFactory, ProcessorLifecycleListener, JobCoordinator)}, with the
```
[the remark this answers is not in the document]
cameronlee314: There was a similar question at https://github.com/apache/samza/pull/606#discussion_r215030081. I'll add a comment about the ticket and see if I can add some Deprecated annotations.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -173,27 +191,42 @@ public StreamProcessor(Config config, Map<String, MetricsReporter> customMetrics
...
    * @param taskFactory task factory to instantiate the Task
+   * @param applicationDefinedContainerContextFactory nullable factory for application-defined container context
+   * @param applicationDefinedTaskContextFactory nullable factory for application-defined task context
    * @param listenerFactory listener to the StreamProcessor life cycle
    * @param jobCoordinator the instance of {@link JobCoordinator}
    */
-  public StreamProcessor(Config config, Map<String, MetricsReporter> customMetricsReporters, TaskFactory taskFactory,
-      StreamProcessorLifecycleListenerFactory listenerFactory, JobCoordinator jobCoordinator) {
+  public StreamProcessor(Config config,
+      Map<String, MetricsReporter> customMetricsReporters,
```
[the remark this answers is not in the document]
cameronlee314: In the cases where the parameter list is long, I thought it helps readability and makes it easier to view a change to the parameter list in the future.
I'll reformat to follow the convention though.

samza-core/src/main/java/org/apache/samza/table/utils/TableMetricsUtil.java
```diff
@@ -46,21 +43,16 @@
...
-    Preconditions.checkNotNull(containerContext);
+  public TableMetricsUtil(Context context, Table table, String tableId) {
+    Preconditions.checkNotNull(context);
     Preconditions.checkNotNull(table);
     Preconditions.checkNotNull(tableId);
 
-    this.metricsRegistry = taskContext == null // The table is at container level, when the task
-        ? containerContext.metricsRegistry     // context passed in is null
-        : taskContext.getMetricsRegistry();
+    this.metricsRegistry = context.getTaskContext().getTaskMetricsRegistry();
```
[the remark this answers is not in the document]
cameronlee314: I was confused by this as well. I don't know how taskContext could be null. I tried searching through the code and couldn't find anything.
I don't think this would be a clean way to decide between task vs. container metrics anyways.
I found the change that added this check (https://github.com/apache/samza/pull/555), and I didn't see any cases where taskContext was ever null. @weisong44 could you please clarify why this check was here?

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
cameronlee314: That is correct.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
cameronlee314: Makes sense.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
cameronlee314: Good catch.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
cameronlee314: I agree that it would be better to avoid the multi-return value. However, I needed to save the taskContext and the applicationTaskContext, and I needed the taskContext to build the applicationTaskContext.
That said, I can just inline this back into the TaskInstance initialization section. It probably doesn't add too much now to move it into a separate method.

samza-core/src/test/java/org/apache/samza/context/TestContextImpl.java
cameronlee314: Oops, missed that. Thanks for pointing it out.

samza-core/src/main/java/org/apache/samza/application/ApplicationDescriptorImpl.java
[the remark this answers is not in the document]
cameronlee314: The "noinspection unchecked" acts like an annotation. I see that the annotation is used more throughout the code, so I'll change to that.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -283,7 +316,13 @@ SamzaContainer getContainer() {
 
   @VisibleForTesting
   SamzaContainer createSamzaContainer(String processorId, JobModel jobModel) {
-    return SamzaContainer.apply(processorId, jobModel, config, ScalaJavaUtil.toScalaMap(customMetricsReporter), taskFactory);
+    return SamzaContainer.apply(processorId,
+        jobModel,
+        ScalaJavaUtil.toScalaMap(this.customMetricsReporter),
+        this.taskFactory,
+        JobContextImpl.fromConfigWithDefaults(this.config),
+        Option.apply(this.applicationDefinedContainerContextFactory),
```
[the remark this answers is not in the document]
cameronlee314: Since this is a Java class, stored it as an Optional and converted to Option.

samza-core/src/main/java/org/apache/samza/context/JobContextImpl.java
```diff
@@ -18,32 +18,38 @@
  */
 package org.apache.samza.context;
 
-import org.apache.samza.SamzaException;
 import org.apache.samza.config.Config;
 import org.apache.samza.config.JobConfig;
+import scala.Option;
```
prateekm: Would prefer not using scala classes from within Java code.

cameronlee314: This is just used as the type of a local var when calling JobConfig.getName, which returns the Option.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
cameronlee314: The type of the object already allows us to know it is an Option, and there isn't any variable in this class that uses the non-Option version, so I would prefer to be more concise here.

prateekm: I prefer suffixing Option/Future etc. because the type declaration can be far from the usage, so this makes it clearer when reviewing code or looking at it without an IDE.

cameronlee314: Ok, I'll update to Option/Optional.

samza-core/src/main/java/org/apache/samza/table/utils/TableMetricsUtil.java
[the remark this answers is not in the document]
weisong44: We don't currently support container level table yet, once that 's added you'd see null taskContext.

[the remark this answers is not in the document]
cameronlee314: Thanks. We will simplify the constructor now to just use task-level metrics, since that is the only current use case. When the container-level table functionality comes in, then this can be modified or a new path can be added.

weisong44:
I would prefer not to make this change, is there specific reason?

–Wei
On Oct 5, 2018, at 11:18, cameronlee314 <notifications@github.com<mailto:notifications@github.com>> wrote:

@cameronlee314 commented on this pull request.

________________________________

In samza-core/src/main/java/org/apache/samza/table/utils/TableMetricsUtil.java<https://eur04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fapache%2Fsamza%2Fpull%2F672%23discussion_r223097945&data=02%7C01%7C%7Cb43c4759baf24bdbaa5808d62aeefd32%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C636743603262673159&sdata=83mwglQhL9O52kgZtb6FXnC7rBBqdoXILR9pL6dvYUw%3D&reserved=0>:

>      Preconditions.checkNotNull(table);
     Preconditions.checkNotNull(tableId);

-    this.metricsRegistry = taskContext == null // The table is at container level, when the task
-        ? containerContext.metricsRegistry     // context passed in is null
-        : taskContext.getMetricsRegistry();
+    this.metricsRegistry = context.getTaskContext().getTaskMetricsRegistry();


Thanks. We will simplify the constructor now to just use task-level metrics, since that is the only current use case. When the container-level table functionality comes in, then this can be modified or a new path can be added.

cameronlee314:
@weisong44 We are updating the public APIs for task/container contexts, so the old TaskContext/SamzaContainerContext no longer exist. Now there is a single Context API which can be used to get to any context information, and we are updating the old context usages to use the new API.
Since there is currently no support for container-level tables, that check is being removed in this PR to keep the implementation clean. Once there is a concrete support for container-level tables, we can discuss how to add that check back in (e.g. maybe instead of passing null context objects, there is a different constructor and/or a config).

samza-core/src/main/java/org/apache/samza/application/ApplicationDescriptorImpl.java
```diff
@@ -64,17 +67,15 @@
...
+   * Applications are not required to define a factory.
+   */
+  private Optional<ApplicationContainerContextFactory<?>> applicationContainerContextFactoryOptional = Optional.empty();
 
-    @Override
-    public void close() {
-    }
-  };
+  /**
+   * Applications are not required to define a factory.
```
prateekm: Minor: Comment is redundant if the type is Optional. Can omit.

samza-core/src/main/java/org/apache/samza/context/TaskContextImpl.java
```diff
@@ -84,4 +95,22 @@ public CallbackScheduler getCallbackScheduler() {
...
+
+  public void registerObject(String name, Object value) {
+    this.objectRegistry.put(name, value);
+  }
+
+  public Object fetchObject(String name) {
+    return this.objectRegistry.get(name);
+  }
+
+  public JobModel getJobModel() {
```
prateekm: Can we update getJobModel() usages to use JobContext and remove this for now?

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -82,16 +81,15 @@
...
-  private TaskContext taskContext;
+  private CallbackScheduler callbackScheduler;
   private ControlMessageSender controlMessageSender;
 
   /**
    * Initialize this {@link OperatorImpl} and its user-defined functions.
    *
-   * @param config  the {@link Config} for the task
-   * @param context  the {@link TaskContext} for the task
+   * @param context the {@link Context} for the task
```
prateekm: Update doc: "the {@link Context} for the application", here and everywhere else.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -102,43 +100,34 @@ public final void init(Config config, TaskContext context) {
...
-          .get(context.getSamzaContainerContext().id);
-      this.taskModel = containerModel.getTasks().get(taskName);
-    } else {
-      this.taskModel = null;
-      this.usedInCurrentTask = true;
-    }
-
-    this.taskContext = taskContext;
-    handleInit(config, taskContext);
+    this.taskModel = taskContext.getTaskModel();
```
prateekm: Do you know what this was doing earlier with the job model null check? When would it be null? Can't tell from the PR, is the `usedInCurrentTask` logic still the same?

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
```diff
@@ -21,14 +21,23 @@
 package org.apache.samza.operators.impl;
 
 import com.google.common.base.Preconditions;
-import org.apache.samza.config.Config;
+import java.util.ArrayList;
```
prateekm: Prefer either reverting import order changes, or if this is the order in majority of classes, doing this for all files in the module for consistency (i.e., not just the ones touched in this PR). If the latter, would also prefer to pick the default IntelliJ sort order and publish it as part of the contributor's guidelines / code style guidelines. We can do the rest of the classes at once in a follow up PR.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -77,9 +85,14 @@ class TaskInstance(
...
     })
-
-  val context = new TaskContextImpl(taskName, metrics, containerContext, systemStreamPartitions.asJava, offsetManager,
-                                    kvStoreSupplier, tableManager, jobModel, streamMetadataCache, timerExecutor)
+  private val taskContext = new TaskContextImpl(taskModel, metrics.registry, kvStoreSupplier, tableManager,
+    new CallbackSchedulerImpl(epochTimeScheduler), offsetManager, jobModel, streamMetadataCache)
+  // need separate field for this instead of using it through Context, since Context throws an exception if it is null
+  private val applicationTaskContextOption = applicationTaskContextFactoryOption.map(_.create(jobContext,
+    containerContext, taskContext, applicationContainerContextOption.orNull))
+  val context: Context = new ContextImpl(jobContext, containerContext, taskContext,
```
prateekm: Minor: don't need type annotation.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -126,7 +141,7 @@ class TaskInstance(
     if (tableManager != null) {
       debug("Starting table manager for taskName: %s" format taskName)
 
-      tableManager.init(containerContext, context)
+      tableManager.init(this.context)
```
prateekm: Minor: No `this.` for consistency with rest of the class. Same for line 154.

samza-core/src/main/java/org/apache/samza/operators/impl/WindowOperatorImpl.java
cameronlee314: I reverted my import order changes to the default. There are still some classes which already don't match the default, so I generally tried to leave those alone for now.

PR #673, opened 2018-09-28
SAMZA-1832: Fix race condition between StreamProcessor and SamzaContainerListener

bharathkk:
The PR addresses the following issues

- We have few scenarios where there is a race condition between `SamzaContainerListener` and `StreamProcessor` that results in incorrect application status being propagated to client

> Consider the scenario when the container runs into an exception in run loop and triggers shutdown sequence and at the same time the user triggers a stop on the `StreamProcessor`. The user request comes in and notices that the container is already shutting down and proceeds to shutdown the job coordinator which in turns shuts down successfully and the `processorListener.onStop()` is invoked. The container eventually invokes the `SamzaContainerListener` callback and updates the exception state after the `StreamProcessor` has finished shutting down.

- Currently, we only propagate failures to `processorListener` when containerException is not null. It is possible for the samza container to take longer than `task.shutdown.ms` to shutdown in which case, we need to propagate a timeout exception to the `processorListener` as opposed assuming the shutdown was successful.

- Make container shutdown idempotent since its only setting a boolean flag to true and there is no need to throw an exception on subsequent invocations.
           1. It simplifies the interaction of other components (`StreamProcessor`) with container and prevents unnecessary check on container state to determine if its safe to call shutdown or not.
           2. Enables to reason about the impact of container state change on `StreamProcessor` easily since we restrict container interaction w/ `StreamProcessor` only via `SamzaContainerListener`.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -309,8 +309,10 @@ public class StreamProcessor {
       LOGGER.info(String.format("Shutdown status of container: %s for stream processor: %s is: %b.", container, processorId, hasContainerShutdown));
     }
 
-    // We want to propagate when container shutdown times out. It is possible that the timeout exception we propagate to the
-    // local application runner maybe overwritten with the actual container failure cause in case of interleaved execution
+    // We want to propagate TimeoutException when container shutdown times out. It is possible that the timeout exception
+    // we propagate to the local application runner maybe overwritten by container failure cause in case of interleaved execution.
+    // It is acceptable since container exception is much more useful compared to timeout exception.
+    // We can infer from the logs about the fact that container shutdown timed out or not for additional inference.
     if (!hasContainerShutdown && containerException == null) {
       containerException = new TimeoutException("Container shutdown timed out.");
     }
--- a/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
+++ b/samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
@@ -28,11 +28,11 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import org.apache.samza.annotation.InterfaceStability;
 import org.apache.samza.config.Config;
 import org.apache.samza.config.JobCoordinatorConfig;
 import org.apache.samza.config.TaskConfigJava;
-import org.apache.samza.container.IllegalContainerStateException;
 import org.apache.samza.container.SamzaContainer;
 import org.apache.samza.container.SamzaContainerListener;
 import org.apache.samza.coordinator.JobCoordinator;
@@ -104,7 +104,7 @@ public class StreamProcessor {
   private final ExecutorService executorService;
   private final Object lock = new Object();
 
-  private Throwable containerException = null;
+  private volatile Throwable containerException = null;
 
   volatile CountDownLatch containerShutdownLatch = new CountDownLatch(1);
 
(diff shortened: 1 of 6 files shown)


samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -298,22 +298,25 @@ private JobCoordinator createJobCoordinator() {
   private boolean stopSamzaContainer() {
     boolean hasContainerShutdown = true;
     if (container != null) {
-      if (!container.hasStopped()) {
-        try {
-          container.shutdown();
-          LOGGER.info("Waiting {} ms for the container: {} to shutdown.", taskShutdownMs, container);
-          hasContainerShutdown = containerShutdownLatch.await(taskShutdownMs, TimeUnit.MILLISECONDS);
-        } catch (IllegalContainerStateException icse) {
```
prateekm: ICSE is not a timeout. Maybe set containerException to this value, or log this and then set hasContainerShutdown = false?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -298,22 +298,25 @@ private JobCoordinator createJobCoordinator() {
...
       }
+      LOGGER.info(String.format("Shutdown status of container: %s for stream processor: %s is: %b.", container, processorId, hasContainerShutdown));
+    }
+
+    // We want to propagate TimeoutException when container shutdown times out. It is possible that the timeout exception
+    // we propagate to the local application runner maybe overwritten by container failure cause in case of interleaved execution.
+    // It is acceptable since container exception is much more useful compared to timeout exception.
+    // We can infer from the logs about the fact that container shutdown timed out or not for additional inference.
+    if (!hasContainerShutdown && containerException == null) {
+      containerException = new TimeoutException("Container shutdown timed out.");
```
prateekm: Also include the timeout value in the message.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -359,8 +362,12 @@ public void onNewJobModel(String processorId, JobModel jobModel) {
       public void onCoordinatorStop() {
         synchronized (lock) {
           LOGGER.info("Shutting down the executor service of the stream processor: {}.", processorId);
-          stopSamzaContainer();
-          executorService.shutdownNow();
+          boolean hasContainerShutdown = stopSamzaContainer();
+
+          // we only want to interrupt when container shutdown times out.
```
prateekm: Why? Is something else running on this executor other than the container?
If not, would also prefer renaming this to containerExecutorService to clarify it's purpose.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
prateekm: I guess the idea is to make container shutdown idempotent? If so, removing ICSE is fine.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
shanthoosh: > Is something else running on this executor other than the container

There will be nothing running other than SamzaContainer within that executor service. Renaming should be  fine.

samza-test/src/test/java/org/apache/samza/test/framework/FaultInjectionTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+public class FaultInjectionTest extends StreamApplicationIntegrationTestHarness {
+  @Test
+  public void testRaceCondition() throws InterruptedException {
+    int taskShutdownInMs = (int) (Math.random() * 10000);
+
+    CountDownLatch containerShutdownLatch = new CountDownLatch(1);
+    Map<String, String> configs = new HashMap<>();
+    configs.put(JobCoordinatorConfig.JOB_COORDINATOR_FACTORY, "org.apache.samza.standalone.PassthroughJobCoordinatorFactory");
+    configs.put(JobCoordinatorConfig.JOB_COORDINATION_UTILS_FACTORY, "org.apache.samza.standalone.PassthroughCoordinationUtilsFactory");
```
shanthoosh: This idea is a good one, but is it possible to test this with `ZkJobCoordinatorFactory` (in zookeeper world). Brooklin is removing support for pass-through coordinator and is recommending existing users to migrate to Zk-Standalone. So it will be useful if we can use ZkJobcoordinator here(since `PassthroughCoordinator` will deprecated sooner and if it works in ZkJobCoordinator(complicated case), it will work in pass-through world as well).

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -878,9 +878,6 @@ class SamzaContainer(
    * @throws SamzaException, Thrown when the container has already been stopped or failed
    */
   def shutdown(): Unit = {
-    if (status == SamzaContainerStatus.STOPPED || status == SamzaContainerStatus.FAILED) {
```
shanthoosh: Rather than removing this if-block, do we think it's useful to add a log-message at warn-log level(rather than throwing up).

shanthoosh:
Looks like there're compilation errors in new tests, please fix them.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -298,22 +298,25 @@ private JobCoordinator createJobCoordinator() {
...
       }
+      LOGGER.info(String.format("Shutdown status of container: %s for stream processor: %s is: %b.", container, processorId, hasContainerShutdown));
+    }
+
+    // We want to propagate TimeoutException when container shutdown times out. It is possible that the timeout exception
+    // we propagate to the local application runner maybe overwritten by container failure cause in case of interleaved execution.
+    // It is acceptable since container exception is much more useful compared to timeout exception.
+    // We can infer from the logs about the fact that container shutdown timed out or not for additional inference.
+    if (!hasContainerShutdown && containerException == null) {
+      containerException = new TimeoutException("Container shutdown timed out after " + taskShutdownMs + " ms.");
```
nickpan47: nit: can't we move this line together with line 307?

bharathkk: I have captured the shutdown failure exception as opposed to moving it since we also need to populate the exception in the case of latch timing out.

PR #676, opened 2018-09-29
SAMZA-1927: Samza-sql - always repartition the stream denoted by stream-table join.

atoomula:


samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
```diff
@@ -80,8 +81,11 @@
   public static final String CFG_UDF_RESOLVER = "samza.sql.udfResolver";
   public static final String CFG_FMT_UDF_RESOLVER_DOMAIN = "samza.sql.udfResolver.%s.";
 
+  public static final String CFG_METADATA_TOPIC_PREFIX = "samza.sql.metadataTopicPrefix";
```
prateekm: If the intention is to reset state from a previous run, I'd strongly prefer not to go down the rout e of introducing SQL specific configs. I think this behavior could also be achieved by incrementing job.id. Would that be sufficient?

We can also introduce a job.version configuration for such use cases.

srinipunuru: @prateekm  We probably need a single config in Samza Framework that says reset.state which takes care of all these things under the hood. When this feature is available we could remove this workaround.
This particular targeted fix is needed to unblock a customer who has been patiently waiting on us while we are fixing this stream-table join issues one by one. We also need to cherry-pick this fix into samza sql hot fix branch. Changing the framework is unfortunately not an option at this point.

PR #682, opened 2018-10-01
SAMZA-1929: Expose Kafka topic level configuration in local store backed table descriptors

weisong44:
The changelog Kafka topic level configuration may be useful for adjustment, when TTL is set on a RocksDB store or when enabled for in-memory DB. User should be able to set Kafka topic level properties as follows

TableDescriptor desc = new RocksDbTableDescriptor(...)
    .withChangelogKafkaTopicLevelProperty("cleanup.policy", "delete")
    .withChangelogKafkaTopicLevelProperty("retention.ms", "3600");


prateekm:
Thanks for adding this. Would prefer to not tie changelogs to Kafka, esp. in the API. Also, now that we have StreamDescriptors, I think it'll be better to accept a StreamDescriptor for the changelog (i.g., withChangelog(OutputStreamDescriptor). This way users can configure these system specific properties (replication.factor, cleanup.policy etc) on the stream itself, which makes more sense. What do you think?

weisong44:
@prateekm Updated based on our discussion

samza-kv-rocksdb/src/main/java/org/apache/samza/storage/kv/RocksDbTableProvider.java
```diff
@@ -65,6 +64,38 @@ public RocksDbTableProvider(TableSpec tableSpec) {
...
+    //  1) Disable log compaction on Kafka changelog topic
+    //  2) Set topic TTL to be slightly larger than RocksDB TTL
+    //
+    String rocksDbTtl = tableSpec.getConfig().get("rocksdb." + RocksDbTableDescriptor.ROCKSDB_TTL_MS);
+    String changelogEnabled = tableSpec.getConfig().get(BaseLocalStoreBackedTableDescriptor.INTERNAL_ENABLE_CHANGELOG);
+    if (rocksDbTtl != null && "true".equalsIgnoreCase(changelogEnabled)) {
+
+      String changelogPrefix = String.format(StorageConfig.CHANGELOG_STREAM(), tableSpec.getId()) + ".";
+
+      // We should use regular Kafka topic instead of log-compacted
```
jmakes: Could use both compact and delete.

jmakes: Why add kafka-specific settings to the RocksDbTableProvider?

prateekm: If we do this this shouldn't be only for the table API. We should have the same behavior for regular stores as well.

weisong44: Moved this to KafkaConfig, logic is still the same. User could override it own config if desired.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -259,7 +259,22 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
     //  kafkaChangeLogProperties.setProperty("cleanup.policy", "compact,delete")
     //  kafkaChangeLogProperties.setProperty("retention.ms", String.valueOf(KafkaConfig.DEFAULT_RETENTION_MS_FOR_BATCH))
     // }
-    kafkaChangeLogProperties.setProperty("cleanup.policy", "compact")
+
```
prateekm: Minor: I think you can remove the code block above and just leave the comment with the ticket.

weisong44: The "compact,delete" is only available on 0.11.1.57 or later, which is fairly new. We will stay with "delete" for now.

weisong44:
Confirmed with @nickpan47

PR #683, opened 2018-10-02
Samza 1771: Test Framework support for stateful testing and assertions over state

Sanil15:
- Gives StateAssert for stateful assertions 
- Gives examples of Testing stateful jobs for TaskApplication and StreamApplication


samza-kv-rocksdb/src/main/java/org/apache/samza/storage/kv/RocksDbKeyValueReader.java
```diff
@@ -80,6 +84,33 @@ public RocksDbKeyValueReader(String storeName, String dbPath, Config config) {
...
   }
 
+  /**
+   * Construct the <code>RocksDbKeyValueReader</code> with store's descriptor,
+   * database's path and Samza's config
+   *
+   * @param descriptor describes rocksdb
+   * @param dbPath path to the db directory
+   */
+  public RocksDbKeyValueReader(RocksDbTableDescriptor descriptor, String dbPath) {
```
cameronlee314: Would it be simpler to use an in-memory store instead of RocksDB? Does RocksDB give something that in-memory does not?

samza-kv-rocksdb/src/main/java/org/apache/samza/storage/kv/RocksDbKeyValueReader.java
```diff
@@ -80,6 +84,33 @@ public RocksDbKeyValueReader(String storeName, String dbPath, Config config) {
...
+    keySerde = descriptor.getSerde().getKeySerde();
+    valueSerde = descriptor.getSerde().getValueSerde();
+
+    // get db options
+    ArrayList<TaskName> taskNameList = new ArrayList<TaskName>();
```
cameronlee314: If you do need this new constructor, then can you please consolidate the repeated logic in here and the existing constructor?

samza-test/src/main/java/org/apache/samza/test/framework/StateAssert.java
```diff
@@ -0,0 +1,85 @@
...
+ *  {@link org.apache.samza.operators.TableDescriptor}.
+ */
+public class StateAssert {
+
+  /**
+   * Verifies that the {@code key} is present in the table described by {@code RocksDbTableDescriptor}
+   * @param key expected key in the table
+   * @param tableDescriptor describes the RocksDb table
+   */
+  public static void contains(Object key, RocksDbTableDescriptor tableDescriptor) {
```
cameronlee314: Is this necessary when you already have the "get" API?
If you do keep this, can you please make the argument ordering consistent with "get"? Also, can you add a javadoc clarifying how this works across multiple partitions?

samza-test/src/main/java/org/apache/samza/test/framework/StateAssert.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /**
+   * Fetches the list of values associated with a {@code key} in all the partitions of the table
+   * described by {@code tableDescriptor}
+   * @param tableDescriptor describes the RocksDb table
+   * @param key expected key in the table
+   * @return list of values associated with the {@code key}
+   */
+  public static List get(RocksDbTableDescriptor tableDescriptor, Object key) {
```
cameronlee314: Does this only work for local RocksDB tables? Since internal state is not exposed outside the container, is it necessary to test local state directly? Instead of checking internal state, could you trigger the actual functionality (e.g. join actually outputs correct events)?
If you do keep this, would it be good to know which partition had the data?
It makes sense to have verification for remote tables, so is there a plan to extend this to those use cases?

samza-test/src/main/java/org/apache/samza/test/framework/StateAssert.java
```diff
@@ -0,0 +1,85 @@
...
+      if (reader.get(key) != null)
+        values.add(reader.get(key));
+      reader.stop();
+    }
+    return values;
+  }
+
+  private static File[] getPartitionDirectories(String store) {
+    File currentDirFile = new File(".");
+    File storeDir = new File(currentDirFile.getAbsolutePath().replace(".", "state/" + store + "/"));
```
cameronlee314: How does the write location for RocksDB get specified, and is there a way to guarantee it is read/written from the same place?

samza-kv-rocksdb/src/main/java/org/apache/samza/storage/kv/RocksDbKeyValueReader.java
Sanil15: InMemory Store gets cleaned once the Samza job is finished, I spoke to @atoomula about it.

samza-kv-rocksdb/src/main/java/org/apache/samza/storage/kv/RocksDbKeyValueReader.java
```diff
@@ -80,6 +84,33 @@ public RocksDbKeyValueReader(String storeName, String dbPath, Config config) {
...
+    taskNameList.add(new TaskName("read-rocks-db"));
```
shanthoosh: Since we're not using RocksDBOptions here, it would be better to get rid of these object constructions completely.

samza-test/src/main/java/org/apache/samza/test/framework/StateAssert.java
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * Verifies that the {@code key} is present in the table described by {@code RocksDbTableDescriptor}
+   * @param key expected key in the table
+   * @param tableDescriptor describes the RocksDb table
+   */
+  public static void contains(Object key, RocksDbTableDescriptor tableDescriptor) {
+    boolean contains = false;
+    for (File partition: getPartitionDirectories(tableDescriptor.getTableId())) {
+      RocksDbKeyValueReader reader = new RocksDbKeyValueReader(tableDescriptor, partition.getAbsolutePath());
```
shanthoosh: SamzaContainer has multiple tasks and each task has a associated state-store.

Two different tasks(T1, T2) running on a single container will have state-store created under different directories viz `{RocksDBBaseDir}/T1/storeName`, `{RocksDBBaseDir}/T2/storeName` . 

In the above case, when the user invokes `get`, `contains`  API, it's not clear from which task state-store the user is reading the key from and what are our `StreamAssert` guarantees?

samza-kv-rocksdb/src/main/java/org/apache/samza/storage/kv/RocksDbKeyValueReader.java
```diff
@@ -130,4 +130,4 @@ public void stop() {
     }
     return Util.getObj(serdeClassName, SerdeFactory.class).getSerde(name, serializerConfig);
   }
-}
+}
```
shanthoosh: Please revert this change.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -262,6 +275,14 @@ public void run(Duration timeout) {
     if (status.getStatusCode() == ApplicationStatus.StatusCode.UnsuccessfulFinish) {
       throw new SamzaException(ExceptionUtils.getStackTrace(status.getThrowable()));
     }
+    // Clean the state directory if any
```
shanthoosh: We're cleaning the state stores are cleaned up after running the stream application. However,  if there're any exceptions thrown by `runner.run` or `runner.waitForFinish` or if the user stops the test before the completion, this store deletion code flow will not get triggered. So it will be safer to clean the rocksDB state before the run of an application(and we will know that there is no pending state before start).

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -262,6 +275,14 @@ public void run(Duration timeout) {
...
+    Path stateDir = Paths.get(configs.get(JobConfig.JOB_NON_LOGGED_STORE_BASE_DIR()));
+    if (!Files.exists(stateDir)) return;
```
shanthoosh: It's a good practice to enclose even single line if-blocks within `{`, `}`.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -262,6 +275,14 @@ public void run(Duration timeout) {
...
+    try {
+      Files.walk(stateDir).sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
```
shanthoosh: Before invoking delete on directory, it will be better to add a logging at INFO level to indicate that we're deleting the store.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -262,6 +275,14 @@ public void run(Duration timeout) {
...
+    } catch (IOException e) {
+      throw new SamzaException("Could not delete the JOB_NON_LOGGED_STORE_BASE_DIR due to: \n" + e.getMessage());
```
shanthoosh: This message has to be warn/error. It will be better to not fail the test for this. What do you think?

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -98,6 +105,12 @@ private TestRunner() {
     configs.put(JobConfig.PROCESSOR_ID(), "1");
     configs.put(JobCoordinatorConfig.JOB_COORDINATOR_FACTORY, PassthroughJobCoordinatorFactory.class.getName());
     configs.put(TaskConfig.GROUPER_FACTORY(), SingleContainerGrouperFactory.class.getName());
+    // RockDb Tables if used needs this to be configured if no changelog is configured
```
shanthoosh: s/RockDb Tables if used needs this to be configured if no changelog is configured/Setup rocksDB stores without changelog.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -98,6 +105,12 @@ private TestRunner() {
...
+    configs.put(JobConfig.JOB_NON_LOGGED_STORE_BASE_DIR(),
+        FileSystems.getDefault().getPath(this.inMemoryScope).toAbsolutePath().toString() + File.separator);
+    // SideInput Tables needs this to be configured for persisting data
```
shanthoosh: s/SideInput Tables needs this to be configured for persisting data/configure sideInput tables

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -251,9 +264,9 @@ public TestRunner addOutputStream(InMemoryOutputDescriptor<?> streamDescriptor,
    * @throws SamzaException if Samza job fails with exception and returns UnsuccessfulFinish as the statuscode
    */
   public void run(Duration timeout) {
-    Preconditions.checkState(app != null,
-        "TestRunner should run for Low Level Task api or High Level Application Api");
+    Preconditions.checkState(app != null, "TestRunner should run for Low Level Task api or High Level Application Api");
```
shanthoosh: Should set either `StreamTaskApplication` or  `StreamApplication` before run.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
shanthoosh: There's an existing utility method `FileUtil.rm(file)` in `FileUtil.scala` in samza which recursively deletes the entire directory. This duplication here can be removed by simply switching to it.

shanthoosh:
@Sanil15 

Currently `TestRunner` API creates the RocksDB state relative to the main samza  directory. For whatever reason, if the RocksDB state is not cleaned up from a test run then the subsequent ./gradlew build will fail with the following rat exception: 
```
:rat
Rat report: build/rat/rat-report.html
Unknown license: /home/svenkata/code/samza/samza-1/samza/samza-test/logged/profile-table/Partition_0/000003.log
Unknown license: /home/svenkata/code/samza/samza-1/samza/samza-test/logged/profile-table/Partition_0/CURRENT
Unknown license: /home/svenkata/code/samza/samza-1/samza/samza-test/logged/profile-table/Partition_0/IDENTITY
Unknown license: /home/svenkata/code/samza/samza-1/samza/samza-test/logged/profile-table/Partition_0/LOCK
Unknown license: /home/svenkata/code/samza/samza-1/samza/samza-test/logged/profile-table/Partition_0/LOG
Unknown license: /home/svenkata/code/samza/samza-1/samza/samza-test/logged/profile-table/Partition_0/OPTIONS-000005
:rat FAILED
```

To fix, we can set the RocksDB state store base directory as `/tmp`

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -98,8 +105,16 @@ private TestRunner() {
     configs.put(JobConfig.PROCESSOR_ID(), "1");
     configs.put(JobCoordinatorConfig.JOB_COORDINATOR_FACTORY, PassthroughJobCoordinatorFactory.class.getName());
     configs.put(TaskConfig.GROUPER_FACTORY(), SingleContainerGrouperFactory.class.getName());
+    // Changing the base directory for non-changelog stores used by Samza application to separate the
+    // on-disk store locations for concurrently executing tests
+    configs.put(JobConfig.JOB_NON_LOGGED_STORE_BASE_DIR(),
+        FileSystems.getDefault().getPath("/tmp/" + this.inMemoryScope).toAbsolutePath().toString() + File.separator);
```
prateekm: Instead of hardcoding path, can you check if `Files.createTempDirectory()` works?

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -98,8 +105,16 @@ private TestRunner() {
...
+    // Changing the base directory for changelog stores (here SideInput) used by Samza application to separate the
```
prateekm: Don't need the same comment twice. Side inputs is not relevant here.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -98,8 +105,16 @@ private TestRunner() {
...
+    // on-disk store locations for concurrently executing tests
+    configs.put(JobConfig.JOB_LOGGED_STORE_BASE_DIR(),
+        FileSystems.getDefault().getPath("/tmp/" + this.inMemoryScope).toAbsolutePath().toString() + File.separator);
     addConfig(JobConfig.JOB_DEFAULT_SYSTEM(), JOB_DEFAULT_SYSTEM);
-    // This is important because Table Api enables host affinity by default for RocksDb
+    // Disabling host affinity since Table Api enables it by default for RocksDb and expects coordinator stream
```
prateekm: "Disabling host affinity since it requires reading locality information from a Kafka coordinator stream."

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -251,17 +266,20 @@ public TestRunner addOutputStream(InMemoryOutputDescriptor<?> streamDescriptor,
    * @throws SamzaException if Samza job fails with exception and returns UnsuccessfulFinish as the statuscode
    */
   public void run(Duration timeout) {
-    Preconditions.checkState(app != null,
-        "TestRunner should run for Low Level Task api or High Level Application Api");
+    Preconditions.checkState(app != null, "No Samza application to run");
```
prateekm: Preconditions.checkNotNull

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -251,17 +266,20 @@ public TestRunner addOutputStream(InMemoryOutputDescriptor<?> streamDescriptor,
...
     Preconditions.checkState(!timeout.isZero() || !timeout.isNegative(), "Timeouts should be positive");
+    // Cleaning store directories to ensure current run does not pick up state from previous run
+    deleteStoreDirectories();
     final LocalApplicationRunner runner = new LocalApplicationRunner(app, new MapConfig(configs));
     runner.run();
     boolean timedOut = !runner.waitForFinish(timeout);
     Assert.assertFalse("Timed out waiting for application to finish", timedOut);
     ApplicationStatus status = runner.status();
     if (status.getStatusCode() == ApplicationStatus.StatusCode.UnsuccessfulFinish) {
+      deleteStoreDirectories();
```
prateekm: Can you just move deleteStoreDirectories before the status check?

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -251,17 +266,20 @@ public TestRunner addOutputStream(InMemoryOutputDescriptor<?> streamDescriptor,
...
       throw new SamzaException(ExceptionUtils.getStackTrace(status.getThrowable()));
```
prateekm: What is this doing? We should not be manipulating exception stack traces.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -369,4 +387,20 @@ public void run(Duration timeout) {
             new EndOfStreamMessage(null)));
       });
   }
+
+  private void deleteStoreDirectories() {
+    Preconditions.checkNotNull(configs.get(JobConfig.JOB_LOGGED_STORE_BASE_DIR()),
```
prateekm: What is this doing? Does checkNotNull take a vararg param?

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
prateekm: Looks like this is stringifying the original exception and using it as the message. You should use the constructor that accepts a (meaningful) message and a throwable. I.e., new SamzaException("Application did not finish running successfully.", e)

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: Files.createTempDirectory() takes in the absolute path and involves manually deletion as well so this works better

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: ok

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: ok

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: ok

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: ok

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -251,17 +266,20 @@ public TestRunner addOutputStream(InMemoryOutputDescriptor<?> streamDescriptor,
...
-    Preconditions.checkState(app != null,
-        "TestRunner should run for Low Level Task api or High Level Application Api");
+    Preconditions.checkState(app != null, "No Samza application to run");
     Preconditions.checkState(!timeout.isZero() || !timeout.isNegative(), "Timeouts should be positive");
+    // Cleaning store directories to ensure current run does not pick up state from previous run
+    deleteStoreDirectories();
     final LocalApplicationRunner runner = new LocalApplicationRunner(app, new MapConfig(configs));
     runner.run();
     boolean timedOut = !runner.waitForFinish(timeout);
     Assert.assertFalse("Timed out waiting for application to finish", timedOut);
```
prateekm: Throw an exception or use Preconditions instead of using junit Assert here. You can also inline this check in the block below. `if (timeout or status == unsuccessful finish) throw meaningful exception`

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: when status is not successful, throwing an exception so Junit reports a failure

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: sanity check, if some users end up removing it,  yes it takes a vararg param

prateekm: Are you sure? The matching signature I see in my IDE is:
`public static <T> T checkNotNull(T reference, @Nullable Object errorMessage)`?
That wouldn't do what I think you want to do - check both configs for existence.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
prateekm: Are you sure? From the javadoc for the 2-arg variant:
`Creates a new directory in the default temporary-file directory, using the given prefix to generate its name.`

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
prateekm: I'm saying that you should throw a meaningful exception and not manipulate exception stack traces. See second comment above.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: I tested it, doc says it creates directory with the prefix in the default temporary directory, that might be not set to /tmp/

it generated for me in /private/var/folders/0p/sjzh971n0nzfnj3k3kjd7qvm000whj/T/sanil-jain1818152723896917306

I guess putting it in /tmp/ is a better option since we are aware where temp state directories get created

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: sure

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
prateekm: Why does it matter? You can always log the directory path somewhere later.

Sanil15: https://softwareengineering.stackexchange.com/questions/314796/should-temporary-files-be-saved-to-tmp-or-the-current-working-directory

prateekm: Advantages of using "OS temporary directory", which that /private/var/folders directory is on OSX.
https://apple.stackexchange.com/questions/94964/where-is-the-temp-folder/131080

PR #684, opened 2018-10-02
SAMZA-1933: Fix NPE in LocalityManager. 

shanthoosh:


shanthoosh:
@xinyuiscool prateekm  Please take a look when you get a chance.

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -128,10 +128,11 @@ public void writeContainerToHostMapping(String containerId, String hostName) {
 
   public void close() {
     metadataStore.close();
-    taskAssignmentManager.close();
   }
 
   public TaskAssignmentManager getTaskAssignmentManager() {
```
prateekm: Why make this change? It's weird for a getter to return new objects. It's also weird that init is called here but caller is responsible for calling close.

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
```diff
@@ -95,50 +95,54 @@ public GroupByContainerCount(int containerCount) {
...
 
     TaskAssignmentManager taskAssignmentManager = localityManager.getTaskAssignmentManager();
-    List<TaskGroup> containers = getPreviousContainers(taskAssignmentManager, tasks.size());
-    if (containers == null || containers.size() == 1 || containerCount == 1) {
-      log.info("Balancing does not apply. Invoking grouper.");
-      Set<ContainerModel> models = group(tasks);
-      saveTaskAssignments(models, taskAssignmentManager);
-      return models;
-    }
+    try {
```
prateekm: Is the try-finally the only difference?

shanthoosh: Yes.

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
shanthoosh: Before using the operations from metadatastore, one has to invoke metadatastore.init(). That wasn't done, hence we'd the NPE bug referenced by @xinyuiscool in the PR. 

As we discussed offline, deleted the `getTaskAssignmentManager` method from `LocalityManager`. I agree that having a getAPI that creates objects like factory does not make much sense. In my initial migration to `MetadataStore`, we missed this part somehow. For now, deleted this method and in couple of places where this was used, simply in-lined the invocation(Other changes in the patch are mocking changes in tests).

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -128,10 +125,9 @@ public void writeContainerToHostMapping(String containerId, String hostName) {
 
   public void close() {
     metadataStore.close();
-    taskAssignmentManager.close();
   }
 
-  public TaskAssignmentManager getTaskAssignmentManager() {
-    return taskAssignmentManager;
+  public Config getConfig() {
```
prateekm: Locality manager shouldn't expose config that was passed to it.

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
```diff
@@ -94,51 +95,56 @@ public GroupByContainerCount(int containerCount) {
...
 
-    TaskAssignmentManager taskAssignmentManager = localityManager.getTaskAssignmentManager();
-    List<TaskGroup> containers = getPreviousContainers(taskAssignmentManager, tasks.size());
-    if (containers == null || containers.size() == 1 || containerCount == 1) {
-      log.info("Balancing does not apply. Invoking grouper.");
-      Set<ContainerModel> models = group(tasks);
-      saveTaskAssignments(models, taskAssignmentManager);
-      return models;
-    }
+    TaskAssignmentManager taskAssignmentManager =  new TaskAssignmentManager(localityManager.getConfig(), new MetricsRegistryMap());
```
prateekm: Why pass a no-op metrics registry map? Either it creates useful metrics and we need to emit them, or it doesn't and we should remove this param.

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
```diff
@@ -30,19 +30,27 @@
 import org.apache.samza.job.model.TaskModel;
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mockito;
+import org.powermock.api.mockito.PowerMockito;
```
prateekm: Prefer not using PowerMock. Can you make this a protected method that's mocked in test?

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
```diff
@@ -320,7 +328,7 @@ public void testBalancerAfterContainerDecrease() {
    *  T8  T7  T3
    */
   @Test
-  public void testBalancerMultipleReblances() {
+  public void testBalancerMultipleReblances() throws Exception {
```
prateekm: "Rebalances". Also, why does the test `throws Exception`?

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/container/grouper/task/GroupByContainerCount.java
shanthoosh: In my next follow-up patch as a part of SEP-11, we'll be passing in the taskAssignments(static map) directly in the grouper API interface. So we will deprecate this API and this change will be no longer relevant and will be removed.

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
shanthoosh: I agree this is better than Powermock, but this mocking should be deleted in my next patch(as part of SEP-11). Since this is a short-lived, not sure the value of changing it in right way.

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
shanthoosh: Mocking via `PowerMock` bubbles up an exception, hence there was a `throws` clause.  I will remove it in my follow-up patch(part of SEP-11).

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
prateekm: Sounds good, let's remove this in the follow up PR.

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
```diff
@@ -25,50 +25,67 @@
...
-  public void setup() {
+  public void setup() throws Exception {
     taskAssignmentManager = mock(TaskAssignmentManager.class);
     localityManager = mock(LocalityManager.class);
-    when(localityManager.getTaskAssignmentManager()).thenReturn(taskAssignmentManager);
+    PowerMockito.whenNew(TaskAssignmentManager.class).withAnyArguments().thenReturn(taskAssignmentManager);
+    Mockito.doNothing().when(taskAssignmentManager).init();
+  }
+
+  Config getConfig(int containerCount) {
```
prateekm: Move private methods to the bottom of the class.

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
prateekm: Please don't resolve comments right after replying. This collapses them in the UI and makes it difficult to follow up.

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerCount.java
shanthoosh: Done.

PR #691, opened 2018-10-04
SAMZA-1937: Make DSL converter work when sql is a query

weiqingy:
To support the use case that users use Samza Sql Console or Samza Sql shell to perform SELECT query. 

## How was this patch tested?
1. Build passed and current unit tests passed
2. Testing in Samza Sql shell


samza-sql/src/main/java/org/apache/samza/sql/dsl/SamzaSqlDslConverter.java
```diff
@@ -38,11 +38,13 @@
 import org.slf4j.LoggerFactory;
 
 import static org.apache.samza.sql.runner.SamzaSqlApplicationConfig.*;
+import static org.apache.samza.sql.testutil.SamzaSqlQueryParser.*;
```
nickpan47: Please try to avoid import *

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
```diff
@@ -52,10 +52,13 @@
 import org.apache.samza.sql.interfaces.UdfResolver;
 import org.apache.samza.sql.testutil.JsonUtil;
 import org.apache.samza.sql.testutil.ReflectionUtils;
+import org.apache.samza.sql.testutil.SamzaSqlQueryParser;
 import org.codehaus.jackson.type.TypeReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.samza.sql.dsl.SamzaSqlDslConverter.*;
```
nickpan47: same here.

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
```diff
@@ -82,6 +85,8 @@
 
   public static final String CFG_GROUPBY_WINDOW_DURATION_MS = "samza.sql.groupby.window.ms";
 
+  private static final String LOG_OUTPUT_STREAM = "log.outputStream";
```
nickpan47: This is dup from SamzaSqlDslConverter's static var definition. Can we move it to a common place?

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
```diff
@@ -202,9 +209,16 @@ private UdfResolver createUdfResolver(Map<String, String> config) {
 
     Collection<RelRoot> relRoots = dslConverter.convertDsl(String.join("\n", dslStmts));
 
-    for (RelRoot relRoot : relRoots) {
-      SamzaSqlApplicationConfig.populateSystemStreams(relRoot.project(), inputSystemStreams, outputSystemStreams);
-    }
+    // the snippet below dose not work when sql is a query
```
nickpan47: Please either do a // TODO or // FIXME, or remove this code.

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
```diff
@@ -52,6 +65,27 @@
...
+    private transient SamzaRelConverter samzaMsgConverter;
+    private final String outputTopic;
+
+    OutputMapFunction(String outputTopic) {
+      this.outputTopic = outputTopic;
+    }
+
+    @Override
+    public void init(Config config, TaskContext taskContext) {
+      TranslatorContext context = (TranslatorContext) taskContext.getUserContext();
```
nickpan47: Does this PR fix the "different TranslatorContext instance for different query statements" problem?

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
```diff
@@ -52,6 +65,27 @@
   private final ModifyTranslator modifyTranslator;
   private final SamzaSqlApplicationConfig sqlConfig;
   private final Map<String, SamzaRelConverter> converters;
+  private static final String LOG_OUTPUT_STREAM = "log.outputStream";
```
nickpan47: Again, please don't repeatedly define the same configure vars everywhere.

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
```diff
@@ -143,10 +182,29 @@ public void init(Config config, TaskContext taskContext) {
...
       public void close() {
-
       }
-
     });
+  }
 
+  private void sendToOutputStream(StreamApplicationDescriptor appDesc, TranslatorContext context, RelNode node) {
+    SqlIOConfig sinkConfig = sqlConfig.getOutputSystemStreamConfigsBySource().get(LOG_OUTPUT_STREAM);
+    MessageStreamImpl<SamzaSqlRelMessage> stream = (MessageStreamImpl<SamzaSqlRelMessage>) context.getMessageStream(node.getId());
```
nickpan47: nit: we can use MessageStream here instead of MessageStreamImpl, since we don't need to access non-public API functions of stream.

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
```diff
@@ -202,9 +209,16 @@ private UdfResolver createUdfResolver(Map<String, String> config) {
...
+    // for (RelRoot relRoot : relRoots) {
+    //   SamzaSqlApplicationConfig.populateSystemStreams(relRoot.project(), inputSystemStreams, outputSystemStreams);
+    // }
```
atoomula: Is there no way around commenting these ? In the case of DSLs, we won't be able to parse the query and get the input/output streams. All we have is relational plan in the form of RelRoot.

samza-sql/src/main/java/org/apache/samza/sql/dsl/SamzaSqlDslConverter.java
weiqingy: Done.

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
weiqingy: Done

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
weiqingy: No. PR https://github.com/apache/samza/pull/697 is for the issue we discussed last week "different TranslatorContext instance for different query statements".

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
weiqingy: yes.

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
weiqingy: done

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
weiqingy: Thanks for reviewing this PR. 
"FIXME" has been added.

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
weiqingy: Before QueryTranslator performs "translate()", "SamzaSqlApplicationConfig" or "SamzaSqlApplication" has to know if the Sql statements has queries . If we still use this approach that passes the query only to Calcite if users perform SELECT sql statement. To around commenting these lines, we could add an API "hasSqlQuery()" in "interface DslConverter" or somewhere. 

Or, we can think of other solutions, for example, we still pass the whole Sql statement to Calcite, but still we need to figure out a way to deal with the issue of missing schema of log.outputStream.

weiqingy:
Now this PR conflicts with master ... I am not familiar with the latest code change. Need more time to fix this.

samza-sql/src/main/java/org/apache/samza/sql/dsl/SamzaSqlDslConverter.java
```diff
@@ -37,12 +37,11 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import static org.apache.samza.sql.runner.SamzaSqlApplicationConfig.*;
-
 
 public class SamzaSqlDslConverter implements DslConverter {
 
   private static final Logger LOG = LoggerFactory.getLogger(SamzaSqlDslConverter.class);
+  private static final String LOG_OUTPUT_STREAM = "log.outputStream";
```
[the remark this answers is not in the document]
weiqingy: Thanks for the review. 

Yes, I've updated the PR to check system name "log" only.

weiqingy:
After merging master branch, this PR looks messy. I'll another PR for this.

PR #695, opened 2018-10-05
Samza SQL Documentation

srinipunuru:
This is initial Samza SQL documentation. @vjagadish I still am tracking your comment to add more details on various queries, Samza SQL components. I think i need more time for those. If i can i will try to add them before Samza 1.0. If not we can target that for the next release. 

snappy

### DIFF
--- a/docs/learn/documentation/versioned/api/samza-sql.md
+++ b/docs/learn/documentation/versioned/api/samza-sql.md
@@ -20,33 +20,171 @@ title: Samza SQL
 -->
 
 
-# Section 1
+# Overview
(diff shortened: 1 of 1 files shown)


vjagadish1989:
>>  @vjagadish I still am tracking your comment to add more details on various queries, Samza SQL components. I think i need more time for those. If i can i will try to add them before Samza 1.0. If not we can target that for the next release.

Thanks @srinipunuru! We still have a week for code-freeze in 1.0, hence - would be great to have those changes in. However, if it ends up taking longer (or if you're working on something higher priority), we can track them for 1.0.1

vjagadish1989 (approved):
approved, looks good to me.

PR #696, opened 2018-10-06
SAMZA-1856: eventhub descriptors

dxichen:
@prateekm @lhaiesp 
Add system descriptors and input/output descriptors for Eventhubs.
We should make an effort to deprecate the `systems.%s.stream.list`.


samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubSystemDescriptor.java
```diff
@@ -0,0 +1,212 @@
...
+   * @param partitioningMethod
+   * @return this system descriptor
+   */
+  public EventHubSystemDescriptor withPartitioningMethod(PartitioningMethod partitioningMethod) {
+    this.producerEventhubsPartitioningMethod = Optional.of(partitioningMethod);
+    return this;
+  }
+
+  /**
+   *  Sending each message key to the eventhub in the properties of the AMQP message.
```
prateekm: Don't understand the first sentence, seems like an incomplete phrase. Can you clarify, both here and in config reference?

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubSystemDescriptor.java
```diff
@@ -0,0 +1,212 @@
...
+   *  If the Samza Eventhub consumer is used, this field is used as the message key if the partition key is not present.
+   *
+   * @param sendKeys Set to true if the message key should be send in the EventData properties, false otherwise
+   * @return this system descriptor
+   */
+  public EventHubSystemDescriptor withSendKeys(Boolean sendKeys) {
```
prateekm: use `boolean` so that this can't be null.

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubSystemDescriptor.java
```diff
@@ -0,0 +1,212 @@
...
+   *   key used for the hash should be a string. If the partition key is not set, the message key is
+   *   used instead.</li>
+   *   <li>PARTITION_KEY_AS_PARTITION:
+   *   Use the integer key specified by the partition key or key of the message to a specific partition
+   *   on Eventhub. If the integer key is greater than the number of partitions in the destination Eventhub,
+   *   a modulo operation will be performed to determine the resulting paritition.
+   *   ie. if there are 6 partitions and the key is 9, the message will end up in partition 3.
+   *   Similarly to EVENT_HUB_HASHING, if the partition key is not set the message key is used instead.</li>
+   * </ul>
+   * @param partitioningMethod
```
prateekm: Add documentation for `@param` and `@returns` everywhere.

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubSystemDescriptor.java
```diff
@@ -0,0 +1,212 @@
...
+  public EventHubSystemDescriptor withPartitioningMethod(PartitioningMethod partitioningMethod) {
+    this.producerEventhubsPartitioningMethod = Optional.of(partitioningMethod);
+    return this;
+  }
+
+  /**
+   *  Sending each message key to the eventhub in the properties of the AMQP message.
+   *  If the Samza Eventhub consumer is used, this field is used as the message key if the partition key is not present.
+   *
+   * @param sendKeys Set to true if the message key should be send in the EventData properties, false otherwise
```
prateekm: Nitpick: "set to true" (lower case s)

lhaiesp:
Thanks for the work. Agree that we should deprecate the stream.list config. But once we make that change, we need to make sure Brooklin connector adjust to that change as well. Also Brooklin EventHubs connector should use these descriptors.

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubSystemDescriptor.java
dxichen: Event Hubs uses eventData, which is a AMQP map under to hood. The kv-pair format from Samza is therefore not kept. Users may specify that they want to add the key in the map in addition to the value. I will clarify

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubOutputDescriptor.java
```diff
@@ -48,52 +50,39 @@
    * @param streamId id of the stream
    * @param serde serde for messages in the stream
    * @param systemDescriptor system descriptor this stream descriptor was obtained from
+   * @param namespace namespace for the Event Hubs entity to produce to, not null
```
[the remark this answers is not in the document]
dxichen: This is resolved, not sure why it still doesn't show the diff in the preview.

samza-azure/src/main/java/org/apache/samza/system/eventhub/EventHubSystemDescriptor.java
```diff
@@ -0,0 +1,191 @@
...
+
+  /**
+   * Constructs a {@link SystemDescriptor} instance.
+   *  @param systemName name of this system
+   */
+  public EventHubSystemDescriptor(String systemName) {
+    super(systemName, FACTORY_CLASS_NAME, null, null);
+  }
+
+  public <StreamMessageType> EventHubInputDescriptor<StreamMessageType> getInputDescriptor(String streamId, String namespace,
```
prateekm: Need to add javadoc for public methods, since they don't override the base class methods anymore.

PR #699, opened 2018-10-08
Fix potential null pointer issues in StreamOperatorTask

bharathkk:


samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -136,7 +136,9 @@ public void close() throws Exception {
     if (this.contextManager != null) {
       this.contextManager.close();
     }
-    operatorImplGraph.close();
+    if (operatorImplGraph != null) {
```
shanthoosh: Please correct me if i'm wrong here.

I think, this  problem occurs when a caller before invoking `init()` which sets `operatorImplGraph`, invokes `close()`(or `process()`)  method.

This problem can happen even in `process()` method as well. Not isolated to `close()` alone. Do we need to fix it as well, may be add `isInitialized` boolean guard in the method?

prateekm: process() will not be called before init() - that's Samza's API contract.

bharathkk: The ordering of the method is guaranteed. We always invoke init before process and close. The problem occurs when we run into an exception during `init` and failed to initialize `operatorImplGraph` and the shutdown sequence runs into a `NPE`.

shanthoosh: My bad, I missed this class was implementing just methods of `StreamTask`, `WindowableTask` etc. Thanks for pointing it out

PR #702, opened 2018-10-10
Application Descriptor implementation cleanup.

prateekm:
Major changes:
1. Move withDefaultSystem to the base ApplicationDescriptor interface so that it can be used for TaskApplication as well.

2. Move some of the common state (input/output/system/table descriptors) and corresponding helper methods from the StreamApplicationDescriptorImpl/TaskApplicationDescriptorImpl to the ApplicationDescriptor class.


prateekm:
cc @nickpan47 @cameronlee314 for review.

samza-core/src/main/java/org/apache/samza/application/StreamApplicationDescriptorImpl.java
```diff
@@ -286,8 +164,8 @@ public String getNextOpId(OpCode opCode, String userDefinedId) {
     }
 
     String nextOpId = String.format("%s-%s-%s-%s",
-        config.get(JobConfig.JOB_NAME()),
-        config.get(JobConfig.JOB_ID(), "1"),
+        getConfig().get(JobConfig.JOB_NAME()),
```
cameronlee314: Since you are changing here: can these use JobConfig.getName and JobConfig.getId?

samza-core/src/main/java/org/apache/samza/execution/JobGraph.java
```diff
@@ -248,7 +248,7 @@ private StreamEdge getOrCreateStreamEdge(StreamSpec streamSpec, boolean isInterm
     String streamId = streamSpec.getId();
     StreamEdge edge = edges.get(streamId);
     if (edge == null) {
-      boolean isBroadcast = appDesc.getBroadcastStreams().contains(streamId);
+      boolean isBroadcast = appDesc.getIntermediateBroadcastStreamIds().contains(streamId);
```
cameronlee314: Is it possible that the streamSpec is for a non-intermediate broadcast stream? Then it seems like this check would return false, but it should be true.

prateekm: > Is it possible that the streamSpec is for a non-intermediate broadcast stream
We currently don't support specifying broadcast inputs using input descriptors. 
SAMZA-1841 will address that later. Currently these stream IDs are those that were generated using the High Level API broadcast() operator, hence the renaming to clarify that this is only for intermediate streams.

Does that answer your question?

cameronlee314: Got it, so appDesc.getBroadcastStreams was already not including broadcast inputs. That makes sense that your change isn't changing anything, but does that mean there is currently an issue with getOrCreateStreamEdge which doesn't consider broadcast inputs through config?

PR #703, opened 2018-10-10
SAMZA-1947: Remove specific js/css files from rat excludes

shanthoosh:
Internally at Linkedin we've observed that excluding specific `js` and `css` through rat gradle-plugin causes the ligradle build to not run the unit-tests. 

To fix it, this patch removes those files from rat excludes and adds the samza license to the individual files.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,7 +67,6 @@ rat {
     '**/bootstrap.min.css',
     '**/bootstrap.min.js',
     '**/build/**',
-    '**/google-fonts.css',
     '**/ionicons.min.css',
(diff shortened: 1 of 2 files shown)


shanthoosh:
@prateekm 
please take a look when you get a chance. I verified that this unblocks the internal build problems.

prateekm:
@shanthoosh It's not OK to change existing licenses on third party dependencies and replace them with Apache license. Also, I'm not sure why adding a few files to rat excludes causes this failure. Let's debug more, and close this PR for now.

PR #706, opened 2018-10-11
SAMZA-1948 Updated hybrid table descriptors to take underlying table descriptors

weisong44:
Hybrid Table Descriptors, e.g. CachingTableDescriptor, should take both tables as a constructor param instead of using withXYZ methods. This is better because they are required for the hybrid descriptor to work, and it helps with type inference for the descriptor.


samza-core/src/main/java/org/apache/samza/table/caching/CachingTableDescriptor.java
```diff
@@ -47,10 +47,30 @@
   private boolean isWriteAround;
 
   /**
-   * {@inheritDoc}
+   * Constructs a table descriptor instance with internal cache
+   *
+   * @param tableId Id of the table, it must confirm to pattern { @literal [\\d\\w-_]+ }
```
pdu-mn1: nit: confirm -> conform

samza-core/src/main/java/org/apache/samza/table/caching/CachingTableDescriptor.java
```diff
@@ -47,10 +47,30 @@
...
+    this.table = table;
+  }
+
+  /**
+   * Constructs a table descriptor instance and specify a cache (as Table descriptor)
+   * to be used for caching. Cache get is not synchronized with put for better parallelism
+   * in the read path of {@link CachingTable}. As such, cache table implementation is
+   * expected to be thread-safe for concurrent accesses.
+   *
+   * @param tableId Id of the table, it must confirm to pattern { @literal [\\d\\w-_]+ }
```
pdu-mn1: nit: confirm -> conform

PR #707, opened 2018-10-11
Fix local side input table tests to use java.io.tmp directory

bharathkk:
Use java.io.tmp directory for tests instead of temporary folder inside module directories.


bharathkk:
@cameronlee314 please take a look

samza-test/src/test/java/org/apache/samza/test/table/TestLocalTableWithSideInputs.java
```diff
@@ -63,6 +64,8 @@
   private static final String PAGEVIEW_STREAM = "pageview";
   private static final String PROFILE_STREAM = "profile";
   private static final String ENRICHED_PAGEVIEW_STREAM = "enrichedpageview";
+  private static final File LOGGED_STORE_BASE_PATH = new File(System.getProperty("java.io.tmpdir"), "logged");
```
prateekm: Does Files.createTempDirectory work?

bharathkk: Are you referring to guava helper? Looks like its unstable and it does do the same thing except it chooses a directory name randomly and creates it under `java.io.tmpdir` and returns the path.
It is sufficient for the tests to provide the path alone while the actual creation is handled by the StorageManager.  Is there any other benefit to use `Files.createTempDirectory`?

prateekm: No, java.nio.Files#createTempDirectory.

bharathkk: The behavior of Files under `java.nio` seems similar as well. Is there added benefit in using it? I feel its better to lazily create directories as opposed to create directories upfront and especially in this scenario, TaskStorageManager does make sure thats the behavior.

prateekm: Looks like they return the same path, so you're right, this is better. 
Btw, do you need a random path on every test instead of hardcoding "logged"?  Or is this going to go away with Sanil's TestRunner changes?

bharathkk: It doesn't have to be a random path. logged and non-logged directories need to be different so that I can verify the directory used by the StorageEngine is the same as the directory used by StorageManager.

It will go away with Sanil's changes provided his test runner uses different directories for logged and non-logged stores. Otherwise, we still need this to ensure StorageEngine and StorageManager use the same directories due to scattered logic of determining store paths between SamzaContainer and StorageManager.

PR #708, opened 2018-10-11
System and Stream Descriptor API cleanup.

prateekm:
Major changes:
1. Made withPhysicalName(String) a protected method in StreamDescriptor. 
The primary reason to set the physical name is to use stream names with special characters in them, which streamId doesn't support. This change is to make it so that the physical name setter should only be exposed by systems where it means something useful - like HDFS.
2. Renamed some methods in StreamDescriptor for clarity.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/operators/descriptors/base/stream/InputDescriptor.java
+++ b/samza-api/src/main/java/org/apache/samza/operators/descriptors/base/stream/InputDescriptor.java
@@ -73,16 +73,15 @@ public abstract class InputDescriptor<StreamMessageType, SubClass extends InputD
   }
 
   /**
-   * If set to true, when a Samza container starts up, it ignores any checkpointed offset for this particular
+   * If set, when a Samza container starts up, it ignores any checkpointed offset for this particular
    * input stream. Its behavior is thus determined by the {@link #withOffsetDefault} setting.
    * Note that the reset takes effect every time a container is started, which may be every time you restart your job,
    * or more frequently if a container fails and is restarted by the framework.
    *
-   * @param resetOffset whether the container should ignore any checkpointed offset when starting
    * @return this input descriptor
    */
-  public SubClass withResetOffset(boolean resetOffset) {
-    this.resetOffsetOptional = Optional.of(resetOffset);
+  public SubClass shouldResetOffset() {
+    this.resetOffsetOptional = Optional.of(true);
     return (SubClass) this;
   }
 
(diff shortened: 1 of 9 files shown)


prateekm:
@cameronlee314 @nickpan47 Please take a look.

samza-api/src/main/java/org/apache/samza/operators/descriptors/base/stream/InputDescriptor.java
```diff
@@ -73,16 +73,15 @@ public InputDescriptor(String streamId, Serde serde, SystemDescriptor systemDesc
...
-   * If set to true, when a Samza container starts up, it ignores any checkpointed offset for this particular
+   * If set, when a Samza container starts up, it ignores any checkpointed offset for this particular
    * input stream. Its behavior is thus determined by the {@link #withOffsetDefault} setting.
    * Note that the reset takes effect every time a container is started, which may be every time you restart your job,
    * or more frequently if a container fails and is restarted by the framework.
    *
-   * @param resetOffset whether the container should ignore any checkpointed offset when starting
    * @return this input descriptor
    */
-  public SubClass withResetOffset(boolean resetOffset) {
```
cameronlee314: I do like the look of this proposed API better, but this is kind of now coupling with the fact that the default is "false", since this can't set it to false anymore. Is it a reasonable assumption that these all always will default to false? Or will someone need to remember to update this if the default changes?

samza-api/src/test/java/org/apache/samza/operators/descriptors/TestExpandingInputDescriptor.java
```diff
@@ -37,15 +37,13 @@ public void testAPIUsage() {
     ExampleExpandingOutputDescriptor<Integer> output1 = expandingSystem.getOutputDescriptor("output1", new IntegerSerde());
 
     input1
-        .withBootstrap(false)
+        .shouldBootstrap()
```
cameronlee314: Is it intentional to change from setting to false to setting to true for bootstrap and reset offset? Since this is just a "compile" test, it seems ok, but wanted to double check.

samza-api/src/main/java/org/apache/samza/operators/descriptors/base/stream/InputDescriptor.java
prateekm: Changing a default value would be a backwards incompatible change, so making a corresponding API change should be OK. In general going forward we will need to keep the configuration and descriptors in sync (including javadocs etc.).

samza-api/src/test/java/org/apache/samza/operators/descriptors/TestExpandingInputDescriptor.java
prateekm: Yeah, this was just for testing the API, so should be OK.

PR #709, opened 2018-10-11
SAMZA-1891: Updated javadoc to include config behavior

weisong44:
Instead of referring users to the Samza Configuration Guide, TableDescriptors should copy the documentation instead. If the goal is to let users configure their job primarily in code, it's simpler for them to see javadocs in their IDE than referring to a separate web page.


samza-core/src/main/java/org/apache/samza/operators/BaseTableDescriptor.java
```diff
@@ -45,28 +45,34 @@
...
+   * @param tableId Id of the table, it must conform to pattern {@literal [\\d\\w-_]+}
    * @param serde the serde for key and value
    */
   protected BaseTableDescriptor(String tableId, KVSerde<K, V> serde) {
     this.tableId = tableId;
     this.serde = serde;
   }
 
+  /**
+   * {@inheritDoc}
```
prateekm: Minor: Don't need @inheritDoc with @override, it already implies it.

samza-kv-rocksdb/src/main/java/org/apache/samza/storage/kv/RocksDbTableDescriptor.java
```diff
@@ -85,7 +94,17 @@ public RocksDbTableDescriptor withWriteBatchSize(int writeBatchSize) {
...
+   * Samza maintains an additional cache in front of RocksDB for frequently-accessed objects.
+   * This cache contains deserialized objects (avoiding the deserialization overhead on cache
+   * hits), in contrast to the RocksDB block cache ({@link #withCacheSize}), which caches
+   * serialized objects. This property determines the number of objects to keep in Samza's
+   * cache, per task instance. This same cache is also used for write buffering
+   * (see {@link #withWriteBatchSize}). A value of 0 disables all caching and batching.
+   * <p>
+   * Default value is 1,000.
+   * <p>
    * Refer to <code>stores.store-name.object.cache.size</code> in Samza configuration guide
```
prateekm: Prefer removing these "Refer to" comments.

PR #717, opened 2018-10-12
SAMZA-1951: Make JMX enabled/disabled using configs for Samza containers

Sanil15:


samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
```diff
@@ -199,4 +205,12 @@ public boolean getJmxEnabled() {
       return true;
     }
   }
+
+  public boolean getJmxEnabledOnContainer() {
```
prateekm: Why do these need to be separate configs?

Sanil15: `cluster-manager.jobcoordinator.jmx.enabled` is the config we use which can be misleading since it has jobcoordinator in it. I thought about it and wanted to rename this config to `cluster-manager.jmx.enabled` but we also have a legacy yarn config for the same. 

To avoid confusion I created a new, I agree one config for JC and Container but is renaming an existing config acceptable?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -791,6 +791,7 @@ class SamzaContainer(
   def hasStopped(): Boolean = status == SamzaContainerStatus.STOPPED || status == SamzaContainerStatus.FAILED
 
   def run {
+    val isJMXEnabledOnContainer = new ClusterManagerConfig(config).getJmxEnabledOnContainer
```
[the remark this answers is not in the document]
Sanil15: refer above

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
prateekm: Is there a need to enable/disable them separately? If not, let's use the existing config.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -800,8 +800,10 @@ class SamzaContainer(
 
       val startTime = System.nanoTime()
       status = SamzaContainerStatus.STARTING
-
-      jmxServer = new JmxServer()
+      val isJMXEnabled = new ClusterManagerConfig(config).getJmxEnabled
+      if (isJMXEnabled) {
```
shanthoosh: We report virtual machine runtime metrics via `JvmMetrics` from `SamzaContainer`. Have we tested if conditionally enabling/disabling `JmxServer` has any effect on that?

samza-core/src/main/scala/org/apache/samza/job/local/ThreadJobFactory.scala
```diff
@@ -73,7 +73,10 @@ class ThreadJobFactory extends StreamJobFactory with Logging {
     ChangelogStreamManager.createChangelogStreams(jobModel.getConfig, jobModel.maxChangeLogStreamPartitions)
 
     val containerId = "0"
-    val jmxServer = new JmxServer
+    var jmxServer: JmxServer = null
+    if (new ClusterManagerConfig(config).getJmxEnabled) {
+      jmxServer = new JmxServer();
```
shanthoosh: Since we're supporting turn off of JMX  for security reasons, it will be beneficial to add a trivial unit-test here, just to safe-guard against cases where this if-guard is unintentionally removed.

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
```diff
@@ -86,7 +86,7 @@
   private static final int DEFAULT_CONTAINER_RETRY_COUNT = 8;
 
   /**
-   * Determines whether a JMX server should be started on the job coordinator
+   * Determines whether a JMX server should be started on the job coordinator and container
```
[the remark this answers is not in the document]
Sanil15: These are the following ways:

- If we don't go with current config we Introduce a single config `job.jmx.enabled` for standalone and yarn. We will have to Deprecate configs like `cluster-manager.jobcoordinator.jmx.enabled` and `yarn.am.jmx.enabled`

- We work this out with two different configs, then container needs to refactor and do that check for standalone vs non-standalone deployement (this makes he code even messier)

- We go with the `cluster-manager.jobcoordinator.jmx.enabled` and document that this applies for standalone, this does not need deprecation

Is there a namespace for stanadalone only?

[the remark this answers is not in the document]
shanthoosh: Existing implementation of `ClusterManagerConfig.getJmxEnabled` checks the following two configuration keys: `cluster-manager.jobcoordinator.jmx.enabled`, `yarn.am.jmx.enabled` and returns true if either of them are true. If we use that config method for standalone , users can turn off/on the jmxServer via `yarn.am.jmx.enabled`. 

I think it'll be wise not to use yarn configuration for standalone  and simply document it. I think both `cluster-manager`, `yarn` namespace doesn't apply for standalone.

>> Is there a namespace for standalone only ? 
Yes, `job.coordinator.`  is predominantly used for most of the standalone configuration(or one can use `job.jmx.enabled` for this case or some other meaningful name). 

@prateekm what do you think?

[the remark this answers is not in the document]
prateekm: IIUC, there is a config to enable/disable JMX on the AM, but not one for the container, where it's always on. 
If so, we can add `job.jmx.enabled` with defaut=true, and use it consistently for both AM and containers. Should respect the values of existing configs. I.e., they should have higher precedence than job.jmx.enabled. Should mark both existing configs as deprecated in the configuration reference.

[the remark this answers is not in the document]
Sanil15: I propose a better plan (already did according to first commit)

That's what I did in the first commit, introduced a new config called `cluster-manager.container.jmx.enabled`, maybe I will rename it to `job.container.jmx.enabled`

and let the current config `cluster-manager.jobcoordinator.jmx.enabled`  manage JMX on JobCoordinator itself, this is more transparent in terms of handling JMX differently in JonCoordinator and Container since they apply differently in different ways of deploying a job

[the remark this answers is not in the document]
prateekm: This way we eventually end up with 2 configs, the other way we end up with one. Unless there's a reason to have separate configs for the AM and the containers, fewer configs seems better to me. WDYT?

samza-core/src/main/scala/org/apache/samza/job/local/ThreadJobFactory.scala
Sanil15: If this was passed a method param, I can use mockito for it. But i don't feel its worth to do that refactor to add a unit test for this. 

"Just to safe-guard against cases where this if-guard is unintentionally removed in the future" 
code reviews ?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
Sanil15: will check and circle back

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
```diff
@@ -88,10 +88,20 @@
   /**
    * Determines whether a JMX server should be started on the job coordinator
    * Default: true
+   *
+   * @deprecated use {@code JOB_JMX_ENABLED} instead
```
prateekm: Reorganize these fields. Put the job.jmx.enabled one on top, move these to below it with a note to use job.jmx.enabled instead, and remove duplicate javadoc.

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
```diff
@@ -189,12 +199,23 @@ public String getContainerManagerClass() {
...
+      log.info("Configuration {} is deprecated. Please use {}", AM_JMX_ENABLED, JOB_JMX_ENABLED);
       return getBoolean(AM_JMX_ENABLED);
+    } else if (containsKey(JOB_JMX_ENABLED)) {
+      return getBoolean(JOB_JMX_ENABLED);
+    } else {
+      return true;
+    }
+  }
+
+  public boolean getJmxEnabled() {
```
prateekm: `job.jmx.enabled` should be in JobConfig

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
Sanil15: sure

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
Sanil15: sure

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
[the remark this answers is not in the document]
Sanil15: I aggree

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -800,8 +800,9 @@ class SamzaContainer(
 
       val startTime = System.nanoTime()
       status = SamzaContainerStatus.STARTING
-
-      jmxServer = new JmxServer()
+      if (new JobConfig(config).getJMXEnabled) {
```
shanthoosh:
```suggestion
      if (config.getJMXEnabled) {
```

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
```diff
@@ -189,12 +197,15 @@ public String getContainerManagerClass() {
     return get(CLUSTER_MANAGER_FACTORY, CLUSTER_MANAGER_FACTORY_DEFAULT);
   }
 
-  public boolean getJmxEnabled() {
+  public boolean getJmxEnabledOnJobCoordinator() {
     if (containsKey(CLUSTER_MANAGER_JMX_ENABLED)) {
+      log.info("Configuration {} is deprecated. Please use {}", CLUSTER_MANAGER_JMX_ENABLED, JOB_JMX_ENABLED);
```
shanthoosh: This message should be logged at the warn log-level.

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
```diff
@@ -85,20 +85,28 @@
...
-
   /**
    * The cluster managed job coordinator sleeps for a configurable time before checking again for termination.
    * The sleep interval of the cluster managed job coordinator.
    */
   public static final String CLUSTER_MANAGER_SLEEP_MS = "cluster-manager.jobcoordinator.sleep.interval.ms";
   private static final int DEFAULT_CLUSTER_MANAGER_SLEEP_MS = 1000;
 
+  /**
+   * Determines whether a JMX server should be started on job coordinator and container
```
shanthoosh: Minor: 
s/job coordinator/JobCoordinator, s/container/SamzaContainer.
s/Default: true/By default set to true if undefined in the config.
```suggestion
   * Determines whether a JMX server should be started on JobCoordinator and SamzaContainer
```

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
```diff
@@ -85,20 +85,28 @@
...
+   */
+  public static final String JOB_JMX_ENABLED = "job.jmx.enabled";
+
+  /**
+   * Determines whether a JMX server should be started on the job coordinator
+   * Default: true
+   *
+   * @deprecated use {@code JOB_JMX_ENABLED} instead
+   */
+  public static final String AM_JMX_ENABLED = "yarn.am.jmx.enabled";
```
shanthoosh: Both of these static constants should be made private. Please correct me if i'm wrong, I don't see them being used anywhere outside of this class.
```suggestion
  private static final String AM_JMX_ENABLED = "yarn.am.jmx.enabled";
```

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
Sanil15: info is fine

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
Sanil15: I will prefer this

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
prateekm: @shanthoosh is right, please use warn since this configuration may go away in future.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
shanthoosh: > I will prefer this, 

I think each of our config class in scala defines implicit conversion methods. For instance, JobConfig.scala defines implicit conversion [here](https://github.com/apache/samza/blob/7780081ea147ee2854a982e020f9a5f6e0ab9fc8/samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala#L104). Can you please share the rationale why you prefer explicitly creating object  here?

shanthoosh: >> I will prefer this

I think each of our config class in scala defines implicit conversion methods. For instance, JobConfig.scala defines implicit conversion [here](https://github.com/apache/samza/blob/7780081ea147ee2854a982e020f9a5f6e0ab9fc8/samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala#L104). Can you please share the rationale why you prefer explicitly creating object here?

Sanil15: This will ensure, method is from the JobConfig subclass of Config and not any other subclass

shanthoosh: Majority of the implementation in `SamzaContainer` involving scala code follows this scala implicit conversion and does not create explicit object here. If I understand right, you're saying we will have different getJmxEnabled from other config class. Can you share the scenarios these can happen(to me at-least the reason sounds far-fetched).

vjagadish1989:
merged and submitted. Thanks @Sanil15 for the contribution and @shanthoosh for reviews!

PR #721, opened 2018-10-12
SAMZA-1937: Make DSL converter work when sql is a query

weiqingy:
To support the use case that users use Samza Sql Console or Samza Sql shell to perform SELECT query. 

## How was this patch tested?
1. Build passed and current unit tests passed
2. Testing in Samza Sql shell


weiqingy:
Hi @srinipunuru @nickpan47 @atoomula,

I've fixed review comments in https://github.com/apache/samza/pull/691 and the conflicts with master branch. Could you please review it again?

PR #728, opened 2018-10-15
SAMZA-1954: User provided configuration should have higher precedence than Samza generated configuration.

prateekm:


samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigurationGenerator.java
```diff
@@ -190,68 +198,27 @@ private long computeTriggerInterval(Collection<OperatorSpec> reachableOperators)
...
   }
 
-  private void configureTables(Map<String, String> configs, Config config, Map<String, TableSpec> tables, Set<String> inputs) {
-    configs.putAll(TableConfigGenerator.generateConfigsForTableSpecs(new MapConfig(configs),
-        tables.values().stream().collect(Collectors.toList())));
+  private void configureTables(Map<String, String> generatedConfig, Config originalConfig,
+      Map<String, TableSpec> tables, Set<String> inputs) {
+    generatedConfig.putAll(
+        TableConfigGenerator.generateConfigsForTableSpecs(
+            new MapConfig(generatedConfig), new ArrayList<>(tables.values())));
```
cameronlee314: Instead of passing generatedConfig, should this pass "config" (or even possibly config merged with the current state of generatedConfig)? I didn't look too closely into it, but something looks like it might be wrong based on the initial read.

samza-core/src/main/java/org/apache/samza/execution/JobPlanner.java
```diff
@@ -70,35 +70,20 @@ ExecutionPlan getExecutionPlan() {
...
 
-    // update application configs
-    Map<String, String> cfg = new HashMap<>();
-    if (StringUtils.isNoneEmpty(runId)) {
-      cfg.put(ApplicationConfig.APP_RUN_ID, runId);
+    // merge user-provided configuration with generated configuration. generated configuration has lower priority.
+    // TODO: This should all be consolidated with ExecutionPlanner after fixing SAMZA-1811
+    Map<String, String> allowedUserConfig = new HashMap<>(userConfig);
+    if (!LegacyTaskApplication.class.isAssignableFrom(appDesc.getAppClass())) {
+      allowedUserConfig.remove(TaskConfig.INPUT_STREAMS()); // must be set using descriptors or operators
```
cameronlee314: The previous code that did this had a log about the override. Would it be good to put that here too?

samza-core/src/main/java/org/apache/samza/execution/LocalJobPlanner.java
```diff
@@ -107,10 +107,10 @@ private void createStreams(String planId, List<StreamSpec> intStreams, StreamMan
     LOG.info("A single processor must create the intermediate streams. Processor {} will attempt to acquire the lock.", uid);
     // Move the scope of coordination utils within stream creation to address long idle connection problem.
     // Refer SAMZA-1385 for more details
-    JobCoordinatorConfig jcConfig = new JobCoordinatorConfig(config);
-    String coordinationId = new ApplicationConfig(config).getGlobalAppId() + APPLICATION_RUNNER_PATH_SUFFIX;
+    JobCoordinatorConfig jcConfig = new JobCoordinatorConfig(userConfig);
```
cameronlee314: Orthogonal question: Does intermediate stream creation need anything from the generated configs?

samza-core/src/main/java/org/apache/samza/execution/JobPlanner.java
prateekm: Sure, will add back logging.

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigurationGenerator.java
cameronlee314: Ah yes, your changes aren't changing any functionality. However, now that you've named these variables better, it looks like there might be an issue, since it would seem that `generateConfigsForTableSpecs` would want `config` and not the currently generated config map.

samza-core/src/main/java/org/apache/samza/execution/LocalJobPlanner.java
cameronlee314: Yep, I just wanted to make sure that the usage of the `userConfig` here does not need the generated config. Thanks for clarifying that actual intermediate stream creation does use generated configs later on.

samza-core/src/main/java/org/apache/samza/execution/JobNodeConfigurationGenerator.java
prateekm: Yeah, I'll create a ticket to fix that separately. Most of this TableSpec/TableProvider code should go away post 1.0 anyway.

PR #732, opened 2018-10-16
SAMZA-1938: Support use cases to run multiple sql statements in one Samza job

weiqingy:

This PR is to support the following user scenarios:
1. support users to run multiple Sql statements in the same job.
- sql1 = "Insert into testavro.simpleOutputTopic select * from testavro.SIMPLE1";
-  sql2 = "Insert into testavro.SIMPLE3 select * from testavro.SIMPLE2";

2. Support fan-out use case. For example, 
- sql1 = "Insert into testavro.SIMPLE2 select * from testavro.SIMPLE1";
- sql2 = "Insert into testavro.SIMPLE3 select * from testavro.SIMPLE1";

3. Support fan-in use case. For example,
- sql1 = "Insert into testavro.simpleOutputTopic select * from testavro.SIMPLE2";
- sql2 = "Insert into testavro.simpleOutputTopic select * from testavro.SIMPLE1";

This PR makes the information about `SystemDescriptors`,  `MessageStream of input sources` and `OutputStream` stored in `QueryTranslator`. The reason is that those information belongs to the job and each job has only one `QueryTranslator` object.

## How was this patch tested?
1. Add unit tests
2. Testing in Samza SQL shell.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplication.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplication.java
@@ -20,14 +20,20 @@
 package org.apache.samza.sql.runner;
 
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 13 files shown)


weiqingy:
Hi @srinipunuru @atoomula, could you please help to review this PR? Thanks.

PR #738, opened 2018-10-17
SAMZA-1888: Kafka consumer improvements

sborya:


samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -16,6 +16,24 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+/*
```
bharathkk: nit: duplicate license header.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -630,11 +601,58 @@ public void validateStream(StreamSpec streamSpec) throws StreamValidationExcepti
...
+  }
+
+  private Supplier<ZkUtils> getZkConnection() {
+    String zkConnect =
+        config.get(String.format(KafkaConfig.CONSUMER_CONFIGS_CONFIG_KEY(), systemName, ZOOKEEPER_CONNECT));
+    if (StringUtils.isBlank(zkConnect)) {
+      throw new SamzaException("Missing zookeeper.connect config for admin for system " + systemName);
+    }
+    return () -> {
+      return ZkUtils.apply(zkConnect, 6000, 6000, false);
```
bharathkk: Can be simplified to 
`
() -> ZkUtils.apply(zkConnect, 6000, 6000, false);
`

bharathkk: Do we need `sessionTimeout` and `connectionTimeout` to be configurable? If not, can we extract this to meaningful constants?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -630,11 +601,58 @@ public void validateStream(StreamSpec streamSpec) throws StreamValidationExcepti
...
+      }
       KafkaSystemAdminUtilsScala.deleteMessages(adminClient, offsets);
       deleteMessageCalled = true;
     }
   }
 
+  private Properties createAdminClientProperties() {
+    // populate brokerList from either consumer or producer configs
+    Properties props = new Properties();
+    // included SSL settings if needed
```
bharathkk: Do we want to make this method protected instead and have the extenders determine how they want to generate properties?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -564,7 +535,7 @@ public boolean clearStream(StreamSpec streamSpec) {
    * @param spec a StreamSpec object
    * @return KafkaStreamSpec object
    */
-  KafkaStreamSpec toKafkaSpec(StreamSpec spec) {
+  public KafkaStreamSpec toKafkaSpec(StreamSpec spec) {
```
bharathkk: why do we need this to be public?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -106,12 +108,13 @@ public KafkaSystemConsumer(Consumer<K, V> kafkaConsumer, String systemName, Conf
 
   /**
    * Create internal kafka consumer object, which will be used in the Proxy.
+   * @param <K> key type for the consumer
+   * @param <V> value type for the consumer
    * @param systemName system name for which we create the consumer
    * @param kafkaConsumerConfig config object for Kafka's KafkaConsumer
-   * @return KafkaConsumer object
+   * @return KafkaConsumer newly create kafka consumer object
```
bharathkk: nit: s/create/created

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -630,11 +601,58 @@ public void validateStream(StreamSpec streamSpec) throws StreamValidationExcepti
   @Override
   public void deleteMessages(Map<SystemStreamPartition, String> offsets) {
     if (deleteCommittedMessages) {
+      if (adminClient == null) {
+        adminClient = AdminClient.create(createAdminClientProperties());
+      }
       KafkaSystemAdminUtilsScala.deleteMessages(adminClient, offsets);
```
bharathkk: We need to invoke start on the adminClient after creation prior to deleting messages right?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
sborya: we discussed it before and decided that in this case we don't want any configs.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
sborya: yes. Will change.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
sborya: this method may be used from a different package.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
sborya: no, there is not start in adminClient.

PR #741, opened 2018-10-18
Changed TaskApplicationDescriptor APIs to be fluent and consistent with StreamApplicationDescriptor API.

prateekm:


prateekm:
@nickpan47 @rmatharu @bharathkk Please take a look.

samza-api/src/main/java/org/apache/samza/application/TaskApplication.java
```diff
@@ -50,9 +50,10 @@
  *     KafkaOutputDescriptor<PageViewEvent>> outputStreamDescriptor =
  *         trackingSystemDescriptor.getOutputDescriptor("recentPageViewEvent", new JsonSerdeV2<>(PageViewEvent.class)));
  *
- *     appDescriptor.addInputStream(inputStreamDescriptor);
- *     appDescriptor.addOutputStream(outputStreamDescriptor);
- *     appDescriptor.setTaskFactory((StreamTaskFactory) () -> new PageViewTask());
+ *     appDescriptor
+ *         .addInputStream(inputStreamDescriptor)
+ *         .addOutputStream(outputStreamDescriptor)
+ *         .withTaskFactory((StreamTaskFactory) () -> new PageViewTask());
```
rmatharu: Doesn't setTaskFactory seem more consistent with addInput and addOutput, 
vs. withTaskFactory consistent with withInput withOutput?

prateekm: Other methods in the base ApplicationDescriptor are already called withXYZ. We use the same convention in I/O and Table descriptors for setters as well, so prefer consistency.

I do think the `withTaskFactory, withInputStream and withOutputStream` set of methods makes sense. @nickpan47 WDYT?

nickpan47: +1. Sounds good to me.

samza-api/src/main/java/org/apache/samza/application/TaskApplication.java
```diff
@@ -50,9 +50,10 @@
  *     KafkaOutputDescriptor<PageViewEvent>> outputStreamDescriptor =
  *         trackingSystemDescriptor.getOutputDescriptor("recentPageViewEvent", new JsonSerdeV2<>(PageViewEvent.class)));
  *
- *     appDescriptor.addInputStream(inputStreamDescriptor);
- *     appDescriptor.addOutputStream(outputStreamDescriptor);
- *     appDescriptor.setTaskFactory((StreamTaskFactory) () -> new PageViewTask());
+ *     appDescriptor
+ *         .addInputStream(inputStreamDescriptor)
```
nickpan47: Can we change the comments here to withXXX().withXXX()?

prateekm: 👍 Will change them ti withInput/OutputStream.

PR #760, opened 2018-10-24
Javadoc cleanup for new Application, Descriptor, Context and Table APIs - Part 2

prateekm:
Currently, we don't allow imports for use only in javadocs. This requires using FQNs in link tags, which is not very readable. Checkstyle's UnusedImport rule has an option to allow imports for use in javadoc comments (processJavadocs=true, should be read as "check javadocs for import usage == true"). 

AFAICT, there's no good way to change the check's properties within a submodule. This PR adds both versions (strict and relaxed) to the checkstyle, and disables the strict validation for samza-api only.

This PR also updates the javadocs to use the class names with imports.

### DIFF
--- a/checkstyle/checkstyle.xml
+++ b/checkstyle/checkstyle.xml
@@ -34,7 +34,14 @@
   <module name="TreeWalker">
 
     <!-- code cleanup -->
-    <module name="UnusedImports"/>
+    <module name="UnusedImports">
+      <property name="id" value="allowJavadocUnusedImports"/>
+      <property name="processJavadoc" value="true"/>
+    </module>
+    <module name="UnusedImports"> <!-- suppressed for samza-api -->
+      <property name="id" value="preventJavadocUnusedImports"/>
+      <property name="processJavadoc" value="false"/>
+    </module>
     <module name="FileContentsHolder"/>
     <module name="RedundantImport"/>
     <module name="IllegalImport" />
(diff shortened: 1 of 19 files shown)


prateekm:
@vjagadish1989 @cameronlee314

checkstyle/checkstyle.xml
```diff
@@ -34,7 +34,14 @@
   <module name="TreeWalker">
 
     <!-- code cleanup -->
-    <module name="UnusedImports"/>
+    <module name="UnusedImports">
+      <property name="id" value="allowJavadocUnusedImports"/>
+      <property name="processJavadoc" value="true"/>
```
cameronlee314: That's nice that this granularity is available. What do you think about just enabling this flag for all modules and deleting the "preventJavadocUnusedImports" rule/supression?
From my understanding, if you have rules for processJavadoc as true and false, then the stricter one just always applies anyways, so this new "allowJavadocUnusedImports" doesn't actually do anything right now. The change that is allowing the build to pass is that you did the suppression.

prateekm: For normal classes it's an AND of strict and relaxed, so it's strict. For samza-api, I disabled strict specifically, so it's relaxed. I wanted to make a minimal change from existing behavior, but I'm in favor of relaxing this check everywhere too.

@nickpan47 WDYT? Are you OK with relaxing checkstyle unused imports check in Samza such that it adding imports for reference in Javadocs?

cameronlee314: Ah right, I had a feeling that I was missing something. If you didn't add the "allowJavadocUnusedImports", then samza-api would have no unused imports checks at all, but we still would want a check for fully unused imports.

PR #764, opened 2018-10-26
Samza 1967: Tests failing when Job uses any serde other than NoOp

Sanil15:
Context: Serde is configured in JobNodeConfigurationGenerator and any StreamDescriptor#toConfig does not generate key and msg serde configs

Problem: Tests failing when Job uses any serde other than NoOp, since ApplicationDescriptor serdes take precedence in absence of any user-supplied configs

Solution: Passing null msg and key serde configs in userConfigs for StreamDescriptors ensures ApplicationDescriptor generated serde configs don't take precedence 

@prateekm @rmatharu please take a look

### DIFF
--- a/samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
+++ b/samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
@@ -39,6 +39,7 @@ import org.apache.samza.config.InMemorySystemConfig;
 import org.apache.samza.config.JobConfig;
 import org.apache.samza.config.JobCoordinatorConfig;
 import org.apache.samza.config.MapConfig;
+import org.apache.samza.config.StreamConfig;
 import org.apache.samza.config.TaskConfig;
 import org.apache.samza.container.grouper.task.SingleContainerGrouperFactory;
 import org.apache.samza.job.ApplicationStatus;
@@ -55,6 +56,7 @@ import org.apache.samza.system.SystemProducer;
 import org.apache.samza.system.SystemStream;
 import org.apache.samza.system.SystemStreamMetadata;
 import org.apache.samza.system.SystemStreamPartition;
+import org.apache.samza.system.descriptors.StreamDescriptor;
 import org.apache.samza.system.inmemory.InMemorySystemFactory;
 import org.apache.samza.task.AsyncStreamTask;
 import org.apache.samza.task.StreamTask;
@@ -167,7 +169,7 @@ public class TestRunner {
    */
   public TestRunner addConfig(String key, String value) {
     Preconditions.checkNotNull(key);
-    Preconditions.checkNotNull(value);
+    //Preconditions.checkNotNull(value);
     configs.put(key, value);
     return this;
   }
@@ -243,6 +245,7 @@ public class TestRunner {
         .createStream(spec);
     addConfig(streamDescriptor.toConfig());
     addConfig(streamDescriptor.getSystemDescriptor().toConfig());
+    addSerdeConfigs(streamDescriptor);
     return this;
   }
 
@@ -359,6 +362,7 @@ public class TestRunner {
     imsd.withInMemoryScope(this.inMemoryScope);
     addConfig(descriptor.toConfig());
     addConfig(descriptor.getSystemDescriptor().toConfig());
+    addSerdeConfigs(descriptor);
     StreamSpec spec = new StreamSpec(descriptor.getStreamId(), streamName, systemName, partitionData.size());
     SystemFactory factory = new InMemorySystemFactory();
     Config config = new MapConfig(descriptor.toConfig(), descriptor.getSystemDescriptor().toConfig());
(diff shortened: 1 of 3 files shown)


samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -391,4 +395,17 @@ private void deleteDirectory(String path) {
...
+  /**
+   * Test Framework only supports NoOpSerde. This method ensures null key and msg serde config for input and output streams
+   * takes preference when configs are merged in {@link org.apache.samza.execution.JobPlanner#getExecutionPlan}
+   * over {@link org.apache.samza.application.descriptors.ApplicationDescriptor} generated configs
+   */
+  private void addSerdeConfigs(StreamDescriptor descriptor) {
+    String streamIdPrefix = String.format(StreamConfig.STREAM_ID_PREFIX(), descriptor.getStreamId());
+    String keySerdeConfigKey = streamIdPrefix + StreamConfig.KEY_SERDE();
+    String msgSerdeConfigKey = streamIdPrefix + StreamConfig.MSG_SERDE();
+    this.configs.put(keySerdeConfigKey, null);
```
rmatharu: Is it possible to handle this use-case without overloading and using "null", e.g., by 
defining a TestFrameworkSerde class. 
"null" in code usually signifies things gone wrong. 
see https://goo.gl/eaxLTp

Sanil15: I agree with https://goo.gl/eaxLTp

The issue here is NoOpSerde is not an actual value in code (would have used that) with a facory to produce it, its absence of a serde, if we handle it the code consistently as a value everywhere, that's the right fix to do. That needs bigger refactor and more testing (using NoOpSerde applies a wider use case)

This problem will go away when I do a cleanup PR and replace appDescriptors with in-memory descriptors. This is just a temporary fix, discussed with @prateekm

Sanil15: 1. It is documented that we only support NoOpsSerde for now. That is intentionally done so that even if users pass a valid one, it always ends up using NoOp

2. for three cases, irrelevant of api you write test in (low level api, high level api, legacy low level)
- In the client library: for example, the library for publishing to Kafka and consuming from Kafka supports pluggable serialization.
A: Does not apply to InMemory System

- In the task implementation: your process method can use raw byte arrays as inputs and outputs, and do any parsing and serialization itself.
- Between the two: Samza provides a layer of serializers and deserializers, or serdes for short.
A: All streams are configured to have NoOpSerde only internally, that means test fw works with deserialized objects only

shanthoosh: @Sanil15 From your response, I think we only support `NoOpSerde` in `TestRunner` API. So the serde part question doesn't matter for which API.

I think then we wouldn't have to consider all the above cases for both low-level and high-level API.  Let's document this and provide examples of using the TestRunner API when users define custom serde's.

shanthoosh: @Sanil15 Just curious, Do you know how does [IntermediateStreamSerde](https://github.com/apache/samza/blob/master/samza-core/src/main/java/org/apache/samza/serializers/IntermediateMessageSerde.java), [ControlMessageSerde](https://github.com/apache/samza/blob/master/samza-api/src/main/java/org/apache/samza/system/ControlMessage.java) works with  `TestRunner`, since we only support NoOpSerde with `TestRunner`. Is the whole message converted to bytes before it reaches the InmemorySystem?

Sanil15: Discussed Offline

prateekm: @shanthoosh Why are intermediate stream and control message serde relevant here? They're framework internal and not used for inputs. 

Also, "test runner only supports NoOpSerde" doesn't mean much, if anything. It'll be clearer to clarify what's expected of the user - that the inputs to any in memory streams must be already deserialized.

samza-test/src/test/java/org/apache/samza/test/framework/StreamApplicationIntegrationTest.java
```diff
@@ -135,10 +135,10 @@ public void describe(StreamApplicationDescriptor appDescriptor) {
               KVSerde.of(new IntegerSerde(), new TestTableData.ProfileJsonSerde())));
 
       KafkaSystemDescriptor ksd = new KafkaSystemDescriptor("test");
-      KafkaInputDescriptor<TestTableData.Profile> profileISD = ksd.getInputDescriptor("Profile", new NoOpSerde<>());
+      KafkaInputDescriptor<TestTableData.Profile> profileISD = ksd.getInputDescriptor("Profile", new JsonSerdeV2<>());
```
prateekm: Set a KVSerde.of() so that this can test key serde usage as well.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: sure

samza-test/src/test/java/org/apache/samza/test/framework/StreamApplicationIntegrationTest.java
Sanil15: sure

PR #774, opened 2018-10-26
Samza-sql: Change Calcite sql type for samza sql rel message __key__ to accept any format

atoomula:


samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
```diff
@@ -76,45 +80,60 @@ public AvroRelConverter(SystemStream systemStream, AvroRelSchemaProvider schemaP
...
     } else {
       String msg = "Avro message converter doesn't support messages of type " + value.getClass();
       LOG.error(msg);
       throw new SamzaException(msg);
     }
 
-    return new SamzaSqlRelMessage(samzaMessage.getKey(), fieldNames, fieldValues);
+    Object key = samzaMessage.getKey();
+    // Key could be an avro record or could just be any other object. Handle avro here.
+    if (key instanceof IndexedRecord) {
```
srinipunuru: The fact that a key can be an Avro should be known to the RelMessageConverter. I don't think we need to have this if else in this layer,Rather it would be cleaner if you move this up to a layer above which uses AvroRelMessageConverter like BrooklinRelMessageConverter or KafkaRelMessageConverter.  It is ok if this layer handles keys of type String as long as there is a clean way to override the behavior.

PR #781, opened 2018-10-30
Add physical name support for InMemoryStreamDescriptors

Sanil15:


Sanil15:
@prateekm please have a look

PR #784, opened 2018-10-30
Fix to make Samza SQL applications work after the Runner refactoring

srinipunuru:
With recent change in Samza, Constructor signature for ApplicationRunner has changed. But the SamzaSQLApplicationRunner was not updated with the new signature. This is to fix the signature of the constructor for SamzaSQLApplicationRunner with the updated signature.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationRunner.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationRunner.java
@@ -1,21 +1,21 @@
 /*
-* Licensed to the Apache Software Foundation (ASF) under one
-* or more contributor license agreements.  See the NOTICE file
-* distributed with this work for additional information
-* regarding copyright ownership.  The ASF licenses this file
-* to you under the Apache License, Version 2.0 (the
-* "License"); you may not use this file except in compliance
-* with the License.  You may obtain a copy of the License at
-*
-*   http://www.apache.org/licenses/LICENSE-2.0
-*
-* Unless required by applicable law or agreed to in writing,
-* software distributed under the License is distributed on an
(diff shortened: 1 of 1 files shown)


srinipunuru:
@atoomula @prateekm Can you take a look at this one?

weiqingy:
How did you find this issue?

srinipunuru:
> How did you find this issue?

I didn't find it. Cameron found it accidentally while he was looking at the code.

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationRunner.java
```diff
@@ -56,9 +57,20 @@
   public static final String RUNNER_CONFIG = "app.runner.class";
   public static final String CFG_FMT_SAMZA_STREAM_SYSTEM = "streams.%s.samza.system";
 
+  /**
+   * NOTE: This constructor is called from {@link ApplicationRunners} through reflection.
+   * Please refrain from updating the signature or removing this constructor unless the caller has changed the interface.
+   */
+  public SamzaSqlApplicationRunner(SamzaApplication app, Config config) {
+    this(app, false, config);
```
prateekm: @srinipunuru Does this assume that this will only be used in YARN? ApplicationRunners#getApplicationRunner will be used for both YARN and Standalone. What does setting localRunner == false do?

Let's also sync up offline about why we need a separate runner for SQL. Maybe we can reuse the existing runners.

srinipunuru: @prateekm Yes it will be used just in yarn. Regarding the need for separate runner we can chat about it. I can't remember who recommended this approach, But this was recommended approach to do config translation rather than use config rewriter. I had to rewrite the config rewriter as applicationRunner.

PR #786, opened 2018-10-30
SAMZA-1971: Fix NPE in partition key computation for InMemorySystemProducer

bharathkk:


Sanil15:
Thanks for the change, lgtm!

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemProducer.java
```diff
@@ -75,11 +75,21 @@ public void send(String source, OutgoingMessageEnvelope envelope) {
...
+    // key as the final resort.
+    if (envelope.getPartitionKey() != null) {
+      partitionKey = envelope.getPartitionKey();
+    } else if (key != null) {
+      partitionKey = key;
+    } else {
+      partitionKey = message;
+    }
+
+    Preconditions.checkNotNull(partitionKey, "Failed to compute partition key for the message: " + envelope);
```
Sanil15: Consider to throw a  SamzaException on this case, rather than preconditions

PR #788, opened 2018-10-31
SAMZA-1970: Support for physical names in InMemorySystem

Sanil15:
if super is not there, java compiles this to this.withPhysicalName which results in StackOverflows


Sanil15:
@vjagadish1989 please merge when you have time

samza-test/src/main/java/org/apache/samza/test/framework/system/descriptors/InMemoryOutputDescriptor.java
```diff
@@ -45,7 +45,7 @@
   }
 
   public InMemoryOutputDescriptor withPhysicalName(String physicalName) {
-    withPhysicalName(physicalName);
+    super.withPhysicalName(physicalName);
```
shanthoosh: Please correct me If i'm wrong here. Before this change, was this method going into a indefinite recursive call?

Sanil15: yes if this is used, I mentioned this in a comment above

PR #790, opened 2018-11-01
SAMZA-1973: Unify the TaskNameGrouper interface for yarn and standalone.

shanthoosh:
This patch consists of the following changes:
* Unify the different methods present in the TaskNameGrouper interface. This will enable us to have a single interface method usable for both the yarn and standalone models.
* Generate locationId aware task assignment to processors in standalone.
* Move the task assignment persistence logic from a custom `TaskNameGrouper` implementation to `JobModelManager`, so that this works for any kind of custom group.
* General code clean up in `JobModelManager`,  `TaskAssignmentManager` and in other samza internal classes.
* Read/write taskLocality of the processors in standalone. 
* Updated the existing java docs and added java docs where they were missing.

Testing: 
* Fixed the existing unit-tests due to the changes. 
* Added new unit tests for the functionality changed added as a part of this patch. 
* Tested this patch with a sample job from `hello-samza` project and verified that it works as expected.

Please refer to [SEP-11](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75957309) for more details.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/task/GrouperContext.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/task/GrouperContext.java
@@ -34,16 +34,24 @@ import java.util.Map;
  */
 @InterfaceStability.Evolving
 public class GrouperContext {
-  private final Map<String, LocationId> processorLocality;
-  private final Map<TaskName, LocationId> taskLocality;
-  private final Map<TaskName, List<SystemStreamPartition>> previousTaskToSSPAssignment;
-  private final Map<TaskName, String> previousTaskToContainerAssignment;
 
-  public GrouperContext(Map<String, LocationId> processorLocality, Map<TaskName, LocationId> taskLocality, Map<TaskName, List<SystemStreamPartition>> previousTaskToSSPAssignments, Map<TaskName, String> previousTaskToContainerAssignment) {
+  // Map of processorId to LocationId.
+  private final Map<String, LocationId> processorLocality;
+
+  // Map of TaskName to LocationId.
+  private final Map<TaskName, LocationId> taskLocality;
+
+  // Map of TaskName to a list of the input SystemStreamPartition's assigned to it.
+  private final Map<TaskName, List<SystemStreamPartition>> previousTaskToSSPAssignment;
+
+  // Map of TaskName to ProcessorId.
+  private final Map<TaskName, String> previousTaskToProcessorAssignment;
+
+  public GrouperContext(Map<String, LocationId> processorLocality, Map<TaskName, LocationId> taskLocality, Map<TaskName, List<SystemStreamPartition>> previousTaskToSSPAssignments, Map<TaskName, String> previousTaskToProcessorAssignment) {
     this.processorLocality = Collections.unmodifiableMap(processorLocality);
     this.taskLocality = Collections.unmodifiableMap(taskLocality);
     this.previousTaskToSSPAssignment = Collections.unmodifiableMap(previousTaskToSSPAssignments);
-    this.previousTaskToContainerAssignment = Collections.unmodifiableMap(previousTaskToContainerAssignment);
+    this.previousTaskToProcessorAssignment = Collections.unmodifiableMap(previousTaskToProcessorAssignment);
   }
 
   public Map<String, LocationId> getProcessorLocality() {
@@ -62,7 +70,7 @@ public class GrouperContext {
     return new ArrayList<>(processorLocality.keySet());
   }
 
-  public Map<TaskName, String> getPreviousTaskToContainerAssignment() {
-    return this.previousTaskToContainerAssignment;
+  public Map<TaskName, String> getPreviousTaskToProcessorAssignment() {
+    return this.previousTaskToProcessorAssignment;
   }
 }
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/task/GrouperContext.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/task/GrouperContext.java
@@ -34,28 +34,28 @@ import java.util.Map;
  */
 @InterfaceStability.Evolving
 public class GrouperContext {
-  private Map<String, LocationId> processorLocality;
-  private Map<TaskName, LocationId> taskLocality;
-  private Map<TaskName, List<SystemStreamPartition>> previousTaskToSSPAssignment;
-  private Map<TaskName, String> previousTaskToContainerAssignment;
+  private final Map<String, LocationId> processorLocality;
+  private final Map<TaskName, LocationId> taskLocality;
+  private final Map<TaskName, List<SystemStreamPartition>> previousTaskToSSPAssignment;
+  private final Map<TaskName, String> previousTaskToContainerAssignment;
 
   public GrouperContext(Map<String, LocationId> processorLocality, Map<TaskName, LocationId> taskLocality, Map<TaskName, List<SystemStreamPartition>> previousTaskToSSPAssignments, Map<TaskName, String> previousTaskToContainerAssignment) {
-    this.processorLocality = processorLocality;
-    this.taskLocality = taskLocality;
-    this.previousTaskToSSPAssignment = previousTaskToSSPAssignments;
-    this.previousTaskToContainerAssignment = previousTaskToContainerAssignment;
+    this.processorLocality = Collections.unmodifiableMap(processorLocality);
+    this.taskLocality = Collections.unmodifiableMap(taskLocality);
+    this.previousTaskToSSPAssignment = Collections.unmodifiableMap(previousTaskToSSPAssignments);
+    this.previousTaskToContainerAssignment = Collections.unmodifiableMap(previousTaskToContainerAssignment);
   }
 
   public Map<String, LocationId> getProcessorLocality() {
-    return Collections.unmodifiableMap(processorLocality);
+    return processorLocality;
   }
 
   public Map<TaskName, LocationId> getTaskLocality() {
-    return Collections.unmodifiableMap(taskLocality);
+    return taskLocality;
   }
 
   public Map<TaskName, List<SystemStreamPartition>> getPreviousTaskToSSPAssignment() {
-    return Collections.unmodifiableMap(previousTaskToSSPAssignment);
+    return previousTaskToSSPAssignment;
   }
 
   public List<String> getProcessorIds() {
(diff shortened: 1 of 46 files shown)


shanthoosh:
@sborya @xinyuiscool @prateekm 

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -131,14 +140,16 @@
     this.barrier =  new ZkBarrierForVersionUpgrade(zkUtils.getKeyBuilder().getJobModelVersionBarrierPrefix(), zkUtils, new ZkBarrierListenerImpl(), debounceTimer);
     systemAdmins = new SystemAdmins(config);
     streamMetadataCache = new StreamMetadataCache(systemAdmins, METADATA_CACHE_TTL_MS, SystemClock.instance());
+    LocationIdProviderFactory locationIdProviderFactory = Util.getObj(new JobConfig(config).getLocationIdProviderFactory(), LocationIdProviderFactory.class);
```
sborya: can this code moved up in the class hierarchy? it is same everywhere.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -19,98 +19,182 @@
...
+  @volatile var currentJobModelManager: JobModelManager = _
   val jobModelRef: AtomicReference[JobModel] = new AtomicReference[JobModel]()
 
   /**
-   * Does the following actions for a job.
+   * Does the following:
    * a) Reads the jobModel from coordinator stream using the job's configuration.
-   * b) Recomputes changelog partition mapping based on jobModel and job's configuration.
+   * b) Recomputes the changelog partition mapping based on jobModel and job's configuration.
    * c) Builds JobModelManager using the jobModel read from coordinator stream.
```
sborya: Put a note, that this is only for Yarn case.

samza-core/src/test/java/org/apache/samza/container/grouper/task/TestGroupByContainerIds.java
```diff
@@ -114,7 +105,8 @@ public void testGroupHappyPath() {
   public void testGroupWithNullContainerIds() {
```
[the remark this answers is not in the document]
shanthoosh: Yes. This was not introduced by this patch. It was done to test the grouper with null containerIds.

samza-core/src/main/java/org/apache/samza/container/grouper/task/GrouperContext.java
```diff
@@ -0,0 +1,68 @@
...
+
+  public Map<String, LocationId> getProcessorLocality() {
+    return processorLocality;
+  }
+
+  public Map<TaskName, LocationId> getTaskLocality() {
+    return taskLocality;
+  }
+
+  public Map<TaskName, List<SystemStreamPartition>> getPreviousTaskToSSPAssignment() {
```
bharathkk: unused method?

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -19,98 +19,183 @@
...
+    val containerToLocationId: util.Map[String, LocationId] = new util.HashMap[String, LocationId]()
+    val existingContainerLocality = localityManager.readContainerLocality()
+
+    for (containerId <- 0 to config.getContainerCount) {
+      val localityMapping = existingContainerLocality.get(containerId.toString)
+      // To handle the case when the container count is increased between two different runs of a samza-yarn job,
+      // set the locality of newly added containers to any_host.
+      var locationId: LocationId = new LocationId("ANY_HOST")
+      if (localityMapping != null) {
+        locationId = new LocationId(localityMapping.get(SetContainerHostMapping.HOST_KEY))
```
bharathkk: Is it guaranteed that `localityMapping.get(SetContainerHostMapping.HOST_KEY)` returns non-null value? Otherwise we would be throwing an IllegalArgumentException instead of using `ANY_HOST` as the location id. Do we want to fall back to `ANY_HOST`?

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskNameGrouper.java
```diff
@@ -44,15 +43,36 @@
...
+   * Group the taskModels into set of {@link ContainerModel}.
    *
-   * @param tasks Set of tasks to group into containers.
-   * @return Set of containers, which contain the tasks that were passed in.
+   * @param taskModels the set of {@link TaskModel} to group into containers.
+   * @return the grouped {@link ContainerModel} built from the provided taskModels.
    */
-  Set<ContainerModel> group(Set<TaskModel> tasks);
+  @Deprecated
+  Set<ContainerModel> group(Set<TaskModel> taskModels);
```
bharathkk: Why are we deprecating these? Are we expecting that non-host affinity jobs also require some metadata from the grouper context to perform grouping?

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskNameGrouper.java
```diff
@@ -44,15 +43,36 @@
...
 public interface TaskNameGrouper {
+
+  /**
+   * Groups the taskModels into set of {@link ContainerModel} using the provided grouperContext.
+   *
+   * @param taskModels the set of tasks to group into containers.
+   * @param grouperContext the existing historical metadata of the samza job.
+   * @return the grouped {@link ContainerModel} built from the provided taskModels.
+   */
+  default Set<ContainerModel> group(Set<TaskModel> taskModels, GrouperContext grouperContext) {
```
bharathkk: Why is the GrouperContext part of the group method? I feel like the context is more of a metadata that the implementor needs in order to perform grouping. It may well be useful beyond the grouping path like some setup before performing grouping.

Although I don't like the idea of having lifecycle methods, can we have an init method that takes the context and keep the group method clean? Are there any implications that requires the current pattern?

Ideally we can pass in the context during construction though I am not sure if it is going to complicate things here since we use reflection.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -200,22 +299,18 @@ object JobModelManager extends Logging {
...
-        case _ => containerGrouper.group(taskModels.asJava, containerIds)
-      }
-    }
-    val containerMap = containerModels.asScala.map { case (containerModel) => containerModel.getId -> containerModel }.toMap
-
-    if (isHostAffinityEnabled) {
-      new JobModel(config, containerMap.asJava, localityManager)
+    var containerModels: util.Set[ContainerModel] = null
+    if(isHostAffinityEnabled) {
+      containerModels = containerGrouper.group(taskModels, grouperContext)
```
bharathkk: It would be much cleaner to have a single path here and let the grouper decide on how it wants to perform grouping? I think host affinity is a detail that shouldn't determine which grouping choice to do in the JobModelManager since that is sort of leaking the grouper responsibility into JobModelManager.

IMO, it is cleaner to have the JobModelManager prepare the context for the grouper and hand it off to the grouper which in turn determines how to perform grouping based on the context it has.

Let me know what you think?

samza-core/src/main/java/org/apache/samza/zk/ZkUtils.java
```diff
@@ -239,6 +243,23 @@ private boolean isValidRegisteredProcessor(final ProcessorNode processor) {
...
+    validatePaths(new String[] {taskLocalityPath});
+    writeData(taskLocalityPath, locationId.getId());
+  }
+
+  public Map<TaskName, LocationId> readTaskLocality() {
+    Map<TaskName, LocationId> taskLocality = new HashMap<>();
+    List<String> tasks = zkClient.getChildren(keyBuilder.getTaskLocalityPath());
+    for (String taskName : tasks) {
+      String taskPath = String.format("%s/%s", keyBuilder.getTaskLocalityPath(), taskName);
+      String locationId = zkClient.readData(taskPath);
```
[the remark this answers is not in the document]
prateekm: Should this be stored using the metadata store API? Please sync up with Ray about how and where to read/write locality (both task and container) before committing.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -977,16 +978,13 @@ class SamzaContainer(
 
   def storeContainerLocality {
     val isHostAffinityEnabled: Boolean = new ClusterManagerConfig(config).getHostAffinityEnabled
-    if (isHostAffinityEnabled) {
-      val localityManager: LocalityManager = new LocalityManager(config, containerContext.getContainerMetricsRegistry)
+    if (isHostAffinityEnabled && localityManager != null) {
```
prateekm: Why would this be null if host affinity is enabled?

samza-core/src/main/java/org/apache/samza/container/grouper/task/MetadataProvider.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.samza.runtime.LocationId;
+import org.apache.samza.system.SystemStreamPartition;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Provides the historical metadata of the samza job.
+ */
+@InterfaceStability.Evolving
+public interface MetadataProvider {
```
prateekm: This sounds like a very general/vague API, and is not conceptually related to MetadataStore. Can we call this GrouperParams/GrouperInfo/GrouperMetadata or something similar? Even GrouperContext is fine if you don't like any of the above, but would definitely prefer not calling this MetadataProvider.

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskGroup.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.*;
+
+/**
+ * A mutable group of tasks and an associated container id.
+ *
+ * Used as a temporary mutable container until the final ContainerModel is known.
+ */
+class TaskGroup {
+  final List<String> taskNames = new LinkedList<>();
```
prateekm: Can these be private?

shanthoosh: Yes. Modified the visibility.

samza-core/src/main/java/org/apache/samza/container/grouper/task/MetadataProvider.java
shanthoosh: Done.

vjagadish1989:
Fix build failures?

vjagadish1989:
Looks like the build failures were unrelated to this PR.

PR #791, opened 2018-11-02
SAMZA-1976: MetadataStore API cleanup.

shanthoosh:
This PR consists of the following changes:
* Switching all the API methods from using byte[] array as key type to string. 
* Fixed `CoordinatorMetadataStore`, `ZkMetadataStore` tests due to the type change of key.

Shortly in a followup PR,  namespace unification for different metadata stored in standalone and YARN model will be done.


shanthoosh:
@prateekm @xinyuiscool @sborya 

PR #794, opened 2018-11-05
SAMZA-1616: Samza-Sql - Support remote table for stream-table join

atoomula:


samza-sql-shell/src/main/java/org/apache/samza/sql/client/impl/SamzaExecutor.java
```diff
@@ -390,6 +390,9 @@ static void saveOutputMessage(OutgoingMessageEnvelope messageEnvelope) {
     String avroSamzaToRelMsgConverterDomain =
         String.format(SamzaSqlApplicationConfig.CFG_FMT_SAMZA_REL_CONVERTER_DOMAIN, "avro");
 
+    String avroSamzaToRelTableKeyConverterDomain =
```
srinipunuru: Where is avroSamzaToRelTableKeyConverterDomain being used?

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SamzaRelTableKeyConverter.java
```diff
@@ -0,0 +1,39 @@
...
+ * The {@link SamzaRelTableKeyConverter} is configurable at a system level, so it is possible to configure different
+ * {@link SamzaRelTableKeyConverter} for different remote table systems.
+ */
+public interface SamzaRelTableKeyConverter {
+  /**
+   * Convert the key in relational record format to the format expected by remote table.
+   * @param relKeyRecord key relational record that needs to be converted.
+   * @return the table key
+   */
+  Object convertToTableKeyFormat(SamzaSqlRelRecord relKeyRecord);
```
srinipunuru: How does this work if the table just supports primitives. Do we require the users to always create a SamzaSqlRelRecord even if the key is a primitive type?

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SamzaRelTableKeyConverterFactory.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * Create a {@link SamzaRelTableKeyConverter}. This method is called when the framework wants to create the
+   * {@link SamzaRelTableKeyConverter} corresponding to the system.
+   * @param systemStream the systemStream to create a key converter for
+   * @param relSchemaProvider the relational schema provider, if there is one
+   * @param config config that is used to create the object
+   * @return the object created.
+   */
+  SamzaRelTableKeyConverter create(SystemStream systemStream, RelSchemaProvider relSchemaProvider, Config config);
```
srinipunuru: Right now relSchemaProvider only returns the payloadSchema, How is the RelSchemaProvider used in the TableKeyConverter?

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplication.java
```diff
@@ -69,13 +66,13 @@ public void describe(StreamApplicationDescriptor appDescriptor) {
       // 3. Translate Calcite plan to Samza stream operators
       QueryTranslator queryTranslator = new QueryTranslator(appDescriptor, sqlConfig);
       SamzaSqlExecutionContext executionContext = new SamzaSqlExecutionContext(sqlConfig);
-      Map<String, SamzaRelConverter> converters = sqlConfig.getSamzaRelConverters();
+      int queryId = 0;
```
srinipunuru: Can you check with Shenoda to make sure we use the same queryId for diagnostics as well?

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlIOConfig.java
```diff
@@ -79,6 +82,14 @@ public SqlIOConfig(String systemName, String streamName, List<String> sourcePart
     Validate.notEmpty(samzaRelConverterName,
         String.format("%s is not set or empty for system %s", CFG_SAMZA_REL_CONVERTER, systemName));
 
+    if (isRemoteTable()) {
+      samzaRelTableKeyConverterName = streamConfigs.get(CFG_SAMZA_REL_TABLE_KEY_CONVERTER);
+      Validate.notEmpty(samzaRelTableKeyConverterName,
```
srinipunuru: Does all remote tables need a tableKeyConverters even if the table providers support primitive types which doesn't need a conversion?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -59,46 +60,53 @@
...
  */
 class JoinTranslator {
 
   private static final Logger log = LoggerFactory.getLogger(JoinTranslator.class);
   private int joinId;
-  private SqlIOResolver ioResolver;
   private final String intermediateStreamPrefix;
+  private final int queryId;
+  enum TableType {
+    NONE, // Not a table == a stream
```
srinipunuru: Why do we have a TableType which is not a table but a stream?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -114,41 +122,51 @@ void translate(final LogicalJoin join, final TranslatorContext context) {
...
-                intermediateStreamPrefix + "stream_" + joinId)
-            .map(KV::getValue)
-            .join(table, joinFn);
+    MessageStream<SamzaSqlRelMessage> outputStream;
+
+    if (isRemoteTable) {
+      List<String> tableNameParts = (isTablePosOnRight ? join.getRight() : join.getLeft())
+          .getTable()
+          .getQualifiedName();
+      String remoteTableName = SqlIOConfig.getSourceFromSourceParts(tableNameParts);
```
srinipunuru: Do we need metrics for the RemoteTableJoins?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -114,41 +122,51 @@ void translate(final LogicalJoin join, final TranslatorContext context) {
...
+      StreamTableJoinFunction joinFn = new SamzaSqlRemoteTableJoinFunction(context.getMsgConverter(remoteTableName),
+          context.getTableKeyConverter(remoteTableName), remoteTableName, join.getJoinType(), isTablePosOnRight,
+          streamKeyIds, streamFieldNames, tableKeyIds, tableFieldNames, queryId);
+
+      outputStream = inputStream.join(table, joinFn);
```
srinipunuru: How are the failures, throttling handled?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -307,19 +334,26 @@ private Table loadLocalTable(boolean isTablePosOnRight, List<Integer> tableKeyId
...
-    // We need to repartition the stream denoted as table to ensure that both the stream and table that are joined
-    // have the same partitioning scheme and partition key.
-    relOutputStream
-        .partitionBy(m -> createSamzaSqlCompositeKey(m, tableKeyIds),
-            m -> m,
-            KVSerde.of(keySerde, valueSerde),
-            intermediateStreamPrefix + "table_" + joinId)
-        .sendTo(table);
+    // If local table, load the table
+    if (!isRemoteTable) {
```
srinipunuru: There are quite a few if else's in JoinTranslator which is one of the code smells. Can you look at a better way to factor this code?

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -92,6 +93,15 @@ void translate(final TableScan tableScan, final TranslatorContext context,
     final String streamName = sqlIOConfig.getStreamName();
     final String source = sqlIOConfig.getSource();
 
+    final boolean isRemoteTable = sqlIOConfig.getTableDescriptor().isPresent() &&
+        (sqlIOConfig.getTableDescriptor().get() instanceof RemoteTableDescriptor);
+
+    // We don't need to define input stream descriptor for a remote table as the table descriptor is already defined by
```
srinipunuru: This is little bit confusing on why only for RemoteTable, ScanTranslator doesn't do anything. It should be no-op for both Remote and local table right?

samza-sql/src/test/java/org/apache/samza/sql/testutil/SampleRelTableKeyConverter.java
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+package org.apache.samza.sql.testutil;
+
+import java.util.stream.Collectors;
+import org.apache.samza.sql.SamzaSqlRelRecord;
+import org.apache.samza.sql.interfaces.SamzaRelTableKeyConverter;
+
+
+public class SampleRelTableKeyConverter implements SamzaRelTableKeyConverter {
```
srinipunuru: Can you add documentation? Also make sure all the new classes have documentation?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: That will be part of TableDecriptor definition. https://github.com/apache/samza/blob/520da15e682d55499224dc2be5b7964e7780eebf/samza-core/src/main/java/org/apache/samza/table/remote/descriptors/RemoteTableDescriptor.java#L176

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: Join has two inputs to it. And the type of input could be either a stream, local table or remote table. And that is what I wanted to represent here in the enum. Should I change the enum name to InputType and rename the values to STREAM, LOCAL_TABLE, REMOTE_TABLE ? Splitting them as a boolean for stream or table and then another boolean for local or remote would be another approach.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: I refactored some part of the code.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: Yes - we will definitely need to have metrics similar to what Shenoda is adding. I will add them as part of a separate effort, once Shenoda commits his changes for filter and projection.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: I think I found a better way to refactor this code. Let me refactor JoinTranslator but please take a look at responses for other comments.

atoomula: Refactored the code. I will refactor the recursion part of the code to get input types as part of subquery support.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: I refactored this part of the code. Please take a look.

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SamzaRelTableKeyConverter.java
srinipunuru: Thanks for the explanation, This works! Do we have validations for not supporting UDFs on the table fields?

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlIOConfig.java
srinipunuru: Based on the above discussion, The answer to this is Yes? We need a tableKeyConverter even for tables that just support primitives?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
srinipunuru: Can you open a Jira to track this as well as the throttling and retry work?

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlIOConfig.java
atoomula: Correct.

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
atoomula: Sure

PR #795, opened 2018-11-05
SAMZA-1977: Shutdown the container threadPooln stopping AsyncRunLoop.

shanthoosh:


shanthoosh:
@xinyuiscool @sborya 

prateekm:
@shanthoosh what prompted this change?

shanthoosh:
@prateekm 

When fixing SAMZA-1638, I noticed that the container threadPool in AsyncRunLoop was not shutdown at all.

Container thread pool is used to run window, commit, timer tasks in ` AsyncRunLoop`.  This will cause problems in standalone in the following scenario:

1. Due to a rebalance phase, `StreamProcessor` shuts down the `SamzaContainer`. 
2. `StreamProcessor` receives a new `JobModel` (v2) and starts the `SamzaContainer` again.
3. Window, commit tasks from the unstopped container thread pool corresponding to  previous JobModel(v1) would still be running in the JVM.

I think this can cause duplicate processing  depending upon number of window tasks scheduled in the container threadPool before `AsyncRunLoop` shutdown. 

Please correct me if I'm missing something here.

xinyuiscool:
Does SamzaContainer.shutdownTask() shut them down?

prateekm:
@shanthoosh I understand, was just asking if you noticed any specific issues that lead to this change. Why is this closed? Is this not an issue anymore?

PR #797, opened 2018-11-06
SAMZA-1817: Long Classpath Support: Generating relative path from absolute paths

Sanil15:
- We removed wildcarding to shorten the classpath since it induces indeterminism in classpath on two consecutive deploys
- This patch constructs classpath (run-class.sh) by generating the relative path from directory executing the script to target directory where jars are present. 
- From observations on deploying jobs on YARN and locally, we saw that we can save over 65% of classpath length by switching to relative paths as compared to the absolute path

### DIFF
--- a/samza-shell/src/main/bash/run-class.sh
+++ b/samza-shell/src/main/bash/run-class.sh
@@ -51,12 +51,13 @@ export JOB_LIB_DIR=$JOB_LIB_DIR
 
 function relative_path_from_to() {
  # returns relative path to $2/$target from $1/$source
- source=$1
- target=$2
+ source=$1 # This is $home_dir for this case
+ target=$2 # This is $BASE_LIB_DIR where JARS are
 
  common_part=$source # for now
  result="" # for now
 
+ # This calculates the longest common prefix
  while [[ "${target#$common_part}" == "${target}" ]]; do
      # no match, means that candidate common part is not correct
      # go up one level (reduce common part)
@@ -89,9 +90,13 @@ function relative_path_from_to() {
  echo "./$result"
 }
 
-
 # Get the relative path from current execution directory to base directory with jars
 RELATIVE_PATH_BASE_LIB=$(relative_path_from_to $home_dir $BASE_LIB_DIR)
+# If the relative path directory does not exist, resolve to absolute path
+if [ ! -d "$RELATIVE_PATH_BASE_LIB" ]; then
+   echo "Resolving to absolute path, $RELATIVE_PATH_BASE_LIB does not exist"
+   RELATIVE_PATH_BASE_LIB=$BASE_LIB_DIR
+fi
 echo RELATIVE_PATH_BASE_LIB=$RELATIVE_PATH_BASE_LIB
 
 if [ -d "$JOB_LIB_DIR" ] && [ "$JOB_LIB_DIR" != "$BASE_LIB_DIR" ]; then
--- a/samza-shell/src/main/bash/run-class.sh
+++ b/samza-shell/src/main/bash/run-class.sh
@@ -45,7 +45,7 @@ DEFAULT_LOG4J2_FILE=$base_dir/lib/log4j2.xml
 BASE_LIB_DIR="$base_dir/lib"
 # JOB_LIB_DIR will be set for yarn container in ContainerUtil.java
 # for others we set it to home_dir/lib
-JOB_LIB_DIR="${JOB_LIB_DIR:-$base_dir/lib}"
+JOB_LIB_DIR="${JOB_LIB_DIR:-$home_dir/lib}"
 
 export JOB_LIB_DIR=$JOB_LIB_DIR
 
--- a/samza-shell/src/main/bash/run-class.sh
+++ b/samza-shell/src/main/bash/run-class.sh
@@ -49,15 +49,58 @@ JOB_LIB_DIR="${JOB_LIB_DIR:-$base_dir/lib}"
 
 export JOB_LIB_DIR=$JOB_LIB_DIR
 
-echo JOB_LIB_DIR=$JOB_LIB_DIR
-echo BASE_LIB_DIR=$BASE_LIB_DIR
+function relative_path_from_to() {
+ # returns relative path to $2/$target from $1/$source
+ source=$1
+ target=$2
+
+ common_part=$source # for now
+ result="" # for now
+
+ while [[ "${target#$common_part}" == "${target}" ]]; do
+     # no match, means that candidate common part is not correct
+     # go up one level (reduce common part)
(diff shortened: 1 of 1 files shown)


Sanil15:
@sborya @abhishekshivanna Please have a look

samza-shell/src/main/bash/run-class.sh
```diff
@@ -45,19 +45,68 @@ DEFAULT_LOG4J2_FILE=$base_dir/lib/log4j2.xml
...
+
+ echo "./$result"
+}
+
+# Get the relative path from current execution directory to base directory with jars
+RELATIVE_PATH_BASE_LIB=$(relative_path_from_to $home_dir $BASE_LIB_DIR)
+# If the relative path directory does not exist, resolve to absolute path
+if [ ! -d "$RELATIVE_PATH_BASE_LIB" ]; then
+   echo "Resolving to absolute path, $RELATIVE_PATH_BASE_LIB does not exist"
+   RELATIVE_PATH_BASE_LIB=$BASE_LIB_DIR
```
abhishekshivanna: can you explain under what scenario this condition is hit ? especially since `RELATIVE_PATH_BASE_LIB` is derived from `BASE_LIB_DIR`

Sanil15: RELATIVE_PATH_BASE_LIB also depends on $home_dir, which might be different when someone externally executes control script vs execution path deployment,  this just acts as a safeguard in those cases

samza-shell/src/main/bash/run-class.sh
```diff
@@ -45,19 +45,68 @@ DEFAULT_LOG4J2_FILE=$base_dir/lib/log4j2.xml
...
+# Purpose: Given two absolute paths return relative path from $source to $destination
+function relative_path_from_to() {
+ source=${1%/} # This is $home_dir for this case
+ target=${2%/} # This is $BASE_LIB_DIR where JARS are
+
+ common_part=$source # start with common_part as source and iterate until common_part becomes $target
+ result=""
+
+ # This calculates the longest common prefix
+ while [[ "${target#$common_part}" == "${target}" ]]; do
```
shanthoosh: Though this idea sounds interesting, can we just cd into the  directory(BASE_LIB_DIR) that has all the jars and run the java launch command with absolute jar path. 

Something along the lines of the following:
```bash 
sh -c 'cd BASE_LIB_DIR &&  exec $JAVA $JAVA_OPTS -cp $CLASSPATH "$@"
```
I have assumed that BASE_LIB_DIR has all the relevant jars. I think this will preserve the current working dir in the script and simplify the code in `run-class.sh`. 

Will that work in this case?

samza-shell/src/main/bash/run-class.sh
```diff
@@ -45,19 +45,68 @@ DEFAULT_LOG4J2_FILE=$base_dir/lib/log4j2.xml
...
+     result="${forward_part:1}"
+ fi
+
+ echo "./$result"
+}
+
+# Get the relative path from current execution directory to base directory with jars
+RELATIVE_PATH_BASE_LIB=$(relative_path_from_to $home_dir $BASE_LIB_DIR)
+# If the relative path directory does not exist, resolve to absolute path
+if [ ! -d "$RELATIVE_PATH_BASE_LIB" ]; then
```
sborya: it would be safer to check for existence of a jar file (samza-api\*.jar for example).

samza-shell/src/main/bash/run-class.sh
Sanil15: This will trouble us when we want to read jar/wars out of JOB_LIB_DIR and BASE_LIB_DIR

samza-shell/src/main/bash/run-class.sh
Sanil15: will do

samza-shell/src/main/bash/run-class.sh
shanthoosh: Even in your current solution, you're using relative_path for one lib directory and absolute path for the other lib directory. You can follow the same approach with the above idea as well. I think it'll be good for us in the long run to go with a simpler solution rather than computing longest common prefix and stripping stuff. 

If you disagree, then please share your rationale for doing so?

Sanil15: - Are you sure `sh -c 'cd BASE_LIB_DIR &&  exec $JAVA $JAVA_OPTS -cp $CLASSPATH'`
works? Please confirm on this
- Also please throw more light on the issues with the current solution? Why do you think it will be a problem?
- One more thing to mention: Look at the shebang, might run into syntax errors on using sh 

Please be considerate that we have spent time and effort thoroughly testing/reviewing the current fix in various deployment modes.

shanthoosh: >> Are you sure sh -c 'cd BASE_LIB_DIR && exec $JAVA $JAVA_OPTS -cp $CLASSPATH'
works? Please confirm on this

It's the contract of [GNU shell](https://linux.die.net/man/1/sh).


>> Also please throw more light on the issues with the current solution? Why do you think it will be a problem?
1. Existing solution is hard to understand of me now. In the middle of the night, admist a GCN, I think it wouldn't be a great to decipher the logic in it. 
2. Any other developer who wants to alter it in the future would have a tough time with it.


>> One more thing to mention: Look at the shebang, might run into syntax errors on using sh

There are multiple ways to solve this problem. Take a look [here](https://stackoverflow.com/questions/10382141/temporarily-change-current-working-directory-in-bash-to-run-a-command). 
I am not suggesting that you should switch to the approach I mentioned. At-least we should try to simplify things here by trying alternate options. I just suggested an alternative. 


>> Please be considerate that we have spent time and effort thoroughly testing/reviewing the current fix in various deployment modes.

IMHO, this is not a strong enough rationale.  I am not going to stand in the way if everyone else thinks this is fine.

Sanil15: Tested with `cd BASE_LIB_DIR && exec $JAVA $JAVA_OPTS -cp $CLASSPATH` and found this breaks control-dev & control-yarn, since they build bash arguments expecting this execution of JAVA from a specific path. We use relative path in those scripts for certs, other config files.

Sanil15:
@shanthoosh Tested with cd BASE_LIB_DIR && exec $JAVA $JAVA_OPTS -cp $CLASSPATH and found this breaks control-dev & control-yarn, since they build bash arguments expecting this execution of JAVA from a specific path. We use relative path in those scripts for certs, other config files.

Changing the execution directory for Java entails a larger refactoring

Sanil15:
https://github.com/apache/samza/pull/824 Simplifies this, discarding this patch for now

PR #800, opened 2018-11-09
SAMZA-1986: Samza-sql: Use system name along with stream name for streamId

atoomula:


samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedIOResolverFactory.java
```diff
@@ -68,10 +68,10 @@ public ConfigBasedIOResolver(Config config) {
 
     @Override
     public SqlIOConfig fetchSourceInfo(String source) {
-      return fetchSystemInfo(source);
+      return fetchSystemInfo(source, false);
     }
 
-    private SqlIOConfig fetchSystemInfo(String name) {
+    private SqlIOConfig fetchSystemInfo(String name, boolean isSink) {
```
srinipunuru: Are we using the sink or source info?  If not, Can we remove this and add it when we are going to use it?

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlIOConfig.java
```diff
@@ -96,10 +97,15 @@ public SqlIOConfig(String systemName, String streamName, List<String> sourcePart
...
-    // Currently, only local table is supported. And it is assumed that all tables are local tables.
-    if (tableDescriptor != null) {
-      streamConfigs.put(String.format(StreamConfig.BOOTSTRAP_FOR_STREAM_ID(), streamName), "true");
-      streamConfigs.put(String.format(StreamConfig.CONSUMER_OFFSET_DEFAULT_FOR_STREAM_ID(), streamName), "oldest");
+    if (!isRemoteTable()) {
+      // The below config is required for local table and streams but not for remote table.
+      streamConfigs.put(String.format(StreamConfig.PHYSICAL_NAME_FOR_STREAM_ID(), streamId), streamName);
+    }
+
+    if (tableDescriptor != null && !isRemoteTable()) {
```
srinipunuru: Can you merge this condition with the previous one?

samza-test/src/test/java/org/apache/samza/test/samzasql/TestSamzaSqlEndToEnd.java
```diff
@@ -110,6 +130,28 @@ public void testEndToEndMultiSqlStmts() {
     Assert.assertTrue(IntStream.range(0, numMessages).boxed().collect(Collectors.toList()).equals(new ArrayList<>(outMessagesSet)));
   }
 
+  @Ignore
+  @Test
+  public void testEndToEndMultiSqlStmtsWithSameSystemStreamAsInputAndOutput() {
```
srinipunuru: Thanks for adding this test. Can you add a comment on why we are ignoring this and possibly point to the Jira ticket tracking the issue?

atoomula: Ok

samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedIOResolverFactory.java
atoomula: Sure.

samza-sql/src/main/java/org/apache/samza/sql/interfaces/SqlIOConfig.java
atoomula: Yup. Thanks for the suggestion.

PR #801, opened 2018-11-09
SAMZA-1988: Properly suffix modules with direct Scala dependencies with the Scala version.

dnishimura:
List of modules without a Scala version suffix that have direct Scala dependencies and the direct Scala API calls in each module are in the JIRA ticket: https://issues.apache.org/jira/browse/SAMZA-1988

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -203,7 +203,7 @@ project(":samza-core_$scalaVersion") {
   }
 }
 
-project(':samza-azure') {
+project(":samza-azure_$scalaVersion") {
   apply plugin: 'java'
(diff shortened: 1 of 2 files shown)


prateekm:
@dnishimura Minor: can you move the description to the JIRA, it shows up as commit message and clutters 'git log'.

dnishimura:
@prateekm It was duplicated from the ticket. I'll link the ticket instead.

dnishimura:
@prateekm - I ran `bin/check-all.sh` and everything passed. The first failure was likely due to my laptop going to sleep mode during the tests. Please review and approve.

PR #803, opened 2018-11-12
SAMZA-1989: SystemStreamGrouper interface change for SEP-5

shanthoosh:
Samza users may need to increase the partition count of the input streams of their stateful samza jobs. For example, Kafka needs to limit the maximum size of each partition to scale up its performance. Thus the number of partitions of a Kafka topic needs to be expanded to reduce the partition size if the average byte-in-rate or retention time of the Kafka topic has doubled.

In order to perform a join between streams, stateful jobs generally have to route the partitions from the different input streams to same task of a container. However, when a input stream repartitioning happens, key space of a partition gets redistributed. This will make the stateful jobs to produce erroneous results.

So if the partition count of input stream is increased then the users have to manually purge the changelog topics, local RocksDb state of their stateful jobs. This  results in an increased operational complexity and data loss.

This patch takes a first stab at solving the above problem and is comprised of the following changes:

* Introduce a new group method in `SystemStreamPartitionGrouper` interface to generate task assignment factoring in the partition expansion of input streams.
* Introduced a `StreamPartitionMapper` abstraction to allow the user to plugin the input stream partitioning function.  
* Fixed the existing unit tests and added new unit tests to validate the new grouper changes.

In a followup PR shortly, these grouper changes would be integrated with `JobModelManager`(Waiting for PR 790 to be landed for this. It had made significant changes to `JobModelManager`)

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/container/grouper/stream/GrouperContext.java
+++ b/samza-api/src/main/java/org/apache/samza/container/grouper/stream/GrouperContext.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.container.grouper.stream;
+
+import java.util.Collections;
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.List;
+import java.util.Set;
+import org.apache.samza.annotation.InterfaceStability;
+import org.apache.samza.container.TaskName;
+import org.apache.samza.runtime.LocationId;
+import org.apache.samza.system.SystemStreamPartition;
+
+/**
+ * A Wrapper class that holds the necessary historical metadata of the samza job which is used
+ * by the {@link org.apache.samza.container.grouper.stream.SystemStreamPartitionGrouper}
+ * to generate optimal task assignments.
+ */
+@InterfaceStability.Evolving
+public class GrouperContext {
+  private Map<String, LocationId> processorLocality;
+  private Map<TaskName, LocationId> taskLocality;
+  private Map<TaskName, Set<SystemStreamPartition>> previousTaskToSSPAssignment;
+  private Map<TaskName, String> previousTaskToContainerAssignment;
+
+  public GrouperContext(Map<String, LocationId> processorLocality, Map<TaskName, LocationId> taskLocality, Map<TaskName, Set<SystemStreamPartition>> previousTaskToSSPAssignments, Map<TaskName, String> previousTaskToContainerAssignment) {
+    this.processorLocality = processorLocality;
+    this.taskLocality = taskLocality;
+    this.previousTaskToSSPAssignment = previousTaskToSSPAssignments;
+    this.previousTaskToContainerAssignment = previousTaskToContainerAssignment;
+  }
+
+  public Map<String, LocationId> getProcessorLocality() {
+    return Collections.unmodifiableMap(processorLocality);
+  }
+
+  public Map<TaskName, LocationId> getTaskLocality() {
+    return Collections.unmodifiableMap(taskLocality);
+  }
+
+  public Map<TaskName, Set<SystemStreamPartition>> getPreviousTaskToSSPAssignment() {
+    return Collections.unmodifiableMap(previousTaskToSSPAssignment);
+  }
+
+  public List<String> getProcessorIds() {
+    return new ArrayList<>(processorLocality.keySet());
+  }
+
+  public Map<TaskName, String> getPreviousTaskToContainerAssignment() {
+    return Collections.unmodifiableMap(this.previousTaskToContainerAssignment);
+  }
+}
--- a/samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapper.java
+++ b/samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapper.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.container.grouper.stream;
+
+import org.apache.samza.system.SystemStreamPartition;
+
+/**
+ * Input streams of a samza job can be either expanded or contracted by the user.
+ * This abstraction determines the previous {@link SystemStreamPartition} for a {@link SystemStreamPartition}
+ * of a input stream after the stream expansion. This will be used in {@link SystemStreamPartitionGrouper}
+ * implementations to generate partition expansion aware task assignments.
+ */
+public interface StreamPartitionMapper {
+
+  /**
+   * Determines the previous {@link SystemStreamPartition} for a {@link SystemStreamPartition}
+   * of a input stream after the stream expansion.
+   * @param currentSystemStreamPartition denotes the current partition after the stream expansion.
+   * @param partitionCountBeforeExpansion the partition count of the stream before the stream expansion.
+   * @param partitionCountAfterExpansion the partition count of the stream after the stream expansion.
+   * @return the mapped {@link SystemStreamPartition}.
+   */
+  SystemStreamPartition getSSPBeforeExpansion(SystemStreamPartition currentSystemStreamPartition,
+                                              int partitionCountBeforeExpansion, int partitionCountAfterExpansion);
+}
--- a/samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapper.java
+++ b/samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapper.java
@@ -23,19 +23,19 @@ import org.apache.samza.system.SystemStreamPartition;
 /**
  * Input streams of a samza job can be either expanded or contracted by the user.
  * This abstraction determines the previous {@link SystemStreamPartition} for a {@link SystemStreamPartition}
- * of a input stream after the stream expansion. This will be used in {@link SystemStreamPartitionGrouper}
+ * of a input stream after the stream expansion or contraction. This will be used in {@link SystemStreamPartitionGrouper}
  * implementations to generate partition expansion aware task assignments.
  */
 public interface StreamPartitionMapper {
 
   /**
    * Determines the previous {@link SystemStreamPartition} for a {@link SystemStreamPartition}
-   * of a input stream after the stream expansion.
-   * @param currentSystemStreamPartition denotes the current partition after the stream expansion.
-   * @param partitionCountBeforeExpansion the partition count of the stream before the stream expansion.
-   * @param partitionCountAfterExpansion the partition count of the stream after the stream expansion.
+   * of a input stream after the stream expansion or contraction.
+   * @param currentSystemStreamPartition denotes the current partition after the stream partition count change.
+   * @param previousPartitionCount the partition count of the stream before the stream partition count change.
+   * @param afterPartitionCount the partition count of the stream after the stream partition count change.
    * @return the mapped {@link SystemStreamPartition}.
    */
   SystemStreamPartition getSSPAfterPartitionChange(SystemStreamPartition currentSystemStreamPartition,
-                                                   int partitionCountBeforeExpansion, int partitionCountAfterExpansion);
+                                                   int previousPartitionCount, int afterPartitionCount);
 }
--- a/samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapper.java
+++ b/samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapper.java
@@ -36,6 +36,6 @@ public interface StreamPartitionMapper {
    * @param partitionCountAfterExpansion the partition count of the stream after the stream expansion.
    * @return the mapped {@link SystemStreamPartition}.
    */
-  SystemStreamPartition getSSPBeforeExpansion(SystemStreamPartition currentSystemStreamPartition,
-                                              int partitionCountBeforeExpansion, int partitionCountAfterExpansion);
+  SystemStreamPartition getSSPAfterPartitionChange(SystemStreamPartition currentSystemStreamPartition,
+                                                   int partitionCountBeforeExpansion, int partitionCountAfterExpansion);
 }
--- a/samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapperFactory.java
+++ b/samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapperFactory.java
@@ -1,0 +1,26 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.container.grouper.stream;
+
+import org.apache.samza.config.Config;
+import org.apache.samza.metrics.MetricsRegistry;
+
+public interface StreamPartitionMapperFactory {
+  StreamPartitionMapper getPartitionExpansionAlgorithm(Config config, MetricsRegistry metricsRegistry);
+}
--- a/samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapperFactory.java
+++ b/samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapperFactory.java
@@ -22,5 +22,5 @@ import org.apache.samza.config.Config;
 import org.apache.samza.metrics.MetricsRegistry;
 
 public interface StreamPartitionMapperFactory {
-  StreamPartitionMapper getPartitionExpansionAlgorithm(Config config, MetricsRegistry metricsRegistry);
+  StreamPartitionMapper getStreamPartitionMapper(Config config, MetricsRegistry metricsRegistry);
 }
--- a/samza-api/src/main/java/org/apache/samza/container/grouper/stream/SystemStreamPartitionGrouper.java
+++ b/samza-api/src/main/java/org/apache/samza/container/grouper/stream/SystemStreamPartitionGrouper.java
@@ -18,11 +18,10 @@
  */
 package org.apache.samza.container.grouper.stream;
 
+import java.util.*;
+
 import org.apache.samza.container.TaskName;
 import org.apache.samza.system.SystemStreamPartition;
-
-import java.util.Map;
-import java.util.Set;
 
 /**
  * Group a set of SystemStreamPartitions into logical taskNames that share a common characteristic, defined
@@ -37,5 +36,9 @@ import java.util.Set;
  * the TaskNames.
  */
 public interface SystemStreamPartitionGrouper {
-  public Map<TaskName, Set<SystemStreamPartition>> group(Set<SystemStreamPartition> ssps);
+  Map<TaskName, Set<SystemStreamPartition>> group(Set<SystemStreamPartition> systemStreamPartitions);
+
+  default Map<TaskName, Set<SystemStreamPartition>> group(Set<SystemStreamPartition> systemStreamPartitions, GrouperContext grouperContext) {
+    return group(systemStreamPartitions);
+  }
 }
--- a/samza-api/src/main/java/org/apache/samza/container/grouper/stream/SystemStreamPartitionGrouper.java
+++ b/samza-api/src/main/java/org/apache/samza/container/grouper/stream/SystemStreamPartitionGrouper.java
@@ -36,9 +36,11 @@ import org.apache.samza.system.SystemStreamPartition;
  * the TaskNames.
  */
 public interface SystemStreamPartitionGrouper {
-  Map<TaskName, Set<SystemStreamPartition>> group(Set<SystemStreamPartition> systemStreamPartitions);
 
-  default Map<TaskName, Set<SystemStreamPartition>> group(Set<SystemStreamPartition> systemStreamPartitions, GrouperContext grouperContext) {
-    return group(systemStreamPartitions);
-  }
+  /**
+   * Groups the input systemStreamPartitions into the logical taskNames.
+   * @param systemStreamPartitions the input system stream partitions.
+   * @return the grouped {@link TaskName} to {@link SystemStreamPartition} assignments.
+   */
+  Map<TaskName, Set<SystemStreamPartition>> group(Set<SystemStreamPartition> systemStreamPartitions);
 }
--- a/samza-api/src/main/java/org/apache/samza/container/grouper/stream/SystemStreamPartitionMapper.java
+++ b/samza-api/src/main/java/org/apache/samza/container/grouper/stream/SystemStreamPartitionMapper.java
@@ -23,8 +23,7 @@ import org.apache.samza.system.SystemStreamPartition;
 /**
  * Input streams of a samza job can be either expanded or contracted by the user.
  * This abstraction determines the previous {@link SystemStreamPartition} for a {@link SystemStreamPartition}
- * of a input stream after the stream expansion or contraction. This will be used in {@link SystemStreamPartitionGrouper}
- * implementations to generate partition expansion aware task assignments.
+ * of a input stream after the stream expansion or contraction.
  */
 public interface SystemStreamPartitionMapper {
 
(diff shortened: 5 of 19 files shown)


samza-api/src/main/java/org/apache/samza/container/grouper/stream/GrouperContext.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.samza.container.TaskName;
+import org.apache.samza.runtime.LocationId;
+import org.apache.samza.system.SystemStreamPartition;
+
+/**
+ * A Wrapper class that holds the necessary historical metadata of the samza job which is used
+ * by the {@link org.apache.samza.container.grouper.stream.SystemStreamPartitionGrouper}
+ * to generate optimal task assignments.
+ */
+@InterfaceStability.Evolving
```
shanthoosh: Added it here since PR 790 is still under review. Will be removed after PR: 790 is landed.

shanthoosh:
@xinyuiscool @sborya @prateem 

shanthoosh:
@dnishimura Can you please review this patch when you have some time.

samza-api/src/main/java/org/apache/samza/container/grouper/stream/GrouperContext.java
```diff
@@ -0,0 +1,69 @@
...
+public class GrouperContext {
+  private Map<String, LocationId> processorLocality;
+  private Map<TaskName, LocationId> taskLocality;
+  private Map<TaskName, Set<SystemStreamPartition>> previousTaskToSSPAssignment;
+  private Map<TaskName, String> previousTaskToContainerAssignment;
```
rmatharu:
```
Please ad these comments: 
private Map<TaskName, LocationId> taskLocality; // Map of task-name to location ID
private Map<TaskName, Set<SystemStreamPartition>> previousTaskToSSPAssignment; // Map of task-name to the set of input SSPs for each task
private Map<TaskName, String> previousTaskToContainerAssignment; // Map of task-name to the container-name on which the task is hosted.
private Map<String, LocationId> processorLocality; // <Meaning goes here> 
```

rmatharu: Also what does ```processorLocality``` mean for YARN deploys?

samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapper.java
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Determines the previous {@link SystemStreamPartition} for a {@link SystemStreamPartition}
+   * of a input stream after the stream expansion.
+   * @param currentSystemStreamPartition denotes the current partition after the stream expansion.
+   * @param partitionCountBeforeExpansion the partition count of the stream before the stream expansion.
+   * @param partitionCountAfterExpansion the partition count of the stream after the stream expansion.
+   * @return the mapped {@link SystemStreamPartition}.
+   */
+  SystemStreamPartition getSSPBeforeExpansion(SystemStreamPartition currentSystemStreamPartition,
```
rmatharu: Please name this generically since this abstraction is used both in case of expansion and reduction, e.g., 
``` getSSPBeforePartitionChange(SSP currentSSP, int partitionCountBeforeChange, int partitionCountAfterChange)```

File: samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapperFactory.java [hunk 4]
```diff
@@ -0,0 +1,26 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.container.grouper.stream;
+
+import org.apache.samza.config.Config;
+import org.apache.samza.metrics.MetricsRegistry;
+
+public interface StreamPartitionMapperFactory {
+  StreamPartitionMapper getPartitionExpansionAlgorithm(Config config, MetricsRegistry metricsRegistry);
```
rmatharu: Going by other Factories in Samza code base, perhaps
```StreamPartitionMapper  getStreamPartitionMapper(config, metricsRegistry)```

File: samza-api/src/main/java/org/apache/samza/container/grouper/stream/SystemStreamPartitionGrouper.java [hunk 5]
```diff
@@ -37,5 +36,9 @@
  * the TaskNames.
  */
 public interface SystemStreamPartitionGrouper {
-  public Map<TaskName, Set<SystemStreamPartition>> group(Set<SystemStreamPartition> ssps);
+  Map<TaskName, Set<SystemStreamPartition>> group(Set<SystemStreamPartition> systemStreamPartitions);
+
+  default Map<TaskName, Set<SystemStreamPartition>> group(Set<SystemStreamPartition> systemStreamPartitions, GrouperContext grouperContext) {
```
rmatharu: Is this interface change really necessary. The ```GrouperContext``` can be passed to the interface-implementing class in its constructor as an Option<GrouperContext>. 
If a GrouperContext is specified, the class can take it into consideration on invocations of ```group(ssps)```.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/DefaultStreamPartitionMapper.java
```diff
@@ -0,0 +1,41 @@
...
+import org.apache.samza.Partition;
+import org.apache.samza.system.SystemStreamPartition;
+
+public class DefaultStreamPartitionMapper implements StreamPartitionMapper {
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public SystemStreamPartition getSSPBeforeExpansion(SystemStreamPartition currentSystemStreamPartition, int partitionCountBeforeExpansion, int partitionCountAfterExpansion) {
```
rmatharu: Please consider using this implementation as the default interface impl in StreamPartitionMapper

samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapper.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.apache.samza.system.SystemStreamPartition;
+
+/**
+ * Input streams of a samza job can be either expanded or contracted by the user.
+ * This abstraction determines the previous {@link SystemStreamPartition} for a {@link SystemStreamPartition}
+ * of a input stream after the stream expansion. This will be used in {@link SystemStreamPartitionGrouper}
+ * implementations to generate partition expansion aware task assignments.
+ */
+public interface StreamPartitionMapper {
```
prateekm: Why a new interface instead of a new config/mode for the existing grouper? What other implementations of this interface do you envision?

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AbstractSystemStreamPartitionGrouper.java
```diff
@@ -0,0 +1,186 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.util.*;
+
+/**
+ * An abstract implementation of {@link SystemStreamPartitionGrouper} that provides a stream expansion
+ * aware task to partition assignments.
+ */
+@InterfaceStability.Evolving
+public abstract class AbstractSystemStreamPartitionGrouper implements SystemStreamPartitionGrouper {
```
prateekm: Are all SSP grouper implementations expected to extend this? If so, document it. If not, why make this an abstract class instead of using the functionality here as util methods in implementations that need it?

rmatharu: An easier way to implement this very same logic maybe to just enrich the existing ```GroupBySystemStreamPartition```, and have it take the GrouperContext as an optional constructor input.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AbstractSystemStreamPartitionGrouper.java
```diff
@@ -0,0 +1,186 @@
...
+      for (SystemStreamPartition currentSystemStreamPartition : entry.getValue()) {
+        if (broadcastSystemStreamPartitions.contains(currentSystemStreamPartition)) {
+          LOGGER.info("SystemStreamPartition: {} is part of broadcast stream. Skipping reassignment.", currentSystemStreamPartition);
+        } else {
+          SystemStream systemStream = currentSystemStreamPartition.getSystemStream();
+
+          Integer previousStreamPartitionCount = previousStreamToPartitionCount.getOrDefault(systemStream, 0);
+          Integer currentStreamPartitionCount = currentStreamToPartitionCount.getOrDefault(systemStream, 0);
+
+          if (previousStreamPartitionCount > 0 && previousStreamPartitionCount < currentStreamPartitionCount
```
rmatharu: Could we also make this work both incase of a partition-expansion and reduction.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AbstractSystemStreamPartitionGrouper.java
```diff
@@ -0,0 +1,186 @@
...
+            LOGGER.info("SystemStream: {} has expanded from: {} to: {}. Performing partition reassignment.", systemStream, previousStreamPartitionCount, currentStreamPartitionCount);
+
+            SystemStreamPartition previousSystemStreamPartition = streamPartitionMapper.getSSPBeforeExpansion(currentSystemStreamPartition, previousStreamPartitionCount, currentStreamPartitionCount);
+            TaskName previouslyAssignedTask = previousSSPToTask.get(previousSystemStreamPartition);
+
+            LOGGER.info("Moving systemStreamPartition: {} from task: {} to task: {}.", currentSystemStreamPartition, currentlyAssignedTask, previouslyAssignedTask);
+
+            taskToPartitionGroup.get(currentlyAssignedTask).removeSSP(currentSystemStreamPartition);
+            taskToPartitionGroup.get(previouslyAssignedTask).addSSP(currentSystemStreamPartition);
+          }
```
rmatharu:
``` else {
if (currentStreamPartitionCount % previousStreamPartitionCount !=0)
logger.error("your partition change is not proportionate. current partition count {} previous {} ... ");

}
```

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -101,6 +101,8 @@ object JobConfig {
   val DIAGNOSTICS_APPENDER_CLASS = "job.diagnostics.appender.class"
   val DEFAULT_DIAGNOSTICS_APPENDER_CLASS = "org.apache.samza.logging.log4j.SimpleDiagnosticsAppender"
 
+  val STREAM_PARTITION_MAPPER_FACTORY = "job.stream.partition.mapper.factory"
```
rmatharu: Given the default functionality of the StreamPartitionMapper above, this config doesnt seem warranted.

samza-core/src/test/java/org/apache/samza/container/grouper/stream/TestGroupBySystemStreamPartition.java
```diff
@@ -21,88 +21,265 @@
...
-  SystemStreamPartition ac0 = new SystemStreamPartition("SystemA", "StreamB", new Partition(0));
-  GroupBySystemStreamPartitionFactory grouperFactory = new GroupBySystemStreamPartitionFactory();
+  private SystemStreamPartition aa0 = new SystemStreamPartition("SystemA", "StreamA", new Partition(0));
+  private SystemStreamPartition aa1 = new SystemStreamPartition("SystemA", "StreamA", new Partition(1));
+  private SystemStreamPartition aa2 = new SystemStreamPartition("SystemA", "StreamA", new Partition(2));
+  private SystemStreamPartition ac0 = new SystemStreamPartition("SystemA", "StreamB", new Partition(0));
+  private GroupBySystemStreamPartitionFactory grouperFactory = new GroupBySystemStreamPartitionFactory();
 
   @Test
   public void testLocalStreamGroupedCorrectly() {
```
rmatharu: Maybe we need to add a test where previous and current task assignments in the grouper-context, are exactly the same in which case the grouping should remain the same.

samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapper.java
shanthoosh: When partition expansion or contraction happens on a input stream, we want to know the producer partitioning function used in the upstream to map the partitions back to correct task.  


1. This interface is for abstracting out that producer partitioning function part in samza for different input sources. `DefaultPartitioner`, `MurmurPartitioner` and `RangePartitioner` are some of the kafka producer partitioning functions  used within  LI through samza, tracker and other libraries. 

2. In SEP-5, this partition mapping was made pluggable as well.

samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapper.java
shanthoosh: Makes sense. Done.

samza-api/src/main/java/org/apache/samza/container/grouper/stream/StreamPartitionMapperFactory.java
shanthoosh: Done.

samza-api/src/main/java/org/apache/samza/container/grouper/stream/SystemStreamPartitionGrouper.java
shanthoosh: Yes, good point. We want to make all the previous metadata of the job execution available to the interface so that it can generate grouping that is partition aware(or in future do other optimizations). 

1. Currently in samza, `SystemStreamPartitionGrouper` is only instantiated through the `SystemStreamPartitionGrouperFactory` using reflection. So, to invoke the appropriate `SystemStreamPartitionGrouper` constructor which takes in the `GrouperContext` optional, the factory has to be aware of the GrouperContext optional. To accomplish this, we need to make a interface change(add a new method) in `SystemStreamPartitionGrouperFactory`. However, the problem then becomes the factory invoking the right constructor and any custom implementations in the future, would come to know that this `GrouperContext` is available by following a trial through the factory which may not be obvious. By adding this to the interface we make the interface contract clear and upfront.
2. I think the grouper context will be present for a job all the time. During the first run of the samza job, it will be empty. I'm not sure of the necessity to make it optional though.

This interface change was approved in SEP-5. If you have strong opinions about this, then lets discuss offline.

samza-api/src/main/java/org/apache/samza/container/grouper/stream/GrouperContext.java
shanthoosh: 1. Those comments are already added in PR 790. This file will not be checked-in as a part of this PR(I added a comment to indicate that). I added this file here, since PR 790 is in review. Apologies for any confusion.
2. Processor locality for a processor denotes the previous host in which a particular processor ran in YARN.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AbstractSystemStreamPartitionGrouper.java
```diff
@@ -0,0 +1,186 @@
...
+    return sspToTaskMapping;
+  }
+
+  /**
+   * Creates a instance of {@link StreamPartitionMapper} using the partition expansion factory class
+   * defined in the {@param config}.
+   * @param config the configuration of the samza job.
+   * @return the instantiated {@link StreamPartitionMapper} object.
+   */
+  private StreamPartitionMapper getPartitionExpansionAlgorithm(Config config) {
```
prateekm: Factory method name should match class name its returning.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/DefaultStreamPartitionMapper.java
```diff
@@ -0,0 +1,41 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.container.grouper.stream;
+
+import com.google.common.base.Preconditions;
+import org.apache.samza.Partition;
+import org.apache.samza.system.SystemStreamPartition;
+
+public class DefaultStreamPartitionMapper implements StreamPartitionMapper {
```
prateekm: What does default mean here?

samza-core/src/main/java/org/apache/samza/container/grouper/stream/GroupByPartition.java
```diff
@@ -23,57 +23,48 @@
...
 
-public class GroupByPartition implements SystemStreamPartitionGrouper {
-  private TaskConfigJava taskConfig = null;
-  private Set<SystemStreamPartition> broadcastStreams = new HashSet<SystemStreamPartition>();
+public class GroupByPartition extends AbstractSystemStreamPartitionGrouper {
+  private final Set<SystemStreamPartition> broadcastSystemStreamPartitions;
 
   /**
-   * default constructor
+   * @param config the configuration of the job.
```
prateekm: Add description for javadocs.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/AbstractSystemStreamPartitionGrouper.java
prateekm: In that case s/Abstract/Base to clarify that this is the expected base class for all groupers. Also, there are no abstract methods in this class, what are the extending classes expected to implement?

prateekm: The `SystemStreamPartitionGrouper` interface doesn't mention this. There's also nothing preventing them from overriding this method. If this functionality is 
1. going to be common to and required for all grouper implementations, and
2. provided by the framework
it will be better to use composition over inheritence, and use the user provided grouper and mapper implementation within this class. 
This will also address @rmatharu's comments about adding a new method to the grouper and moving some of this logic to the default method implementation - it will be unnecessary. Does that sound reasonable?

shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/SSPGrouperProxy.java
```diff
@@ -0,0 +1,195 @@
...
+      for (SystemStreamPartition currentSystemStreamPartition : entry.getValue()) {
+        if (broadcastSystemStreamPartitions.contains(currentSystemStreamPartition)) {
+          LOGGER.info("SystemStreamPartition: {} is part of broadcast stream. Skipping reassignment.", currentSystemStreamPartition);
+        } else {
+          SystemStream systemStream = currentSystemStreamPartition.getSystemStream();
+
+          Integer previousStreamPartitionCount = previousStreamToPartitionCount.getOrDefault(systemStream, 0);
+          Integer currentStreamPartitionCount = currentStreamToPartitionCount.getOrDefault(systemStream, 0);
+
+          if (previousStreamPartitionCount > 0 && currentStreamPartitionCount % previousStreamPartitionCount == 0) {
```
dnishimura: Maybe this check for "previous count needing to be a multiplier of the current partition count" should be enforced only by the `StreamPartitionMapper`? Here, if the condition is false, then it loosely fails by logging a warning and will never hit the same Precondition check in `HashStreamPartitionMapper#getSSPAfterPartitionChange`. Might be cleaner to consolidate the logic.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -213,4 +215,11 @@ class JobConfig(config: Config) extends ScalaMapConfig(config) with Logging {
   def getJMXEnabled = {
     getBoolean(JobConfig.JOB_JMX_ENABLED, true);
   }
+
+  /**
+   * @return the partition expansion algorithm factory defined in the config.
```
prateekm: Add javadoc description, or preferably remove boilerplate javadoc.

shanthoosh: Removed.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -213,4 +215,11 @@ class JobConfig(config: Config) extends ScalaMapConfig(config) with Logging {
...
+   */
+  def getStreamPartitionMapperFactory: String = {
+    get(JobConfig.STREAM_PARTITION_MAPPER_FACTORY, classOf[HashStreamPartitionMapperFactory].getCanonicalName)
```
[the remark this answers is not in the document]
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/HashSystemStreamPartitionMapper.java
```diff
@@ -0,0 +1,42 @@
...
+ * under the License.
+ */
+package org.apache.samza.container.grouper.stream;
+
+import com.google.common.base.Preconditions;
+import org.apache.samza.Partition;
+import org.apache.samza.system.SystemStreamPartition;
+
+/**
+ * A SystemStreamPartitionMapper that uses the default hash partitioning function of kafka producer to map a {@link SystemStreamPartition} to
```
dnishimura: Is this partition mapper Kafka specific, or is it be applicable to any system? If the former, Kafka should be part of the implemented class name. If the latter, maybe just replace `kafka producer` with `underlying producer` or something similar in the comments.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -213,4 +214,8 @@ class JobConfig(config: Config) extends ScalaMapConfig(config) with Logging {
   def getJMXEnabled = {
     getBoolean(JobConfig.JOB_JMX_ENABLED, true);
   }
+
+  def getSystemStreamPartitionMapperFactoryName: String = {
+    get(JobConfig.SYSTEM_STREAM_PARTITION_MAPPER_FACTORY, classOf[HashSystemStreamPartitionMapperFactory].getCanonicalName)
```
dnishimura: Related to my other comment. If this SSP mapper is intended to be Kafka specific, then it should return a more generic SSP mapper factory by default.

shanthoosh: I think this is not kafka specific and can be extended to any custom producer partitioning function.

samza-api/src/main/java/org/apache/samza/container/grouper/stream/SystemStreamPartitionGrouper.java
```diff
@@ -18,12 +18,11 @@
  */
 package org.apache.samza.container.grouper.stream;
 
+import java.util.*;
```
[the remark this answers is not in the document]
prateekm: Still see it, please fix.

samza-api/src/main/java/org/apache/samza/container/grouper/stream/SystemStreamPartitionMapper.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package org.apache.samza.container.grouper.stream;
+
+import org.apache.samza.system.SystemStreamPartition;
+
+/**
+ * Input streams of a samza job can be either expanded or contracted by the user.
+ * This abstraction determines the previous {@link SystemStreamPartition} for a {@link SystemStreamPartition}
+ * of a input stream after the stream expansion or contraction. This will be used in {@link SystemStreamPartitionGrouper}
+ * implementations to generate partition expansion aware task assignments.
```
prateekm: Remove "This will be used in SSPGrouper implementations ...", not true anymore.

samza-api/src/main/java/org/apache/samza/container/grouper/stream/SystemStreamPartitionMapper.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public interface SystemStreamPartitionMapper {
+
+  /**
+   * Determines the previous {@link SystemStreamPartition} for a {@link SystemStreamPartition}
+   * of a input stream after the stream expansion or contraction.
+   * @param currentSystemStreamPartition denotes the current partition after the stream partition count change.
+   * @param previousPartitionCount the partition count of the stream before the stream partition count change.
+   * @param afterPartitionCount the partition count of the stream after the stream partition count change.
```
prateekm: currentPartitionCount for consistency.

samza-api/src/main/java/org/apache/samza/container/grouper/stream/SystemStreamPartitionMapper.java
```diff
@@ -0,0 +1,41 @@
...
+   * @return the mapped {@link SystemStreamPartition}.
+   */
+  SystemStreamPartition getPreviousSSP(SystemStreamPartition currentSystemStreamPartition,
+                                       int previousPartitionCount, int afterPartitionCount);
```
prateekm: Indent by 4, don't align params to parens open.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/GroupBySystemStreamPartition.java
```diff
@@ -29,20 +29,16 @@
 import org.apache.samza.container.TaskName;
 import org.apache.samza.system.SystemStreamPartition;
 
+/**
+ * An implementation of {@link SystemStreamPartitionGrouper} that assigns the input {@link SystemStreamPartition} to a separate task.
```
prateekm: "assigns each input"

samza-core/src/main/java/org/apache/samza/container/grouper/stream/GroupByPartition.java
```diff
@@ -23,61 +23,50 @@
...
 import java.util.Set;
-
 import org.apache.samza.config.Config;
 import org.apache.samza.config.TaskConfigJava;
 import org.apache.samza.container.TaskName;
 import org.apache.samza.system.SystemStreamPartition;
 
+/**
+ * An implementation of {@link SystemStreamPartitionGrouper} that groups the input stream partitions according to their partition number.
+ * This leads to a single task processing all the messages of a single partition(e.g. partition 0) across all the input streams.
```
prateekm: "single partition (e.g. partition 0)" <- space before parens open.

Thanks for adding this.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/HashSystemStreamPartitionMapper.java
```diff
@@ -0,0 +1,42 @@
...
+ * A SystemStreamPartitionMapper that uses the hash partitioning function of the producer to map a {@link SystemStreamPartition} to
+ * correct previous {@link SystemStreamPartition} after the stream expansion.
+ */
+public class HashSystemStreamPartitionMapper implements SystemStreamPartitionMapper {
+
+  @Override
+  public SystemStreamPartition getPreviousSSP(SystemStreamPartition currentSystemStreamPartition, int previousPartitionCount, int afterPartitionCount) {
+    Preconditions.checkNotNull(currentSystemStreamPartition);
+    Preconditions.checkArgument(afterPartitionCount % previousPartitionCount == 0,
+                                String.format("New partition count: %d should be a multiple of previous partition count: %d.", afterPartitionCount, previousPartitionCount));
```
prateekm: Indent by 4, dont' align to parens open.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
prateekm: Use getName instead of getCanonicalName.

samza-core/src/test/java/org/apache/samza/container/grouper/stream/TestGroupByPartition.java
```diff
@@ -21,110 +21,275 @@
 
 import static org.junit.Assert.*;
```
prateekm: Static imports at bottom.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/GroupByPartition.java
shanthoosh: Done.

samza-core/src/test/java/org/apache/samza/container/grouper/stream/TestGroupByPartition.java
shanthoosh: Done.

PR #805, opened 2018-11-13
SAMZA-1972: Make Operator Timer metrics calculation configurable

xinyuiscool:
This patch introduces two changes:
1. Make the timer metrics in OperatorImpl to be optional, and disabled by default. Adding TimerMetrics has quite a big performance impact which affects jobs with large number of operators, so it should be turned on for debugging only.
2. Register operator-level metrics on the container metrics registry. The task level registry has too many metrics which are usually ignored by the users. Having it in the container level will reduce the total amount of metrics published as well as the memory footprint.

Tested by hello-samza and works as expected.

### DIFF
--- a/docs/learn/documentation/versioned/jobs/configuration-table.html
+++ b/docs/learn/documentation/versioned/jobs/configuration-table.html
@@ -2239,6 +2239,23 @@
                         60 seconds.
                     </td>
                 </tr>
+                <tr>
+                    <td class="property" id="metrics-timer-enabled">metrics.timer.enabled</td>
+                    <td class="default">true</td>
+                    <td class="description">
+                        This setting enables the common timer metrics for your job, which include container metrics
+                        such as process-ns, window-ns, commit-ns, and block-ns, as well as key-value storage engine
+                        metrics such as get-ns, put-ns and range-ns.
+                    </td>
+                </tr>
+                <tr>
+                    <td class="property" id="metrics-timer-debug-eanbled">metrics.timer.debug.enabled</td>
+                    <td class="default">false</td>
+                    <td class="description">
+                        This setting enables the additional timer metrics for debugging of your job. These metrics
+                        include operator metrics such as handle-message-ns and handle-timer-ns.
+                    </td>
+                </tr>
 
                 <tr>
                     <th colspan="3" class="section" id="hdfs-system-producer"><a href="../hdfs/producer.html">Writing to HDFS</a></th>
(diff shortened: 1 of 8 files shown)


docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -2239,6 +2239,23 @@ <h1>Samza Configuration Reference</h1>
                         60 seconds.
                     </td>
                 </tr>
+                <tr>
```
vjagadish1989: What do you think about consolidating metrics-reporting configs into "metric.reporting.level", which can be set to either "info", "debug" or "critical". By default, all metric levels are reported with "info". 

A nice property is that we probably wouldn't require 2 configs: `metrics.timer.debug.enabled` and `metrics.timer.enabled`.. It would also allow you to broaden the scope instead of just timers.

xinyuiscool: This idea is confusing to the users since in the code they don't define the levels.

vjagadish1989: thanks, maybe add it to the config-reference table in the new documentation? 

also, document the relationship between the 2 metrics? eg: the override behavior when metrics.timer.enabled = false and metrics.timer.debug.enabled = true

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImpl.java
```diff
@@ -495,7 +498,9 @@ protected String getOpImplId() {
   }
 
   private HighResolutionClock createHighResClock(Config config) {
-    if (new MetricsConfig(config).getMetricsTimerEnabled()) {
+    MetricsConfig metricsConfig = new MetricsConfig(config);
+    // The timer metrics calculation here is only enabled for debugging
+    if (metricsConfig.getMetricsTimerDebugEnabled()) {
```
sborya: doesn't this have to be both Timer and TimerDebug enabled?

xinyuiscool: My old impl has both, but i was thinking it's clearer to have them separated so each flag controls a different set of timers. If you think the first one should have an overall control, I can add that back.

docs/learn/documentation/versioned/jobs/configuration-table.html
xinyuiscool: Right, I made the condition to be based on both timer and timer.debug enabled.

PR #806, opened 2018-11-15
[SAMZA-1997] Samza-sql diagnostics - instrument project operator

shenodaguirguis:
When the user uses Samza-SQL, they use high level declarative language (SQL) for ease and speed of implementation of their Samza Job. Therefore, monitoring the job should provide metrics at this high/logical level. This is the goal of the Samza-SQL diagnostics project. In this first baby-step, we start with instrumenting the Project operator to provide run-time metrics.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/translator/ProjectTranslator.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/translator/ProjectTranslator.java
@@ -60,20 +60,24 @@ class ProjectTranslator {
     this.queryId = queryId;
   }
 
+  /**
+   * ProjectMapFunction implements MapFunction to map input SamzaSqlRelMessages, one at a time, to a new
+   * SamzaSqlRelMessage which consists of the projected fields
+   */
   @VisibleForTesting
   public static class ProjectMapFunction implements MapFunction<SamzaSqlRelMessage, SamzaSqlRelMessage> {
     private transient Project project;
     private transient Expression expr;
-    private transient TranslatorContext context;
+    private transient TranslatorContext translatorContext;
     private transient MetricsRegistry metricsRegistry;
-    private transient SamzaHistogram processingTimeHistogram; // milli-seconds
-    private transient Counter messagesCounter;
+    private transient SamzaHistogram processingTime; // milli-seconds
+    private transient Counter numEvents;
 
     private final int queryId;
     private final int projectId;
     private final String logicalOpId;
-    private final String PROCESSING_TIME_HISTOGRAM_NAME = "processingTimeHistogram";
-    private final String MESSAGES_COUNTER_NAME = "messagesCounter";
+    private final String PROCESSING_TIME_NAME = "processingTime";
+    private final String NUM_EVENTS_NAME = "numEvents";
 
     ProjectMapFunction(int projectId, int queryId, String logicalOpId) {
       this.projectId = projectId;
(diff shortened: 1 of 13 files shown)


samza-sql/src/main/java/org/apache/samza/sql/translator/ProjectTranslator.java
```diff
@@ -83,9 +104,20 @@ public SamzaSqlRelMessage apply(SamzaSqlRelMessage message) {
...
         names.add(index, project.getNamedProjects().get(index).getValue());
       }
-
+      updateMetrics(arrivalTime, Instant.now());
       return new SamzaSqlRelMessage(names, Arrays.asList(output));
     }
+
+    private void updateMetrics(Instant arrivalTime, Instant outputTime) {
+      messagesCounter.inc();
+      processingTimeHistogram.update(Duration.between(arrivalTime, outputTime).toMillis());
```
atoomula: Would a metric smaller than millis be better considering that for many jobs projection would be trivial processing ?

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
```diff
@@ -131,6 +131,7 @@ public void translate(RelRoot relRoot, TranslatorContext translatorContext, int
     node.accept(new RelShuttleImpl() {
       int windowId = 0;
       int joinId = 0;
```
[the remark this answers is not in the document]
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/ProjectTranslator.java
shenodaguirguis: discussed offline: unit need to be consistent across all operators, and milliseconds makes more sense for joins for example, so will keep it milliseconds (and might revisit this when we see actual  stats)

samza-sql/src/main/java/org/apache/samza/sql/translator/ProjectTranslator.java
```diff
@@ -53,28 +60,42 @@
...
   }
 
-  private static class ProjectMapFunction implements MapFunction<SamzaSqlRelMessage, SamzaSqlRelMessage> {
+  @VisibleForTesting
+  public static class ProjectMapFunction implements MapFunction<SamzaSqlRelMessage, SamzaSqlRelMessage> {
     private transient Project project;
     private transient Expression expr;
     private transient TranslatorContext context;
+    private transient MetricsRegistry metricsRegistry;
+    private transient SamzaHistogram processingTimeHistogram; // milli-seconds
```
srinipunuru: nit: As a general practice, Samza code base doesn't have type related details to the variable Name. You can just call it as processingTime and numMessages or messages.

samza-sql/src/main/java/org/apache/samza/sql/translator/ProjectTranslator.java
```diff
@@ -53,28 +60,42 @@
...
+    private transient Counter messagesCounter;
 
     private final int queryId;
     private final int projectId;
+    private final String logicalOpId;
+    private final String PROCESSING_TIME_HISTOGRAM_NAME = "processingTimeHistogram";
+    private final String MESSAGES_COUNTER_NAME = "messagesCounter";
 
-    ProjectMapFunction(int projectId, int queryId) {
+    ProjectMapFunction(int projectId, int queryId, String logicalOpId) {
```
srinipunuru: We should have added doc on this method before. Looks like we missed it. Can you add it please?

samza-sql/src/test/java/org/apache/samza/sql/testutil/MetricsRegistry.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.samza.metrics.Counter;
+import org.apache.samza.metrics.Gauge;
+import org.apache.samza.metrics.ListGauge;
+import org.apache.samza.metrics.Timer;
+
+
+public class MetricsRegistry implements org.apache.samza.metrics.MetricsRegistry {
```
srinipunuru: since it is created for testing purposes. Can we call this as a TestMetricsRegistryImpl?

srinipunuru: CAn you add class and method documentations

samza-sql/src/main/java/org/apache/samza/sql/translator/ProjectTranslator.java
shenodaguirguis: done

samza-sql/src/test/java/org/apache/samza/sql/testutil/MetricsRegistry.java
shenodaguirguis: done

shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/ProjectTranslator.java
shenodaguirguis: done

srinipunuru:
@shenodaguirguis Looks like Travis is failing, Please take a look and fix the issue before we can merge this PR.

shenodaguirguis:
@srinipunuru it seems to be some sort of flake issue as it runs successfully locally (multiple times) and the code it touches has the only change of using logicalOpId (of JoinTranslator) instead of joinId

PR #807, opened 2018-11-15
SAMZA-1998: Table API refactoring

weisong44:
Table API refactoring
     - Removed TableSpec
     - Consolidated configuration for tables to table descriptors
     - Refactored constructor so that only local table would require serde's
     - Removed table provider for RocksDB- and in-memory tables, and added LocalTableProvider
     - Various refactoring for unit tests


samza-api/src/main/java/org/apache/samza/table/descriptors/BaseTableDescriptor.java
```diff
@@ -78,33 +68,33 @@ public String getTableId() {
...
   }
 
-  /**
-   * Generate config for {@link TableSpec}; this method is used internally.
-   * @param tableSpecConfig configuration for the {@link TableSpec}
-   */
-  protected void generateTableSpecConfig(Map<String, String> tableSpecConfig) {
-    tableSpecConfig.putAll(config);
+  public String getProviderFactoryClassName() {
+    throw new SamzaException("Not implemented");
```
prateekm: Is the idea that extending classes should provide this? If so, maybe make this an abstract method instead?

samza-api/src/main/java/org/apache/samza/table/descriptors/BaseTableDescriptor.java
```diff
@@ -78,33 +68,33 @@ public String getTableId() {
...
+  @Override
+  public Map<String, String> toConfig(Config jobConfig) {
+
+    validate();
+
+    Map<String, String> tableConfig = new HashMap<>(config);
+    tableConfig.put(
+        String.format(JavaTableConfig.TABLE_PROVIDER_FACTORY, tableId),
+        getProviderFactoryClassName());
+    generateConfig(jobConfig, tableConfig);
```
prateekm: The pattern we're using in StreamDescriptors is to have the extending classes also override the toConfig method from the base class and call `super.toConfig` as the first step. Might be a simpler interface for extending classes since there is one less config related interface method. If we do that, you might also not need to call addTableConfig to mutate the passed in `tableConfig` map in generateConfig.

samza-api/src/main/java/org/apache/samza/table/descriptors/CachingTableDescriptor.java
```diff
@@ -158,6 +130,34 @@ public TableSpec getTableSpec() {
     return this;
   }
 
+  @Override
+  public String getProviderFactoryClassName() {
+    return PROVIDER_FACTORY_CLASS_NAME;
+  }
+
+  @Override
+  protected void generateConfig(Config jobConfig, Map<String, String> tableConfig) {
```
prateekm: What's `tableConfig` in the method params?

samza-api/src/main/java/org/apache/samza/table/descriptors/LocalTableDescriptor.java
```diff
@@ -130,20 +125,64 @@ public D withChangelogReplicationFactor(int replicationFactor) {
...
+      addStoreConfig(tableConfig, "side.inputs.processor.serialized.instance",
+          SerdeUtils.serialize("Side Inputs Processor", sideInputsProcessor));
+    }
+
+    // Changelog configuration
     if (enableChangelog) {
-      if (changelogStream != null) {
-        tableSpecConfig.put(INTERNAL_CHANGELOG_STREAM, changelogStream);
+      if (StringUtils.isEmpty(changelogStream)) {
+        String jobName = jobConfig.containsKey("job.name")
```
prateekm: Can use JobConfig.getName.get(). Not providing a job.name should be an error, we shouldn't default to "unknown".

samza-api/src/main/java/org/apache/samza/table/descriptors/LocalTableDescriptor.java
```diff
@@ -130,20 +125,64 @@ public D withChangelogReplicationFactor(int replicationFactor) {
...
+            ? jobConfig.get("job.name")
+            : "unknown";
+        String jobId = jobConfig.containsKey("job.id")
```
prateekm: Can use JobConfig.getJobId().

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -208,6 +210,9 @@ JobGraph createJobGraph(ApplicationDescriptorImpl<? extends ApplicationDescripto
         OperatorSpecGraphAnalyzer.getJoinToInputOperatorSpecs(
             jobGraph.getApplicationDescriptorImpl().getInputOperators().values());
 
+    Map<String, TableDescriptor> tableDescMap = jobGraph.getTables().stream()
```
prateekm: Minor: `tableDescriptors`

samza-core/src/main/java/org/apache/samza/operators/TableImpl.java
```diff
@@ -20,22 +20,26 @@
...
 import org.apache.samza.table.Table;
-import org.apache.samza.table.TableSpec;
+import org.apache.samza.table.descriptors.TableDescriptor;
 
 
 /**
- * This class is the holder of a {@link TableSpec}
+ * This class is the holder of a table Id
  */
 public class TableImpl implements Table, Serializable {
```
prateekm: Do we still need this class? Looks like this isn't doing much.

samza-core/src/main/java/org/apache/samza/table/BaseTableProvider.java
```diff
@@ -48,24 +48,10 @@ public BaseTableProvider(TableSpec tableSpec) {
...
-    // Insert table_id prefix to config entries
-    tableSpec.getConfig().forEach((k, v) -> {
-        String realKey = String.format(JavaTableConfig.TABLE_ID_PREFIX, tableSpec.getId()) + "." + k;
-        tableConfig.put(realKey, v);
-      });
-
-    logger.info("Generated configuration for table " + tableSpec.getId());
-
-    return tableConfig;
+  public void close() {
```
prateekm: Minor: Add Override annotation?

samza-core/src/main/java/org/apache/samza/table/TableConfigGenerator.java
```diff
@@ -19,80 +19,24 @@
...
-        // Generate additional configuration
-        TableProviderFactory tableProviderFactory =
-            Util.getObj(tableSpec.getTableProviderFactoryClassName(), TableProviderFactory.class);
-        TableProvider tableProvider = tableProviderFactory.getTableProvider(tableSpec);
-        tableConfigs.putAll(tableProvider.generateConfig(config, tableConfigs));
-      });
-
-    LOG.info("TableConfigGenerator has generated configs {}", tableConfigs);
-    return tableConfigs;
+  static public Map<String, String> generate(Config jobConfig, List<TableDescriptor> tableDescriptors) {
```
prateekm: Minor: From what I can tell, this is only used in one place (JobNodeConfigurationGenerator). Since this is fairly simple now, would prefer removing this class and inlining the generate method there.

samza-api/src/main/java/org/apache/samza/table/descriptors/LocalTableDescriptor.java
weisong44: JobConfig currently resides in core, and it has a bunch of dependencies in core as well, we could change to use it, once it's moved to api as well.

samza-api/src/main/java/org/apache/samza/table/descriptors/LocalTableDescriptor.java
weisong44: Same as above

samza-core/src/main/java/org/apache/samza/operators/TableImpl.java
weisong44: It is used in building the DAG in fluent API, but replacing it with table Id might have side effects in the future, if we need to add attach any properties to a table. I'd say let's keep it for now.

samza-core/src/main/java/org/apache/samza/table/TableConfigGenerator.java
weisong44: It is actually used in samba-frame, which is a separate MP.

samza-api/src/main/java/org/apache/samza/table/descriptors/CachingTableDescriptor.java
weisong44: It's now refactored away

samza-core/src/main/java/org/apache/samza/table/TableManager.java
```diff
@@ -70,26 +67,11 @@
...
-            serdes.get(tableConfig.getValueSerde(tableId)));
-
-        TableSpec tableSpec = new TableSpec(tableId, serde, tableProviderFactory,
-            config.subset(String.format(JavaTableConfig.TABLE_ID_PREFIX, tableId) + "."));
-
-        addTable(tableSpec);
-
-        logger.info("Added table " + tableSpec.getId());
+        addTable(tableId, config);
+        logger.info("Added table " + tableId);
```
prateekm: Minor: logger.debug?

samza-api/src/main/java/org/apache/samza/table/descriptors/TableDescriptor.java
```diff
@@ -72,4 +75,18 @@
...
+   *
+   * Note: although the serdes may have already been set in this instance, their
+   * corresponding configuration needs to be generated centrally for consistency
+   * and efficiency reasons. Therefore the serde configuration for this table
+   * is expected to have already been generated and stored in the {@code jobConfig}.
+   *
+   * @param jobConfig job configuration
+   * @return table configuration
+   */
+  Map<String, String> toConfig(Config jobConfig);
```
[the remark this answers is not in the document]
weisong44: Yeah, will consolidate serde generation to JobNodeConfigGenerator.

PR #810, opened 2018-11-19
SAMZA-1707: Samza onTimer method triggering before init

xinyuiscool:
Currently there was a bug when registering a timer with a very short amount of delay, it might not be invoked since it depends on the creation of the run loop. This patch fixed the problem by double checking the ready timers when run loop is created (listener is registered.)

### DIFF
--- a/samza-test/src/test/java/org/apache/samza/test/functions/SchedulerFunctionTest.java
+++ b/samza-test/src/test/java/org/apache/samza/test/functions/SchedulerFunctionTest.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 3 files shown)


xinyuiscool:
gradle build works fine locally. The travis failure is not related to this change.

samza-test/src/test/java/org/apache/samza/test/functions/SchedulerFunctionTest.java
```diff
@@ -0,0 +1,89 @@
...
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class SchedulerFunctionTest {
```
prateekm: 'TestSchedulerFunction' to keep existing convention.

vjagadish1989 (approved):
approved, looks great to me.

PR #811, opened 2018-11-19
SAMZA-1994 Table API: Add missed key lookups metric for table reads

atoomula:


weisong44:
LGTM, thanks for the enhancement!!

PR #814, opened 2018-11-26
SAMZA-2002 SamzaSQL Diagnostics: instrument rest of operators (except join) and at Query level

shenodaguirguis:
After instrumenting Project operator in Samza-1997, this PR instruments the rest of operators, except for Join. It also instrument the Query Level to add its metrics. Latency measured at query level is actually the time span of the message/event since when it is arrived to end of processing. This skips the pre-arrival network latency, which we are considering adding later, whenever feasible.


shenodaguirguis:
closing this as it has some git error, creating a replacement PR

PR #825, opened 2018-11-28
SAMZA-2002 SamzaSQL Diagnostics: instrument rest of operators (except join) and at Query level

shenodaguirguis:


samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -64,6 +77,7 @@ public SamzaSqlRelMessage(List<String> fieldNames, List<Object> fieldValues) {
     }
     this.key = key;
     this.samzaSqlRelRecord = new SamzaSqlRelRecord(fieldNames, fieldValues);
+    this.samzaSqlRelMsgMetadata = new SamzaSqlRelMsgMetadata(Instant.now().toString());
```
atoomula: Would it better for the caller to decide on what time to put here ? IOW, should we make this constructor private ? Given that this is core datastructure in Samza SQL code, I would avoid having too many public constructors. We should also strive to make SamzaSqlRelMessage to not have any more smarts that it already has.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -89,13 +109,21 @@ public SamzaSqlRelMessage(Object key, List<String> fieldNames, List<Object> fiel
     tmpFieldValues.addAll(fieldValues);
 
     this.samzaSqlRelRecord = new SamzaSqlRelRecord(tmpFieldNames, tmpFieldValues);
+    this.samzaSqlRelMsgMetadata = new SamzaSqlRelMsgMetadata(Instant.now().toString());
```
atoomula: Same above. Make this constructor private.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -89,13 +109,21 @@ public SamzaSqlRelMessage(Object key, List<String> fieldNames, List<Object> fiel
...
-  public SamzaSqlRelMessage(@JsonProperty("samzaSqlRelRecord") SamzaSqlRelRecord samzaSqlRelRecord) {
+  public SamzaSqlRelMessage(@JsonProperty("samzaSqlRelRecord") SamzaSqlRelRecord samzaSqlRelRecord,
+      @JsonProperty("samzaSqlRelMsgMetadata") SamzaSqlRelMsgMetadata metadata) {
+    this(samzaSqlRelRecord);
+    this.samzaSqlRelMsgMetadata = metadata;
+  }
+
+  public SamzaSqlRelMessage(SamzaSqlRelRecord samzaSqlRelRecord) {
     this(samzaSqlRelRecord.getFieldNames(), samzaSqlRelRecord.getFieldValues());
   }
```
atoomula: This constructor irrespective of what we decide on metadata should be private.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -104,6 +132,16 @@ public SamzaSqlRelRecord getSamzaSqlRelRecord() {
...
 
+  @JsonProperty("samzaSqlRelMsgMetadata")
+  public void setSamzaSqlRelMsgMetadata(SamzaSqlRelMsgMetadata metadata) {
+    samzaSqlRelMsgMetadata = metadata;
+  }
+
+  @JsonProperty("samzaSqlRelMsgMetadata")
+  public SamzaSqlRelMsgMetadata getSamzaSqlRelMsgMetadata() { return samzaSqlRelMsgMetadata; }
+
+  public boolean hasMeatadata() { return samzaSqlRelMsgMetadata != null; }
```
atoomula: Meat -> Meta

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
```diff
@@ -0,0 +1,51 @@
...
+
+package org.apache.samza.sql.data;
+
+import java.io.Serializable;
+import java.time.Instant;
+import org.codehaus.jackson.annotate.JsonProperty;
+
+
+/**
+ * Metadata of Samza Sql Rel Message. Conains metadata about the corresponding event or
```
atoomula: Contains

PR #826, opened 2018-11-29
SAMZA-2009: Add ability to disable table metrics

weisong44:
This is to back port SAMZA-2004 to 0.14.1, due to the extend of changed made in Samza 1.0, the back port is actually done manually.

More details can be found here
https://github.com/apache/samza/pull/822


prateekm:
@weisong44 Does this need to be merged in? If not, let's close the PR.

weisong44:
This PR won't be needed.

PR #827, opened 2018-11-29
SAMZA-2010: Handle null value in LocalReadWriteTable.putAll()

weisong44:
To be consistent with put(), null values in input should be delete operation


samza-kv/src/main/java/org/apache/samza/storage/kv/LocalReadWriteTable.java
```diff
@@ -69,7 +70,20 @@ public void put(K key, V value) {
...
+    List<K> toDelete = new LinkedList<>();
+    entries.forEach(e -> {
+        if (e.getValue() != null) {
+          toPut.add(e);
+        } else {
+          toDelete.add(e.getKey());
+        }
+      });
+
+    instrument(writeMetrics.numPutAlls, writeMetrics.putAllNs, () -> kvStore.putAll(toPut));
```
ahmedahamid: We probably need to check if `toPut` is empty as well.

PR #828, opened 2018-11-29
SAMZA-2011: Add ability to disable store timers

weisong44:
Add same ability for store timers as in SAMZA-2004
https://issues.apache.org/jira/browse/SAMZA-2004


samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -143,7 +143,7 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
 
     // create the storage engine and return
     // TODO: Decide if we should use raw bytes when restoring
-    val keyValueStorageEngineMetrics = new KeyValueStorageEngineMetrics(storeName, registry)
+    val keyValueStorageEngineMetrics = new KeyValueStorageEngineMetrics(storeName, registry, jobContext.getConfig)
```
prateekm: Isn't the code below sufficient to disable timers? IIRC it returns the same value for each timer update which reduces the overhead.

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngineMetrics.scala
```diff
@@ -38,16 +40,18 @@ class KeyValueStorageEngineMetrics(
...
 
-  val getNs = newTimer("get-ns")
-  val getAllNs = newTimer("get-all-ns")
-  val putNs = newTimer("put-ns")
-  val putAllNs = newTimer("put-all-ns")
-  val deleteNs = newTimer("delete-ns")
-  val deleteAllNs = newTimer("delete-all-ns")
+  val timerEnabled = new MetricsConfig(config).getMetricsTimerEnabled
+
+  val getNs = if (timerEnabled) newTimer("get-ns") else null
```
prateekm: Instead of returning null values I'd prefer to do this check at call sites and not update timers. Otherwise we can run into NPEs if caller doesn't always check.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
weisong44: Thanks, it seems this is sufficient, will close this PR.

PR #829, opened 2018-11-29
SAMZA-2012: Add API for wiring an external context through to application processing code

cameronlee314:
This PR also refactors TestSamzaSqlRemoteTable to be in samza-test instead of samza-sql, since it seems to actually be an integration test. It is useful to move that test in this PR so that tests that may need an external context can be consolidated.


samza-api/src/main/java/org/apache/samza/context/ApplicationContainerContextFactory.java
```diff
@@ -35,14 +36,42 @@
...
  */
+@InterfaceStability.Evolving
 public interface ApplicationContainerContextFactory<T extends ApplicationContainerContext> extends Serializable {
-
   /**
    * Creates an instance of the application-defined {@link ApplicationContainerContext}.
+   * <p>
+   * Applications should implement this to provide a context for container initialization.
    *
+   * @param externalContext external context provided for the application; null if it was not provided
```
prateekm: For consistency: "application-defined external context ...".
Same for all other classes.

samza-api/src/main/java/org/apache/samza/context/ApplicationContainerContextFactory.java
```diff
@@ -35,14 +36,42 @@
...
+   * @param containerContext framework-provided container context
+   * @return a new instance of the application-defined {@link ApplicationContainerContext}
+   *
+   * Deprecated: Applications should implement {@link #create(ExternalContext, JobContext, ContainerContext)} directly.
+   * This is being left here for backwards compatibility.
+   */
+  @Deprecated
+  default T create(JobContext jobContext, ContainerContext containerContext) {
+    // adding this here so that new apps do not need to implement this
+    throw new UnsupportedOperationException("Please implement #create(ExternalContext, JobContext, ContainerContext) "
```
prateekm: For existing users already implementing this method, this exception doesn't matter.
For new users, this will only happen if they don't override either create. If the idea is to prevent that, maybe remove ", or at least ..." from the message?

samza-api/src/main/java/org/apache/samza/context/Context.java
```diff
@@ -76,4 +76,13 @@
    * @throws IllegalStateException if no {@link ApplicationTaskContextFactory} was provided for the application
    */
   ApplicationTaskContext getApplicationTaskContext();
+
+  /**
+   * Gets the {@link ExternalContext} that was created outside of the application.
+   * <p>
+   * Use {@link org.apache.samza.runtime.ApplicationRunner#run(ExternalContext)} to provide this context.
```
prateekm: Since this is Samza API, can import the class for readability instead of using the FQN. Same for other classes.

samza-core/src/main/java/org/apache/samza/context/ContextImpl.java
```diff
@@ -30,24 +30,26 @@
...
   private final Optional<ApplicationTaskContext> applicationTaskContextOptional;
+  private final Optional<ExternalContext> externalContextOptional;
 
   /**
    * @param jobContext non-null job context
    * @param containerContext non-null framework container context
    * @param taskContext non-null framework task context
    * @param applicationContainerContextOptional optional application-defined container context
    * @param applicationTaskContextOptional optional application-defined task context
+   * @param externalContextOptional optional external context
```
prateekm: Nitpick: application-defined external context

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
```diff
@@ -71,4 +73,13 @@ public static void main(String[] args) throws Exception {
         throw new IllegalArgumentException("Unrecognized operation: " + op);
     }
   }
+
+  private static Optional<ExternalContext> buildExternalContext(Config config) {
+    /*
+     * By default, use an empty ExternalContext here. In a custom fork of Samza, this can be implemented to pass
+     * a non-empty ExternalContext. Only config should be used to build the external context. In the future, components
+     * like the application descriptor may not be available.
```
prateekm: Did you mean "may be available"? I don't think we need to mention that, it's also weird to have the descriptors (or any app logic) be available for this, since this is expected to be external to the application.

prateekm: Actually I'm not sure if we need the external context here, since there's no application processing (and init) going on here - it's only planning.

samza-api/src/main/java/org/apache/samza/context/ApplicationContainerContextFactory.java
cameronlee314: I had considered this, but since external context is passed through the ApplicationRunner and not through the SamzaApplication (unlike ApplicationContainerContext/ApplicationTaskContext), I didn't want to couple it with being "application-defined".
Do you still think we should add that label?

samza-api/src/main/java/org/apache/samza/context/ApplicationContainerContextFactory.java
cameronlee314: I wanted to avoid requiring new users to implement the deprecated version of create, so I needed some default impl for the deprecated version. I was trying to note in the log message that implementing the deprecated version of create was not ideal, but the javadoc should be clear enough, so I'll simplify the log message.

samza-api/src/main/java/org/apache/samza/context/Context.java
cameronlee314: I thought you hadn't pushed the change yet to avoid the unused import check. It looked like other Samza API classes still were using the FQN. Could you confirm?

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
cameronlee314: This was related to the future plan of only doing planning once, and then passing a serialized version of the app around. I thought that the app descriptor was only needed for planning, and once the serialization happened, it would not be available. I wanted to clarify in the comment that external context should not be built with the app descriptor in case it is not available in the future. Is my understanding correct?
I agree that it might be odd to use app logic to build this, but you never know, so I added a note.
If ApplicationRunnerMain gets used to launch the LocalApplicationRunner, then I thought it might be needed.

samza-api/src/main/java/org/apache/samza/context/ApplicationContainerContextFactory.java
prateekm: I think the more important contrast is between "framework-provided" and "application-defined", so I'd rather keep it.

samza-api/src/main/java/org/apache/samza/context/Context.java
prateekm: I did, although I only cleaned up javadocs for classes in High Level API (https://github.com/apache/samza/pull/760).

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
prateekm: No, ApplicationRunnerMain is only for YARN application. Standalone users have thier own "main" (that of the application container), so they don't need this class.

samza-api/src/main/java/org/apache/samza/context/ApplicationContainerContextFactory.java
cameronlee314: The example I was thinking about was if someone implemented LocalContainerRunner.buildExternalContext to return some pre-defined set of beans which was not dependent on an application. To me, LocalContainerRunner is "framework" and not "application", so in this case, I felt that the external context was "framework provided".
We will be using the external context abstraction layer as "application-defined", but it seemed like there could be use cases where it falls more under the "framework-provided" category. Maybe we could chat in-person when you have some time.

samza-api/src/main/java/org/apache/samza/context/Context.java
cameronlee314: Thanks for confirming. I'll change this.

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
cameronlee314: Oh ok, since there was an ApplicationRunner interface and we classloaded the concrete ApplicationRunner, I thought that ApplicationRunnerMain could potentially run any app runner. I can remove this. I guess if someone is forking Samza already and needs external context here, then they could just wire that in on their own.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -211,19 +213,22 @@ public StreamProcessor(Config config, Map<String, MetricsReporter> customMetrics
...
-      StreamProcessorLifecycleListenerFactory listenerFactory, JobCoordinator jobCoordinator) {
+      Optional<ExternalContext> externalContextOptional, StreamProcessorLifecycleListenerFactory listenerFactory,
+      JobCoordinator jobCoordinator) {
     Preconditions.checkNotNull(listenerFactory, "StreamProcessorListenerFactory cannot be null.");
     this.config = config;
     this.customMetricsReporter = customMetricsReporters;
     this.taskFactory = taskFactory;
     this.applicationDefinedContainerContextFactoryOptional = applicationDefinedContainerContextFactoryOptional;
     this.applicationDefinedTaskContextFactoryOptional = applicationDefinedTaskContextFactoryOptional;
+    this.externalContextOptional = externalContextOptional;
```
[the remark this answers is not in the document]
cameronlee314: 1. Each of the context types have different scopes. ApplicationTaskContext is created per-task and associated with the lifecycle of a task. ApplicationContainerContext is per-container. ExternalContext is "outside" of Samza. Users have to manage lifecycle of ExternalContext on their own outside of running the container/processor.
2. If you want to do this, then you should not use ExternalContext. You should use ApplicationTaskContext/ApplicationContainerContext instead.
3. It is relevant in both modes. See LocalContainerRunner.buildExternalContext. Users will need an external fork to implement that method if they need it.
4. The lifecycle of ExternalContext is not considered within appRunner.run(), so there is no coupling. It must be managed outside of appRunner.run().

samza-api/src/main/java/org/apache/samza/context/ApplicationContainerContextFactory.java
cameronlee314: Discussed in-person. We will leave the documentation vague since it could be "framework-provided" or "application-provided" in different cases.

samza-core/src/main/java/org/apache/samza/context/ContextImpl.java
cameronlee314: See https://github.com/apache/samza/pull/829#discussion_r238491709

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
```diff
@@ -59,7 +61,7 @@ public static void main(String[] args) throws Exception {
 
     switch (op) {
       case RUN:
-        appRunner.run();
+        appRunner.run(null);
```
prateekm: I think can revert all changes in this class.

prateekm: Nvm, we didn't keep the old method in RemoteApplicationRunner. This is fine.

cameronlee314: I need to remove the unused imports here. I'm just running a local check before pushing the change.

PR #831, opened 2018-11-30
[SAMZA-2002] SamzaSQL Diagnostics: instrument rest of operators (except join & aggregate) and at Query level

shenodaguirguis:
Second phase of instrumenting SamzaSQL operators to add and maintain metrics. All operators, except join and aggregate, are instrumented to add Processing Time and Input Rate metrics. Whenever output rate could be different (e.g., filter operator) the output rate is also added. At query level, we have Query Latency, and input and output rates.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
@@ -20,6 +20,7 @@
 package org.apache.samza.sql.data;
 
 import java.io.Serializable;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
@@ -45,6 +46,12 @@ public class SamzaSqlRelMessage implements Serializable {
   private final SamzaSqlRelRecord samzaSqlRelRecord;
 
   /**
+   * hold metadata about the message or event, e.g., the eventTime timestamp
+   */
+  @JsonProperty("samzaSqlRelMsgMetadata")
+  private SamzaSqlRelMsgMetadata samzaSqlRelMsgMetadata;
+
+  /**
    * Creates a {@link SamzaSqlRelMessage} from the list of relational fields and values.
    * If the field list contains KEY, then it extracts the key out of the fields to create a
    * {@link SamzaSqlRelRecord} along with key, otherwise creates a {@link SamzaSqlRelRecord}
@@ -53,9 +60,11 @@ public class SamzaSqlRelMessage implements Serializable {
    * @param fieldValues  Ordered list of all the values in the row. Some of the fields can be null, This could be
    *                     result of delete change capture event in the stream or because of the result of the outer join
    *                     or the fields themselves are null in the original stream.
+   * @param metadata the message/event's metadata
    */
-  public SamzaSqlRelMessage(List<String> fieldNames, List<Object> fieldValues) {
+  public SamzaSqlRelMessage(List<String> fieldNames, List<Object> fieldValues, SamzaSqlRelMsgMetadata metadata) {
     Validate.isTrue(fieldNames.size() == fieldValues.size(), "Field Names and values are not of same length.");
+    Validate.notNull(metadata, "Message metadata is NULL");
 
     int keyIndex = fieldNames.indexOf(KEY_NAME);
     Object key = null;
@@ -64,7 +73,9 @@ public class SamzaSqlRelMessage implements Serializable {
     }
     this.key = key;
     this.samzaSqlRelRecord = new SamzaSqlRelRecord(fieldNames, fieldValues);
+    this.samzaSqlRelMsgMetadata = metadata;
   }
+
 
   /**
    * Create the SamzaSqlRelMessage, Each rel message represents a row in the table.
@@ -74,9 +85,11 @@ public class SamzaSqlRelMessage implements Serializable {
    * @param fieldValues Ordered list of all the values in the row. Some of the fields can be null, This could be result of
    *               delete change capture event in the stream or because of the result of the outer join or the fields
    *               themselves are null in the original stream.
+   * @param metadata the message/event's metadata
    */
-  public SamzaSqlRelMessage(Object key, List<String> fieldNames, List<Object> fieldValues) {
+  public SamzaSqlRelMessage(Object key, List<String> fieldNames, List<Object> fieldValues, SamzaSqlRelMsgMetadata metadata) {
     Validate.isTrue(fieldNames.size() == fieldValues.size(), "Field Names and values are not of same length.");
+    Validate.notNull(metadata, "Message metadata is NULL");
 
     List<String> tmpFieldNames = new ArrayList<>();
     List<Object> tmpFieldValues = new ArrayList<>();
@@ -89,20 +102,26 @@ public class SamzaSqlRelMessage implements Serializable {
     tmpFieldValues.addAll(fieldValues);
 
     this.samzaSqlRelRecord = new SamzaSqlRelRecord(tmpFieldNames, tmpFieldValues);
+    this.samzaSqlRelMsgMetadata = metadata;
   }
 
   /**
    * Creates the SamzaSqlRelMessage from {@link SamzaSqlRelRecord}.
    * @param samzaSqlRelRecord represents the rel record.
+   * @param metadata the message/event's metadata
    */
-  public SamzaSqlRelMessage(@JsonProperty("samzaSqlRelRecord") SamzaSqlRelRecord samzaSqlRelRecord) {
-    this(samzaSqlRelRecord.getFieldNames(), samzaSqlRelRecord.getFieldValues());
+  public SamzaSqlRelMessage(@JsonProperty("samzaSqlRelRecord") SamzaSqlRelRecord samzaSqlRelRecord,
+      @JsonProperty("samzaSqlRelMsgMetadata") SamzaSqlRelMsgMetadata metadata) {
+    this(samzaSqlRelRecord.getFieldNames(), samzaSqlRelRecord.getFieldValues(), metadata);
   }
 
   @JsonProperty("samzaSqlRelRecord")
   public SamzaSqlRelRecord getSamzaSqlRelRecord() {
     return samzaSqlRelRecord;
   }
+
+  @JsonProperty("samzaSqlRelMsgMetadata")
+  public SamzaSqlRelMsgMetadata getSamzaSqlRelMsgMetadata() { return samzaSqlRelMsgMetadata; }
 
   public Object getKey() {
     return key;
@@ -127,7 +146,7 @@ public class SamzaSqlRelMessage implements Serializable {
 
   @Override
   public String toString() {
-    return "RelMessage: {" + samzaSqlRelRecord + "}";
+    return "RelMessage: {" + samzaSqlRelRecord + " " + samzaSqlRelMsgMetadata + "}";
   }
 
   /**
@@ -172,4 +191,5 @@ public class SamzaSqlRelMessage implements Serializable {
     }
     return keyPartNames;
   }
+
 }
--- a/samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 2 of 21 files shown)


samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
```diff
@@ -67,28 +76,56 @@
...
+      outputEvents = metricsRegistry.newCounter(queryLogicalId, TranslatorConstants.OUTPUT_EVENTS_NAME);
+      outputEvents.clear();
     }
 
     @Override
     public KV<Object, Object> apply(SamzaSqlRelMessage message) {
-      return this.samzaMsgConverter.convertToSamzaMessage(message);
+      if (!message.hasMetadata()) {
+        LOG.warn("Output message with no Metadata: " + message);
+        message.setSamzaSqlRelMsgMetadata(new SamzaSqlRelMsgMetadata(Instant.now().toString()));
```
atoomula: If we agree that when SamzaSqlRelMessage is created, event time or current time is always set, then we won't have to add this code. There is only one place where this determination of event time vs current time needs to be done and that is in Rel converter in convertToRelMessage API implementation. And at all other places where SamzaSqlRelMessage is created (like ProjectTranslator, Join and QueryTranslator outputfunction), we get the time from the input rel message.

samza-sql/src/main/java/org/apache/samza/sql/translator/FilterTranslator.java
```diff
@@ -45,49 +52,83 @@
     this.queryId = queryId;
   }
 
-  private static class FilterTranslatorFunction implements FilterFunction<SamzaSqlRelMessage> {
+  /**
+   * FilterTranslatorFunction to process input events, apply the filter and produce output
+   * events accordingly
+   */
+  @VisibleForTesting
```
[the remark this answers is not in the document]
shenodaguirguis: good catch! 
Reverted back to private (there was a phase when I needed to public it for debugging)

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -53,8 +60,9 @@
    * @param fieldValues  Ordered list of all the values in the row. Some of the fields can be null, This could be
    *                     result of delete change capture event in the stream or because of the result of the outer join
    *                     or the fields themselves are null in the original stream.
+   * @param metadata the message/event's metadata
    */
-  public SamzaSqlRelMessage(List<String> fieldNames, List<Object> fieldValues) {
+  public SamzaSqlRelMessage(List<String> fieldNames, List<Object> fieldValues, SamzaSqlRelMsgMetadata metadata) {
     Validate.isTrue(fieldNames.size() == fieldValues.size(), "Field Names and values are not of same length.");
```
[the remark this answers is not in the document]
shenodaguirguis: agree. done

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
```diff
@@ -0,0 +1,94 @@
...
+   * the timestamp when SamzaSQL query starts processing the event
+   * set by the SamzaSQL Scan operator
+   */
+  @JsonProperty("scanTime")
+  private String scanTime;
+
+  public SamzaSqlRelMsgMetadata(@JsonProperty("eventTime") String eventTime, @JsonProperty("arrivalTime") String arrivalTime,
+      @JsonProperty("scanTime") String scanTime) {
+    this.eventTime = eventTime;
+    this.arrivalTime =arrivalTime;
```
atoomula: space after =

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
```diff
@@ -0,0 +1,94 @@
...
+  public boolean hasEventTime() { return eventTime != null && !eventTime.isEmpty(); }
+
+  @JsonProperty("arrivalTime")
+  public String getarrivalTime() { return arrivalTime;}
+
+  public void setArrivalTime(String arrivalTime) {
+    this.arrivalTime = arrivalTime;
+  }
+
+  public boolean hasArrivalTime() { return arrivalTime != null && !arrivalTime.isEmpty(); }
```
atoomula: maybe I'm paranoid, but do we need these checks for arrival time and scan time ? I understand event time could be set or not. The reason for my paranoia is because this is the prime path where each and every event will have to go through these checks in every operator.

shenodaguirguis: I had the same worry, and for now we do need to check if arrivalTime is set or not, until we do set the arrivalTIme. But even after we do set arrivalTime, we will check/validate it is set (as in scanTime case now)

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
shenodaguirguis: discussed offline: added comments on scanTime and arrivalTime checks that they need to be removed once code is stable, and that eventTime check needs to stay there because it's availability depends on source of event

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
```diff
@@ -190,9 +245,9 @@ public RelNode visit(LogicalAggregate aggregate) {
     sendToOutputStream(outputSystemStream, streamAppDescriptor, translatorContext, node, queryId);
   }
 
-  private void sendToOutputStream(String sinkStream, StreamApplicationDescriptor appDesc, TranslatorContext context, RelNode node, int queryId) {
+  private void sendToOutputStream(String sinkStream, StreamApplicationDescriptor appDesc, TranslatorContext translatorContext, RelNode node, int queryId) {
     SqlIOConfig sinkConfig = sqlConfig.getOutputSystemStreamConfigsBySource().get(sinkStream);
```
srinipunuru: We don't have an operator for SendTo, I think we should represent sendTo as an operator and can we have the following metrics

1. processing time
2. event rate.

samza-sql/src/main/java/org/apache/samza/sql/translator/TranslatorConstants.java
```diff
@@ -0,0 +1,32 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.sql.translator;
+
+public class TranslatorConstants {
+  public static final String PROCESSING_TIME_NAME = "processingTime";
+  public static final String TOTAL_LATENCY_NAME = "totalLatency";
```
srinipunuru: We generally add unit of time(Ms) to these metrics for e.g. totalLatencyMs so that it's clear to the user.

shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/ProjectTranslator.java
```diff
@@ -64,20 +64,17 @@
...
-  @VisibleForTesting
-  public static class ProjectMapFunction implements MapFunction<SamzaSqlRelMessage, SamzaSqlRelMessage> {
+  private static class ProjectMapFunction implements MapFunction<SamzaSqlRelMessage, SamzaSqlRelMessage> {
     private transient Project project;
     private transient Expression expr;
     private transient TranslatorContext translatorContext;
     private transient MetricsRegistry metricsRegistry;
     private transient SamzaHistogram processingTime; // milli-seconds
-    private transient Counter numEvents;
+    private transient Counter inputEvents;
```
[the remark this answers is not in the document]
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -80,33 +86,69 @@ public boolean apply(KV<Object, Object> message) {
...
 
     @Override
     public void init(Context context) {
       TranslatorContext translatorContext =
           ((SamzaSqlApplicationContext) context.getApplicationTaskContext()).getTranslatorContexts().get(queryId);
       this.msgConverter = translatorContext.getMsgConverter(streamName);
+      ContainerContext containerContext = context.getContainerContext();
+      metricsRegistry = containerContext.getContainerMetricsRegistry();
+      processingTime = new SamzaHistogram(metricsRegistry, logicalOpId, TranslatorConstants.PROCESSING_TIME_NAME);
+      inputEvents = metricsRegistry.newCounter(logicalOpId, TranslatorConstants.INPUT_EVENTS_NAME);
```
srinipunuru: Same as the output, If the queryInput events and Scan level inputEvents are going to be the same. I think it may be good idea to just remove the scan level input Events.

samza-sql/src/main/java/org/apache/samza/sql/translator/TranslatorConstants.java
```diff
@@ -0,0 +1,32 @@
...
+
+package org.apache.samza.sql.translator;
+
+public class TranslatorConstants {
+  public static final String PROCESSING_TIME_NAME = "processingTimeMs";
+  public static final String TOTAL_LATENCY_NAME = "totalLatencyMs";
+  public static final String QUERY_LATENCY_NAME = "queryLatencyMs";
+  public static final String QUEUEING_LATENCY_NAME = "queueingLatencyMs";
+  public static final String INPUT_EVENTS_NAME = "inputEvents";
+  public static final String FILTERED_EVENTS_NAME = "filteredOutEvents";
```
srinipunuru: We can just call it as filteredEvents

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
```diff
@@ -180,20 +259,21 @@ public RelNode visit(LogicalJoin join) {
...
-        String logicalOpId = "sql" + Integer.toString(queryId) + "_window" + Integer.toString(opId++);
+        String logicalOpId = String.format(TranslatorConstants.LOGOPID_TEMPLATE, queryId, "window", opId++);
         new LogicalAggregateTranslator(logicalOpId, sqlConfig.getMetadataTopicPrefix())
             .translate(aggregate, translatorContext);
         return node;
       }
     });
 
-    sendToOutputStream(outputSystemStream, streamAppDescriptor, translatorContext, node, queryId);
+    String logicalOpId = String.format(TranslatorConstants.LOGOPID_TEMPLATE, queryId, "sendTo", opId);
```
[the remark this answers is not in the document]
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/TranslatorConstants.java
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
```diff
@@ -67,29 +76,95 @@
...
+      queueingLatency = new SamzaHistogram(metricsRegistry, queryLogicalId, TranslatorConstants.QUEUEING_LATENCY_NAME);;
+      outputEvents = metricsRegistry.newCounter(queryLogicalId, TranslatorConstants.OUTPUT_EVENTS_NAME);
+      outputEvents.clear();
     }
 
     @Override
     public KV<Object, Object> apply(SamzaSqlRelMessage message) {
-      return this.samzaMsgConverter.convertToSamzaMessage(message);
+      Instant beginProcessing = Instant.now();
+      KV<Object, Object> retKV = this.samzaMsgConverter.convertToSamzaMessage(message);
```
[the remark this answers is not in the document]
shenodaguirguis: as discussed offline, it is not currently feasible to include sendTo() time. This is to be revisited later, tracked by [SAMZA-2031](https://issues.apache.org/jira/browse/SAMZA-2031) and commented in the code

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
```diff
@@ -67,29 +76,95 @@
...
     @Override
     public void init(Context context) {
       TranslatorContext translatorContext =
           ((SamzaSqlApplicationContext) context.getApplicationTaskContext()).getTranslatorContexts().get(queryId);
       this.samzaMsgConverter = translatorContext.getMsgConverter(outputTopic);
+      ContainerContext containerContext = context.getContainerContext();
+      metricsRegistry = containerContext.getContainerMetricsRegistry();
+      /* sendTo Metrics */
+      processingTime = new SamzaHistogram(metricsRegistry, sendToLogicalId, TranslatorConstants.TOTAL_LATENCY_NAME);;
+      inputEvents = metricsRegistry.newCounter(sendToLogicalId, TranslatorConstants.OUTPUT_EVENTS_NAME);
```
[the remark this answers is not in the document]
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
shenodaguirguis: done

PR #837, opened 2018-12-03
SAMZA-2019: add Is broadcast per stream config

sborya:


samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
```diff
@@ -53,9 +53,10 @@
   StreamEdge(StreamSpec streamSpec, boolean isIntermediate, boolean isBroadcast, Config config) {
     this.streamSpec = streamSpec;
     this.isIntermediate = isIntermediate;
-    this.isBroadcast = isBroadcast;
+    this.isBroadcast =
+          isBroadcast || config.getBoolean(String.format("streams.%s.samza.broadcast", streamSpec.getId()), false);;
```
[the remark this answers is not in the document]
sborya: yeah, I was working on it. Should be in now.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1180,6 +1180,13 @@ <h1>Samza Configuration Reference</h1>
                     </td>
                 </tr>
 
+                    <td class="property" id="streams-samza-broadcast">streams.<span class="stream">stream-id</span>.<br>samza.broadcast</td>
```
prateekm: I think this is the old config page. Should add it to the md file instead.

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
```diff
@@ -111,20 +113,21 @@ boolean isIntermediate() {
   }
 
   Config generateConfig() {
-    Map<String, String> config = new HashMap<>();
+    Map<String, String> newConfig = new HashMap<>();
```
prateekm: Revert rename?

sborya: 'config' override a member field 'config'. Although it is intentional, it makes the code confusing.

docs/learn/documentation/versioned/jobs/configuration-table.html
sborya: Oh, ok

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
prateekm: 'streamConfig' would be clearer in that case.

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
```diff
@@ -53,9 +53,11 @@
   StreamEdge(StreamSpec streamSpec, boolean isIntermediate, boolean isBroadcast, Config config) {
     this.streamSpec = streamSpec;
     this.isIntermediate = isIntermediate;
-    this.isBroadcast = isBroadcast;
+    // broadcast can be configured either by an operator or via the configs
+    this.isBroadcast =
+          isBroadcast || (config == null) ? false : new StreamConfig(config).getBroadcastEnabled(streamSpec.toSystemStream());
```
prateekm: Why would config be null here? I'd also prefer to extract variables for clarify instead of a ternary operator inside an or. It would also be clearer to make the caller do if possible. Otherwise the param isBroadcast doesn't mean what it says it means.

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
sborya: will change.

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
sborya: I was trying to address some tests, that pass null. I agree it is not a valid case, so I modified the tests. I will remove this condition in PR846

PR #838, opened 2018-12-03
Moving store test to  TestContainerStorageManager from TestSamzaContainer

rmatharu:
There was a test in TestSamzaContainer that needs to be moved to TestContainerStorageManager because the restore logic is moved there. 

Minor change in TestSamzaContainer and ContainerStorageManager


samza-core/src/test/scala/org/apache/samza/container/TestSamzaContainer.scala
```diff
@@ -152,18 +154,6 @@ class TestSamzaContainer extends AssertionsForJUnit with MockitoSugar {
     verify(this.runLoop).run()
   }
 
-  @Test
```
rmatharu: This test has now been moved to TestContainerStorageManager

PR #839, opened 2018-12-03
SAMZA-2021: Adding an API to rel converter to filter out system messages.

atoomula:


samza-sql/src/main/java/org/apache/samza/sql/interfaces/SamzaRelConverter.java
```diff
@@ -30,6 +30,15 @@
  * {@link SamzaRelConverter} for different systems.
  */
 public interface SamzaRelConverter {
+  /**
+   * Determine if the input samza message is a system message.
+   * @param message input samza message.
+   * @return true if the input message is system message.
+   */
+  default boolean isSystemMessage(KV<Object, Object> message) {
```
srinipunuru: Can you add a comment saying that this api will be removed soon, When the Creation of the descriptor becomes part of the resolver?

PR #845, opened 2018-12-05
SAMZA-2027: Supports environment variables for SqlExecutors. 

weiqingy:
1. User can now set environment variables for the shell itself and for any SqlExecutor, by executing the '`SET`' command in the shell or by using a configuration file. User can also choose to use any SqlExecutor they want in the same way. 

2. Refine the abstraction of environment variables. New interfaces and abstract classes are added so the shell itself and any SqlExecutor can now give a specification about what variables they support, what the possible values are, the default value, and whether they accept unknown variables or not. The shell and SqlExecutors now handle environment variables in a consistent way.

3. The `SET` command controls environment variables which affects corresponding behaviors of the Shell. Conforming to Unix philosophy, the usage and behavior of command `SET` is easy to expect. `SET` without parameters displays all the environment variables and their current values, no matter whether they have been explicitly set by user or configuration file or not, as they are actually always there affecting the Shell and is ready to be modified. “`SET <EnvName>`” displays the value of that specific environment variable. “`SET <EnvName=Value>`” sets the value of the environment variable. All the values of environment variables set by user via interactive set command are NOT saved, meaning they only take effect till the end of the session. To change an environment variable permanently, use the configuration file.
In implementation, the Shell itself has no knowledge of the Executor and handles those variables starting with “`shell.`” only. The Shell, however, asks the active Executor to handle other variables(e.g. “`samza.sql.output`”, “`samza.sql.relSchemaProvider.config.schemaDir`”, etc) by asking the Executor if it recognizes a certain environment variable, what the valid values of the variable are, and informs the Executor when the value of the variable is set.
## How was this patch tested?
Tested in Samza SQL shell.

### DIFF
--- a/samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
+++ b/samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
@@ -19,112 +19,168 @@
 
 package org.apache.samza.sql.client.cli;
 
-import java.io.IOException;
-import java.io.OutputStream;
-import java.io.PrintStream;
-import java.io.Writer;
-import java.util.ArrayList;
+import org.apache.samza.sql.client.interfaces.EnvironmentVariableHandler;
+import org.apache.samza.sql.client.interfaces.EnvironmentVariableSpecs;
+import org.apache.samza.sql.client.interfaces.SqlExecutor;
+import org.apache.samza.sql.client.util.CliException;
+import org.apache.samza.sql.client.util.CliUtil;
+import org.apache.samza.sql.client.util.Pair;
+
+import java.io.PrintWriter;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
- * CliEnvironment contains "environment variables" that configures the shell behavior.
+ * CliEnvironment handles "environment variables" that configures the shell behavior.
  */
-public class CliEnvironment {
-  private static final String debugEnvVar = "shell.debug";
-  private static PrintStream stdout = System.out;
-  private static PrintStream stderr = System.err;
-  private Boolean debug = false;
+class CliEnvironment {
+  private EnvironmentVariableHandler shellEnvHandler;
+  private EnvironmentVariableHandler executorEnvHandler;
+  private SqlExecutor executor;
+  private Map<String, String> delayedExecutorVars;
 
-  boolean isDebug() {
-    return debug;
-  }
+  // shell.executor is special and is specifically handled by CliEnvironment
+  private String activeExecutorClassName;
 
-  void setDebug(Boolean debug) {
-    this.debug = debug;
+  public CliEnvironment() {
+    shellEnvHandler = new CliShellEnvironmentVariableHandler();
   }
 
   /**
-   * @param var Environment variable
-   * @param val Value of the environment variable
+   * @param envName Environment variable name
+   * @param value   Value of the environment variable
    * @return 0 : succeed
-   * -1: invalid var
-   * -2: invalid val
+   * -1: invalid envName
+   * -2: invalid value
    */
-  int setEnvironmentVariable(String var, String val) {
-    switch (var.toLowerCase()) {
-      case debugEnvVar:
-        val = val.toLowerCase();
-        if (val.equals("true")) {
-          debug = true;
-          enableJavaSystemOutAndErr();
-        } else if (val.equals("false")) {
-          debug = false;
-          disableJavaSystemOutAndErr();
-        } else
-          return -2;
-        break;
-      default:
-        return -1;
+  public int setEnvironmentVariable(String envName, String value) {
+    envName = envName.toLowerCase();
+    if(envName.equals(CliConstants.CONFIG_EXECUTOR)) {
+      if(createShellExecutor(value)) {
+        activeExecutorClassName = value;
+        executorEnvHandler = executor.getEnvironmentVariableHandler();
+        return 0;
+      } else {
+        return -2;
+      }
     }
 
-    return 0;
+    EnvironmentVariableHandler handler = getAppropriateHandler(envName);
+    if(handler == null) {
+      // Shell doesn't recognize this variable. There's no executor handler yet. Save for future executor
+      if(delayedExecutorVars == null) {
+        delayedExecutorVars = new HashMap<>();
+      }
+      delayedExecutorVars.put(envName, value);
+      return 0;
+    }
+
+    return handler.setEnvironmentVariable(envName, value);
   }
 
-  // TODO: Separate the values out of the logic part
-  List<String> getPossibleValues(String var) {
-    List<String> vals = new ArrayList<>();
-    switch (var.toLowerCase()) {
-      case debugEnvVar:
-        vals.add("true");
-        vals.add("false");
-        return vals;
-      default:
-        return null;
+  public String getEnvironmentVariable(String envName) {
+    if(envName.equalsIgnoreCase(CliConstants.CONFIG_EXECUTOR)) {
+      return activeExecutorClassName;
+    }
+
+    EnvironmentVariableHandler handler = getAppropriateHandler(envName);
+    if(handler == null)
+      return null;
+
+    return handler.getEnvironmentVariable(envName);
+  }
+
+  public String[] getPossibleValues(String envName)
+  {
+    EnvironmentVariableHandler handler = getAppropriateHandler(envName);
+    if(handler == null)
+      return null;
+
+    EnvironmentVariableSpecs.Spec spec = handler.getSpecs().getSpec(envName);
+    return spec == null ? null : spec.getPossibleValues();
+  }
+
+  public void printAll(PrintWriter writer) {
+    printVariable(writer, CliConstants.CONFIG_EXECUTOR, activeExecutorClassName);
+    printAll(writer, shellEnvHandler);
+    if(executorEnvHandler != null) {
+      printAll(writer, executorEnvHandler);
     }
   }
 
-  void printAll(Writer writer) throws IOException {
-    writer.write(debugEnvVar);
-    writer.write('=');
-    writer.write(debug.toString());
-    writer.write('\n');
+  /**
+   * Gives CliEnvironment a chance to apply settings, especially during initialization, things like
+   * making default values take effect
+   */
+  public void finishInitialization() {
+    if(executor == null) {
+      if(createShellExecutor(CliConstants.DEFAULT_EXECUTOR_CLASS)) {
+        activeExecutorClassName = CliConstants.DEFAULT_EXECUTOR_CLASS;
+        executorEnvHandler = executor.getEnvironmentVariableHandler();
+        if(delayedExecutorVars != null) {
+          for (Map.Entry<String, String> entry : delayedExecutorVars.entrySet()) {
+            setEnvironmentVariable(entry.getKey(), entry.getValue());
+          }
+          delayedExecutorVars = null;
+        }
+      } else {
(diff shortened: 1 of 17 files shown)


weiqingy:
@srinipunuru @atoomula Could you please help to review this PR? Thanks.

weiqingy:
@atoomula Yes we could support SET in the samza sql file in the next PR. Just one logic we need to clarify: shall the configs in the sql file be kept effective even after a samza.sql is executed, or shall the previous shell settings be restored after a sql file being executed?

atoomula:
I think this warrants a one pager describing what the behavior of set in Samza sql file would be and its interaction with the regular config properties.


On Jan 3, 2019, at 10:16 AM, Weiqing Yang <notifications@github.com<mailto:notifications@github.com>> wrote:


@atoomula<https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fatoomula&data=02%7C01%7Catoomula%40linkedin.com%7C16a2dafbb0dd4977845908d671a79f22%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C636821362081764723&sdata=0VhNpnSpXS%2FDbb50tsN5zBl7bfeLrgdtgfPAogY0o9k%3D&reserved=0> Yes we could support SET in the samza sql file in the next PR. Just one logic we need to clarify: shall the configs in the sql file be kept effective even after a samza.sql is executed, or shall the previous shell settings be restored after a sql file being executed?

weiqingy:
Per the discussion offline with @atoomula ,the exact behavior of supporting SET command is yet to be defined and we've agreed not to put it in this PR but leave it as a future work instead.

weiqingy:
@atoomula @srinipunuru Could you help to review this PR?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
```diff
@@ -19,112 +19,168 @@
...
+      executor = (SqlExecutor) ctor.newInstance();
+    } catch (ClassNotFoundException | NoSuchMethodException
+            | IllegalAccessException | InstantiationException | InvocationTargetException e) {
+      return false;
     }
+    return true;
+  }
 
-    public void write(byte[] b, int off, int len) {
+  private EnvironmentVariableHandler getAppropriateHandler(String envName) {
```
srinipunuru: Can you add comments for all these methods?

srinipunuru: it looks like envName is the environmentVariableName right? But it gives an impression that it is environmentName. Can you rename this variable across the usage appropriately?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
```diff
@@ -19,112 +19,168 @@
...
 
-    public void write(byte[] b) {
+  private boolean createShellExecutor(String executorClassName) {
+    try {
+      Class<?> clazz = Class.forName(executorClassName);
+      Constructor<?> ctor = clazz.getConstructor();
+      executor = (SqlExecutor) ctor.newInstance();
+    } catch (ClassNotFoundException | NoSuchMethodException
+            | IllegalAccessException | InstantiationException | InvocationTargetException e) {
+      return false;
```
srinipunuru: Whenever an exception is swallowed, We should log it otherwise it will be hard to debug. Can you please log the exception that is swallowed here?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
```diff
@@ -19,112 +19,168 @@
...
 
-  private class NullOutputStream extends OutputStream {
-    public void close() {
+  /**
+   * Gives CliEnvironment a chance to apply settings, especially during initialization, things like
+   * making default values take effect
+   */
+  public void finishInitialization() {
+    if(executor == null) {
+      if(createShellExecutor(CliConstants.DEFAULT_EXECUTOR_CLASS)) {
```
srinipunuru: rather than have the method return boolean do you think it is a good idea to have the exception thrown and logged below?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliShellEnvironmentVariableHandler.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.io.OutputStream;
+import java.io.PrintStream;
+
+
+public class CliShellEnvironmentVariableHandler extends EnvironmentVariableHandlerImpl {
+  private static final String SHELL_DEBUG = "shell.debug";
+
+  public CliShellEnvironmentVariableHandler() {
+    specs.put(SHELL_DEBUG, new String[] {"true", "false"}, "false");
```
srinipunuru: Can we make specs immutable and have it initialized only in the constructor?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/EnvironmentVariableHandlerImpl.java
```diff
@@ -0,0 +1,107 @@
...
+    return -2;
+  }
+
+  @Override
+  public String getEnvironmentVariable(String envName) {
+    return envVars.get(envName);
+  }
+
+  @Override
+  public List<Pair<String, String>> getAllEnvironmentVariables() {
```
srinipunuru: Does it return only the environment variables that are set or the ones that have default values as well?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/EnvironmentVariableHandlerImpl.java
```diff
@@ -0,0 +1,107 @@
...
+  protected boolean processEnvironmentVariable(String envName, String value) {
+    return true;
+  }
+
+  /**
+   * Subclasses shall override this method to if they want to accept environment variables
+   * with unknown names
+   * @return true if environment variables with unknown names are acceptable. This method returns false by default.
+   */
+  protected boolean isAcceptUnknowName() {
```
srinipunuru: I am not sure i understand when this will be set to true. Can you give me a scenarios when this should be true?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/EnvironmentVariableHandlerImpl.java
```diff
@@ -0,0 +1,107 @@
...
+    return specs;
+  }
+
+  /**
+   * Subclasses shall override this method to process their settings
+   * @param envName name of the environment variable
+   * @param value value of the environment variable
+   * @return succeed or not
+   */
+  protected boolean processEnvironmentVariable(String envName, String value) {
```
srinipunuru: Can we use exception handling for success or failure that ways it's more consistent with other parts of the code?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/EnvironmentVariableSpecs.java
```diff
@@ -0,0 +1,80 @@
...
+   * @param defaultValue default value of the environment variable
+   */
+  public void put(String envName, String[] possibleValues, String defaultValue) {
+    if(CliUtil.isNullOrEmpty(envName) || CliUtil.isNullOrEmpty(defaultValue))
+      throw new IllegalArgumentException();
+
+    specMap.put(envName, new Spec(possibleValues, defaultValue));
+  }
+
+  public Spec getSpec(String envName) {
```
srinipunuru: Can you add method documentation, Also for all other methods in this PR?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
weiqingy: Sure.

weiqingy: Sure.

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
weiqingy: Of course. Good point.

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/EnvironmentVariableHandlerImpl.java
weiqingy: During initialization, after having read the configuration file, all environment variables in the specification that have not been set by the configuration file are SET to the default value. So the answer is both.

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/EnvironmentVariableSpecs.java
weiqingy: Sure.

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/EnvironmentVariableHandlerImpl.java
weiqingy: It's actually a temporary workaround for SamzaExecutor. It has so many configuration entries I am not yet able to go through them and input them as a specification, like ""serializers.registry.string.class", "producer.bootstrap.servers", etc.

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/EnvironmentVariableHandlerImpl.java
weiqingy: At the moment Samza-Sql-Shell prefers return values for expected error or must-handle errors. I know it's controversial but as I mentioned in a previous comment, we need to change the way of reporting errors entirely if we decide to use exceptions. Can we keep this issue for a while and see what we can do as a future work?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/EnvironmentVariableHandlerImpl.java
srinipunuru: Can you add this as a comment in the code.

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
srinipunuru: Good point @weiqingy  As we discussed offline, Let's do regular exceptions for all the internal methods and in the outermost layer before the information is returned to the user, You can handle these exceptions and transform them into user understandable form.

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/EnvironmentVariableHandlerImpl.java
srinipunuru: I think our offline discussion should clarify this as well :).

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/EnvironmentVariableHandlerImpl.java
weiqingy: Sure.

weiqingy:
Thanks for the review, @srinipunuru . I've updated the PR ([5f7cb40](https://github.com/apache/samza/pull/845/commits/5f7cb40aa1674372ac06f326ce496c4c1aad3ee3)) to handle the exceptions.

PR #848, opened 2018-12-06
SAMZA-2028 Samza-SQL Diagnostics: add metrics to Join and Aggregate operators

shenodaguirguis:
by adding metrics to Join and Aggregate, it concludes the first phase (adding metrics) of Samza-SQL Diagnostics.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
@@ -36,6 +36,12 @@ public class SamzaSqlRelMsgMetadata implements Serializable {
   public boolean isNewInputMessage = true;
 
   /**
+   *
+   */
+  public String operatorBeginProcessingInstant = null;
+
+
+  /**
    * The timestamp of when the events actually happened
    * set by and copied from the event source
    * TODO: copy eventTime through from source to RelMessage
--- a/samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
@@ -19,11 +19,11 @@
 
 package org.apache.samza.sql.translator;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
 
-import java.util.Objects;
 import org.apache.calcite.adapter.enumerable.EnumerableTableScan;
 import org.apache.calcite.plan.RelOptUtil;
 import org.apache.calcite.rel.RelNode;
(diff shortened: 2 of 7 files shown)


samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
```diff
@@ -35,6 +35,12 @@
    */
   public boolean isNewInputMessage = true;
 
+  /**
+   *
```
weiqingy: Missing comments?

shenodaguirguis: added

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -154,6 +161,7 @@ void translate(final LogicalJoin join, final TranslatorContext translatorContext
     // the names from the stream as the lookup needs to be done based on what is stored in the local table.
     return
         inputStream
+            .map(inputMetricsMF)
```
atoomula: For local table, we probably need more metrics. This is the stream side of the metrics. We also want to have some metrics for the table where the source stream is repartitioned and sent to table. We anyways do not support it for now. Could you file a jira for it ?

samza-sql/src/main/java/org/apache/samza/sql/translator/TranslatorOutputMetricsMapFunction.java
```diff
@@ -0,0 +1,71 @@
...
+   * update metrics given a message
+   * @param message  the input message
+   * @return the same message
+   */
+  @Override
+  public SamzaSqlRelMessage apply(SamzaSqlRelMessage message) {
+    Instant endProcessing = Instant.now();
+    Instant beginProcessing = Instant.parse(message.getSamzaSqlRelMsgMetadata().operatorBeginProcessingInstant);
+    outputEvents.inc();
+    processingTime.update(Duration.between(beginProcessing, endProcessing).toMillis());
```
atoomula: Is processing time for remote table join == latency for lookup ?

samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
shenodaguirguis: created JIRA: SAMZA-2029

samza-sql/src/main/java/org/apache/samza/sql/translator/TranslatorOutputMetricsMapFunction.java
shenodaguirguis: Yes.currently, for both remote and local table join, the processingTime = total join operator latency, whatever processing that includes. So, for remote table, since processing is simply a remote lookup operation then processingTime

PR #855, opened 2018-12-11
Minor: Log full thread stacks in thread dumps.

prateekm:


samza-core/src/main/java/org/apache/samza/util/ThreadUtil.java
```diff
@@ -0,0 +1,123 @@
...
+
+  public static void logThreadDump(String message) {
+    try {
+      ThreadInfo[] threadInfo = THREAD_MX_BEAN.dumpAllThreads(true, true);
+      StringBuilder sb = new StringBuilder();
+      sb.append(message).append("\n");
+      for (ThreadInfo ti: threadInfo) {
+        sb.append(toString(ti)).append("\n");
+      }
+      System.out.println(sb.toString());
```
cameronlee314: It looks like the previous impl used log4j to log. Did you intend to just print to stdout now?

samza-core/src/main/java/org/apache/samza/util/ThreadUtil.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+    if (info.isSuspended()) {
+      sb.append(" (suspended)");
+    }
+    if (info.isInNative()) {
+      sb.append(" (in native)");
+    }
+    sb.append('\n');
+    int i = 0;
+    for (; i < info.getStackTrace().length; i++) {
```
cameronlee314: Is there any risk in having too large of a log message? Such as if you try to send this to StreamAppender?

samza-core/src/main/java/org/apache/samza/util/ThreadUtil.java
prateekm: Oh, this was for testing, will fix.

samza-core/src/main/java/org/apache/samza/util/ThreadUtil.java
prateekm: It's possible. Within LinkedIn we use the open source li-apache-kafka-clients with large message support. This might be an issue for a regular Kafka producer. Any suggestions for a workaround? We can print to stdout instead if you prefer that.

cameronlee314: One thing that comes to mind is to use a large but finite MAX_FRAMES. But doing that estimate will be kind of arbitrary. Using stdout would work, but it's probably nice if we could do the same logging as everything else, since then users won't have to know to look in a different place.
Although, maybe we don't have to worry about this. I think if StreamAppender send fails, the app still will end up writing to the local log, and since drop.producer.errors is true, the producer will just reset itself. We would miss the log on kafka, but maybe that is ok. Is that understanding correct?

prateekm: That's right. `drop.producer.errors` is enabled for StreamAppender, so this (and possibly other log messages in the batch) will get dropped if the batch is too large. It will still write to local.

samza-core/src/main/java/org/apache/samza/util/ThreadUtil.java
```diff
@@ -0,0 +1,123 @@
...
+      LOGGER.info(sb.toString());
+    } catch (Exception e) {
+      LOGGER.error("Could not get and log a thread dump.", e);
+    }
+  }
+
+  /**
+   * Copy of ThreadInfo#toString() without the hardcoded MAX_FRAMES = 8 restriction on thread stack depth.
+   *
+   * Returns a string representation of this thread info.
```
vjagadish1989: trivial: fix line width

PR #857, opened 2018-12-12
SAMZA-2041: add hdfs and kinesis descriptor

lhaiesp:


lhaiesp:
@xinyuiscool  @atoomula   can you take a look?

prateekm:
@dxichen Can you take a look as well?

samza-aws/src/main/java/org/apache/samza/system/kinesis/descriptors/KinesisInputDescriptor.java
```diff
@@ -0,0 +1,122 @@
...
+  public KinesisInputDescriptor<StreamMessageType> withKCLConfig(Map<String, String> kclConfig) {
+    this.kclConfig = kclConfig;
+    return this;
+  }
+
+  @Override
+  public Map<String, String> toConfig() {
+    final Map<String, String> config = new HashMap<>(super.toConfig());
+
+    final String systemName = getSystemName();
```
vjagadish1989: nit: The rest of the code-base does not use `final` for local variables since it's a bit boiler-platey.

samza-aws/src/main/java/org/apache/samza/system/kinesis/descriptors/KinesisInputDescriptor.java
```diff
@@ -0,0 +1,122 @@
...
+
+
+  /**
+   * Constructs an {@link InputDescriptor} instance.
+   *
+   * @param streamId id of the stream
+   * @param valueSerde serde the values in the messages in the stream
+   * @param systemDescriptor system descriptor this stream descriptor was obtained from
+   */
+  KinesisInputDescriptor(String streamId, Serde valueSerde, SystemDescriptor systemDescriptor) {
```
vjagadish1989: Should you make the serde typed for KinesisInputDescriptor?

samza-hdfs/src/main/java/org/apache/samza/system/hdfs/descriptors/HdfsInputDescriptor.java
```diff
@@ -0,0 +1,48 @@
...
+ * A {@link HdfsInputDescriptor} can be used for specifying Samza and HDFS specific properties of HDFS
+ * input streams.
+ * <p>
+ * Use {@link HdfsSystemDescriptor#getInputDescriptor} to obtain an instance of this descriptor.
+ * <p>
+ * Stream properties provided in configuration override corresponding properties specified using a descriptor.
+ *
+ */
+public class HdfsInputDescriptor
+    extends InputDescriptor<Object, HdfsInputDescriptor> {
```
vjagadish1989: Any chance we can expose a "typed" API instead of InputDescriptor\<Object\>?

samza-aws/src/main/java/org/apache/samza/system/kinesis/descriptors/KinesisInputDescriptor.java
```diff
@@ -0,0 +1,122 @@
...
+    super(streamId, KVSerde.of(new NoOpSerde<>(), valueSerde), systemDescriptor, null);
+  }
+
+  /**
+   * Kinesis region for the system stream.
+   * @param region Kinesis region
+   * @return this input descriptor
+   */
+  public KinesisInputDescriptor<StreamMessageType> withRegion(String region) {
+    this.region = Optional.of(region);
```
dxichen: Nit: For defensive programming, I would prefer `String.stripToNull(region)`; Please update for similar invocations.

docs/learn/documentation/versioned/connectors/kinesis.md
```diff
@@ -66,40 +60,63 @@ job.systemstreampartition.grouper.factory=org.apache.samza.container.grouper.str
...
+
+final KinesisInputDescriptor<KV<String, byte[]>> kid = 
+    ksd.getInputDescriptor("STREAM-NAME", new NoOpSerde<byte[]>())
+          .withRegion("STREAM-REGION")
+          .withAccessKey("YOUR-ACCESS_KEY")
+          .withSecretKey("YOUR-SECRET-KEY")
+          .withKCLConfig(kclConfig);
 {% endhighlight %}
 
 As an example, the below configuration is equivalent to invoking `kclClient#WithTableName(myTable)` on the KCL instance.
```
dxichen: Is there a way that we may truncate these 2 code examples since they only differ by one line? IE add`kclConfig.put("TableName", "myTable");` to the first sample.

samza-hdfs/src/main/java/org/apache/samza/system/hdfs/descriptors/HdfsSystemDescriptor.java
```diff
@@ -0,0 +1,254 @@
...
+  private static final String FACTORY_CLASS_NAME = HdfsSystemFactory.class.getName();
+
+  private Optional<String> datePathFormat = Optional.empty();
+  private Optional<String> outputBaseDir = Optional.empty();
+  private Optional<Long> writeBatchSizeBytes = Optional.empty();
+  private Optional<Long> writeBatchSizeRecords = Optional.empty();
+  private Optional<String> writeCompressionType = Optional.empty();
+  private Optional<String> writerClass = Optional.empty();
+
+  private Optional<Long> consumerBufferCapacity = Optional.empty();
```
dxichen: I would prefer if we add all the consumer/producer configs in the the input and output descriptor respectively instead of the system descriptor.

samza-aws/src/main/java/org/apache/samza/system/kinesis/descriptors/KinesisInputDescriptor.java
lhaiesp: done

samza-hdfs/src/main/java/org/apache/samza/system/hdfs/descriptors/HdfsInputDescriptor.java
lhaiesp: Unfortunately, HDFS system is not able to leverage the serde concept in samza. Essentially the data type is determined by file type itself (csv, avro, orc, etc.). Users today need to config different reader to read different types of files. We don't just get bytes and deserialize them.

It may be possible to support a "typed" API after more thoughts and refactor. But it would be beyond the scope of this change

samza-aws/src/main/java/org/apache/samza/system/kinesis/descriptors/KinesisInputDescriptor.java
lhaiesp: done

docs/learn/documentation/versioned/connectors/kinesis.md
lhaiesp: It's not completely straightforward to do so. But I have tried to reduce the duplicate codes to make it less verbose.

samza-hdfs/src/main/java/org/apache/samza/system/hdfs/descriptors/HdfsSystemDescriptor.java
lhaiesp: This is another unfortunate unresolved issue in HDFS. Where all the configs are per system not per stream. Once we move to per stream config then it definitely makes sense to place them respectively in input and output descriptor.

samza-aws/src/main/java/org/apache/samza/system/kinesis/descriptors/KinesisInputDescriptor.java
xinyuiscool: @vjagadish1989 : using final is good habit and should be encouraged in samza code base. I don't think your suggestion makes sense here.

lhaiesp: Style and good practice discussion will never end. We should talk more about these in team meeting and agree upon things together.  I will just go for consistency for now

lhaiesp:
@dxichen  @vjagadish1989    addressed or responded to all your feedback. let me know if you have more comments

PR #859, opened 2018-12-12
SAMZA-2042: Add support for inspecting/modifying checkpoints before they're written.

prateekm:


prateekm:
cc. @sborya @vjagadish1989 for review.

prateekm:
This feature can be used, for example, to adjust the to-be-checkpointed offset to a safe offset when using li-apache-kafka-consumer with large message support.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -231,24 +231,64 @@ class OffsetManager(
...
     }
   }
 
+  /**
+    * Call the beforeCheckpoint() method on [[CheckpointListener]] SystemConsumers to give them a chance to inspect and
+    * potentially modify the offsets about to be checkpointed.
+    */
+  def getModifiedOffsets(taskStartingOffsets: Map[SystemStreamPartition, String],
+      taskLastProcessedOffsets: Map[SystemStreamPartition, String]): HashMap[SystemStreamPartition, String] = {
+    val modifiedOffsets = new HashMap[SystemStreamPartition, String]()
```
vjagadish1989: slightly cleaner to do : `new HashMap(taskLastProcessedOffsets.asJava)`

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -231,24 +231,64 @@ class OffsetManager(
...
+
+        // if we haven't processed any messages from the system since we started, we don't need to call
+        // the checkpoint listener. also, in case of Kafka, getting the safe offset is only possible after
+        // first poll() is successful. check if we need to adjust the offset for any of the ssp in the system.
+        val needModifiedOffsets = systemSSPLastProcessedOffsets.exists { case (ssp, sspLastProcessedOffset) =>
+          // starting offsets are checkpointed offsets + 1 (see 'loadStartingOffsets'),
+          // while on container start last processed offsets are equal to checkpointed offsets.
+          // subtract one from starting offsets to compare them with last processed offsets.
+          val adjustedSSPStartingOffset = taskStartingOffsets.get(ssp) match {
+            case Some(offset) if offset != null => (offset.toLong - 1).toString
```
vjagadish1989: There's a hidden assumption here on contiguous offsets - since, we're by-passing the offset comparator. maybe that's ok?

prateekm: That's true, however there's no SystemAdmin#getOffsetsBefore() API that we can use here.

This is ported from existing samza-li code for using li-apache-kafka-consumer with safe offsets. I think this was a workaround for the KafkaConsumer specific behavior that poll() must be called on the consumer before requesting a safe offset.

Now that you mention it, I think the bigger issue is that this is also assuming that the offset is a long, which is definitely not true in general. I can check if we can handle the KafkaConsumer specific issue by catching whatever exception safeOffset returns.

samza-api/src/main/java/org/apache/samza/checkpoint/CheckpointListener.java
```diff
@@ -22,11 +22,28 @@
...
 public interface CheckpointListener {
-  void onCheckpoint(Map<SystemStreamPartition, String> offsets);
+  /**
+   * Called before writing the checkpoint for each Task in the Container processing SSPs on this system.
+   * Implementations may modify the Task offsets to be checkpointed, or return the provided offsets as-is.
+   *
+   * @param offsets offsets about to be checkpointed.
+   * @return potentially modified offsets to be checkpointed.
+   */
+  default Map<SystemStreamPartition, String> beforeCheckpoint(Map<SystemStreamPartition, String> offsets) {
```
shanthoosh: Please correct me if i'm wrong here.

1. Why do we need this API? When does the offsets tracked by the TaskInstance/RunLoop  for a stream has to be invalidated/altered by an external listener API?  

2. Would it make sense to add an API in systemAdmin like getCommitableOffsets/getValidOffsets which would return the valid offsets based upon the provided input offsets that we can then commit. 

In samza, checkpoint listeners are the system consumers which implement the CheckpointListener interface in addition to SystemConsumer interface. The problem with adding this API in checkpointListener is that every systemconsumer now has to be aware that it had to implement CheckpointListener to have an option to manipulate checkpoints before commit.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
shanthoosh: Please correct me if i'm wrong here.

Does subtracting -1 from the offset is a reasonable way to calculate the previous offset for any input stream?
For instance, in case of brooklin system, the  offsets are serialized json values. Casting the brooklin offset to long and subtracting 1 from the offset might not be a good idea.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -231,24 +231,64 @@ class OffsetManager(
...
+            case _ => ""
+          }
+          !adjustedSSPStartingOffset.equals(sspLastProcessedOffset)
+        }
+
+        if (needModifiedOffsets) {
```
shanthoosh: Please correct me if i'm wrong here.

This is more like a optimization right, why can't we bluntly call the listener.beforeCheckpoint without the needModifiedOffsets check. For most of the I/0 systems, checkpoint listener implementation will simply be a passthrough filter. For kafka system alone, this optimization might save few kafka API calls before we commit the first checkpoint.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
prateekm: Yes, see discussion with Jagadish above.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
prateekm: See discussion with Jagadish above.

samza-api/src/main/java/org/apache/samza/checkpoint/CheckpointListener.java
prateekm: > Why do we need this API?
See earlier discussion. This is a port for the LI internal logic for getting safe offset from KafkaConsumer for large message support.

Checkpoint listener interface already exists, which gives consumers a callback after the checkpoint has been written. We need a new callback before the checkpoint is written with the ability to modify it. It makes sense to me to put this callback int the CheckpointListener as well. WDYT?

SystemConsumer is typically the one that returns/mints the offsets for incoming messages, so it's seems reasonable to give the checkpoint callback to the SystemConsumer, instead of proxying to the same SystemConsumer through the SystemAdmin. WDYT?

samza-api/src/main/java/org/apache/samza/checkpoint/CheckpointListener.java
```diff
@@ -22,11 +22,28 @@
...
+    return offsets;
+  }
+
+  /**
+   * Called after writing the checkpoint for each Task in the Container processing SSPs on this system.
+   *
+   * @param offsets offsets that were checkpointed.
+   */
+  default void onCheckpoint(Map<SystemStreamPartition, String> offsets) { }
```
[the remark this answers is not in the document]
prateekm: Sure, will do.

samza-api/src/main/java/org/apache/samza/checkpoint/CheckpointListener.java
shanthoosh: Makes sense. Thanks for the explanation.

PR #860, opened 2018-12-13
SAMZA-1985: Startpoint and StartpointManager implementation.

dnishimura:
This is the first PR for [SEP-18](https://cwiki.apache.org/confluence/display/SAMZA/SEP-18%3A+Startpoints+-+Manipulating+Starting+Offsets+for+Input+Streams). Please see updated SEP-18 for details.
This PR implements the StartpointManager and Startpoint model and the initial integration with the OffsetManager. The OffsetManager manages the deletion of Startpoints when the initial checkpoint commits happen per task after start-up.

The immediate follow-ons to this PR are:
1. Have the various `JobCoordinators` to re-map the Startpoints appropriately to each task by utilizing the `StartpointManager#groupStartpointsPerTask(SystemStreamPartition, JobModel)` method implemented in this PR. SEP-18 describes this in more detail.
2. Implement `StartpointConsumerVisitor` for each of the provided `SystemConsumer`s.


dnishimura:
@cameronlee314 @prateekm @shanthoosh please take a look when you have some cycles.

samza-api/src/main/java/org/apache/samza/startpoint/PositionType.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.apache.samza.startpoint;
+
+/**
+ * Defines how a position in the {@link Startpoint} is interpreted.
+ */
+public enum PositionType {
+  SPECIFIC_OFFSET,
+  TIMESTAMP,
+  EARLIEST,
```
vjagadish1989: how does earliest differ from bootstrap? also maybe document individual fields?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -0,0 +1,273 @@
...
+   * @param startpoint Reference to a Startpoint object.
+   */
+  public void writeStartpoint(SystemStreamPartition ssp, Startpoint startpoint) {
+    Preconditions.checkNotNull(metadataStore, "Underlying metadata store not available");
+    Preconditions.checkNotNull(ssp, "SystemStreamPartition cannot be null");
+    Preconditions.checkNotNull(startpoint, "Startpoint cannot be null");
+
+    try {
+      metadataStore.put(toStoreKey(ssp),
+          startpointSerde.toBytes(startpoint.copyWithStoredAt(Instant.now().toEpochMilli())));
```
vjagadish1989: Consider always minting the timestamp on creation instead of cloning object later?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -0,0 +1,273 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * The StartpointManager reads and writes {@link Startpoint} to the {@link MetadataStore} defined by
+ * the configuration task.startpoint.metadata.store.factory.
+ *
+ * Startpoints are keyed in the MetadataStore by two different formats:
+ * 1) Only by {@link SystemStreamPartition}
+ * 2) A combination of {@link SystemStreamPartition} and {@link TaskName}
```
vjagadish1989: Do we need a notion of "TaskName" in the `StartpointManager`? For eg: do we have cases to support per-partition , per-task startpoints for broad-cast streams?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -0,0 +1,273 @@
...
+   * For {@link Startpoint}s keyed only by {@link SystemStreamPartition}, this method remaps the Startpoints for the specified
+   * SystemStreamPartition to SystemStreamPartition+{@link TaskName} for all tasks provided by the {@link SystemStreamPartitionGrouper}
+   * This method is not atomic or thread-safe. The intent is for the Samza Processor's coordinator to use this
+   * method to assign the Startpoints to the appropriate tasks.
+   * @param ssp The {@link SystemStreamPartition} that a {@link Startpoint} is keyed to.
+   * @param grouper The {@link SystemStreamPartitionGrouper} is used to determine what the task names are.
+   * @return The list of {@link TaskName}s
+   */
+  public Set<TaskName> groupStartpointsPerTask(SystemStreamPartition ssp, SystemStreamPartitionGrouper grouper) {
+    Preconditions.checkNotNull(metadataStore, "Underlying metadata store not available");
```
vjagadish1989: Can you explain why we require this Read-Modify-Write operation?

samza-api/src/main/java/org/apache/samza/startpoint/PositionType.java
dnishimura: Change-capture connectors treat earliest and bootstrap types of offsets differently. 
I'll document each of the enum values as suggested.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: That's correct, this is mainly in support for broadcast streams, but also custom SSP groupers may assign a partition to multiple tasks as well. In addition to that, when a Samza processor starts, the SystemConsumer registers the starting offsets per task instance. The StartpointManager has a mechanism [`groupStartpointsPerTask(SystemStreamPartition, SystemStreamPartitionGrouper)`] to copy Startpoints from an SSP to SSP+TaskName. Hence, the SystemConsumers do not need to be aware of any SSP=>TaskName mapping logic and simplifies things for the SystemConsumer.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: My thinking was to set the creation timestamp as close to the time it is stored in the metadata store, but it's probably not necessary to be that precise. I'll set the timestamp upon construction.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: Please see my response to your comment about TaskName awareness above. Also SEP-18 document provides more details. Any suggestions to convey the reasoning in the javadoc better?

samza-api/src/main/java/org/apache/samza/startpoint/Startpoint.java
```diff
@@ -0,0 +1,134 @@
...
+   */
+  public String getPosition() {
+    return position;
+  }
+
+  /**
+   * The timestamp when this {@link Startpoint} was written to the storage layer.
+   * @return a timestamp in epoch milliseconds.
+   */
+  public Long getStoredAt() {
```
shanthoosh: Would be better to rename this to `getCreationTimeStamp` or `getCreatedTimeStamp`.

samza-api/src/main/java/org/apache/samza/startpoint/PositionType.java
```diff
@@ -0,0 +1,30 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.startpoint;
+
+/**
+ * Defines how a position in the {@link Startpoint} is interpreted.
+ */
+public enum PositionType {
+  SPECIFIC_OFFSET,
```
shanthoosh: Would be better to document here in comments the semantic meaning of each one of these options and in which kind of streams they're applicable.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -0,0 +1,273 @@
...
+ *
+ * Startpoints are keyed in the MetadataStore by two different formats:
+ * 1) Only by {@link SystemStreamPartition}
+ * 2) A combination of {@link SystemStreamPartition} and {@link TaskName}
+ *
+ * The intention for the StartpointManager is to maintain a strong contract between the caller
+ * and how Startpoints are stored in the underlying MetadataStore.
+ */
+@InterfaceStability.Unstable
+public class StartpointManager {
```
shanthoosh: Since this is not in public interface would be better to remove the interface stabliity annotation.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -0,0 +1,273 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(StartpointManager.class);
+  private static final String NAMESPACE = "samza-startpoint-v1";
+
+  private final StartpointSerde startpointSerde = new StartpointSerde();
+
+  private MetadataStore metadataStore;
```
shanthoosh: May be mark this field as final as well?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -0,0 +1,273 @@
...
+    deleteStartpoint(ssp);
+
+    return ImmutableSet.copyOf(group.keySet());
+  }
+
+  /**
+   * Relinquish resources held by the underlying {@link MetadataStore}
+   */
+  public void stop() {
+    if (Objects.nonNull(metadataStore)) {
```
shanthoosh: 1. Do we need to make stop idempotent?
2. Why do we need to set metadatastore as null here. Would be better to use a different boolean variable here to make the stop idempotent.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -0,0 +1,273 @@
...
+  public void writeStartpoint(SystemStreamPartition ssp, Startpoint startpoint) {
+    Preconditions.checkNotNull(metadataStore, "Underlying metadata store not available");
+    Preconditions.checkNotNull(ssp, "SystemStreamPartition cannot be null");
+    Preconditions.checkNotNull(startpoint, "Startpoint cannot be null");
+
+    try {
+      metadataStore.put(toStoreKey(ssp),
+          startpointSerde.toBytes(startpoint.copyWithStoredAt(Instant.now().toEpochMilli())));
+    } catch (Exception ex) {
+      throw new SamzaException(String.format(
```
shanthoosh: Why do we need to catch the exception here wrap the exception and throw it back. What is the value-add we get compared to  throwing up the base exception back  to the caller?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -0,0 +1,273 @@
...
+    Preconditions.checkNotNull(ssp, "SystemStreamPartition cannot be null");
+    Preconditions.checkNotNull(grouper, "SystemStreamPartitionGrouper cannot be null");
+
+    Startpoint startpoint = readStartpoint(ssp);
+    HashMap<TaskName, Startpoint> result = new HashMap<>();
+
+    LOG.info("Grouping Startpoint keyed on SSP: {} to tasks determined by the grouper: {}", ssp, grouper.getClass().getCanonicalName());
+
+    // Re-map startpoints from SSP-only keys to SSP+TaskName keys
+    Map<TaskName, Set<SystemStreamPartition>> group = grouper.group(Sets.newHashSet(ssp));
```
shanthoosh: 1. Why do we need the grouper here to generate the task name, because kafka checkpoint  [key](https://github.com/apache/samza/blob/master/samza-kafka/src/main/java/org/apache/samza/checkpoint/kafka/KafkaCheckpointLogKey.java) needs it?
2. What if the grouper is changed by the user after publishing the startpoint and before the subsequent deployment.

samza-core/src/test/java/org/apache/samza/startpoint/TestStartpointManager.java
```diff
@@ -0,0 +1,292 @@
...
+import org.apache.samza.container.grouper.stream.SystemStreamPartitionGrouper;
+import org.apache.samza.system.SystemStreamPartition;
+import org.apache.samza.util.NoOpMetricsRegistry;
+import org.junit.Assert;
+import org.junit.Test;
+
+
+public class TestStartpointManager {
+  @Test
+  public void TestDefaultMetadataStore() {
```
shanthoosh: Would be better to add a integration test by bringing a embedded broker and testing the entire flow from publish startpoints, bringing up container and seeing the update value.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -284,6 +319,14 @@ class OffsetManager(
     } else {
       debug("Skipping checkpoint manager shutdown because no checkpoint manager is defined.")
     }
+
+    if (startpointManager != null) {
```
shanthoosh: I think it would be better to move the startpoint manager outside of offsetmanager. The utility of the startpointmanager ends after the container startup sequence. But the offsetmanager lifecycle is tied to the container lifecycle. It would be better to stop the startpoint manager after we update the offset and delete the startpoint and not have the idle metadata-store connection lurking around.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -404,6 +447,23 @@ class OffsetManager(
...
+    * Load Startpoints for each SystemStreamPartition+TaskName
+    */
+  private def loadStartpoints: Unit = {
+    if (startpointManager != null) {
+      val taskNameToSSPs: Map[TaskName, Set[SystemStreamPartition]] = systemStreamPartitions
+
+      taskNameToSSPs.foreach {
+        case (taskName, systemStreamPartitionSet) => {
+          startpoints += taskName -> systemStreamPartitionSet
+            .map(ssp => (ssp, startpointManager.readStartpointForTask(ssp, taskName)))
```
shanthoosh: When we update the starting offset for a taskName using the defined startpoint instead of the checkpointed offset, it would be better to log it to aid debuggability.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -261,7 +285,15 @@ class OffsetManager(
         val sspToOffsets = checkpoint.getOffsets
         if(sspToOffsets != null) {
           sspToOffsets.asScala.foreach {
-            case (ssp, cp) => offsetManagerMetrics.checkpointedOffsets.get(ssp).set(cp)
+            case (ssp, cp) => {
+              offsetManagerMetrics.checkpointedOffsets.get(ssp).set(cp)
+
+              // Delete startpoints once a new checkpoint is committed.
+              startpoints.get(taskName) match {
+                case Some(_) => startpointManager.deleteStartpointForTask(ssp, taskName)
```
shanthoosh: Do we need to check here if the ssp was defined in a startpoint for a task before deleting it. Let's say there're 1000 ssps mapped to a task in the commit and  the startpoint is defined for only one SSP of the task. Wouldn't this  make unnecessary delete calls?

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -93,6 +93,7 @@ object JobConfig {
   val JOB_SECURITY_MANAGER_FACTORY = "job.security.manager.factory"
 
   val METADATA_STORE_FACTORY = "metadata.store.factory"
+  val STARTPOINT_METADATA_STORE_FACTORY = "startpoint.metadata.store.factory"
```
shanthoosh: 1. Since the configuration `metadata.store.factory` allows us to have a  metadata store defined per samza job, just wondering if this is really necessary and what value add it provides? 
2. Do we have a use-case where some metadata of the job is stored in one store and the rest of them is stored in a different metadata store?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointSerde.java
```diff
@@ -0,0 +1,68 @@
...
+  private static final String POSITION_TYPE_FIELD = "positionType";
+  private static final String POSITION_FIELD = "position";
+  private static final String STORED_AT_FIELD = "storedAt";
+
+  private final ObjectMapper mapper = new ObjectMapper();
+
+  @Override
+  public Startpoint fromBytes(byte[] bytes) {
+    try {
+      LinkedHashMap<String, String> deserialized = mapper.readValue(bytes, LinkedHashMap.class);
```
shanthoosh: 1. Why not use the JsonSerde<V2> here for the serialization/deserialization?
2. We've defined custom message format  for startpoint here which is different from our existing message formats: [ContainerLocality](https://github.com/apache/samza/blob/master/samza-core/src/main/java/org/apache/samza/coordinator/stream/messages/SetContainerHostMapping.java), [TaskContainerMapping](https://github.com/apache/samza/blob/master/samza-core/src/main/java/org/apache/samza/coordinator/stream/messages/SetContainerHostMapping.java). I'm okay with it as long as it's json. Just sync up with prateek and see if he has any concerns. Only thing we lose is the consistency across  the different message formats.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
cameronlee314: If someone writes a startpoint for an SSP+taskName and then the job model changes, could the store end up with an entry that will never get deleted?

samza-core/src/test/java/org/apache/samza/startpoint/TestStartpointManager.java
cameronlee314: Minor: could you please use lower case for methods?

samza-core/src/test/scala/org/apache/samza/checkpoint/TestOffsetManager.scala
```diff
@@ -63,7 +69,7 @@ class TestOffsetManager {
     val checkpointManager = getCheckpointManager(systemStreamPartition, taskName)
     val systemAdmins = mock(classOf[SystemAdmins])
     when(systemAdmins.getSystemAdmin("test-system")).thenReturn(getSystemAdmin)
-    val offsetManager = OffsetManager(systemStreamMetadata, config, checkpointManager, systemAdmins, Map(), new OffsetManagerMetrics)
+    val offsetManager = OffsetManager(systemStreamMetadata, config, checkpointManager, getStartpointManager(), systemAdmins, Map(), new OffsetManagerMetrics)
```
cameronlee314: Could you please add more tests here to validate calls to startpoint manager?

samza-core/src/test/java/org/apache/samza/startpoint/TestStartpointManager.java
```diff
@@ -0,0 +1,292 @@
...
+import org.apache.samza.config.MapConfig;
+import org.apache.samza.container.TaskName;
+import org.apache.samza.container.grouper.stream.SystemStreamPartitionGrouper;
+import org.apache.samza.system.SystemStreamPartition;
+import org.apache.samza.util.NoOpMetricsRegistry;
+import org.junit.Assert;
+import org.junit.Test;
+
+
+public class TestStartpointManager {
```
cameronlee314: Could you please also add tests for StartpointKey and StartpointSerde?

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
cameronlee314: TTL would work. If the metadata store itself doesn't give that functionality, you could set a time in the entry and then check if it is "new enough" when handling it.
One undesirable rare case here would be if you don't delete that entry due to a changed job model, and then some time in the future, the mapping happens to work out again such that the job reads the entry and uses a very old startpoint.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -0,0 +1,273 @@
...
+    Preconditions.checkNotNull(metricsRegistry, "MetricsRegistry cannot be null");
+
+    String metadataStoreFactoryName = new JobConfig(config).getStartpointMetadataStoreFactory();
+    MetadataStore metadataStore;
+    try {
+      MetadataStoreFactory metadataStoreFactory = Util.getObj(metadataStoreFactoryName, MetadataStoreFactory.class);
+      metadataStore = metadataStoreFactory.getMetadataStore(NAMESPACE, config, metricsRegistry);
+      metadataStore.init();
+    } catch (ConfigException | NullPointerException ex) {
+      LOG.error("StartpointManager cannot be created because the underlying metadata store could not be configured or created.", ex);
```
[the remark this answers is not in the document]
dnishimura: Good point. I'll pass in the factory instead to create the metadata store.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
dnishimura: That's reasonable. Will introduce the notion of staleness and handle it on the reads.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -0,0 +1,273 @@
...
+   * @return {@link StartpointManager} instance.
+   */
+  public static StartpointManager getInstance(Config config, MetricsRegistry metricsRegistry) {
+    Preconditions.checkNotNull(config, "Config cannot be null");
+    Preconditions.checkNotNull(metricsRegistry, "MetricsRegistry cannot be null");
+
+    String metadataStoreFactoryName = new JobConfig(config).getStartpointMetadataStoreFactory();
+    MetadataStore metadataStore;
+    try {
+      MetadataStoreFactory metadataStoreFactory = Util.getObj(metadataStoreFactoryName, MetadataStoreFactory.class);
```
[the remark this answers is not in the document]
dnishimura: It is a good argument. Will pass in the factory instead.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: Change of direction. Took your suggestion instead.

samza-core/src/test/java/org/apache/samza/startpoint/TestStartpointManager.java
dnishimura: Jira created here: https://issues.apache.org/jira/browse/SAMZA-2045

dnishimura:
@cameronlee314 @shanthoosh there are new changes regarding your comments. Please take a look when you get a chance.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointSerde.java
```diff
@@ -0,0 +1,63 @@
...
+
+  void register(Class<? extends Startpoint> clazz) {
+    registry.put(clazz.getCanonicalName(), clazz);
+  }
+
+  @Override
+  public Startpoint fromBytes(byte[] bytes) {
+    try {
+      LinkedHashMap<String, String> deserialized = mapper.readValue(bytes, LinkedHashMap.class);
+      return mapper.readValue(deserialized.get(STARTPOINT_OBJ), registry.get(deserialized.get(STARTPOINT_CLASS)));
```
cameronlee314: You could use Class.forName to avoid needing a "registry". The disadvantage of using Class.forName is that you have to make sure to handle unexpected (or even malicious, if possible) class names in the store. Up to you.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -0,0 +1,265 @@
...
+   */
+  public Startpoint readStartpointForTask(SystemStreamPartition ssp, TaskName taskName) {
+    Preconditions.checkState(started, "Underlying metadata store not available");
+    Preconditions.checkNotNull(ssp, "SystemStreamPartition cannot be null");
+
+    byte[] startpointBytes = metadataStore.get(toStoreKey(ssp, taskName));
+
+    if (Objects.nonNull(startpointBytes)) {
+      Startpoint startpoint = startpointSerde.fromBytes(startpointBytes);
+      if (Instant.now().minus(expirationDuration).isBefore(Instant.ofEpochMilli(startpoint.getCreatedTimestamp()))) {
```
cameronlee314: Should there be a warning logged if the startpoint expired?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -0,0 +1,265 @@
...
+   */
+  public void writeStartpointForTask(SystemStreamPartition ssp, TaskName taskName, Startpoint startpoint) {
+    Preconditions.checkState(started, "Underlying metadata store not available");
+    Preconditions.checkNotNull(ssp, "SystemStreamPartition cannot be null");
+    Preconditions.checkNotNull(startpoint, "Startpoint cannot be null");
+
+    try {
+      metadataStore.put(toStoreKey(ssp, taskName), startpointSerde.toBytes(startpoint));
+    } catch (Exception ex) {
+      throw new SamzaException(String.format(
```
cameronlee314: read/delete don't wrap the storage operation in a try-catch, so would it be ok to get rid of this one too?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -0,0 +1,265 @@
...
+   * @param ssp The {@link SystemStreamPartition} that a {@link Startpoint} is keyed to.
+   * @param jobModel The {@link JobModel} is used to determine which {@link TaskName} each {@link SystemStreamPartition} maps to.
+   * @return The list of {@link TaskName}s mapped to the ssp.
+   */
+  public Set<TaskName> groupStartpointsPerTask(SystemStreamPartition ssp, JobModel jobModel) {
+    Preconditions.checkState(started, "Underlying metadata store not available");
+    Preconditions.checkNotNull(ssp, "SystemStreamPartition cannot be null");
+    Preconditions.checkNotNull(jobModel, "JobModel cannot be null");
+
+    Startpoint startpoint = readStartpoint(ssp);
```
cameronlee314: Instead of reading from an existing startpoint and rewriting the entry, would it be possible to make an API which just takes in the startpoint to expand and then directly write the expanded entries? Then we could do everything in a single step instead of what looks like two steps.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -274,6 +293,25 @@ class OffsetManager(
...
+        .intersect(startpoints.getOrElse(taskName, Map.empty[SystemStreamPartition, Startpoint]).keySet)
+
+      // delete startpoints for this task and the intersection of SSPs between checkpoint and startpoint.
+      sspStartpoints.foreach(ssp => {
+        startpointManager.deleteStartpointForTask(ssp, taskName)
+        info("Deleted startpoint for SSP: %s and task: %s" format (ssp, taskName))
+      })
+      startpoints -= taskName
+
+      if (startpoints.isEmpty) {
```
cameronlee314: Should we leave the StartpointManager running so that new startpoints can be written?

samza-api/src/main/java/org/apache/samza/startpoint/Startpoint.java
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Apply the visitor {@link StartpointConsumerVisitor}'s register methods to the instance of this {@link Startpoint}
+   * class.
+   * @param systemStreamPartition The {@link SystemStreamPartition} needed to register with the {@link StartpointConsumerVisitor}
+   * @param startpointConsumerVisitor The visitor to register with.
+   */
+  public abstract void apply(SystemStreamPartition systemStreamPartition, StartpointConsumerVisitor startpointConsumerVisitor);
+
+  @Override
+  public String toString() {
```
cameronlee314: Do the concrete versions of Startpoint need toString methods too, or does Objects.toStringHelper handle that?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -418,12 +420,23 @@ object SamzaContainer extends Logging {
       .orNull
     info("Got checkpoint manager: %s" format checkpointManager)
 
+    val metadataStoreFactory = Option(config.getStartpointMetadataStoreFactory)
+      .map(Util.getObj(_, classOf[MetadataStoreFactory]))
+      .orNull
+    var startpointManager = None: Option[StartpointManager]
```
cameronlee314: Minor: Is the type needed? If so, could you put it after the variable?
e.g. "var startpointManager: Option[StartpointManager] = None"

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -418,12 +420,23 @@ object SamzaContainer extends Logging {
...
+    try {
+      startpointManager = Some(StartpointManager.getWithMetadataStore(metadataStoreFactory, config, samzaContainerMetrics.registry))
+    } catch {
+      case e: Throwable =>
```
cameronlee314: Minor: maybe catch Exception instead? If it's a Throwable, it's probably pretty bad, so maybe let that exception get thrown.

samza-api/src/main/java/org/apache/samza/metadatastore/InMemoryMetadataStoreFactory.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.apache.samza.metadatastore;
+
+import java.util.concurrent.ConcurrentHashMap;
+import org.apache.samza.config.Config;
+import org.apache.samza.metrics.MetricsRegistry;
+
+
+/**
+ * Factory for an in-memory {@link MetadataStore}. Not intended for use in production.
```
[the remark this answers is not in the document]
cameronlee314: samza-test has a bunch of other test utilities too that people depend on for local/integration testing (e.g. TestRunner). If users want to depend on this for testing, I think it would make sense for them to depend on samza-test directly.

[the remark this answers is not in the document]
dnishimura: It's also used in `samza-core` unit tests.

samza-api/src/main/java/org/apache/samza/startpoint/Startpoint.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Startpoint represents a position in a stream partition.
+ */
+@InterfaceStability.Unstable
+public abstract class Startpoint {
+
+  private final Long createdTimestamp;
+
+  protected Startpoint() {
```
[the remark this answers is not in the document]
cameronlee314: I was thinking of the case where someone writes the Startpoint, and then resets the job, and then the Startpoint gets read. Then the write time is going to be significantly before the read time.
In my opinion, it would be nice to avoid setters, since it is slightly harder to ensure that it gets called correctly. If you require it as a constructor param, then the caller of the constructor could pass in Instant.now(). Maybe one way to also be consistent with @vjagadish1989 's suggestion is to not have a Startpoint.copyWithStoredAt and directly use the Startpoint (including the createdTimestamp in it) passed to StartpointManager.writeStartpointForTask.

samza-api/src/main/java/org/apache/samza/startpoint/Startpoint.java
```diff
@@ -0,0 +1,59 @@
...
+    this.createdTimestamp = Instant.now().toEpochMilli();
+  }
+
+  /**
+   * The timestamp when this {@link Startpoint} was written to the storage layer.
```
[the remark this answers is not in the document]
cameronlee314: If it's simpler, I don't think you necessarily need to only set it after it is successfully written.

samza-api/src/main/java/org/apache/samza/startpoint/StartpointConsumerVisitor.java
```diff
@@ -0,0 +1,63 @@
...
+ * Visitor interface for system consumers to implement to support {@link Startpoint}s.
+ */
+public interface StartpointConsumerVisitor {
+
+  /**
+   * Seek to specific offset represented by {@link StartpointSpecific}
+   * @param systemStreamPartition The {@link SystemStreamPartition} to seek the offset to.
+   * @param startpointSpecific The {@link Startpoint} that represents the specific offset.
+   */
+  void register(SystemStreamPartition systemStreamPartition, StartpointSpecific startpointSpecific);
```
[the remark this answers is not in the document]
cameronlee314: One way you could do it is to have a SystemCounsumer.register(SSP, Startpoint), and then have that delegate to a visitor implementation. That would give a nice consistent interface when using SystemConsumer, but also be able to leverage the visitor pattern.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
cameronlee314: That's helpful clarification. Do you have a use case for why the expansion needs to be done after the job starts up instead of when the Startpoint is written?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointSerde.java
cameronlee314: I think you would need to cast the resulting Class to a Class<? extends Startpoint>. But I think there is a method in Class to check if it extends Startpoint and then do the cast.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
cameronlee314: What if someone wanted to write some more Startpoints before restarting?

samza-api/src/main/java/org/apache/samza/metadatastore/InMemoryMetadataStoreFactory.java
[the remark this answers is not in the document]
cameronlee314: Thanks for clarifying that. I still feel that it is odd to put a class in samza-api which is only for testing. Is it a requirement for a metadata store to be persistent? Maybe we could just say that this is valid in production as long as you understand that it gets cleared on restart.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
prateekm: 👍 Btw, the type should be Option, not None or Some.

vjagadish1989:
Even though the current PR is functionally correct, I wonder if it'll be cleaner to take a different approach to the implementation?..

1. Intents to (re)set a startpoint for a given ssp are recorded in the metadata store
2. Each task writes back to the store when it's acknowledged processing the intent
 
Subsequent reads on the startpointManager reconcile [1] and [2] when constructing the startpoint. 

Also, this will potentially eliminate read-modify-write methods on the startpointManager.

cameronlee314:
> Even though the current PR is functionally correct, I wonder if it'll be cleaner to take a different approach to the implementation?..
> 
>     1. Intents to (re)set a startpoint for a given ssp are recorded in the metadata store
> 
>     2. Each task writes back to the store when it's acknowledged processing the intent
> 
> 
> Subsequent reads on the startpointManager reconcile [1] and [2] when constructing the startpoint.
> 
> Also, this will potentially eliminate read-modify-write methods on the startpointManager.

Would your suggestion treat the metadata store more like a log of actions than a KV store? Would the StartpointManager need to read a few different entries to figure out if something needs to be done?

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
cameronlee314: That makes sense that the job wouldn't apply the new Startpoints until the next restart.
However, does stopping the StartpointManager prevent it from writing the new Startpoints? To me, if something gets stopped, then none of its functionality should be expected to work.

samza-api/src/main/java/org/apache/samza/startpoint/StartpointVisitor.java
```diff
@@ -24,40 +24,48 @@
...
-public interface StartpointConsumerVisitor {
+public interface StartpointVisitor {
 
   /**
    * Seek to specific offset represented by {@link StartpointSpecific}
    * @param systemStreamPartition The {@link SystemStreamPartition} to seek the offset to.
    * @param startpointSpecific The {@link Startpoint} that represents the specific offset.
    */
-  void register(SystemStreamPartition systemStreamPartition, StartpointSpecific startpointSpecific);
+  void visit(SystemStreamPartition systemStreamPartition, StartpointSpecific startpointSpecific);
```
cameronlee314: Instead of overloading the "visit" method, could you please have different methods for each type? It makes the calls a little more explicit to make sure the correct visit call is made.
e.g. visitSpecific, visitTimestamp, visitOldest

samza-api/src/main/java/org/apache/samza/startpoint/StartpointVisitor.java
```diff
@@ -24,40 +24,48 @@
...
 
   /**
-   * Bootstrap signal represented by {@link StartpointBootstrap}
+   * Bootstrap signal represented by {@link StartpointCustom}
    * @param systemStreamPartition The {@link SystemStreamPartition} to seek the offset to.
-   * @param startpointBootstrap The {@link Startpoint} that represents the bootstrap signal.
+   * @param startpointCustom The {@link Startpoint} that represents the bootstrap signal.
    */
-  void register(SystemStreamPartition systemStreamPartition, StartpointBootstrap startpointBootstrap);
+  default void visit(SystemStreamPartition systemStreamPartition, StartpointCustom startpointCustom) {
```
cameronlee314: How can we ensure that implementations of StartpointCustom are serializable into the store? Also, will visitor implementations need to cast this to the actual concrete custom type being expected?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -207,18 +190,21 @@ public void deleteStartpointForTask(SystemStreamPartition ssp, TaskName taskName
...
+          if (startpoint == null) {
+            LOG.debug("No Startpoint for SSP: {} in task: {}", ssp, taskName);
+            continue;
           }
+
+          LOG.info("Grouping Startpoint keyed on SSP: {} to tasks determined by the job model.", ssp);
+          Startpoint startpointForTask = readStartpoint(ssp, taskName);
+          if (startpointForTask == null || startpointForTask.getCreationTimestamp() < startpoint.getCreationTimestamp()) {
+            writeStartpoint(ssp, taskName, startpoint);
+            sspsToDelete.add(ssp); // Mark for deletion
```
cameronlee314: Just to confirm: Is the idea here that if multiple taskNames map to a single SSP, then that SSP will just be put into this sspsToDelete set multiple times?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointSerde.java
```diff
@@ -30,21 +29,16 @@
...
 
   private final ObjectMapper mapper = new ObjectMapper();
 
   @Override
   public Startpoint fromBytes(byte[] bytes) {
     try {
       LinkedHashMap<String, String> deserialized = mapper.readValue(bytes, LinkedHashMap.class);
-      return mapper.readValue(deserialized.get(STARTPOINT_OBJ), SERDE_TYPES.get(deserialized.get(STARTPOINT_CLASS)));
+      Class<? extends Startpoint> startpointClass =
+          (Class<? extends Startpoint>) Class.forName(deserialized.get(STARTPOINT_CLASS));
```
cameronlee314: Would it be useful to check if startpointClass extends Startpoint before casting? Not sure how worried we should be that something else unexpectedly writes to this store.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
dnishimura: It shouldn't matter here. Only reads and deletes happen in the OffsetManager. Once all of that is completed, no other startpoint operation is expected. The writes would originate externally from a tool or service that writes the startpoints into the store.

samza-api/src/main/java/org/apache/samza/startpoint/StartpointVisitor.java
dnishimura: I like being explicit as well, although it sort of contradicts @prateekm 's suggestion in the `StartpointManager`, to refactor `writeStartpoint` and `writeStartpointForTask` into the same overloaded name `writeStartpoint` since the parameter types describe what it does. @prateekm - any guidance on when to overload a method name vs. having different but similar method names for the sake of being explicit on the action?

samza-api/src/main/java/org/apache/samza/startpoint/StartpointVisitor.java
dnishimura: I have a unit test for this in `TestStartpointSerde`, though I forgot to rename `testStartpointBootstrapSerde` to `testStartpointCustomSerde`, so it's good you commented here. Will change the test method name.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: That's correct and that's why I used a set instead of an arraylist.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointSerde.java
dnishimura: Might not be necessary to check, but this should throw a cast exception if it's not a `Startpoint` subclass and at least provide a small layer of safety.

cameronlee314: Ok, if it throws a class cast exception at this line, then you should be good.
I thought in some cases, incorrect casts might reveal themselves later, and I wasn't sure how far the code would go here before failing (if at all).

samza-api/src/main/java/org/apache/samza/startpoint/StartpointVisitor.java
cameronlee314: Your test case has a couple of primitive fields, which JSON can handle easily, but when there is an option to do arbitrary structures, some cases might not work well with JSON. Would it be useful to doc that JSON will be used and that implementations should be simple POJOs or something?

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
cameronlee314: Ah, thanks for clarifying the intention that writes would come from somewhere else.
Would it be simpler though to just let it close when the container shuts down? Is there a benefit in shutting down early?

samza-api/src/main/java/org/apache/samza/startpoint/StartpointVisitor.java
dnishimura: Good point. Will add documentation for it. Thanks.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
dnishimura: It would be simpler, but the underlying store may keep an open idle connection for no reason. If we can free up resources when not in use, I think that'd be better.

dnishimura:
Thanks @cameronlee314 for the review and approval. @prateekm @shanthoosh @vjagadish1989 please let me know if you would like me to address anything else; otherwise, your 👍  is appreciated 😄

samza-core/src/test/java/org/apache/samza/startpoint/TestStartpointManager.java
```diff
@@ -0,0 +1,265 @@
...
+  @Test
+  public void testDefaultMetadataStore() {
+    MapConfig config = new MapConfig();
+    StartpointManager startpointManager = new StartpointManager(new InMemoryMetadataStoreFactory(), config, new NoOpMetricsRegistry());
+    Assert.assertNotNull(startpointManager);
+    Assert.assertEquals(InMemoryMetadataStore.class, startpointManager.getMetadataStore().getClass());
+  }
+
+  @Test
+  public void testSpecifiedMetadataStore() {
```
jmakes: Is this test identical to the one above?

dnishimura: Ah good catch, it is redundant. It was needed before we came up with an `InMemoryMetadataStoreFactory` for testing. I'll remove it. Thanks.

dnishimura:
@prateekm - Last commit is in, CI has passed and is ready for your final blessing. Thanks!

PR #861, opened 2018-12-13
SAMZA-2043: Consolidate ReadableTable and ReadWriteTable

weisong44:
So far we've not seen a lot of use in maintaining separate implementation for ReadableTable and ReadWriteTable, which adds quite a bit complexity. Hence consolidating them.


samza-api/src/main/java/org/apache/samza/table/ReadWriteTable.java
```diff
@@ -32,7 +34,51 @@
  * @param <V> the type of the value in this table
  */
 @InterfaceStability.Unstable
-public interface ReadWriteTable<K, V> extends ReadableTable<K, V> {
+public interface ReadWriteTable<K, V> extends Table {
```
xinyuiscool: Seems with this patch we can replace Table with ReadWriteTable. If it's a too big change or it breaks too many apis, we can choose not do it or leave it for later.

weisong44: At this point it looks fine, the difference is do you cast to a RO or RW table in a task. With this patch we won't need the cast as well.

PR #863, opened 2018-12-17
SAMZA-2029: Samza-SQL Diagnostics: propagate eventTime and assign arrivalTime

shenodaguirguis:
a) upon receiving a new event message, check if eventTime is available, and if so copy it to message metadata, to propagate to SamzaSQL layer.

b) upon converting the event message to SamzaSQLRelMessage, assign current-time to be the event's arrival time (in the message metadata)


build.gradle
```diff
@@ -179,6 +179,7 @@ project(":samza-core_$scalaVersion") {
     compile("com.101tec:zkclient:$zkClientVersion") {
       exclude module: 'junit:junit'
     }
+    compile "org.apache.avro:avro:$avroVersion"
```
prateekm: We don't want an avro dependency in samza-core.

samza-core/src/main/java/org/apache/samza/operators/impl/InputOperatorImpl.java
```diff
@@ -48,10 +51,21 @@ protected void handleInit(Context context) {
...
     if (transformer != null) {
-      message = transformer.apply(ime);
+      tmpMessage = transformer.apply(ime);
+      // copy eventTime to output Samza (KV) message
+      if (tmpMessage instanceof KV && ime.getTimestamp() > 0L) {
+        Object kvValue = ((KV) tmpMessage).getValue();
+        if (kvValue != null && kvValue instanceof IndexedRecord) {
+          IndexedRecord retRec  = checkAndAddField((IndexedRecord) kvValue, "eventTime", Instant.ofEpochMilli(ime.getTimestamp()).toString());
+          retRec = checkAndAddField(retRec, "arrivalTime", Instant.now().toString());
+          message = new KV<>(((KV) tmpMessage).getKey(), retRec);
```
prateekm: Why add special logic for avro?

build.gradle
shenodaguirguis: removed & code refactored as discussed offline

samza-core/src/main/java/org/apache/samza/operators/impl/InputOperatorImpl.java
shenodaguirguis: refactored as discussed offline

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -68,12 +71,20 @@ public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, Stri
     this.size = size;
   }
 
-  public void setTimestamp(long timestamp) {
-    this.timestamp = timestamp;
+  public void setEventTime(long eventTime) {
```
prateekm: I'd prefer to add these params to constructors instead of setting via setters.

samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
```diff
@@ -209,7 +209,7 @@ private Windows() { }
    *    MessageStream<WindowPane<Void, Collection<Long>>> windowedStream =
    *        integerStream.window(Windows.tumblingWindow(Duration.ofSeconds(10)));
    *    MessageStream<Long> windowedPercentiles =
-   *        windowedStream.map(windowPane -> percentile99(windowPane.getMessage());
+   *        windowedStream.map(windowPane -> percentile99(windowPane.getEventMessage());
```
prateekm: Typo?

samza-api/src/main/java/org/apache/samza/serializers/SamzaMessageEnvelopeSerde.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.nio.ByteBuffer;
+import org.apache.samza.system.SamzaMessageEnvelope;
+
+
+/**
+ * a serializer for SamzaMessageEnvelope
+ */
+
+public class SamzaMessageEnvelopeSerde<M> implements Serde<SamzaMessageEnvelope> {
```
prateekm: Not sure why we're introducing a new framework-wide wrapper message, transformer and serde here. Let's talk offline.

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
shenodaguirguis: done

samza-api/src/main/java/org/apache/samza/serializers/SamzaMessageEnvelopeSerde.java
shenodaguirguis: this was an outdated upload

samza-api/src/main/java/org/apache/samza/operators/windows/Windows.java
shenodaguirguis: this was an outdated upload

samza-api/src/main/java/org/apache/samza/system/descriptors/GenericInputDescriptor.java
```diff
@@ -40,6 +40,10 @@
     super(streamId, serde, systemDescriptor, null);
   }
 
+  GenericInputDescriptor(String streamId, SystemDescriptor systemDescriptor, Serde serde, InputTransformer transformer) {
```
prateekm: This shouldn't be here. There was a conscious decision to not support stream level transformers. If this being used would like to understand the use case.

samza-core/src/main/scala/org/apache/samza/serializers/SerdeManager.scala
```diff
@@ -148,7 +148,10 @@ class SerdeManager(
         envelope.getSystemStreamPartition,
         envelope.getOffset,
         key,
-        message)
+        message,
+        0,
```
prateekm: Is this the size field? If so, shouldn't hardcode to 0, should use getSize instead.

samza-kafka/src/main/scala/org/apache/samza/system/kafka_deprecated/KafkaSystemConsumer.scala
```diff
@@ -284,12 +284,12 @@ private[kafka_deprecated] class KafkaSystemConsumer(
       }
 
       if(fetchLimitByBytesEnabled ) {
-        val ime = new IncomingMessageEnvelope(systemStreamPartition, offset, key, message, getMessageSize(msg.message))
-        ime.setTimestamp(if (!msg.message.isNull) msg.message.timestamp else 0L)
+        val ime = new IncomingMessageEnvelope(systemStreamPartition, offset, key, message, getMessageSize(msg.message),
+          (if (!msg.message.isNull) msg.message.timestamp else 0L), 0L)
```
prateekm: Extract as a field outside the if(fetchLimitByBytesEnabled) block.
Also, should we set arrivalTime to currentTimeMilis here?

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputTransformer.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.commons.logging.LogFactory;
+import org.apache.samza.operators.KV;
+import org.apache.samza.sql.data.SamzaSqlRelMsgMetadata;
+import org.apache.samza.system.IncomingMessageEnvelope;
+import org.apache.samza.system.descriptors.InputTransformer;
+
+
+/**
+ * SamzaSqlInputTransformer:
+ *   Given a userTransformer, this transformer applies the usreTransformer.apply
```
prateekm: Currently, transformers aren't user specified. They're determined by the SystemDescriptor implementation. Hence there shouldn't be a user specified input transformer in Samza SQL, since they don't create a SystemDescriptor, we do. Maybe we can remove the 'appSpecifiedTransformer' code?

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputTransformer.java
```diff
@@ -0,0 +1,74 @@
...
+ *   and when WrapMessage is true, it wraps transformed message into a SamzaMessageEnvelope
+ *   copying eventTime and arrivalTime from the input InputMessageEnvelope.
+ */
+
+public class SamzaSqlInputTransformer implements InputTransformer {
+
+  private InputTransformer appSpecifiedTransformer; // user provided transformer
+  private boolean isKeyed; // output is KV or not, in case userTransformer is null
+  private boolean wrapMessage; // wrap transformed message in SamzaMessageEnvelope or not
```
prateekm: When would this be false?

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -125,9 +127,12 @@ public void init(Context context) {
...
     @Override
-    public SamzaSqlRelMessage apply(KV<Object, Object> message) {
+    public SamzaSqlRelMessage apply(Object message) {
+      MessageAndMetadata messageAndMetadata = (MessageAndMetadata) message;
       Instant startProcessing = Instant.now();
-      SamzaSqlRelMessage retMsg = this.msgConverter.convertToRelMessage(message);
+      SamzaSqlRelMessage retMsg = this.msgConverter.convertToRelMessage((KV<Object, Object>) messageAndMetadata.message);
+      retMsg.getSamzaSqlRelMsgMetadata().setEventTime(messageAndMetadata.metadata.getEventTime());
+      retMsg.getSamzaSqlRelMsgMetadata().setArrivalTime(messageAndMetadata.metadata.getarrivalTime());
       retMsg.getSamzaSqlRelMsgMetadata().setScanTime(startProcessing.toString());
```
prateekm: Just wondering: Why is scanTime a string?

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -169,15 +174,27 @@ void translate(final TableScan tableScan, final String queryLogicalId, final Str
...
     }
 
-    KVSerde<Object, Object> noOpKVSerde = KVSerde.of(new NoOpSerde<>(), new NoOpSerde<>());
-    DelegatingSystemDescriptor
-        sd = systemDescriptors.computeIfAbsent(systemName, DelegatingSystemDescriptor::new);
-    GenericInputDescriptor<KV<Object, Object>> isd = sd.getInputDescriptor(streamId, noOpKVSerde);
+    // set the wrapper input transformer (SamzaSqlInputTransformer) in system descriptor
+    DelegatingSystemDescriptor sd = systemDescriptors.get(systemName);
+    if (sd == null) {
+      sd = new DelegatingSystemDescriptor(systemName,// null);
```
prateekm: Delete '// null);'.

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -169,15 +174,27 @@ void translate(final TableScan tableScan, final String queryLogicalId, final Str
       return;
     }
 
-    KVSerde<Object, Object> noOpKVSerde = KVSerde.of(new NoOpSerde<>(), new NoOpSerde<>());
-    DelegatingSystemDescriptor
-        sd = systemDescriptors.computeIfAbsent(systemName, DelegatingSystemDescriptor::new);
-    GenericInputDescriptor<KV<Object, Object>> isd = sd.getInputDescriptor(streamId, noOpKVSerde);
+    // set the wrapper input transformer (SamzaSqlInputTransformer) in system descriptor
+    DelegatingSystemDescriptor sd = systemDescriptors.get(systemName);
+    if (sd == null) {
```
prateekm: Minor: Maybe use Map#computeIfAbsent?

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -169,15 +174,27 @@ void translate(final TableScan tableScan, final String queryLogicalId, final Str
...
+        InputTransformer existingTransformer = sd.getTransformer().get();
+        if (!(existingTransformer instanceof SamzaSqlInputTransformer)) {
+          throw new SamzaException("SamzaSQL Exception: existing transformer for " + systemName + " is not SamzaSqlInputTransformer");
+        }
+      }
+    }
 
-    MessageStream<KV<Object, Object>> inputStream =
-        inputMsgStreams.computeIfAbsent(source, v -> streamAppDesc.getInputStream(isd));
+    InputDescriptor isd = sd.getInputDescriptor(streamId, new NoOpSerde<>(), sd.getTransformer().get());
```
prateekm: Is specifying transformer explicitly required? I think InputDescriptor should use the transformer for the SystemDescriptor it was obtained from by default.

samza-core/src/main/scala/org/apache/samza/serializers/SerdeManager.scala
shenodaguirguis: that would be a change in behavior: original code was using the instructor that hardcodes size=0

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
shenodaguirguis: you are right, this is related to the above comment on getInputDescriptor taking InputTransformer as input change which I rolled back.

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
shenodaguirguis: I throw exception if not absent

samza-api/src/main/java/org/apache/samza/system/descriptors/GenericInputDescriptor.java
shenodaguirguis: rolled back

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputTransformer.java
shenodaguirguis: when appSpecifiedTransformer is not null. Now appSpecifiedTransformer is removed, wrapMessage is also removed.

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputTransformer.java
shenodaguirguis: agreed & done

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -68,12 +71,19 @@ public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, Stri
     this.size = size;
```
prateekm: Maybe default arrival time to currentTimeMillis here?

samza-core/src/main/java/org/apache/samza/system/descriptors/DelegatingSystemDescriptor.java
```diff
@@ -47,6 +45,11 @@ public DelegatingSystemDescriptor(String systemName) {
     super(systemName, null, null, null);
   }
 
+  public DelegatingSystemDescriptor(String systemName, InputTransformer transformer) {
+    super(systemName, null, transformer, null);
+  }
+
```
prateekm: Nitpick: Delete extra newline.

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
```diff
@@ -68,7 +69,8 @@ void put(SystemStreamPartition ssp, Object key, Object message) {
       offset = IncomingMessageEnvelope.END_OF_STREAM_OFFSET;
     }
 
-    IncomingMessageEnvelope messageEnvelope = new IncomingMessageEnvelope(ssp, offset, key, message);
+    IncomingMessageEnvelope messageEnvelope = new IncomingMessageEnvelope(ssp, offset, key, message,
+        0, Instant.now().toEpochMilli(), 0L);
```
prateekm: I think both event and arrival time should be current time.

samza-core/src/main/scala/org/apache/samza/serializers/SerdeManager.scala
prateekm: I think that might have been an oversight. Wouldn't hurt to carry the value over.

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputTransformer.java
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * SamzaSqlInputTransformer:
+ *   Given a userTransformer, this transformer applies the usreTransformer.apply
+ *   and when WrapMessage is true, it wraps transformed message into a SamzaMessageEnvelope
+ *   copying eventTime and arrivalTime from the input InputMessageEnvelope.
+ */
+
+public class SamzaSqlInputTransformer implements InputTransformer {
+
+  private boolean isKeyed; // output is KV or not, in case userTransformer is null
```
prateekm: This would also always be false, right? Since there is no userTransformer?

shenodaguirguis: I changed  to depend on the message itself, if it is not KV already, create one

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputTransformer.java
```diff
@@ -0,0 +1,54 @@
...
+ *   Instance of KV, it creates KV. In both cases the KV is wrapped into a {@link MessageAndMetadata}
+ *   object, to copy over the message's meatadata (eventTime and arrivalTime)
+ */
+
+public class SamzaSqlInputTransformer implements InputTransformer {
+
+  @Override
+  public Object apply(IncomingMessageEnvelope ime) {
+    Object samzaMsg;
+    if (ime.getMessage() instanceof KV) {
```
prateekm: KV type is only used in High Level API after the transformer. It essentially contains (ime.key(), ime.message()). You can remove this check completely. `ime` should also not be null.

shenodaguirguis: done

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -66,14 +70,22 @@ public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, Stri
     this.key = key;
     this.message = message;
     this.size = size;
+    this.arrivalTime = Instant.now().toEpochMilli();
   }
 
-  public void setTimestamp(long timestamp) {
-    this.timestamp = timestamp;
+  public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, String offset,
```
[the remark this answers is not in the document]
shenodaguirguis: done

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -66,14 +70,22 @@ public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, Stri
...
+      Object key, Object message, int size, long eventTime, long arrivalTime) {
+    this(systemStreamPartition, offset, key, message, size);
+    this.eventTime = eventTime;
+    this.arrivalTime = arrivalTime;
+  }
+
+  public long getEventTime() {
```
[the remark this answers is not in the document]
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/MessageAndMetadata.java
```diff
@@ -0,0 +1,38 @@
...
+ * under the License.
+ */
+
+package org.apache.samza.sql;
+
+import org.apache.samza.sql.data.SamzaSqlRelMsgMetadata;
+
+
+public class MessageAndMetadata {
+  public Object message;
```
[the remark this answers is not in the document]
shenodaguirguis: right, but, we need to use generic Object to avoid an issue with java type inference related to samza predefined/default system, in setting the transformer and the NoOpSerde in ScanTranslator in particular

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputTransformer.java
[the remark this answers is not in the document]
shenodaguirguis: message is always not KV (I ran experiment and verify that). We get here from InputOperatorImpl when transformer is not null. Before introducing the SamzaSQLInputTransformer, InputOperatorImpl had transformer is null & isKeyed is true (because noOpKVSerde is KVSerde), and would  always created a KV. Now, with transformer is not null, the output is whatever we generate, so to have same effect we need to create the KV. I changed the code to always create the KV.

samza-sql/src/main/java/org/apache/samza/sql/MessageAndMetadata.java
```diff
@@ -0,0 +1,38 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.sql;
+
+import org.apache.samza.sql.data.SamzaSqlRelMsgMetadata;
+
+
+public class MessageAndMetadata {
```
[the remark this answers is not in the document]
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -169,15 +173,26 @@ void translate(final TableScan tableScan, final String queryLogicalId, final Str
...
     }
 
-    KVSerde<Object, Object> noOpKVSerde = KVSerde.of(new NoOpSerde<>(), new NoOpSerde<>());
-    DelegatingSystemDescriptor
-        sd = systemDescriptors.computeIfAbsent(systemName, DelegatingSystemDescriptor::new);
-    GenericInputDescriptor<KV<Object, Object>> isd = sd.getInputDescriptor(streamId, noOpKVSerde);
+    // set the wrapper input transformer (SamzaSqlInputTransformer) in system descriptor
+    DelegatingSystemDescriptor sd = systemDescriptors.get(systemName);
+    if (sd == null) {
+      sd = new DelegatingSystemDescriptor(systemName, new SamzaSqlInputTransformer());
```
sborya: nit. sd-> something more descriptive

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -169,15 +173,26 @@ void translate(final TableScan tableScan, final String queryLogicalId, final Str
...
+        InputTransformer existingTransformer = sd.getTransformer().get();
+        if (!(existingTransformer instanceof SamzaSqlInputTransformer)) {
+          throw new SamzaException("SamzaSQL Exception: existing transformer for " + systemName + " is not SamzaSqlInputTransformer");
+        }
+      }
+    }
 
-    MessageStream<KV<Object, Object>> inputStream =
-        inputMsgStreams.computeIfAbsent(source, v -> streamAppDesc.getInputStream(isd));
+    InputDescriptor isd = sd.getInputDescriptor(streamId, new NoOpSerde<>());
```
sborya: nit. isd->to something more descriptive

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputTransformer.java
```diff
@@ -0,0 +1,51 @@
...
+ *   from the IME to the SamzaSQL layer through the {@link SamzaSqlMessage}
+ */
+
+public class SamzaSqlInputTransformer implements InputTransformer {
+
+  @Override
+  public Object apply(IncomingMessageEnvelope ime) {
+    Assert.notNull(ime, "ime is null");
+    KV<Object, Object> keyAndMessageKV = KV.of(ime.getKey(), ime.getMessage());
+    if (keyAndMessageKV == null) {
```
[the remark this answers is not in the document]
shenodaguirguis: removed

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlMessage.java
```diff
@@ -0,0 +1,54 @@
...
+public class SamzaSqlMessage {
+  public KV<Object, Object> keyAndMessageKV;
+  public SamzaSqlRelMsgMetadata metadata;
+
+  /**
+   * Constructs a new SamzaSqMessage given the input arguments
+   * @param keyAndMessageKV
+   * @param metadata
+   */
+  public SamzaSqlMessage(KV<Object, Object> keyAndMessageKV, SamzaSqlRelMsgMetadata metadata) {
```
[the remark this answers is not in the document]
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -58,7 +59,7 @@
   private final int queryId;
 
   // FilterFunction to filter out any messages that are system specific.
-  private static class FilterSystemMessageFunction implements FilterFunction<KV<Object, Object>> {
+  private static class FilterSystemMessageFunction implements FilterFunction<Object> {
```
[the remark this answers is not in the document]
shenodaguirguis: it is now SamzaSqlMessage

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -76,8 +77,8 @@ public void init(Context context) {
     }
 
     @Override
-    public boolean apply(KV<Object, Object> message) {
-      return !relConverter.isSystemMessage(message);
+    public boolean apply(Object samzaSqlMessage) {
```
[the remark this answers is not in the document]
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -125,9 +126,12 @@ public void init(Context context) {
     }
 
     @Override
-    public SamzaSqlRelMessage apply(KV<Object, Object> message) {
+    public SamzaSqlRelMessage apply(Object message) {
```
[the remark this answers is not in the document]
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
shenodaguirguis: renamed to systemDescriptor

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -169,15 +173,26 @@ void translate(final TableScan tableScan, final String queryLogicalId, final Str
...
-        sd = systemDescriptors.computeIfAbsent(systemName, DelegatingSystemDescriptor::new);
-    GenericInputDescriptor<KV<Object, Object>> isd = sd.getInputDescriptor(streamId, noOpKVSerde);
+    // set the wrapper input transformer (SamzaSqlInputTransformer) in system descriptor
+    DelegatingSystemDescriptor sd = systemDescriptors.get(systemName);
+    if (sd == null) {
+      sd = new DelegatingSystemDescriptor(systemName, new SamzaSqlInputTransformer());
+      systemDescriptors.put(systemName, sd);
+    } else {
+      /* in SamzaSQL, there should be no systemDescriptor setup by user, so this branch happens only
+      * in case of Fan-OUT (i.e., same input stream used in multiple sql statements), or wehn same input
```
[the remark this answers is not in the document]
shenodaguirguis: done, thanks!

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
shenodaguirguis: renamed to inputDescriptor

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
```diff
@@ -68,7 +69,9 @@ void put(SystemStreamPartition ssp, Object key, Object message) {
       offset = IncomingMessageEnvelope.END_OF_STREAM_OFFSET;
     }
 
-    IncomingMessageEnvelope messageEnvelope = new IncomingMessageEnvelope(ssp, offset, key, message);
+    long currentTime = Instant.now().toEpochMilli();
+    IncomingMessageEnvelope messageEnvelope = new IncomingMessageEnvelope(ssp, offset, key, message,
+        0, currentTime, currentTime);
```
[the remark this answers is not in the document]
shenodaguirguis: done, but I will also cycle-back to Prateek who asked for it (I might have misunderstood his request)

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlMessage.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.samza.operators.KV;
+import org.apache.samza.sql.data.SamzaSqlRelMsgMetadata;
+
+
+/**
+ * Represents the message transformed from IncomingMessageEnvelope by SamzaSqlInputTransformer
+ * it contains the event message and its metadata
+ */
+public class SamzaSqlMessage {
+  public KV<Object, Object> keyAndMessageKV;
```
srinipunuru: Why not have key and value directly here rather than have this wrap KV?

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlMessage.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.apache.samza.operators.KV;
+import org.apache.samza.sql.data.SamzaSqlRelMsgMetadata;
+
+
+/**
+ * Represents the message transformed from IncomingMessageEnvelope by SamzaSqlInputTransformer
+ * it contains the event message and its metadata
+ */
+public class SamzaSqlMessage {
```
srinipunuru: Can you call this as SamzaSqlInputMessage.

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
```diff
@@ -79,7 +79,7 @@
   private final SamzaSqlApplicationConfig sqlConfig;
   private final StreamApplicationDescriptor streamAppDescriptor;
   private final Map<String, DelegatingSystemDescriptor> systemDescriptors;
-  private final Map<String, MessageStream<KV<Object, Object>>> inputMsgStreams;
+  private final Map<String, MessageStream<Object>> inputMsgStreams;
```
srinipunuru: Should this be of type Map<String, MessageStream<SamzaSqlMessage>>?

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -125,9 +126,11 @@ public void init(Context context) {
     }
 
     @Override
-    public SamzaSqlRelMessage apply(KV<Object, Object> message) {
+    public SamzaSqlRelMessage apply(SamzaSqlMessage samzaSqlMessage) {
       Instant startProcessing = Instant.now();
-      SamzaSqlRelMessage retMsg = this.msgConverter.convertToRelMessage(message);
+      SamzaSqlRelMessage retMsg = this.msgConverter.convertToRelMessage(samzaSqlMessage.keyAndMessageKV);
+      retMsg.getSamzaSqlRelMsgMetadata().setEventTime(samzaSqlMessage.metadata.getEventTime());
```
srinipunuru: This is an anti-pattern. 

1. When you are doing a get, You should expect that the object that is being returned is immutable. 
2. And in the implementation of get it is always a good idea to not return a mutable object. 

Ideally you should be able to initialize the SamzaSqlRelMessage with the metadata in the conversion layer that converts the SamzaSqlMessage to SamzaSqlRelMessage. It looks like you did this to not change the RelConverter interface. Can you add a comment saying that this should be fixed? RelConverter interface needs a cleanup, We can cleanup this code as part of that work.

samza-kafka/src/main/scala/org/apache/samza/system/kafka_deprecated/KafkaSystemConsumer.scala
```diff
@@ -283,15 +284,16 @@ private[kafka_deprecated] class KafkaSystemConsumer(
...
-      if(fetchLimitByBytesEnabled ) {
-        val ime = new IncomingMessageEnvelope(systemStreamPartition, offset, key, message, getMessageSize(msg.message))
-        ime.setTimestamp(if (!msg.message.isNull) msg.message.timestamp else 0L)
-        put(systemStreamPartition, ime)
-      } else {
-        val ime = new IncomingMessageEnvelope(systemStreamPartition, offset, key, message)
-        ime.setTimestamp(if (!msg.message.isNull) msg.message.timestamp else 0L)
-        put(systemStreamPartition, ime)
-      }
+      val eventTime = if (!msg.message.isNull) msg.message.timestamp else 0L
```
srinipunuru: This looks like a deprecated code. I am not sure whether we really need to update this.

shenodaguirguis: it is deprecated code, the setTimestamp API does no longer exist

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
```diff
@@ -169,15 +174,26 @@ void translate(final TableScan tableScan, final String queryLogicalId, final Str
...
-    DelegatingSystemDescriptor
-        sd = systemDescriptors.computeIfAbsent(systemName, DelegatingSystemDescriptor::new);
-    GenericInputDescriptor<KV<Object, Object>> isd = sd.getInputDescriptor(streamId, noOpKVSerde);
+    // set the wrapper input transformer (SamzaSqlInputTransformer) in system descriptor
+    DelegatingSystemDescriptor sd = systemDescriptors.get(systemName);
+    if (sd == null) {
+      sd = new DelegatingSystemDescriptor(systemName, new SamzaSqlInputTransformer());
+      systemDescriptors.put(systemName, sd);
+    } else {
+      /* in SamzaSQL, there should be no systemDescriptor setup by user, so this branch happens only
```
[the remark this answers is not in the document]
shenodaguirguis: Now transformer can be set in construction of the DelegatingSystemDescriptor, I throw error if Transformer is set to something other than SamzaSqlInputTransformer. And this path is hit in case of multiple usage of an input stream (as in the comment).

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlMessage.java
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
shenodaguirguis: it could. I changed it to Map<String, MessageStream<SamzaSqlInputMessage>>

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputMessage.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.samza.operators.KV;
+import org.apache.samza.sql.data.SamzaSqlRelMsgMetadata;
+
+
+/**
+ * Represents the message transformed from IncomingMessageEnvelope by SamzaSqlInputTransformer
+ * it contains the event message and its metadata
+ */
+public class SamzaSqlInputMessage {
+  public KV<Object, Object> keyAndMessageKV;
```
srinipunuru: Please make these private final.

samza-sql/src/main/java/org/apache/samza/sql/translator/ScanTranslator.java
shenodaguirguis: created SAMZA-2089 to track this & added comment

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputMessage.java
shenodaguirguis: done

PR #869, opened 2018-12-21
SAMZA-2022: Integrate startpoints with SystemConsumers

dnishimura:
This PR integrates the `Startpoint` with the `SystemConsumers` multiplexer class. The general logic is if there is a `Startpoint` for an SSP in a task and the `SystemConsumer` implements the `StartpointVisitor` interface, it will apply the visitor method via `Startpoint#apply(SystemStreamPartition, StartpointVisitor)`. Otherwise, the `SystemConsumer#register(SystemStreamPartition, checkpointOffset:String)` method is called as it currently does.

@shanthoosh @cameronlee314 @prateekm - Please take a look... after the break is fine. If I don't respond to comments within 24-48 hours during my leave, please feel free to hijack this PR into another PR.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/system/SystemConsumers.scala
+++ b/samza-core/src/main/scala/org/apache/samza/system/SystemConsumers.scala
@@ -25,6 +25,7 @@ import java.util.concurrent.TimeUnit
 import scala.collection.JavaConverters._
 import org.apache.samza.serializers.SerdeManager
 import org.apache.samza.util.{Logging, TimerUtil}
+import org.apache.samza.startpoint.{Startpoint, StartpointVisitor}
 import org.apache.samza.system.chooser.MessageChooser
 import org.apache.samza.SamzaException
 import java.util.ArrayDeque
@@ -33,6 +34,7 @@ import java.util.HashSet
 import java.util.HashMap
 import java.util.Queue
 import java.util.Set
+
 
 object SystemConsumers {
   val DEFAULT_POLL_INTERVAL_MS = 50
@@ -181,7 +183,7 @@ class SystemConsumers (
   }
 
 
-  def register(systemStreamPartition: SystemStreamPartition, offset: String) {
+  def register(systemStreamPartition: SystemStreamPartition, offset: String, startpoint: Startpoint) {
     debug("Registering stream: %s, %s" format (systemStreamPartition, offset))
 
     if (IncomingMessageEnvelope.END_OF_STREAM_OFFSET.equals(offset)) {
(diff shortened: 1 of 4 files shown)


dnishimura:
@shanthoosh @cameronlee314 @prateekm - If you have lots of comments and concerns about this PR, feel free to hijack into another PR. Otherwise, if there are only a few minor comments, I'm happy to address them and have this ready to commit. Thanks.
CC: @PawasChhokra @jmakes

shanthoosh:
@Sanil15 
I am assigned to work on the save point feature. So I will work on addressing your concern.

samza-core/src/main/scala/org/apache/samza/system/SystemConsumers.scala
```diff
@@ -192,10 +194,22 @@ class SystemConsumers (
 
     metrics.registerSystemStreamPartition(systemStreamPartition)
     unprocessedMessagesBySSP.put(systemStreamPartition, new ArrayDeque[IncomingMessageEnvelope]())
+
+    // Note regarding Startpoints and MessageChooser:
```
cameronlee314: Do we always want the startpoint to override the current offset? It seems to me like the chooser should not use the current offset when there is a startpoint, because the startpoint might correspond to a different current offset.

samza-core/src/main/scala/org/apache/samza/system/SystemConsumers.scala
```diff
@@ -192,10 +194,22 @@ class SystemConsumers (
...
+    // and needs to decide what's the lowest starting offset for an SSP that spans across multiple tasks so it knows
+    // to keep the highest priority on the SSP starting from the lowest starting offset until the SSP is fully
+    // bootstrapped to the UPCOMING offset. The offset here is ignored otherwise.
     chooser.register(systemStreamPartition, offset)
 
     try {
-      consumers(systemStreamPartition.getSystem).register(systemStreamPartition, offset)
+      val consumer = consumers(systemStreamPartition.getSystem)
+      if (startpoint != null && consumer.isInstanceOf[StartpointVisitor]) {
+        startpoint.apply(systemStreamPartition, consumer.asInstanceOf[StartpointVisitor])
```
cameronlee314: I don't remember if this was discussed in the previous PR, but what do you think about adding a `SystemConsumer.register(systemStreamPartition, startpoint)` method? Then each implementation could provide their own visitor and call `startpoint.apply`.
The `StartpointVisitor` interface itself does not necessarily mean that something has to be registered. It just is a way to specify that something should happen to a `Startpoint`, so it is a little bit odd to use `StartpointVisitor` as a marker to denote that something should be registered.

samza-core/src/main/scala/org/apache/samza/system/SystemConsumers.scala
```diff
@@ -192,10 +194,22 @@ class SystemConsumers (
...
+      } else {
+        consumer.register(systemStreamPartition, offset)
```
cameronlee314: Could this put the `offset` into a `StartpointSpecific` and then just always call `consumer.register(systemStreamPartition, startpoint)`?

samza-core/src/test/scala/org/apache/samza/system/TestSystemConsumers.scala
```diff
@@ -384,6 +420,30 @@ class TestSystemConsumers {
     def stop {}
     def register { super.register(systemStreamPartition, "0") }
   }
+
+  private class TestStartpointConsumer extends SerializingConsumer with StartpointVisitor {
```
cameronlee314: Instead of implementing a test object, would it be possible to use a mock and then verify the call to `visit`?

shanthoosh:
@dnishimura I discussed offline with @cameronlee314 . Due to these suggestions in reviews, it will be better for me to take this PR over and not bother you during vacation :). Feel free to close this PR.

dnishimura:
@shanthoosh since this doesn't have any changes to existing behavior, an alternative is to commit this PR since it has a couple of approvals and apply @cameronlee314 and @Sanil15 's comments in a separate PR. Just a suggestion to save any additional work. I'm :ok: with either way... let me know what you think. Thanks!

shanthoosh:
@prateekm @cameronlee314 I agree to @dnishimura suggestion. I can create a followup ticket to address the feedback in this PR.

Do you guys agree to checking this in?

prateekm:
I'm ok with fixing forward. Let's check this in if @cameronlee314 doesn't have an objection / isn't available by EOD.

cameronlee314:
> I'm ok with fixing forward. Let's check this in if @cameronlee314 doesn't have an objection / isn't available by EOD.

@shanthoosh, can you please double check that if we check this in, then Samza should still be in a working state (don't spend time doing extra testing, but maybe reread the changes one more time)? If so, then I am good to check this in and do a follow-up PR. I just want to make sure that no one else is blocked in testing other features until the follow-up gets checked in.

shanthoosh:
@cameronlee314
If there're no startpoints published, then this should be a No-OP. 

Startpoint store is defaulted to coordinator stream by default. This would instantiate a coordinator-stream consumer and read all messages from the stream unnecessarily even when there're no startpoints. 

In addition to checking this in, before this feature is completed we could remove the defaults for the startpoint store.

What do you think?

shanthoosh:
@cameronlee314 @prateekm 
Created this follow-up ticket: https://issues.apache.org/jira/browse/SAMZA-2083 to track and address the review-comments from this PR.

shanthoosh:
@dnishimura Thanks for the changes. Merged this patch to trunk.

dnishimura:
Thanks @shanthoosh!

PR #871, opened 2019-01-08
SAMZA-2056 : Adding a TaskMode in the TaskModel

rmatharu:
This PR adds a TaskMode (an enum) to the TaskModel. 
This assignment is persisted to the metastore using a SetTaskModeMapping type.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
+++ b/samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.job.model;
+
+public enum TaskMode {
+  Active ("active"),
+  HostState ("hostState");
+
+  private final String name;
+
+  private TaskMode(String s) {
+    name = s;
+  }
+
+  public String toString() {
+    return this.name;
+  }
+}
--- a/samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
+++ b/samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
@@ -21,10 +21,10 @@ package org.apache.samza.job.model;
 /**
  * This defines the logical mode of a taskInstance.
  * Active is the defacto mode for a task, i.e., tasks processing input, reading/writing state, producing output, etc.
- * StandbyState is the mode for tasks, that maintain warmed-up KV state by reading from its changelog.
+ * Standby is the mode for tasks, that maintain warmed-up KV state by reading from its changelog.
  */
 public enum TaskMode {
-  Active("active"), StandbyState("standbyState");
+  Active("active"), Standby("standby");
 
   private final String mode;
 
--- a/samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
+++ b/samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
@@ -18,9 +18,13 @@
  */
 package org.apache.samza.job.model;
 
+/**
+ * This defines the logical mode of a taskInstance.
+ * Active is the defacto mode for a task, i.e., tasks processing input, reading/writing state, producing output, etc.
+ * StandbyState is the mode for tasks, that maintain warmed-up KV state by reading from its changelog.
+ */
 public enum TaskMode {
-  Active ("active"),
-  HostState ("hostState");
+  Active("active"), StandbyState("standbyState");
 
   private final String name;
 
--- a/samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
+++ b/samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
@@ -24,7 +24,7 @@ package org.apache.samza.job.model;
  * Standby is the mode for tasks, that maintain warmed-up KV state by reading from its changelog.
  */
 public enum TaskMode {
-  Active("active"), Standby("standby");
+  Active("Active"), Standby("Standby");
 
   private final String mode;
 
--- a/samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
+++ b/samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
@@ -26,13 +26,13 @@ package org.apache.samza.job.model;
 public enum TaskMode {
   Active("active"), StandbyState("standbyState");
 
-  private final String name;
+  private final String mode;
 
-  private TaskMode(String s) {
-    name = s;
+  private TaskMode(String mode) {
+    this.mode = mode;
   }
 
   public String toString() {
-    return this.name;
+    return this.mode;
   }
 }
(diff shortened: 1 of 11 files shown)


samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * This defines the logical mode of a taskInstance.
+ * Active is the defacto mode for a task, i.e., tasks processing input, reading/writing state, producing output, etc.
+ * StandbyState is the mode for tasks, that maintain warmed-up KV state by reading from its changelog.
+ */
+public enum TaskMode {
+  Active("active"), StandbyState("standbyState");
+
+  private final String name;
```
shanthoosh: Minor: may be rename to mode.

samza-api/src/main/java/org/apache/samza/job/model/TaskModel.java
```diff
@@ -35,11 +35,23 @@
...
+  private final TaskMode taskMode;
 
+  /**
+   *  Create a TaskModel for an active task with the given taskName, SSPs, and changelogPartition.
+   * @param taskName The desired taskName
+   * @param systemStreamPartitions SSPs assigned to this task.
+   * @param changelogPartition The changelog SSP for this task.
+   */
   public TaskModel(TaskName taskName, Set<SystemStreamPartition> systemStreamPartitions, Partition changelogPartition) {
+    this(taskName, systemStreamPartitions, changelogPartition, TaskMode.Active);
```
shanthoosh: I think you have to change the TaskModelMixin for this new field addition.

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
```diff
@@ -131,12 +141,12 @@ public void writeTaskContainerMapping(String taskName, String containerId) {
...
     for (String taskName : taskNames) {
-      metadataStore.delete(taskName);
+      taskContainerMappingMetadataStore.delete(taskName);
       taskNameToContainerId.remove(taskName);
     }
   }
 
   public void close() {
-    metadataStore.close();
+    taskContainerMappingMetadataStore.close();
```
shanthoosh: I think it will be better to close the other store as well.

samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
```diff
@@ -0,0 +1,38 @@
...
+ * under the License.
+ */
+package org.apache.samza.job.model;
+
+/**
+ * This defines the logical mode of a taskInstance.
+ * Active is the defacto mode for a task, i.e., tasks processing input, reading/writing state, producing output, etc.
+ * StandbyState is the mode for tasks, that maintain warmed-up KV state by reading from its changelog.
+ */
+public enum TaskMode {
```
shanthoosh: Can we make this static enum within the TaskModel itself. There seems to be no other public usage other than in TaskModel.

samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
rmatharu: done

samza-api/src/main/java/org/apache/samza/job/model/TaskModel.java
rmatharu: done

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
rmatharu: done

samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
rmatharu: I have a couple of other PRs inflight (for standby tasks) that make extensive use of this.

samza-api/src/main/java/org/apache/samza/job/model/TaskMode.java
```diff
@@ -0,0 +1,38 @@
...
+  Active("active"), StandbyState("standbyState");
```
prateekm: Minor: would prefer Standby instead of StandbyState

samza-api/src/main/java/org/apache/samza/job/model/TaskModel.java
```diff
@@ -35,11 +35,23 @@
...
+  }
+
+  public TaskModel(TaskName taskName, Set<SystemStreamPartition> systemStreamPartitions, Partition changelogPartition, TaskMode taskMode) {
```
prateekm: Nitpick: Move the javadoc to this constructor instead and document what @param taskMode is. Maybe move the other constructor below this one.

samza-api/src/main/java/org/apache/samza/job/model/TaskModel.java
```diff
@@ -66,6 +78,10 @@ public Partition getChangelogPartition() {
     return changelogPartition;
   }
 
+  public TaskMode getTaskMode() {
+    return this.taskMode;
+  }
+
   @Override
   public boolean equals(Object o) {
     if (this == o) {
```
prateekm: Need to update equals and hashcode?

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
```diff
@@ -91,8 +98,8 @@ public TaskAssignmentManager(Config config, MetricsRegistry metricsRegistry, Ser
    */
   public Map<String, String> readTaskAssignment() {
```
prateekm: Does this need to be updated to return the task mode too? Alternatively, do we need a separate getter for TaskMode?

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
```diff
@@ -106,8 +113,9 @@ public TaskAssignmentManager(Config config, MetricsRegistry metricsRegistry, Ser
    *
    * @param taskName    the task name
    * @param containerId the SamzaContainer ID or {@code null} to delete the mapping
+   * @param taskMode the mode of the task
    */
-  public void writeTaskContainerMapping(String taskName, String containerId) {
+  public void writeTaskContainerMapping(String taskName, String containerId, TaskMode taskMode) {
     String existingContainerId = taskNameToContainerId.get(taskName);
     if (existingContainerId != null && !existingContainerId.equals(containerId)) {
       LOG.info("Task \"{}\" moved from container {} to container {}", new Object[]{taskName, existingContainerId, containerId});
```
prateekm: Minor: update logs to include mode.

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
```diff
@@ -131,12 +141,13 @@ public void writeTaskContainerMapping(String taskName, String containerId) {
    */
   public void deleteTaskContainerMappings(Iterable<String> taskNames) {
```
prateekm: Do we need to delete task mode mapping wherever this is being called?

In general, would it make sense to get/put/delete them together as a complex value?

samza-core/src/main/java/org/apache/samza/serializers/model/JsonTaskModelMixIn.java
```diff
@@ -45,4 +46,7 @@ public JsonTaskModelMixIn(@JsonProperty("task-name") TaskName taskName, @JsonPro
 
   @JsonProperty("changelog-partition")
   abstract Partition getChangelogPartition();
+
+  @JsonProperty("task-mode")
```
prateekm: Do you need to do anything to mark this as optional or will this work as-is if the property is missing?

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
rmatharu: yes, I have a separate PR that introduces and uses a separate getter for TaskModes.

samza-core/src/main/java/org/apache/samza/serializers/model/JsonTaskModelMixIn.java
rmatharu: It'll be ok asis because at the class level we have @JsonIgnoreProperties(ignoreUnknown = true). 
It'll set it to java default, and my other PR will handle the case in code.

rmatharu:
Addressed all review comments.

samza-core/src/main/java/org/apache/samza/serializers/model/JsonTaskModelMixIn.java
rmatharu: Thanks. Added deserializer for handling this case. 
Also added a test for this case in TestSamzaObjectMapper.

PR #873, opened 2019-01-10
Samza 2070 : Upgrade calcite and  add tests for array, map and nested records

srinipunuru:
Upgraded calcite version which adds support for array, map. It also adds basic support for nested records through a DOT Operator. But the DOT operator doesn't work item with its execution engine. because of couple of issues.

1. When there is a structured record, Calcite flattens the nested record through a LogicalProject using RelStructuredTypeFlattener. which generates plan like the one below. So it introduces RexFieldAccess operator to access fields "$4.zip". But RexToLixTranslator doesn't support RexFieldAccess yet. 

`
LogicalProject(EXPR$0=[DOT(ITEM($7, 0), 'kind')])
  LogicalProject(__key__=[$0], id=[$1], name=[$2], companyId=[$3], zip=[$4.zip], number=[$4.streetnum.number], selfEmployed=[$5], phoneNumbers=[$6], mapValues=[$7])
    EnumerableTableScan(table=[[testavro, PROFILE]])
`

2. There is no Operator implementation for DOT Operator in the RexImpTable.

Hence disabled the NestedRecord test for now.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -318,6 +318,12 @@ project(":samza-elasticsearch_$scalaVersion") {
 project(":samza-sql_$scalaVersion") {
   apply plugin: 'java'
 
+  // Disabling assertions while running test because of a calcite bug in {@link RelOptUtil#eq} method
+  // it checks for the type1 != type2 rather than !type1.equals(type2)
+  tasks.withType(Test) {
+    enableAssertions = false
+  }
+
   dependencies {
     compile project(':samza-api')
     compile project(":samza-kafka_$scalaVersion")
@@ -786,6 +792,12 @@ project(":samza-test_$scalaVersion") {
   // tasks.compileTestJava.enabled = false
   sourceSets.main.java.srcDirs = []
   sourceSets.test.java.srcDirs = []
+
+  // Disabling assertions while running test because of a calcite bug in {@link RelOptUtil#eq} method
+  // it checks for the type1 != type2 rather than !type1.equals(type2)
+  tasks.withType(Test) {
+    enableAssertions = false
+  }
 
   configurations {
     // Remove transitive dependencies from Zookeeper that we don't want.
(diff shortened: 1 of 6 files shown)


samza-test/src/test/java/org/apache/samza/test/samzasql/TestSamzaSqlEndToEnd.java
```diff
@@ -283,6 +283,46 @@ public void testEndToEndFlatten() throws Exception {
...
+
+  @Test
+  public void testEndToEndComplexRecord() {
+    int numMessages = 10;
+    TestAvroSystemFactory.messages.clear();
+    Map<String, String> staticConfigs = SamzaSqlTestConfig.fetchStaticConfigsWithFactories(configs, numMessages);
+
+    String sql1 =
+        "Insert into testavro.outputTopic"
+            + " select map_values['key0'] as string_value, array_values[0] as string_value, map_values, id, bytes_value, fixed_value, float_value "
```
shenodaguirguis: map_string_value & array_string_value?

build.gradle
```diff
@@ -787,6 +793,12 @@ project(":samza-test_$scalaVersion") {
   sourceSets.main.java.srcDirs = []
   sourceSets.test.java.srcDirs = []
 
+  // Disabling assertions while running test because of a calcite bug in {@link RelOptUtil#eq} method
```
atoomula: Is this a known issue in Calcite ? Could we have the ticket here so that we could enable the assertions when it is fixed ?

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroTypeFactoryImpl.java
```diff
@@ -79,9 +80,9 @@ private RelDataType convertRecordType(Schema schema) {
   private RelDataType getRelDataType(Schema fieldSchema) {
     switch (fieldSchema.getType()) {
       case ARRAY:
-        // TODO JavaTypeFactoryImpl should convert Array into Array(ANY, ANY)
-        // return new ArraySqlType(createSqlType(SqlTypeName.ANY), true);
-        return createTypeWithNullability(createSqlType(SqlTypeName.ANY), true);
+        RelDataType elementType = getRelDataType(fieldSchema.getElementType());
+         return new ArraySqlType(elementType, true);
+//        return createTypeWithNullability(createSqlType(SqlTypeName.ANY), true);
```
atoomula: do we need this comment anymore now that we have added array support ?

build.gradle
srinipunuru: Yup it's a bug in calcite code in RelOptUtil. Opened https://issues.apache.org/jira/browse/SAMZA-2069 to track this.

PR #878, opened 2019-01-15
SAMZA-2065: Moving InMemory Stream and System Descriptors to samza-core

Sanil15:
@xinyuiscool @bharathkk for review


prateekm:
@Sanil15 Why are these in core?

Sanil15:
@prateekm consumer/producer for InMemorySystem is in core and descriptors were in test, hence we need this to be in core. In addition to that these descriptors now do not tightly couple with TestRunner

Sanil15:
@xinyuiscool will do

cameronlee314:
> @prateekm consumer/producer for InMemorySystem is in core and descriptors were in test, hence we need this to be in core. In addition to that these descriptors now do not tightly couple with TestRunner

In general, users should not have to depend on samza-core. Descriptors are part of the API. Maybe we could have a separate module called samza-in-memory (or something like that), similar to how the kafka descriptors are in samza-kafka.

Sanil15:
I remember samza-inmemory used to exist in samza-test and then the consumer/producer were moved to core @bharathkk do you remember why this was done.

shanthoosh:
@Sanil15 @prateekm 
Is this migration intended for [BoundedSystemStream](https://github.com/apache/beam/blob/master/runners/samza/src/main/java/org/apache/beam/runners/samza/adapter/BoundedSourceSystem.java) use-cases in beam? 


Please correct me if i'm wrong. Looking at the implementation,  InMemoryConsumer/InMemoryProducer generates a testId for each run based on some user-defined config. Is that semantics applicable outside of the samza test environment.

xinyuiscool:
Agree with @cameronlee314. The descriptors should go to api. THe consumer/producer will be better if kept inside core. that'll be a nice separation.

prateekm:
@Sanil15 @xinyuiscool why are we doing this though? Why make these non-test in the first place? Are you using this for something?

xinyuiscool:
@prateekm : inmemory can be useful in other scenarios, e.g. populate some fixed data set into the pipeline. Or it can be useful for other application to use, e.g. in my beam runner, I also need to create some data into Samza pipeline without using the test runner. Generally this can be quite useful.

As a fair comparison, both Beam and Flink supports creating dataset from java collections in their apI. For beam, it's Create.of(). For flink, here is how it looks:

        DataSet<String> text = env.fromElements(
            "Who's there?",
            "I think I hear them. Stand, ho! Who's there?");

prateekm:
@xinyuiscool Even for the use cases you mentioned, I'm not sure if there are non-test/demo uses cases for creating streams with small deterministic in-memory collections?

I'm OK with moving them to API, but it's not sufficient to just change package names. We should clearly document what capabilities these "in memory streams" provide, and what can/should they be used for. E.g., if they're primarily meant for dev environments, testing and demo apps, I don't want users accidentally using them in place of actual streams expecting them to work.

xinyuiscool:
This is actually for usability, I believe. With this api, I can quickly create a samza pipeline and test it out some logic without setting up the heavy lifting kafka or other streams. In another word, it's good for prototyping. There is some special use case that might require injecting some fixed data set too, e.g. Beam Impulse can use this to inject the binary array to init the following operators to read the read data source, which today we create a very simple factory which does what similar things.

I agree we need good documentation for this and clean it up so it can be user-facing. I do see this is an opportunity to make this Inmemory api better and have no coupling with samza test runner.

prateekm:
@xinyuiscool Agree that this helps with prototyping. I think we're on the same page regarding goals. Would you mind creating tickets for tracking documentation and API changes (if necessary) for making this a public API? Will @Sanil15 be working on them? I'd prefer those changes to go in first, or at least along with the PR to move these classes from samza-test to samza-api.

prateekm:
@Sanil15 Are you still planning to check this in?

Sanil15:
@prateekm let me close this for now, I will pick it up later since this needs way more refactoring and documentation

PR #879, opened 2019-01-15
SAMZA-2059: Persist configuration in coordinator stream for standalone.

shanthoosh:
Prior to Samza 1.0, users plugged in the properties of an I/O system through a configuration file. Samza employed rewriters in the user-defined order to compute the configuration of a job.

Post Samza 1.0, we introduced new abstractions viz` StreamDescriptor` and `SystemDescriptor` in samza, with the purpose of performing configuration expansion for predefined systems at run-time.

Configuration computed at run-time is not persisted at a centralized storage in samza-standalone. This breaks the functionality of the tools viz checkpoint-tool, coordinator-stream-writer, etc in samza standalone. This patch addresses this problem by storing the configuration in coordinator stream for standalone. 

In the follow up PR's:
1. We'll switch from zookeeper to coordinator-stream as JobModel storage layer in standalone
2.  Samza tools(checkpoint-tool) will be migrated to read the configuration from coordinator stream rather than from disk.


shanthoosh:
@xinyuiscool @sborya 

PR #880, opened 2019-01-15
SAMZA-2066 Refactor remote tables to separate various concerns

weisong44:
Currently different features such as rate limiting, retries, etc. are implemented together in remote table, the implementation will become more and more complex and error prone as we add more functionality. It will be necessary to separate theses features into their own class/module. This is also a necessary step as we move on to add batching support.

 - Converted to a composition-based implementation
 - The remote table would only provide core functionality and basic metrics
 - Splitted ReadWriteTable into sync and async part (AsyncReadWriteTable)
 - Introduced AsyncRateLimitedTable, AsyncRetriableTable


samza-core/src/main/java/org/apache/samza/table/BaseReadWriteTable.java
```diff
@@ -66,4 +67,12 @@ public void init(Context context) {
   public String getTableId() {
     return tableId;
   }
+
+  public interface Func0 {
+    void apply();
+  }
+
+  public interface Func1<T> {
+    CompletableFuture<T> apply();
```
vjagadish1989: Can't you use a `Supplier<CompletableFuture<T>>` if all you need is an interface that returns a future?

samza-core/src/main/java/org/apache/samza/table/BaseReadWriteTable.java
```diff
@@ -66,4 +67,12 @@ public void init(Context context) {
   public String getTableId() {
     return tableId;
   }
+
+  public interface Func0 {
+    void apply();
```
vjagadish1989: Can you take in a `Function` instead of defining our own `Func0`?

samza-core/src/main/java/org/apache/samza/table/retry/AsyncRetriableTable.java
```diff
@@ -0,0 +1,158 @@
...
+    this.tableId = tableId;
+    this.table = table;
+    this.retryExecutor = retryExecutor;
+
+    if (readRetryPolicy != null && readFn != null) {
+      Predicate<Throwable> readRetryPredicate = readRetryPolicy.getRetryPredicate();
+      readRetryPolicy.withRetryPredicate((ex) -> readFn.isRetriable(ex) || readRetryPredicate.test(ex));
+      this.readRetryPolicy = FailsafeAdapter.valueOf(readRetryPolicy);
+    } else {
+      this.readRetryPolicy = null;
```
dengpan-yin: can the "else" part be removed since by default the object variable of a class is null?
Is it possible to refactor the code for readRetryPolicy and writeRetryPolicy since they are very similar?

samza-api/src/main/java/org/apache/samza/table/remote/TableRateLimiter.java
```diff
@@ -79,11 +75,12 @@
    * @param tag tag to be used with the rate limiter
    */
   public TableRateLimiter(String tableId, RateLimiter rateLimiter, CreditFunction<K, V> creditFn, String tag) {
+    Preconditions.checkNotNull(rateLimiter);
```
dengpan-yin: should tableId and creditFn be null checked as well?

samza-core/src/main/java/org/apache/samza/table/remote/AsyncRemoteTable.java
```diff
@@ -0,0 +1,104 @@
...
+
+
+/**
+ * A composable asynchronous table implementation that delegates read/write operations
+ * to the underlying table read and write functions.
+ *
+ * @param <K> the type of the key in this table
+ * @param <V> the type of the value in this table
+ */
+public class AsyncRemoteTable<K, V> implements AsyncReadWriteTable<K, V> {
```
dengpan-yin: should we have a AsyncLocalTable class as well?

samza-core/src/main/java/org/apache/samza/table/retry/AsyncRetriableTable.java
weisong44: They are final variables, so they would need a value assignment.
Can you elaborate more re readRetryPolicy and writeRetryPolicy, the intent is to provide different retry behavior for read and write.

samza-api/src/main/java/org/apache/samza/table/remote/TableRateLimiter.java
weisong44: Good catch, creditFn is optional though.

samza-core/src/main/java/org/apache/samza/table/remote/AsyncRemoteTable.java
weisong44: AsyncRemoteTable is separated out to facilitate the composition construct, primarily due to the number of different features on remote side. I'd keep this option for local table for a later point.

samza-core/src/main/java/org/apache/samza/table/BaseReadWriteTable.java
weisong44: The closest one is Runnable, which feels odd. Do you have a suggestion?

vjagadish1989: >> The closest one is Runnable, which feels odd

I can also see why you think Runnable can be odd - esp., given its association with the Thread class.

>> Do you have a suggestion?

As a personal preference, I would use a `Runnable` instead of defining a Samza-specific interface with the same signature. I'll leave the final call to you.

samza-core/src/main/java/org/apache/samza/table/remote/RemoteTable.java
```diff
@@ -90,33 +101,46 @@
...
     super(tableId);
-    Preconditions.checkNotNull(readFn, "null read function");
+    Preconditions.checkNotNull(readFn, "null readFn");
+
     this.readFn = readFn;
     this.writeFn = writeFn;
     this.readRateLimiter = readRateLimiter;
     this.writeRateLimiter = writeRateLimiter;
-    this.tableExecutor = tableExecutor;
+    this.rateLimitingExecutor = rateLimitingExecutor;
```
vjagadish1989: You get better encapsulation by moving the rateLimits and retries entirely out of the RemoteTable. 

Here's one organization that comes to mind:

BaseRemoteTable <-- a basic RWTable implementation using reader functions and writer functions 
RetriableTable <- wraps a table with retries
RateLimitedTable <- wraps a table with rate-limits

A factory class like `RemoteTableProvider` can then instantiate the desired `RemoteTable` instance by wrapping it with the right set of decorators.

samza-core/src/main/java/org/apache/samza/table/BaseReadWriteTable.java
weisong44: Yeah, I have a weird feeling about Runnable, so let's keep Func0 for now.

samza-core/src/main/java/org/apache/samza/table/ratelimit/AsyncRateLimitedTable.java
```diff
@@ -0,0 +1,148 @@
...
+import org.apache.samza.table.utils.TableMetricsUtil;
+
+
+/**
+ * A composable read and/or write rate limited asynchronous table implementation
+ *
+ * @param <K> the type of the key in this table
+ * @param <V> the type of the value in this table
+ */
+public class AsyncRateLimitedTable<K, V> implements AsyncReadWriteTable<K, V> {
```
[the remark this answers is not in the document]
weisong44: As we spoke, the core of remote table is pure async and sync part is added only at last as a thin layer in remote table to avoid unnecessary duplication. However, It may be good optimization to move the construction of the actually remote table instance to table provider, which is much cleaner. I will open another Jira to track this.

samza-core/src/main/java/org/apache/samza/table/remote/RemoteTable.java
weisong44: Please see reply regarding sync/asncy

samza-api/src/main/java/org/apache/samza/table/ReadWriteTable.java
```diff
@@ -20,29 +20,18 @@
...
-public interface ReadWriteTable<K, V> extends Table {
-
-  /**
-   * Initializes the table during container initialization.
-   * Guaranteed to be invoked as the first operation on the table.
-   * @param context {@link Context} corresponding to this table
-   */
-  default void init(Context context) {
-  }
+public interface ReadWriteTable<K, V> extends AsyncReadWriteTable<K, V> {
```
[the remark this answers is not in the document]
weisong44: See reply re sync/async.

PR #885, opened 2019-01-16
Samza 2081 : Type system for Samza SQL and Support for types in UDFS

srinipunuru:
This checkin adds 
1. Type system for Samza SQL. Previously Samza SQL was using Calcite's relational type system. We need an intermediate type system that is specific to Samza SQL so that we could support Beam SQL in future. This intermediate type system also allows to provide typing to Samza SQL UDFs.
2. Java annotations for Samza SQL that allows us to discover the Samza SQL UDFs easily and also provide users to configure name of the UDF, whether it is disabled. 
3. Initial support for adding types in Samza SQL UDFs. Right now we are not using these types for validations. Future checkin will add that capability.


samza-api/src/main/java/org/apache/samza/sql/schema/SqlFieldSchema.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+package org.apache.samza.sql.schema;
+
+/**
+ * Types of Samza Sql fields.
+ */
+public class SqlFieldSchema {
+
+  private SamzaSqlFieldType typeName;
```
shenodaguirguis: typeName --> fieldType?

samza-api/src/main/java/org/apache/samza/sql/schema/SqlFieldSchema.java
```diff
@@ -0,0 +1,76 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.sql.schema;
+
+/**
+ * Types of Samza Sql fields.
```
shenodaguirguis: comment not right

samza-api/src/main/java/org/apache/samza/sql/schema/SqlFieldSchema.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+public class SqlFieldSchema {
+
+  private SamzaSqlFieldType typeName;
+  private SqlFieldSchema elementType;
+  private SqlFieldSchema valueType;
```
shenodaguirguis: [element | value]Schema?

samza-api/src/main/java/org/apache/samza/sql/schema/SqlSchemaBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+
+/**
+ * Builder class to build the {@link SqlSchema}.
+ */
+public class SqlSchemaBuilder {
+  private List<String> names = new ArrayList<>();
+  private List<SqlFieldSchema> typeNames = new ArrayList<>();
```
shenodaguirguis: names --> fieldsNames
typeNames --> fieldsSchemas ?

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroTypeFactoryImpl.java
```diff
@@ -56,78 +52,69 @@ public RelDataType createType(Schema schema) {
...
-      String fieldName = field.name();
-      int fieldPos = field.pos() + 1;
-      RelDataType dataType = getRelDataType(field.schema());
-      relFields.add(new RelDataTypeFieldImpl(fieldName, fieldPos, dataType));
+      SqlFieldSchema fieldSchema = convertField(field.schema());
+      schemaBuilder.addField(field.name(), fieldSchema);
     }
 
-    return relFields;
+    return schemaBuilder.build();
```
shenodaguirguis: need to guard this that fields is not empty

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMessage.java
```diff
@@ -42,6 +46,10 @@
   // key could be a record in itself.
   private final Object key;
 
+  // Metadata associated with the message.
+  @JsonIgnore
+  private final Map<String, Object> metadata = new HashMap<>();
```
shenodaguirguis: I don't see where this is used/needed

samza-api/src/main/java/org/apache/samza/sql/schema/SamzaSqlFieldType.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.sql.schema;
```
weiqingy: The definition of SamzaSqlFieldType is duplicate with https://github.com/apache/samza/blob/e203db2aa29614c15ff0ee1593bf41c79b3b5c42/samza-sql-shell/src/main/java/org/apache/samza/sql/client/impl/SamzaSqlFieldType.java#L77

samza-api/src/main/java/org/apache/samza/sql/schema/SqlFieldSchema.java
```diff
@@ -0,0 +1,76 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.sql.schema;
+
+/**
+ * Types of Samza Sql fields.
+ */
+public class SqlFieldSchema {
```
weiqingy: The `SqlFieldSchema` here is duplicate with https://github.com/apache/samza/blob/e203db2aa29614c15ff0ee1593bf41c79b3b5c42/samza-sql-shell/src/main/java/org/apache/samza/sql/client/impl/SamzaSqlFieldType.java#L27. We need to remove one of them.

samza-api/src/main/java/org/apache/samza/sql/schema/SamzaSqlFieldType.java
weiqingy: Could we clean the code and remove one of them?

samza-api/src/main/java/org/apache/samza/sql/udfs/SamzaSqlUdf.java
```diff
@@ -0,0 +1,43 @@
...
+package org.apache.samza.sql.udfs;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+/**
+ * Java annotation to identity a Samza SQL Udf
```
weiqingy: I wonder if this PR also resolved the TODO described [here](https://github.com/apache/samza/blob/master/samza-sql-shell/src/main/java/org/apache/samza/sql/client/impl/SamzaExecutor.java#L323) ?

srinipunuru: Yes that's the idea. In my subsequent checkin. I will integrate this into the shell.

PR #886, opened 2019-01-16
SAMZA-2074: Use configuration stored in coordinator stream for CheckpointTool.

shanthoosh:
Currently, each run of a `CheckpointTool` requires the `TaskModel` from the `JobModel` for updating the checkpoints. `JobModel` generation involves reading the `SystemStreamMetadata`, `SystemStreamPartitionMetadata` of the input streams of the job. Post samza 1.0, this would require the entire configuration bag stored in coordinator stream. 

As a followup to `SAMZA-2059`, this PR changes the `CheckpointTool` to read the configuration from the coordinator stream and use it when generating the `JobModel`.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
+++ b/samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
@@ -167,20 +167,20 @@ class CheckpointTool(newOffsets: TaskNameToCheckpointMap, coordinatorStreamManag
     coordinatorStreamManager.register(getClass.getSimpleName)
     coordinatorStreamManager.start()
     coordinatorStreamManager.bootstrap()
-    val coordinatorStreamConfig: Config = coordinatorStreamManager.getConfig
+    val configFromCoordinatorStream: Config = coordinatorStreamManager.getConfig
 
     // Instantiate the checkpoint manager with coordinator stream configuration.
-    val checkpointManager: CheckpointManager = coordinatorStreamConfig.getCheckpointManagerFactory() match {
+    val checkpointManager: CheckpointManager = configFromCoordinatorStream.getCheckpointManagerFactory() match {
       case Some(className) =>
         Util.getObj(className, classOf[CheckpointManagerFactory])
-          .getCheckpointManager(coordinatorStreamConfig, new MetricsRegistryMap)
+          .getCheckpointManager(configFromCoordinatorStream, new MetricsRegistryMap)
       case _ =>
         throw new SamzaException("Configuration: task.checkpoint.factory is not defined.")
     }
 
     // Find all the TaskNames that would be generated for this job config
     val changelogManager = new ChangelogStreamManager(coordinatorStreamManager)
-    val jobModelManager = JobModelManager(coordinatorStreamConfig, changelogManager.readPartitionMapping())
+    val jobModelManager = JobModelManager(configFromCoordinatorStream, changelogManager.readPartitionMapping())
     val taskNames = jobModelManager
       .jobModel
       .getContainers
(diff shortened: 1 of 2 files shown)


shanthoosh:
@PawasChhokra @prateekm 
Please take a look when you have some time.

samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
```diff
@@ -152,25 +153,34 @@ object CheckpointTool {
...
 
-class CheckpointTool(config: Config, newOffsets: TaskNameToCheckpointMap, manager: CheckpointManager) extends Logging {
+class CheckpointTool(newOffsets: TaskNameToCheckpointMap, coordinatorStreamManager: CoordinatorStreamManager) extends Logging {
+
+  def run() {
+    // Read the configuration stored in the coordinator stream.
+    coordinatorStreamManager.register(getClass.getSimpleName)
+    coordinatorStreamManager.start()
+    coordinatorStreamManager.bootstrap()
+    val coordinatorStreamConfig: Config = coordinatorStreamManager.getConfig
```
cameronlee314: `coordinatorStreamConfig` seems to be used elsewhere to indicate the Config about a coordinator stream which is used to build a coordinator stream consumer/manager. This here seems to be the full job config loaded from the coordinator stream, so maybe use `configFromCoordinatorStream` to clarify that?

samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
```diff
@@ -152,25 +153,34 @@ object CheckpointTool {
...
 
-  def run {
-    info("Using %s" format manager)
+    // Instantiate the checkpoint manager with coordinator stream configuration.
+    val checkpointManager: CheckpointManager = coordinatorStreamConfig.getCheckpointManagerFactory() match {
+      case Some(className) =>
+        Util.getObj(className, classOf[CheckpointManagerFactory])
+          .getCheckpointManager(coordinatorStreamConfig, new MetricsRegistryMap)
```
cameronlee314: Is this being moved to `run` because `CheckpointManagerFactory` now needs the full Config to build a checkpoint manager?

samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
```diff
@@ -179,46 +189,37 @@ class CheckpointTool(config: Config, newOffsets: TaskNameToCheckpointMap, manage
...
       .toSet
 
-    taskNames.foreach(manager.register)
-    manager.start
+    taskNames.foreach(checkpointManager.register)
+    checkpointManager.start()
 
-    val lastCheckpoints = taskNames.map(tn => tn -> readLastCheckpoint(tn)).toMap
+    val lastCheckpoints = taskNames.map(taskName => {
+      taskName ->  Option(checkpointManager.readLastCheckpoint(taskName))
```
cameronlee314: Minor: check whitespace

samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
shanthoosh: Yes, to instantiate CheckpointManager with full config.

samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
shanthoosh: Sure. Applied the suggestion.

samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
shanthoosh: Done.

PR #888, opened 2019-01-18
SAMZA-2078: Add zookeeper session metrics in standalone

shanthoosh:


shanthoosh:
@sborya @prateekm 

samza-core/src/main/java/org/apache/samza/zk/ZkSessionMetrics.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package org.apache.samza.zk;
+
+import org.apache.samza.metrics.Counter;
+import org.apache.samza.metrics.MetricsBase;
+import org.apache.samza.metrics.MetricsRegistry;
+
+public class ZkSessionMetrics extends MetricsBase {
+
+  /**
```
vjagadish1989: consider wrapping these in ZkJobCoordinatorMetrics so that all metrics for monitoring Zk are in one place?

samza-core/src/main/java/org/apache/samza/zk/ZkSessionMetrics.java
```diff
@@ -0,0 +1,60 @@
...
+   * Denotes the number of times the zookeeper client session had expired.
+   */
+  public final Counter zkSessionExpirations;
+
+  /**
+   *  Denotes the number of times the zookeeper client session disconnects had occurred.
```
vjagadish1989: nit: remove "Denotes" .. consider removing comments in places where it's obvious from the name.

shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkSessionMetrics.java
shanthoosh: Unifying would keep all metrics in one place, but ZkSessionMetrics and ZkCoordinatorMetrics are different category of metrics. It would be logical to keep them separate. What do you think?

vjagadish1989: well, they are not entirely different because the coordinator handles Zk-session management, barrier lifecycle etc. but, it's fine by me if you prefer to keep them in 2 separate classes

PR #889, opened 2019-01-23
SEP-19: Allocator changes for standby-aware container allocation

rmatharu:
This PR makes the following changes: 

* Adds a map called standbyContainerConstraints which stores standby constraints for each container in the job model. The logic for populating the map using the job model is added as Util class.

* Adds a check before runStreamProcessor, where the standbyContainerConstraints is checked before launching, if the check succeeds container is launched. If standbyTasks are not enabled in config, existing behaviour is retained. 

* Adds logic to handle standbyContainerConstraints check failures, which entails releasing the resource, and making an any-host request for the container. 

* Adds logic in SamzaResourceRequest to order container request such that active-containers take precedence over standby, otherwise based on request-timestamp. 

* Adds metrics to capture FailedStandbyAllocations (due to standby constraints) and successful standby allocations.

Tested on dev setup and a test cluster.


rmatharu:
Closing it, in lieu of https://github.com/apache/samza/pull/903.

PR #890, opened 2019-01-23
Bugfix : Passing localityManager to JobModel when created from JobModelManager

rmatharu:
Samza Standalone required JobModel refactoring, due to which jobModelRef (and localityManager) is set explicitly in JobModelManager: https://github.com/apache/samza/pull/790.

However the JobModel instances with the localityManager (jobModelRef) was not being used when apply() was being called on the JobModelManager by the ClusterBasedJC.
Due to which host-locality was not being read, causing host affinity to break in YARN.


shanthoosh:
@rmatharu Thanks for the patch. Merged to trunk.

prateekm:
@shanthoosh I did not agree to get this change in. I recommended trying to extract the locality manager as a constructor / apply parameter and passing it in from the ClusterManagerJobCoordinator first.

@rmatharu can you give that a try? Let's fix cleanly if it's not too much work.

rmatharu:
Followup PR: https://github.com/apache/samza/pull/897/files

PR #892, opened 2019-01-24
Samza-2018: Using Rocksdb bulk load only for clean-slate restores

rmatharu:
Rocksdb has a problem that its bulk-load option cannot be used on existing stores seamlessly. 
See https://github.com/facebook/rocksdb/issues/2734

Because of this when host-affinity puts a container onto an old host, the restore path's bulk-load mode fails. 

To fix this, 
We use rocksdb's bulk load option only when the underlying storeDirectory does not exist. 
If the storeDirectory exists, then we open the store without bulk-load options. 

Note that, an existing storeDirectory may also be "stale", in which case we delete it, and should ideally be using the bulk load option. We will add this additional optimization separately.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
+++ b/samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
@@ -119,7 +119,7 @@ public class StorageManagerUtil {
    * Check if a store's disk footprint exists.
    *
    * @param storeDir the base directory of the store
-   * @return true if the storeDirectory exists, false otherwise
+   * @return true if a non-empty storeDir exists, false otherwise
    */
   public static boolean storeExists(File storeDir) {
     return storeDir.exists() && storeDir.list().length > 0;
--- a/samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
+++ b/samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
@@ -116,6 +116,16 @@ public class StorageManagerUtil {
   }
 
   /**
+   * Check if the storeDirectory exists.
+   *
+   * @param storeDir the base directory of the store
+   * @return true if the storeDirectory exists, false otherwise
+   */
+  public static boolean storeDirectoryExists(File storeDir) {
+    return storeDir.exists();
+  }
+
+  /**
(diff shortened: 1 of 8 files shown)


samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
```diff
@@ -115,6 +115,16 @@ public static boolean isOffsetFileValid(File storeDir, String offsetFileName) {
     return hasValidOffsetFile;
   }
 
+  /**
+   * Check if the storeDirectory exists.
```
mynameborat: I am not sure if the current check needs a separate method by itself; especially under storage util. 
We use this to determine only if the store exists. Maybe we can rename this method to `storeExists` and also check for non-zero file count for now? 

I am also fine with just the existence of the directory check. Although, i feel we can inline that as opposed to adding a new method in the storage util.

Thoughts?

rmatharu: Done, 
yup, makes sense to rename to storeExists and add non-zero file count.

samza-kv-rocksdb/src/main/java/org/apache/samza/storage/kv/RocksDbOptionsHelper.java
```diff
@@ -107,7 +107,11 @@ public static Options options(Config storeConfig, int numTasksForContainer, Stor
     options.setMaxLogFileSize(storeConfig.getLong(ROCKSDB_MAX_LOG_FILE_SIZE_BYTES, 64 * 1024 * 1024L));
     options.setKeepLogFileNum(storeConfig.getLong(ROCKSDB_KEEP_LOG_FILE_NUM, 2));
 
-    if(storeMode.equals(StorageEngineFactory.StoreMode.BulkLoad)) {
+    // use prepareForBulk load only when i. the store is being requested in BulkLoad mode
+    // and ii. the storeDirectory does not exist (fresh restore), because bulk load does not work seamlessly with
+    // bulk-load : https://github.com/facebook/rocksdb/issues/2734
```
jmakes:
```suggestion
    // existing stores : https://github.com/facebook/rocksdb/issues/2734
```

PR #893, opened 2019-01-24
SAMZA-2085 Add table serde generation

weisong44:
Currently the serde configuration generation is consolidated to JobNodeConfigurationGenerator, which runs during planning stage, it is not executed when tables are dynamically generated at earlier stage e.g. in config rewriter. In such situations, we would need to generated table serdes separately.


samza-core/src/main/java/org/apache/samza/table/TableConfigGenerator.java
```diff
@@ -19,26 +19,90 @@
...
     return tableConfig;
   }
 
+  /**
+   * Generate serde configuration for provided tables
+   *
+   * @param tableDescriptors table descriptors, for which serde configuration to be generated
+   * @return serde configuration for tables
+   */
+  static public Map<String, String> generateSerdeConfig(List<TableDescriptor> tableDescriptors) {
```
prateekm: Nitpick: public static

PR #897, opened 2019-01-26
SEP-19 : Refactoring to remove localityManager from JobModel

rmatharu:
This PR removes localityManager from JobModel. 
There was some localityMapping creation location in JobModel that was based on localityManager being null. 
That has, for now, been moved to JobModelManager. 
A subsequent PR will completely get rid of that null-based logic.


samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -365,6 +366,44 @@ class JobModelManager(
     }
   }
 
+  private def populateContainerLocalityMappings(): Unit = {
```
shanthoosh: 1. I think the agreed upon solution in offline-discussion with @prateekm  was to manage the lifecycle of `LocalityManager` in `ClusterBaseJobCoordinator` and inject it in `JobModelManager` when generating the `JobModel`. Why was that route not taken?  
2. There're test failures( `TestSamzaYarnAppMasterService`) in CI. Does this patch break any existing functionality or is it simply un-updated mocks/tests?

rmatharu: 1. The ContainerProcessManager.java needs a handle to the localityManager to query containers' hosts, when handling container failures. 
So moving the LocalityManager to JC means, the CPM needs to access the LM via the JC. The JC is the one that creates and manages the CPM, thus creating a circular dependency. 

2. I have to talk to someone who knows about the SamzaYarnAppMasterService to investigate why the test is failing. The code-path being tested doesnt seem to be used in production.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -365,6 +366,44 @@ class JobModelManager(
...
+    val allMappings = localityManager.readContainerLocality
+    import scala.collection.JavaConversions._
```
mynameborat: nit: remove?

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -365,6 +366,44 @@ class JobModelManager(
...
+    for (containerId <- jobModel.getContainers.keySet) {
+      if (allMappings.containsKey(containerId)) localityMappings.put(containerId, allMappings.get(containerId).get(SetContainerHostMapping.HOST_KEY))
+      else localityMappings.put(containerId, null)
+    }
+  }
+
+  def getAllContainerLocality: util.Map[String, String] = {
+    if (localityManager != null) populateContainerLocalityMappings
+    else if(localityMappings.isEmpty) jobModel.getContainers.keys.foreach(containerId => localityMappings.put(containerId, null))
```
mynameborat: would returning an empty map be sufficient here?

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -365,6 +366,44 @@ class JobModelManager(
...
+    if (mappings == null) return EMPTY_STRING
+    if (!mappings.containsKey(key)) return EMPTY_STRING
+    mappings.get(key)
+  }
+
+  def getAllContainerToHostValues(key: String): util.Map[String, String] = {
+    if (localityManager == null)
+      return new util.HashMap[String, String]()
+    val allValues = new util.HashMap[String, String]
+    import scala.collection.JavaConversions._
```
mynameborat: is this intended? I noticed imports in the middle somewhere else too.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -365,6 +366,44 @@ class JobModelManager(
...
+    for (entry <- localityManager.readContainerLocality.entrySet) {
```
mynameborat: Can we rename these to meaningful ones? It is hard to know what entry, key, value(key inside value) corresponds to?

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
shanthoosh: > Circular dependency
1.  If I understand right, ClusterBasedJobCoordinator is at one level higher than both the JobModelManager, ContainerProcessManager.  So how does managing the LocalityManager lifecycle above both  the JobModelManager, ContainerProcessManager in ClusterBasedJC and injecting it down during instantiation/jobModelGeneration would create circular dependency? Can you shed some light.
2. Does both these locality methods logically belong in JobModelManager? It would make the refactoring easier, but we have a  full blown out LocalityManager just to manage locality. Moving these locality computation outside of it breaks the ownership boundaries. 

> SamzaYarnAppMasterService:

Would be great to fix the existing  tests and test this end to end just to ensure that we're not breaking anything. Can you  share some light on the integration tests we did for this change?

>  code-path being tested doesnt seem to be used in production.

Are you really sure of that? AFAIK, it sets up ApplicationMasterWebServlet to list the config, appState, rmHttpAddress etc. It might not be a good idea to unnecessarily remove an endpoint as a part of this change.

samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
```diff
@@ -39,34 +36,14 @@
  * </p>
  */
 public class JobModel {
-  private static final String EMPTY_STRING = "";
   private final Config config;
   private final Map<String, ContainerModel> containers;
-
-  private final LocalityManager localityManager;
-  private final Map<String, String> localityMappings;
```
cameronlee314: What do you think about leaving this in here (but make it fully immutable)? Would we be able to always read the localityMappings before creating a JobModel?
It seems like the localityMappings might be a conceptual fit in the JobModel. I kind of see it a bit similarly to the container-to-task mappings which are in the ContainerModel.
Other historical info: One thing that we considered when doing the Context API changes was to add JobModel to the JobContext, but that required putting JobModel in samza-api, and that didn't work due to LocalityManager still being there. However, now that LocalityManager is being removed, maybe we could move JobModel. It might be useful to provide access to the locality mappings in the Context. On the other hand, we could just leave it out for now and then add it later.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -355,6 +355,7 @@ class JobModelManager(
    */
   val localityManager: LocalityManager = null) extends Logging {
 
+  val localityMappings : Map[String, String] = scala.collection.mutable.Map[String, String]()
```
cameronlee314: Does this need to be mutable? Can we read the locality mappings up front and then pass them directly to the JobModelManager?

prateekm:
@rmatharu Let's get this checked in once you have some spare time.

prateekm:
@rmatharu We also want to move the JobModel to samza-api (where ContainerModel and TaskModel already are).

cameronlee314:
@rmatharu do you still plan to check this in?

rmatharu:
yes, most of the comments are addressed but i need to rebase on master and send a revision.

xinyuiscool:
@rmatharu : could you please create a JIRA ticket for this change in samza open source and associate the pr with the jira (setting the subject SAMZA-XXXX:...)? It would be much easier for the release person to figure out what got into the release with a JIRA ticket. SEP shouldn't be used as the PR titile.

Please also read "Contribute using Pull Requests" in Samza contributor corner (https://cwiki.apache.org/confluence/display/SAMZA/Contributor%27s+Corner) to get familiar with expected Samza contributor flow. Thanks.

cameronlee314:
@rmatharu : Could we please get this rebased so it can be checked in?

mynameborat:
This refactor along with moving JobModel to samza-api are committed as part of #1421 . Closing this PR

PR #899, opened 2019-01-31
 SAMZA-2091: Update ZkClient in samza-standalone to use string serializer

shanthoosh:

1. Switch from using SerializableSerializer to ZkStringSerializer in the zkClient for standalone.
  ZkClient uses pluggable serde to serialize/deserialize user-defined types to bytes before it's written to zookeeper. In the current implementation, the predefined SerializableSerializer(which uses the java native serialization) provided as default by the ZkClient library is used by samza-standalone. SerializableSerializer augments special characters  on top of the serialized data as auxiliary data. The presence of these special characters makes the parsing of zookeeper data impossible  when building custom tools in languages(such as python) that does not have java native serializer. To solve this problem, this patch introduces string serializer to serialize/deserialize data before it's written to zookeeper.
2. Adds a new string serializer which implements the ZkSerializer serialization abstraction provided by the ZkClient.
3. Add protocol version to the zookeeper base path  to enable us to make data format changes easily in the future. 
4. Fix the usages where there're direct type conversion are done.

**Testing:**

1. Added unit tests for the newly introduced serializer.
2. Tested this patch  with the sample standalone test jobs from the samza-hello-samza repository.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
@@ -92,10 +92,6 @@ public class ZkBarrierForVersionUpgrade {
       this.str = str;
     }
 
-    public String getValue() {
-      return str;
-    }
-
     @Override
     public String toString() {
       return str;
@@ -129,7 +125,7 @@ public class ZkBarrierForVersionUpgrade {
         barrierParticipantsPath,
         barrierStatePath});
     LOG.info("Marking the barrier state: {} as {}.", barrierStatePath, State.NEW);
-    zkUtils.writeData(barrierStatePath, State.NEW.getValue());
+    zkUtils.writeData(barrierStatePath, State.NEW.toString());
 
     LOG.info("Subscribing child changes on the path: {} for barrier version: {}.", barrierParticipantsPath, version);
     zkUtils.subscribeChildChanges(barrierParticipantsPath, new ZkBarrierChangeHandler(version, participants, zkUtils));
@@ -164,7 +160,7 @@ public class ZkBarrierForVersionUpgrade {
     State barrierState = State.valueOf(zkUtils.getZkClient().readData(barrierStatePath));
     if (Objects.equals(barrierState, State.NEW)) {
       LOG.info(String.format("Expiring the barrier version: %s. Marking the barrier state: %s as %s.", version, barrierStatePath, State.TIMED_OUT));
-      zkUtils.writeData(keyBuilder.getBarrierStatePath(version), State.TIMED_OUT.getValue());
+      zkUtils.writeData(keyBuilder.getBarrierStatePath(version), State.TIMED_OUT.toString());
     } else {
       LOG.debug(String.format("Barrier version: %s is at: %s state. Not marking barrier as %s.", version, barrierState, State.TIMED_OUT));
     }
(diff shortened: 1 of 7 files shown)


shanthoosh:
@xinyuiscool @sborya 

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -155,10 +161,10 @@ public void join(String version, String participantId) {
    */
   public void expire(String version) {
     String barrierStatePath = keyBuilder.getBarrierStatePath(version);
-    State barrierState = zkUtils.getZkClient().readData(barrierStatePath);
+    State barrierState = State.valueOf(zkUtils.getZkClient().readData(barrierStatePath));
```
vjagadish1989: Are there compatibility concerns? eg: when data written using serde format is read using the string serializer? 
May not be an issue if we always do a write before we do a read. 

Either-ways, would be good to understand the behavior wrt. rolling upgrades

shanthoosh: In addition to switching to string serializer, I've also changed the zookeeper base path that is used to store all the coordination data in zookeeper(Refer protocol version comment).  When an application picks-up this change, the old state will not be considered for `JobModel` generation. We will not de-serialize the old state with the new serializer. 

> Compatibility concerns:

Currently there are only stateless standalone jobs in production. So the previous state is not relevant and not useful for the task redistribution on processor shuffle. 

> Rolling upgrade. 

For rolling upgrade scenario, the newer processors will be storing the coordination data to new zookeeper namespace and the older processors will be storing the coordination data to old zookeeper namespace. So there will be no serialization exceptions. Though, there could be a brief rolling-bounce window, where in some processors could be doing duplicate stream processing. I think we can document these situations in the release wiki and guide users. 

Despite this notable problem, I think this patch is the way forward. It will help in viewing the standalone zookeeper state with tools in LinkedIn and open source. Will allow us to build tools in non-java languages for displaying standalone-state(Building dashboard for standlaone).

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
```diff
@@ -82,14 +82,20 @@
...
+    TIMED_OUT("TIMED_OUT"),
+    DONE("DONE");
 
     private String str;
 
     State(String str) {
       this.str = str;
     }
 
+    public String getValue() {
```
sborya: you can just use toString()

shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/zk/ZkBarrierForVersionUpgrade.java
vjagadish1989: FYI, I don't think anyone disagrees that this patch is the way forward..

prateekm:
@shanthoosh Please merge if you don't need anything else for this PR.

prateekm:
@shanthoosh Please merge if you don't need anything else for this PR.

shanthoosh:
@prateekm Apologies for the delay in merging this in. I was side-tracked with save-points and standalone -prod issues with on-boarding new user.

PR #901, opened 2019-02-01
SAMZA-2094: Implement the StartpointVisitor for KafkaSystemConsumer.

shanthoosh:


shanthoosh:
@cameronlee314 
Can you take a look when you have a chance.

cameronlee314:
Please take a look at the comment https://github.com/apache/samza/pull/869#discussion_r246982218 regarding not having KafkaSystemConsumer directly implement StartpointVisitor.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -330,6 +341,57 @@ public String getSystemName() {
     return systemName;
   }
 
+  @VisibleForTesting
+  static class KafkaStartpointVisitor implements StartpointVisitor {
```
cameronlee314: Could you please change the name to indicate it is responsible for startpoint registration? "StartpointVisitor" can be anything that needs to act on a `Startpoint`, so just using `KafkaStartpointVisitor` isn't specific enough.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -60,6 +70,7 @@
   private final Config config;
   private final boolean fetchThresholdBytesEnabled;
   private final KafkaSystemConsumerMetrics metrics;
+  private final KafkaStartpointVisitor kafkaStartpointVisitor;
```
cameronlee314: It doesn't look like this is getting used anywhere.
Maybe at least add a public API for this class, even if you don't wire it in yet.
Ideally, could you put out a PR which shows this being used? That other PR can be separate from this one, but it would be good to see how it gets used.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -330,6 +341,57 @@ public String getSystemName() {
...
+    }
+
+    @Override
+    public void visit(SystemStreamPartition systemStreamPartition, StartpointTimestamp startpointTimestamp) {
+      Long timestampInStartpoint = startpointTimestamp.getTimestampOffset();
+      TopicPartition topicPartition = toTopicPartition(systemStreamPartition);
+      Map<TopicPartition, Long> topicPartitionsToTimeStamps = ImmutableMap.of(topicPartition, timestampInStartpoint);
+
+      // Look up the offset by timestamp.
+      LOG.info("Looking up the offsets of the topic partition: {} by timestamp: {}.",  topicPartition, timestampInStartpoint);
```
cameronlee314: Minor: watch double whitespace

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -330,6 +341,57 @@ public String getSystemName() {
...
+      Map<TopicPartition, OffsetAndTimestamp> topicPartitionToOffsetTimestamps = kafkaConsumer.offsetsForTimes(topicPartitionsToTimeStamps);
+      OffsetAndTimestamp offsetAndTimeStamp = topicPartitionToOffsetTimestamps.get(topicPartition);
```
cameronlee314: Is there any case where there is a null entry in the map for `topicPartition`?

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -205,6 +215,125 @@ public void testFetchThresholdBytesDiabled() {
...
+    final KafkaConsumer consumer = Mockito.mock(KafkaConsumer.class);
+    KafkaStartpointVisitor kafkaStartpointVisitor = new KafkaStartpointVisitor(consumer);
+
+    final StartpointSpecific testStartpointSpecific = new StartpointSpecific(offset);
+
+    // Mock the consumer interactions.
+    Mockito.doNothing().when(consumer).seek(testTopicPartition, Long.valueOf(offset));
+
+    // Invoke the consumer with startpoint.
+    kafkaStartpointVisitor.visit(testSystemStreamPartition, testStartpointSpecific);
```
cameronlee314: If there is a public API for this functionality, it might be better to call that directly in the unit test.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -205,6 +215,125 @@ public void testFetchThresholdBytesDiabled() {
...
+
+    // Mock verifications.
+    Mockito.verify(consumer).seekToEnd(ImmutableList.of(testTopicPartition));
+
+  }
+
+  private MockKafkaSystemConsumer createConsumerWithMock(final KafkaConsumer kafkaConsumer) {
```
cameronlee314: Is this used?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
cameronlee314: Even though we currently won't be able to trigger this flow, it's not obvious that there could be a problem in the future if someone changes some things around. Could you please at least document in the public API that it needs to be called with valid SSPs?

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
cameronlee314: 1. In general, we should try to test public APIs as much as possible instead of testing pieces of implementation. The production usage of our code is through public APIs, so it is best to test those directly. Sometimes, it is impractical to test public APIs, but in this case, it doesn't really seem any harder to test them directly. In addition, if someone adds logic in the future outside of the visitor, then the current tests won't catch that code.
2. Let's see what that other PR looks like, and then maybe we can update this PR based on that. This PR could just be committed after your other PR gets committed.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
shanthoosh: Sure. Added the documentation.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
shanthoosh: 1. Of-course, the plan is to test the public `SystemConsumer#register` API as well. But while testing it, I would just verify if the visitor mocks were invoked with right parameters. 
2. Yes, will send out the PR for it shortly.

cameronlee314: It doesn't seem necessary to have unit tests for both the public `register` API and the internal visitor methods, since those tests are going to cover pretty much the same thing.

PR #903, opened 2019-02-02
SEP-19: Allocator changes for standby-aware container allocation, and active container failover

rmatharu:
This PR makes the following changes: 

* Adds a map called standbyContainerConstraints which stores standby constraints for each container in the job model. The logic for populating the map using the job model is added as Util class.

* Adds a check before runStreamProcessor, where the standbyContainerConstraints is checked before launching, if the check succeeds container is launched. If standbyTasks are not enabled in config, existing behaviour is retained. 

* Adds logic to handle standbyContainerConstraints check failures, which entails releasing the resource, and making an any-host request for the container. 

* Adds logic in SamzaResourceRequest to order container request such that active-containers take precedence over standby, otherwise based on request-timestamp.

* Adds logic in HostAwareContainerAllocator to trap resourceRequests issued by the CPM and the HACA, and 
a. if it is an any-host for an active container,  translate it for a stop a standby. 
b. if it is for a standby stopped by us, translate it for a resource request for active on standby-host and standby on anyhost. 
c. in all other cases, proceeds asis.

* Adds metrics to capture FailedStandbyAllocations (due to standby constraints), successful standby allocations, number of failovers of active to a valid standby, number of failovers of active to anyhost, in case no standby was found, and number of standby-container-stops completed.

Tested on dev setup, a VPC, and a test cluster.


samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java
```diff
@@ -119,4 +153,219 @@ private void updateExpiryMetrics(SamzaResourceRequest request) {
...
+  }
+
+  /**
+   * Intercept resource requests, which are due to either a launch-failure or resource-request expired or standby
+   * 1. a standby container, we proceed to make a anyhost request
+   * 2. an activeContainer, we try to fail-it-over to a standby
+   * @param containerID Identifier of the container that will be run when a resource is allocated
+   */
+  @Override
+  public void requestResourceDueToLaunchFailOrExpiredRequest(String containerID) {
```
vjagadish1989: shouldn't have to override this method  (or expose this method in the parent class). Instead it's cleaner to expose the most general API and special-case them when they're used

samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java
```diff
@@ -119,4 +153,219 @@ private void updateExpiryMetrics(SamzaResourceRequest request) {
...
+      requestResource(containerID, ResourceRequestState.ANY_HOST); // invoke local method & select a new standby if possible
+    }
+  }
+
+  // Intercept issuing of resource requests from the CPM
+  // 1. for ActiveContainers and instead choose a StandbyContainer to stop
+  // 2. for a StandbyContainer (after it has been chosen for failover), to put the active on the standby's host
+  // and request another resource for the standby
+  // 3. for a standbyContainer (if not for a failover)
+  @Override
```
vjagadish1989: would be preferable to not override `requestResource`..Additionally, `requestResource` should not perform any actions related to failover handling.

samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java
```diff
@@ -119,4 +153,219 @@ private void updateExpiryMetrics(SamzaResourceRequest request) {
...
+    if (!new JobConfig(config).getStandbyTasksEnabled()) {
+      super.requestResource(containerID, preferredHost);
+      return;
+    }
+
+    // If its an anyhost request for an active container, then we select a standby container to stop and place this activeContainer on that standby's host
+    // we may have already chosen a standby (which didnt work for a failover)
+    if (!StandbyTaskUtil.isStandbyContainer(containerID) && preferredHost.equals(ResourceRequestState.ANY_HOST)) {
+      initiateActiveContainerFailover(containerID);
+    } else if (StandbyTaskUtil.isStandbyContainer(containerID)) {
```
vjagadish1989: move failover handling logic up-stream - maybe to the `ContainerProcessManager`

samza-core/src/main/java/org/apache/samza/clustermanager/SamzaApplicationState.java
```diff
@@ -141,6 +143,36 @@
...
+  */
+  public final AtomicInteger failoversToAnyHost = new AtomicInteger(0);
+
+  /**
+   * Number of stops of standby containers that completed.
+   */
+  public final AtomicInteger standbyStopsComplete = new AtomicInteger(0);
+
+  // Map of active containers that are in failover, indexed by the active container's resourceID (at the time of failure)
+  public final ConcurrentMap<String, ContainerFailoverState> failovers = new ConcurrentHashMap<String, ContainerFailoverState>(0);
```
vjagadish1989: move state that pertains to failover to a new class.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterResourceManager.java
```diff
@@ -122,6 +122,10 @@ public ClusterResourceManager(Callback callback) {
   public abstract void launchStreamProcessor(SamzaResource resource, CommandBuilder builder);
 
 
+  // Requests the stopping of a StreamProcessor, identified by the given resource
+  public abstract void stopStreamProcessor(SamzaResource resource);
+
+
   public abstract void stop(SamzaApplicationState.SamzaAppStatus status);
```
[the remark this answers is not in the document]
rmatharu: done

samza-core/src/main/java/org/apache/samza/clustermanager/SamzaApplicationState.java
```diff
@@ -141,6 +143,36 @@
    */
   public final AtomicInteger redundantNotifications = new AtomicInteger(0);
 
+  /**
+   * Number of container allocations that proceeded because they met standby container constraints.
```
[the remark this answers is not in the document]
rmatharu: done

samza-core/src/main/java/org/apache/samza/clustermanager/AbstractContainerAllocator.java
```diff
@@ -195,14 +195,25 @@ protected final SamzaResourceRequest peekPendingRequest() {
...
   }
 
+  /**
+   *  Method to handle re-requesting of resource for a given container, in cases where
+   *  a prior resource request for the container, expired or the launch failed (due to a startup-fail, or
+   *  standby constraints not being met).
+   * @param containerID Identifier of the container that will be run when a resource is allocated
+   *
+   */
+  public void requestResourceDueToLaunchFailOrExpiredRequest(String containerID) {
```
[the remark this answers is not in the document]
rmatharu: removed

samza-core/src/main/java/org/apache/samza/clustermanager/SamzaApplicationState.java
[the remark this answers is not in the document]
rmatharu: removed

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerFailoverState.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Encapsulates metadata concerning the failover of an active container.
+ */
+public class ContainerFailoverState {
+  public final String activeContainerID;
+  public final String activeContainerResourceID;
+
+  // Map of samza-container-resource ID to host, for each standby container selected for failover
+  public final Map<String, String> selectedStandbyContainers;
```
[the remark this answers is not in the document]
rmatharu: done

samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java
rmatharu: done, this method is removed.

samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java
rmatharu: yup, done

samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java
rmatharu: moved to ContainerProcessManager and StandbyContainerManager

samza-core/src/main/java/org/apache/samza/clustermanager/SamzaApplicationState.java
rmatharu: Moved to ContainerFailoverState class

rmatharu:
Addressed all comments. Will pursue the yarn-client-upgrade separately.

samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java
```diff
@@ -0,0 +1,466 @@
...
+    } else {
+      log.info("Issuing request for standby container {} on host {}, since this is not for a failover",
+          standbyContainerID, preferredHost);
+      containerAllocator.requestResource(standbyContainerID, preferredHost);
+      return;
+    }
+  }
+
+  /** Method to handle failover for a container.
+   *  If it is an active container,
```
vjagadish1989: perhaps rename method too? since the function includes "active" in its name

samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java
```diff
@@ -0,0 +1,466 @@
...
+
+    // Iterate over the list of running standby containers, to find a standby resource that we have not already
+    // used for a failover for this active resoruce
+    for (String standbyContainerID : this.standbyContainerConstraints.get(activeContainerID)) {
+
+      if (samzaApplicationState.runningContainers.containsKey(standbyContainerID)) {
+        SamzaResource standbyContainerResource = samzaApplicationState.runningContainers.get(standbyContainerID);
+
+        // use this standby if there was no previous failover or if this standbyResource was not used for it
+        if (!failoverMetadata.isPresent() || !failoverMetadata.get()
```
vjagadish1989: maybe negate this condition and use '&&" instead?

"isPresent && isUsed()"

samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java
```diff
@@ -0,0 +1,466 @@
...
+    // obtain any existing failover metadata
+    Optional<StandbyContainerManager.FailoverMetadata> failoverMetadata =
+        activeContainerResourceID == null ? Optional.empty() : this.getFailoverMetadata(activeContainerResourceID);
+
+    // Iterate over the list of running standby containers, to find a standby resource that we have not already
+    // used for a failover for this active resoruce
+    for (String standbyContainerID : this.standbyContainerConstraints.get(activeContainerID)) {
+
+      if (samzaApplicationState.runningContainers.containsKey(standbyContainerID)) {
+        SamzaResource standbyContainerResource = samzaApplicationState.runningContainers.get(standbyContainerID);
```
vjagadish1989: Can the state class expose this API?

SamzaResource getActiveContainerFor(SamzaResource r1). a nice thing is that you don't have to keep extracting specific fields and doing lookups

samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java
```diff
@@ -0,0 +1,444 @@
...
+  /** Method to handle failover for a container.
+   *  If it is an active container,
+   *  We try to find a standby for the active container, and issue a stop on it.
+   *  If we do not find a standby container, we simply issue an anyhost request to place it.
+   *  If it is standby container, we simply forward the request to the containerAllocator
+   *
+   * @param containerID the samzaContainerID of the active-container
+   * @param resourceID  the samza-resource-ID of the container when it failed (used to index failover-state)
+   */
+  private void initiateActiveContainerFailover(String containerID, String resourceID,
```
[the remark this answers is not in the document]
rmatharu: Sorry the doc for this method was incorrect. 
This method only handles activeContainerIDs now, so the initiateActiveContainerFailover is accurate.
The "standby" case has been inlined.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -484,5 +497,13 @@ private String getPendingContainerId(String resourceId) {
     return null;
   }
 
+  private void handleContainerStop(String containerID, String resourceID, String preferredHost, int exitStatus) {
+    if (standbyContainerManager.isPresent()) {
+      standbyContainerManager.get().handleContainerStop(containerID, resourceID, preferredHost, exitStatus, containerAllocator);
```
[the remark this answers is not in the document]
rmatharu: The HACA needs/has a handle on standbyContainerManager (e.g., to handle expired requests),

samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java
```diff
@@ -0,0 +1,444 @@
...
+  }
+
+  /**
+   * Method to select a standby container for a given active container that has stopped.
+   * TODO: enrich this method to select standby's intelligently based on lag, timestamp, load-balencing, etc.
+   * @param activeContainerID Samza containerID of the active container
+   * @param activeContainerResourceID ResourceID of the active container at the time of its last failure
+   * @return
+   */
+  private Optional<Entry<String, SamzaResource>> selectStandby(String activeContainerID,
```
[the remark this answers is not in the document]
rmatharu: FailoverMetadata currently only has metadata on any failovers in progress. 
but selecting a standby also requires access to samzaApplicationState. 
I'll keep this as is for now.

samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java
```diff
@@ -0,0 +1,466 @@
...
+      String activeContainerResourceID) {
+
+    log.info("Standby containers {} for active container {}", this.standbyContainerConstraints.get(activeContainerID), activeContainerID);
```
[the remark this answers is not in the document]
rmatharu: But the same map also contains standyContainerID to active and other standby containerIDs.
Its a map of samza containerIDs to their corresponding active and standby containers, e.g., 0 -> {0-0, 0-1}, 0-0 -> {0, 0-1}
This is used for checking no two standbys or active-standby-pair are started on the same host

samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java
```diff
@@ -81,13 +87,27 @@ public void assignResourceRequests()  {
 
         if (expired) {
           updateExpiryMetrics(request);
-          if (resourceAvailableOnAnyHost) {
-            log.info("Request for container: {} on {} has expired. Running on ANY_HOST", request.getContainerID(), request.getPreferredHost());
-            runStreamProcessor(request, ResourceRequestState.ANY_HOST);
+
+          if (standbyContainerManager.isPresent()) {
```
vjagadish1989: what if standby is not enabled? the current behavior is to issue any_host request. is that changed?

samza-core/src/main/java/org/apache/samza/clustermanager/SamzaApplicationState.java
```diff
@@ -141,6 +142,28 @@
...
+  public final AtomicInteger successfulStandbyAllocations = new AtomicInteger(0);
+
+  /**
+   * Number of container allocations that were dis-allowed because they
+   * did not meet standby container constraints.
+   */
+  public final AtomicInteger failedStandbyAllocations = new AtomicInteger(0);
+
+  /**
+   * Number of active container standbyContainerState initiated in which a standby container was found.
```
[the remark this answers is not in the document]
rmatharu: Added docs to explain the difference.

samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java
```diff
@@ -81,13 +87,27 @@ public void assignResourceRequests()  {
...
+
+            // if standby is enabled and an alternative-anyhost-resource is available, we try to use it
+            if (resourceAvailableOnAnyHost) {
+              standbyContainerManager.get().handleExpiredResourceRequest(containerID, request, Optional.of(peekAllocatedResource(ResourceRequestState.ANY_HOST)), this, resourceRequestState);
+            } else {
+              standbyContainerManager.get().handleExpiredResourceRequest(containerID, request, Optional.empty(), this, resourceRequestState);
```
vjagadish1989: why invoke this with an empty resource? what state are we tracking?

samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java
```diff
@@ -81,13 +87,27 @@ public void assignResourceRequests()  {
...
+            }
+
           } else {
-            log.info("Request for container: {} on {} has expired. Requesting additional resources on ANY_HOST.", request.getContainerID(), request.getPreferredHost());
-            resourceRequestState.cancelResourceRequest(request);
-            requestResource(containerID, ResourceRequestState.ANY_HOST);
+
+            if (resourceAvailableOnAnyHost) {
```
vjagadish1989: prefer consolidating resourceAvailableOnAnyHost checks instead of having them checked at multiple places

samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java
rmatharu: Current behavior is unchanged.
This if is only triggerred if (standbyContainerManager.isPresent())
otherwise it falls into the if below, 
```else {

            if (resourceAvailableOnAnyHost) {
              log.info("Request for container: {} on {} has expired. Running on ANY_HOST", request.getContainerID(), request.getPreferredHost());
              runStreamProcessor(request, ResourceRequestState.ANY_HOST);
            } else {
              log.info("Request for container: {} on {} has expired. Requesting additional resources on ANY_HOST.", request.getContainerID(), request.getPreferredHost());
              resourceRequestState.cancelResourceRequest(request);
              requestResource(containerID, ResourceRequestState.ANY_HOST);
            }

          }```

File: samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java (hunk 18 above)
Comment (rmatharu) [reply samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java]: Because the StandbyContainerManager has a single method to handle expired requests, if there is no alternate resource on anyHost available, we invoke it with an empty optional resource.

File: samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java (hunk 19 above)
Comment (rmatharu) [reply samza-core/src/main/java/org/apache/samza/clustermanager/HostAwareContainerAllocator.java]: This separation keeps the path taken in case standbyContainerManager is and isnt present clear.

File: samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java [hunk 20]
```diff
@@ -0,0 +1,459 @@
...
+      // A standby container can be started on the anyhost-alternative-resource rightaway provided it passes all the
+      // standby constraints
+
+      log.info("Handling expired request, standby container {} can be started on alternative resource {}", containerID, alternativeResource.get());
+
+      checkStandbyConstraintsAndRunStreamProcessor(request, ResourceRequestState.ANY_HOST, alternativeResource.get(),
+          containerAllocator, resourceRequestState);
+
+    } else if (StandbyTaskUtil.isStandbyContainer(containerID) && !alternativeResource.isPresent()) {
+      // If there is no alternative-resource for the standby container we make a new anyhost request
```
vjagadish1989: The logic of checking for empty resources feels a bit mis-placed in this class.

it introduces multiple multiple branches in this class and in the HostAwareAllocator as well :)  is there any room to reduce the complexity?

rmatharu: Let me try separating it out by container-type (standby v. active) : handleExpiredRequestForActive and handleExpiredRequestForStandby. 
That would help i think.

This overall method is basically saying a resource request has expired, and optionally an alternate anyHost resource is available.

PR #904, opened 2019-02-04
SAMZA-2096: Support getVersion in Sql Shell

weiqingy:
1. Print version information when Shell is just initialized. 
2. Support "VERSION" command.

## How was this patch tested?
Test in Shell.


samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliShell.java
```diff
@@ -242,6 +248,18 @@ void open(String message) {
     }
   }
 
+  private void commandVersion(CliCommand command) {
+    printVersion();
+  }
+
+  private void printVersion() {
+    String version = String.format("Shell version %s, Executor is %s, version %s",
```
atoomula: I'm still wondering if we need two different versions ? Let's talk offline.

atoomula:
LGTM

PR #906, opened 2019-02-05
SAMZA-2097: Allow table parts configuration generation

weisong44:
Sometimes a table part may use some 3rd party libraries that require certain configuration, these can't be handled from within a table descriptor itself. We should allow table parts to participate in the configuration generation in toConfig() method.


samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -243,29 +239,52 @@ public String getProviderFactoryClassName() {
...
+    // Handle write retry policy
     if (writeRetryPolicy != null) {
       addTableConfig(WRITE_RETRY_POLICY, SerdeUtils.serialize("write retry policy", writeRetryPolicy), tableConfig);
+      tableConfig.putAll(writeRetryPolicy.toConfig(jobConfig, new MapConfig(tableConfig)));
     }
 
     addTableConfig(ASYNC_CALLBACK_POOL_SIZE, String.valueOf(asyncCallbackPoolSize), tableConfig);
 
+    // Handle table reader function
+    addTableConfig(READ_FN, SerdeUtils.serialize("read function", readFn), tableConfig);
```
xinyuiscool: Do we need a null check for readFn?

weisong44: No, there is a check to ensure it's not null.

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -243,29 +239,52 @@ public String getProviderFactoryClassName() {
         throw new SamzaException("Failed to create default rate limiter", ex);
       }
       addTableConfig(RATE_LIMITER, SerdeUtils.serialize("rate limiter", defaultRateLimiter), tableConfig);
+      if (defaultRateLimiter instanceof TablePart) {
+        tableConfig.putAll(((TablePart) defaultRateLimiter).toConfig(jobConfig, new MapConfig(tableConfig)));
```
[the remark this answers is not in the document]
weisong44: Sure, make sense. Added a new method to simplify the logic.

PR #908, opened 2019-02-06
SAMZA-2083: Adds the register(SystemStreamPartition, startpoint) API to the SystemConsumer.

shanthoosh:
This is a followup PR to address the comments from #869. It is comprised of the following changes:
1. SAMZA-1985 introduced startpoint abstraction, a comprehensive mechanism to define starting positions for a input SystemStreamPartition in samza. This patch merely adds the register(SystemStreamPartition, Startpoint) API to SystemConsumer. 
2. Removes the default metadata store defined for startpoints. This is helpful in not creating the unnecessary metastore connections as a part of container initialization before the startpoint feature is fully completed.

Follow-ups:

1. In a follow-up PR,  new register API and start-point visitors will be implemented for different I/O connectors.
2. Usages of SystemConsumer#register(SystemStreamPartition, offset) is very wide-spread in samza codebase. After the above step is completed, will deprecate SystemConsumer#register(SystemStreamPartition, offset) after moving all it's usages to SystemConsumer#register(SystemStreamPartition, StartpointSpecific).

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/system/SystemConsumer.java
+++ b/samza-api/src/main/java/org/apache/samza/system/SystemConsumer.java
@@ -22,6 +22,8 @@ package org.apache.samza.system;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import org.apache.samza.annotation.InterfaceStability;
+import org.apache.samza.startpoint.Startpoint;
 
 /**
  * <p>
@@ -29,7 +31,7 @@ import java.util.Set;
  * wishes to integrate with Samza. Examples of systems that one might want to
  * integrate would be systems like Kafka, Hadoop, Kestrel, SQS, etc.
  * </p>
- * 
+ *
  * <p>
  * SamzaContainer uses SystemConsumer to read messages from the underlying
  * system, and funnels the messages to the appropriate StreamTask instances. The
@@ -38,11 +40,11 @@ import java.util.Set;
  * repeat. If no IncomingMessageEnvelopes are returned, the SamzaContainer polls
  * again, but with a timeout of 10ms.
  * </p>
- * 
+ *
  * <p>
  * The SamzaContainer treats SystemConsumer in the following way:
  * </p>
- * 
+ *
  * <ul>
  * <li>Start will be called before stop.</li>
  * <li>Register will be called one or more times before start.</li>
@@ -63,17 +65,17 @@ import java.util.Set;
  * <li>Any exception thrown by the SystemConsumer means that the SamzaContainer
  * should halt.</li>
  * </ul>
- * 
+ *
  * <p>
  * There are generally three implementation styles to this interface:
  * </p>
- * 
+ *
  * <ol>
  * <li>Thread-based</li>
  * <li>Selector-based</li>
  * <li>Synchronous</li>
  * </ol>
- * 
+ *
  * <p>
  * Thread-based implementations typically use a series of threads to read from
  * an underlying system asynchronously, and put the resulting messages into a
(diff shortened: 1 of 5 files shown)


shanthoosh:
@cameronlee314 This patch addresses your comment here: [#869 (comment)](https://github.com/apache/samza/pull/869#discussion_r246982218)

samza-api/src/main/java/org/apache/samza/system/SystemConsumer.java
```diff
@@ -136,18 +138,31 @@
...
+   * Registers the {@link Startpoint} to the SystemConsumer. SystemConsumer
+   * should read the messages from all the registered SystemStreamPartitions.
+   * SystemStreamPartitions should be registered before the start is called.
+   *
+   * @param systemStreamPartition represents the SystemStreamPartition to be registered.
+   * @param startpoint represents the position in the SystemStreamPartition.
+   */
+  @InterfaceStability.Evolving
+  default void register(SystemStreamPartition systemStreamPartition, Startpoint startpoint) {
+    throw new UnsupportedOperationException("This operation is not supported.");
```
cameronlee314: Could you please make the exception message more descriptive?

samza-core/src/test/scala/org/apache/samza/system/TestSystemConsumers.scala
```diff
@@ -340,7 +340,7 @@ class TestSystemConsumers {
     assertTrue(consumer.lastPoll.contains(systemStreamPartition1))
   }
 
-  @Test
+  @Test(expected = classOf[Exception])
```
cameronlee314: Why is this expected to throw an exception?

shanthoosh: Previously, we checked if the Consumer is a sub-type of StartpointVisitor and now that is removed. This test invokes the default implementation of `SystemConsumer.register(ssp, startpoint)` and hits the `UnsupportedOperationException`. I will improve this test, once we have one consumer which implements both `register(ssp, startpoint)` and `StartpointVisitor`.

cameronlee314: This class works against the SystemConsumer interface, so this test should be mocking a SystemConsumer, which means you don't need a consumer with a concrete implementation. You should be able to update this test within this PR.

shanthoosh: 1. Just to be clear, the suggestion you're pointing out is not based on the changes done in this patch.
2. Yes. I removed the SystemConsumer implementation present in the test class. Switched to using `InmemorySystemConsumer`.  Using it over mock, since the mock doesn't invoke the default implementation of `SystemConsumer` that throws up the expected exception.

cameronlee314: This test shouldn't depend on any concrete SystemConsumer, including InmemorySystemConsumer. You don't need a concrete SystemConsumer for this test. You should just have a mock SystemConsumer.

shanthoosh: Done.

samza-core/src/test/scala/org/apache/samza/system/TestSystemConsumers.scala
```diff
@@ -347,32 +348,15 @@ class TestSystemConsumers {
     val systemStreamPartition1 = new SystemStreamPartition(system, stream, new Partition(1))
     val systemStreamPartition2 = new SystemStreamPartition(system, stream, new Partition(2))
 
-    val consumer = new TestStartpointConsumer
```
cameronlee314: Please rename the test, since there is no visitor anymore.

shanthoosh: Done.

PR #911, opened 2019-02-07
Support for types in Samza SQL UDF

srinipunuru:
This PR adds following capabilities to UDF

1. Adds typed UDFS : Previously samza sql udfs were un-typed all the params used to be Objects, Although this allowed flexibility of polymorphism without having to implement multiple methods. This was fragile. We couldn't do validations during initialization phase. This change adds types to the UDFs.
2. Decouples name of the method from udf: previously all udf methods have to be named "execute". Now the Udfs methods can be named whatever the developer prefers. The UDFMethods are identified using the annotation. 
3. Adds polymorphism to UDFs : Adds capability where you can have mulitple methods for the same UDF name. i.e. multiple methods within the class can be annotated as SamzaSQLUDFMethod.
4. Minor : Adds description to the UDF. which can be used by the Shell.
5. Still keeps the support for un-typed functions that needs dynamic number of arguments like "BuildOutputRecord"

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/sql/udfs/ScalarUdf.java
+++ b/samza-api/src/main/java/org/apache/samza/sql/udfs/ScalarUdf.java
@@ -36,5 +36,4 @@ public interface ScalarUdf {
    * @param udfConfig Config specific to the udf.
    */
   void init(Config udfConfig);
-
 }
--- a/samza-api/src/main/java/org/apache/samza/sql/udfs/ScalarUdf.java
+++ b/samza-api/src/main/java/org/apache/samza/sql/udfs/ScalarUdf.java
@@ -23,8 +23,8 @@ import org.apache.samza.config.Config;
 
 
 /**
- * The base class for the Scalar UDFs. All the scalar UDF classes needs to extend this and implement a method named
- * "execute". The number and type of arguments for the method annotated with {@link SamzaSqlUdfMethod} in the UDF class should match the number and type of fields
+ * The base class for the Scalar UDFs. All the scalar UDF classes needs to extend this.
+ * The number and type of arguments for the method annotated with {@link SamzaSqlUdfMethod} in the UDF class should match the number and type of fields
  * used while invoking this UDF in SQL statement.
  * Say for e.g. User creates a UDF class with signature int execute(int var1, String var2). It can be used in a SQL query
  *     select myudf(id, name) from profile
--- a/samza-api/src/main/java/org/apache/samza/sql/udfs/ScalarUdf.java
+++ b/samza-api/src/main/java/org/apache/samza/sql/udfs/ScalarUdf.java
@@ -24,7 +24,7 @@ import org.apache.samza.config.Config;
 
 /**
  * The base class for the Scalar UDFs. All the scalar UDF classes needs to extend this and implement a method named
- * "execute". The number and type of arguments for the execute method in the UDF class should match the number and type of fields
+ * "execute". The number and type of arguments for the method annotated with {@link SamzaSqlUdfMethod} in the UDF class should match the number and type of fields
  * used while invoking this UDF in SQL statement.
  * Say for e.g. User creates a UDF class with signature int execute(int var1, String var2). It can be used in a SQL query
  *     select myudf(id, name) from profile
--- a/samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlExecutionContext.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlExecutionContext.java
@@ -19,7 +19,9 @@
 
 package org.apache.samza.sql.data;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -38,7 +40,7 @@ public class SamzaSqlExecutionContext implements Cloneable {
    * The variables that are shared among all cloned instance of {@link SamzaSqlExecutionContext}
    */
   private final SamzaSqlApplicationConfig sqlConfig;
-  private final Map<String, UdfMetadata> udfMetadata;
+  private final Map<String, List<UdfMetadata>> udfMetadata;
 
   /**
    * The variable that are not shared among all cloned instance of {@link SamzaSqlExecutionContext}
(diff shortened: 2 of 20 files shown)


samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlExecutionContext.java
```diff
@@ -38,7 +40,7 @@
    * The variables that are shared among all cloned instance of {@link SamzaSqlExecutionContext}
    */
   private final SamzaSqlApplicationConfig sqlConfig;
-  private final Map<String, UdfMetadata> udfMetadata;
+  private final Map<String, List<UdfMetadata>> udfMetadata;
```
weiqingy: So here, `UdfMetadata`  is for `SamzaSQLUDFMethod` ,  and `Map<String, List<UdfMetadata>>` is for `SamzaSqlUdf`?

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlExecutionContext.java
```diff
@@ -52,16 +54,19 @@ private SamzaSqlExecutionContext(SamzaSqlExecutionContext other) {
...
+    }
   }
 
   public ScalarUdf getOrCreateUdf(String clazz, String udfName) {
     return udfInstances.computeIfAbsent(udfName, s -> createInstance(clazz, udfName));
   }
 
   public ScalarUdf createInstance(String clazz, String udfName) {
-    Config udfConfig = udfMetadata.get(udfName).getUdfConfig();
+    Config udfConfig = udfMetadata.get(udfName).get(0).getUdfConfig();
```
weiqingy: Here, is it possible to move "[private final Config udfConfig](https://github.com/apache/samza/blob/c6ca30cc4869548a7766848c90833b4c133d6282/samza-sql/src/main/java/org/apache/samza/sql/interfaces/UdfMetadata.java#L38)" from UdfMetadata (SamzaSQLUDFMethod level) to UDF level?

srinipunuru: Good catch!! UdfMetadata is associated at the udflevel, We could consider creating a class level and then method level udf metadata. Creating such a structure just for holding the config seemed unncessary. I will add the comment here clarifying this behavior.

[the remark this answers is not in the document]
shenodaguirguis: I see. Thanks for the clarification & the  comment. Don't we still need to guard code against erroneously empty list?

samza-sql/src/main/java/org/apache/samza/sql/fn/BuildOutputRecordUdf.java
```diff
@@ -61,13 +61,13 @@
  * If no args is provided, it returns an empty SamzaSqlRelRecord (with empty field names and values list).
  */
 
-@SamzaSqlUdf(name="BuildOutputRecord")
+@SamzaSqlUdf(name="BuildOutputRecord" , description = "Creates an Output record.")
```
atoomula: typo: space after name. name = "BuildOutputRecord"

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlScalarFunctionImpl.java
```diff
@@ -61,17 +63,23 @@ public String getUdfName() {
     return udfName;
   }
 
+  public int numberArguments() {
```
atoomula: typo: numberOfArguments

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlExecutionContext.java
```diff
@@ -38,7 +40,10 @@
    * The variables that are shared among all cloned instance of {@link SamzaSqlExecutionContext}
    */
   private final SamzaSqlApplicationConfig sqlConfig;
-  private final Map<String, UdfMetadata> udfMetadata;
+
+  // Maps the UDF name to list of all UDF methods associated with the name.
+  // Since we support polymorphism there can multiple udfMetadata associated with the single name.
```
atoomula: How can these different methods be invoked via Udf in a sql statement ? Is there an example or a test ? I see one with overload but can we have different method names and make it work ?

samza-sql/src/main/java/org/apache/samza/sql/fn/GetSqlFieldUdf.java
```diff
@@ -51,22 +52,21 @@
  *           - sessionKey (Scalar)
  *
  */
-@SamzaSqlUdf(name = "GetSqlField")
+@SamzaSqlUdf(name = "GetSqlField", description = "Get an element from complex Sql field.")
```
atoomula: GetSqlField is kind of limiting as it returns only a String. For instance, we cannot get a sub-record or an array or any other type. Just wondering if other cases work if we return an object from execute ? Or since now we have ability to add more methods to an Udf (although I'm still not clear on how these different methods can be invoked in Udf unless they are fn overloads), is there any other way we could make it return different types ?

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlExecutionContext.java
srinipunuru: There was already a testcase (testEndToEndStreamTableTableJoinWithPrimaryKeys) which covers this, because we are passing both object as we all as an integer to the MyTest. I just added another test targeting this specifically.

samza-sql/src/main/java/org/apache/samza/sql/fn/GetSqlFieldUdf.java
srinipunuru: Function overloads will not work if only the return types change. They work if the arguments are different. If a UDF has to return a type that can be different, Then it has to return an object in which case the return type will be "ANY". Does it answer your question?

samza-api/src/main/java/org/apache/samza/sql/udfs/ScalarUdf.java
```diff
@@ -24,7 +24,7 @@
 
 /**
  * The base class for the Scalar UDFs. All the scalar UDF classes needs to extend this and implement a method named
- * "execute". The number and type of arguments for the execute method in the UDF class should match the number and type of fields
+ * "execute". The number and type of arguments for the method annotated with {@link SamzaSqlUdfMethod} in the UDF class should match the number and type of fields
```
weiqingy: Shall we also remove the "execute" from the first sentence?

samza-sql/src/main/java/org/apache/samza/sql/fn/GetSqlFieldUdf.java
atoomula: Sure. Could we change the description to reflect that it returns a string. May be: "Get an element as a String from Complex Sql field." ? As a separate effort, we will have to add another such Udf that returns ANY.

PR #916, opened 2019-02-12
End to end test case for union type with subRecord

srinipunuru:
1. Fix the bug in AvroRelConverter where it throws when a union type has a subRecord
2. Adding a test case with union type with a subrecord.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
@@ -207,7 +207,11 @@ public class AvroRelConverter implements SamzaRelConverter {
             .collect(Collectors.toMap(Map.Entry::getKey, e -> convertToAvroObject(e.getValue(),
                 getNonNullUnionSchema(schema).getValueType())));
       case UNION:
-        return convertToAvroObject(relObj, getNonNullUnionSchema(schema));
+        for (Schema unionSchema : schema.getTypes()) {
+          if (isSchemaCompatibleWithRelObj(relObj, unionSchema)) {
+            return convertToAvroObject(relObj, unionSchema);
+          }
+        }
       case ENUM:
         return new GenericData.EnumSymbol(schema, (String) relObj);
       case FIXED:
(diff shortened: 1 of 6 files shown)


samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
```diff
@@ -207,7 +207,11 @@ static public Object convertToAvroObject(Object relObj, Schema schema) {
             .collect(Collectors.toMap(Map.Entry::getKey, e -> convertToAvroObject(e.getValue(),
                 getNonNullUnionSchema(schema).getValueType())));
       case UNION:
-        return convertToAvroObject(relObj, getNonNullUnionSchema(schema));
+        for (Schema unionSchema : schema.getTypes()) {
```
atoomula: We have an AvroRelConverter version in Linkedin. We will have to merge these two and get rid of the Linkedin one.

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
```diff
@@ -283,4 +278,66 @@ public Object convertToJavaObject(Object avroObj, Schema schema) {
...
+        return relObj instanceof ByteString;
+      case FLOAT:
+        return relObj instanceof Float || relObj instanceof Double;
+      default:
+        return true;
+    }
+  }
+
+  // Two non-nullable types in a union is not yet supported.
+  public static Schema getNonNullUnionSchema(Schema schema) {
```
atoomula: How would this work if it is called from line 198 or from other complex types below it ? Wouldn't this function always return the same schema as the type would be the complex type it is called from ? I think I added that code. But it doesn't make sense looking at it now. It makes sense if it is called from convertToGenericRecord().

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
srinipunuru: I already updated the linkedIn version to unblock the customer. I will open a bug to remove the LI one and use the open source version.

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
srinipunuru: Yup getNonNullUnionSchema just returns the original schema if it is not a union schema.

srinipunuru:
> General comment not entirely related to this specific change:
> All the schemas in our tests have an outer record with primitive and complex types within it. We should also add tests with schemas containing just the primitive types or complex types (other than the record). I'm not sure if the code handles such cases.

Yup we need to support the primitive types in Samza SQL. I will open a bug to add a support for it.

PR #917, opened 2019-02-14
Adding bounded-time start of SamzaContainer from StreamProcessor

rmatharu:
This bounded-time start is needed to account for cases where the start of a container never returns within a bounded amount of time, e.g., when querying metadata starts taking a long amount of time. 
We would like the container to exit gracefully in that case.


samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -401,11 +403,29 @@ public void onJobModelExpired() {
...
+                LOGGER.error("Timed out {} ms while waiting for the container: {} to startup. Interrupting the container: {} thread to die", taskStartMs, container, container);
+                containerExcecutorService.shutdownNow();
+                state = State.STOPPED;
+                jobCoordinator.stop();
+                processorListener.afterFailure(new TimeoutException("Container start timed out after "+taskStartMs+ " ms"));
+              }
+            } catch (InterruptedException e) {
+              LOGGER.error("Exception occurred when starting the container: {}.", container, e);
+              state = State.STOPPED;
+              jobCoordinator.stop();
```
Sanil15: Shouldn't this be in finally?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -401,11 +403,29 @@ public void onJobModelExpired() {
...
             container = createSamzaContainer(processorId, jobModel);
             container.setContainerListener(new ContainerListener());
             LOGGER.info("Starting the container: {} for the stream processor: {}.", container, processorId);
             containerExcecutorService.submit(container);
+
+            try {
+              LOGGER.info("Waiting {} ms for the container: {} to startup.", taskStartMs, container);
+
+              if (!containerStartLatch.await(taskStartMs, TimeUnit.MILLISECONDS)) {
+                LOGGER.error("Timed out {} ms while waiting for the container: {} to startup. Interrupting the container: {} thread to die", taskStartMs, container, container);
```
Sanil15: nit: "Timed out after {} ms, waiting for the container ...."

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -401,11 +403,29 @@ public void onJobModelExpired() {
...
         synchronized (lock) {
           if (state == State.IN_REBALANCE) {
+            containerStartLatch = new CountDownLatch(1);
             containerShutdownLatch = new CountDownLatch(1);
             container = createSamzaContainer(processorId, jobModel);
             container.setContainerListener(new ContainerListener());
             LOGGER.info("Starting the container: {} for the stream processor: {}.", container, processorId);
             containerExcecutorService.submit(container);
+
+            try {
```
Sanil15: Consider this: Move try inside if and then move:

`state = State.STOPPED;
jobCoordinator.stop();
processorListener.afterFailure(new TimeoutException("Container start timed out after "+taskStartMs+ " ms"));`

inside a finally block

rmatharu:
Closing this since
deciding the bound on the start of a container is non-trivial, and depends on a large number of parameters. 
Instead, we can introduce bounded-start on components such as SystemsConsumer.

PR #918, opened 2019-02-14
SAMZA-2094: Implement the StartpointVisitor for the KafkaSystemConsumer.

shanthoosh:


shanthoosh:
@cameronlee314 

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -285,6 +321,27 @@ public void register(SystemStreamPartition systemStreamPartition, String offset)
...
+      LOG.warn("{}: ignoring SSP {}, because this consumer's system doesn't match.", this, systemStreamPartition);
+      return;
+    }
+
+    LOG.debug("Registering the ssp: {}, startpoint: {} with the consumer.", systemStreamPartition, startpoint);
+
+    super.register(systemStreamPartition, startpoint);
+
+    TopicPartition topicPartition = toTopicPartition(systemStreamPartition);
+    topicPartitionToStartpointMap.put(topicPartition, startpoint);
```
cameronlee314: There is logic in the old `register` which checks that the requested offset is older than the existing offset, in order to avoid missing messages. Just double checking that we want to intentionally leave that out?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -285,6 +321,27 @@ public void register(SystemStreamPartition systemStreamPartition, String offset)
     metrics.registerTopicAndPartition(toTopicAndPartition(tp));
   }
 
+  @Override
+  public void register(SystemStreamPartition systemStreamPartition, Startpoint startpoint) {
```
cameronlee314: Do you plan to remove the other `register` method once 1.0.1 is released, or do you plan to support the other `register` for more time?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -285,6 +321,27 @@ public void register(SystemStreamPartition systemStreamPartition, String offset)
...
+    }
+
+    if (!Objects.equals(systemStreamPartition.getSystem(), systemName)) {
+      LOG.warn("{}: ignoring SSP {}, because this consumer's system doesn't match.", this, systemStreamPartition);
+      return;
+    }
+
+    LOG.debug("Registering the ssp: {}, startpoint: {} with the consumer.", systemStreamPartition, startpoint);
+
+    super.register(systemStreamPartition, startpoint);
```
cameronlee314: Does `BlockingEnvelopeMap` implement this `register`? Seems like this will throw an exception.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -205,6 +215,156 @@ public void testFetchThresholdBytesDiabled() {
     consumer.stop();
   }
 
+  @Test
+  public void testStartpointSpecificOffsetVisitorShouldUpdateTheFetchOffsetInConsumer() {
+    // Define dummy variables for testing.
+    final Integer testPartitionId = 0;
```
cameronlee314: Maybe extract some of these dummy variables to static variables in the test class, so they can be shared with your other visitor tests?

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -205,6 +215,156 @@ public void testFetchThresholdBytesDiabled() {
...
+
+    // Mock verifications.
+    Mockito.verify(consumer).seekToEnd(ImmutableList.of(testTopicPartition));
+  }
+
+  @Test
+  public void testStartInvocationAfterStartPointsRegistrationShouldInvokeTheStartPointApplyMethod() {
+    // Initialize the constants required for the test.
+    Consumer mockConsumer = Mockito.mock(Consumer.class);
+    String testSystemName = "testSystem";
```
cameronlee314: Looks like some of these are already defined as static variables for the test class.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -205,6 +215,156 @@ public void testFetchThresholdBytesDiabled() {
...
+        mockStartPointVisitor = Mockito.mock(KafkaStartpointRegistrationHandler.class);
+    Mockito.doNothing().when(mockStartPointVisitor).visit(testSystemStreamPartition1, startPointSpecific);
+    Mockito.doNothing().when(mockStartPointVisitor).visit(testSystemStreamPartition2, startpointTimestamp);
+    Mockito.doNothing().when(mockStartPointVisitor).visit(testSystemStreamPartition3, startpointOldest);
+    Mockito.doNothing().when(mockStartPointVisitor).visit(testSystemStreamPartition4, startpointUpcoming);
+
+    // Instantiate KafkaSystemConsumer for testing.
+    KafkaSystemConsumer kafkaSystemConsumer = new KafkaSystemConsumer(mockConsumer, testSystemName, testConfig,
+                                                                      testClientId, kafkaSystemConsumerMetrics, new TestClock(), mockStartPointVisitor);
+    kafkaSystemConsumer.proxy = Mockito.mock(KafkaConsumerProxy.class);
```
cameronlee314: Could you please try to avoid directly accessing and mutating fields? Maybe include the proxy in your VisibleForTesting constructor.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
cameronlee314: Oops, I lost track of your change to `BlockingEnvelopeMap` above. Ignore the above comment.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -60,16 +74,22 @@
...
   final KafkaConsumerMessageSink messageSink;
 
   // This proxy contains a separate thread, which reads kafka messages (with consumer.poll()) and populates
   // BlockingEnvelopMap's buffers.
-  final private KafkaConsumerProxy proxy;
+  @VisibleForTesting
+  KafkaConsumerProxy proxy;
 
   // keep registration data until the start - mapping between registered SSPs and topicPartitions, and their offsets
   final Map<TopicPartition, String> topicPartitionsToOffset = new HashMap<>();
```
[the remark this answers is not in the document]
cameronlee314: 2A. I agree that it is odd for specific `SystemConsumer` implementations to each compare offsets. It seems outside of the scope of this PR though. Can we move that comparator logic refactoring to a separate PR so that we can cleanly separate the changes? That PR could be committed before this PR.
2B. I looked at some other code which uses `register` (e.g. `ContainerStorageManager`, `CoordinatorStreamStore`, and `CoordinatorStreamSystemConsumer`). Those don't really seem to do any comparisons to make sure they are registering the lowest offset. How can you tell all of the upper-layers already pass the lowest offset? I just want to double check that there wasn't a bug in an upper-layer which happened to be prevented by having offset comparator logic directly in `SystemConsumer` impls.

samza-core/src/main/scala/org/apache/samza/system/SystemConsumers.scala
```diff
@@ -208,7 +223,11 @@ class SystemConsumers (
       if (startpoint != null) {
         consumer.register(systemStreamPartition, startpoint)
       } else {
-        consumer.register(systemStreamPartition, offset)
+        val existingOffset = sspToRegisteredOffsets.get(systemStreamPartition)
+        val systemAdmin = systemAdmins.getSystemAdmin(systemStreamPartition.getSystem)
+        if (existingOffset == null || systemAdmin.offsetComparator(existingOffset, offset) > 0) {
+          sspToRegisteredOffsets.put(systemStreamPartition, offset)
```
cameronlee314: It seems conceptually a bit odd to call `SystemConsumer.register` in `SystemConsumers.start`. If `SystemConsumer.register` works with multiple calls (so it just registers the offset from the latest call to `register`), then you could call `register` here, and then you wouldn't need to call it in `start`.
Not sure if this is feasible, but I think it would make the flow a bit cleaner, since you wouldn't have to worry about doing anything extra in `start`.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -330,6 +320,73 @@ public String getSystemName() {
...
+
+    @Override
+    public void visit(SystemStreamPartition systemStreamPartition, StartpointTimestamp startpointTimestamp) {
+      Long timestampInStartpoint = startpointTimestamp.getTimestampOffset();
+      TopicPartition topicPartition = toTopicPartition(systemStreamPartition);
+      Map<TopicPartition, Long> topicPartitionsToTimeStamps = ImmutableMap.of(topicPartition, timestampInStartpoint);
+
+      // Look up the offset by timestamp.
+      LOG.info("Looking up the offsets of the topic partition: {} by timestamp: {}.", topicPartition, timestampInStartpoint);
+      Map<TopicPartition, OffsetAndTimestamp> topicPartitionToOffsetTimestamps = kafkaConsumer.offsetsForTimes(topicPartitionsToTimeStamps);
```
dnishimura: According to the [Kafka documentation](https://kafka.apache.org/0110/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#offsetsForTimes(java.util.Map)), `OffsetAndTimestamp` may be null for a partition if no events exist at or after the specified timestamp. Probably need to handle null in this case. Seeking to `upcoming` probably makes the most sense.

samza-core/src/main/scala/org/apache/samza/system/SystemConsumers.scala
dnishimura: Adding to @cameronlee314's comment. What's the motivation for only accepting the older offset if `register` is called multiple times? When would this happen? Is it the broadcast stream scenario?

samza-core/src/main/scala/org/apache/samza/system/SystemConsumers.scala
```diff
@@ -154,6 +164,11 @@ class SystemConsumers (
   metrics.setUnprocessedMessages(() => totalUnprocessedMessages)
 
   def start {
+    for ((systemStreamPartition, offset) <- sspToRegisteredOffsets.asScala) {
```
dnishimura: Just wondering why is this necessary? When would consumers restart multiple times and seek to the last registered offset? Is it some error scenario you're handling?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -159,30 +179,16 @@ private void startSubscription() {
...
-
-      LOG.info("{}: Changing consumer's starting offset for tp = {} to {}", this, tp, startingOffsetString);
-
+    topicPartitionToStartpointMap.forEach((topicPartition, startpoint) -> {
+      Partition partition = new Partition(topicPartition.partition());
+      SystemStreamPartition systemStreamPartition = new SystemStreamPartition(systemName, topicPartition.topic(), partition);
+      startpoint.apply(systemStreamPartition, kafkaStartpointRegistrationHandler);
       // add the partition to the proxy
-      proxy.addTopicPartition(topicPartitionsToSSP.get(tp), startingOffset);
+      proxy.addTopicPartition(systemStreamPartition, kafkaConsumer.position(topicPartition));
```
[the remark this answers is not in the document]
dnishimura: What was the resolution here?

samza-core/src/main/scala/org/apache/samza/system/SystemConsumers.scala
shanthoosh: 1. This would happen in the broadcast stream scenario. This older-offset-comparator check is done by some of the SystemConsumer implementation and not in the others. I'm moving it to a common layer like `SystemConsumers` above to ensure functional correctness. More context [here](https://github.com/apache/samza/pull/918#discussion_r257523873).
2. Currently this odd behavior is what all the system consumer implementations that support broadcast streams do.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
shanthoosh: Good question. 
Excerpt the kafka javadoc linked in the above comment:
```
If the message format version in a partition is before 0.10.0, i.e. the messages do not have timestamps, null will be returned for that partition
```

Null would be returned by  KafkaConsumer `offsetsForTimes` API when used with the kafka broker version < `0.10.0`. 

Starting from 1.0, samza supports kafka version: `0.11.0.2`. From the [kafka-compatibility-matrix](https://cwiki.apache.org/confluence/display/KAFKA/Compatibility+Matrix), kafka-client version used by samza is not compatible with broker versions < `0.10.0`. 

Since samza doesn't support kafka brokers < 0.10.0, I'm not sure if we should handle null value returned by this API.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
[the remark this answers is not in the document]
shanthoosh: 1. Created PR: 954 which addresses the comment of moving the refactor into separate PR. 
2. Added the rationale in description of PR: 954 as to why this refactor would not break any existing functionality.
    * `CoordinatorStreamSystemConsumer`: Reads the coordinator stream from oldest offset. Please check here(https://github.com/apache/samza/blob/master/samza-core/src/main/java/org/apache/samza/coordinator/stream/CoordinatorStreamSystemConsumer.java#L128) for more details.
    * `CoordinatorStreamStore`: Reads the coordinator stream from oldest offset. Please check [here](https://github.com/apache/samza/blob/master/samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStore.java#L183) for more details. 
    *  `ContainerStorageManager`: Chooses the lowest offset before registering with consumer. [Relevant code](https://github.com/apache/samza/blob/master/samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java#L68) .

Please let me know if there're any other questions.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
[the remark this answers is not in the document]
cameronlee314: Your comment from a while back says that you moved this, and the comment was resolved. However, the PR still seems to have the proxy in the old place. Can you please clarify?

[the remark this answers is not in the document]
shanthoosh: @cameronlee314  My changes in `KafkaSystemConsumer` were operating on `KafkaConsumerProxy`. However, after  SAMZA-2120, this was not the case. My local changes of registering SSP with proxy in `RegistrationHandler` did not apply cleanly with the recent master(Since the recent master had  `KafkaConsumerProxyFactory`). Updated the patch after re-implementing some of my changes and now SSP registration with proxy happens in `RegistrationHandler`.

dnishimura:
Is the timeout error shown in Travis-CI just a transient thing and not reproducible locally?
```testShouldCreateCoordinatorStream FAILED
    org.apache.samza.SamzaException: Creation of topic test-coordinator-stream failed.
        Caused by:
        java.util.concurrent.TimeoutException```

**Comment by shanthoosh** [comment]
@dnishimura No, I tried it locally multiple times but was not able to reproduce it. But that failure is orthogonal to this change. That test failure originates from the API: `KafkaSystemAdmin.createStream()`

File: samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java [hunk 12]
```diff
@@ -330,6 +325,88 @@ public String getSystemName() {
...
+      long offsetInStartpoint = Long.parseLong(startpointSpecific.getSpecificOffset());
+      LOG.info("Updating the consumer fetch offsets of topic partition: {} to {}.", topicPartition, offsetInStartpoint);
+
+      // KafkaConsumer is not thread-safe.
+      synchronized (kafkaConsumer) {
+        kafkaConsumer.seek(topicPartition, offsetInStartpoint);
+      }
+
+      // add the partition to the proxy
+      proxy.addTopicPartition(systemStreamPartition, kafkaConsumer.position(topicPartition));
```
cameronlee314: Does `kafkaConsumer.position` need to be protected by the lock too?
Also, please check below for other usages of `kafkaConsumer` outside of the lock.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -330,6 +325,88 @@ public String getSystemName() {
...
+
+    @Override
+    public void visit(SystemStreamPartition systemStreamPartition, StartpointTimestamp startpointTimestamp) {
+      Long timestampInStartpoint = startpointTimestamp.getTimestampOffset();
+      TopicPartition topicPartition = toTopicPartition(systemStreamPartition);
+      Map<TopicPartition, Long> topicPartitionsToTimeStamps = ImmutableMap.of(topicPartition, timestampInStartpoint);
+
+      // Look up the offset by timestamp.
+      LOG.info("Looking up the offsets of the topic partition: {} by timestamp: {}.", topicPartition, timestampInStartpoint);
+      Map<TopicPartition, OffsetAndTimestamp> topicPartitionToOffsetTimestamps = kafkaConsumer.offsetsForTimes(topicPartitionsToTimeStamps);
```
cameronlee314: Does this need to be protected by the `kafkaConsumer` lock?

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -119,9 +136,11 @@ public void testConsumerRegisterOlderOffsetOfTheSamzaSSP() {
     consumer.register(ssp1, "3");
     consumer.register(ssp2, "0");
 
-    assertEquals("0", consumer.topicPartitionsToOffset.get(KafkaSystemConsumer.toTopicPartition(ssp0)));
-    assertEquals("2", consumer.topicPartitionsToOffset.get(KafkaSystemConsumer.toTopicPartition(ssp1)));
-    assertEquals("0", consumer.topicPartitionsToOffset.get(KafkaSystemConsumer.toTopicPartition(ssp2)));
+    consumer.start();
+
+    assertEquals("5", ((StartpointSpecific) consumer.topicPartitionToStartpointMap.get(KafkaSystemConsumer.toTopicPartition(ssp0))).getSpecificOffset());
```
cameronlee314: The name of the test doesn't seem to match the actual test logic, since the oldest offset is no longer being used.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -206,6 +225,166 @@ public void testFetchThresholdBytesDiabled() {
     consumer.stop();
   }
 
+  @Test
+  public void testStartpointSpecificOffsetVisitorShouldUpdateTheFetchOffsetInConsumer() {
+    final KafkaConsumer consumer = Mockito.mock(KafkaConsumer.class);
+    final KafkaConsumerProxy kafkaConsumerProxy = Mockito.mock(KafkaConsumerProxy.class);
+    KafkaSystemConsumer kafkaSystemConsumer = new KafkaSystemConsumer(consumer, TEST_SYSTEM, new MapConfig(), TEST_CLIENT_ID, kafkaConsumerProxy, Mockito.mock(KafkaSystemConsumerMetrics.class), new TestClock(), REGISTRATION_HANDLER);
+    KafkaStartpointRegistrationHandler kafkaStartpointRegistrationHandler = kafkaSystemConsumer.new KafkaStartpointRegistrationHandler();
```
cameronlee314: It seems kind of awkward to test like this, where you put in the mock registration handler, and then create a new one to actually test with. Could you make `KafkaStartpointRegistrationHandler` a static inner class and then pass the necessary objects to it? In my opinion, non-static inner classes are harder to read/maintain anyways.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -206,6 +225,166 @@ public void testFetchThresholdBytesDiabled() {
...
+
+    // Mock the consumer interactions.
+    Mockito.doNothing().when(consumer).seek(TEST_TOPIC_PARTITION, Long.valueOf(TEST_OFFSET));
+
+    // Invoke the consumer with startpoint.
+    kafkaStartpointRegistrationHandler.visit(TEST_SYSTEM_STREAM_PARTITION, testStartpointSpecific);
+
+    // Mock verifications.
+    Mockito.verify(consumer).seek(TEST_TOPIC_PARTITION, Long.valueOf(TEST_OFFSET));
+    Mockito.verify(kafkaConsumerProxy).addTopicPartition(Mockito.any(SystemStreamPartition.class), Mockito.anyLong());
```
cameronlee314: Can this verify that the correct `SystemStreamPartition` and offset get used, instead of just doing "any"?

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -206,6 +225,166 @@ public void testFetchThresholdBytesDiabled() {
...
+    // Mock the consumer interactions.
+    Mockito.when(consumer.offsetsForTimes(Mockito.anyMap())).thenReturn(offsetForTimesResult);
+    Mockito.doNothing().when(consumer).seek(TEST_TOPIC_PARTITION, Long.valueOf(TEST_OFFSET));
+
+    kafkaStartpointRegistrationHandler.visit(TEST_SYSTEM_STREAM_PARTITION, startpointTimestamp);
+
+    // Mock verifications.
+    Mockito.verify(consumer).seek(TEST_TOPIC_PARTITION, Long.valueOf(TEST_OFFSET));
+    Mockito.verify(consumer).offsetsForTimes(Mockito.anyMap());
+    Mockito.verify(kafkaConsumerProxy).addTopicPartition(Mockito.any(SystemStreamPartition.class), Mockito.anyLong());
```
cameronlee314: Same as above about not using "any".

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -206,6 +225,166 @@ public void testFetchThresholdBytesDiabled() {
...
+    KafkaSystemConsumer kafkaSystemConsumer = new KafkaSystemConsumer(consumer, TEST_SYSTEM, new MapConfig(), TEST_CLIENT_ID, kafkaConsumerProxy, Mockito.mock(KafkaSystemConsumerMetrics.class), new TestClock(), REGISTRATION_HANDLER);
+
+    KafkaStartpointRegistrationHandler kafkaStartpointRegistrationHandler = kafkaSystemConsumer.new KafkaStartpointRegistrationHandler();
+
+    final StartpointTimestamp startpointTimestamp = new StartpointTimestamp(testTimeStamp);
+    final Map<TopicPartition, OffsetAndTimestamp> offsetForTimesResult = ImmutableMap.of(
+        TEST_TOPIC_PARTITION, new OffsetAndTimestamp(Long.valueOf(TEST_OFFSET), testTimeStamp));
+
+    // Mock the consumer interactions.
+    Mockito.when(consumer.offsetsForTimes(Mockito.anyMap())).thenReturn(offsetForTimesResult);
```
cameronlee314: We should be able to check a specific argument instead of using `anyMap`, right?

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -206,6 +225,166 @@ public void testFetchThresholdBytesDiabled() {
...
+    Mockito.doNothing().when(consumer).seek(TEST_TOPIC_PARTITION, Long.valueOf(TEST_OFFSET));
+
+    kafkaStartpointRegistrationHandler.visit(TEST_SYSTEM_STREAM_PARTITION, startpointTimestamp);
+
+    // Mock verifications.
+    Mockito.verify(consumer).seek(TEST_TOPIC_PARTITION, Long.valueOf(TEST_OFFSET));
+    Mockito.verify(consumer).offsetsForTimes(Mockito.anyMap());
```
cameronlee314: I don't think `offsetsForTimes` has side effects, so you probably don't need to verify this particular call. The test won't work if it doesn't get called anyways.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -206,6 +225,166 @@ public void testFetchThresholdBytesDiabled() {
...
+    SystemStreamPartition testSystemStreamPartition3 = new SystemStreamPartition(TEST_SYSTEM, TEST_STREAM, new Partition(2));
+    SystemStreamPartition testSystemStreamPartition4 = new SystemStreamPartition(TEST_SYSTEM, TEST_STREAM, new Partition(3));
+
+    // Different kinds of {@code Startpoint}.
+    StartpointSpecific startPointSpecific = new StartpointSpecific("100");
+    StartpointTimestamp startpointTimestamp = new StartpointTimestamp(100L);
+    StartpointOldest startpointOldest = new StartpointOldest();
+    StartpointUpcoming startpointUpcoming = new StartpointUpcoming();
+
+    // Mock the visit methods of KafkaStartpointRegistrationHandler.
```
cameronlee314: Is it necessary to mock the visitor to do nothing?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
shanthoosh: Moved all the kafkaConsumer invocations into a lock.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
cameronlee314: I don't think that you need to mock the object if you are just verifying it. Maybe I'm missing something though, so if the test fails when you remove this section, then leave it in.

PR #920, opened 2019-02-15
SAMZA-2106: Samza app and job config refactor

Sanil15:
Proposal doc: https://docs.google.com/document/d/1Q8tNF6zlXSyefOO6koXWgiDuyV6eLngKoU-5KLvq3I4/edit

Testing:

- Added/Refactored unit tests 
- Tested by deploying samza-hello-samza jobs


Sanil15:
@vjagadish1989 please have a look.

samza-core/src/main/java/org/apache/samza/execution/JobPlanner.java
```diff
@@ -155,4 +158,33 @@ final void writePlanJsonFile(String planJson) {
...
+  /**
+   * Generates job.id from app.id and job.name from app.name config
+   * If both job.id and app.id is defined, app.id takes precedence and job.id is set to value of app.id
+   * If both job.name and app.name is defined, app.name takes precedence and job.name is set to value of app.name
+   *
+   * @param allowedUserConfig configs passed from user
+   */
+  void generateJobIdAndName(Map<String, String> allowedUserConfig) {
+    if (allowedUserConfig.containsKey(JobConfig.JOB_ID())) {
+      LOG.warn("{} is a deprecated configuration, use app.id instead.", JobConfig.JOB_ID());
```
shanthoosh: 1. Can you please deprecate the job.id, job.name configurations from the samza configuration table. 
2. There're some samples in [samza-hello-samza](https://github.com/apache/samza-hello-samza/blob/master/src/main/config/azure-application-local-runner.properties) in which job.id, job.name are defined as configuration. Could you please update them to use app.id, app.name. Could you check if the same point applies to the samza-hello-samza samples within linkedin? 
3. Could you please update the integration test files located in samza-test/src/test/ to use the new app.id, app.name configuration. [Example](https://github.com/apache/samza/blob/master/samza-test/src/main/config/join/emitter.samza)
4. From what I see, we're doing configuration aliasing in the `LocalApplicationRunner`, `JobPlanner` directly. Could you share what is the rationale behind this decision? Why not do this in a `Rewriter`?
5. Consider a case where user has defined only `app.id`, `app.name` in the configuration and launches their low-level application. Could you please share how will submitting `YarnJob` through `JobRunner` work? `run-job.sh`  launches `YarnJob` which requires the  `job.name`, `job.id` to built through `CoordinatorStreamUtil.buildCoordinatorStreamConfig` ?

shanthoosh: Have you planned to migrate the existing tools like `CheckpointTool`, `CoordinatorStreamWriter` etc to use the same configuration aliasing ?

Sanil15: Thanks for the review @shanthoosh 

1. Sure will do

2. We will be able to migrate this once a release containing these changes is published but I have that in mind

3. I changed a few, once again this change is user-facing for not using job.id and job.name. Hence we need to only change user-facing samples/prototypes/tests

4. Rewriters are a pluggable component, that is defined by a user. Changes I am proposing is independent of rewriting/aliasing. Changes I have proposed will make Samza framework generate job.id and job.name consistently and internally use it.  

5. I was planning to do those in another RB, but the changes are small enough so I included them here. 

Thanks

samza-core/src/main/scala/org/apache/samza/coordinator/stream/CoordinatorStreamWriterCommandLine.scala
```diff
@@ -68,4 +72,27 @@ class CoordinatorStreamWriterCommandLine extends CommandLine {
 
     value
   }
+
+  def genJobConfigs(config: MapConfig): MapConfig = {
```
shanthoosh: Rather than duplicating this code at two places, could you please extract them into a util method and move the duplicated code there.  Invoke the util method at all the places where you're calling this method.

samza-core/src/main/java/org/apache/samza/execution/JobPlanner.java
```diff
@@ -155,4 +158,33 @@ final void writePlanJsonFile(String planJson) {
...
+    }
+
+    if (allowedUserConfig.containsKey(JobConfig.JOB_NAME())) {
+      LOG.warn("{} is a deprecated configuration, use use app.name instead.", JobConfig.JOB_NAME());
+    }
+
+    if (allowedUserConfig.containsKey(ApplicationConfig.APP_NAME)) {
```
shanthoosh: It is a good practice not to mutate the arguments(`allowedUserConfig`) as a part of the method implementation. It will improve readability and be simpler when someone has to re-read again in the future. Rather than having return type void in this implementation, may be you can return the updated configuration, like 

```
MapConfig calculateJobIdAndJobName(String jobName, String jobId)
```

At caller, you can simply add the return value you received from this method invocation to the final configuration bag. 

Please change if this makes sense.

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
```diff
@@ -104,7 +105,7 @@
   private TestRunner() {
     this.configs = new HashMap<>();
     this.inMemoryScope = RandomStringUtils.random(10, true, true);
-    configs.put(JobConfig.JOB_NAME(), JOB_NAME);
+    configs.put(ApplicationConfig.APP_NAME, JOB_NAME);
```
shanthoosh: Rename the `JOB_NAME`  constant to `APP_NAME`.

samza-core/src/test/java/org/apache/samza/runtime/TestLocalApplicationRunner.java
```diff
@@ -75,8 +75,8 @@ public void setUp() {
   public void testRunStreamTask() {
     final Map<String, String> cfgs = new HashMap<>();
     cfgs.put(ApplicationConfig.APP_PROCESSOR_ID_GENERATOR_CLASS, UUIDGenerator.class.getName());
-    cfgs.put(JobConfig.JOB_NAME(), "test-task-job");
-    cfgs.put(JobConfig.JOB_ID(), "jobId");
+    cfgs.put(ApplicationConfig.APP_NAME, "test-task-job");
+    cfgs.put(ApplicationConfig.APP_ID, "jobId");
```
shanthoosh: s/jobId/testAppId
s/test-task-job/testAppName

samza-core/src/main/scala/org/apache/samza/coordinator/stream/CoordinatorStreamWriterCommandLine.scala
Sanil15: Yea was thinking to do that, will do!

samza-core/src/main/java/org/apache/samza/execution/JobPlanner.java
Sanil15: Will do

samza-test/src/main/java/org/apache/samza/test/framework/TestRunner.java
Sanil15: Ok

samza-core/src/test/java/org/apache/samza/runtime/TestLocalApplicationRunner.java
Sanil15: ok

samza-core/src/main/java/org/apache/samza/execution/JobPlanner.java
```diff
@@ -82,10 +84,13 @@ ExecutionPlan getExecutionPlan(String runId) {
             userConfig.get(TaskConfig.INPUT_STREAMS()));
         allowedUserConfig.remove(TaskConfig.INPUT_STREAMS());
       }
-
       generatedConfig.putAll(getGeneratedConfig(runId));
     }
 
+    // Generate job.id and job.name configs from app.id and app.name if defined
+    MapConfig generatedJobConfigs = JobConfigUtil.generateJobIdAndName(new MapConfig(allowedUserConfig));
```
xinyuiscool: this is not correct. The job name and id got generated for all the jobs here. This logic won't work if we split the app into multiple jobs. The job name and id should be set up correctly in the JobNode.

samza-core/src/main/java/org/apache/samza/util/JobConfigUtil.java
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+   * Generates job.id from app.id and job.name from app.name config
+   * If both job.id and app.id is defined, app.id takes precedence and job.id is set to value of app.id
+   * If both job.name and app.name is defined, app.name takes precedence and job.name is set to value of app.name
+   *
+   * @param userConfigs configs passed from user
+   *
+   */
+  public static MapConfig generateJobIdAndName(MapConfig userConfigs) {
```
xinyuiscool: the input type should be map<string, String>

samza-core/src/main/java/org/apache/samza/execution/JobPlanner.java
Sanil15: will do

samza-core/src/main/java/org/apache/samza/util/JobConfigUtil.java
Sanil15: will do

Sanil15:
@xinyuiscool not all the code that uses JOB.NAME, reads it of JobConfig#getName method and the same applies to JOB.ID

With this approach, we can easily remove job.id and job.name from user-defined configs in next major release

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
```diff
@@ -113,7 +115,7 @@ private void validateConfig() {
    */
   /* package private */
   JobGraph createJobGraph(ApplicationDescriptorImpl<? extends ApplicationDescriptor> appDesc) {
-    JobGraph jobGraph = new JobGraph(config, appDesc);
+    JobGraph jobGraph = new JobGraph(new MapConfig(config), appDesc);
```
xinyuiscool: why we need to wrap MapConfig on the config?

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -84,7 +85,7 @@ public void kill() {
     // since currently we only support single actual remote job, we can get its status without
     // building the execution plan.
     try {
-      JobConfig jc = new JobConfig(appDesc.getConfig());
+      JobConfig jc = new JobConfig(JobConfigUtil.generateJobIdAndName(appDesc.getConfig()));
```
xinyuiscool: This JobConfigUtil.generateJobIdAndName should be in the planner. You can add a function directly there like 
   planer.createSingleJobConfig(). This will be better handled in a single place.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -98,7 +99,7 @@ public ApplicationStatus status() {
     // since currently we only support single actual remote job, we can get its status without
     // building the execution plan
     try {
-      JobConfig jc = new JobConfig(appDesc.getConfig());
+      JobConfig jc = new JobConfig(JobConfigUtil.generateJobIdAndName(appDesc.getConfig()));
```
xinyuiscool: same above.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -112,7 +113,7 @@ public void waitForFinish() {
 
   @Override
   public boolean waitForFinish(Duration timeout) {
-    JobConfig jobConfig = new JobConfig(appDesc.getConfig());
+    JobConfig jobConfig = new JobConfig(JobConfigUtil.generateJobIdAndName(appDesc.getConfig()));
```
xinyuiscool: same above.

samza-core/src/main/java/org/apache/samza/util/JobConfigUtil.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Map;
+import org.apache.samza.SamzaException;
+import org.apache.samza.config.ApplicationConfig;
+import org.apache.samza.config.JobConfig;
+import org.apache.samza.config.MapConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class JobConfigUtil {
```
xinyuiscool: this logic should belong to the planer. See the comments above.

samza-core/src/main/java/org/apache/samza/execution/ExecutionPlanner.java
Sanil15: yes that should not exist

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
Sanil15: will do

samza-core/src/main/java/org/apache/samza/util/JobConfigUtil.java
Sanil15: will do

PR #921, opened 2019-02-15
SAMZA-2105: [elasticsearch, hdfs, kafka] code cleanup and refactoring

strkkk:
1. Added/removed modifiers
2. Guava optional -> java optional and removed guava from samza-elasticsearch module dependencies. Guava optional was added in https://issues.apache.org/jira/browse/SAMZA-853 but it is not clear why guava is better than default API.
3. Few code snippets are simplified


strkkk:
@vjagadish1989 thanks. Can you merge it?

PR #924, opened 2019-02-19
SAMZA-2108: Check for host affinity config before resolving preferred host matching

dxichen:
@vjagadish @rmatharu  Added check in the AbstractContainerAllocator to check the host-affinity config before using the preferred host mapping

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/clustermanager/AbstractContainerAllocator.java
+++ b/samza-core/src/main/java/org/apache/samza/clustermanager/AbstractContainerAllocator.java
@@ -55,6 +55,8 @@ public abstract class AbstractContainerAllocator implements Runnable {
    */
   private final TaskConfig taskConfig;
 
+  private final ClusterManagerConfig clusterManagerConfig;
+
   private final Config config;
 
   /**
@@ -92,6 +94,7 @@ public abstract class AbstractContainerAllocator implements Runnable {
     this.containerMemoryMb = clusterManagerConfig.getContainerMemoryMb();
     this.containerNumCpuCores = clusterManagerConfig.getNumCores();
     this.taskConfig = new TaskConfig(config);
+    this.clusterManagerConfig = new ClusterManagerConfig(config);
     this.state = state;
     this.config = config;
   }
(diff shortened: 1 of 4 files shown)


samza-core/src/main/java/org/apache/samza/clustermanager/AbstractContainerAllocator.java
```diff
@@ -171,7 +174,7 @@ public void requestResources(Map<String, String> resourceToHostMappings) {
     for (Map.Entry<String, String> entry : resourceToHostMappings.entrySet()) {
       String containerId = entry.getKey();
       String preferredHost = entry.getValue();
-      if (preferredHost == null)
+      if (preferredHost == null || !clusterManagerConfig.getHostAffinityEnabled())
```
rmatharu: You can make AbstractContainerAllocator.java ignore all present resourceToHostMappings, since AbstractContainerAllocator is instantiated only in case host-affinity is disabled, otherwise we instantiate 
HostAwareContainerAllocator.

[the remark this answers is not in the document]
dxichen: good catch, there is already logics in ContainerProcessManager to differentiate between host-aware/ host unaware container allocators, I will leverage that check to these 2 requestResources methods. (ie have specific implementations in ContainerAllocator and HostAwareContainerAllocator for requestResources)

samza-core/src/main/java/org/apache/samza/clustermanager/AbstractContainerAllocator.java
```diff
@@ -55,6 +55,8 @@
    */
   private final TaskConfig taskConfig;
 
+  private final ClusterManagerConfig clusterManagerConfig;
```
vjagadish1989: this should be no longer necessary right? since, we've delegated allocator-specific behavior to the respective sub-classes

dxichen: Oops removed, good catch!

PR #925, opened 2019-02-19
SAMZA-2109: Reduce per-partition default buffer sizes

vjagadish1989:


shanthoosh:
Is it possible to back-port this patch to samza internal releases done at linkedin?

vjagadish1989:
Since, hot-fixes are reserved for critical bug fixes - probably not worth back-porting new features/enhancements. I'd wait for 1.0.1, since it's around the corner anyways :-)

PR #926, opened 2019-02-20
Records with all null values are transformed to null records

srinipunuru:
Right now Samza SQL AvroRelConverter transforms the records with null values into null records. I think this was added to support `select * from` usecase when the input has null records, But ideally that should be supported through a mechanism where the entire record is passed as null from input to the output rather than add a current hack where we are flattening the record with null values into null record.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
@@ -153,7 +153,6 @@ public class AvroRelConverter implements SamzaRelConverter {
     GenericRecord record = new GenericData.Record(schema);
     List<String> fieldNames = relRecord.getFieldNames();
     List<Object> values = relRecord.getFieldValues();
-    int nullFieldValueCount = 0;
     for (int index = 0; index < fieldNames.size(); index++) {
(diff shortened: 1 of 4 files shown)


atoomula:
> Right now Samza SQL AvroRelConverter transforms the records with null values into null records. I think this was added to support `select * from` usecase when the input has null records, But ideally that should be supported through a mechanism where the entire record is passed as null from input to the output rather than add a current hack where we are flattening the record with null values into null record.

While converting a null avro record to rel record on line 90 (which is for the top level root record), we are explicitly setting each field in that record to null. The reverse is done here only for the root record for symmetry. Without the change that is reverted in the review, we cannot delete a record in a remote table, where puts with null values are considered deletes.

srinipunuru:
> > Right now Samza SQL AvroRelConverter transforms the records with null values into null records. I think this was added to support `select * from` usecase when the input has null records, But ideally that should be supported through a mechanism where the entire record is passed as null from input to the output rather than add a current hack where we are flattening the record with null values into null record.
> 
> While converting a null avro record to rel record on line 90 (which is for the top level root record), we are explicitly setting each field in that record to null. The reverse is done here only for the root record for symmetry. Without the change that is reverted in the review, we cannot delete a record in a remote table, where puts with null values are considered deletes.

Can we use `__avroRecord__` to figure out whether the payload is null or not?
 
Looks like we are missing an end to end table test case for this scenario. Can we add a test case for this scenario?

PR #927, opened 2019-02-21
SAMZA-2111: Refactoring ApplicationRunnerMain code

atoomula:
Adding invokeApplicationRunner public static API in preparation for fast sql where this API will be called directly within the same process.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
@@ -50,8 +50,12 @@ public class ApplicationRunnerMain {
     ApplicationRunnerCommandLine cmdLine = new ApplicationRunnerCommandLine();
     OptionSet options = cmdLine.parser().parse(args);
     Config orgConfig = cmdLine.loadConfig(options);
-    Config config = Util.rewriteConfig(orgConfig);
     ApplicationRunnerOperation op = cmdLine.getOperation(options);
+    invokeApplicationRunner(orgConfig, op);
+  }
+
+  public static ApplicationRunner invokeApplicationRunner(Config orgConfig, ApplicationRunnerOperation op) {
+    Config config = Util.rewriteConfig(orgConfig);
 
     ApplicationRunner appRunner =
         ApplicationRunners.getApplicationRunner(ApplicationUtil.fromConfig(config), config);
@@ -69,5 +73,6 @@ public class ApplicationRunnerMain {
       default:
         throw new IllegalArgumentException("Unrecognized operation: " + op);
     }
+    return appRunner;
   }
 }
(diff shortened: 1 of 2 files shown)


samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
```diff
@@ -50,8 +50,12 @@ public static void main(String[] args) throws Exception {
     ApplicationRunnerCommandLine cmdLine = new ApplicationRunnerCommandLine();
     OptionSet options = cmdLine.parser().parse(args);
     Config orgConfig = cmdLine.loadConfig(options);
-    Config config = Util.rewriteConfig(orgConfig);
     ApplicationRunnerOperation op = cmdLine.getOperation(options);
+    invokeApplicationRunner(orgConfig, op);
+  }
+
+  public static ApplicationRunner invokeApplicationRunner(Config orgConfig, ApplicationRunnerOperation op) {
```
prateekm: Do you need to return the ApplicationRunner?

Nitpick: s/orgConfig/config/

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
```diff
@@ -50,8 +50,12 @@ public static void main(String[] args) throws Exception {
...
+    Config config = Util.rewriteConfig(orgConfig);
```
prateekm: Config rewrite shouldn't happen in this method since it's orthogonal to rest of the functionality and is already available as a util.

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
atoomula: It will be needed to get the status.

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
```diff
@@ -51,28 +50,6 @@ public static void main(String[] args) throws Exception {
...
-        appRunner.kill();
-        break;
-      case STATUS:
-        System.out.println(appRunner.status());
-        break;
-      default:
-        throw new IllegalArgumentException("Unrecognized operation: " + op);
-    }
-    return appRunner;
+    ApplicationRunnerUtil.invokeApplicationRunner(orgConfig, op);
```
prateekm: Minor: s/invokeApplicationRunner/invoke
Minor: s/orgConfig/originalConfig

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
atoomula: We agreed to keep it the same.

PR #928, opened 2019-02-21
SAMZA-1990: Samza framework should let using the same system stream as both input and output.

dnishimura:
**Symptom:** An `IllegalArgumentException` is thrown when the same `streamId` is referred from multiple input/output stream descriptors.
**Cause:** The `ApplicationDescriptorImpl` caches the serde instances for streams by a `streamId` and there's a check to ensure the expected stream serde matches when using the same stream from multiple input/output descriptors. However the check is incorrect because it compares serde instances and not serde types. This check always fails in this scenario.
**Fix:** Compare the stream serdes for a particular `streamId` by type.

Please take a look @prateekm @nickpan47 
CC: @atoomula

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/application/descriptors/ApplicationDescriptorImpl.java
+++ b/samza-core/src/main/java/org/apache/samza/application/descriptors/ApplicationDescriptorImpl.java
@@ -311,9 +311,12 @@ public abstract class ApplicationDescriptorImpl<S extends ApplicationDescriptor>
             ". Values will not be (de)serialized");
       }
       streamSerdes.put(streamId, KV.of(keySerde, valueSerde));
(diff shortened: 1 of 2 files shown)


prateekm:
@dnishimura the equality check was intentional, because for some serdes (e.g. LI's AvroSpecificRecordSerde, JsonSerdeV2) two instances of the same class may not be equal since they only act on specific message types. However relaxing the check to check for type equality is fine for now.

dnishimura:
Thanks for the review @prateekm !
In case you're waiting for the Travis-CI build to complete to merge this change, it's already completed: https://travis-ci.org/apache/samza/builds/496764468
I'm not sure why it still shows as pending in github.

PR #934, opened 2019-02-27
SAMZA-1935 : Refactor TaskContextImpl to not include access to objects that are only used internally

vwidin:
We created a new class called JobContextMetadata which hold the methods `registerObject`, `fetchObject`, `getJobModel` and `getStreamMetadataCache` and is instantiated in the `OperatorImplGraph` and is passed to multiple `OperatorImpl`. 

The class `EmbeddedTaggedRateLimiter` implements the public API `RateLimiter`, which has a method called init with a `Context` object as inparameter. This means that the Context object is the only source from which the JobModel and the `StreamMetadataCache` can be fetched. Hence, the methods `getJobModel` and `getStreamMetadataCache` are needed outside of `JobContextMetadata`. Therefore only copies of these methods were moved to `JobContextMetadata`. This means that the `StreamMetadataCache` and the `JobModel` are created with the `Context` therefore the `JobContextMetadata` is initiated with the `StreamMetadataCache` and `JobModel` retrieved by calling `getJobModel` and `getStreamMetadataCache` on the `TaskContextImpl`, so there are still casts of the `TaskContext` to the `TaskContextImpl`.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/context/JobContextMetadata.java
+++ b/samza-core/src/main/java/org/apache/samza/context/JobContextMetadata.java
@@ -1,0 +1,35 @@
+package org.apache.samza.context;
+
+import org.apache.samza.job.model.JobModel;
+import org.apache.samza.system.StreamMetadataCache;
+
+import java.util.HashMap;
+import java.util.Map;
+
+
+public class JobContextMetadata {
+    private final JobModel jobModel;
+    private final StreamMetadataCache streamMetadataCache;
+    private final Map<String, Object> objectRegistry = new HashMap<>();
+
+    public JobContextMetadata(JobModel jobModel, StreamMetadataCache streamMetadataCache) {
+        this.jobModel = jobModel;
+        this.streamMetadataCache = streamMetadataCache;
+    }
+
+    public void registerObject(String name, Object value) {
+        this.objectRegistry.put(name, value);
+    }
+
+    public Object fetchObject(String name) {
+        return this.objectRegistry.get(name);
+    }
+
+    public JobModel getJobModel() {
+        return this.jobModel;
+    }
+
+    public StreamMetadataCache getStreamMetadataCache() {
+        return this.streamMetadataCache;
+    }
+}
--- a/samza-core/src/main/java/org/apache/samza/context/JobContextMetadata.java
+++ b/samza-core/src/main/java/org/apache/samza/context/JobContextMetadata.java
@@ -1,3 +1,22 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
 package org.apache.samza.context;
 
 import org.apache.samza.job.model.JobModel;
--- a/samza-core/src/main/java/org/apache/samza/context/JobContextMetadata.java
+++ b/samza-core/src/main/java/org/apache/samza/context/JobContextMetadata.java
@@ -27,28 +27,28 @@ import java.util.Map;
 
 
 public class JobContextMetadata {
-    private final JobModel jobModel;
-    private final StreamMetadataCache streamMetadataCache;
-    private final Map<String, Object> objectRegistry = new HashMap<>();
+  private final JobModel jobModel;
+  private final StreamMetadataCache streamMetadataCache;
+  private final Map<String, Object> objectRegistry = new HashMap<>();
 
-    public JobContextMetadata(JobModel jobModel, StreamMetadataCache streamMetadataCache) {
-        this.jobModel = jobModel;
-        this.streamMetadataCache = streamMetadataCache;
-    }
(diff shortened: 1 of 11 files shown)


prateekm:
Thanks for the contribution @vwidin. cc @cameronlee314 for review.

samza-core/src/test/java/org/apache/samza/context/TestJobContextMetadata.java
```diff
@@ -0,0 +1,51 @@
...
+package org.apache.samza.context;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class TestJobContextMetadata {
+
```
cameronlee314: Minor: Looks like you have a couple extra lines of whitespace here

samza-core/src/test/java/org/apache/samza/operators/impl/TestOperatorImpl.java
```diff
@@ -47,10 +48,12 @@
 
 public class TestOperatorImpl {
   private Context context;
+  private JobContextMetadata jobContextMetadata;
 
   @Before
   public void setup() {
     this.context = new MockContext();
+    this.jobContextMetadata = new JobContextMetadata(null, null);
```
cameronlee314: Consider mocking this object with Mockito instead of using a real instance. It makes the test a bit more resilient to changes to JobContextMetadata in the future.

samza-core/src/test/java/org/apache/samza/operators/impl/TestWindowOperator.java
```diff
@@ -436,13 +434,6 @@ public void testCancelationOfRepeatingNestedTriggers() throws Exception {
 
   @Test
   public void testEndOfStreamFlushesWithEarlyTriggerFirings() throws Exception {
-    EndOfStreamStates endOfStreamStates = new EndOfStreamStates(ImmutableSet.of(new SystemStreamPartition("kafka",
```
cameronlee314: Just double checking: It looks like you removed some code but did not replace it with anything. Was this code unnecessary for the test?

samza-core/src/main/java/org/apache/samza/context/JobContextMetadata.java
```diff
@@ -0,0 +1,54 @@
...
+package org.apache.samza.context;
+
+import org.apache.samza.job.model.JobModel;
+import org.apache.samza.system.StreamMetadataCache;
+
+import java.util.HashMap;
+import java.util.Map;
+
+
+public class JobContextMetadata {
```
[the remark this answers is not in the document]
Cpptz: Indeed, it is much better

samza-core/src/test/java/org/apache/samza/operators/impl/TestWindowOperator.java
Cpptz: Well first of all, as the JobContextMetadata is instanciated inside the OperatorImplGraph, we have no reference to it in this test, therefore we didn't know how to mock it.
And infact, we found that the mocking wasn't necessary if we simply set the correct stream name in the setup, namely `integers` instead of  `integTestExecutionPlannerers` since tests are done with the stream name `integers`

Cpptz:
Thank you very much for the review, we will clean code according to your comments .

samza-core/src/test/java/org/apache/samza/operators/impl/TestWindowOperator.java
cameronlee314: Sounds good. If it's not needed for the test, it is best to remove it.

In the future, if you ever run into a use case where you do need to mock something that is instantiated inside the constructor, here are a couple of things you could consider:
Create a second package-private constructor in the class which accepts the object you need to mock. The test class should have the same package as the class under test, so the test class should be able to access this second constructor. Guava has an annotation called @VisibleForTesting which you could use to mark this second constructor. Any tests can use this second constructor to inject a mock. Unfortunately, this pattern does mean you need some test-only code in your class, which is generally not ideal.
Another thing you could look into is refactoring the class so that it does not instantiate anything inside the constructor. Everything it needs gets passed in directly. In my opinion, this is probably the ideal state, but sometimes it is not practical or not possible to refactor something that way.

samza-core/src/main/java/org/apache/samza/context/JobContextMetadata.java
```diff
@@ -44,11 +42,14 @@ public Object fetchObject(String name) {
...
-  public JobModel getJobModel() {
-    return this.jobModel;
+  public Context getContext() {
+    return context;
   }
 
+  public JobModel getJobModel() {
+    return ((TaskContextImpl) this.context.getTaskContext()).getJobModel();
+  }
   public StreamMetadataCache getStreamMetadataCache() {
```
cameronlee314: Minor: Please add a line of whitespace between method implementations.

samza-core/src/main/java/org/apache/samza/operators/impl/OperatorImplGraph.java
```diff
@@ -85,7 +101,7 @@ public OperatorImplGraph(OperatorSpecGraph specGraph, Context context, Clock clo
     this.clock = clock;
     StreamConfig streamConfig = new StreamConfig(context.getJobContext().getConfig());
     TaskContextImpl taskContext = (TaskContextImpl) context.getTaskContext();
```
cameronlee314: I think this variable is unused now.

samza-core/src/main/java/org/apache/samza/context/JobContextMetadata.java
```diff
@@ -0,0 +1,55 @@
...
+
+package org.apache.samza.context;
+
+import org.apache.samza.job.model.JobModel;
+import org.apache.samza.system.StreamMetadataCache;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class JobContextMetadata {
```
cameronlee314: Apologies for missing this earlier: I just realized that the `objectRegistry` is specific to each task here, not the whole job. Therefore, could you please name this object (and its usages) to something that reflects the task scope? Maybe `InternalTaskContext`, since it is not part of the Samza public API.

Cpptz: Done

PR #935, opened 2019-03-04
SAMZA-2118: Improve the shutdown sequence of AsyncRunLoop.

shanthoosh:
A samza container is comprised of multiple tasks and event-loop. Event loop is the main orchestration layer between the different stages of a task(process, commit, window) and delivers messages to task by polling them from different consumers. 

Currently in samza, the event loop dispatches messages to the tasks as per user-defined configuration `task.max.concurrency`. After dispatching messages with count equal to `task.max.concurrency` to the tasks,  the event loop is blocked until the processing of at-least a single dispatched message finishes(aka the task implementation invokes `taskCallBack.complete()` for at-least a dispatched message). 

When the event loop is blocked by this, the user thread would sometimes try to shutdown the event loop. This shutdown intent marked by the user-thread will take effect only when the processing of a single message completes. In some-cases, when the task implementation does retry on failure for a message, this could prolong the shutdown of the samza event loop.

This introduces tight coupling between the task processing time and the event loop shutdown time.

This behavior was observed predominantly in standalone applications where the samza container is shutdown as a part of the re-balancing phase and task processing time was significantly larger.

To alleviate this problem, as a part of this patch, we wake-up the samza event-loop thread when the user-thread marks the intent to shutdown the event-loop.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
+++ b/samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
@@ -193,6 +193,7 @@ public class AsyncRunLoop implements Runnable, Throttleable {
 
   public void shutdown() {
     shutdownNow = true;
+    resume();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -193,6 +193,7 @@ public double getWorkFactor() {
 
   public void shutdown() {
     shutdownNow = true;
+    resume();
```
prateekm: Update javadoc for resume.

shanthoosh:
@xinyuiscool 

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
shanthoosh: Done.

PR #938, opened 2019-03-05
SAMZA-1531: Support run.id in standalone for batch processing.

lakshmi-manasa-g:
Adds run.id in standalone: only for BATCH mode and no changes for STREAM mode.

Known issues:

- No support for Azure in this PR. a follow up PR will be made for it. However, this PR does not break Azure/Passthrough and resorts to current behavior for Azure/Passthrough.

- upon an unclean shut down of the job, a restart before ZK server session timeout would result in reuse of the old runid.


samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -280,4 +368,47 @@ private void setApplicationFinalStatus() {
...
+        return;
+      }
+      DistributedDataAccess runIdAccess = coordinationUtils.getDataAccess();
+      String globalRunId = (String) runIdAccess.readData(RUNID_PATH);
+      if( runId != globalRunId){
+        processors.forEach(StreamProcessor::stop);
+        cleanup();
+        appStatus = ApplicationStatus.UnsuccessfulFinish;
+        String msg = String.format("run.id %s on processor %s differs from the global run.id %s", runId, uid, globalRunId);
+        throw new SamzaException(msg);
```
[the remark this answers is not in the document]
lakshmi-manasa-g: Done. doing shutdownLatch.countDown() also.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -280,4 +368,47 @@ private void setApplicationFinalStatus() {
...
+          // acquire lock to recreate active processor ephemeral node
+          DistributedReadWriteLock.AccessType lockAccess = runIdLock.lock(LOCK_TIMEOUT, LOCK_TIMEOUT_UNIT);
+          if(lockAccess == DistributedReadWriteLock.AccessType.WRITE || lockAccess == DistributedReadWriteLock.AccessType.READ) {
+            LOG.info("Processor {} creates its active processor ephemeral node in read write lock again" + uid);
+            runIdLock.unlock();
+          } else {
+            throw new SamzaException(msg);
+          }
+        } catch (TimeoutException e) {
+          throw new SamzaException(msg, e);
```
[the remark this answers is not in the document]
lakshmi-manasa-g: Done.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -280,4 +368,47 @@ private void setApplicationFinalStatus() {
...
+        cleanup();
+        appStatus = ApplicationStatus.UnsuccessfulFinish;
+        String msg = String.format("run.id %s on processor %s differs from the global run.id %s", runId, uid, globalRunId);
+        throw new SamzaException(msg);
+      } else if(runIdLock != null) {
+        String msg = String.format("Processor {} failed to get the lock for run.id", uid);
+        try {
+          // acquire lock to recreate active processor ephemeral node
+          DistributedReadWriteLock.AccessType lockAccess = runIdLock.lock(LOCK_TIMEOUT, LOCK_TIMEOUT_UNIT);
+          if(lockAccess == DistributedReadWriteLock.AccessType.WRITE || lockAccess == DistributedReadWriteLock.AccessType.READ) {
```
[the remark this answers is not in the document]
lakshmi-manasa-g: Session expiration does delete all the ephemeral nodes are cleaned up. But according to the ZkClient code, after an expiration two things can happen -- a successful reconnect in which case listener.handleNewSession is invoked or failure to reconnect in which case listener.handleSessionEstablishmentError is invoked. Handling both of these in LocalApplicationRunner now.

samza-core/src/main/java/org/apache/samza/coordinator/CoordinationUtils.java
```diff
@@ -39,6 +42,12 @@
 
   DistributedLockWithState getLockWithState(String lockId);
 
+  DistributedReadWriteLock getReadWriteLock();
+
+  DistributedDataAccess getDataAccess();
+
+  void setCoordinationSessionListener(CoordinationSessionListener sessionListener);
```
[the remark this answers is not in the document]
lakshmi-manasa-g: done.

samza-core/src/main/java/org/apache/samza/coordinator/CoordinationUtils.java
```diff
@@ -39,6 +42,12 @@
 
   DistributedLockWithState getLockWithState(String lockId);
 
+  DistributedReadWriteLock getReadWriteLock();
```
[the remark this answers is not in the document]
lakshmi-manasa-g: Added lock id.

samza-core/src/main/java/org/apache/samza/execution/LocalJobPlanner.java
```diff
@@ -119,7 +119,12 @@ private void createStreams(String planId, List<StreamSpec> intStreams, StreamMan
       return;
     }
 
-    DistributedLockWithState lockWithState = coordinationUtils.getLockWithState(planId);
+    Boolean isAppModeBatch = new ApplicationConfig(userConfig).getAppMode() == ApplicationConfig.ApplicationMode.BATCH;
+    String lockId = planId;
+    if(isAppModeBatch && runId != null) {
```
sborya: add comment

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -59,13 +67,23 @@
...
+  private static final TimeUnit LOCK_TIMEOUT_UNIT = TimeUnit.MINUTES;
 
   private final ApplicationDescriptorImpl<? extends ApplicationDescriptor> appDesc;
   private final LocalJobPlanner planner;
   private final Set<StreamProcessor> processors = ConcurrentHashMap.newKeySet();
   private final CountDownLatch shutdownLatch = new CountDownLatch(1);
   private final AtomicInteger numProcessorsToStart = new AtomicInteger();
   private final AtomicReference<Throwable> failure = new AtomicReference<>();
+  private final String uid = UUID.randomUUID().toString();
+  private CoordinationUtils coordinationUtils = null;
```
sborya: make final.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -59,13 +67,23 @@
...
+  private DistributedReadWriteLock runIdLock = null;
+  private DistributedDataAccess runIdAccess = null;
+  private String runId = null;
```
sborya: make final. Assign null if needed in the constructor.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -77,23 +95,80 @@
...
+
+  private void getRunId(){
+    Boolean isAppModeBatch = new ApplicationConfig(appDesc.getConfig()).getAppMode() == ApplicationConfig.ApplicationMode.BATCH;
+    if(coordinationUtils == null ||  !isAppModeBatch) {
+      return;
+    }
+
+    runIdLock = coordinationUtils.getReadWriteLock(RUNID_LOCK_ID);
+    if(runIdLock == null) {
+      LOG.warn("Processor {} failed to create the lock for run.id generation", uid);
```
sborya: better to find more specific info than UID, uid is not very helpful.

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedReadWriteLock.java
```diff
@@ -0,0 +1,233 @@
...
+import org.apache.samza.coordinator.DistributedReadWriteLock;
+import org.apache.zookeeper.Watcher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Distributed lock primitive for Zookeeper.
+ */
+public class ZkDistributedReadWriteLock implements DistributedReadWriteLock {
```
sborya: Is this a new code or moved from somewhere?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -200,6 +279,22 @@ static String createProcessorId(ApplicationConfig appConfig) {
...
+    }
+    if(coordinationUtils != null) {
+      coordinationUtils.close();
+    }
+  }
+
+  private void stopProcessingAndShutDown() {
+    processors.forEach(StreamProcessor::stop);
+    cleanup();
+    appStatus = ApplicationStatus.UnsuccessfulFinish;
```
sborya: qq. why is it Unsuccessful?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -280,4 +376,68 @@ private void setApplicationFinalStatus() {
...
+   */
+  private final class LocalDistributedDataStateListener implements DistributedDataStateListener {
+
+    /**
+     * upon reconnect check if global runid differs from local runid
+     */
+    @Override
+    public void handleReconnect() {
+      if(coordinationUtils == null || runIdLock == null || runIdAccess == null ) {
+        LOG.warn("Stopping processor {} and shutting down due to failure reading global runid after reconnect", uid);
```
sborya: where is it reading the global runid?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -280,4 +376,68 @@ private void setApplicationFinalStatus() {
...
+    }
+  }
+
+  /**
+   * Defines a specific implementation of {@link DistributedDataWatcher} for local {@link DistributedDataAccess}
+   */
+  private final class LocalDistributedDataWatcher implements DistributedDataWatcher {
+    @Override
+    public void handleDataChange(Object newData) {
+      if(runId != (String) newData) {
```
sborya: you should use equal()

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedDataAccess.java
```diff
@@ -0,0 +1,61 @@
...
+
+  public ZkDistributedDataAccess(ZkUtils zkUtils) {
+    if (zkUtils == null) {
+      throw new RuntimeException("Cannot operate ZkDistributedDataAccess without ZkUtils.");
+    }
+    this.zkUtils = zkUtils;
+    this.keyBuilder = zkUtils.getKeyBuilder();
+  }
+
+  public Object readData(String key, DistributedDataWatcher watcher) {
```
sborya: Won't watcher override the one set in writeData? If they are the same (or only one should be there) then make it final and pass it in the constructor.

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedDataAccess.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  class ZkDistributedDataChangeHandler extends ZkUtils.GenerationAwareZkDataListener {
+
+    public ZkDistributedDataChangeHandler(ZkUtils zkUtils) {
+      super(zkUtils, "ZkDistributedDataChangeHandler");
+    }
+
+    /**
+     * Invoked when there is a change to the  data z-node.
```
sborya: nit. extra space.

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedReadWriteLock.java
```diff
@@ -0,0 +1,233 @@
...
+  private final ZkUtils zkUtils;
+  private final String lockPath;
+  private final String particpantsPath;
+  private final String processorsPath;
+  private final String participantId;
+  private final ZkKeyBuilder keyBuilder;
+  private final Random random = new Random();
+  private String activeParticipantPath = null;
+  private String activeProcessorPath = null;
+  private Object mutex;
```
sborya: final

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedReadWriteLock.java
```diff
@@ -0,0 +1,233 @@
...
+      synchronized (mutex) {
+        AccessType accessType = checkAndAcquireLock();
+        if(accessType != AccessType.NONE) {
+          isInCriticalSection = true;
+          if(isStateLost) {
+            throw new SamzaException("Lock's state lost due to connection expiry");
+          }
+          return accessType;
+        } else {
+          if(isStateLost) {
```
sborya: can this be moved out of the if()?

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedReadWriteLock.java
```diff
@@ -0,0 +1,233 @@
...
+    while((System.currentTimeMillis() - startTime) < lockTimeout) {
+      synchronized (mutex) {
+        AccessType accessType = checkAndAcquireLock();
+        if(accessType != AccessType.NONE) {
+          isInCriticalSection = true;
+          if(isStateLost) {
+            throw new SamzaException("Lock's state lost due to connection expiry");
+          }
+          return accessType;
+        } else {
```
sborya: when this happens?

samza-azure/src/main/java/org/apache/samza/coordinator/AzureCoordinationUtils.java
```diff
@@ -52,6 +52,16 @@ public DistributedLockWithState getLockWithState(String lockId) {
     return new AzureLock(blob);
   }
 
+  @Override
+  public DistributedReadWriteLock getReadWriteLock(String lockId) throws UnsupportedOperationException {
```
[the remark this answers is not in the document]
lakshmi-manasa-g: getReadWriteLock() is used in LocalApplicationRunner (LAR) when app.mode=BATCH. If we throw an exception here instead of returning null, none of the standalone deployments for BATCH mode using Azure will even work. 

Currently, no run.id is generated for standalone (for both ZK and Azure) and thus, if we return null here instead of an exception, it would revert to the existing behavior until we support run.id for Azure.

Would it help if we log a warning explaining that run.id will not be generated (meaning previous state will be reused) and this batch job wont starting with fresh state?

samza-azure/src/main/java/org/apache/samza/coordinator/AzureCoordinationUtils.java
```diff
@@ -52,6 +52,16 @@ public DistributedLockWithState getLockWithState(String lockId) {
...
+    return null;
+  }
+
+  @Override
+  public DistributedDataAccess getDataAccess(DistributedDataStateListener listener) throws UnsupportedOperationException {
```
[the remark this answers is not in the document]
lakshmi-manasa-g: Same as above.

samza-core/src/main/java/org/apache/samza/coordinator/DistributedDataStateListener.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package org.apache.samza.coordinator;
+
+/**
+ * Listen to changes in state of distributed data connection
+ */
+public interface DistributedDataStateListener {
+  /**
+   * Called when the connected has reconnected after a disconnect.
```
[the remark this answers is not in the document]
lakshmi-manasa-g: Thank you for pointing this out. However, to clean up LAR, removing this listener altogether.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -59,13 +67,23 @@
 public class LocalApplicationRunner implements ApplicationRunner {
 
   private static final Logger LOG = LoggerFactory.getLogger(LocalApplicationRunner.class);
+  private static final String RUNID_PATH = "runId";
+  private static final String APPLICATION_RUNNER_PATH_SUFFIX = "/ApplicationRunnerData";
+  private static final String RUNID_LOCK_ID = "runId";
+  private static final int LOCK_TIMEOUT = 10;
```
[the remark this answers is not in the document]
lakshmi-manasa-g: Done.

samza-core/src/main/java/org/apache/samza/execution/LocalJobPlanner.java
lakshmi-manasa-g: done. please let me know if the comment added is clear.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
lakshmi-manasa-g: same issue as above. runIdLock and runIdAccess are not inited in case of STREAM and not in the constructor. runId is also inited in a separate method and not the constructor.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
lakshmi-manasa-g: Actually, I am using uid is several Log statements. pl let me know if i should remove it out of all of them. 
here there are two cases runIdLock could be null -- ZK failure which will anways throw an exception trying to create ZkDistributedReadWriteLock or in Azure where it is not supported. For Azure, we really dont have any other info to provide.

Would it help to provide an elaborate warning -- something like runid lock creation failed and that old run's state (intermediate streams etc) will be used?

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedReadWriteLock.java
lakshmi-manasa-g: this is new code

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
lakshmi-manasa-g: stopProcessingAndShutDown is called when the global runid differs from this processor's runid or if the runid node itself is deleted. In the first case, it is a failure as this processor was using different intermediate streams than the rest of the processors and in the latter, there is no way for processors to agree upon the streams to use. THus, is it an unsuccessful run and needs to be indicated.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
lakshmi-manasa-g: actually meant to say it could not read global runid. HOwever, this code is now moved out of LAR to make it cleaner

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
lakshmi-manasa-g: Yes, figured that out when i spun a processor with this code. THanks for pointing it out.

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedDataAccess.java
lakshmi-manasa-g: Actually, a watcher is specific to a key/path and hence can not be set in the constructor (as we do not know what the key will be). ZkDistributedDataAccess is used to provide access to (any/all) data in ZK and hence the constructor does not take in a key. 

here, I am not making in generic enough to use this class to access data for any number of keys. Updating this code to keep a map of <key,watcher> so that several keys could be accessed. However, the issue you mentioned would still be there.. as in for a key if a watcher is passed during readData and then a different watcher is passed in during writeData then it does get overwritten. I believe that is correct as we should associate a watcher with a key and not to the specific way (read vs write) the data at the key was accessed. I do not forsee a necessity for differentiating between read/write when watching the data. please let me know if you think otherwise.

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedDataAccess.java
lakshmi-manasa-g: actually removing this comment as it feels uncessary

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedReadWriteLock.java
lakshmi-manasa-g: made mutex final. 
the paths (activeParticipantPath and activeProcessorPath) can not be inited in the constructor and are set to null when the corresponding ephemeral nodes are deleted. Thus these can not be made final.

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedReadWriteLock.java
lakshmi-manasa-g: unfortunately no. cause, isStateLost is set to true in some cases only if isInCriticalSection is true.

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedReadWriteLock.java
lakshmi-manasa-g: As in what is the scenario when this exception is thrown?

Suppose processor P1 is trying to acquire this lock. It is the first participant and the only one in the processors list and thus would have accessType = WRITE. Now, if there is a connection expiry then ZK will clean up all its ephemeral nodes meaning its participant and processor node will be deleted leading to another processor P2 acquiring the lock as WRITE. P2 would go ahead and write a run.id (say r2). Now, if we did not throw an exception in P1 and instead returned WRITE, then P1 will go ahead (upon connection restore) will overwrite the run.id (say r1). Thus P1 and P2 are using two different run.ids. It is a corner case.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
mynameborat: I see why this is a confusion. can we take in an exception instead and set the state based on the exception?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -77,16 +90,70 @@
...
+        runId = (String) runIdAccess.readData(CoordinationConstants.RUNID_PATH, new LocalDistributedDataWatcher());
+        runIdLock.unlock();
+      } else {
+        String msg = String.format("Processor {} failed to get the lock for run.id", uid);
+        throw new SamzaException(msg);
+      }
+    } catch (TimeoutException e) {
+      String msg = String.format("Processor {} timed out waiting to acquire lock for run.id generation", uid);
+      throw new SamzaException(msg, e);
+    }
```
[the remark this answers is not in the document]
lakshmi-manasa-g: done.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -77,16 +90,70 @@
...
+      LOG.warn("Processor {} failed to create utils for run.id generation", uid);
+      return;
+    }
+
+    try {
+      // acquire lock to write or read run.id
+      DistributedReadWriteLock.AccessType lockAccess = runIdLock.lock(CoordinationConstants.LOCK_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      if(lockAccess == DistributedReadWriteLock.AccessType.WRITE) {
+        LOG.info("write lock acquired for run.id generation by Processor " + uid);
+        runId = String.valueOf(System.currentTimeMillis()) + "-" + UUID.randomUUID().toString().substring(0, 8);
```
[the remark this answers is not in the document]
lakshmi-manasa-g: It would not lead to a collision cause the runid is taken to be SystemTime in millisecs and then the UUID substring.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -77,16 +90,70 @@
...
+  }
+
+  private void initializeCoordinationUtils(Config config) {
+    JobCoordinatorConfig jcConfig = new JobCoordinatorConfig(config);
+    String coordinationId = new ApplicationConfig(config).getGlobalAppId() + CoordinationConstants.APPLICATION_RUNNER_PATH_SUFFIX;
+    this.coordinationUtils = jcConfig.getCoordinationUtilsFactory().getCoordinationUtils(coordinationId, uid, config);
+  }
+
+  private void initializeRunId(){
+    LOG.info("Manasa: LocalApplicationRunner: getRunId() : entered ");
```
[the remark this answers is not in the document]
lakshmi-manasa-g: done. thanks for catching this.. was meant for testing purposes :)

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -77,16 +90,70 @@
    */
   public LocalApplicationRunner(SamzaApplication app, Config config) {
     this.appDesc = ApplicationDescriptorUtil.getAppDescriptor(app, config);
-    this.planner = new LocalJobPlanner(appDesc);
+    Boolean isAppModeBatch = new ApplicationConfig(appDesc.getConfig()).getAppMode() == ApplicationConfig.ApplicationMode.BATCH;
+    if(isAppModeBatch) {
+      initializeCoordinationUtils(config);
+      initializeRunId();
+    }
+    this.planner = new LocalJobPlanner(appDesc, coordinationUtils, uid, runId);
```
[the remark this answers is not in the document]
lakshmi-manasa-g: done.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -77,16 +90,70 @@
    */
   public LocalApplicationRunner(SamzaApplication app, Config config) {
     this.appDesc = ApplicationDescriptorUtil.getAppDescriptor(app, config);
-    this.planner = new LocalJobPlanner(appDesc);
+    Boolean isAppModeBatch = new ApplicationConfig(appDesc.getConfig()).getAppMode() == ApplicationConfig.ApplicationMode.BATCH;
+    if(isAppModeBatch) {
+      initializeCoordinationUtils(config);
```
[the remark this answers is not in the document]
lakshmi-manasa-g: done. using Optional.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -77,16 +90,70 @@
    */
   public LocalApplicationRunner(SamzaApplication app, Config config) {
     this.appDesc = ApplicationDescriptorUtil.getAppDescriptor(app, config);
-    this.planner = new LocalJobPlanner(appDesc);
+    Boolean isAppModeBatch = new ApplicationConfig(appDesc.getConfig()).getAppMode() == ApplicationConfig.ApplicationMode.BATCH;
```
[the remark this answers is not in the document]
lakshmi-manasa-g: done.

samza-core/src/main/java/org/apache/samza/execution/LocalJobPlanner.java
```diff
@@ -119,7 +124,17 @@ private void createStreams(String planId, List<StreamSpec> intStreams, StreamMan
...
 
-    DistributedLockWithState lockWithState = coordinationUtils.getLockWithState(planId);
+    // If BATCH, then need to create new intermediate streams every run.
+    // planId does not change every run and hence, need to use runid
+    // as the lockId to create a new lock with state each run
+    // to create new streams each run.
+    // If run.id is null, defaults to old behavior of using planId
+    Boolean isAppModeBatch = new ApplicationConfig(userConfig).getAppMode() == ApplicationConfig.ApplicationMode.BATCH;
+    String lockId = planId;
+    if(isAppModeBatch && runId != null) {
```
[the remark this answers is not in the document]
lakshmi-manasa-g: UnsupportedOperation is thrown in LocalApplicationRunner and not in LocalJobPlanner. Thus there is no way to know why runId is null and hence keeping this as is.

samza-core/src/main/java/org/apache/samza/execution/LocalJobPlanner.java
```diff
@@ -119,7 +124,17 @@ private void createStreams(String planId, List<StreamSpec> intStreams, StreamMan
       return;
     }
 
-    DistributedLockWithState lockWithState = coordinationUtils.getLockWithState(planId);
+    // If BATCH, then need to create new intermediate streams every run.
+    // planId does not change every run and hence, need to use runid
+    // as the lockId to create a new lock with state each run
+    // to create new streams each run.
+    // If run.id is null, defaults to old behavior of using planId
+    Boolean isAppModeBatch = new ApplicationConfig(userConfig).getAppMode() == ApplicationConfig.ApplicationMode.BATCH;
```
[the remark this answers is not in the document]
lakshmi-manasa-g: done.

samza-core/src/main/java/org/apache/samza/coordinator/CoordinationConstants.java
```diff
@@ -0,0 +1,32 @@
...
+package org.apache.samza.coordinator;
+
+import java.util.concurrent.TimeUnit;
+
+
+public class CoordinationConstants {
+  private CoordinationConstants() {}
+
+  public static final String RUNID_PATH = "runId";
+  public static final String APPLICATION_RUNNER_PATH_SUFFIX = "/ApplicationRunnerData";
```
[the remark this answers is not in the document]
lakshmi-manasa-g: done.

samza-core/src/main/java/org/apache/samza/execution/LocalJobPlanner.java
```diff
@@ -107,10 +111,11 @@ private void createStreams(String planId, List<StreamSpec> intStreams, StreamMan
     LOG.info("A single processor must create the intermediate streams. Processor {} will attempt to acquire the lock.", uid);
     // Move the scope of coordination utils within stream creation to address long idle connection problem.
     // Refer SAMZA-1385 for more details
-    JobCoordinatorConfig jcConfig = new JobCoordinatorConfig(userConfig);
-    String coordinationId = new ApplicationConfig(userConfig).getGlobalAppId() + APPLICATION_RUNNER_PATH_SUFFIX;
-    CoordinationUtils coordinationUtils =
-        jcConfig.getCoordinationUtilsFactory().getCoordinationUtils(coordinationId, uid, userConfig);
+    if(coordinationUtils == null) {
+      JobCoordinatorConfig jcConfig = new JobCoordinatorConfig(appDesc.getConfig());
+      String coordinationId = new ApplicationConfig(appDesc.getConfig()).getGlobalAppId() + CoordinationConstants.APPLICATION_RUNNER_PATH_SUFFIX;
```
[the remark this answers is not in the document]
lakshmi-manasa-g: done.

samza-core/src/main/java/org/apache/samza/coordinator/CoordinationUtils.java
```diff
@@ -39,6 +42,10 @@
 
   DistributedLockWithState getLockWithState(String lockId);
 
+  DistributedReadWriteLock getReadWriteLock(String lockId);
+
+  DistributedDataAccess getDataAccess(DistributedDataStateListener listener);
```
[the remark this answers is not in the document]
lakshmi-manasa-g: updated to take a listener per key while reading/writing.

samza-azure/src/main/java/org/apache/samza/coordinator/AzureCoordinationUtils.java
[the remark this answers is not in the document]
lakshmi-manasa-g: done.

samza-core/src/main/java/org/apache/samza/coordinator/DistributedReadWriteLock.java
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Release the lock
+   */
+  void unlock();
+
+  /**
+   * Clean state of the lock
+   */
+  void cleanState();
```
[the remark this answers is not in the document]
lakshmi-manasa-g: done.

samza-core/src/main/java/org/apache/samza/coordinator/DistributedDataAccess.java
```diff
@@ -0,0 +1,42 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.coordinator;
+
+/**
+ * Concurrent writes are not allowed but concurrent reads are.
```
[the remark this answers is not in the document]
lakshmi-manasa-g: Gave it a shot. Please leave another comment if it doesnt look okay. resolving this comment.

samza-core/src/main/java/org/apache/samza/coordinator/DistributedReadWriteLock.java
```diff
@@ -0,0 +1,70 @@
...
+
+
+/**
+ * Lock to acquire read access or write access
+ * At any point in time, only one processor can hold the lock
+ * The processor that has acquired the lock, holds it until it does `unlock` explicitly
+ * The first processor to acquire the lock should be given a WRITE access
+ * All other subsequent requests should acquire only READ access
+ */
+public interface DistributedReadWriteLock {
```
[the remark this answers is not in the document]
lakshmi-manasa-g: yes, that would be nice to have. But for now keeping this as is to wrap up sooner.

vjagadish1989:
@bharathkk @lakshmi-manasa-g 

A quick summary of next steps: The current implementation is way too complex for the problem at hand. 

Part of the complexity stems from (i) Added interfaces to coordinationUtils that are too specific to this feature (ii) combining coordination and state management as a part of the same interface eg: this leads to pathologies like `DistributedLockWithState`

Instead, we should compose new features using these building-blocks:
1. State-management: interface across k-v read-writes across processors; This should be `MetadataStore`
2. Cluster-membership: interface for queries to membership of the cluster + getting notified participants change. 
3. Coordination: interface for doing an atomic operation across processors ie., distributed-lock

With the above goal, we will:
- Rip apart state-management from the existing `DistributedLockWithState` interface and make it a `DistributedLock`; 
- Move all state-management logic to use the `MetadataStore`
- Add support to the `DistributedLock` interface to provide notifications during disconnects/lock revocations. 
- Use `DistributedLock` consistently wherever we require distributed atomic operations in the code-base. 
- Nuke the `DistributedReadWriteLock` interface and its corresponding implementation; In its current form, its semantics are hard to reason with and there appears to be no use-case that can't be solved with the `DistributedLock` 
- Add a new interface to coordinationUtils to query cluster-membership + getting notified for ongoing changes
- Combine these building-blocks and generate run.id at each processor by performing the following actions within the critical-section:
* Register as a participant and use the cluster-membership interface to query membership
* If it's the first processor in the fleet, generate and write run.id to the metadata-store
* If not, read run.id from metadata-store
- As a simpler first solution, any lock revocation due to a Zk-error can shutdown the process. We can explore fancier ways of handling re-registration later

samza-core/src/main/java/org/apache/samza/coordinator/DistributedReadWriteLock.java
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * Try to acquire lock: returns the type of access lock was acquired for
+   * @param timeout Duration of lock acquiring timeout.
+   * @param unit Time Unit of the timeout defined above.
+   * @return AccessType.READ if lock is acquired for read, AccessType.WRITE if acquired for write
+   * @throws TimeoutException if could not acquire the lock.
+   */
+  AccessType lock(long timeout, TimeUnit unit) throws TimeoutException;
```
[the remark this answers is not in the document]
lakshmi-manasa-g: removed this lock as per suggestion.

samza-core/src/main/java/org/apache/samza/coordinator/DistributedReadWriteLock.java
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * Release the lock
+   */
+  void unlock();
+
+  /**
+   * Clean state of the lock
```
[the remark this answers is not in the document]
lakshmi-manasa-g: removed this lock as per suggestion

samza-core/src/main/java/org/apache/samza/coordinator/DistributedDataAccess.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public interface DistributedDataAccess {
+  /**
+   * Read data associated with key.
+   * Does not modify the value at key.
+   * @param key
+   * @param watcher through which notifications of data changes at key are sent
+   * @return data present at key
+   */
+  Object readData(String key, DistributedDataWatcher watcher);
```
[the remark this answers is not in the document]
lakshmi-manasa-g: Removed this API as per suggestion

samza-core/src/main/java/org/apache/samza/coordinator/DistributedReadWriteLock.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+package org.apache.samza.coordinator;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+
+/**
+ * Lock to acquire read access or write access.
```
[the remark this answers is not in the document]
lakshmi-manasa-g: Removed this API as per suggestion.
Now replaced with a simple lock, cluster membership and metadata store

samza-core/src/main/java/org/apache/samza/coordinator/DistributedDataAccess.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * DistributedDataAccess is an interface for accessing data
+ * shared between several processors in the quorum.
+ * 1. Not thread safe
+ * 2. No internal synchronization
+ * 3. A watcher is registered to listen to data changes at the key.
+ */
+public interface DistributedDataAccess {
+  /**
+   * Read data associated with key.
```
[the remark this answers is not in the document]
lakshmi-manasa-g: removed this api as per suggestion

samza-core/src/main/java/org/apache/samza/coordinator/CoordinationUtils.java
```diff
@@ -39,6 +41,10 @@
 
   DistributedLockWithState getLockWithState(String lockId);
 
+  DistributedReadWriteLock getReadWriteLock(String lockId);
+
+  DistributedDataAccess getDataAccess();
```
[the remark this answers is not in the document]
lakshmi-manasa-g: using metadata store as per suggestion.

samza-core/src/main/java/org/apache/samza/coordinator/DistributedReadWriteLock.java
```diff
@@ -0,0 +1,82 @@
...
+ * The processor that has acquired the lock, holds it until it does `unlock` explicitly.
+ * If there is only the one processor in the quorum and it is trying to acquire the lock
+ * then it should be given a WRITE access.
+ * All other requests should acquire only READ access
+ *
+ * `State` is information pertaining to the requesting processor maintained by the lock
+ * to determine how to grant read or write access
+ *
+ *
+ *                                       lock()                     unlock()                   cleanState()
```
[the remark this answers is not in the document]
lakshmi-manasa-g: removed this api as per suggestion.

samza-core/src/main/java/org/apache/samza/coordinator/DistributedReadWriteLock.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+
+/**
+ * Lock to acquire read access or write access.
+ * At any point in time, only one processor can hold the lock.
+ * The processor that has acquired the lock, holds it until it does `unlock` explicitly.
+ * If there is only the one processor in the quorum and it is trying to acquire the lock
+ * then it should be given a WRITE access.
```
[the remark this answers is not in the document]
lakshmi-manasa-g: removed this api as per suggestion.and using cluster membership

samza-core/src/main/java/org/apache/samza/coordinator/ClusterMembership.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * @return number of processors in the list
+   */
+  int getNumberOfProcessors();
+
+  /**
+   * remove processor from the list
+   */
+  void unregisterProcessor();
```
mynameborat: Same here, should this take a `processorId` as an argument?

samza-core/src/main/java/org/apache/samza/coordinator/ClusterMembership.java
```diff
@@ -0,0 +1,40 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.coordinator;
+
+/**
+ * Coordination Primitive to maintain the list of processors in the quorum
```
mynameborat: Can you add more docs to this interface? I'd like to specifically see 
* Expectation around the register and unregister methods. 
* Read after write guarantee
* Since there isn't any listeners what is required from the implementors in case of disconnects or loss of membership.
* Thread safety and concurrent access

samza-core/src/main/java/org/apache/samza/coordinator/CoordinationConstants.java
```diff
@@ -0,0 +1,30 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.coordinator;
+
+
+public class CoordinationConstants {
```
mynameborat: nit: mark the class final

samza-core/src/main/java/org/apache/samza/coordinator/RunIdGenerator.java
```diff
@@ -0,0 +1,114 @@
...
+  public String getRunId() {
+    DistributedLock runIdLock;
+    String runId = null;
+
+
+    try {
+      runIdLock = coordinationUtils.getLock(CoordinationConstants.RUNID_LOCK_ID);
+      clusterMembership = coordinationUtils.getClusterMembership();
+      if (runIdLock == null || metadataStore == null || clusterMembership == null) {
+        throw new SamzaException("Failed to create utils for run id generation");
```
mynameborat: Is the idea here to return null as well or throw exception to the caller? It seems like this exception is going to be caught few lines later and we end up returning null anyways?

samza-core/src/main/java/org/apache/samza/coordinator/RunIdGenerator.java
```diff
@@ -0,0 +1,114 @@
...
+
+  public String getRunId() {
+    DistributedLock runIdLock;
+    String runId = null;
+
+
+    try {
+      runIdLock = coordinationUtils.getLock(CoordinationConstants.RUNID_LOCK_ID);
+      clusterMembership = coordinationUtils.getClusterMembership();
+      if (runIdLock == null || metadataStore == null || clusterMembership == null) {
```
mynameborat: We can remove the metadataStore and clusterMembership (once you address the other comment around initializing) checks since they are already done in the constructor.

samza-core/src/main/java/org/apache/samza/coordinator/RunIdGenerator.java
```diff
@@ -0,0 +1,114 @@
...
+              String.valueOf(System.currentTimeMillis()) + "-" + UUID.randomUUID().toString().substring(0, 8);
+          LOG.info("Writing the run id for this run as {}", runId);
+          metadataStore.put(CoordinationConstants.RUNID_STORE_KEY, runId.getBytes("UTF-8"));
+        } else {
+          runId = new String(metadataStore.get(CoordinationConstants.RUNID_STORE_KEY));
+          LOG.info("Read the run id for this run as {}", runId);
+        }
+      }
+    } catch (TimeoutException e) {
+      String msg = String.format("Processor timed out waiting to acquire lock for run.id generation");
```
mynameborat: Seems like a plain string. Why use `String.format`?
Can we eliminate the interim local variable and have it plugged in directly to the exception creation?

samza-core/src/main/java/org/apache/samza/coordinator/RunIdGenerator.java
```diff
@@ -0,0 +1,114 @@
...
+      throw new SamzaException(msg, e);
+    } catch (UnsupportedEncodingException e) {
+      String msg = String.format("Processor could not serialize/deserialize string for run.id generation");
```
mynameborat: same as above.

samza-core/src/main/java/org/apache/samza/coordinator/RunIdGenerator.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+    return runId;
+  }
+
+  /**
+   * might be called several times and hence should be idempotent
+   * that means the clusterMembership.unregisterProcessor should be idempotent
+   */
+  public void close() {
+    clusterMembership.unregisterProcessor();
```
mynameborat: can we instead track invocation and short circuit here?

samza-core/src/main/java/org/apache/samza/execution/LocalJobPlanner.java
```diff
@@ -43,19 +49,35 @@
  */
 public class LocalJobPlanner extends JobPlanner {
   private static final Logger LOG = LoggerFactory.getLogger(LocalJobPlanner.class);
-  private static final String APPLICATION_RUNNER_PATH_SUFFIX = "/ApplicationRunnerData";
+  private static final String METADATA_STORE = "StreamCreationCoordinationStore";
```
mynameborat: nit: rename it to explicitly say `STREAM_CREATION_METADATA_STORE`?

samza-core/src/main/java/org/apache/samza/execution/LocalJobPlanner.java
```diff
@@ -119,21 +137,79 @@ private void createStreams(String planId, List<StreamSpec> intStreams, StreamMan
...
+                "Processor {} did not obtain the lock for streams creation. They must've been created by another processor.",
+                uid);
+          }
+        } catch (TimeoutException e) {
+          LOG.warn("Processor {} failed to get the lock for stream initialization. Will try again until time out", uid);
+        } catch (UnsupportedEncodingException e) {
+          String msg = String.format("Processor {} failed to encode string for stream initialization", uid);
+          throw new SamzaException(msg, e);
+        } finally {
+          distributedLock.unlock();
```
mynameborat: It seems possible for the processor to not obtain `lock` but still call `unlock`. Can that potentially have any side effects?

samza-core/src/main/java/org/apache/samza/execution/LocalJobPlanner.java
```diff
@@ -119,21 +137,79 @@ private void createStreams(String planId, List<StreamSpec> intStreams, StreamMan
...
+      long startTime = System.currentTimeMillis();
+      long lockTimeout = TimeUnit.MILLISECONDS.convert(CoordinationConstants.LOCK_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+
+      // If "stream created state" exists in store then skip stream creation
+      // Else acquire lock, create streams, set state in store and unlock
+      // Checking for state before acquiring lock to prevent all processors from acquiring lock
+      // In a while loop so that if two processors check state simultaneously then
+      // to make sure the processor not acquiring the lock
+      // does not die of timeout exception and comes back and checks for state and proceeds
+      while ((System.currentTimeMillis() - startTime) < lockTimeout) {
```
mynameborat: Can we modularize the following code? It is hard to read through especially with try and catch and throw statements everywhere.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -59,13 +68,19 @@
 public class LocalApplicationRunner implements ApplicationRunner {
 
   private static final Logger LOG = LoggerFactory.getLogger(LocalApplicationRunner.class);
+  private static final String UID = UUID.randomUUID().toString();
```
mynameborat: nit: can we give it a meaningful name?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -59,13 +68,19 @@
...
+  private final  static String METADATA_STORE = "RunIdCoordinationStore";
```
mynameborat: nit: rename to explicitly refer to `RUN_ID_METADATA_STORE`?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -200,6 +283,21 @@ static String createProcessorId(ApplicationConfig appConfig) {
     }
   }
 
+  private void cleanup() {
+    if (runIdGenerator.isPresent()) {
+      runIdGenerator.get().close();
```
mynameborat: nit: can be simplified with `runIdGenerator::ifPresent(RunIdGenerator::close);`
same for `coordinationUtils`

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedLock.java
```diff
@@ -19,48 +19,47 @@
...
   public static final Logger LOG = LoggerFactory.getLogger(ZkDistributedLock.class);
   private static final String STATE_INITED = "sate_initialized";
   private final ZkUtils zkUtils;
   private final String lockPath;
   private final String participantId;
   private final ZkKeyBuilder keyBuilder;
-  private final Random random = new Random();
   private String nodePath = null;
-  private final String statePath;
+  private  Object mutex;
```
mynameborat: nit: extra space

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedLock.java
```diff
@@ -19,48 +19,47 @@
...
+    zkUtils.getZkClient().subscribeChildChanges(lockPath, new ParticipantChangeHandler(zkUtils));
   }
 
   /**
    * Tries to acquire a lock in order to create intermediate streams. On failure to acquire lock, it keeps trying until the lock times out.
    * Creates a sequential ephemeral node to acquire the lock. If the path of this node has the lowest sequence number, the processor has acquired the lock.
    * @param timeout Duration of lock acquiring timeout.
    * @param unit Unit of the timeout defined above.
-   * @return true if lock is acquired successfully, false if it times out.
+   * @return true if lock is acquired successfully or throw TimeOutException if could not acquire within timeout
```
mynameborat: nit: use `@throws` instead in the java docs

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedLock.java
```diff
@@ -114,4 +107,29 @@ public void unlockAndSet() {
...
+    public ParticipantChangeHandler(ZkUtils zkUtils) {
+      super(zkUtils, "ParticipantChangeHandler");
+    }
+
+    // Called when the children of the given path changed.
+    @Override
+    public void doHandleChildChange(String parentPath, List<String> currentChildren)
+        throws Exception {
+      synchronized (mutex) {
+        if (currentChildren == null) {
```
mynameborat: When does this happen?

lakshmi-manasa-g: When the Znode at parentPath has been deleted.
https://github.com/sgroschupf/zkclient/blob/03ccf12c70aca2f771bfcd94d44dc7c4d4a1495e/src/main/java/org/I0Itec/zkclient/ZkClient.java#L846

since Im logging the currentChildren and then later fetching the list of children in lock, thought it might be best to warn and avoid going to lock with null children.

samza-core/src/main/java/org/apache/samza/coordinator/ClusterMembership.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public interface ClusterMembership {
+  /**
+   * add processor to the list
+   */
+  void registerProcessor();
```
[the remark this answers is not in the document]
lakshmi-manasa-g: do we need a processor id? 
at this point, we are using just the size of the list (aka number of processors registered) and not using the id in any way.

samza-core/src/main/java/org/apache/samza/coordinator/RunIdGenerator.java
```diff
@@ -101,14 +103,19 @@ public String getRunId() {
     } finally {
       runIdLock.unlock();
     }
-    return runId;
+    if (runId != null) {
+      return Optional.of(runId);
+    }
```
mynameborat: Can be simplified to `Optional.ofNullable(runId);`

samza-core/src/main/java/org/apache/samza/coordinator/RunIdGenerator.java
lakshmi-manasa-g: moved to throw exceptions and have the caller deal with exceptions arising when RunIdGenerator is created and when getRunId() is called.

samza-core/src/main/java/org/apache/samza/coordinator/ClusterMembership.java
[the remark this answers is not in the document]
mynameborat: If the implementor is to generate the processor id then I assume we have 1-1 association between `clusterMembership` and `StreamProcessor`. I see this has certain downsides.
1.  It is an implicit requirement and not apparent from the CoordinationUtils interface standpoint which is responsible for creation of `ClusterMembership`. 
2. Having multiple instances of `ClusterMembership` in case of multiple `StreamProcessors` might not be ideal especially in terms of maintaining multiple connections to underlying system etc.

Would it extra effort or are there any downsides taking a processor id?

samza-core/src/main/java/org/apache/samza/coordinator/DistributedLock.java
```diff
@@ -19,24 +19,22 @@
...
+public interface DistributedLock {
 
   /**
-   * Try to acquire the lock, but first check if the state flag is set. If it is set, return false.
-   * If the flag is not set, and lock is acquired - return true.
+   * Try to acquire the lock
    * @param timeout Duration of lock acquiring timeout.
-   * @param unit Time Unit of the timeout defined above.
-   * @return true if lock is acquired successfully, false if state is already set.
+   * @return true if lock is acquired successfully
```
mynameborat: what scenario does `false` correspond to? Is there a need to differentiate between Timeout and returning `false`?

samza-core/src/main/java/org/apache/samza/coordinator/DistributedLock.java
```diff
@@ -19,24 +19,22 @@
...
    * @throws TimeoutException if could not acquire the lock.
```
mynameborat: nit: update comment to reflect this happens only when the wait time to acquire lock is > timeout. not, in the general case of not able to acquire the lock.

samza-core/src/main/java/org/apache/samza/coordinator/RunIdGenerator.java
```diff
@@ -0,0 +1,111 @@
...
+  private final MetadataStore metadataStore;
+  private final ClusterMembership clusterMembership;
+
+  public RunIdGenerator(CoordinationUtils coordinationUtils, MetadataStore metadataStore) {
+    Preconditions.checkNotNull(coordinationUtils, "CoordinationUtils cannot be null");
+    Preconditions.checkNotNull(metadataStore, "MetadataStore cannot be null");
+    this.coordinationUtils = coordinationUtils;
+    this.metadataStore = metadataStore;
+    this.clusterMembership = coordinationUtils.getClusterMembership();
+    if (clusterMembership == null) {
```
mynameborat: I'd prefer to have `Preconditions.checkNotNull(...)` so that we throw a same category of exception in the event of invalid arguments.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -77,20 +92,86 @@
...
+    boolean isAppModeBatch = new ApplicationConfig(appDesc.getConfig()).getAppMode() == ApplicationConfig.ApplicationMode.BATCH;
+    if (!isAppModeBatch) {
+      return Optional.empty();
+    }
+    JobCoordinatorConfig jcConfig = new JobCoordinatorConfig(config);
+    CoordinationUtils coordinationUtils = jcConfig.getCoordinationUtilsFactory().getCoordinationUtils(CoordinationConstants.APPLICATION_RUNNER_PATH_SUFFIX, PROCESSOR_ID, config);
+    if (coordinationUtils != null) {
+      return Optional.of(coordinationUtils);
+    }
+    return Optional.empty();
```
mynameborat: can be simplified to `Optional.oNullable(coordinationUtils)`

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -77,20 +92,86 @@
...
+  }
+
+  /**
+   *
```
mynameborat: nit: fill in a description or remove the extra line

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -77,20 +92,86 @@
...
+    if (this.runId.isPresent()) {
+      runId = this.runId.get();
+    }
+    return new LocalJobPlanner(appDesc, coordinationUtils, PROCESSOR_ID, runId);
+  }
+
+
+  private void initializeRunId() {
+    try {
+      boolean isAppModeBatch = new ApplicationConfig(appDesc.getConfig()).getAppMode() == ApplicationConfig.ApplicationMode.BATCH;
```
mynameborat: looks like we extract isAppModeBatch in quite a few places within this class. Can you extract it to a class variable and initialize it in constructor.

samza-core/src/main/java/org/apache/samza/coordinator/ClusterMembership.java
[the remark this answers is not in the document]
lakshmi-manasa-g: added processor id.

samza-azure/src/main/java/org/apache/samza/coordinator/AzureCoordinationUtils.java
```diff
@@ -45,11 +44,17 @@ public Latch getLatch(int size, String latchId) throws UnsupportedOperationExcep
     return null;
   }
 
+  // To support DistributedLock in Azure, even MetadataStore needs to be implemented
```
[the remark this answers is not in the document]
lakshmi-manasa-g: added bit more context to the comment.

samza-core/src/main/java/org/apache/samza/coordinator/DistributedLock.java
lakshmi-manasa-g: removed Timeout Exception and returning `false` instead

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -77,20 +93,81 @@
...
+    try {
+      MetadataStore metadataStore = getMetadataStore();
+      if (coordinationUtils.isPresent() && metadataStore != null) {
+        runIdGenerator = Optional.of(new RunIdGenerator(coordinationUtils.get(), metadataStore));
+      }
+      if (!coordinationUtils.isPresent() || !isAppModeBatch || !runIdGenerator.isPresent()) {
+        LOG.warn("coordination utils or run id generator could not be created successfully!");
+        return;
+      }
+      runId = runIdGenerator.get().getRunId();
```
mynameborat: runIdGenerator check can be removed and you can simplify this to `runId = runIdGenerator.ifPresent(RunIdGenerator::getRunId);`

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -77,20 +93,81 @@
...
+  public String getRunid() {
+    if (runId.isPresent()) {
+      return runId.get();
+    }
+    return null;
   }
 
   @Override
   public void run(ExternalContext externalContext) {
+    if (isAppModeBatch) {
```
mynameborat: can we remove this check since we already do this check inside the `initializeRunId()`

samza-core/src/main/java/org/apache/samza/zk/ZkCoordinationUtils.java
```diff
@@ -52,11 +53,12 @@ public Latch getLatch(int size, String latchId) {
   }
 
   @Override
-  public DistributedLockWithState getLockWithState(String lockId) {
+  public DistributedLock getLock(String lockId) {
     return new ZkDistributedLock(processorIdStr, zkUtils, lockId);
   }
 
   public void close() {
+    System.out.println("Manasa: closing coordination utils");
```
mynameborat: nit: remove

samza-test/src/test/java/org/apache/samza/test/processor/TestZkLocalApplicationRunner.java
```diff
@@ -1011,7 +1011,7 @@ public void testAgreeingOnSameRunIdForBatch() throws InterruptedException {
     LocalApplicationRunner localApplicationRunner1 = (LocalApplicationRunner) appRunner1;
     LocalApplicationRunner localApplicationRunner2 = (LocalApplicationRunner) appRunner2;
 
-    assertEquals("RunId of the two processors does not match", localApplicationRunner2.getRunid(), localApplicationRunner1.getRunid());
+    assertEquals("RunId of the two processors does not match", localApplicationRunner2.getRunId().orElse(null), localApplicationRunner1.getRunId().orElse(null));
```
mynameborat: You don't have to unwrap the optional for assertEquals. 
I think it should work out of the box with object comparison.

samza-core/src/main/java/org/apache/samza/execution/LocalJobPlanner.java
```diff
@@ -104,36 +127,89 @@ private void createStreams(String planId, List<StreamSpec> intStreams, StreamMan
...
+    long lockTimeout = TimeUnit.MILLISECONDS.convert(CoordinationConstants.LOCK_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+
+    // If "stream created state" exists in store then skip stream creation
+    // Else acquire lock, create streams, set state in store and unlock
+    // Checking for state before acquiring lock to prevent all processors from acquiring lock
+    // In a while loop so that if two processors check state simultaneously then
+    // to make sure the processor not acquiring the lock
+    // does not die of timeout exception and comes back and checks for state and proceeds
+    while ((System.currentTimeMillis() - startTime) < lockTimeout) {
+      if (metadataStore.get(String.format(STREAM_CREATED_STATE_KEY, lockId)) != null) {
```
vjagadish1989: too many if-else's

I would re-think the control-flow : get-rid off breaks, group similar conditions together

Here's a better organization, imho:

```
if (metadataStore.get() != null) {
  log "streams already created"
} else {
  if (lock.lock()) {
      create, update, unlock
   } else {
      log "unable to acquire lock"
   }
}
```

samza-core/src/main/java/org/apache/samza/execution/LocalJobPlanner.java
```diff
@@ -104,36 +127,89 @@ private void createStreams(String planId, List<StreamSpec> intStreams, StreamMan
...
+        } else {
+          LOG.info("Processor {} failed to get the lock for stream initialization. Will try again until time out", processorId);
+        }
+      } catch (UnsupportedEncodingException e) {
+        String msg = String.format("Processor {} failed to encode string for stream initialization", processorId);
+        throw new SamzaException(msg, e);
+      }
+    }
+    if ((System.currentTimeMillis() - startTime) >= lockTimeout) {
+      throw new TimeoutException(String.format("Processor {} failed to get the lock for stream initialization within {} milliseconds.", processorId, CoordinationConstants.LOCK_TIMEOUT_MS));
```
vjagadish1989: throw a SamzaException directly with the right message - instead of catching/wrapping/re-throwing higher up the call stack

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -77,20 +93,74 @@
...
+    if (!isAppModeBatch) {
+      return new LocalJobPlanner(appDesc, PROCESSOR_ID);
+    }
+    CoordinationUtils coordinationUtils = this.coordinationUtils.orElse(null);
+    String runId = this.runId.orElse(null);
+    return new LocalJobPlanner(appDesc, coordinationUtils, PROCESSOR_ID, runId);
+  }
+
+
+  private void initializeRunId() {
```
vjagadish1989: make the function stateless by return "getRunId()" instead of mutating class-state

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedLock.java
```diff
@@ -18,93 +18,84 @@
...
   public static final Logger LOG = LoggerFactory.getLogger(ZkDistributedLock.class);
   private static final String STATE_INITED = "sate_initialized";
   private final ZkUtils zkUtils;
   private final String lockPath;
   private final String participantId;
   private final ZkKeyBuilder keyBuilder;
-  private final Random random = new Random();
   private String nodePath = null;
-  private final String statePath;
+  private Object mutex;
```
vjagadish1989: final?

samza-core/src/main/java/org/apache/samza/zk/ZkDistributedLock.java
```diff
@@ -18,93 +18,84 @@
...
 
     nodePath = zkUtils.getZkClient().createEphemeralSequential(lockPath + "/", participantId);
 
     //Start timer for timeout
     long startTime = System.currentTimeMillis();
-    long lockTimeout = TimeUnit.MILLISECONDS.convert(timeout, unit);
+    long lockTimeout = timeout.toMillis();
 
     while ((System.currentTimeMillis() - startTime) < lockTimeout) {
+      synchronized (mutex) {
```
vjagadish1989: Why change the existing implementation that relies on thread.sleep? - If it's a simple enough operation, sleep + re-poll may be more understandable than a mutex + callback combo

PR #939, opened 2019-03-06
SAMZA-2122: Fix the task caught-up logic which doesn't handle no incoming messages

xinyuiscool:
Currently the TaskInstance.checkCaughtUp() logic has a bug that if there is no incoming messages for a partition, it will not mark the ssp to be caught up. Instead, it should mark ssp to be caught up if the starting offset is already the upcoming offset for a ssp, meaning it has consumed all the messages in this partition and waiting for the future incoming messages. This indicates the ssp is fully caught up.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
+++ b/samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
@@ -323,6 +323,9 @@ class TaskInstance(
     }
   }
 
+  /**
+    * Check each partition assigned to the task is caught to the last offset
+    */
   def initCaughtUpMapping() {
     if (taskContext.getStreamMetadataCache != null) {
       systemStreamPartitions.foreach(ssp => {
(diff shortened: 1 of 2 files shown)


samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -321,6 +323,25 @@ class TaskInstance(
...
+          .getStreamMetadataCache
+          .getSystemStreamMetadata(ssp.getSystemStream, false)
+          .getSystemStreamPartitionMetadata.get(ssp.getPartition)
+
+        val upcomingOffset = partitionMetadata.getUpcomingOffset
+        val startingOffset = offsetManager.getStartingOffset(taskName, ssp)
+          .getOrElse(throw new SamzaException("No offset defined for SystemStreamPartition: %s" format ssp))
+
+        if(Objects.equals(upcomingOffset, startingOffset)) {
+          ssp2CaughtupMapping(ssp) = true
```
sborya: comment here.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -321,6 +323,25 @@ class TaskInstance(
     }
   }
 
+  def initCaughtUpMapping() {
```
sborya: Comment for the method.

xinyuiscool: Added the comments.

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
xinyuiscool: Added.

PR #940, opened 2019-03-06
SAMZA-2121: Add checkpoint offset field to IncomingMessageEnvelope

cameronlee314:
This is the first step for implementing https://issues.apache.org/jira/browse/SAMZA-2120.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
+++ b/samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
@@ -33,17 +33,35 @@ public class IncomingMessageEnvelope {
   public static final String END_OF_STREAM_OFFSET = new String(END_OF_STREAM_BYTES, Charset.defaultCharset());
 
   private final SystemStreamPartition systemStreamPartition;
+  /**
+   * The offset in the partition that the message was received from
+   */
   private final String offset;
+  /**
+   * Offset that can be checkpointed when this {@link IncomingMessageEnvelope} is processed.
+   * This might be different from {@link #offset} in a case where the direct offset is not a valid offset from which to
+   * resume consumption.
+   */
+  private final String checkpointOffset;
   private final Object key;
   private final Object message;
+  /**
+   * Size of the message and key in bytes. This may be set to 0 if the event source did not provide a value for it.
+   */
   private final int size;
-  // the timestamp when this event occured, should be set by the event source, 0 means unassigned
-  private long eventTime = 0L;
-  // the timestamp when this event is pickedup by samza, 0 means unassgined
-  private long arrivalTime = 0L;
+  /**
+   * The timestamp when this event occurred, should be set by the event source. This may be set to 0 if the event source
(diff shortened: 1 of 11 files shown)


cameronlee314:
@rmatharu @mynameborat could you please take a look?

samza-core/src/test/java/org/apache/samza/task/TestAsyncRunLoop.java
```diff
@@ -51,7 +51,7 @@
 import scala.Option;
 import scala.collection.JavaConverters;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
```
mynameborat: nit: use explicit imports instead of `*`

cameronlee314:
> Looks good.
> General question around checkpoint offset vs offset:
> Are there other systems that will benefit with this separation? To me it seems this separation arises only in the case of Kafka specifically large message scenario.

We don't currently have any other concrete use cases for this, but I was thinking that this was still a general enough abstraction that it could apply to other systems if they needed to modify offsets.

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -86,7 +101,29 @@ public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, Stri
...
+   * @param message A deserialized message received from the partition offset.
+   * @param size size of the message and key in bytes.
+   * @param eventTime the timestamp (in epochMillis) of when this event happened
+   * @param arrivalTime the timestamp (in epochMillis) of when this event arrived to (i.e., was picked-up by) Samza
+   */
+  public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, String offset, String checkpointOffset,
+      Object key, Object message, int size, long eventTime, long arrivalTime) {
+    this.systemStreamPartition = systemStreamPartition;
+    this.offset = offset;
+    this.checkpointOffset = checkpointOffset;
```
shanthoosh: 1. Can you please share the rationale behind adding checkpoint offset to `IncomingMessageEnvelope` public API. 
2. Can you please throw some light on the other options considered here? Is it not plausible to do this making this change in public API. 

IMHO, it'll be better not to proliferate the public API with these internal details which will be hard to remove later on. Anyone who reads this in the future, would wonder why there are two different offsets in `IncomingMessageEnvelope` contract. Is it possible to maintain the mapping(association we want for safe-checkpoint) internally within the samza framework side and expose it to users.

cameronlee314: 1. Some more details and an example are described at https://issues.apache.org/jira/browse/SAMZA-2120.
2. We could make this part of the internal API by creating an `InternalIncomingMessageEnvelope` which is passed around internally and extends `IncomingMessageEnvelope`. Would that help avoid complicating the public API?

cameronlee314:
This is no longer necessary. See SAMZA-2121 for more details.

PR #941, opened 2019-03-07
SAMZA-2120: Enable custom handling of ConsumerRecords consumed by Kafka 

cameronlee314:
This includes the code from https://github.com/apache/samza/pull/940.
For this PR, it would be practical to just look at commit https://github.com/apache/samza/commit/e09f03022802b0757146915647a20a56478a7e4e.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
+++ b/samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
@@ -74,7 +74,7 @@ public class TaskSideInputStorageManager {
   private final SystemAdmins systemAdmins;
   private final TaskName taskName;
   private final TaskMode taskMode;
-  private final Map<SystemStreamPartition, String> lastProcessedOffsets = new ConcurrentHashMap<>();
+  private final Map<SystemStreamPartition, String> checkpointOffsets = new ConcurrentHashMap<>();
 
   private Map<SystemStreamPartition, String> startingOffsets;
 
@@ -128,8 +128,8 @@ public class TaskSideInputStorageManager {
     startingOffsets = getStartingOffsets(fileOffsets, oldestOffsets);
     LOG.info("Starting offsets for the task {}: {}", taskName, startingOffsets);
 
-    lastProcessedOffsets.putAll(fileOffsets);
-    LOG.info("Last processed offsets for the task {}: {}", taskName, lastProcessedOffsets);
+    checkpointOffsets.putAll(fileOffsets);
+    LOG.info("Last processed offsets for the task {}: {}", taskName, checkpointOffsets);
 
     initializeStoreDirectories();
   }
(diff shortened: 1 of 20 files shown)


cameronlee314:
@rmatharu @mynameborat could you please take a look?

samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
```diff
@@ -74,7 +74,7 @@
   private final SystemAdmins systemAdmins;
   private final TaskName taskName;
   private final TaskMode taskMode;
-  private final Map<SystemStreamPartition, String> lastProcessedOffsets = new ConcurrentHashMap<>();
```
rmatharu: Perhaps undo this change. 
Since this map actually stores the "last processed offset for the given side input" (which is used to emit a metric). 
SideInputs are only checkpointed locally (see writeOffsetFiles).

samza-core/src/main/java/org/apache/samza/task/AsyncRunLoop.java
```diff
@@ -611,10 +611,12 @@ public void run() {
             List<TaskCallbackImpl> callbacksToUpdate = callbackManager.updateCallback(callbackImpl);
             for (TaskCallbackImpl callbackToUpdate : callbacksToUpdate) {
               IncomingMessageEnvelope envelope = callbackToUpdate.envelope;
-              log.trace("Update offset for ssp {}, offset {}", envelope.getSystemStreamPartition(), envelope.getOffset());
+              log.trace("Update offset for ssp {}, checkpoint offset {}", envelope.getSystemStreamPartition(),
```
rmatharu: log.trace("Update offset for ssp {}, offset {}, checkpoint offset {}", envelope.getSystemStreamPartition(), envelope.getOffset(), envelope.getCheckpointOffset());

samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
cameronlee314: I was trying to clarify that this field no longer stored the actual "last processed offsets", since we now need to store the "checkpoint offset" for writing to the offset files.
Do you think we should also track the actual "last processed offsets" explicitly for the metric? Or would the "checkpoint offset" be sufficient for the metric?

prateekm:
@cameronlee314 I'd prefer to not add new fields/semantics to IME if possible. I'm assuming you want to return the safe offset from LiKafka as the checkpoint offset? Any other alternatives to handle it?

cameronlee314:
> @cameronlee314 I'd prefer to not add new fields/semantics to IME if possible. I'm assuming you want to return the safe offset from LiKafka as the checkpoint offset? Any other alternatives to handle it?

That's correct.
I have sent over some alternative solutions separately.

cameronlee314:
Replaced by https://github.com/apache/samza/pull/947.

PR #942, opened 2019-03-07
Bugfix: Recent CSM refactor was causing some metrics to not be emitted. Fixed <task>-restore-time metric.

rmatharu:


samza-core/src/main/scala/org/apache/samza/system/SystemConsumersMetrics.scala
```diff
@@ -19,12 +19,10 @@
 
 package org.apache.samza.system
 
-import org.apache.samza.metrics.MetricsRegistry
-import org.apache.samza.metrics.MetricsRegistryMap
-import org.apache.samza.metrics.Counter
-import org.apache.samza.metrics.MetricsHelper
+import org.apache.samza.metrics._
```
prateekm: Minor: Prefer explicit imports.

samza-core/src/main/scala/org/apache/samza/storage/ContainerStorageManager.java
```diff
@@ -573,6 +576,25 @@ private StorageEngine createStore(String storeName, TaskName taskName, TaskModel
     return this.sideInputStorageManagers.values().stream().collect(Collectors.toSet());
   }
 
+  /**
+   * Registers any CSM created metrics such as side-inputs related metrics, and standby-task related metrics.
+   * @param metricsReporters metrics reporters to use
+   */
+  public void registerMetrics(Map<String, MetricsReporter> metricsReporters) {
```
prateekm: I don't think this is the standard pattern. The right way would be to pass in the shared MetricsRegistry (created in SamzaContainer) to the CSM, and register metrics on that in CSM#register(). That MetricsRegistry is already registered with the MetricsReporters in SamzaContainer. Does this make sense?

rmatharu: That's true, but that pattern doesn't work in this case because both the CSM and SamzaContainer need to have a different registry because they both create SystemConsumersMetrics. Using the same registry causes a conflict in the metrics' name. 

We do use this pattern in TaskInstance, where each TaskInstance is given a set of reporters and it registers its metrics there.

mynameborat: +1, we would need to change the `SystemConsumerMetrics` to take a configurable group name instead of using the default class name as the group id. I feel that change is much simpler compared to copying over the boiler plate code for metrics registration to CSM.

shanthoosh: > SystemConsumerMetrics to take a configurable group name

Alerting in ingraphs and auto-scaling rules at LinkedIn are defined based upon the samza generated metric names. Group name is prefix and initial part of the generated metric name. It will be better to factor these use-cases into account and maintain backwards compatibility when changing the metric names.

rmatharu: Done. 
It did not require a group-name change, so all existing metrics remain as is.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -879,9 +880,13 @@ class SamzaContainer(
...
   def startMetrics {
-    info("Registering task instances with metrics.")
-
-    taskInstances.values.foreach(_.registerMetrics)
+    info("Registering taskInstanceMetrics.")
+    reporters.values.foreach(reporter => {
+      taskInstanceMetrics.values.foreach(taskMetrics => {
+        reporter.register(taskMetrics.source, taskMetrics.registry)
+      }
+      )
```
prateekm: Nitpick: Move to previous line.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -879,9 +880,13 @@ class SamzaContainer(
   }
 
   def startMetrics {
-    info("Registering task instances with metrics.")
-
-    taskInstances.values.foreach(_.registerMetrics)
+    info("Registering taskInstanceMetrics.")
```
prateekm: Nitpick: "task instance metrics"

samza-core/src/main/scala/org/apache/samza/system/SystemConsumersMetrics.scala
```diff
@@ -19,12 +19,13 @@
...
 
-import org.apache.samza.metrics.MetricsRegistry
 import org.apache.samza.metrics.MetricsRegistryMap
 import org.apache.samza.metrics.Counter
 import org.apache.samza.metrics.MetricsHelper
+import org.apache.samza.metrics.ReadableMetricsRegistry
 
-class SystemConsumersMetrics(val registry: MetricsRegistry = new MetricsRegistryMap) extends MetricsHelper {
+class SystemConsumersMetrics(val registry: ReadableMetricsRegistry = new MetricsRegistryMap,
+  val prefix: String = "") extends MetricsHelper {
```
prateekm: Can we call it 'source' instead? Or are the semantics very different?

samza-core/src/main/scala/org/apache/samza/storage/ContainerStorageManager.java
```diff
@@ -229,15 +230,15 @@ public ContainerStorageManager(ContainerModel containerModel, StreamMetadataCach
...
           this.sideInputSystemStreams.values().stream().flatMap(Set::stream).collect(Collectors.toSet())).toSet(), false);
 
-      SystemConsumersMetrics systemConsumersMetrics = new SystemConsumersMetrics(
-          new MetricsRegistryMap(String.format(SIDEINPUTS_METRICS_NAME, containerModel.getId())));
+      SystemConsumersMetrics sideInputSystemConsumersMetrics = new SystemConsumersMetrics(samzaContainerMetrics.registry(), SIDEINPUTS_METRICS_PREFIX);
+      // we use the same registry as samza-container-metrics
 
       MessageChooser chooser = DefaultChooser.apply(inputStreamMetadata, new RoundRobinChooserFactory(), config,
-          systemConsumersMetrics.registry(), systemAdmins);
+          sideInputSystemConsumersMetrics.registry(), systemAdmins);
```
prateekm: Minor: Consider passing in the registry explicitly in the constructor instead of getting it from the container metrics.

samza-core/src/main/scala/org/apache/samza/system/SystemConsumersMetrics.scala
prateekm: We discussed offline, source is a TaskInstance metrics specific thing and should be removed. 'prefix' has similar semantics and is the better way to do this. @rmatharu please add the ticket here with the discussion summary.

samza-core/src/main/scala/org/apache/samza/storage/ContainerStorageManager.java
rmatharu: Will do in fix for https://issues.apache.org/jira/browse/SAMZA-2125

samza-core/src/main/scala/org/apache/samza/system/SystemConsumersMetrics.scala
rmatharu: https://issues.apache.org/jira/browse/SAMZA-2125

rmatharu:
Addressed all comments.

PR #944, opened 2019-03-08
Release version updates

dxichen:
Updated hard coded version for the upcoming release
As per https://github.com/apache/samza/blob/master/docs/README.md

### DIFF
--- a/docs/learn/tutorials/versioned/samza-rest-getting-started.md
+++ b/docs/learn/tutorials/versioned/samza-rest-getting-started.md
@@ -48,7 +48,7 @@ Run the following commands:
 {% highlight bash %}
 cd samza-rest/build/distributions/
 mkdir -p deploy/samza-rest
-tar -xvf ./samza-rest-1.0.1-SNAPSHOT.tgz -C deploy/samza-rest
+tar -xvf ./samza-rest-1.1.1-SNAPSHOT.tgz -C deploy/samza-rest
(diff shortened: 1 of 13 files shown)


docs/learn/tutorials/versioned/samza-rest-getting-started.md
```diff
@@ -48,7 +48,7 @@ Run the following commands:
 {% highlight bash %}
 cd samza-rest/build/distributions/
 mkdir -p deploy/samza-rest
-tar -xvf ./samza-rest-1.0.1-SNAPSHOT.tgz -C deploy/samza-rest
+tar -xvf ./samza-rest-1.1.1-SNAPSHOT.tgz -C deploy/samza-rest
```
shanthoosh: 1. Shouldn't this be `1.1.0` instead of `1.1.1` 
2. Why are we using the snapshotted version for `samza-rest`?

dxichen: 1. I believe since this is on master it should be the SNAPSHOTTED version of the next release (ie 1.1.1-SNAPSHOT) instead of the latest released version.
2. Following convention of the previous releases, master should be named as a SNAPSHOT to be picked up in maven local

dxichen:
@shanthoosh @jagadish-v0 @prateekm Please take a look

PR #947, opened 2019-03-09
SAMZA-2120: Enable custom handling of ConsumerRecords consumed by Kafka

cameronlee314:
This replaces https://github.com/apache/samza/pull/941. Instead of adding an object hook into KafkaConsumerProxy, this allows KafkaConsumerProxy to be directly extended for record processing.


samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -51,7 +53,7 @@
  * This class is not thread safe. There will be only one instance of this class per KafkaSystemConsumer object.
  * We still need some synchronization around kafkaConsumer. See pollConsumer() method for details.
  */
-class KafkaConsumerProxy<K, V> {
+public class KafkaConsumerProxy<K, V> {
```
prateekm: Do the K, V type parameters matter for this class (and the underlying KafkaConsumer)?. If not, maybe remove for simplicity.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -310,11 +312,7 @@ private void fetchMessages() {
         results.put(ssp, messages);
       }
 
-      K key = record.key();
-      Object value = record.value();
-      IncomingMessageEnvelope imEnvelope =
-          new IncomingMessageEnvelope(ssp, String.valueOf(record.offset()), key, value, getRecordSize(record),
-              record.timestamp(), Instant.now().toEpochMilli());
+      IncomingMessageEnvelope imEnvelope = handleNewRecord(record, ssp);
```
prateekm: Nipick: ime

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -328,9 +326,21 @@ private void fetchMessages() {
...
+   * Convert a {@link ConsumerRecord} to an {@link IncomingMessageEnvelope}.
+   *
+   * This has a protected visibility so that {@link KafkaConsumerProxy} can be extended to add special handling logic
+   * for custom Kafka systems.
+   *
+   * @param consumerRecord {@link ConsumerRecord} from Kafka that was consumed
+   * @param systemStreamPartition {@link SystemStreamPartition} corresponding to the record
+   * @return {@link IncomingMessageEnvelope} corresponding to the {@code consumerRecord}
+   */
+  protected IncomingMessageEnvelope handleNewRecord(ConsumerRecord<K, V> consumerRecord,
```
prateekm: Maybe: 'toIME'?

samza-kafka/src/main/scala/org/apache/samza/util/KafkaUtilJava.java
```diff
@@ -0,0 +1,12 @@
+package org.apache.samza.util;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+
+
+public class KafkaUtilJava {
```
prateekm: Prefer inlining, or just calling this KafkaUtil, or if one already exists, moving method there (even if it's Scala).

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
cameronlee314: The <K, V> does get used for the kafka `Consumer` which is wired in. Ultimately, it doesn't get used since `IncomingMessageEnvelope` doesn't have type params, but it looks like it helps to ensure the wiring matches up for the proxy and Samza/Kafka consumer objects.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
cameronlee314: Renamed to `incomingMessageEnvelope`

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
cameronlee314: I originally considered a method name like that, but I felt that this hook could include some other custom logic as well, such as managing managing offsets for committing them back to the kafka broker.
We could use `toIncomingMessageEnvelope` for now and then change it later, but I figured `handleNewRecord` leaves it a bit more open.

samza-kafka/src/main/scala/org/apache/samza/util/KafkaUtilJava.java
cameronlee314: I figured that we need to be strict in avoiding new Scala code (or try to convert Scala code to Java as much as possible). It doesn't matter for this case though, since now this can just be kept within `KafkaConsumerProxy`. I'll move it back.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
prateekm: I'd prefer to remove the types from the Consumer as well, since they're mostly meaningless. But we can do it in a separate PR.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -66,7 +66,7 @@
 
   // This proxy contains a separate thread, which reads kafka messages (with consumer.poll()) and populates
   // BlockingEnvelopMap's buffers.
-  final private KafkaConsumerProxy proxy;
+  final private KafkaConsumerProxy<K, V> proxy;
```
shanthoosh: nit: s/final private/private final.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -328,7 +325,28 @@ private void fetchMessages() {
...
+      SystemStreamPartition systemStreamPartition) {
+    return new IncomingMessageEnvelope(systemStreamPartition, String.valueOf(consumerRecord.offset()),
```
shanthoosh: Are we planning to override this API internally at LI to inject the safe-offset into `IncomingMessageEnvelope`?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
cameronlee314: Oops, I should have taken the opportunity to do that since I was changing this line anyways, but the PR already got merged. It was already like that anyways, and it can be done in some other PR.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
cameronlee314: Yes

PR #950, opened 2019-03-12
SAMZA-2126: Bug fixes for batch-mode generated stream specs

xinyuiscool:
- Kafka properties doesn't support replication.factor when creating topic, so change the way to override intermediate stream properties. 
- Due to planner created twice, we need to check the physical stream name has been created for intermediate streams to avoid the run.id got appended twice.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
+++ b/samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
@@ -75,9 +75,12 @@ public class StreamEdge {
     StreamSpec spec = (partitions == PARTITIONS_UNKNOWN) ?
         streamSpec : streamSpec.copyWithPartitionCount(partitions);
 
-    // Append unique id to the batch intermediate streams
-    // Check the physical stream name is already generated first
+    // For intermediate stream that physical name is the same as id,
+    // meaning the physical name is auto-generated, and not overrided
+    // by user or batch processing.
     if (isIntermediate && spec.getId().equals(spec.getPhysicalName())) {
+      // Append unique id to the batch intermediate streams
+      // Note this will only happen for batch processing
       String physicalName = StreamManager.createUniqueNameForBatch(spec.getPhysicalName(), config);
       if (!physicalName.equals(spec.getPhysicalName())) {
         spec = spec.copyWithPhysicalName(physicalName);
--- a/samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
+++ b/samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
@@ -75,7 +75,9 @@ public class StreamEdge {
     StreamSpec spec = (partitions == PARTITIONS_UNKNOWN) ?
         streamSpec : streamSpec.copyWithPartitionCount(partitions);
 
-    if (isIntermediate) {
+    // Append unique id to the batch intermediate streams
+    // Check the physical stream name is already generated first
+    if (isIntermediate && spec.getId().equals(spec.getPhysicalName())) {
       String physicalName = StreamManager.createUniqueNameForBatch(spec.getPhysicalName(), config);
       if (!physicalName.equals(spec.getPhysicalName())) {
         spec = spec.copyWithPhysicalName(physicalName);
--- a/samza-core/src/test/java/org/apache/samza/execution/TestStreamEdge.java
+++ b/samza-core/src/test/java/org/apache/samza/execution/TestStreamEdge.java
@@ -50,9 +50,15 @@ public class TestStreamEdge {
     Map<String, String> config = new HashMap<>();
     config.put(ApplicationConfig.APP_MODE, ApplicationConfig.ApplicationMode.BATCH.name());
     config.put(ApplicationConfig.APP_RUN_ID, "123");
+
+    // For batch, if the physical name hasn't been generated, it should append the unique id
     StreamSpec batchSpec = new StreamSpec("stream-1", "stream-1", "system-1");
     StreamEdge edge = new StreamEdge(batchSpec, true, false, new MapConfig(config));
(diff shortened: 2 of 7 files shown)


xinyuiscool:
@lhaiesp and @dxichen : please take a look!

samza-core/src/test/java/org/apache/samza/execution/TestStreamEdge.java
```diff
@@ -50,8 +50,9 @@ public void testGetStreamSpec_Batch() {
     Map<String, String> config = new HashMap<>();
     config.put(ApplicationConfig.APP_MODE, ApplicationConfig.ApplicationMode.BATCH.name());
     config.put(ApplicationConfig.APP_RUN_ID, "123");
-    StreamEdge edge = new StreamEdge(spec, true, false, new MapConfig(config));
-    assertEquals(edge.getStreamSpec().getPhysicalName(), spec.getPhysicalName() + "-123");
+    StreamSpec batchSpec = new StreamSpec("stream-1", "stream-1", "system-1");
+    StreamEdge edge = new StreamEdge(batchSpec, true, false, new MapConfig(config));
+    assertEquals(edge.getStreamSpec().getPhysicalName(), batchSpec.getPhysicalName() + "-123");
```
lhaiesp: shouldn't you have another test case where the run.id is already applied and you verify it's not re-applied?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
```diff
@@ -106,8 +106,8 @@ class KafkaSystemFactory extends SystemFactory with Logging {
     if (appConfig.getAppMode == ApplicationMode.BATCH) {
       val streamConfig = new StreamConfig(config)
       streamConfig.getStreamIds().filter(streamConfig.getIsIntermediateStream(_)).map(streamId => {
+        // only the override here
         val properties = new Properties()
-        properties.putAll(streamConfig.getStreamProperties(streamId))
```
lhaiesp: It's not obvious to me how this is different from the kafkaSpec.getProperties(). Will follow up with you

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
```diff
@@ -75,7 +75,9 @@ StreamSpec getStreamSpec() {
     StreamSpec spec = (partitions == PARTITIONS_UNKNOWN) ?
         streamSpec : streamSpec.copyWithPartitionCount(partitions);
 
-    if (isIntermediate) {
+    // Append unique id to the batch intermediate streams
+    // Check the physical stream name is already generated first
+    if (isIntermediate && spec.getId().equals(spec.getPhysicalName())) {
```
dxichen: Could there be a case where the Id and the physical name are not set to the same value for batch? In either case, we should add a comment about how we can guarantee this block is only for batch.

samza-core/src/test/java/org/apache/samza/execution/TestStreamEdge.java
dxichen: +1, could there also add a test such that this does not trigger for intermediate streams for nonbatch?

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
dxichen: I believe `kafkaSpec.getProperties()` strictly returns kafka properties. Did we want to avoid a subset of the stream properties?

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemFactoryJava.java
```diff
@@ -45,15 +46,13 @@ public void testGetIntermediateStreamProperties() {
     // no properties for stream
     config.put("streams.test.samza.intermediate", "true");
     config.put("streams.test.compression.type", "lz4"); //some random config
-    properties = JavaConversions.mapAsJavaMap(
-        factory.getIntermediateStreamProperties(new MapConfig(config)));
-    assertTrue(properties.isEmpty());
```
dxichen: I think these lines should be able stay here, since they should be filtered out as well but kafkaStreamSpec?

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemFactoryJava.java
```diff
@@ -22,6 +22,7 @@
 import org.apache.samza.config.ApplicationConfig;
```
dxichen: I don't see any test that satisfy `kafka properties doesn't support replication.factor when creating topic` as per Jira description.

shanthoosh:
> Kafka properties doesn't support replication.factor when creating topic, so change the way to override intermediate stream properties.

Just curious. Please correct me if I'm wrong.

1. Even kafka-topic creation shell-command supports a configuration like `replication.factor` when creating the topic(( kafka-streams supports configuration like replication.factor)). Is the change entails in the way how  kafka-client accepts this config(replication.factor) when creating the streams?
2. If kafka-client API does not support `replication.factor`, wouldn't it apply to other internal metadata-topics managed by samza viz `coordinator-stream`, `checkpoint` etc?

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemFactoryJava.java
xinyuiscool: Good catch! I added them back.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaSystemFactory.scala
lhaiesp: ok. make sense now

PR #951, opened 2019-03-14
SAMZA-2127: Upgrade to Kafka 2.0

mynameborat:
Initial PR for Kafka 2.0 upgrade


samza-test/src/test/scala/org/apache/samza/test/integration/StreamTaskTestUtil.scala
```diff
@@ -257,23 +250,27 @@ class StreamTaskTestUtil {
...
-      if (message.message == null) {
-        messages += null
+      if (stream.hasNext) {
+        message = stream.next
+        if (message.value() == null) {
+          messages += null
+        } else {
+          messages += new String(message.value, "UTF-8")
+        }
+        System.err.println("StreamTaskTestUtil.readAll(): offset=%s, message=%s" format (message.offset, messages.last))
```
cameronlee314: I know this was already here, but it doesn't seem like an error. Could it just be removed, or at least changed to System.out?

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemAdmin.scala
```diff
@@ -257,17 +211,22 @@ class TestKafkaSystemAdmin {
...
-    var stream = connector.createMessageStreams(Map(TOPIC -> 1))(TOPIC).head.iterator
-    var message = stream.next
-    var text = new String(message.message, "UTF-8")
-    connector.shutdown
+    val consumer = getKafkaConsumer
+    consumer.subscribe(JavaConverters.setAsJavaSetConverter(Set(TOPIC)).asJava)
+    val records = consumer.poll(Duration.ofMillis(1000))
+    consumer.close()
+
+    assertTrue(records.count() > 2)
```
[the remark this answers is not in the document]
mynameborat: Forgot to push the local change.
Fixed it.

samza-kafka/src/test/scala/org/apache/samza/checkpoint/kafka/TestKafkaCheckpointManager.scala
```diff
@@ -110,15 +110,12 @@ class TestKafkaCheckpointManager extends KafkaServerTestHarness {
     kcm1.stop
 
     // check that start actually creates the topic with log compaction enabled
-    val zkClient = ZkUtils(zkConnect, 6000, 6000, JaasUtils.isZkSecurityEnabled())
-    val topicConfig = AdminUtils.fetchEntityConfig(zkClient, ConfigType.Topic, checkpointTopic)
+    val topicConfig = adminZkClient.getAllTopicConfigs().getOrElse(checkpointTopic, new Properties())
```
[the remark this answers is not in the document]
mynameborat: `adminZkClient` is a very light weight utility wrapper around `zkClient` and doesn't have any lifecycle methods. However, `ZookeeperTestHarness` parent of `KafkaServerTestHarness` takes care of managing the lifecycle of zkClient.

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemAdmin.scala
```diff
@@ -301,12 +260,6 @@ class TestKafkaSystemAdmin {
     val spec = StreamSpec.createCoordinatorStreamSpec(topic, "kafka")
     systemAdmin.createStream(spec)
     validateTopic(topic, 1)
-    val topicMetadataMap = TopicMetadataCache.getTopicMetadata(Set(topic), "kafka", metadataStore.getTopicInfo)
-    assertTrue(topicMetadataMap.contains(topic))
-    val topicMetadata = topicMetadataMap(topic)
-    val partitionMetadata = topicMetadata.partitionsMetadata.head
-    assertEquals(0, partitionMetadata.partitionId)
-    assertEquals(3, partitionMetadata.replicas.size)
```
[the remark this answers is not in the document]
mynameborat: Yeah. I tried to keep this by using APIs from `adminClient` and `adminZkClient`but in vain. I'd have to replicate the logic in the file `ClientsUtilTopicMetadataStore.scala` that was removed as part of the PR. Since we no longer have any active code path within our codebase that reads this data, I chose to remove it.

mynameborat:
Summary of 44d8c97
* Combined StandaloneIntegrationTestHarness and AbstractIntegrationTestHarness into IntegrationTestHarness. 
* Modified some more tests to use `adminClient` APIs and got rid of deprecated `zkUtils`
* Addressed Cameron's review comments.
* Fixed a bug around metric where in records-lag is not longer prefixed with the topic partition rather topic partitions are part of tags within the metric names.

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemAdmin.scala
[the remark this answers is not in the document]
cameronlee314: The end-to-end check seems unnecessary anyways for a unit test.
However, since you are removing that check, it might be good to have a test which uses Mockito to verify that the correct configurations are passed to the admin client when creating topics. If I recall correctly, in the past, we may have had some issues which involved replication factors (not sure if they were this exact flow though).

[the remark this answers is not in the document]
mynameborat: Agreed. Let me do the Mockito way 👍

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemAdmin.scala
```diff
@@ -89,49 +91,49 @@ object TestKafkaSystemAdmin extends KafkaServerTestHarness {
...
-
-          val topicDescription =
-            adminClient.describeTopics(JavaConverters.asJavaCollectionConverter(Set(topic)).asJavaCollection)
-              .all().get().get(topic)
+        val topicDescriptionFutures = adminClient.describeTopics(
+            JavaConverters.asJavaCollectionConverter(Set(topic)).asJavaCollection).all()
+        val topicDescription = topicDescriptionFutures.get(500, TimeUnit.MILLISECONDS)
+            .get(topic)
 
         done = expectedPartitionCount == topicDescription.partitions().size()
```
cameronlee314: 1. Can `topicDescription` be null?
2. If done is true, then can we break out of this loop?

samza-test/src/test/java/org/apache/samza/test/framework/StreamApplicationIntegrationTestHarness.java
```diff
@@ -295,4 +227,20 @@ public Config getConfig() {
       return this.config;
     }
   }
+
+  @Override
+  protected KafkaProducer createProducer() {
```
cameronlee314: It seems a bit odd that this test harness needs slightly different serialization/deserialization for createProducer/createConsumer compared to `IntegrationTestHarness`. If it is difficult to make them consistent, could you please put a comment about why you have these overrides?
You could also consider having helpers in `IntegrationTestHarness` to build the configs, just in case someone needs to add something in the future.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkLocalApplicationRunner.java
```diff
@@ -159,42 +152,25 @@ public void setUp() {
...
-        Assert.fail(String.format("Unable to create kafka topic: %s.", kafkaTopic));
-      }
-    }
+    List<NewTopic> newTopics =
+        ImmutableList.of(inputKafkaTopic, outputKafkaTopic, inputSinglePartitionKafkaTopic, outputSinglePartitionKafkaTopic)
+            .stream()
+            .map(topic -> new NewTopic(topic, 5, (short) 1))
+            .collect(Collectors.toList());
+
+    createTopics(newTopics);
```
cameronlee314: Should this fail if `createTopics` returns false?

samza-test/src/test/scala/org/apache/samza/test/integration/StreamTaskTestUtil.scala
```diff
@@ -133,43 +130,36 @@ object StreamTaskTestUtil {
...
   def createTopics {
-    AdminUtils.createTopic(
-      zkUtils,
-      INPUT_TOPIC,
-      TOTAL_TASK_NAMES,
-      REPLICATION_FACTOR)
+    adminClient.createTopics(Collections.singleton(new NewTopic(INPUT_TOPIC, TOTAL_TASK_NAMES, REPLICATION_FACTOR.shortValue())))
   }
 
   def validateTopics {
```
cameronlee314: Not sure if this is feasible/practical to combine, but it looks like some of this code is similar to code from `TestKafkaSystemAdmin`, such as `validateTopics`.

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemAdmin.scala
mynameborat: 1. Nope. 
2. Thanks for bringing it up. I found a corner case with current implementation. I changed it a little without introducing additional control flow statements.

gradle/dependency-versions.gradle
```diff
@@ -34,7 +34,7 @@
   jodaTimeVersion = "2.2"
   joptSimpleVersion = "3.2"
   junitVersion = "4.12"
-  kafkaVersion = "0.11.0.2"
+  kafkaVersion = "2.0.1"
```
shanthoosh: After this, through kafka 2.0, we would pull in slf4j-log4j12 transitively. 

Do we know if this would cause logging problems with different log4j-1/log4j-2 combinations that exists in samza.  Or it doesn't matter?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -668,22 +631,86 @@ protected Properties createAdminClientProperties() {
...
+    if (appConfig.getAppMode() == ApplicationConfig.ApplicationMode.BATCH) {
+      StreamConfig streamConfig = new StreamConfig(config);
+      intermedidateStreamProperties = JavaConverters.asJavaCollectionConverter(streamConfig.getStreamIds())
+          .asJavaCollection()
+          .stream()
+          .filter(streamConfig::getIsIntermediateStream)
+          .collect(Collectors.toMap(Function.identity(), streamId -> {
+            Properties properties = new Properties();
+            properties.putAll(streamConfig.getStreamProperties(streamId));
+            properties.putIfAbsent("retention.ms", String.valueOf(KafkaConfig.DEFAULT_RETENTION_MS_FOR_BATCH()));
```
shanthoosh: Would be better to switch from hard-coded string: `retention.ms` to `TopicConfig.RETENTION_MS_CONFIG ` here.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -668,22 +631,86 @@ protected Properties createAdminClientProperties() {
...
+            return properties;
+          }));
+    }
+
+    return intermedidateStreamProperties;
+  }
+
+  private Properties getCoordinatorStreamProperties(KafkaConfig config) {
+    Properties coordinatorStreamProperties = new Properties();
+    coordinatorStreamProperties.put("cleanup.policy", "compact");
```
shanthoosh: 1. Better to replace hard-coded string(`cleanup.policy`) with `TopicConfig.RETENTION_MS_CONFIG `. 
2. Better to replace hard-coded string: `compact` with `TopicConfig.CLEANUP_POLICY_COMPACT `.
3. Better to replace `segment.policy` with `TopicConfig. SEGMENT_BYTES_CONFIG `

samza-test/src/test/java/org/apache/samza/test/framework/StreamApplicationIntegrationTestHarness.java
mynameborat: I added a comment and wired in the serde through configuration.

samza-test/src/test/scala/org/apache/samza/test/integration/StreamTaskTestUtil.scala
mynameborat: It is doable and I considered it. I deferred it due to following

1. I didn't want to add more churn to scala code. StreamTaskTestUtil still uses some of the TestUtils deprecated code and I chose to clean up all at once as a separate PR (maybe rewrite them in java)
2. I didn't want to expand the scope of this PR.

I have created a follow up ticket (https://issues.apache.org/jira/browse/SAMZA-2145) to clean up the test and consolidate the common code to some helper or move it to test harness.

gradle/dependency-versions.gradle
mynameborat: We already pull in `slf4j-log4j12` as part of build time dependency in samza-log4j module. Since we have a direct pinned dependency, I suppose it shouldn't matter.

Let me keep an eye for any potential logging issues.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -628,10 +589,12 @@ public void validateStream(StreamSpec streamSpec) throws StreamValidationExcepti
...
-      if (adminClientForDelete == null) {
-        adminClientForDelete = kafka.admin.AdminClient.create(createAdminClientProperties());
-      }
-      KafkaSystemAdminUtilsScala.deleteMessages(adminClientForDelete, offsets);
+      Map<TopicPartition, RecordsToDelete> recordsToDelete = offsets.entrySet()
+          .stream()
+          .collect(Collectors.toMap(entry ->
+              new TopicPartition(entry.getKey().getStream(), entry.getKey().getPartition().getPartitionId()),
+              entry -> RecordsToDelete.beforeOffset(Long.parseLong(entry.getValue()) + 1)));
+      adminClient.deleteRecords(recordsToDelete);
```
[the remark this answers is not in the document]
mynameborat: Done 1 & 2 
3. We have been using AdminClient for all flows except deletion path. We don't have any of the backoff configurations configurable for the other flows and I plan to keep it the same for deletion too.

We can follow up if there is a need for it in a separate PR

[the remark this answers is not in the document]
mynameborat: Good point. I am caught up on this since the existing behavior wasn't blocking either and does a best effort on the delete. I debated between retaining existing behavior vs blocking. Since blocking on this would mean blocking commit thread which does seem significant especially given this is an optimization to reduce the data footprint for our intermediate topics.

Let me log an error in case of failures on this future.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -594,7 +593,13 @@ public void deleteMessages(Map<SystemStreamPartition, String> offsets) {
           .collect(Collectors.toMap(entry ->
               new TopicPartition(entry.getKey().getStream(), entry.getKey().getPartition().getPartitionId()),
               entry -> RecordsToDelete.beforeOffset(Long.parseLong(entry.getValue()) + 1)));
-      adminClient.deleteRecords(recordsToDelete);
+
+      adminClient.deleteRecords(recordsToDelete).all().whenComplete((ignored, exception) -> {
```
cameronlee314: Just to double check, `whenComplete` is async, right?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -594,7 +593,13 @@ public void deleteMessages(Map<SystemStreamPartition, String> offsets) {
...
+        if (exception != null) {
+          LOG.error("Delete message failed for SSPs {} due to {}", offsets.keySet(), exception);
```
cameronlee314: I think this will end up not printing out the stack trace. I think you need to use the `LOG.error(String, Exception)` instead.

samza-test/src/test/java/org/apache/samza/test/framework/StreamApplicationIntegrationTestHarness.java
```diff
@@ -229,18 +232,32 @@ public Config getConfig() {
...
-    return new KafkaProducer<>(kafkaConfig, STRING_SERIALIZER, STRING_SERIALIZER);
+  protected Properties createProducerConfigs() {
+    Properties producerProps = new Properties();
+    producerProps.setProperty(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers());
+    /*
+     * Stream application tests uses string serde for both key and value.
+     * The default serde for the test producer in IntegrationTestHarness uses
+     * byte serde for value and string for key.
+     */
+    producerProps.setProperty(KEY_SERIALIZER_CLASS_CONFIG, STRING_SERIALIZER);
```
cameronlee314: Could you implement this by calling `super.createProducerConfigs`, and then just override the serializers? Same for `createConsumerConfigs`.

gradle/dependency-versions.gradle
shanthoosh: Discussed offline.

Since we're bumping up kafka-client version to 2.0, it would be better to run the  perf tests. If possible try running [kafka-read-write-perf](https://github.com/apache/samza/blob/master/samza-test/src/main/config/perf/kafka-read-write-performance.properties) test before landing this patch and verify there's no performance degradation(in terms of QPS/poll-latency).

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
mynameborat: 1. Since this is an optimization in the background to clean up data for intermediate topics, it is not be desirable to bubble it up and kill the application.
2. The return future is of type Void and ignored doesn't carry any information.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
mynameborat: Yes

samza-kafka/src/test/scala/org/apache/samza/system/kafka/TestKafkaSystemAdmin.scala
```diff
@@ -169,59 +165,21 @@ class TestKafkaSystemAdmin {
...
-    assertEquals(expectedSystemStream2Partition0Metadata, stream2PartitionMetadata.get(new Partition(0)))
-    assertEquals(expectedSystemStream2Partition1Metadata, stream2PartitionMetadata.get(new Partition(1)))
-  }
-
   @Test
   def testShouldGetOldestNewestAndNextOffsets {
-    // Create an empty topic with 50 partitions, but with no offsets.
-    createTopic(TOPIC, 50)
-    validateTopic(TOPIC, 50)
+    // Reduced the partition count to 25 instead of 50 to prevent tests from failing. In the current setup,
```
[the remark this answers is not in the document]
mynameborat: Discussed offline. The essence of the tests remains the same except the higher number of partitions doesn't play well with the current Embedded Kafka setup (3 brokers) in terms of replication assignment especially with the new create APIs which used in the tests which use RackAwareMode as Enforced by default.

I didn't want to spend too much time on getting it to work since the essence of the tests still remains the same.

gradle/dependency-versions.gradle
mynameborat: will do

samza-test/src/test/java/org/apache/samza/test/harness/IntegrationTestHarness.java
```diff
@@ -0,0 +1,192 @@
...
+
+  protected boolean createTopic(String topicName, int numPartitions, int replicationFactor) {
+    return createTopics(Collections.singleton(new NewTopic(topicName, numPartitions, (short) replicationFactor)));
+  }
+
+  protected boolean createTopics(Collection<NewTopic> newTopics) {
+    boolean createStatus = true;
+
+    try {
+      CreateTopicsResult resultFuture =
```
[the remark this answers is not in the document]
mynameborat: Synced up offline. 
Taking up dependency on the Samza code path for setting up environment (in this case topic creation) is unnecessary and doesn't buy in terms of code reuse.

Also, createStreams API in KafkaSystemAdmin also has bunch of boiler plate code and would mean the creation path is constrained by the parameters that are used to create SystemAdmin which compromises flexibility.

shanthoosh:
@mynameborat 
Thanks for the changes. Merged it to trunk.

PR #952, opened 2019-03-15
SEP-19: Improved standby-aware container allocation for active-containers on job redeploys

rmatharu:


samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java
```diff
@@ -226,15 +244,35 @@ private void initiateActiveContainerFailover(String containerID, String resource
...
+
+    // Now, we iterate over the list of last-known standbyHosts to check if anyone of them has not already been tried
+    Map<String, String> containerToHostMapping = this.samzaApplicationState.jobModelManager.jobModel().getAllContainerLocality();
+
+    for (String standbyContainerID : this.standbyContainerConstraints.get(activeContainerID)) {
+      String standbyHost = containerToHostMapping.get(standbyContainerID);
+
+      // if there is no last-known host for the standby, continue
+      if (standbyHost == null) {
+        continue;
```
vjagadish1989: Should log here

vjagadish1989: the flow is a bit circuitous. avoid the `continue` and consider returning `any-host` here?

rmatharu: i agree, simplified it a bit. 
but we cant return any-host only once we're done iterating over all standbyContainerIDs.

samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java
```diff
@@ -161,54 +161,72 @@ private void handleStandbyContainerStop(String standbyContainerID, String resour
...
-      failoverMetadata.updateStandbyContainer(standbyResourceID, standbyHost);
+      log.info("No running standby container to stop on host {}, making a resource-request for placing {} on {}", standbyHost, activeContainerID, standbyHost);
+      FailoverMetadata failoverMetadata = this.registerActiveContainerFailure(activeContainerID, resourceID);
 
-      log.info("Initiating failover and stopping standby container, found standbyContainer {} = resource {}, "
-          + "for active container {}", standbyContainerId, standbyResourceID, containerID);
+      // record the resource request, before issuing it to avoid race with allocation-thread
+      SamzaResourceRequest resourceRequestForActive = containerAllocator.getResourceRequest(activeContainerID, standbyHost);
+      failoverMetadata.recordResourceRequest(resourceRequestForActive);
+      containerAllocator.issueResourceRequest(resourceRequestForActive);
```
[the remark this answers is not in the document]
rmatharu: i need to recordResource requests, to handle expired requests, so i create one and then use
issueResourceRequest to issue it.

rmatharu:
Addressed all comments. 
thumbsup => comment addressed.

samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java
```diff
@@ -224,17 +257,40 @@ private void initiateActiveContainerFailover(String containerID, String resource
...
+    // We iterate over the list of last-known standbyHosts to check if anyone of them has not already been tried
+    for (String standbyContainerID : this.standbyContainerConstraints.get(activeContainerID)) {
+
+      String standbyHost = this.samzaApplicationState.jobModelManager.jobModel().
+          getContainerToHostValue(standbyContainerID, SetContainerHostMapping.HOST_KEY);
+
+      if (standbyHost == null) {
+        log.info("No last known standbyHost for container {}", standbyContainerID);
+
+      } else if (standbyHost != null && failoverMetadata.isPresent() && failoverMetadata.get()
```
vjagadish1989: by definition, the condition `standbyHost != null` is true at L:280. Can we remove it

samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java
```diff
@@ -161,56 +163,87 @@ private void handleStandbyContainerStop(String standbyContainerID, String resour
...
+        FailoverMetadata failoverMetadata = this.registerActiveContainerFailure(activeContainerID, resourceID);
+
+        // record the resource request, before issuing it to avoid race with allocation-thread
+        SamzaResourceRequest resourceRequestForActive =
+            containerAllocator.getResourceRequest(activeContainerID, standbyHost);
+        failoverMetadata.recordResourceRequest(resourceRequestForActive);
+        containerAllocator.issueResourceRequest(resourceRequestForActive);
+        samzaApplicationState.failoversToStandby.incrementAndGet();
+      } else {
+        // if there is a running standby-containers on the standbyHost, we issue a stop (the stopComplete callback completes the remainder of the flow)
```
vjagadish1989: nice explanations;

PR #953, opened 2019-03-16
Set job coordinator replication factor configuration for standalone.

shanthoosh:


shanthoosh:
@vjagadish1989
Can you take a look.

PR #954, opened 2019-03-17
SAMZA-2129: Move the offset comparison check to choose the lowest offset to SystemConsumers.

shanthoosh:

* Moves the offset comparison check to SystemConsumers
 
     * Currently the offset comparator check to find a lowest for a `SystemStreamPartition` is duplicated in some implementations of  SystemConsumer API.
     * Some `SystemConsumer` implementations do not perform this offset comparator check
in the implementation of register method.
     * Moving this one level up from `SystemConsumer.register(SystemStreamPartition, offset)` API implementation to  `SystemConsumers.register(SystemStreamPartition, offset)` API implementation removes unnecessary duplication and ensures functional correctness.

* Fixes to the test-cases in `TestSystemConsumer`.

There should be no functional breakages introduced by moving the offset-comparator check to `SystemConsumers` layer. Here's why. Metadata and I/O topics currently use SystemConsumer API to read data. 
* Checkpoint topic: This is a stream with one partition and is log-compacted. Samza-container reads from the beginning of this stream. There's no offset comparisons required for this topic.
* ChangeLog topic: To read change-log topic-partition, SamzaContainer uses `StorageManagerUtil.getStartingOffset` to read from the change-log topic and the offset comparator check to choose the lowest offset is currently performed within it.
* Coordinator topic: This is a stream with one partition and is log-compacted. SamzaContainer/ApplicationMaster is currently reads from the beginning of this stream to get the entire state.
*  I/O topics: `SamzaContainer` currently reads  data from input and output topics through `SystemConsumers` API. Moving this offset comparator check should not affect the existing functionality.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
+++ b/samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
@@ -38,6 +38,10 @@ public class SystemAdmins {
     this.systemAdminMap = systemConfig.getSystemAdmins();
   }
 
+  public SystemAdmins(Map<String, SystemAdmin> systemAdminMap) {
+    this.systemAdminMap = systemAdminMap;
+  }
+
   /**
    * Creates a new instance of {@link SystemAdmins} with an empty admin mapping.
    * @return New empty instance of {@link SystemAdmins}
--- a/samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
+++ b/samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
@@ -38,10 +38,6 @@ public class SystemAdmins {
     this.systemAdminMap = systemConfig.getSystemAdmins();
   }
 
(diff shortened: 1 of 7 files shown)


shanthoosh:
@dnishimura Please take a look when you have a chance.

samza-core/src/main/java/org/apache/samza/system/SystemAdmins.java
```diff
@@ -38,6 +38,10 @@ public SystemAdmins(Config config) {
     this.systemAdminMap = systemConfig.getSystemAdmins();
   }
 
+  public SystemAdmins(Map<String, SystemAdmin> systemAdminMap) {
```
cameronlee314: Is this only for tests? Could you mock it instead in the tests? It would be good to avoid having public methods just for tests.

shanthoosh: Done.

PR #956, opened 2019-03-18
SAMZA-2132: Flatten startpoint key when serialized.

dnishimura:
StartpointKey has unnecessary redundancies when serialized. Example: 
```
"{\"systemStreamPartition\":{\"system\":\"test\",\"stream\":\"test\",\"partition\":{\"partitionId\":2},\"systemStream\":{\"system\":\"test\",\"stream\":\"test\"}},\"taskName\":{\"taskName\":\"Partition 2\"}}"
```
The serialized key should should be flattened to 4 elements: system, stream, partition ID and taskName.

@shanthoosh please take a look.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/startpoint/StartpointKey.java
+++ b/samza-core/src/main/java/org/apache/samza/startpoint/StartpointKey.java
@@ -22,11 +22,11 @@ import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
 import org.apache.samza.container.TaskName;
 import org.apache.samza.system.SystemStreamPartition;
-import org.codehaus.jackson.annotate.JsonIgnore;
-import org.codehaus.jackson.annotate.JsonProperty;
+import org.codehaus.jackson.map.annotate.JsonSerialize;
 
 
-final class StartpointKey {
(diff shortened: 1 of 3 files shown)


samza-core/src/main/java/org/apache/samza/startpoint/StartpointKey.java
```diff
@@ -41,12 +41,29 @@
     this.taskName = taskName;
   }
 
+  @JsonIgnore
```
shanthoosh: The idea of including only relevant fields from `StartpointKey` to reduce redundancy is a good one. 

However, rather than marking individual fields in model POJO as included or not, it would be better to have a custom `JsonSerializer` implementation and have this logic to include fields we want in it. Similar to [SystemStreamPartitionSerializer](https://github.com/apache/samza/blob/master/samza-core/src/main/java/org/apache/samza/serializers/model/SamzaObjectMapper.java#L203)

What do you think?

dnishimura: @shanthoosh good idea. I'm always for being explicit. Please see the latest commit.

PR #961, opened 2019-03-19
SAMZA-2134:Enable table rate limiter by default.

dengpanyin:
By default rate limiter will be enabled. The user need to provide a rate limiter for the
table descriptor or disable read/write rate limiter explicitly.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
+++ b/samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
@@ -83,6 +83,12 @@ public class RemoteTableDescriptor<K, V> extends BaseTableDescriptor<K, V, Remot
   // Rate limiter for client-side throttling; it is set by withRateLimiter()
   private RateLimiter rateLimiter;
 
+  // Indicate whether read rate limiter is enabled or not
+  private boolean enableReadRateLimiter = true;
+
+  // Indicate whether write rate limiter is enabled or not
+  private boolean enableWriteRateLimiter = true;
+
   // Rates for constructing the default rate limiter when they are non-zero
   private Map<String, Integer> tagCreditsMap = new HashMap<>();
 
@@ -170,6 +176,41 @@ public class RemoteTableDescriptor<K, V> extends BaseTableDescriptor<K, V, Remot
     this.rateLimiter = rateLimiter;
     this.readCreditFn = readCreditFn;
     this.writeCreditFn = writeCreditFn;
+    return this;
+  }
+
+  /**
+   * Disable both read and write rate limiter. If the read rate limiter is enabled, the user must provide a rate limiter
+   * by calling {@link #withRateLimiter(RateLimiter, TableRateLimiter.CreditFunction, TableRateLimiter.CreditFunction)}
+   * or {@link #withReadRateLimit(int)}. If the write rate limiter is enabled, the user must provide a rate limiter
+   * by calling {@link #withRateLimiter(RateLimiter, TableRateLimiter.CreditFunction, TableRateLimiter.CreditFunction)}
+   * or {@link #withWriteRateLimit(int)}. By default, both read and write rate limiters are enabled.
+   *
+   * @return this table descriptor instance.
+   */
+  public RemoteTableDescriptor<K, V> withDisableRateLimiter() {
+    withDisableReadRateLimiter();
+    withDisableWriteRateLimiter();
+    return this;
+  }
+
+  /**
+   * Disable the read rate limiter.
+   *
+   * @return this table descriptor instance.
+   */
+  public RemoteTableDescriptor<K, V> withDisableReadRateLimiter() {
+    this.enableReadRateLimiter = false;
+    return this;
+  }
+
+  /**
+   * Disable the write rate limiter.
+   *
+   * @return this table descriptor instance.
+   */
+  public RemoteTableDescriptor<K, V> withDisableWriteRateLimiter() {
+    this.enableWriteRateLimiter = false;
     return this;
   }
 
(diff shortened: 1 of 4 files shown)


dengpanyin:
@weisong44  Can you help view this RB?  Thanks

dengpanyin:
Add @sborya to take a look as well.

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -173,6 +179,41 @@ public RemoteTableDescriptor(String tableId) {
...
+  /**
+   * Disable both read and write rate limiter. If the read rate limiter is enabled, the user must provide a rate limiter
+   * by calling {@link #withRateLimiter(RateLimiter, TableRateLimiter.CreditFunction, TableRateLimiter.CreditFunction)}
+   * or {@link #withReadRateLimit(int)}. If the write rate limiter is enabled, the user must provide a rate limiter
+   * by calling {@link #withRateLimiter(RateLimiter, TableRateLimiter.CreditFunction, TableRateLimiter.CreditFunction)}
+   * or {@link #withWriteRateLimit(int)}. By default, both read and write rate limiters are enabled.
+   *
+   * @return this table descriptor instance.
+   */
+  public RemoteTableDescriptor<K, V> withDisableRateLimiter() {
```
vjagadish1989: Can you explain what we are solving here? If the goal is to always provide a rate-limiter with remote-table,  maybe there's a simple way to get there.

For eg: we could mandate that users provide a readQPS/writeQPS or a rateLimiter when they construct a remote table. Users who prefer to avoid rate-limiting can set read/write QPS to infinity.

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -173,6 +179,41 @@ public RemoteTableDescriptor(String tableId) {
...
+   *
+   * @return this table descriptor instance.
+   */
+  public RemoteTableDescriptor<K, V> withDisableReadRateLimiter() {
+    this.enableReadRateLimiter = false;
+    return this;
+  }
+
+  /**
+   * Disable the write rate limiter.
```
vjagadish1989: There are multiple combination of APIs in this class.

- APIs that set read rate-limits and write rate-limits
- APIs that set read-rate-limiter and write-rate-limiter
- APIs that turn off rate-limiting

It maybe confusing to deal with these combinations / overrides and validate which of them are compatible at run-time. 

Instead, I'd rather us not expose the "disable" API combinations and enforce limits during construction
```
RemoteTableDescriptor(int readQPS, int writeQPS);
RemoteTableDescriptor(RateLimiter limiter);
```

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
dengpanyin: @vjagadish1989 
Having a rate limier for a remote read/write table is highly recommended, and we will enable the rate limiter by default so that the user can either explicitly set a rate limiter or disable it. There is no change if the user is currently using rate limiter. Otherwise, the user either needs to set one or disable it using this API based on their use case. 
Ideally we should set a default one, but we can't get a correct estimation, and could result in problems, so we push the user to make the decision. 

This rate limiter is still an optional field. For optional field we don't put in the constructors.

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
dengpanyin: @vjagadish1989 
As mentioned earlier, we typically don't put optional field in the constructor. 
Also when talking to @weisong44 earlier, he recommended to have this API for the users.

On the other hand, currently, we already have the following APIs to set rate limiter:

(1) withRateLimiter(RateLimiter rateLimiter,
      TableRateLimiter.CreditFunction<K, V> readCreditFn,
      TableRateLimiter.CreditFunction<K, V> writeCreditFn)

(2) withReadRateLimit(int creditsPerSec)

(3) withWriteRateLimit(int creditsPerSec)

If we further allow the user to config rate limiter from the constructor, the user may get confused if they have a rate limiter value in the constructor and a rate limiter configured using one of the above APIs.

Also, the user have to give a magic number in order to disable the corresponding rate limiter. 
If the user config a very large value (not -1) with the intension to disable rate limiter, 
we still need to create a rate limiter using that value and apply the rate limiter towards read/write (wasting memory and CPU). 

Thinking all above, it seems cleaner to have an explicit API to disable rate limiters if the user has to.

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -173,6 +179,41 @@ public RemoteTableDescriptor(String tableId) {
...
+    withDisableWriteRateLimiter();
+    return this;
+  }
+
+  /**
+   * Disable the read rate limiter.
+   *
+   * @return this table descriptor instance.
+   */
+  public RemoteTableDescriptor<K, V> withDisableReadRateLimiter() {
```
weisong44: can we call this withReadRateLimiterDisabled() ?

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -173,6 +179,41 @@ public RemoteTableDescriptor(String tableId) {
...
+    this.enableReadRateLimiter = false;
+    return this;
+  }
+
+  /**
+   * Disable the write rate limiter.
+   *
+   * @return this table descriptor instance.
+   */
+  public RemoteTableDescriptor<K, V> withDisableWriteRateLimiter() {
```
weisong44: Can we call this withWriteRateLimiterDisabled() ?

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
dengpanyin: @weisong44  
API names updated to be:
withRateLimiterDisabled
withReadRateLimiterDisabled
withWriteRateLimiterDisabled

Thanks

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
dengpanyin: Renamed. Thanks

PR #962, opened 2019-03-20
SAMZA-2135: Provide a way inject ExternalContext to TestRunner

Sanil15:
- Added a public API to pass ExternalContext to TestRunner
- Refactored existing tests to test it

### DIFF
--- a/samza-test/src/test/java/org/apache/samza/test/framework/StreamTaskIntegrationTest.java
+++ b/samza-test/src/test/java/org/apache/samza/test/framework/StreamTaskIntegrationTest.java
@@ -107,7 +107,7 @@ public class StreamTaskIntegrationTest {
         .of(MyStreamTestTask.class)
         .addInputStream(imid, inputList)
         .addOutputStream(imod, 1)
-        .addExternalContext(new TestContext())
+        .addExternalContext(new TestContext(10))
         .run(Duration.ofSeconds(1));
 
     Assert.assertThat(TestRunner.consumeStream(imod, Duration.ofMillis(1000)).get(0),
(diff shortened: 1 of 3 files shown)


Sanil15:
@cameronlee314 please have a look!

samza-test/src/test/java/org/apache/samza/test/framework/StreamTaskIntegrationTest.java
```diff
@@ -263,4 +268,10 @@ public void genData(Map<Integer, List<KV>> inputPartitionData, Map<Integer, List
     }
   }
 
-}
\ No newline at end of file
+}
+
+class TestContext implements ExternalContext {
+  public int getMultiplier() {
+    return 10;
```
cameronlee314: To give a slightly better example of how `ExternalContext` might be used, maybe pass the multiplier as an argument when creating a `TestContext` instead of hardcoding it into `getMultiplier`.

Sanil15: sure

samza-test/src/test/java/org/apache/samza/test/framework/StreamTaskIntegrationTest.java
```diff
@@ -263,4 +268,16 @@ public void genData(Map<Integer, List<KV>> inputPartitionData, Map<Integer, List
     }
   }
 
-}
\ No newline at end of file
+}
+
+class TestContext implements ExternalContext {
+  private int multiplier;
```
cameronlee314: Minor: please add final modifier for immutable field

Sanil15: sure

PR #966, opened 2019-03-22
Moving the Fixes from internal AvroRelConverter

srinipunuru:
Misc fixes that are made in internal copy of AvroRelConverter.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
@@ -31,6 +31,7 @@ import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.generic.IndexedRecord;
 import org.apache.calcite.avatica.util.ByteString;
+import org.apache.commons.lang3.Validate;
 import org.apache.samza.SamzaException;
 import org.apache.samza.config.Config;
 import org.apache.samza.operators.KV;
@@ -180,8 +181,6 @@ public class AvroRelConverter implements SamzaRelConverter {
           continue;
         }
         Object relObj = values.get(index);
-        if (relObj == null) {
-        }
         Schema fieldSchema = schema.getField(fieldName).schema();
         record.put(fieldName, convertToAvroObject(relObj, getNonNullUnionSchema(fieldSchema)));
       }
@@ -190,7 +189,7 @@ public class AvroRelConverter implements SamzaRelConverter {
     return record;
   }
 
-  static Object convertToAvroObject(Object relObj, Schema schema) {
+  public static Object convertToAvroObject(Object relObj, Schema schema) {
     if (relObj == null) {
       return null;
     }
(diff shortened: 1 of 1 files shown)


samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
```diff
@@ -170,22 +179,22 @@ static public GenericRecord convertToGenericRecord(SamzaSqlRelRecord relRecord,
               schema.getName(), schema.getNamespace(), fieldName);
           continue;
         }
-
         Object relObj = values.get(index);
+        if (relObj == null) {
```
shenodaguirguis: ?

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
```diff
@@ -271,45 +281,46 @@ public Object convertToJavaObject(Object avroObj, Schema schema) {
     }
   }
 
-  private boolean isSchemaCompatible(Object avroObj, Schema unionSchema) {
-    if (unionSchema.getType() == Schema.Type.NULL && avroObj == null) {
-      return true;
+  private static boolean isSchemaCompatibleWithRelObj(Object relObj, Schema unionSchema) {
+    if (unionSchema.getType() == Schema.Type.NULL) {
```
shenodaguirguis: verify/handle unionSchema != null?

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
```diff
@@ -271,45 +281,46 @@ public Object convertToJavaObject(Object avroObj, Schema schema) {
...
+        return relObj instanceof Float || relObj instanceof Double;
       default:
         return true;
     }
   }
 
-  private static boolean isSchemaCompatibleWithRelObj(Object relObj, Schema unionSchema) {
-    if (unionSchema.getType() == Schema.Type.NULL && relObj == null) {
+  private static boolean isSchemaCompatible(Object avroObj, Schema unionSchema) {
+    if (unionSchema.getType() == Schema.Type.NULL && avroObj == null) {
```
shenodaguirguis: shouldn't (avroObj == null) move to be the return?

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
srinipunuru: Dead code, needs to be removed.

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroRelConverter.java
srinipunuru: I believe you are asking whether we should handle null values for unionSchema?

shenodaguirguis: I was actually wondering if we should guard against NullPointerException when  unionSchema of Schema is null

PR #971, opened 2019-03-25
SAMZA-2144: In TaskInstance, bypass "SSP caught-up" check if offset is set by a startpoint

dnishimura:
TaskInstance maintains a "SSP caught-up map" to ensure that all tasks for the same SSP have caught up to he same checkpoint offset before processing any messages. However, this does not apply to startpoints since it expects to begin processing messages starting from the startpoint's offset position.

@shanthoosh please take a look.


dnishimura:
Closing since incoming API modifications will not require this PR.
FYI - @shanthoosh

PR #975, opened 2019-03-26
SAMZA-2148: Fix the rendering of bash and sql scripts  in samza-sql release validation page.

shanthoosh:
Currently, the bash/sql scripts available in samza-sql release validation page were not using the proper markdown rules and hence they're not rendered properly. 

Please refer to current [samza-sql-validation-page](https://github.com/apache/samza/blob/master/docs/learn/tutorials/versioned/samza-sql.md) to know more about the rendering problems.

We identified this problem during samza-1.1 release-validation and this PR fixes the above problem.


shanthoosh:
@srinipunuru Please take a look when you have a chance. I've attached the screenshot of the rendered validation page created with this patch. 
screencapture-file-home-svenkata-Downloads-Samza-Release-Procedure-md-1-html-2019-03-25-18_14_50

PR #977, opened 2019-03-27
SAMZA-2150: Update the config/index.md for 1.1 release.

shanthoosh:
* Update the release variable in config.yml to generate javadocs.
* Update the index.md to point to 1.1-src url that would allow users to download the samza-1.1 source.


shanthoosh:
@vjagadish1989 Can you take a look.

PR #978, opened 2019-03-27
SAMZA-1733 : Making MetricsSnapshotReporter look for appclass then taskclass

rmatharu:


samza-core/src/main/scala/org/apache/samza/metrics/reporter/MetricsSnapshotReporterFactory.scala
```diff
@@ -45,10 +45,8 @@ class MetricsSnapshotReporterFactory extends MetricsReporterFactory with Logging
     val jobId = config
       .getJobId
 
-    val taskClass = config
-      .getTaskClass
-      .orElse(Option(new ApplicationConfig(config).getAppClass()))
-      .getOrElse(throw new SamzaException("No task or app class defined for config."))
+    val taskClass = Option(new ApplicationConfig(config).getAppClass())
```
prateekm: @rmatharu Just wondering, what issue does changing the order fix?

shanthoosh: 1. I think when this gets used with low-level API(`StreamTask`/`AsyncStreamTask`) jobs, then `app.class` won't be defined.  
2. Would be better to log a message at warn-level that we're falling back to `task.class`  when we don't find `app.class`  in config.
3. Within LI, there's yet another configuration `task.subtask.class` where users specify their task implementation when using offspring-helper. In that case `task.class` always defaults to a specific samza-predefined-class. Would be better to think if that would affect any of the functionality here.

rmatharu: At LI, task.class is defined to empty in the global-app-def, but app.class is not predefined. That necessitates this fix.

The logic here is simple, if there is an app.class, use it. If there is none, use task.class, if there is no task.class (and no app.class), raise an exception. 
I dont see any point in logging a warning, since we can't improve-upon/act-upon it.

shanthoosh: >> I dont see any point in logging a warning, since we can't improve-upon/act-upon it

task.class is deprecated post 1.0. This log-warning is required to communicate to users that they're using deprecated config in their job.

>> Task.subtask.class is not valid post samza 1.0. 

No, it is just deprecated and it is not removed. Couple of high-end jobs(incl atc, concourse) still use it after 1.0 release.

PR #979, opened 2019-03-27
SAMZA-2110 : Support for deleting records from a table

srinipunuru:
1. Commit adds support for deleting records from table by adding a special __op__ field. SQL users can use this field to say whether the row needs to be deleted in the output table. This is a stop mechanism till we support DELETE operation in SQL. 
2. Adds support for pluggable RelProviders to add the special fields __key__. This is useful if the relProvider knows the type of __key__ and wants to set it explicitly.
3. Removes the hack where field positions are incremented by 1 in RelSchemaConverter to support the special field __key__. With this change the knowledge of the special fields exists only in queryTranslator. In future we could consider moving this logic to RelSchemaProvider.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/sql/SamzaSqlRelRecord.java
+++ b/samza-api/src/main/java/org/apache/samza/sql/SamzaSqlRelRecord.java
@@ -120,4 +120,8 @@ public class SamzaSqlRelRecord implements Serializable {
     String valueStr = Joiner.on(",").useForNull("null").join(fieldValues);
     return "[Names:{" + nameStr + "} Values:{" + valueStr + "}]";
   }
+
+  public boolean containsField(String name) {
+    return fieldNames.indexOf(name) != -1;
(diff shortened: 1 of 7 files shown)


srinipunuru:
@atoomula 

> * delete events from cc streams. eg: "select * from brooklin.stream", "select f1, f2 from brooklin.stream"

With this new model there needs to be separate sql statement for delete. I think this model may be better longer term, because When we have delete from support. users need to write two sql statements on for propagating inserts and updates and other one for deletes. This model is inline with that.

> * Non-explicit delete events. eg: "select f1, f2 from kafka.topic". In this case both f1 and f2 could be null. Deleting such an event from destination store is unclear.

When user does select f1, f2 and when both f1 and f2 are null, it is possible that user wants us to insert a record with each fields being null.  Assuming that the behavior in this case should be delete is non-intuitive and also problematic because we may never be able to support the above requested behavior.

PR #981, opened 2019-04-02
SAMZA-2152 : Bugfix: Making KafkaSytemAdmin's metadataConsumer accesses thread-safe, enabling StreamRegexMonitors only when required

rmatharu:
StreamPartitionCountMOnitor and StreamRegexMonitor use the KafkaSystemAdmin, which has a kafka-consumer. 
Kafka consumer is not thread-safe, therefore explicit synchronization is required for metadata accesses. 

StreamRegexMonitor is not to be enabled when there is no regex input.


mynameborat:
Can you create a JIRA and add explanation around the bug and why the synchronization is needed?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -231,11 +231,13 @@ public String getUpcomingOffset() {
...
 
-              List<PartitionInfo> partitionInfos = metadataConsumer.partitionsFor(streamName);
-              LOG.debug("Stream {} has partitions {}", streamName, partitionInfos);
-
-              partitionInfos.forEach(partitionInfo -> partitionMetadata.put(new Partition(partitionInfo.partition()), dummySspm));
+              synchronized (metadataConsumer) {
+                List<PartitionInfo> partitionInfos = metadataConsumer.partitionsFor(streamName);
+                LOG.debug("Stream {} has partitions {}", streamName, partitionInfos);
 
+                partitionInfos.forEach(
```
prateekm: Can move out of synchronized.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -231,11 +231,13 @@ public String getUpcomingOffset() {
             streamNames.forEach(streamName -> {
               Map<Partition, SystemStreamMetadata.SystemStreamPartitionMetadata> partitionMetadata = new HashMap<>();
 
-              List<PartitionInfo> partitionInfos = metadataConsumer.partitionsFor(streamName);
-              LOG.debug("Stream {} has partitions {}", streamName, partitionInfos);
-
-              partitionInfos.forEach(partitionInfo -> partitionMetadata.put(new Partition(partitionInfo.partition()), dummySspm));
+              synchronized (metadataConsumer) {
```
prateekm: Let's add a comment on the metadataConsumer field that all access needs to be synchronized (and explain why). 

Also, if possible let's make it private.

rmatharu:
addressed your comments.
adminClient accesses seem all to be from runLoop (single thread).

PR #982, opened 2019-04-03
SAMZA-2153: Implement to Config for TableRetryPolicy

Zhangyx39:
Added setTableId and getTableId to TablePart interface.
Implemented toConfig method in TableRetryPolicy to serialize the class to json format so that it can be visible in container log.
Tested with TestRemoteTableDescriptor.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
+++ b/samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
@@ -74,7 +74,6 @@ public class RemoteTableDescriptor<K, V> extends BaseTableDescriptor<K, V, Remot
   public static final String ASYNC_CALLBACK_POOL_SIZE = "io.async.callback.pool.size";
   public static final String READ_RETRY_POLICY = "io.read.retry.policy";
   public static final String WRITE_RETRY_POLICY = "io.write.retry.policy";
-  public static final String READ_AND_WRITE_RETRY_POLICY = "io.read.and.write.retry.policy";
 
   // Input support for a specific remote store (required)
   private TableReadFunction<K, V> readFn;
@@ -114,7 +113,6 @@ public class RemoteTableDescriptor<K, V> extends BaseTableDescriptor<K, V, Remot
   public RemoteTableDescriptor<K, V> withReadFunction(TableReadFunction<K, V> readFn) {
     Preconditions.checkNotNull(readFn, "null read function");
     this.readFn = readFn;
-    this.readFn.setTableId(String.format("%s.%s", tableId, READ_FN));
     return this;
   }
 
@@ -126,7 +124,6 @@ public class RemoteTableDescriptor<K, V> extends BaseTableDescriptor<K, V, Remot
   public RemoteTableDescriptor<K, V> withWriteFunction(TableWriteFunction<K, V> writeFn) {
     Preconditions.checkNotNull(writeFn, "null write function");
     this.writeFn = writeFn;
-    this.writeFn.setTableId(String.format("%s.%s", tableId, WRITE_FN));
     return this;
   }
 
@@ -139,8 +136,6 @@ public class RemoteTableDescriptor<K, V> extends BaseTableDescriptor<K, V, Remot
     Preconditions.checkNotNull(readFn, "null read function");
     Preconditions.checkNotNull(retryPolicy, "null retry policy");
     this.readRetryPolicy = retryPolicy;
-    this.readRetryPolicy.setTableId(String.format("%s.%s", tableId,
-        readRetryPolicy != writeRetryPolicy ? READ_RETRY_POLICY : READ_AND_WRITE_RETRY_POLICY));
     return this;
   }
 
@@ -153,8 +148,6 @@ public class RemoteTableDescriptor<K, V> extends BaseTableDescriptor<K, V, Remot
     Preconditions.checkNotNull(writeFn, "null write function");
     Preconditions.checkNotNull(retryPolicy, "null retry policy");
     this.writeRetryPolicy = retryPolicy;
-    this.writeRetryPolicy.setTableId(String.format("%s.%s", tableId,
-        readRetryPolicy != writeRetryPolicy ? WRITE_RETRY_POLICY : READ_AND_WRITE_RETRY_POLICY));
     return this;
   }
 
@@ -177,12 +170,6 @@ public class RemoteTableDescriptor<K, V> extends BaseTableDescriptor<K, V, Remot
     this.rateLimiter = rateLimiter;
     this.readCreditFn = readCreditFn;
     this.writeCreditFn = writeCreditFn;
-    if (this.readCreditFn != null) {
-      this.readCreditFn.setTableId(String.format("%s.%s", tableId, READ_CREDIT_FN));
-    }
-    if (this.writeCreditFn != null) {
-      this.writeCreditFn.setTableId(String.format("%s.%s", tableId, WRITE_CREDIT_FN));
-    }
     return this;
   }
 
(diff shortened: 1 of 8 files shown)


Zhangyx39:
@weisong44 @dengpanyin

samza-api/src/main/java/org/apache/samza/table/retry/TableRetryPolicy.java
```diff
@@ -256,4 +265,30 @@ public BackoffType getBackoffType() {
...
     return retryPredicate;
   }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Map<String, String> toConfig(Config jobConfig, Config tableConfig) {
+    final Gson gson = new GsonBuilder().excludeFieldsWithModifiers(Modifier.TRANSIENT, Modifier.STATIC).create();
+    return Collections.singletonMap(String.format("%s.%s", getTableId(), this.getClass().getCanonicalName()),
```
weisong44: Can we use key format that is consistent with table configuration items format?

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -135,6 +137,7 @@ public RemoteTableDescriptor(String tableId) {
     Preconditions.checkNotNull(readFn, "null read function");
     Preconditions.checkNotNull(retryPolicy, "null retry policy");
     this.readRetryPolicy = retryPolicy;
+    this.readRetryPolicy.setTableId(String.format("%s.%s", tableId, READ_RETRY_POLICY));
```
weisong44: What happens when someone is reusing the same instance, would tableId be overwritten? Can we add a check to avoid this?

Zhangyx39: Good point. In that case I think we can use a prefix "io.read.and.write.retry.policy" to indicate that this TableRetryPolicy is used for both read and write.
Another solution could be preventing users to use same instance. But that's not convenient for users.

samza-api/src/main/java/org/apache/samza/table/retry/TableRetryPolicy.java
Zhangyx39: We can do "tables.<tableId>.<className>" to keep the similar format and avoid collision.

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
weisong44: I don't think adding prefix will help. Well, I am afraid the later is the the only option.

samza-api/src/main/java/org/apache/samza/table/retry/TableRetryPolicy.java
```diff
@@ -256,4 +266,30 @@ public BackoffType getBackoffType() {
   public RetryPredicate getRetryPredicate() {
     return retryPredicate;
   }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Map<String, String> toConfig(Config jobConfig, Config tableConfig) {
+    final Gson gson = new GsonBuilder().excludeFieldsWithModifiers(Modifier.TRANSIENT, Modifier.STATIC).create();
```
dengpanyin: Is TableRetryPolicy the only one need to override the toConfig function? 
How about read/write function?
Should this implementation be put in somewhere that can be shared?

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -164,12 +172,17 @@ public RemoteTableDescriptor(String tableId) {
...
   public RemoteTableDescriptor<K, V> withRateLimiter(RateLimiter rateLimiter,
-      TableRateLimiter.CreditFunction<K, V> readCreditFn,
-      TableRateLimiter.CreditFunction<K, V> writeCreditFn) {
+      TableRateLimiter.CreditFunction<K, V> readCreditFn, TableRateLimiter.CreditFunction<K, V> writeCreditFn) {
     Preconditions.checkNotNull(rateLimiter, "null read rate limiter");
     this.rateLimiter = rateLimiter;
     this.readCreditFn = readCreditFn;
     this.writeCreditFn = writeCreditFn;
+    if (this.readCreditFn != null) {
+      this.readCreditFn.setTableId(String.format("%s.%s", tableId, READ_CREDIT_FN));
```
dengpanyin: remove this.
The formatted strings are the same,  you can write a method to do the formatting.

samza-api/src/main/java/org/apache/samza/table/remote/TablePart.java
```diff
@@ -44,4 +44,18 @@
...
+   * @param tableId table id
+   */
+  default void setTableId(String tableId) {
+  }
+
+  /**
+   * Return the tableId that has been set. Return empty string by default.
+   */
+  default String getTableId() {
+    return "";
```
dengpanyin: is throwing an exception better than an default value? 
Then the user is not allowed to get the id if it's not set.

samza-api/src/main/java/org/apache/samza/table/retry/TableRetryPolicy.java
```diff
@@ -256,4 +266,30 @@ public BackoffType getBackoffType() {
...
+    return Collections.singletonMap(JavaTableConfig.buildKey(tableId, this.getClass().getCanonicalName()),
```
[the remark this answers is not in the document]
Zhangyx39: Right. After adding `io.read.retry.policy` as part of the tableId, I should remove this `ClassName`.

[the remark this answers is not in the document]
Zhangyx39: But if we don't keep the ClassName, this key will be the same as the key we have been using to (de)serialize the TableRetryPolicy. And thus it will override the existing key. I suggest we use a short ClassName here without the namespace.

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -112,6 +114,7 @@ public RemoteTableDescriptor(String tableId) {
   public RemoteTableDescriptor<K, V> withReadFunction(TableReadFunction<K, V> readFn) {
     Preconditions.checkNotNull(readFn, "null read function");
     this.readFn = readFn;
+    this.readFn.setTableId(String.format("%s.%s", tableId, READ_FN));
```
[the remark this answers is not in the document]
Zhangyx39: This allows the `RemoteTableDescriptor` to set the `tableId` for `TableReadFunction` and `TableWriteFunction` and other `TablePart`s. So that when `TableReadFunction` implements `toConfig` method, it knows it's `tabldId`.
This makes it easier for users since they don't need to type in same `tableId` several times when using `RemoteTableDescritptor`.

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -123,6 +126,7 @@ public RemoteTableDescriptor(String tableId) {
   public RemoteTableDescriptor<K, V> withWriteFunction(TableWriteFunction<K, V> writeFn) {
     Preconditions.checkNotNull(writeFn, "null write function");
     this.writeFn = writeFn;
+    this.writeFn.setTableId(String.format("%s.%s", tableId, WRITE_FN));
```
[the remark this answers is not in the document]
Zhangyx39: same as above

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -135,6 +139,8 @@ public RemoteTableDescriptor(String tableId) {
     Preconditions.checkNotNull(readFn, "null read function");
     Preconditions.checkNotNull(retryPolicy, "null retry policy");
     this.readRetryPolicy = retryPolicy;
+    this.readRetryPolicy.setTableId(String.format("%s.%s", tableId,
```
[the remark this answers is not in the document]
Zhangyx39: Another way could be: when TableRetryPolicy implements toConfig method, it doesn't put the tableId as part of the key. We then append the tableId to the key here in RemoteTableDescriptor`.
This way, a TableRetryFunction can be shared by any number of tables, because there is no tableId associated to it.

Zhangyx39:
Synced with Wei and we think it's better to allow table parts to be reusable. So we don't set a tableId for each table part. Instead, we append the tableId as part of key to the table part configs to distinguish different instances.

samza-api/src/main/java/org/apache/samza/table/retry/TableRetryPolicy.java
Zhangyx39: Maybe we could put it in a class in `package org.apache.samza.util`?

samza-core/src/test/java/org/apache/samza/table/remote/descriptors/TestRemoteTableDescriptor.java
```diff
@@ -148,42 +146,39 @@ public void testSpecifyBothRateAndRateLimiter() {
...
     verify(readFn, times(1)).toConfig(any(), any());
-    Assert.assertEquals("v1", tableConfig.get("k1"));
+    Assert.assertEquals("v1", tableConfig.get("tables.1.io.read.func.k1"));
   }
 
   @Test
   public void testTablePartToConfig() {
 
-    int keys = 0;
+    int key = 0;
```
weisong44: Why this change?

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -164,8 +165,7 @@ public RemoteTableDescriptor(String tableId) {
    * @return this table descriptor instance
    */
   public RemoteTableDescriptor<K, V> withRateLimiter(RateLimiter rateLimiter,
-      TableRateLimiter.CreditFunction<K, V> readCreditFn,
-      TableRateLimiter.CreditFunction<K, V> writeCreditFn) {
+      TableRateLimiter.CreditFunction<K, V> readCreditFn, TableRateLimiter.CreditFunction<K, V> writeCreditFn) {
```
weisong44: Can you make sure you don't change the original formatting, if a line is not changed. There is a few cases in the PR

samza-api/src/main/java/org/apache/samza/table/retry/TableRetryPolicy.java
weisong44: Yeah it's probably not a bad idea.

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -304,8 +304,11 @@ protected void validate() {
    * @param jobConfig job configuration
```
dengpanyin: please update the doc as well.

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -304,8 +304,11 @@ protected void validate() {
...
    * @param tableConfig table configuration
    */
-  protected void addTablePartConfig(TablePart tablePart, Config jobConfig, Map<String, String> tableConfig) {
-    tableConfig.putAll(tablePart.toConfig(jobConfig, new MapConfig(tableConfig)));
+  protected void addTablePartConfig(String tablePartKey, TablePart tablePart, Config jobConfig,
+      Map<String, String> tableConfig) {
+    Map<String, String> tablePartConfig = tablePart.toConfig(jobConfig, new MapConfig(tableConfig));
+    for (String configKey : tablePartConfig.keySet()) {
```
dengpanyin: how about 
tablePart.toConfig(jobConfig, new MapConfig(tableConfig))
.forEach((k, v) ->
addTableConfig(String.format("%s.%s", tablePartKey, k), v,  tableConfig) );

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
Zhangyx39: I see... Seems like my auto format setting is different from the original file... I will change them back.

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
Zhangyx39: Looks better.

samza-core/src/test/java/org/apache/samza/table/remote/descriptors/TestRemoteTableDescriptor.java
Zhangyx39: The configs used to be {k0 -> v0, k1 -> v1, ...}. But after the change, it becomes {tables.1.io.read.func.k0 -> v0, tables.1.io.read.func.k1 -> v1, ...}. And the other of the pairs is unpredictable. So I made them all to be <prefix>.k0 -> v0, without increasing the key each time.

PR #984, opened 2019-04-04
Samza 2116: Making sendTo(table), sendTo(stream) non-terminal

Sanil15:


prateekm:
@bharathkk can you take a look?

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -282,7 +284,7 @@
    * @param <K> the type of key in the table
    * @param <V> the type of record value in the table
    */
-  <K, V> void sendTo(Table<KV<K, V>> table);
+  <K, V> MessageStream<KV<K, V>> sendTo(Table<KV<K, V>> table);
```
prateekm: Document return value.

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -105,6 +105,7 @@ public MessageStreamImpl(StreamApplicationDescriptorImpl streamAppDesc, Operator
   }
 
   @Override
+<<<<<<< HEAD
```
prateekm: Fix merge conflict.

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
```diff
@@ -113,18 +114,22 @@ public MessageStreamImpl(StreamApplicationDescriptorImpl streamAppDesc, Operator
...
 
   @Override
-  public void sink(SinkFunction<? super M> sinkFn) {
+=======
+>>>>>>> b22baf9a6b35e5cb0b190be52f8a93cbc9c086cf
+  public MessageStream<M> sink(SinkFunction<? super M> sinkFn) {
     String opId = this.streamAppDesc.getNextOpId(OpCode.SINK);
     SinkOperatorSpec<M> op = OperatorSpecs.createSinkOperatorSpec(sinkFn, opId);
     this.operatorSpec.registerNextOperatorSpec(op);
+    return new MessageStreamImpl<>(this.streamAppDesc, op);
```
prateekm: Why new message stream instead of returning 'this'?

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -117,8 +117,9 @@
    * Samza SystemProducer implementation.
    *
    * @param sinkFn the function to send messages in this stream to an external system
+   * @return the input {@link MessageStream}
```
prateekm: Minor: '@return this {@link MessageStream}' everywhere.

samza-core/src/main/java/org/apache/samza/operators/impl/SendToTableOperatorImpl.java
```diff
@@ -54,11 +62,10 @@ protected void handleInit(Context context) {
   }
 
   @Override
-  protected CompletionStage<Collection<Void>> handleMessageAsync(KV<K, V> message, MessageCollector collector,
+  protected CompletionStage<Collection<KV<K, V>>> handleMessageAsync(KV<K, V> message, MessageCollector collector,
       TaskCoordinator coordinator) {
     table.put(message.getKey(), message.getValue());
-    // there should be no further chained operators since this is a terminal operator.
-    return CompletableFuture.completedFuture(Collections.emptyList());
+    return CompletableFuture.completedFuture(new ArrayList<KV<K, V>>(Arrays.asList(message)));
```
prateekm: Do you need to wrap Arrays.asList in the ArrayList?

samza-api/src/main/java/org/apache/samza/operators/functions/SinkFunction.java
```diff
@@ -42,6 +42,6 @@
    * @param messageCollector  the {@link MessageCollector} to send the message
    * @param taskCoordinator  the {@link TaskCoordinator} to request commits or shutdown
    */
-  void apply(M message, MessageCollector messageCollector, TaskCoordinator taskCoordinator);
+  M apply(M message, MessageCollector messageCollector, TaskCoordinator taskCoordinator);
```
prateekm: The sink _operator_ needs to return the original MessageStream it was applied on. The function doesn't need to return anything

samza-core/src/main/java/org/apache/samza/operators/impl/OutputOperatorImpl.java
```diff
@@ -66,15 +66,15 @@ protected void handleInit(Context context) {
     }
 
     collector.send(new OutgoingMessageEnvelope(systemStream, null, key, value));
-    return CompletableFuture.completedFuture(Collections.emptyList());
+    return CompletableFuture.completedFuture(new ArrayList<M>(Arrays.asList(message)));
```
sborya: do we need new Array here? You can just use ImmutableList

samza-core/src/main/java/org/apache/samza/operators/impl/SendToTableOperatorImpl.java
sborya: is it sufficient to use ImmutableList

samza-core/src/test/java/org/apache/samza/execution/TestJobGraphJsonGenerator.java
```diff
@@ -204,7 +204,9 @@ public void testRepartitionedJoinStreamApplication() throws Exception {
                 (JoinFunction<Object, KV<Object, Object>, KV<Object, Object>, KV<Object, Object>>) mock(JoinFunction.class),
                 mock(Serde.class), mock(Serde.class), mock(Serde.class), Duration.ofHours(2), "j1")
             .sendTo(outputStream1);
-        messageStream2.sink((message, collector, coordinator) -> { });
+        messageStream2.sink((message, collector, coordinator) -> {
+            return message;
```
sborya: why do you return the message here?

samza-core/src/test/java/org/apache/samza/operators/TestJoinOperator.java
```diff
@@ -350,6 +350,7 @@ private StreamApplicationDescriptorImpl getTestJoinStreamGraph(TestJoinFunction
             .sink((message, messageCollector, taskCoordinator) -> {
                 SystemStream outputSystemStream = new SystemStream("outputSystem", "outputStream");
                 messageCollector.send(new OutgoingMessageEnvelope(outputSystemStream, message));
+                return message;
```
sborya: why do you need to return message?

samza-core/src/test/java/org/apache/samza/operators/TestOperatorSpecGraph.java
```diff
@@ -83,7 +83,9 @@ public void setUp() {
     String streamId2 = "test-input-2";
     InputOperatorSpec testInput2 = new InputOperatorSpec(streamId2, new NoOpSerde(), new NoOpSerde(), null, true, "test-input-4");
     StreamOperatorSpec testMap = OperatorSpecs.createMapOperatorSpec(m -> m, "test-map-5");
-    SinkOperatorSpec testSink = OperatorSpecs.createSinkOperatorSpec((m, mc, tc) -> { }, "test-sink-6");
+    SinkOperatorSpec testSink = OperatorSpecs.createSinkOperatorSpec((m, mc, tc) -> {
+        return m;
```
sborya: why do you need to return message?

samza-core/src/test/java/org/apache/samza/operators/impl/TestWindowOperator.java
```diff
@@ -539,6 +539,7 @@ private StreamApplicationDescriptorImpl getKeyedTumblingWindowStreamGraph(Accumu
           .sink((message, messageCollector, taskCoordinator) -> {
               SystemStream outputSystemStream = new SystemStream("outputSystem", "outputStream");
               messageCollector.send(new OutgoingMessageEnvelope(outputSystemStream, message));
+              return message;
```
sborya: see above

samza-core/src/test/java/org/apache/samza/operators/impl/TestWindowOperator.java
```diff
@@ -557,6 +558,7 @@ private StreamApplicationDescriptorImpl getTumblingWindowStreamGraph(Accumulatio
...
```
sborya: see above?

samza-test/src/test/java/org/apache/samza/test/framework/StreamApplicationIntegrationTest.java
```diff
@@ -149,21 +161,30 @@ public void describe(StreamApplicationDescriptor appDescriptor) {
...
-
       KafkaInputDescriptor<KV<String, TestTableData.PageView>> pageViewISD =
           ksd.getInputDescriptor("PageView", KVSerde.of(new StringSerde(), new JsonSerdeV2<>()));
       KafkaOutputDescriptor<TestTableData.EnrichedPageView> enrichedPageViewOSD =
           ksd.getOutputDescriptor("EnrichedPageView", new JsonSerdeV2<>());
 
+      appDescriptor.getInputStream(profileISD)
+          .map(m -> new KV(m.getValue().getMemberId(), m.getValue()))
+          .sendTo(table)
+          .sink((kv, collector, coordinator) -> {
```
sborya: should you test that sink() returns message stream?

samza-core/src/main/java/org/apache/samza/operators/impl/OutputOperatorImpl.java
mynameborat: `Collections.singleton(message)` seems simpler and concise.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -117,8 +117,9 @@
...
    */
-  void sink(SinkFunction<? super M> sinkFn);
+  MessageStream<M> sink(SinkFunction<? super M> sinkFn);
```
mynameborat: Can we add more documentation around what message is propagated to the downstream operators and its characteristics within the scope of the `SinkFunction`? 

- E.g. It is immutable within the scope of the sink function. 
- When is the message propagated to the downstream? Is it after the completion of SinkFunction? What if the sink function is asynchronous like remote call or kafka send? Do we guarantee any sort of happens before?

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -127,8 +128,9 @@
    * When sending messages to any other {@code OutputStream<M>}, messages are partitioned using a null partition key.
    *
    * @param outputStream the output stream to send messages to
+   * @return the input {@link MessageStream}
    */
-  void sendTo(OutputStream<M> outputStream);
+  MessageStream<M> sendTo(OutputStream<M> outputStream);
```
mynameborat: more documentation please. refer to previous comment on sink api.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
mynameborat: same as above. Are there any ordering guarantees between the table writes and message propagation?

samza-api/src/main/java/org/apache/samza/operators/functions/SinkFunction.java
mynameborat: +1

It introduces confusion from the perspective of how is this even different from MapFunction?
Further, it opens up possibility for applications to return a different message than the input which makes it harder to reason about the semantics of SinkFunction.

samza-core/src/main/java/org/apache/samza/operators/impl/SendToTableOperatorImpl.java
mynameborat: +1. `Collections.singleton(message)`

samza-test/src/test/java/org/apache/samza/test/framework/StreamApplicationIntegrationTest.java
```diff
@@ -50,15 +50,18 @@
...
 import static org.apache.samza.test.controlmessages.TestData.PageView;
 
 public class StreamApplicationIntegrationTest {
+  private static final Logger LOG = LoggerFactory.getLogger(StreamApplicationIntegrationTest.class);
 
   private static final String[] PAGEKEYS = {"inbox", "home", "search", "pymk", "group", "job"};
 
   @Test
-  public void testStatefulJoinWithLocalTable() {
+  public void testStatefulJoinWithLocalTable() throws InterruptedException {
```
mynameborat: Why do you need this method to throw InterruptedException?

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
prateekm: +1, table writes need to happen before the propagation.

samza-api/src/main/java/org/apache/samza/operators/functions/SinkFunction.java
prateekm: @bharathkk sink function is meath as an "escape hatch", in case something isn't currently expressible in the high level API. We should already treat it as a black-box. On second thought, might not be  bad idea to let it return a value, although in that case the new MessageStream should have the returned value, not original. WDYT?

samza-core/src/main/java/org/apache/samza/operators/MessageStreamImpl.java
prateekm: On second thought, returning new MessageStream will be required to get the correct chaining behavior, such that the next operator runs after this one, not 'in parallel'. Is that right?

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
prateekm: Also clarify how this stream is partitioned.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
mynameborat: It is. I wanted to add that to the java documentation so that users don't have to dig through the code to understand the guarantees of the API

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
mynameborat: Seems a little complicated. Not sure what written here refers to `sending to the underlying system or flushing to it`. 

Can we simplify the note along the lines of "no happens before guarantee between the message that is propagated to downstream DAG and the send to the underlying output system". 

WDYT?

samza-core/src/main/java/org/apache/samza/operators/spec/PartitionByOperatorSpec.java
```diff
@@ -30,8 +30,6 @@
  * The spec for an operator that re-partitions a {@link org.apache.samza.operators.MessageStream} to a
  * {@link org.apache.samza.system.SystemStream}. This is usually paired with a corresponding
  * {@link InputOperatorSpec} that consumes the {@link org.apache.samza.system.SystemStream} again.
- * <p>
- * This is a terminal operator and does not allow further operator chaining.
```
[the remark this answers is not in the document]
Sanil15: will do

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -121,14 +121,20 @@
   void sink(SinkFunction<? super M> sinkFn);
 
   /**
-   * Allows sending messages in this {@link MessageStream} to an {@link OutputStream}.
+   * Allows sending messages in this {@link MessageStream} to an {@link OutputStream} and then propogates the this
```
prateekm: Make this consistent for the two methods. Also, typo: "the this".

prateekm: Typo: s/propogate/propagate

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -121,14 +121,20 @@
...
+   * {@link MessageStream} to the next chained operator
    * <p>
    * When sending messages to an {@code OutputStream<KV<K, V>>}, messages are partitioned using their serialized key.
    * When sending messages to any other {@code OutputStream<M>}, messages are partitioned using a null partition key.
-   *
+   * <p>
+   *  Note: There is no guarantee that message is first written to your configured output system and then only it is
```
prateekm: "Note: The message will be written but not flushed to the underlying output system before its propagated to the chained operators."

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -276,13 +282,19 @@
 
   /**
    * Sends messages in this {@link MessageStream} to a {@link Table}. The type of input message is expected
-   * to be {@link KV}, otherwise a {@link ClassCastException} will be thrown.
-   *
+   * to be {@link KV}, otherwise a {@link ClassCastException} will be thrown. After doing a put to table this operator
```
prateekm: Can be more concise, like the method above.

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -276,13 +282,19 @@
...
+   * propagates the message in this {@link MessageStream} to the next chained operator.
+   * <p>
+   *  Note: It is a guarantee that message is first written to table and then only it is propagated to next operator
```
prateekm: What do you mean by "It is a guarantee that message is first written to table". Are you saying that we flush the table before calling next?

"Note: The message will be written but may not be flushed to the underlying table before its propagated to the chained operators. Whether the message can be read back from the Table in the chained operator depends on whether it was flushed and whether the Table offers read after write consistency."

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
prateekm: Do you need to mention that the stream retains the original partitioning?

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
Sanil15: sure

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
Sanil15: Will do

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
Sanil15: What do you mean by "It is a guarantee that message is first written to table". Are you saying that we flush the table before calling next? 
No, just put()

Will do the necessary updates

samza-api/src/main/java/org/apache/samza/operators/MessageStream.java
```diff
@@ -121,14 +121,19 @@
...
   /**
-   * Allows sending messages in this {@link MessageStream} to an {@link OutputStream}.
+   * Allows sending messages in this {@link MessageStream} to an {@link OutputStream} and then propagates this
+   * {@link MessageStream} to the next chained operator
    * <p>
    * When sending messages to an {@code OutputStream<KV<K, V>>}, messages are partitioned using their serialized key.
    * When sending messages to any other {@code OutputStream<M>}, messages are partitioned using a null partition key.
-   *
+   * <p>
+   *  Note: The message will be written but not flushed to the underlying output system before its propagated to the
```
prateekm: Nitpick: Fix indentation (off by 1). Don't need the last <p> (line 132). Same for the method below.

PR #986, opened 2019-04-04
SAMZA-2156: Couchbase support for Samza Table API

Zhangyx39:
Implemented CouchbaseTableReadFunction
Implemented CouchbaseTableWriteFunction
Tested with unit tests and integration test


Zhangyx39:
@weisong44 @dengpanyin I created a new merge request against the real master instead of my forked master.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
```diff
@@ -0,0 +1,277 @@
...
+ * @param <V> Type of values to read from / write to couchbase
+ */
+public abstract class BaseCouchbaseTableFunction<V> implements InitableFunction, ClosableFunction, Serializable {
+
+  // Clients
+  private final static CouchbaseBucketRegistry COUCHBASE_BUCKET_REGISTRY = new CouchbaseBucketRegistry();
+  protected transient Bucket bucket;
+
+  // Function Settings
+  protected Serde<V> valueSerde = null;
```
rmatharu: I think the default serde, shouldn't be null, since a non-null valueSerde is required for working of this class.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
```diff
@@ -0,0 +1,277 @@
...
+    COUCHBASE_BUCKET_REGISTRY.closeBucket(bucketName, clusterNodes);
+  }
+
+  /**
+   * Check whether the exception is caused by one of the temporary failure exceptions, which are
+   * likely to be retriable.
+   * @param exception exception thrown by the table provider
+   * @return a boolean
+   */
+  public boolean isRetriable(Throwable exception) {
```
rmatharu: Seems like a helper function, which could just be a static. 
Also, maybe moved down in the class.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseTableReadFunction.java
```diff
@@ -0,0 +1,152 @@
...
+    documentType = JsonObject.class.isAssignableFrom(valueClass) ? JsonDocument.class : BinaryDocument.class;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void init(Context context) {
+    super.init(context);
+    LOGGER.info(String.format("Read function for bucket %s initialized successfully", bucketName));
```
rmatharu: can be simplified into:
LOGGER.info("Read function for bucket {} initialized successfully", bucketName);

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseTableWriteFunction.java
```diff
@@ -0,0 +1,113 @@
...
+    super(bucketName, clusterNodes, valueClass);
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void init(Context context) {
+    super.init(context);
+    LOGGER.info(String.format("Write function for bucket %s initialized successfully", bucketName));
```
rmatharu: LOGGER.info("Write function for bucket {} initialized successfully", bucketName);

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
```diff
@@ -0,0 +1,277 @@
...
+public abstract class BaseCouchbaseTableFunction<V> implements InitableFunction, ClosableFunction, Serializable {
+
+  // Clients
+  private final static CouchbaseBucketRegistry COUCHBASE_BUCKET_REGISTRY = new CouchbaseBucketRegistry();
+  protected transient Bucket bucket;
+
+  // Function Settings
+  protected Serde<V> valueSerde = null;
+  protected Duration timeout = Duration.ZERO; // default value 0 means no timeout
+  protected Duration ttl = Duration.ZERO; // default value 0 means no ttl, data will be stored forever
```
rmatharu: These defaults are a bit counter-intuitive. 
You can use java's Optional<Duration> on timeout and ttl to mean that these features are enabled, else disabled.

Alternately, would it be possible for a negative value of timeout to mean "no timeout", similarly for ttl.
Duration does allow negative.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
Zhangyx39: Thank you Raymond for review.
There is one case when the valueType is JsonObject.class, we do not require a valueSerde. The (de)serialization is handled by Couchbase client in that case. That's why we don't enforce valueSerde to be not null.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
Zhangyx39: This method is inherited by the child classes and is required by TableReadFunction and TableWriteFunction. Although it's not required by any interface in BaseCouchbaseTableFunction class, we cannot change its signature.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseTableReadFunction.java
Zhangyx39: Sounds good.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
Zhangyx39: These defaults are consistent with Couchbase client. Actually, even if I don't set ttl, it will still be 0 in a Couchbase document. So I use those default values to be consistent with Couchbase client API. Otherwise, we need to add extra logic to check those values in read/write function, making it a little more complicated. How do you think?

rmatharu: I see, in that case, if you can please add the 0-special-case behavior to the javadoc for methods withTimeout() and withTtl().

samza-kv-remote/src/test/java/org/apache/samza/table/remote/couchbase/TestCouchbaseBucketRegistry.java
```diff
@@ -0,0 +1,132 @@
...
+import org.junit.runner.RunWith;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+import static org.junit.Assert.*;
+import static org.mockito.Matchers.*;
+import static org.powermock.api.mockito.PowerMockito.*;
+
+
+@RunWith(PowerMockRunner.class)
```
rmatharu: If you can add a brief description of the class here, similar to how you've added for TestCouchbaseRemoteTableEndToEnd. 
It will greatly help for future development, when a change in code breaks this test, and requires someone to debug the test breakage.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
Zhangyx39: No problem.

samza-kv-remote/src/test/java/org/apache/samza/table/remote/couchbase/TestCouchbaseBucketRegistry.java
Zhangyx39: Good point.

Zhangyx39:
> Minor comments, feel free to fix and checkin.

Thanks for the review @rmatharu 
But I don't have permission to checkin the code. Can you help me to do that?

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
```diff
@@ -0,0 +1,277 @@
...
+
+  /**
+   * Constructor for BaseCouchbaseTableFunction. This constructor abstracts the shareable logic of the read and write
+   * functions. It is not intended to be called directly.
+   * @param bucketName Name of the Couchbase bucket
+   * @param clusterNodes Some Hosts of the Couchbase cluster. Recommended to provide more than one nodes so that if
+   *                     the first node could not be connected, other nodes can be tried.
+   * @param valueClass type of values
+   */
+  public BaseCouchbaseTableFunction(String bucketName, List<String> clusterNodes, Class<V> valueClass) {
```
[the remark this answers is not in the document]
Zhangyx39: Yep. This could be nicer.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
```diff
@@ -0,0 +1,277 @@
...
+   * @param password password
+   * @param <T> type of this instance
+   * @return Self
+   */
+  public <T extends BaseCouchbaseTableFunction<V>> T withUsernameAndPassword(String username, String password) {
+    if (environmentConfigs.sslEnabled != null && environmentConfigs.sslEnabled) {
+      throw new IllegalArgumentException(
+          "Role-Based Access Control and Certificate-Based Authentication cannot be used together.");
+    }
+    environmentConfigs.username = username;
```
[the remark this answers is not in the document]
Zhangyx39: I considered about it twice. We can add check for username. But we shouldn't enforce the password to be not empty because there could be some buckets migrated from pre 5.0 Couchbase versions that has no password. In that case, simple an username without password should be able to access the bucket.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
```diff
@@ -0,0 +1,277 @@
...
+
+  /**
+   * Enable certificate-based authentication. If ssl is enabled sslKeystore or sslTrustStore should also be provided
+   * accordingly.
+   * @param sslEnabled allows to enable certificate-based authentication
+   * @param certAuthEnabled allows to enable X.509 client certificate authentication
+   * @param <T> type of this instance
+   * @return Self
+   */
+  public <T extends BaseCouchbaseTableFunction<V>> T withSslEnabledAndCertAuthEnabled(boolean sslEnabled,
```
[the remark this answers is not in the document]
Zhangyx39: There are four combinations here:

- Both true, enable certificate-based authentication
- Both false, disable
- sslEnabled: true, certAuthEnabled: false, only server side certificate checking is needed
- sslEnabled: false, certAuthEnabled: true, not allowed, which is checked in line 176

build.gradle
```diff
@@ -697,6 +698,34 @@ project(":samza-kv-rocksdb_$scalaSuffix") {
   }
 }
 
+project(":samza-kv-remote_$scalaSuffix") {
```
prateekm: Should this be samza-kv-couchbase?

build.gradle
```diff
@@ -188,6 +188,7 @@ project(":samza-core_$scalaSuffix") {
     compile "org.scala-lang:scala-library:$scalaVersion"
     compile "org.slf4j:slf4j-api:$slf4jVersion"
     compile "net.jodah:failsafe:$failsafeVersion"
+    compile "org.scala-lang:scala-library:$scalaVersion"
```
prateekm: Why is this required?

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseBucketRegistry.java
```diff
@@ -0,0 +1,173 @@
...
+    if (configs.bootstrapHttpSslPort != null) {
+      envBuilder.bootstrapHttpSslPort(configs.bootstrapHttpSslPort);
+    }
+    CouchbaseEnvironment env = envBuilder.build();
+    Cluster cluster = CouchbaseCluster.create(env, clusterNodes);
+    if (configs.sslEnabled != null && configs.sslEnabled) {
+      cluster.authenticate(CertAuthenticator.INSTANCE);
+    } else if (configs.username != null && configs.password != null) {
+      cluster.authenticate(configs.username, configs.password);
+    }
```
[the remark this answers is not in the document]
Zhangyx39: I think we can log a warning here since in some situations we don't need authentication. E.g. mocked couchbase cluster.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseTableReadFunction.java
```diff
@@ -0,0 +1,152 @@
...
+   * {@inheritDoc}
+   */
+  @Override
+  public void init(Context context) {
+    super.init(context);
+    LOGGER.info("Read function for bucket {} initialized successfully", bucketName);
+  }
+
+  /**
+   * {@inheritDoc}
```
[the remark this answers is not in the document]
Zhangyx39: I see. Then I think we can remove these.

build.gradle
Zhangyx39: We are thinking to use samza-kv-remote as the package for all remote table providers.

build.gradle
Zhangyx39: This could be a mistake when I rebased the branch with latest master.

build.gradle
prateekm: Why? Is there something shared across all providers? If so, does it makes sense to separate them into two packages - samza-kv-remote and samza-kv-couchbase? If we add another remote table provider in the future, it doesn't make sense to require users to pull in couchbase dependencies as well.

Zhangyx39: What you said made sense. What I thought was that keeping the remote table providers together makes it easy for users to find them. But I don't have strong opinion on this since I am not familiar with how Samza organizes modules. @weisong44 How do you think about it?

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseBucketRegistry.java
```diff
@@ -0,0 +1,198 @@
...
+  /**
+   * A synchronized method to open a bucket given the bucket name and cluster nodes. Once a bucket has been opened,
+   * the reference of the bucket will be stored in the registry, and the same reference will be returned given same
+   * bucket name. Each time this method is called, the counter of usage of the bucket is increased by one.
+   * @param bucketName name of the bucket to be opened
+   * @param clusterNodes list of cluster nodes
+   * @param configs couchbase environment configs
+   * @return A Bucket instance associated with the bucket name and cluster nodes
+   */
+  public synchronized Bucket getBucket(String bucketName, List<String> clusterNodes,
```
mynameborat: Do you need synchronized here? Would concurrent map work instead?

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseBucketRegistry.java
```diff
@@ -0,0 +1,198 @@
...
+
+
+/**
+ * The CouchbaseBucketRegistry is intended to reuse the same {@link Cluster} instance given same clusterNodes and reuse
+ * the same {@link Bucket} given same bucketName. Instantiating a Bucket is expensive. Different tasks within a
+ * container that is using the same bucket should use this registry to avoid creating multiple Buckets.
+ */
+public class CouchbaseBucketRegistry {
+  private static final Logger LOGGER = LoggerFactory.getLogger(CouchbaseBucketRegistry.class);
+  private Map<String, Bucket> openedBuckets;
```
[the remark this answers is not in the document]
Zhangyx39: Sure. That's more appropriate.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseTableWriteFunction.java
```diff
@@ -0,0 +1,103 @@
...
+  @Override
+  public void init(Context context) {
+    super.init(context);
+    LOGGER.info("Write function for bucket {} initialized successfully", bucketName);
+  }
+
+  @Override
+  public CompletableFuture<Void> putAsync(String key, V record) {
+    Preconditions.checkNotNull(key);
+    Preconditions.checkArgument(!key.contains(" "), "key must not contain spaces.");
```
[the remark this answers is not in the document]
Zhangyx39: Right. Empty string is also not valid. Should check this.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseBucketRegistry.java
Zhangyx39: We do need synchronized here. The purpose is that we not only want to keep the map correct, but we also want to prevent creating multiple Bucket instances in different thread. Think about this:
1. Thread 1 and thread 2 both check that the "bucket-id" key does not exist in the map.
2. Then they both try to create a new bucket instance and put to the map
3. Although the map is thread safe and only one Bucket instance can be put to the map with key "bucket-id", two instances of Bucket have been created.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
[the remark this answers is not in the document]
weisong44: Good, so if someone calls this method, sslEnabled must be true. So it doesn't need to be a parameter, correct?

build.gradle
weisong44: Yeah makes sense, let's make the change.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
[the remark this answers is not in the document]
Zhangyx39: Right. Let's remove it.

samza-kv-remote/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseEnvironmentConfigs.java
```diff
@@ -0,0 +1,138 @@
...
+  protected String sslTruststoreFile;
+  protected String sslTruststorePassword;
+  protected Integer bootstrapCarrierDirectPort;
+  protected Integer bootstrapCarrierSslPort;
+  protected Integer bootstrapHttpDirectPort;
+  protected Integer bootstrapHttpSslPort;
+  protected String username;
+  protected String password;
+
+  public Boolean getSslEnabled() {
```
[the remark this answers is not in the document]
weisong44: I didn't mean to make it an inner class in CouchbaseEnvironmentConfigs, it makes it odd. Can we just keep it a separate class, just not use public in class declaration?

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
```diff
@@ -0,0 +1,277 @@
...
+   *                     the first node could not be connected, other nodes can be tried.
+   */
+  public BaseCouchbaseTableFunction(String bucketName, Class<V> valueClass, String... clusterNodes) {
+    Preconditions.checkArgument(StringUtils.isNotEmpty(bucketName), "Bucket name is not allowed to be null or empty.");
+    Preconditions.checkArgument(valueClass != null, "Value class is not allowed to be null.");
+    Preconditions.checkArgument(ArrayUtils.isNotEmpty(clusterNodes),
+        "Cluster nodes is not allowed to be null or empty.");
+    this.bucketName = bucketName;
+    this.clusterNodes = ImmutableList.copyOf(clusterNodes);
+    this.environmentConfigs = new CouchbaseEnvironmentConfigs();
```
dengpanyin: only use "this" when necessary

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseBucketRegistry.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+    return bucketClosed && clusterClosed;
+  }
+
+  /**
+   * Helper method to open a cluster given cluster nodes and environment configurations.
+   */
+  private Cluster openCluster(List<String> clusterNodes, CouchbaseEnvironmentConfigs configs) {
+    DefaultCouchbaseEnvironment.Builder envBuilder = new DefaultCouchbaseEnvironment.Builder();
+    if (configs.sslEnabled != null) {
```
dengpanyin: do you need all the null check?

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseBucketRegistry.java
```diff
@@ -0,0 +1,181 @@
...
+
+
+/**
+ * The CouchbaseBucketRegistry is intended to reuse the same {@link Cluster} instance given same clusterNodes and reuse
+ * the same {@link Bucket} given same bucketName. Instantiating a Bucket is expensive. Different tasks within a
+ * container that is using the same bucket should use this registry to avoid creating multiple Buckets.
+ */
+public class CouchbaseBucketRegistry {
+  private static final Logger LOGGER = LoggerFactory.getLogger(CouchbaseBucketRegistry.class);
+  private final Map<String, Bucket> openedBuckets;
```
dengpanyin: probably a ConcurrentHashMap can satisfy the thread-safe-requirement instead of using "synchronized" keyword to the method.

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseEnvironmentConfigs.java
```diff
@@ -0,0 +1,39 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.table.remote.couchbase;
+
+import java.io.Serializable;
+
+
+class CouchbaseEnvironmentConfigs implements Serializable {
```
dengpanyin: Why creating this class other than use DefaultCouchbaseEnvironment.Builder  directly?

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseTableReadFunction.java
```diff
@@ -0,0 +1,146 @@
...
+  @Override
+  public CompletableFuture<V> getAsync(String key) {
+    Preconditions.checkArgument(StringUtils.isNotBlank(key), "key must not be null, empty or blank");
+    CompletableFuture<V> future = new CompletableFuture<>();
+    Single<? extends Document<?>> singleObservable =
+        bucket.async().get(key, documentType, timeout.toMillis(), TimeUnit.MILLISECONDS).toSingle();
+    singleObservable.subscribe(new SingleSubscriber<Document<?>>() {
+      @Override
+      public void onSuccess(Document<?> document) {
+        if (document != null) {
```
dengpanyin: when will the document be null?

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
```diff
@@ -0,0 +1,277 @@
...
+   * X.509 client cert authentication you might as well just use {@link #withSslTruststoreFileAndPassword}
+   * alone.
+   * @param sslKeystoreFile  path of ssl keystore file
+   * @param sslKeystorePassword password of ssl keystore
+   * @param <T> type of this instance
+   * @return Self
+   */
+  public <T extends BaseCouchbaseTableFunction<V>> T withSslKeystoreFileAndPassword(String sslKeystoreFile,
+      String sslKeystorePassword) {
+    Preconditions.checkArgument(StringUtils.isNotEmpty(sslKeystoreFile), "Null or empty sslKeystoreFile");
```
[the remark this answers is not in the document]
Zhangyx39: The Couchbase client is using `sslKeystoreFile` so I want to keep the name the same. I guess they think keystore is a word?

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
Zhangyx39: ok

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseBucketRegistry.java
Zhangyx39: I think so because we don't have default values for the environment configs. We can't provide default values because it depends on use cases. So it's better to let user set it when they need to.

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseEnvironmentConfigs.java
Zhangyx39: Good question. I really wish to use DefaultCouchbaseEnvironment.Builder. But the problem is that class is not serializable. So we have to create this serializable CouchbaseEnvironmentConfigs to keep the configs.

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseTableReadFunction.java
Zhangyx39: When the key doesn't exist in Couchbase. Instead of returning a document with null value, it will return a null.

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseBucketRegistry.java
Zhangyx39: We not only want to keep it thread safe here, we also want to prevent multiple calls of `cluster.openBucket()`. See my previous reply:
https://github.com/apache/samza/pull/986#discussion_r276424654

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/BaseCouchbaseTableFunction.java
```diff
@@ -0,0 +1,277 @@
...
+  protected Serde<V> valueSerde = null;
+  protected Duration timeout = Duration.ZERO; // default value 0 means no timeout
+  protected Duration ttl = Duration.ZERO; // default value 0 means no ttl, data will be stored forever
+
+  // Cluster Settings
+  protected final List<String> clusterNodes;
+  protected final String bucketName;
+
+  // Environment Settings
+  protected CouchbaseEnvironmentConfigs environmentConfigs;
```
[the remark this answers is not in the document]
Zhangyx39: sure

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseTableReadFunction.java
Zhangyx39: I see. The if it returns a null, it will go to the onError method. In onSuccess method, a document should not be null.

samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseBucketRegistry.java
Zhangyx39: I see. We can use ConcurrentHashMap.computeIfAbsent can do the job. But it makes it harder to count the usage of a bucket. Since this method will only be called once per task per table, it's ok to use a synchronized method.

PR #989, opened 2019-04-09
Added an AM metric for configured container cpu and memory.

prateekm:


prateekm:
@rmatharu can you take a look.

samza-core/src/main/scala/org/apache/samza/metrics/ContainerProcessManagerMetrics.scala
```diff
@@ -73,6 +75,9 @@ class ContainerProcessManagerMetrics(
      val mFailoversToAnyHost = newGauge("failovers-to-any-host", () => state.failoversToAnyHost.get())
      val mFailoversToStandby = newGauge("failovers-to-standby", () => state.failoversToStandby.get())
 
+     val mContainerMemoryMb = newGauge("container-memory-mb", () => clusterManagerConfig.getContainerMemoryMb)
+     val mContainerCpuCores = newGauge("container-cpu-cores", () => clusterManagerConfig.getNumCores)
+
```
rmatharu: Making a note: We may want to in future consolidate this information, so that its only emitted from either the Containers themselves (in case of heterogenous containers), or from the AM (including other container details such as XmX size).

PR #991, opened 2019-04-10
[SAMZA-2162] - Update Table APIs to be consistent with High level Async API

mynameborat:
@weisong44 please take a look


samza-api/src/main/java/org/apache/samza/table/AsyncReadWriteTable.java
```diff
@@ -39,7 +39,7 @@
    * @return completableFuture for the requested value
```
rmatharu: Return CompletionStage for the requested values

samza-api/src/main/java/org/apache/samza/table/AsyncReadWriteTable.java
```diff
@@ -39,7 +39,7 @@
...
    * @throws NullPointerException if the specified {@code key} is {@code null}.
    */
-  CompletableFuture<V> getAsync(K key);
+  CompletionStage<V> getAsync(K key);
```
dengpanyin: why make this change? It may break LI or the external users as well.

samza-core/src/main/java/org/apache/samza/operators/impl/SendToTableOperatorImpl.java
```diff
@@ -56,9 +55,9 @@ protected void handleInit(Context context) {
   @Override
   protected CompletionStage<Collection<Void>> handleMessageAsync(KV<K, V> message, MessageCollector collector,
       TaskCoordinator coordinator) {
-    table.put(message.getKey(), message.getValue());
     // there should be no further chained operators since this is a terminal operator.
-    return CompletableFuture.completedFuture(Collections.emptyList());
+    return table.putAsync(message.getKey(), message.getValue())
```
dengpanyin: with this change, the put order is not guaranteed any more.
Let's say a sequence of put event:
1. put(1, 10)
2. put (2, 20)
3. put(1, 5)
The 1st put fails at the 1st attempt and retry triggered.
2nd put succeed
3rd put succeed
then the retry for the 1st put succeeds.

Now the value for key 1 in the store is 10, while 5 is expected.

samza-api/src/main/java/org/apache/samza/table/remote/TableWriteFunction.java
```diff
@@ -102,43 +95,37 @@ default void putAll(List<Entry<K, V>> records) {
...
    * The default implementation calls deleteAllAsync and blocks on the completion afterwards.
    * @param keys keys for the table records to be written
    */
   default void deleteAll(Collection<K> keys) {
-    try {
-      deleteAllAsync(keys).get();
-    } catch (InterruptedException | ExecutionException e) {
-      throw new SamzaException("DELETE failed for " + keys, e);
-    }
+    deleteAllAsync(keys).toCompletableFuture().join();
```
dengpanyin: remote the exception may break some apps which is relying on the exception type.

weisong44: +1
Should catch and re-throw

samza-api/src/main/java/org/apache/samza/table/AsyncReadWriteTable.java
shanthoosh: +1.

This would affect for not just high-level API users, but also for any samza user who is using the `Table` API through `TaskContext` through low-level API.  

Would be better not to break backwards-compatibility here just for unification.

samza-core/src/main/java/org/apache/samza/operators/impl/SendToTableOperatorImpl.java
mynameborat: Synced up offline about this. This doesn't change semantics in any ways than what it is today. 
The scenario outlined is only possible when task.max.concurrency is set to > 1. 
In case of task.max.concurrency, we expect the application to coordinate its shared states since there is no happens-before guarantee between messages delivered to the DAG.

samza-api/src/main/java/org/apache/samza/table/AsyncReadWriteTable.java
weisong44: As we don't have a lot of customer yet, let's make table API consistent with the rest of the low/high level API.

prateekm:
@mynameborat Is this PR still relevant?

mynameborat:
@prateekm  let me fix the merge conflicts and get this merged.

PR #992, opened 2019-04-11
SAMZA-2163: Add try-catch to get the version in MetricsSnapshotReporterFactory

xinyuiscool:
Currently we relying on app.class or task.class to get the version. For beam the app.class is anonymous class which caused exception when loading the class of it. This patch will do a try catch for this exception and use 0.0.1 as default.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/metrics/reporter/MetricsSnapshotReporterFactory.scala
+++ b/samza-core/src/main/scala/org/apache/samza/metrics/reporter/MetricsSnapshotReporterFactory.scala
@@ -45,14 +45,17 @@ class MetricsSnapshotReporterFactory extends MetricsReporterFactory with Logging
     val jobId = config
       .getJobId
 
-    val taskClass = Option(new ApplicationConfig(config).getAppClass())
-      .getOrElse(config.getTaskClass.getOrElse(throw new SamzaException("No task or app class defined for config.")))
-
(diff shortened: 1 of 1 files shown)


xinyuiscool:
@lhaiesp : please take a look!

samza-core/src/main/scala/org/apache/samza/metrics/reporter/MetricsSnapshotReporterFactory.scala
```diff
@@ -45,14 +45,17 @@ class MetricsSnapshotReporterFactory extends MetricsReporterFactory with Logging
...
-        "0.0.1"
-      })
+    val version =
+      try {
+        val taskClass = Option(new ApplicationConfig(config).getAppClass())
+          .orElse(config.getTaskClass).get
+        Option(Class.forName(taskClass).getPackage.getImplementationVersion).get
+      } catch {
+        case e: Exception => {
+          warn("Unable to find implementation version in jar's meta info. Defaulting to 0.0.1.")
```
lhaiesp: Should we log the exception? And we should document why we need to catch exception here (anonymous class..), so that we don't lose context in the codes.

xinyuiscool: Seems the warning line should be good here, since mostly likely it will be NoSuchElementException if no class is found or no version is found. Currently we are mostly getting 0.0.1 right now.

PR #993, opened 2019-04-12
SAMZA-2164: Close the consumer after reading checkpoints

xinyuiscool:
We read checkpoints at the start of a container for once. We should close the consumer after that so it won't keep filling the BEM with checkpoints written later. Currently I saw a lot of these checkpoint messages in the BEM.

### DIFF
--- a/samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
+++ b/samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
@@ -106,6 +106,7 @@ class KafkaCheckpointManager(checkpointSpec: KafkaStreamSpec,
     info(s"Starting the checkpoint SystemConsumer from oldest offset $oldestOffset")
     systemConsumer.register(checkpointSsp, oldestOffset)
     systemConsumer.start()
+    // the consumer will be closed after first time reading the checkpoint
   }
 
   /**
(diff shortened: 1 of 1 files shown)


samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -128,11 +128,11 @@ class KafkaCheckpointManager(checkpointSpec: KafkaStreamSpec,
     info(s"Reading checkpoint for taskName $taskName")
 
     if (taskNamesToCheckpoints == null) {
-      debug("Reading checkpoints for the first time")
+      info("Reading checkpoints for the first time")
       taskNamesToCheckpoints = readCheckpoints()
-    } else {
-      debug("Updating existing checkpoint mappings")
-      taskNamesToCheckpoints ++= readCheckpoints()
+      // Stop the system consumer since we only need to read checkpoints once
```
prateekm: Don't need the else anymore?

If not, does it make sense to move the register and start here as well? Maybe even make it local to the method instead of a field? Is it  used for anything else?

prateekm: 'it' == systemConsumer. That way we won't break anything if something does rely on reading multiple times, and will still get the perf improvement for the happy path where we don't.

xinyuiscool: No, i don't think that update is needed at all. It's weird after the first pulling we are still trying to update. It doesn't make it better if there is any problem in the first read (some task also has wrong checkpoints).

I feel it might be useful to start a little bit early to just give a bit more time to let the proxy thread running. It's not used anywhere else.

prateekm: Them maybe leave a comment near the start that this is stopped after first checkpoint read. Otherwise LGTM, nice find.

xinyuiscool: cool! Let me put a comment there.

PR #995, opened 2019-04-15
SAMZA-2165: Account for coordinator restarts in calls to status

vjagadish1989:
Currently status of a Samza job is determined by a combination of:
1. Obtain YARN's status for the job by querying the RM
2. Obtain the AM/coordinator URL for the job
3. If (1) is "Running", Query the coordinator URL if all containers have started

YARN may restart the coordinator between (2) and (3) and the old coordinator process may no longer be alive, triggering a ConnectException in (3). This causes the status-call to fail;

A better alternative to handle these retriable errors is to return a "New" status from the API - so that applications can continue polling for status.


vjagadish1989:
@prateekm for review

samza-yarn/src/main/scala/org/apache/samza/job/yarn/ClientHelper.scala
```diff
@@ -350,6 +350,11 @@ class ClientHelper(conf: Configuration) extends Logging {
       } else {
         false
       }
+    } catch {
+      case e: IOException => {
```
prateekm: Minor: Would prefer to not guess the exception cause in the exception message. Maybe instead add a comment explaining why returning false here is better.

What was the previous behavior? Does the caller catch the exception?

samza-yarn/src/main/scala/org/apache/samza/job/yarn/ClientHelper.scala
```diff
@@ -350,6 +350,11 @@ class ClientHelper(conf: Configuration) extends Logging {
...
+        warn("Exception when querying AM metrics. Likely because YARN was re-starting a new AM attempt", e)
```
rmatharu: an IOException in querying AM could be because of a few other reasons as well, right?

vjagadish1989: > an IOException in querying AM could be because of a few other reasons as well, right?

I've modified the log line.

> maybe we want to catch any exception?

 probably not - for eg: parse errors due to an invalid format in the response should be bubbled-up

PR #996, opened 2019-04-16
 SAMZA-2166: [Scala cleanup] Convert StorageConfig.scala to Java 

cameronlee314:
It might be easier to view this PR by looking at individual commits. The commit a34e320 contains the change to just remove `StorageConfig.scala` and replace with `JavaStorageConfig.java`. The next commit 501ae36 does a rename from `JavaStorageConfig.java` to `StorageConfig.java`, since `StorageConfig.scala` was removed.

Note: This also seems to fix a bug in `BaseKeyValueStorageEngineFactory` which involved reading the access log sample ratio config. Previously, the config being passed in to `BaseKeyValueStorageEngineFactory` was stripped of the `stores.<store name>.` prefix, but it was implicitly converted into a `StorageConfig`, so when `getAccessLogSamplingRatio` was called, it seems like it would never actually find the full `stores.<store name>.accesslog.sampling.ratio`. This meant it always fell back to the default value.


samza-kv/src/main/scala/org/apache/samza/storage/kv/AccessLoggedStore.scala
```diff
@@ -21,6 +21,7 @@ package org.apache.samza.storage.kv
 
 
 import java.util
+
```
shanthoosh: Remove the empty line.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -23,41 +23,48 @@
...
-  private static final String ACCESSLOG_ENABLED = "stores.%s.accesslog.enabled";
-  private static final int DEFAULT_ACCESSLOG_SAMPLING_RATIO = 50;
-
-  public static final String SIDE_INPUTS = "stores.%s.side.inputs";
-  public static final String SIDE_INPUTS_PROCESSOR_FACTORY = "stores.%s.side.inputs.processor.factory";
-  public static final String SIDE_INPUTS_PROCESSOR_SERIALIZED_INSTANCE = "stores.%s.side.inputs.processor.serialized.instance";
-
-  public JavaStorageConfig(Config config) {
+
+  public static final String FACTORY = "stores.%s" + FACTORY_SUFFIX;
```
shanthoosh: Would be better to use the STORE_PREFIX literal in all of the following constants, rather than hardcoding it everywhere.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -23,41 +23,48 @@
...
+  public static final String KEY_SERDE = "stores.%s.key.serde";
+  public static final String MSG_SERDE = "stores.%s.msg.serde";
+  public static final String CHANGELOG_STREAM = "stores.%s" + CHANGELOG_SUFFIX;
+  static final String CHANGELOG_SYSTEM = "job.changelog.system";
+  static final String CHANGELOG_DELETE_RETENTION_MS = "stores.%s.changelog.delete.retention.ms";
+  static final long DEFAULT_CHANGELOG_DELETE_RETENTION_MS = TimeUnit.DAYS.toMillis(1);
+  public static final String ACCESSLOG_STREAM_SUFFIX = "access-log";
+  static final String ACCESSLOG_SAMPLING_RATIO = "stores.%s.accesslog.sampling.ratio";
```
shanthoosh: There doesn't seem to be dependency between the static literals in this class. Should group the public, private and package-protected constants together. Now they're scattered across the class.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -89,7 +98,7 @@ public String getChangelogStream(String storeName) {
     if (systemStreamRes != null) {
       systemStreamRes = StreamManager.createUniqueNameForBatch(systemStreamRes, this);
```
shanthoosh: Not related to your change. But `getChangelogStream` is not a pure value-accessor contrary to its name. Would be better to move the creation outside of this get method.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -160,8 +169,8 @@ public String getChangelogSystem() {
    * @param storeName name of the store
    * @return the class name of SideInputsProcessorFactory if present, null otherwise
```
shanthoosh: Would be better to update the java doc to indicate that this method returns Optional.empty if the side-input factory is not defined.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -170,7 +179,30 @@ public String getSideInputsProcessorFactory(String storeName) {
    * @param storeName name of the store
    * @return the serialized instance of SideInputsProcessor if present, null otherwise
```
shanthoosh: Would be better to update the java doc to indicate that this method returns Optional.empty if the side-input factory is not defined.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -131,8 +140,8 @@ public String getStorageMsgSerde(String storeName) {
    *
    * @return the name of the system to use by default for all changelogs, if defined.
    */
-  public String getChangelogSystem() {
```
shanthoosh: Fix the following first line of `getChangelogSystem` javadoc:
`Gets the System to use for reading/writing checkpoints`

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -58,6 +57,7 @@ import org.apache.samza.util.ScalaJavaUtil.JavaOptionals
 import org.apache.samza.util.{Util, _}
 import org.apache.samza.{SamzaContainerStatus, SamzaException}
 
+import scala.collection.JavaConverters
```
shanthoosh: Doesn't line 60, 61 doing the same import. Is both of them necessary?

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
cameronlee314: Are you concerned about the call to a method called "create"? `createUniqueNameForBatch` just looks up some configs; it doesn't actually have a side effect.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
cameronlee314: Good catch

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
cameronlee314: Good catch. I think the `JavaConverters` one without the `_` is only necessary if you directly use `JavaConverters` as a static qualifier, but that's not done in this class, so it can be removed.

PR #1001, opened 2019-04-18
SAMZA-2168: Remove redundant SystemAdmin creation in ApplicationMaster

shanthoosh:
Samza ApplicationMaster is a process responsible for scheduling, orchestrating and managing the lifecycle of containers of a samza job.

SystemAdmin abstraction is used in samza to validate, create and fetch the metadata of input and metadata streams of the samza job. Creating a SystemAdmin instance for a system is an expensive operation which entails creating the connection with the broker and setting up resources.

Currently, SystemAdmin for a system is created multiple times across the different components of the samza ApplicationMaster. This duplicate SystemAdmin creation happens as a part of the startup sequence of ApplicationMaster and unnecessarily increases the startup time of ApplicationMaster.


shanthoosh:
@xinyuiscool

shanthoosh:
@xinyuiscool 
1. I tested this patch with the hello-samza  test jobs in open-source. 
2. Verified that this patch works fine with samza-yarn  and a beam job in LinkedIn. 

For the large stateful jobs which consumes a lot of input topics from different systems, I verified that this patch reduces the startup time of ApplicationMaster by 1.5 minutes.

samza-core/src/main/scala/org/apache/samza/job/local/ProcessJobFactory.scala
```diff
@@ -52,54 +53,62 @@ class ProcessJobFactory extends StreamJobFactory with Logging {
 
     val configFromCoordinatorStream: Config = CoordinatorStreamUtil.readConfigFromCoordinatorStream(coordinatorStreamStore)
 
-    val changelogStreamManager = new ChangelogStreamManager(new NamespaceAwareCoordinatorStreamStore(coordinatorStreamStore, SetChangelogMapping.TYPE))
+    val systemAdmins = new SystemAdmins(configFromCoordinatorStream)
 
-    val coordinator = JobModelManager(configFromCoordinatorStream, changelogStreamManager.readPartitionMapping(), coordinatorStreamStore, metricsRegistry)
-    val jobModel = coordinator.jobModel
+    try {
+      systemAdmins.start()
```
cameronlee314: Should this move out of the try-catch? It looks like the pattern you are using for lifecycle management of `systemAdmins` is a bit inconsistent across classes.
In some classes, you put the `start` inside the try-finally which has the `stop`, but in other classes you have the `start` outside. Could you please try to make it consistent?

samza-test/src/test/scala/org/apache/samza/test/integration/StreamTaskTestUtil.scala
```diff
@@ -283,7 +283,9 @@ class StreamTaskTestUtil {
       case _ => assert(checkpointManager != null, "No checkpoint manager factory configured")
     }
 
-    ChangelogStreamManager.createChangelogStreams(jobModel.getConfig, jobModel.maxChangeLogStreamPartitions)
+    val systemAdmins = new SystemAdmins(mapConfig)
+    systemAdmins.start()
```
cameronlee314: Need to call `stop`?

samza-core/src/main/scala/org/apache/samza/job/local/ProcessJobFactory.scala
```diff
@@ -52,54 +53,62 @@ class ProcessJobFactory extends StreamJobFactory with Logging {
...
-        }
-        case _ => {
-          info("Defaulting to ShellCommandBuilder")
-          new ShellCommandBuilder
+      //create necessary checkpoint and changelog streams
+      val checkpointManager = new TaskConfigJava(jobModel.getConfig).getCheckpointManager(metricsRegistry)
+      if (checkpointManager != null) {
+        checkpointManager.createResources()
+      }
+      ChangelogStreamManager.createChangelogStreams(jobModel.getConfig, jobModel.maxChangeLogStreamPartitions, systemAdmins)
```
rmatharu: We don't seem to have a test for this method and seems to be called from quite a few critical places.

samza-core/src/main/scala/org/apache/samza/job/local/ProcessJobFactory.scala
shanthoosh: Done.

samza-test/src/test/scala/org/apache/samza/test/integration/StreamTaskTestUtil.scala
shanthoosh: Done.

samza-core/src/main/scala/org/apache/samza/job/local/ProcessJobFactory.scala
shanthoosh: Just to add some context, functionality of `ChangelogStreamManager.createChangelogStreams` is not altered in this patch. Only thing, we did is to remove the proliferation of systemAdmin instances created by that method. 

However, I agree that that method did not have adequate testing. Added tests.

mynameborat:
Is this PR still needed?

shanthoosh:
This patch reduces the number of SystemAdmin instances created in ApplicationMaster startup control flow. This reduces the ApplicationMaster startup time. 

Let me fix the merge conflicts and update the patch.

prateekm:
@shanthoosh FYI, is this PR still relevant?

PR #1002, opened 2019-04-18
Fix ZkLocalApplicationRunner regression introduced by PR #951

mynameborat:
https://github.com/apache/samza/pull/951 introduced a regression in TestZkLocalApplicationRunner by creating topics with 5 partitions instead of 1.


shanthoosh:
@mynameborat Thanks for the fix. Merged  to trunk.

PR #1003, opened 2019-04-18
Removing job.diagnostics.appender.class config, and populating it based on if log4j or log4j2 is class-loaded

rmatharu:
Removing job.diagnostics.appender.class config, and populating it based on if log4j or log4j2 is class-loaded, defaults to assuming log4j.


samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -62,6 +62,21 @@ object Util extends Logging {
     }
   }
 
+  /**
+    * Instantiate an object from given className, and given constructor parameters.
+    */
+  def getObj(className: String, constructorParameterTypes: (Class[_], Object)*) = {
```
prateekm: As a general util method, I think it makes more sense for this to throw instead for returning an Option. You can then catch at your call site. 
Should also add explicit return type for public methods.
Maybe s/contructorParamaterTypes/constructorParams

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -62,6 +62,21 @@ object Util extends Logging {
...
+    try {
+      Class.forName(className).getDeclaredConstructor(constructorParameterTypes.map(x => x._1): _*)
+        .newInstance(constructorParameterTypes.map(x => x._2): _*)
+    } catch {
+      case e: Throwable => {
```
prateekm: Should not ignore Throwables :)

samza-core/src/main/scala/org/apache/samza/util/Util.scala
prateekm: Also, I think as written the return type will be Any (== Object) since in the happy path it returns an instance?

rmatharu: I was trying to avoid the nested try-catch. 
fixed.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -903,14 +903,22 @@ class SamzaContainer(
...
         case e@(_: ClassNotFoundException | _: InstantiationException | _: InvocationTargetException) => {
-          error("Failed to instantiate diagnostic appender", e)
-          throw new ConfigException("Failed to instantiate diagnostic appender class " +
-            config.getDiagnosticsAppenderClass, e)
+          try {
+            val diagnosticsAppender = Util.getObj("org.apache.samza.logging.log4j2.SimpleDiagnosticsAppender", (classOf[SamzaContainerMetrics], this.metrics))
+            info("Attached log4j2 diagnostics appender.")
+          } catch {
+            case e : Throwable => {
+              warn("Failed to instantiate neither diagnostic appender for sending error information to diagnostics stream", e)
```
prateekm: Should not ignore throwables. Maybe catch specific exceptions / Exception?

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -62,6 +62,21 @@ object Util extends Logging {
     }
   }
 
+  /**
+    * Instantiate an object from given className, and given constructor parameters.
+    */
+  def getObj(className: String, constructorParams: (Class[_], Object)*) : Any = {
```
prateekm: Should not return Any, can take [T] as a generic param and require passing the type (see other variant)

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
prateekm: Same for the catch below.

rmatharu: I dont want the uncaught exception handler to trigger container failure, if diagnostics cannot be started.

prateekm: Yeah, but I think you can catch the specific exceptions / Exception. E.g., all three exceptions above extend from Exception.

rmatharu:
Done, addressed all comments.

PR #1004, opened 2019-04-19
SAMZA-2169: Preventing task-shuffle after task mode addition

rmatharu:
The issue is described in detail https://issues.apache.org/jira/browse/SAMZA-2169

The fix is to treat a task with no-mode as an active task and still read the remaining task-container mapping from existing coord-stream.


samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -122,9 +122,9 @@ object JobModelManager extends Logging {
     val processorLocality: util.Map[String, LocationId] = getProcessorLocality(config, localityManager)
     val taskModes: util.Map[TaskName, TaskMode] = taskAssignmentManager.readTaskModes()
 
-    // We read the taskAssignment only for ActiveTasks
+    // We read the taskAssignment only for ActiveTasks, i.e., tasks that have no task-mode or have an active task mode
```
prateekm: Is it true that after upgrading to 1.1 every task will have the task mode? Or is mode only set if standby containers are enabled?

rmatharu: Every task will have a mode after the first re-deployment.

PR #1005, opened 2019-04-19
SAMZA-2170: Enabling writing of both new and old format offset files for stores and side-input-stores

rmatharu:
After Samza 1.1, the offset file for stores and sideinputs has been unified and is versioned. 
However, this Jira adds the logic in code to read and write both this new and old format. Because of this apps can switch between 1.0 and 1.1 versions seamlessly.

Note that the old format and filenames for store and side-input offset differed.


samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
```diff
@@ -139,13 +161,26 @@ public static boolean isOffsetFileValid(File storeDir, Set<SystemStreamPartition
...
     FileUtil.writeWithChecksum(offsetFile, fileContents);
+
+    // Now we write the old format offset file, which are different for store-offset and side-inputs
+    if (isSideInput) {
+      offsetFile = new File(getStorePartitionDir(storeBaseDir, storeName, taskName, taskMode), SIDE_INPUT_OFFSET_FILE_NAME_LEGACY);
+      fileContents = OBJECT_WRITER.writeValueAsString(offsets);
+      FileUtil.writeWithChecksum(offsetFile, fileContents);
+    } else {
+      offsetFile = new File(getStorePartitionDir(storeBaseDir, storeName, taskName, taskMode), OFFSET_FILE_NAME_LEGACY);
+      FileUtil.writeWithChecksum(offsetFile, offsets.entrySet().iterator().next().getValue());
```
cameronlee314: Would it be good to have some validation here to make sure there is exactly one entry in `offsets`?
Or maybe have two different methods: one for side inputs which takes in an `offsets` map and one for non-side-inputs which takes in a single `offset`.

samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
```diff
@@ -155,7 +190,15 @@ public static void writeOffsetFile(File storeBaseDir, String storeName, TaskName
    * @param taskName the task name which is referencing the store
    */
   public static void deleteOffsetFile(File storeBaseDir, String storeName, TaskName taskName) {
-    File offsetFile = new File(getStorePartitionDir(storeBaseDir, storeName, taskName, TaskMode.Active), OFFSET_FILE_NAME);
+    deleteOffsetFile(storeBaseDir, storeName, taskName, OFFSET_FILE_NAME_NEW);
```
cameronlee314: Do you need to delete the side-inputs offset file too?

samza-core/src/test/scala/org/apache/samza/storage/TestTaskStorageManager.scala
```diff
@@ -84,7 +84,7 @@ class TestTaskStorageManager extends MockitoSugar {
     val ssp = new SystemStreamPartition(ss, partition)
     val storeDirectory = StorageManagerUtil.getStorePartitionDir(TaskStorageManagerBuilder.defaultLoggedStoreBaseDir, loggedStore, taskName, TaskMode.Active)
     val storeFile = new File(storeDirectory, "store.sst")
-    val offsetFile = new File(storeDirectory, StorageManagerUtil.OFFSET_FILE_NAME)
```
cameronlee314: 1) Do you have a test for only having the old offset file with the old format?
2) Do you have a test for writing the old offset file with the old format?

samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
```diff
@@ -45,7 +45,9 @@
 
 public class StorageManagerUtil {
```
cameronlee314: Are there tests for this class?

rmatharu: No its tested from its consumers -- task-storage-manager, and sideinputstorage manager

samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
rmatharu: yup, 
i'll start a separate PR with these changes, and tests

samza-core/src/main/java/org/apache/samza/storage/StorageManagerUtil.java
rmatharu: no, this is only valid for task-stores

samza-core/src/test/scala/org/apache/samza/storage/TestTaskStorageManager.scala
rmatharu: yes, sending for review in a separate pr

PR #1006, opened 2019-04-22
SAMZA-2171 : Implement MetadataResourceManager to encapsulate loading of metadata resources

dnishimura:
**Motivation:** Before this PR, the code to create the checkpoint and changelog streams are copy and pasted to each JobCoordinator implementation. This PR encapsulates that logic in a `MetadataResourceManager` so that future JobCoordinator implementations don't have to copy and paste duplicate logic and we can add other metadata resources that we may want to create in the future.

@shanthoosh @prateekm @cameronlee314 - please take a look when you get a chance. The likely follow on is to add the startpoint manager fan out call in the `MetadataResourceManager#createResources` method.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceLoader.java
+++ b/samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceLoader.java
@@ -1,0 +1,99 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.coordinator;
+
+import org.apache.samza.checkpoint.CheckpointManager;
+import org.apache.samza.config.Config;
+import org.apache.samza.config.TaskConfigJava;
+import org.apache.samza.coordinator.metadatastore.NamespaceAwareCoordinatorStreamStore;
+import org.apache.samza.coordinator.stream.messages.SetChangelogMapping;
+import org.apache.samza.job.model.JobModel;
+import org.apache.samza.metadatastore.MetadataStore;
+import org.apache.samza.metrics.MetricsRegistry;
+import org.apache.samza.startpoint.StartpointManager;
+import org.apache.samza.storage.ChangelogStreamManager;
+
+
+/**
+ * Loads the managers responsible for the creation and loading of metadata related resources
+ */
+public class MetadataResourceLoader {
+  private final Config config;
+  private final CheckpointManager checkpointManager;
+  private final ChangelogStreamManager changelogStreamManager;
+  private final StartpointManager startpointManager;
+  private final JobModel jobModel; // TODO: Should be loaded by metadata store in the future
+
+  /**
+   * @param metadataStore the {@link MetadataStore} that manages the metadata. Assumption is that the metadataStore is
+   *                      initialized and not stopped.
+   * @param jobModel the loaded {@link JobModel}
+   * @param metricsRegistry the registry for reporting metrics.
+   * @param configOverride the config override to use instead of the one in the jobModel.
+   */
+  public MetadataResourceLoader(MetadataStore metadataStore, JobModel jobModel, MetricsRegistry metricsRegistry, Config configOverride) {
+    this.config = configOverride;
+    this.jobModel = jobModel;
+    this.checkpointManager = new TaskConfigJava(config).getCheckpointManager(metricsRegistry);
+    this.changelogStreamManager = metadataStore == null
+        ? null
+        : new ChangelogStreamManager(new NamespaceAwareCoordinatorStreamStore(metadataStore, SetChangelogMapping.TYPE));
+    this.startpointManager = metadataStore == null
+        ? null
+        : new StartpointManager(new NamespaceAwareCoordinatorStreamStore(metadataStore, StartpointManager.NAMESPACE));
+  }
+
+  /**
+   * @param metadataStore the {@link MetadataStore} that manages the metadata
+   * @param jobModel the loaded {@link JobModel}
+   * @param metricsRegistry the registry for reporting metrics.
+   */
+  public MetadataResourceLoader(MetadataStore metadataStore, JobModel jobModel, MetricsRegistry metricsRegistry) {
+    this(metadataStore, jobModel, metricsRegistry, jobModel.getConfig());
+  }
+
+  /**
+   * Creates and loads the required metadata resources for checkpoints, changelog stream and other
+   * resources related to the metadata system
+   */
+  public void createResources() {
+    CheckpointManager checkpointManager = getCheckpointManager();
+    if (checkpointManager != null) {
+      checkpointManager.createResources();
+    }
+    createChangelogStreams();
+  }
+
+  public CheckpointManager getCheckpointManager() {
+    return checkpointManager;
+  }
+
+  public ChangelogStreamManager getChangelogStreamManager() {
+    return changelogStreamManager;
+  }
+
+  public StartpointManager getStartpointManager() {
+    return startpointManager;
+  }
+
+  void createChangelogStreams() {
+    ChangelogStreamManager.createChangelogStreams(config, jobModel.maxChangeLogStreamPartitions);
+  }
+}
--- a/samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
+++ b/samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
@@ -32,9 +32,10 @@ import org.apache.samza.storage.ChangelogStreamManager;
 
 
 /**
- * Loads the managers responsible for the creation and loading of metadata related resources
+ * Loads the managers responsible for the creation and loading of metadata related resources.
+ * TODO: Replace with a metadata admin interface when the {@link MetadataStore} is fully augmented to handle all metadata sources.
  */
-public class MetadataResourceLoader {
+public class MetadataResourceManager {
   private final Config config;
   private final CheckpointManager checkpointManager;
   private final ChangelogStreamManager changelogStreamManager;
@@ -48,7 +49,7 @@ public class MetadataResourceLoader {
    * @param metricsRegistry the registry for reporting metrics.
    * @param configOverride the config override to use instead of the one in the jobModel.
    */
-  public MetadataResourceLoader(MetadataStore metadataStore, JobModel jobModel, MetricsRegistry metricsRegistry, Config configOverride) {
+  public MetadataResourceManager(MetadataStore metadataStore, JobModel jobModel, MetricsRegistry metricsRegistry, Config configOverride) {
     this.config = configOverride;
     this.jobModel = jobModel;
     this.checkpointManager = new TaskConfigJava(config).getCheckpointManager(metricsRegistry);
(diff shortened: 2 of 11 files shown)


samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceLoader.java
```diff
@@ -0,0 +1,99 @@
...
+   */
+  public MetadataResourceLoader(MetadataStore metadataStore, JobModel jobModel, MetricsRegistry metricsRegistry) {
+    this(metadataStore, jobModel, metricsRegistry, jobModel.getConfig());
+  }
+
+  /**
+   * Creates and loads the required metadata resources for checkpoints, changelog stream and other
+   * resources related to the metadata system
+   */
+  public void createResources() {
```
dnishimura: I plan to add either the fan out call here for the fan out model or the ack clean up call here for the intent-ack model.

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
```diff
@@ -0,0 +1,100 @@
...
+   */
+  public MetadataResourceManager(MetadataStore metadataStore, JobModel jobModel, MetricsRegistry metricsRegistry) {
+    this(metadataStore, jobModel, metricsRegistry, jobModel.getConfig());
+  }
+
+  /**
+   * Creates and loads the required metadata resources for checkpoints, changelog stream and other
+   * resources related to the metadata system
+   */
+  public void createResources() {
```
dnishimura: For the follow on startpoint work, I plan to add either the fan out call here for the fan out model or the ack clean up call here for the intent-ack model.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -296,15 +288,19 @@ void doOnProcessorChange() {
...
     MetadataStore metadataStore = null;
     try {
       // Creates the coordinator stream if it does not exists.
       createCoordinatorStream();
-
       MetadataStoreFactory metadataStoreFactory = Util.getObj(new JobConfig(config).getMetadataStoreFactory(), MetadataStoreFactory.class);
       metadataStore = metadataStoreFactory.getMetadataStore(SetConfig.TYPE, config, metrics.getMetricsRegistry());
       metadataStore.init();
+
+      MetadataResourceManager metadataResourceManager =
```
cameronlee314: Minor: Just to keep the same ordering, could this logic to above `createCoordinatorStream`? Maybe it would be slightly better since then the coordinator stream is created right before its usage (e.g. if `metadataResourceManager` fails, then we can skip trying to create the coord stream).

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
```diff
@@ -0,0 +1,100 @@
...
+
+  public CheckpointManager getCheckpointManager() {
+    return checkpointManager;
+  }
+
+  public ChangelogStreamManager getChangelogStreamManager() {
+    return changelogStreamManager;
+  }
+
+  public StartpointManager getStartpointManager() {
```
cameronlee314: Is this used?

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+public class MetadataResourceManager {
+  private final Config config;
+  private final CheckpointManager checkpointManager;
+  private final ChangelogStreamManager changelogStreamManager;
+  private final StartpointManager startpointManager;
+  private final JobModel jobModel; // TODO: Should be loaded by metadata store in the future
+
+  /**
+   * @param metadataStore the {@link MetadataStore} that manages the metadata. Assumption is that the metadataStore is
```
cameronlee314: Could you please doc about what happens if this is null.

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * Creates and loads the required metadata resources for checkpoints, changelog stream and other
+   * resources related to the metadata system
+   */
+  public void createResources() {
+    CheckpointManager checkpointManager = getCheckpointManager();
+    if (checkpointManager != null) {
+      checkpointManager.createResources();
```
cameronlee314: Do you need to stop the `checkpointManager`? Looks like some existing places might have forgotten to do it already though.

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.samza.metrics.MetricsRegistry;
+import org.apache.samza.startpoint.StartpointManager;
+import org.apache.samza.storage.ChangelogStreamManager;
+
+
+/**
+ * Loads the managers responsible for the creation and loading of metadata related resources.
+ * TODO: Replace with a metadata admin interface when the {@link MetadataStore} is fully augmented to handle all metadata sources.
+ */
+public class MetadataResourceManager {
```
prateekm: Is this concept specific to Kafka streams? E.g., what does it mean to create resources if the metadata store is a DB? If the former, can we do this as a util instead of a new interface?

dnishimura: No, it's to consolidate managers that interact with the metadata store or other metadata streams. It is essentially a util class. I can change the name to `*Util` instead of `*Manager` if that makes better sense. Ideally, when the metadata store extensions are fully designed and implemented, that should replace this.

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
dnishimura: Sure will do

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
dnishimura: Not when creating the checkpoint stream itself. Only when reading/writing to it.

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
```diff
@@ -0,0 +1,100 @@
...
+   */
+  public void createResources() {
+    CheckpointManager checkpointManager = getCheckpointManager();
+    if (checkpointManager != null) {
+      checkpointManager.createResources();
+    }
+    createChangelogStreams();
+  }
+
+  public CheckpointManager getCheckpointManager() {
```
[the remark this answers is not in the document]
dnishimura: I thought to add getters for all the encapsulated managers, but I can remove it or reduce the visibility until they are needed.

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
```diff
@@ -0,0 +1,100 @@
...
+    return checkpointManager;
+  }
+
+  public ChangelogStreamManager getChangelogStreamManager() {
```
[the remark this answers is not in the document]
dnishimura: See comment about `getCheckpointManager()`

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
dnishimura: See comment about `getCheckpointManager()`

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
```diff
@@ -0,0 +1,100 @@
...
+    return changelogStreamManager;
+  }
+
+  public StartpointManager getStartpointManager() {
+    return startpointManager;
+  }
+
+  void createChangelogStreams() {
```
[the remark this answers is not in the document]
dnishimura: Sure will look into it.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
dnishimura: Unfortunately no. The metadata store assumes the coordinator stream is created before being initialized. It does seem a little disconnected, but it's an issue with the metadata store design and implementation. I'll create a ticket to note this for the continued metadata store work.

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceUtil.java
```diff
@@ -0,0 +1,84 @@
...
+    this.config = jobModel.getConfig();
+    this.jobModel = jobModel;
+    this.checkpointManager = checkpointManager;
+  }
+
+  /**
+   * Creates and loads the required metadata resources for checkpoints, changelog stream and other
+   * resources related to the metadata system
+   */
+  public void createResources() {
```
dnishimura: For the follow on startpoint work, I plan to add either the fan out call here for the fan out model or the ack clean up call here for the intent-ack model.

dnishimura:
@cameronlee314 & @prateekm please see my responses and changes. Thanks.

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceManager.java
prateekm: What about LocalityManager / TaskAssignmentManager etc? Should they be managed by this class as well? Looking at just this class in isolation, it's unclear what its responsibilities are. +1 for calling it XUtil if it's temporary and for code cleanup/consolidation.

dnishimura: Any suggestions to make the responsibility more obvious? I already renamed the class to `*Util` and already have a `TODO:` comment in the document to replace this in the future once the full design and implementation of the metadata store is in place. Think it's sufficient for now?

prateekm: My bad, was looking at outdated version. Sounds good to me, thanks.

samza-core/src/test/java/org/apache/samza/coordinator/TestMetadataResourceUtil.java
```diff
@@ -20,24 +20,30 @@
...
 
   @Test
   public void testLoad() {
-    CheckpointManager checkpointManager = Mockito.mock(CheckpointManager.class);
-    MetadataResourceManager metadataResourceManager = Mockito.mock(MetadataResourceManager.class);
-    Mockito.doCallRealMethod().when(metadataResourceManager).createResources();
-    Mockito.doReturn(checkpointManager).when(metadataResourceManager).getCheckpointManager();
-    Mockito.doNothing().when(metadataResourceManager).createChangelogStreams();
+    MetadataResourceUtil metadataResourceUtil = Mockito.spy(new MetadataResourceUtil(checkpointManager, jobModel));
+    Mockito.doCallRealMethod().when(metadataResourceUtil).createResources();
```
cameronlee314: You don't need to do `doCallReadMethod` for `spy` objects. As long as you don't stub a method on a `spy`, it will call the real method by default.

samza-core/src/main/scala/org/apache/samza/job/local/ThreadJobFactory.scala
```diff
@@ -68,12 +68,8 @@ class ThreadJobFactory extends StreamJobFactory with Logging {
     changelogStreamManager.writePartitionMapping(taskPartitionMappings)
 
     //create necessary checkpoint and changelog streams
-    val checkpointManager = new TaskConfigJava(jobModel.getConfig).getCheckpointManager(metricsRegistry)
-    if (checkpointManager != null) {
-      checkpointManager.createResources()
-      checkpointManager.stop()
-    }
-    ChangelogStreamManager.createChangelogStreams(jobModel.getConfig, jobModel.maxChangeLogStreamPartitions)
+    val metadataResourceLoader = new MetadataResourceUtil(coordinatorStreamStore, jobModel, metricsRegistry)
```
shanthoosh: nit: s/metadataResourceLoader/metadataResourceUtil

cameronlee314:
@prateekm do you have further comments/concerns? Otherwise, will merge.

dnishimura:
> @prateekm do you have further comments/concerns? Otherwise, will merge.

@cameronlee314 - He told me offline he wasn't actively reviewing this PR and he gave a 👍 for my response on his sole comment.

PR #1007, opened 2019-04-23
SAMZA-2173: Fix NullPointerException in SetConfig MessageFormat when configurations are deleted.

shanthoosh:
Currently the configurations in the coordinator stream are deleted by sending a  Delete message with  message-map set as null and type set as `SetConfig.TYPE`. 

To read configurations from the coordinator stream, there should be a check at the caller API to find if configuration message has been deleted or not. Otherwise it results in a NPE. 

This patch encapsulates the isDelete check into the CoordinatorStreamMessage class and returns null if the configuration key has been deleted.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/coordinator/stream/messages/CoordinatorStreamMessage.java
+++ b/samza-core/src/main/java/org/apache/samza/coordinator/stream/messages/CoordinatorStreamMessage.java
@@ -191,7 +191,11 @@ public class CoordinatorStreamMessage {
   }
 
   protected String getMessageValue(String key) {
-    return getMessageValues().get(key);
+    if (isDelete) {
+      return null;
+    } else {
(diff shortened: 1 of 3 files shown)


shanthoosh:
@prateekm 

shanthoosh:
@prateekm 

Currently to delete a key in coordinator stream, samza produces the message (key, value: null) into the coordinator-stream. This was the behavior since 0.9 version of samza.

Previously upto 0.8 version of samza, to delete a key, we used to serialize the null value into the json message and publish it to coordinator stream. Here's a sample message taken from the coordinator stream which follows this pattern in production. 

```json
offset: 314005  LogAppendTime: 1555029967663  key: [1,"set-config","random-key"] payload: {"host":"random-host","source":"job-runner","values":{"value":null},"username":"random","timestamp":1255429967479}
```

PR #1008, opened 2019-04-24
SAMZA-2174: Throw a record too large exception for oversized records in changelog

PawasChhokra:
Right now, serialization of values happens after we create a CachedStore.
CachedStore does a put/putAll during flush, after it has stored the large message in its cache. This means that the large message will be stored in the cache but rejected when it reaches the addition to the changelog in the LoggedStore.
One option is to let the message remain in the cache and reject the message when it reaches the LoggedStore.
The other option is to serialize the message first in order to decipher the size of the message, then check the size of this message, and later invoke the CachedStore if required.

We are going to go ahead with both these options that the user can enable with a config value.
2 config values:
First option: expect.large.message
Second option: drop.large.messages

Behaviour (True:T, False:F):
TF, TT: throw RecordTooLargeException before adding to cached store
FT: ignore the large message (cached store will store it but it won’t be written to the db)
FF: current state as it is right now, kv store ingests the value but code breaks while adding to changelog stream

@prateekm Kindly take a look.


samza-kv/src/main/scala/org/apache/samza/storage/kv/LoggedStore.scala
```diff
@@ -19,34 +19,44 @@
...
   val collector: MessageCollector,
-  val metrics: LoggedStoreMetrics = new LoggedStoreMetrics) extends KeyValueStore[K, V] with Logging {
+  val metrics: LoggedStoreMetrics = new LoggedStoreMetrics) extends KeyValueStore[Array[Byte], Array[Byte]] with Logging {
 
   val systemStream = systemStreamPartition.getSystemStream
   val partitionId = systemStreamPartition.getPartition.getPartitionId
 
+  private val DEFAULT_CHANGELOG_MAX_MSG_SIZE_BYTES = 1000000
+  private val CHANGELOG_MAX_MSG_SIZE_BYTES = "changelog.max.message.size.bytes"
+  private val maxMessageSize = storeConfig.getInt(CHANGELOG_MAX_MSG_SIZE_BYTES, DEFAULT_CHANGELOG_MAX_MSG_SIZE_BYTES) // slightly less than 1 MB
```
jmakes: nit: comment should be moved to line 44

samza-kv/src/main/scala/org/apache/samza/storage/kv/LoggedStore.scala
```diff
@@ -114,7 +128,14 @@ class LoggedStore[K, V](
...
   }
 
-  override def snapshot(from: K, to: K): KeyValueSnapshot[K, V] = {
+  override def snapshot(from: Array[Byte], to: Array[Byte]): KeyValueSnapshot[Array[Byte], Array[Byte]] = {
     store.snapshot(from, to)
   }
+
+  private def validateMessageSize(key: Array[Byte], message: Array[Byte]): Unit = {
+    if (message.length > maxMessageSize) {
+      throw new SamzaException("RocksDB message size " + message.length + " for store " + storeName
```
jmakes: nit: exception message is for the logged store, not RocksDB. AFAIK RocksDB has no limit.

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestLoggedStore.scala
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.samza.task.MessageCollector
+import org.apache.samza.{Partition, SamzaException}
+import org.junit.{Before, Test}
+import org.mockito.Matchers
+import org.mockito.Mockito._
+
+
+class TestLoggedStore {
+
+  val store = mock(classOf[KeyValueStore[Array[Byte], Array[Byte]]])
```
jmakes: Has this been tested with a real store implementation? Most samza jobs do not create KVStore<byte[], byte[]>, they create something typed and then after SerializedKVStore serializes the entries, it becomes byte[], byte[]. 

It would be good to ensure the types are compatible in a real scenario with each of the KVStore wrappers.

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestLoggedStore.scala
```diff
@@ -0,0 +1,105 @@
...
+    try {
+      safeLoggedStore.put(key2, largeMessage)
+    } catch {
+      case e: SamzaException =>
+        verify(store, times(0)).put(Matchers.eq(key2), Matchers.eq(largeMessage))
+    }
+  }
+
+  @Test
+  def testSmallMessagePutAllSuccess = {
```
rmatharu: Maybe add tests with sample-messages serialized using serializers in org.apache.samza.serializers.

samza-kv/src/main/scala/org/apache/samza/storage/kv/LoggedStore.scala
PawasChhokra: Done.

samza-kv/src/main/scala/org/apache/samza/storage/kv/LoggedStore.scala
PawasChhokra: Done.

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestLoggedStore.scala
PawasChhokra: Done.

samza-kv/src/main/scala/org/apache/samza/storage/kv/LoggedStore.scala
```diff
@@ -19,34 +19,44 @@
...
+  val storeConfig: Config,
   val systemStreamPartition: SystemStreamPartition,
   val collector: MessageCollector,
-  val metrics: LoggedStoreMetrics = new LoggedStoreMetrics) extends KeyValueStore[K, V] with Logging {
+  val metrics: LoggedStoreMetrics = new LoggedStoreMetrics) extends KeyValueStore[Array[Byte], Array[Byte]] with Logging {
 
   val systemStream = systemStreamPartition.getSystemStream
   val partitionId = systemStreamPartition.getPartition.getPartitionId
 
+  private val DEFAULT_CHANGELOG_MAX_MSG_SIZE_BYTES = 1000000
```
shanthoosh: Just curious. 

Maximum message size accepted by a kafka broker is defined by the configuration: `message.max.bytes`.  There're new asynchronous API's added in kafka-AdminClient which allows the users to query the broker configuration through the `AdminClient` API. Broker configurations can be accessed using  `AdminClient.describeConfigs();` API

Rather than hardcoding the maximum value from samza's-end, would it be beneficial to query the configured value directly once from the broker during container-startup. By that, for different kind of brokers(with different supported max-message-size), users might not have to tune this configuration correctly at samza-end. 

What do you think?

PawasChhokra: I've designed the current code in a way where this change will be easy to do in the future. Created a follow up ticket for this here: https://issues.apache.org/jira/browse/SAMZA-2212

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestLoggedStore.scala
PawasChhokra: I've changed the implementation to have a dedicated LargeMessageStore that runs on 2 configs given by the user to check for large messages and handle them, and added the necessary integration tests for this. I've added more information about this in the description.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -35,15 +35,22 @@ import org.apache.samza.util.ScalaJavaUtil.JavaOptionals
...
+  * A key value storage engine factory implementation
+  *
+  * This trait encapsulates all the steps needed to create a key value storage engine. It is meant to be extended
+  * by the specific key value store factory implementations which will in turn override the getKVStore method.
+  */
 trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V] {
 
   private val INMEMORY_KV_STORAGE_ENGINE_FACTORY =
     "org.apache.samza.storage.kv.inmemory.InMemoryKeyValueStorageEngineFactory"
+  private val CHANGELOG_MAX_MSG_SIZE_BYTES = "changelog.max.message.size.bytes"
```
cameronlee314: Consider moving these configs to `StorageConfig`.
Also, it would probably be good to be able to configure these per store (see `StorageConfig`).

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -124,22 +135,56 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
       new LoggedStore(rawStore, changeLogSystemStreamPartition, collector, loggedStoreMetrics)
     }
 
-    // wrap with serialization
-    val serializedMetrics = new SerializedKeyValueStoreMetrics(storeName, registry)
-    val serialized = new SerializedKeyValueStore[K, V](maybeLoggedStore, keySerde, msgSerde, serializedMetrics)
+    var toBeAccessLoggedStore: KeyValueStore[K, V] = null
+
+    if (largeMessagesExpected) {
```
cameronlee314: The `getStorageEngine` method is already pretty long, and it's a little hard to keep track of all of the different wrapped stores. Please consider pulling out some of your new logic to other helper methods for readability. Also, it might allow you to share some methods with the "dropLargeMessage" case.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -35,15 +35,22 @@ import org.apache.samza.util.ScalaJavaUtil.JavaOptionals
...
+  * This trait encapsulates all the steps needed to create a key value storage engine. It is meant to be extended
+  * by the specific key value store factory implementations which will in turn override the getKVStore method.
+  */
 trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V] {
 
   private val INMEMORY_KV_STORAGE_ENGINE_FACTORY =
     "org.apache.samza.storage.kv.inmemory.InMemoryKeyValueStorageEngineFactory"
+  private val CHANGELOG_MAX_MSG_SIZE_BYTES = "changelog.max.message.size.bytes"
+  private val DEFAULT_CHANGELOG_MAX_MSG_SIZE_BYTES = 1000000 // slightly less than 1 MB
+  private val EXPECT_LARGE_MESSAGES = "expect.large.message"
```
cameronlee314: Are there any docs/javadocs explaining the consequences of enabling these configs?
e.g.:
1) "expect.large.message" causes a perf degredation if using the cached store
2) "drop.large.messages" can result in inconsistent state between the local store and the changelog

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestLargeMessageSafeStore.scala
```diff
@@ -0,0 +1,187 @@
...
+
+    val entries = new ArrayList[Entry[Array[Byte], Array[Byte]]]
+    entries.add(new Entry(key, largeMessage))
+
+    try {
+      largeMessageSafeKeyValueStore.putAll(entries)
+      fail("The test case should have failed due to a large message being passed to the changelog, but it didn't.")
+    } catch {
+      case e: SamzaException =>
+        verify(store, times(0)).putAll(Matchers.eq(entries))
```
cameronlee314: You can do `verifyZeroInteractions(store)` to make sure it just never got used at all.

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestLargeMessageSafeStore.scala
```diff
@@ -0,0 +1,187 @@
...
+    obj.put("jack", "jill")
+    obj.put("john", new java.lang.Integer(2))
+    val smallMessage = jsonSerde.toBytes(obj)
+
+    val entries = new ArrayList[Entry[Array[Byte], Array[Byte]]]
+    entries.add(new Entry(key, smallMessage))
+
+    largeMessageSafeKeyValueStore.putAll(entries)
+    verify(store, times(1)).putAll(Matchers.eq(entries))
+    verify(store, times(0)).put(Matchers.any(classOf[Array[Byte]]), Matchers.any(classOf[Array[Byte]]))
```
cameronlee314: Instead of `verify(store, times(0))`, you can do `verify(store, never())`
Same for other cases below

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestLargeMessageSafeStore.scala
```diff
@@ -0,0 +1,187 @@
...
+    val obj = new java.util.HashMap[String, Object]();
+    obj.put("jack", "jill")
+    obj.put("john", new java.lang.Integer(2))
+    val smallMessage = jsonSerde.toBytes(obj)
+
+    val entries = new ArrayList[Entry[Array[Byte], Array[Byte]]]
+    entries.add(new Entry(key, smallMessage))
+
+    largeMessageSafeKeyValueStore.putAll(entries)
+    verify(store, times(1)).putAll(Matchers.eq(entries))
```
cameronlee314: Default is already `times(1)`, so you can just do `verify(store)`
Same for other cases below

samza-test/src/test/scala/org/apache/samza/storage/kv/TestLargeMessageSafeKeyValueStores.scala
```diff
@@ -0,0 +1,184 @@
...
+import org.apache.samza.task.MessageCollector
+import org.apache.samza.{Partition, SamzaException}
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+import org.junit.runners.Parameterized.Parameters
+import org.junit.{After, Before, Test}
+import org.scalatest.Assertions.fail
+
+@RunWith(value = classOf[Parameterized])
+class TestLargeMessageSafeKeyValueStores(typeOfStore: String, storeConfig: String, dropLargeMessageStr: String) {
```
cameronlee314: Can you please put a javadoc about what you are testing with this?

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestLargeMessageSafeStore.scala
```diff
@@ -0,0 +1,187 @@
...
+import org.apache.samza.serializers.{JsonSerdeV2, LongSerde, StringSerde}
+import org.apache.samza.system.SystemStreamPartition
+import org.apache.samza.task.MessageCollector
+import org.apache.samza.{Partition, SamzaException}
+import org.junit.{Before, Test}
+import org.mockito.Matchers
+import org.mockito.Mockito._
+import org.junit.Assert._
+
+class TestLargeMessageSafeStore {
```
cameronlee314: Please write all new classes in Java.

samza-test/src/test/scala/org/apache/samza/storage/kv/TestLargeMessageSafeKeyValueStores.scala
cameronlee314: Please write all new classes in Java.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -124,22 +135,56 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
...
-    val serialized = new SerializedKeyValueStore[K, V](maybeLoggedStore, keySerde, msgSerde, serializedMetrics)
+    var toBeAccessLoggedStore: KeyValueStore[K, V] = null
+
+    if (largeMessagesExpected) {
+      // Execute "if" when we expect to get large messages and user defined config -> EXPECT_LARGE_MESSAGES = true
+      // This code path will throw a SamzaException for large message size.
+
+      // maybe wrap with caching
+      val maybeCachedStore = if (enableCache) {
+        val cachedStoreMetrics = new CachedStoreMetrics(storeName, registry)
```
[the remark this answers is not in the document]
PawasChhokra: I've changed the documentation to reflect that there will be no deserialization when the key itself is an array of bytes.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
PawasChhokra: Done.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
PawasChhokra: Done. Thanks!

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
PawasChhokra: Sorry I missed this earlier. Added to the docs now.

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestLargeMessageSafeStore.scala
PawasChhokra: Done.

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestLargeMessageSafeStore.scala
PawasChhokra: Done.

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestLargeMessageSafeStore.scala
PawasChhokra: Done.

samza-test/src/test/scala/org/apache/samza/storage/kv/TestLargeMessageSafeKeyValueStores.scala
PawasChhokra: Done.

PawasChhokra: Done.

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestLargeMessageSafeStore.scala
PawasChhokra: Done.

samza-kv/src/main/scala/org/apache/samza/storage/kv/LargeMessageSafeStore.scala
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+  * A wrapper over KeyValueStore that checks to see if any large messages have been passed to the store.
+  *
+  * @param store the store to persist to if messages are not large
+  * @param storeName store name
+  * @param dropLargeMessage a flag that determines if we want to ignore and drop large messages
+  * @param maxMessageSize maximum allowed size of a message, after which every message will be considered large
+  */
+class LargeMessageSafeStore(
```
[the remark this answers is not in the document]
PawasChhokra: Done.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1687,6 +1687,50 @@ <h1>Samza Configuration Reference</h1>
                     </td>
                 </tr>
 
+                <tr>
+                    <td class="property" id="stores-changelog-max-message-size-bytes">stores.<span class="store">store-name</span>.changelog.max.message.size.bytes</td>
+                    <td class="default"></td>
+                    <td class="description">
+                        This property sets the maximum size of the messages allowed in the changelog.
+                        The maximum allowed value and the default value both are 1 MB.
```
cameronlee314: Why is the max allowed value 1MB? If the changelog supports a larger value, it seems like it would be good to allow that.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1687,6 +1687,50 @@ <h1>Samza Configuration Reference</h1>
...
+                        <a href="#stores-drop-large-messages" class="property">stores.*.drop.large.messages</a>.
+                        The default value for this config is false. When this property is not set,
+                        <a href="#stores-drop-large-messages" class="property">stores.*.drop.large.messages</a>
+                        defines the behaviour to be executed.
+                    </td>
+                </tr>
+
+                <tr>
+                    <td class="property" id="stores-drop-large-messages">stores.<span class="store">store-name</span>.drop.large.messages</td>
+                    <td class="default"></td>
```
cameronlee314: default is "false", right?

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1687,6 +1687,50 @@ <h1>Samza Configuration Reference</h1>
...
+                    <td class="description">
+                        This property sets the maximum size of the messages allowed in the changelog.
+                        The maximum allowed value and the default value both are 1 MB.
+                    </td>
+                </tr>
+
+                <tr>
+                    <td class="property" id="stores-expect-large-messages">stores.<span class="store">store-name</span>.expect.large.messages</td>
+                    <td class="default"></td>
```
cameronlee314: default is "false", right?

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1687,6 +1687,50 @@ <h1>Samza Configuration Reference</h1>
                     </td>
                 </tr>
 
+                <tr>
+                    <td class="property" id="stores-changelog-max-message-size-bytes">stores.<span class="store">store-name</span>.changelog.max.message.size.bytes</td>
+                    <td class="default"></td>
```
cameronlee314: Please fill in the default.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -89,6 +90,10 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
     var storePropertiesBuilder = new StoreProperties.StorePropertiesBuilder()
     val accessLog = storageConfig.getAccessLogEnabled(storeName)
 
+    val maxMessageSize = storageConfig.getInt(StorageConfig.CHANGELOG_MAX_MSG_SIZE_BYTES, StorageConfig.DEFAULT_CHANGELOG_MAX_MSG_SIZE_BYTES)
```
cameronlee314: The config keys are format strings which require the store names.

docs/learn/documentation/versioned/jobs/configuration-table.html
PawasChhokra: Changed to 5MB.

docs/learn/documentation/versioned/jobs/configuration-table.html
PawasChhokra: Yes, added.

docs/learn/documentation/versioned/jobs/configuration-table.html
PawasChhokra: Yes, added.

docs/learn/documentation/versioned/jobs/configuration-table.html
PawasChhokra: Done.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
PawasChhokra: Sorry for the mistake. I've changed it to retrieve from storageConfigSubset instead.

docs/learn/documentation/versioned/jobs/configuration-table.html
cameronlee314: I wasn't referring to the 1MB number specifically. It's fine as 1MB or 5MB.
I was referring to the language used of "maximum allowed value". Why is there a maximum allowed value for this config?

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1689,16 +1689,16 @@ <h1>Samza Configuration Reference</h1>
 
                 <tr>
                     <td class="property" id="stores-changelog-max-message-size-bytes">stores.<span class="store">store-name</span>.changelog.max.message.size.bytes</td>
-                    <td class="default"></td>
+                    <td class="default">1000000</td>
                     <td class="description">
                         This property sets the maximum size of the messages allowed in the changelog.
-                        The maximum allowed value and the default value both are 1 MB.
+                        The maximum allowed value is 5 MB and the default value is 1 MB.
```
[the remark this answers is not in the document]
PawasChhokra: There is a maximum allowed value set called max.request.size which is 1 MB as seen here: https://kafka.apache.org/documentation/#producerconfigs. Hence, changed it to 1 MB.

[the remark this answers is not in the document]
PawasChhokra: Also added a check to see if someone sets it to greater than 1 MB.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -90,9 +90,9 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
     var storePropertiesBuilder = new StoreProperties.StorePropertiesBuilder()
     val accessLog = storageConfig.getAccessLogEnabled(storeName)
 
-    val maxMessageSize = storageConfig.getInt(StorageConfig.CHANGELOG_MAX_MSG_SIZE_BYTES, StorageConfig.DEFAULT_CHANGELOG_MAX_MSG_SIZE_BYTES)
-    val largeMessagesExpected = storageConfig.getBoolean(StorageConfig.EXPECT_LARGE_MESSAGES, StorageConfig.DEFAULT_EXPECT_LARGE_MESSAGES)
-    val dropLargeMessage = storageConfig.getBoolean(StorageConfig.DROP_LARGE_MESSAGES, StorageConfig.DEFAULT_DROP_LARGE_MESSAGES)
+    val maxMessageSize = storageConfigSubset.getInt(StorageConfig.CHANGELOG_MAX_MSG_SIZE_BYTES, StorageConfig.DEFAULT_CHANGELOG_MAX_MSG_SIZE_BYTES)
```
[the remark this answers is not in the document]
PawasChhokra: Added tests to TestStorageConfig related to this. Thanks.

docs/learn/documentation/versioned/jobs/configuration-table.html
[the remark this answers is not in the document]
cameronlee314: This flow should not assume a Kafka is the changelog system. Also, Kafka allows for setting the max.request.size to be higher than 1MB, so we should not restrict that.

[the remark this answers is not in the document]
PawasChhokra: Removed the comment and the check about maximum allowed value.

docs/learn/documentation/versioned/jobs/configuration-table.html
PawasChhokra: Removed the comment about the maximum allowed value.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -32,15 +32,15 @@ import org.apache.samza.storage.{StorageEngine, StorageEngineFactory, StorePrope
...
+import org.apache.samza.util.{HighResolutionClock, Logging}
 
 /**
   * A key value storage engine factory implementation
   *
   * This trait encapsulates all the steps needed to create a key value storage engine. It is meant to be extended
   * by the specific key value store factory implementations which will in turn override the getKVStore method.
   */
-trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V] {
+trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V] with Logging {
```
cameronlee314: Minor: Is the `Logging` trait still needed?

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -162,4 +184,28 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
       keyValueStorageEngineMetrics, batchSize, () => clock.nanoTime())
   }
 
+  def createCachedStore[K, V](enableCache: Boolean, storeName: String, registry: MetricsRegistry,
```
prateekm: Instead of passing a boolean as a method parameter, it's better to just do the if-else at the call site and call/not call the method.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -162,4 +184,28 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
...
+    val maybeCachedStore = if (enableCache) {
+      val cachedStoreMetrics = new CachedStoreMetrics(storeName, registry)
+      new CachedStore(underlyingStore, cacheSize, batchSize, cachedStoreMetrics)
+    } else {
+      underlyingStore
+    }
+    maybeCachedStore
+  }
+
+  def createLargeMessageSafeStore(underlyingStore: KeyValueStore[Array[Byte], Array[Byte]], storeName: String,
```
prateekm: Why does this need to be a separate method? It's should be the same amount of code to inline this.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -162,4 +184,28 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
...
+    dropLargeMessage: Boolean, maxMessageSize: Int ): LargeMessageSafeStore = {
+    // wrap with large message checking
+    new LargeMessageSafeStore(underlyingStore, storeName, dropLargeMessage, maxMessageSize)
+  }
+
+  def createSerializedStore(underlyingStore: KeyValueStore[Array[Byte], Array[Byte]], storeName: String,
```
prateekm: Same as above.

samza-kv/src/main/scala/org/apache/samza/storage/kv/LoggedStore.scala
```diff
@@ -24,8 +24,8 @@ import org.apache.samza.system.{OutgoingMessageEnvelope, SystemStreamPartition}
 import org.apache.samza.task.MessageCollector
 
 /**
- * A key/value store decorator that adds a changelog for any changes made to the underlying store
- */
+  * A key/value store decorator that adds a changelog for any changes made to the underlying store
```
prateekm: Please revert indentation changes from javadocs in all classes, unless this is what IntelliJ does by default.

prateekm: Looks like all changes in this class can be reverted?

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
cameronlee314: I think I had suggested this. It wasn't for saving code. It was just to help with readability. When everything was inlined, I thought it was a bit difficult to keep track of all of the different "store" variables through all of the wiring being done. I thought helped to isolate some things so the reader doesn't have to keep track of everything all at once.

prateekm: Yeah, agree that the new code is significantly more complex and we should make it more readable. But I think we need to go further than this and fix the store wrapping pattern and make it consistent. This PR introduces a new `toBeXXX` variable naming convention + var reassignments, which hurt readability a lot more. See also the boolean parameter on the other extracted method.

E.g., the call site for this vs the earlier version will be:
```
createLargeMessageSafeStore(underlyingStore, storeName, dropLargeMessage, maxMessageSize)
new LargeMessageSafeStore(underlyingStore, storeName, dropLargeMessage, maxMessageSize)
```
Which is basically the same from a readability perspective.

cameronlee314: Ah, that's a good point for the `createLargeMessageSafeStore` case. I guess I was thinking of some of the other helpers.

samza-kv/src/main/java/org/apache/samza/storage/kv/LargeMessageSafeStore.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @Override
+  public byte[] get(byte[] key) {
+    return store.get(key);
+  }
+
+  @Override
+  public void put(byte[] key, byte[] value) {
+    validateMessageSize(value);
+    if (!isLargeMessage(value)) {
```
prateekm: Need a metric for ignored large messages rate. Should also log them, not ignore them silently.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1687,6 +1687,50 @@ <h1>Samza Configuration Reference</h1>
                     </td>
                 </tr>
 
+                <tr>
```
prateekm: This is the old configuration reference, please update the new samza-configurations.md instead.

samza-kv/src/main/java/org/apache/samza/storage/kv/LargeMessageSafeStore.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @Override
+  public void flush() {
+    store.flush();
+  }
+
+  private void validateMessageSize(byte[] message) {
+    if (!dropLargeMessage && isLargeMessage(message)) {
+      throw new SamzaException("Logged store message size " + message.length + " for store " + storeName
```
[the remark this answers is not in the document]
prateekm: Throw a new specific exception that users can try-catch on instead of the generic SamzaException. SamzaException can come from many places and is not always safe to ignore.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1687,6 +1687,50 @@ <h1>Samza Configuration Reference</h1>
...
+                    <td class="property" id="stores-changelog-max-message-size-bytes">stores.<span class="store">store-name</span>.changelog.max.message.size.bytes</td>
+                    <td class="default">1000000</td>
+                    <td class="description">
+                        This property sets the maximum size of the messages allowed in the changelog.
+                        The default value is 1 MB.
+                    </td>
+                </tr>
+
+                <tr>
+                    <td class="property" id="stores-expect-large-messages">stores.<span class="store">store-name</span>.expect.large.messages</td>
```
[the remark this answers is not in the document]
PawasChhokra: Changed to disallow.large.messages.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1687,6 +1687,50 @@ <h1>Samza Configuration Reference</h1>
...
+                    <td class="default">false</td>
+                    <td class="description">
+                        This property, when turned on, tells the system to expect large messages to be put in the stores.
+                        It will then look out for any large messages greater than
+                        <a href="#stores-changelog-max-message-size-bytes" class="property">stores.*.changelog.max.message.size.bytes</a>
+                        and throw a SamzaException when it finds one, stating that the record is too large.
+                        In the case of using CachedStore, it will serialize the message first, validate
+                        its size and then cache it if the size is of permissible limit.
```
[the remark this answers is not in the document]
PawasChhokra: Done.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1687,6 +1687,50 @@ <h1>Samza Configuration Reference</h1>
                     </td>
                 </tr>
 
+                <tr>
+                    <td class="property" id="stores-changelog-max-message-size-bytes">stores.<span class="store">store-name</span>.changelog.max.message.size.bytes</td>
+                    <td class="default">1000000</td>
```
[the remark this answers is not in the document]
PawasChhokra: Changed to exact 1 MB.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
PawasChhokra: No. Removed.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1687,6 +1687,50 @@ <h1>Samza Configuration Reference</h1>
...
+                    <td class="property" id="stores-expect-large-messages">stores.<span class="store">store-name</span>.expect.large.messages</td>
+                    <td class="default">false</td>
+                    <td class="description">
+                        This property, when turned on, tells the system to expect large messages to be put in the stores.
+                        It will then look out for any large messages greater than
+                        <a href="#stores-changelog-max-message-size-bytes" class="property">stores.*.changelog.max.message.size.bytes</a>
+                        and throw a SamzaException when it finds one, stating that the record is too large.
+                        In the case of using CachedStore, it will serialize the message first, validate
+                        its size and then cache it if the size is of permissible limit.
+                        This particular case of using CachedStore causes a performance degradation since
```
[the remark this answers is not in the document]
PawasChhokra: Changed.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1687,6 +1687,50 @@ <h1>Samza Configuration Reference</h1>
...
+                        we end up serializing every time before putting the values in the cache.
+                        When this property is turned on, we ignore the value of
```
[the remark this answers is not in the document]
PawasChhokra: Done.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -124,22 +129,39 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
...
-    // wrap with serialization
-    val serializedMetrics = new SerializedKeyValueStoreMetrics(storeName, registry)
-    val serialized = new SerializedKeyValueStore[K, V](maybeLoggedStore, keySerde, msgSerde, serializedMetrics)
+    var toBeAccessLoggedStore: KeyValueStore[K, V] = null
+
+    if (largeMessagesExpected) {
+      // Execute "if" when we expect to get large messages and user defined config -> EXPECT_LARGE_MESSAGES = true, no matter the value of DROP_LARGE_MESSAGES
+      // This code path will throw a SamzaException for large message size.
+
+      // maybe wrap with caching
```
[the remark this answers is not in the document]
PawasChhokra: Removed.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
PawasChhokra: Done.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
PawasChhokra: Done.

samza-kv/src/main/scala/org/apache/samza/storage/kv/LoggedStore.scala
PawasChhokra: These changes were done by IntelliJ by default hence not reverted back.

docs/learn/documentation/versioned/jobs/configuration-table.html
PawasChhokra: Done.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1687,6 +1687,50 @@ <h1>Samza Configuration Reference</h1>
...
+                        <a href="#stores-drop-large-messages" class="property">stores.*.drop.large.messages</a>.
+                        The default value for this config is false. When this property is not set,
```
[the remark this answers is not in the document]
PawasChhokra: Done.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1687,6 +1687,50 @@ <h1>Samza Configuration Reference</h1>
...
+                        <a href="#stores-drop-large-messages" class="property">stores.*.drop.large.messages</a>
+                        defines the behaviour to be executed.
+                    </td>
+                </tr>
+
+                <tr>
+                    <td class="property" id="stores-drop-large-messages">stores.<span class="store">store-name</span>.drop.large.messages</td>
+                    <td class="default">false</td>
+                    <td class="description">
+                        This property, when turned on, tells the system to drop any large messages instead of
```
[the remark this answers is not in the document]
PawasChhokra: Done.

docs/learn/documentation/versioned/jobs/configuration-table.html
```diff
@@ -1687,6 +1687,50 @@ <h1>Samza Configuration Reference</h1>
...
+                        attempting to put them in the stores. It drops all large messages greater than
+                        <a href="#stores-changelog-max-message-size-bytes" class="property">stores.*.changelog.max.message.size.bytes</a>
+                        and continues to function as if it did not receive any large messages.
+                        No exception is thrown. In the case of using CachedStore, when this config is
+                        turned on, the large message is stored in the Cache but is not written to the
+                        changelog and underlying store, resulting in an inconsistent state temporarily.
+                        When this property is turned off, the system will not be able to handle large messages of any type.
```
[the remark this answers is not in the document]
PawasChhokra: Done.

samza-kv/src/main/java/org/apache/samza/storage/kv/LargeMessageSafeStore.java
```diff
@@ -0,0 +1,117 @@
...
+
+
+public class LargeMessageSafeStore implements KeyValueStore<byte[], byte[]> {
+
+  private KeyValueStore<byte[], byte[]> store;
+  private String storeName;
+  private Boolean dropLargeMessage;
+  private int maxMessageSize;
+
+  public LargeMessageSafeStore(KeyValueStore<byte[], byte[]> store, String storeName, Boolean dropLargeMessage, int maxMessageSize) {
```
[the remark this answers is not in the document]
PawasChhokra: Done.

samza-kv/src/main/java/org/apache/samza/storage/kv/LargeMessageSafeStore.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.samza.SamzaException;
+
+
+public class LargeMessageSafeStore implements KeyValueStore<byte[], byte[]> {
+
+  private KeyValueStore<byte[], byte[]> store;
+  private String storeName;
+  private Boolean dropLargeMessage;
```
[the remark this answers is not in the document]
PawasChhokra: Done.

samza-kv/src/main/java/org/apache/samza/storage/kv/LargeMessageSafeStore.java
```diff
@@ -0,0 +1,117 @@
...
+package org.apache.samza.storage.kv;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.samza.SamzaException;
+
+
+public class LargeMessageSafeStore implements KeyValueStore<byte[], byte[]> {
+
+  private KeyValueStore<byte[], byte[]> store;
```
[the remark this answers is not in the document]
PawasChhokra: Done.

samza-kv/src/main/java/org/apache/samza/storage/kv/LargeMessageSafeStore.java
```diff
@@ -0,0 +1,117 @@
...
+      throw new SamzaException("Logged store message size " + message.length + " for store " + storeName
+          + " was larger than the maximum allowed message size " + maxMessageSize + ".");
+    }
+  }
+
+  private boolean isLargeMessage(byte[] message) {
+    return message != null && message.length > maxMessageSize;
+  }
+
+  private List<Entry<byte[], byte[]>> removeLargeMessage(List<Entry<byte[], byte[]>> entries) {
```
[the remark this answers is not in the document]
PawasChhokra: Done.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
PawasChhokra: Done.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -124,22 +129,39 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
...
+      var maybeCachedStore = createCachedStore(enableCache, storeName, registry, maybeLoggedStore, cacheSize, batchSize)
```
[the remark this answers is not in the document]
PawasChhokra: Changed to val.

samza-kv/src/main/java/org/apache/samza/storage/kv/LargeMessageSafeStore.java
```diff
@@ -0,0 +1,117 @@
...
+ * under the License.
+ */
+package org.apache.samza.storage.kv;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.samza.SamzaException;
+
+
+public class LargeMessageSafeStore implements KeyValueStore<byte[], byte[]> {
```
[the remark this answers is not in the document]
PawasChhokra: Done.

samza-kv/src/main/java/org/apache/samza/storage/kv/LargeMessageSafeStore.java
```diff
@@ -0,0 +1,117 @@
...
+    this.maxMessageSize = maxMessageSize;
+  }
+
+  @Override
+  public byte[] get(byte[] key) {
+    return store.get(key);
+  }
+
+  @Override
+  public void put(byte[] key, byte[] value) {
```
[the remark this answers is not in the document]
PawasChhokra: Done.

samza-kv/src/main/java/org/apache/samza/storage/kv/LargeMessageSafeStore.java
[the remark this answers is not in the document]
PawasChhokra: Removed the reference to LoggedStore and added a new RecordTooLargeException.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -124,22 +129,39 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
       new LoggedStore(rawStore, changeLogSystemStreamPartition, collector, loggedStoreMetrics)
     }
 
-    // wrap with serialization
-    val serializedMetrics = new SerializedKeyValueStoreMetrics(storeName, registry)
-    val serialized = new SerializedKeyValueStore[K, V](maybeLoggedStore, keySerde, msgSerde, serializedMetrics)
+    var toBeAccessLoggedStore: KeyValueStore[K, V] = null
+
+    if (largeMessagesExpected) {
+      // Execute "if" when we expect to get large messages and user defined config -> EXPECT_LARGE_MESSAGES = true, no matter the value of DROP_LARGE_MESSAGES
```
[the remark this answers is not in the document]
PawasChhokra: Done.

samza-kv/src/main/java/org/apache/samza/storage/kv/LargeMessageSafeStore.java
PawasChhokra: Added a metric and logs.

samza-kv/src/main/scala/org/apache/samza/storage/kv/LargeMessageSafeStoreMetrics.scala
```diff
@@ -0,0 +1,30 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.storage.kv
+
+import org.apache.samza.metrics.{MetricsHelper, MetricsRegistry, MetricsRegistryMap}
+
+class LargeMessageSafeStoreMetrics(
+  val storeName: String = "unknown",
```
cameronlee314: Maybe you can just not use default values for the constructor args here.
Default values make it harder to ensure that you passed all of the desired constructor arguments, since the compiler doesn't force you to specify them.

samza-kv/src/test/java/org/apache/samza/storage/kv/TestLargeMessageSafeStore.java
```diff
@@ -0,0 +1,180 @@
...
+    byte[] smallMessage = new byte[32];
+    largeMessageSafeKeyValueStore.put(key1, smallMessage);
+    Mockito.verify(store).put(Matchers.eq(key1), Matchers.eq(smallMessage));
+
+    byte[] key2 = new byte[16];
+    byte[] largeMessage = new byte[maxMessageSize + 1];
+    try {
+      largeMessageSafeKeyValueStore.put(key2, largeMessage);
+      Assert.fail("The test case should have failed due to a large message being passed to the changelog, but it didn't.");
+    } catch (SamzaException e) {
```
cameronlee314: Can you check for `RecordTooLargeException` instead of `SamzaException`?

samza-kv/src/test/java/org/apache/samza/storage/kv/TestLargeMessageSafeStore.java
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  @Test
+  public void testLargeMessagePutWithDropLargeMessageDisabled() {
+    LargeMessageSafeStore largeMessageSafeKeyValueStore = new LargeMessageSafeStore(store, storeName, false, maxMessageSize);
+
+    byte[] key1 = new byte[16];
+    byte[] smallMessage = new byte[32];
+    largeMessageSafeKeyValueStore.put(key1, smallMessage);
+    Mockito.verify(store).put(Matchers.eq(key1), Matchers.eq(smallMessage));
```
cameronlee314: Can you please pull this out to a separate test case? It's easier to tell that you have a test for a small message that way.

samza-kv/src/test/java/org/apache/samza/storage/kv/TestLargeMessageSafeStore.java
```diff
@@ -0,0 +1,180 @@
...
+    try {
+      largeMessageSafeKeyValueStore.putAll(entries);
+      Assert.fail("The test case should have failed due to a large message being passed to the changelog, but it didn't.");
+    } catch (SamzaException e) {
+      Mockito.verifyZeroInteractions(store);
+    }
+  }
+
+  @Test
+  public void testSmallMessagePutWithSerdeAndDropLargeMessageDisabled() {
```
cameronlee314: It might not be necessary to have these extra "with serde" tests. The interface for the `LargeMessageSafeStore` is already bytes, and you already tested that case above.

samza-kv/src/test/java/org/apache/samza/storage/kv/TestLargeMessageSafeStore.java
```diff
@@ -0,0 +1,180 @@
...
+
+    try {
+      largeMessageSafeKeyValueStore.putAll(entries);
+    } catch (SamzaException e) {
+      Mockito.verifyZeroInteractions(store);
+    }
+  }
+
+  @Test
+  public void testPutSuccessWithDropLargeMessageEnabled() {
```
cameronlee314: Consider also adding a test case where the message is small and drop large message is enabled, making sure that the message is put into the store.

samza-test/src/test/java/org/apache/samza/storage/kv/TestLargeMessageSafeKeyValueStores.java
```diff
@@ -0,0 +1,240 @@
...
+    }
+    if (dir != null && dir.listFiles() != null) {
+      for (File file : dir.listFiles())
+        file.delete();
+      dir.delete();
+    }
+  }
+
+  @Test
+  public void testLargeMessagePutFailureForLoggedStoreWithWrappedStore() {
```
cameronlee314: This test method (and the test methods below) seem to be a bit misleading. For example, this says "put failure", but if "drop large message" is enabled, then there is no expected failure. Also, it isn't really useful to have the "ForLoggedStoreWithWrappedStore" suffix, since not all executions of this test will do that.
Please try to rename them to clarify what they are testing in all cases. For example, this could just be called "testPutLargeMessage".

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -46,6 +46,12 @@
   public static final String CHANGELOG_STREAM = STORE_PREFIX + "%s" + CHANGELOG_SUFFIX;
   public static final String ACCESSLOG_STREAM_SUFFIX = "access-log";
   public static final String CHANGELOG_REPLICATION_FACTOR = STORE_PREFIX + "%s.changelog.replication.factor";
+  public static final String CHANGELOG_MAX_MSG_SIZE_BYTES = STORE_PREFIX + "%s.changelog.max.message.size.bytes";
+  public static final int DEFAULT_CHANGELOG_MAX_MSG_SIZE_BYTES = 1000000; // slightly less than 1 MB
```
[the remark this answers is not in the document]
PawasChhokra: Done. Thanks.

samza-kv/src/main/java/org/apache/samza/storage/kv/LargeMessageSafeStore.java
```diff
@@ -0,0 +1,150 @@
...
+public class LargeMessageSafeStore implements KeyValueStore<byte[], byte[]> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(LargeMessageSafeStore.class);
+  private final KeyValueStore<byte[], byte[]> store;
+  private final String storeName;
+  private final boolean dropLargeMessages;
+  private final int maxMessageSize;
+  private final LargeMessageSafeStoreMetrics largeMessageSafeStoreMetrics;
+
+  public LargeMessageSafeStore(KeyValueStore<byte[], byte[]> store, String storeName, Boolean dropLargeMessages, int maxMessageSize) {
```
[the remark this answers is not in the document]
PawasChhokra: Done.

samza-kv/src/main/scala/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.scala
```diff
@@ -124,22 +129,46 @@ trait BaseKeyValueStorageEngineFactory[K, V] extends StorageEngineFactory[K, V]
...
+
+      // wrap with large message checking
+      val largeMessageSafeKeyValueStore = new LargeMessageSafeStore(maybeCachedStore, storeName, false, maxMessageSize)
+      // wrap with serialization
+      val serializedMetrics = new SerializedKeyValueStoreMetrics(storeName, registry)
+      toBeAccessLoggedStore = new SerializedKeyValueStore[K, V](largeMessageSafeKeyValueStore, keySerde, msgSerde, serializedMetrics)
+
+    }
+    else {
+      var toBeSerializedStore = maybeLoggedStore
```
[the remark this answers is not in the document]
PawasChhokra: Done.

samza-kv/src/main/scala/org/apache/samza/storage/kv/LargeMessageSafeStoreMetrics.scala
PawasChhokra: Removed the usage of default values.

samza-kv/src/test/java/org/apache/samza/storage/kv/TestLargeMessageSafeStore.java
PawasChhokra: Done.

samza-kv/src/test/java/org/apache/samza/storage/kv/TestLargeMessageSafeStore.java
PawasChhokra: Done.

samza-kv/src/test/java/org/apache/samza/storage/kv/TestLargeMessageSafeStore.java
PawasChhokra: Done.

samza-test/src/test/java/org/apache/samza/storage/kv/TestLargeMessageSafeKeyValueStores.java
PawasChhokra: Renamed these methods.

samza-kv/src/test/java/org/apache/samza/storage/kv/TestLargeMessageSafeStore.java
PawasChhokra: Rayman had asked me to add these previously, hence I haven't removed them. Let me know if they absolutely need to be removed.

cameronlee314: You can leave them in if you believe they are useful.
Just keep in mind in the future that unit tests should test against the API of the class. So if two tests are testing the same flow, then one of them might not be necessary. Every test does have to be maintained, so there is some cost in having more tests.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -47,7 +47,7 @@
   public static final String ACCESSLOG_STREAM_SUFFIX = "access-log";
   public static final String CHANGELOG_REPLICATION_FACTOR = STORE_PREFIX + "%s.changelog.replication.factor";
   public static final String CHANGELOG_MAX_MSG_SIZE_BYTES = STORE_PREFIX + "%s.changelog.max.message.size.bytes";
-  public static final int DEFAULT_CHANGELOG_MAX_MSG_SIZE_BYTES = 1000000; // slightly less than 1 MB
+  public static final int DEFAULT_CHANGELOG_MAX_MSG_SIZE_BYTES = 1048576; // slightly less than 1 MB
```
cameronlee314: Comment is no longer consistent with value

PawasChhokra: Removed the comment.

samza-kv/src/test/java/org/apache/samza/storage/kv/TestLargeMessageSafeStore.java
PawasChhokra: Got it. Leaving them as it is and will keep in mind for the future.

PR #1010, opened 2019-04-26
SAMZA-2176: Ignore the configuration keys with serialized null values from the coordinator stream.

shanthoosh:
Samza stores the metadata of the job into a log-compacted kafka topic called coordinator stream.

Due to some operational issue, we converted serialized null value into bytes and stored them as values for some keys in the coordinator stream. These keys will not be log-compacted by kafka, since they're not actually null messages. 

This patch ignores the configurations from coordinator stream with serialized-null values and buffers only keys  with non-null values in memory in ApplicationMaster.

Verified the changes in this patch by testing with a samza-hello-samza job by injecting samza-reserved configuration keys with serialized null values into the coordinator stream.  Here is the entire local [testing](https://gist.github.com/shanthoosh/88ef289cb98d6d3be9db9fb76bc04472#file-test-deserializing-serialized-null-values-in-coordinator-stream) log indicating successful verification.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/util/CoordinatorStreamUtil.scala
+++ b/samza-core/src/main/scala/org/apache/samza/util/CoordinatorStreamUtil.scala
@@ -22,6 +22,7 @@
 package org.apache.samza.util
 
 import java.util
+import org.apache.commons.lang3.StringUtils
 import org.apache.samza.SamzaException
 import org.apache.samza.config._
 import org.apache.samza.system.{SystemFactory, SystemStream}
@@ -34,7 +35,7 @@ import org.apache.samza.util.ScalaJavaUtil.JavaOptionals
 import scala.collection.immutable.Map
 import scala.collection.JavaConverters._
 
-object CoordinatorStreamUtil {
+object CoordinatorStreamUtil extends Logging {
   /**
     * Given a job's full config object, build a subset config which includes
     * only the job name, job id, and system config for the coordinator stream.
@@ -106,9 +107,17 @@ object CoordinatorStreamUtil {
     val configFromCoordinatorStream: util.Map[String, Array[Byte]] = namespaceAwareCoordinatorStreamStore.all
     val configMap: util.Map[String, String] = new util.HashMap[String, String]
     for ((key: String, valueAsBytes: Array[Byte]) <- configFromCoordinatorStream.asScala) {
-      val valueSerde: CoordinatorStreamValueSerde = new CoordinatorStreamValueSerde(SetConfig.TYPE)
-      val valueAsString: String = valueSerde.fromBytes(valueAsBytes)
-      configMap.put(key, valueAsString)
+      if (valueAsBytes == null) {
+        warn("Value for key: %s in config is null. Ignoring it." format key)
+      } else {
+        val valueSerde: CoordinatorStreamValueSerde = new CoordinatorStreamValueSerde(SetConfig.TYPE)
+        val valueAsString: String = valueSerde.fromBytes(valueAsBytes)
+        if (StringUtils.isBlank(valueAsString)) {
+          warn("Value for key: %s in config is empty or null. Ignoring it." format key)
+        } else {
+          configMap.put(key, valueAsString)
+        }
+      }
     }
     new MapConfig(configMap)
   }


shanthoosh:
@cameronlee314 
In addition to introducing non-null checks for configuration, I also added non-null checks for all types of metadata in different utility class classes viz TaskAssignmentManager, `LocalityManager`, `TaskPartitionAssignmentManager` etc. 

In the future, when someone by mistake writes serialized null(random-empty) for these other message types in coordinator stream, these null-checks would guard the application from failing. Please take a look when you get a chance.

samza-core/src/main/scala/org/apache/samza/util/CoordinatorStreamUtil.scala
```diff
@@ -106,9 +107,17 @@ object CoordinatorStreamUtil {
     val configFromCoordinatorStream: util.Map[String, Array[Byte]] = namespaceAwareCoordinatorStreamStore.all
     val configMap: util.Map[String, String] = new util.HashMap[String, String]
     for ((key: String, valueAsBytes: Array[Byte]) <- configFromCoordinatorStream.asScala) {
-      val valueSerde: CoordinatorStreamValueSerde = new CoordinatorStreamValueSerde(SetConfig.TYPE)
-      val valueAsString: String = valueSerde.fromBytes(valueAsBytes)
-      configMap.put(key, valueAsString)
+      if (valueAsBytes == null) {
+        info("Value for key: %s is null. Ignoring it." format key)
```
rmatharu: Could this be logged as a warning, because a value for a config-key ideally should not be null.

samza-core/src/main/java/org/apache/samza/container/LocalityManager.java
```diff
@@ -63,9 +64,11 @@ public LocalityManager(MetadataStore metadataStore) {
     metadataStore.all().forEach((containerId, valueBytes) -> {
         if (valueBytes != null) {
           String locationId = valueSerde.fromBytes(valueBytes);
-          Map<String, String> values = new HashMap<>();
-          values.put(SetContainerHostMapping.HOST_KEY, locationId);
-          allMappings.put(containerId, values);
+          if (StringUtils.isNotBlank(locationId)) {
```
rmatharu: It appears that we're invoking StringUtils.isNotBlank immediately after <serde>.fromBytes(). 
Is it possible to create a wrapper for serde and call this method there. 

Then you check like so:
`
String locationId = new MySerdeWrapper(valueSerde).fromBytes(valueBytes);

public class MySerdeWrapper {
  public Deserializer deserializer;

  public MySerdeWrapper(Deserializer deserializer) {
    this.deserializer = deserializer;
  }

  public Optional<String> fromBytes(byte[] bytes) {
    return StringUtils.isNotBlank(deserializer.fromBytes(bytes)) ? Optional.of(deserializer.fromBytes(bytes)) : Optional.empty();
  }


}`

shanthoosh: I thought of doing this change before sending this patch out for review.  Since this is a critical bug-fix, I didn't want to couple it with a refactor/cleanup. If it's acceptable then I will do it in a follow-up [JIRA](https://issues.apache.org/jira/browse/SAMZA-2177).

samza-core/src/main/scala/org/apache/samza/util/CoordinatorStreamUtil.scala
shanthoosh: Done.

cameronlee314:
For the non-config cases, did the old 1.0 logic do the null checks?

shanthoosh:
@cameronlee314 

For non-config cases, old 1.0 logic did not do the null checks. 

This whole PR was necessary because coordinator stream was corrupted with serialized null messages for config-type messages. The same serialized-null messages can be published to coordinator stream for any type of keys(container-host, task-partition) in the future.

I'm adding this null-checks as a safe-guard to prevent against it.

cameronlee314:
> @cameronlee314
> 
> For non-config cases, old 1.0 logic did not do the null checks.
> 
> This whole PR was necessary because coordinator stream was corrupted with serialized null values for config-type keys. The same serialized-null values can be published to coordinator stream for any type of keys(container-host, task-partition) in the future.
> 
> I'm adding this null-checks as a safe-guard to prevent against it.

I suggest doing these two changes in separate PRs. First PR to just fix the config case to bring it back to Samza 1.0 functionality. Second PR to add the safe guards. Then we can manage the changes separately (e.g. if it turns out that the extra non-config safe guards are undesirable, we don't want to roll back the config changes)

shanthoosh:
>> if it turns out that the extra non-config safe guards are undesirable, we don't want to roll back the config changes

Removed the null-checks for non-config key types from this PR.

PR #1011, opened 2019-04-26
SAMZA-2172: Make High Level applications respect job.container.thread.pool.size

mynameborat:
Refer https://issues.apache.org/jira/browse/SAMZA-2172 for more details.

`2019-04-25 14:41:17.132 [Samza Container Thread-2] WikipediaAsyncApplication [INFO] Executing blocking filter on the thread Samza Container Thread-2
2019-04-25 14:41:17.133 [Samza Container Thread-2] WikipediaAsyncApplication [INFO] Finished executing blocking filter on the thread Samza Container Thread-2
2019-04-25 14:41:17.442 [Samza Container Thread-3] WikipediaAsyncApplication [INFO] Executing blocking filter on the thread Samza Container Thread-3
2019-04-25 14:41:17.442 [Samza Container Thread-3] WikipediaAsyncApplication [INFO] Finished executing blocking filter on the thread Samza Container Thread-3
2019-04-25 14:41:17.487 [Samza Container Thread-0] WikipediaAsyncApplication [INFO] Executing blocking filter on the thread Samza Container Thread-0
`

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/task/StreamOperatorTaskFactory.java
+++ b/samza-api/src/main/java/org/apache/samza/task/StreamOperatorTaskFactory.java
@@ -1,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.task;
+
+import org.apache.samza.annotation.InterfaceStability;
+
+
+/**
+ * Build {@link AsyncStreamTask} instances.
+ * <p>
+ * Implementations should return a new instance of {@link AsyncStreamTask} for each {@link #createInstance()} invocation.
+ */
+@InterfaceStability.Evolving
+public interface StreamOperatorTaskFactory extends TaskFactory<AsyncStreamTask> {
+}
--- a/samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
+++ b/samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
@@ -21,6 +21,7 @@ package org.apache.samza.task;
 import com.google.common.base.Preconditions;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ExecutorService;
 import org.apache.samza.SamzaException;
 import org.apache.samza.context.Context;
 import org.apache.samza.operators.OperatorSpecGraph;
@@ -47,6 +48,14 @@ public class StreamOperatorTask implements AsyncStreamTask, InitableTask, Window
   private final OperatorSpecGraph specGraph;
   private final Clock clock;
 
+  /*
+   * Thread pool used by the task to schedule processing of incoming messages. If job.container.thread.pool.size is
+   * not configured, this will be null. We don't want to create an executor service within StreamOperatorTask due to
+   * following reasons
+   *   1. It is harder to reason about the lifecycle of the executor service
+   *   2. We end up with thread pool proliferation. Especially for jobs with high number of tasks.
+   */
+  private ExecutorService taskThreadPool;
   private OperatorImplGraph operatorImplGraph;
 
   /**
@@ -86,7 +95,9 @@ public class StreamOperatorTask implements AsyncStreamTask, InitableTask, Window
 
   /**
    * Passes the incoming message envelopes along to the {@link InputOperatorImpl} node
-   * for the input {@link SystemStream}.
+   * for the input {@link SystemStream}. It is non-blocking and dispatches the message to an internally managed thread
+   * pool. The thread pool size is configured through job.container.thread.pool.size. In the absence of the config,
+   * the task defaults to a single thread executor.
    * <p>
    * From then on, each {@link org.apache.samza.operators.impl.OperatorImpl} propagates its transformed output to
    * its chained {@link org.apache.samza.operators.impl.OperatorImpl}s itself.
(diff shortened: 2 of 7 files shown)


mynameborat:
@prateekm @xinyuiscool @vjagadish1989 please take a look

samza-core/src/main/java/org/apache/samza/task/RunLoop.java
```diff
@@ -457,17 +457,25 @@ private void process() {
...
+            containerMetrics.processes().inc();
+            return callbackManager.createCallback(task.taskName(), envelope, coordinator);
+          }
+        };
+
+        task.process(envelope, coordinator, callbackFactory);
       };
 
-      task.process(envelope, coordinator, callbackFactory);
+      if (threadPool != null) {
```
rmatharu: This is a bad pattern in code. Can we fix atleast this instance of it. 
This threadPool is being passed from SamzaContainer as a null. 

```val taskThreadPool = if (threadPoolSize > 0) {
      Executors.newFixedThreadPool(threadPoolSize,
        new ThreadFactoryBuilder().setNameFormat("Samza Container Thread-%d").build())
    } else {
      null
    }```
a. We can either push the thread-pool creation down to the place its actually being used. 
b. Alternatively, can create it as a single-threaded executor pool.
c. or pass it as an optional.

Comment (mynameborat) [reply samza-core/src/main/java/org/apache/samza/task/RunLoop.java]: I am all in for this. However, I'd prefer to defer it to a follow up PR to clean up the semantics of threadPool and update the RunLoop code all at once instead of fixing this place alone. 
In my opinion, it is going to add more noise to the PR and makes it hard to debug issues associated with the PR. 
what do you think?

File: samza-core/src/main/java/org/apache/samza/task/AsyncStreamTaskAdapter.java [hunk 2]
```diff
@@ -19,24 +19,20 @@
...
 
 
 /**
- * AsyncStreamTaskAdapter allows a StreamTask to be executed in parallel.The class
- * uses the build-in thread pool to invoke StreamTask.process and triggers
- * the callbacks once it's done. If the thread pool is null, it follows the legacy
- * synchronous model to execute the tasks on the run loop thread.
+ * AsyncStreamTaskAdapter adapts a StreamTask to be a AsyncStreamTask.
+ * If job.container.thread.pool.size &gt; 1 the framework executes the adapted task in parallel.
+ * Otherwise, it follows the legacy synchronous model to execute the tasks on the run loop thread.
```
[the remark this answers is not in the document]
mynameborat: I updated the comment. It uses the thread pool as long as `job.container.thread.pool.size` is specified.
Only in the absence of the configuration, it falls back to legacy synchronous model using the run loop thread.

samza-core/src/main/java/org/apache/samza/task/RunLoop.java
```diff
@@ -457,17 +457,25 @@ private void process() {
...
-      final ReadableCoordinator coordinator = new ReadableCoordinator(task.taskName());
-      TaskCallbackFactory callbackFactory = new TaskCallbackFactory() {
-        @Override
-        public TaskCallback createCallback() {
-          state.startProcess();
-          containerMetrics.processes().inc();
-          return callbackManager.createCallback(task.taskName(), envelope, coordinator);
-        }
+      Runnable processWorker = () -> {
+        final ReadableCoordinator coordinator = new ReadableCoordinator(task.taskName());
```
shanthoosh: Just curious. 

Previously, we used to have in-order delivery of messages to the task implementation(where max-in-flight messages controlled by `task.max.concurrency` configuration). This behavior was possible since the delivery/dispatch of messages to task implementation would happen only from the event-loop. 

Now, we enqueue process-tasks into the `ExecutorService` from event-loop. Tasks that are submitted to `ExecutorService` can be executed in any order(Interface contract does not guarantee in-order execution of submitted tasks). This could potentially deliver/dispatch messages out of order to the task implementation(i.e message with offset 2 can be delivered before message with offset 1). 

Can you please confirm if the above message-dispatch behavior to task is plausible?

samza-core/src/main/java/org/apache/samza/task/RunLoop.java
```diff
@@ -457,17 +457,25 @@ private void process() {
...
+        TaskCallbackFactory callbackFactory = new TaskCallbackFactory() {
+          @Override
+          public TaskCallback createCallback() {
+            state.startProcess();
```
shanthoosh: Would be better to add a test-case to ensure if `job.container.threadpool.size` configuration is defined, then the process-tasks are submitted to executor service.

What do you think?

samza-core/src/main/java/org/apache/samza/task/RunLoop.java
```diff
@@ -457,17 +457,25 @@ private void process() {
...
+            containerMetrics.processes().inc();
+            return callbackManager.createCallback(task.taskName(), envelope, coordinator);
+          }
```
shanthoosh: This is not related to your change. Currently the life-cycle of the executor service is controlled two levels up at the `SamzaContainer` layer, i.e, the executor service is stopped as a part of the `SamzaContainer` shutdown sequence. 

This makes reasoning about life-cycle of the executor service harder. It would be better to manage the life-cycle entirely within the RunLoop and stop it as a part of `RunLoop` shutdown sequence.

What do you think?

samza-api/src/main/java/org/apache/samza/task/StreamOperatorTaskFactory.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.samza.annotation.InterfaceStability;
+
+
+/**
+ * Build {@link AsyncStreamTask} instances.
+ * <p>
+ * Implementations should return a new instance of {@link AsyncStreamTask} for each {@link #createInstance()} invocation.
+ */
+@InterfaceStability.Evolving
+public interface StreamOperatorTaskFactory extends TaskFactory<AsyncStreamTask> {
```
prateekm: Does this need to be in samza-api? Seems like it should be possible to do this in core, and I don't think we need to expose this to users.

samza-api/src/main/java/org/apache/samza/task/StreamOperatorTaskFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.samza.task;
+
+import org.apache.samza.annotation.InterfaceStability;
+
+
+/**
+ * Build {@link AsyncStreamTask} instances.
```
prateekm: Update javadocs.

samza-core/src/main/java/org/apache/samza/task/StreamOperatorTask.java
```diff
@@ -86,7 +95,9 @@ public final void init(Context context) throws Exception {
 
   /**
    * Passes the incoming message envelopes along to the {@link InputOperatorImpl} node
-   * for the input {@link SystemStream}.
+   * for the input {@link SystemStream}. It is non-blocking and dispatches the message to an internally managed thread
```
prateekm: s/an internally managed/the container.

samza-core/src/main/java/org/apache/samza/task/TaskFactoryUtil.java
```diff
@@ -112,6 +112,24 @@ public static TaskFactory finalizeTaskFactory(TaskFactory factory, ExecutorServi
...
+
+    /*
+     * Note: Even though StreamOperatorTask is an instanceof AsyncStreamTask, we still need to
+     * adapt it in order to inject the container thread pool. Long term, this will go away once we have the
+     * InternalTaskContext refactor, which would then become the entry point for exposing any of the runtime objects
+     * created in the container.
+     * Refer to SAMZA-2203 for more details.
+     */
+    if (isStreamOperatorTaskClass) {
+      log.info("Adapting StreamOperatorTask to AsyncStreamTaskAdapter");
```
prateekm: Update log message (not adapting to AsyncStreamTaskAdapter)

samza-api/src/main/java/org/apache/samza/task/StreamOperatorTaskFactory.java
mynameborat: Agreed. This is will go away with the InternalTaskContext refactor.
Moved it to samza-core.

samza-api/src/main/java/org/apache/samza/task/StreamOperatorTaskFactory.java
mynameborat: I initially had it only return AsyncStreamTask since StreamOperatorTask is not part of samza-api. Now that we have moved it to samza-core. Updated it to return StreamOperatorTask

samza-core/src/main/java/org/apache/samza/task/RunLoop.java
mynameborat: Not applicable anymore.

samza-core/src/main/java/org/apache/samza/task/RunLoop.java
mynameborat: Not applicable anymore

mynameborat:
@prateekm Addressed all the comments and updated the JIRA and chosen solution.

PR #1013, opened 2019-04-26
added table delete api

sborya:
Call sendto(key, null) to delete from a table by key.


prateekm:
@sborya Can you please update the title / description. Delete API already exists, and I don't see sendTo in the PR anywhere. What issue is this fixing?

PR #1014, opened 2019-04-27
Upgrading gradle to 4.8

rmatharu:


prateekm:
Do you need to update gradle.version in gradle.properties?

PR #1015, opened 2019-04-29
SAMZA-2178: Utils to directly inject custom IME to InMemorySystem streams via TestRunner

Sanil15:
Please read the jira tagged for description: https://issues.apache.org/jira/browse/SAMZA-2178

Use cases where this is helpful: 
Systems like Brooklin at LinkedIn where they have a custom BrooklinIncomingMessageEnvelope (BIMEs), this helps to directly feed BIMEs to InMemorySystem for Testing 

@bharathkk @prateekm please have a look when you have time!

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
+++ b/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
@@ -76,6 +76,19 @@ class InMemoryManager {
   }
 
   /**
+   * Handles produce request from {@link InMemorySystemProducer} for case where a job has a custom IME and
+   * populates the underlying message queue with the IME.
+   *
+   * @param ssp
+   * @param envelope
+   */
+  void put(SystemStreamPartition ssp, IncomingMessageEnvelope envelope) {
+    List<IncomingMessageEnvelope> messages = bufferedMessages.get(ssp);
+    bufferedMessages.get(ssp).add(envelope);
+  }
+
+
+  /**
    * Handles the poll request from {@link InMemorySystemConsumer}. It uses the input offset as the starting offset for
    * each {@link SystemStreamPartition}.
    *
--- a/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
+++ b/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
@@ -29,6 +29,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import org.apache.samza.Partition;
+import org.apache.samza.SamzaException;
 import org.apache.samza.system.EndOfStreamMessage;
 import org.apache.samza.system.IncomingMessageEnvelope;
 import org.apache.samza.system.StreamSpec;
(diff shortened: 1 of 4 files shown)


samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
```diff
@@ -75,6 +75,19 @@ void put(SystemStreamPartition ssp, Object key, Object message) {
         .add(messageEnvelope);
   }
 
+  /**
+   * Handles produce request from {@link InMemorySystemProducer} for case where a job has a custom IME and
+   * populates the underlying message queue with the IME.
+   *
+   * @param ssp
```
mynameborat: more docs for the parameter or remove them.

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemProducer.java
```diff
@@ -95,6 +96,15 @@ public void send(String source, OutgoingMessageEnvelope envelope) {
     memoryManager.put(ssp, key, message);
   }
 
+  /**
+   * Populates the IME to the ssp configured, this gives user more control to set up Test environment
+   *
+   * @param envelope
+   */
+  public void send(IncomingMessageEnvelope envelope) {
```
mynameborat: 1. More docs on the parameter or remove it.
2. Can we clarify that this is not intentional for production use?
3. `Populates the IME to the ssp configured` seems a little confusing. I'd explicitly call out that the method takes payloads wrapped in Samza internal representation and clients need to adhere to the semantics IME. e.g. populating the offset field and order in which put is invoked so that IMEs has monotonically increasing offsets.

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
```diff
@@ -75,6 +75,19 @@ void put(SystemStreamPartition ssp, Object key, Object message) {
...
+   * @param envelope
+   */
+  void put(SystemStreamPartition ssp, IncomingMessageEnvelope envelope) {
+    List<IncomingMessageEnvelope> messages = bufferedMessages.get(ssp);
+    bufferedMessages.get(ssp).add(envelope);
```
mynameborat: Need to ensure the offset inserted into the queue is increasing and equals to last message `offset + 1` otherwise, poll logic will be impacted

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemProducer.java
Sanil15: Added more docs, documented not to be used for production use cases, added additional checks to fail if offset policy does not adhere.

samza-test/src/test/java/org/apache/samza/test/framework/StreamTaskIntegrationTest.java
```diff
@@ -191,6 +193,32 @@ public void testSyncTaskWithMultiplePartitionMultithreaded() throws Exception {
     Map<Integer, List<KV>> inputPartitionData = new HashMap<>();
     Map<Integer, List<Integer>> expectedOutputPartitionData = new HashMap<>();
     genData(inputPartitionData, expectedOutputPartitionData);
+    syncTaskWithMultiplePartitionMultithreadedHelper(inputPartitionData, expectedOutputPartitionData);
+  }
+
+  @Test
+  public void testSyncTaskWithMultiplePartitionMultithreadedWithCutomIME() throws Exception {
```
mynameborat: nit: typo `custom`

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemProducer.java
mynameborat: Thanks! Can you also add it is not thread safe?

PR #1016, opened 2019-04-30
SAMZA-2179: Move the StartpointVisitor abstraction to SystemAdmin interface.

shanthoosh:

* Migrates the `StartpointVisitor` abstraction from `SystemConsumer` to `SystemAdmin`. Currently, startpoints are acted upon at the `SystemConsumer` layer. This patch enables resolving startpoint to the offset before starting off the downstream components of SamzaContainer. This provides easier integration path for Startpoint by not mandating that every component of SamzaContainer from OffsetManager, SystemConsumer, SystemConsumers etc to support startpoint semantics and avoids proliferation of Samza public API with startpoints.
* Moves the `KafkaStartpointVisitor` implementation from `KafkaSystemConsumer` to `KafkaSystemAdmin`. Reverts back the refactor done in `KafkaSystemConsumer` for startpoints. 
* Moved the startpoint tests from `KafkaSystemConsumer` to `KafkaSystemAdmin`.
* Removes the `StartpointCustom`, which is not required for supporting any type of seek-back for I/O connectors.


shanthoosh:
@dnishimura  

dnishimura:
+ @cameronlee314

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -167,12 +142,11 @@ public void start() {
...
   private void startSubscription() {
     //subscribe to all the registered TopicPartitions
-    Set<TopicPartition> registeredTopicPartitions = new HashSet<>(topicPartitionToStartpointMap.keySet());
-    LOG.info("{}: Consumer subscribes to {}", this, registeredTopicPartitions);
+    LOG.info("{}: Consumer subscribes to {}", this, topicPartitionsToSSP.keySet());
     try {
       synchronized (kafkaConsumer) {
         // we are using assign (and not subscribe), so we need to specify both topic and partition
-        kafkaConsumer.assign(registeredTopicPartitions);
+        kafkaConsumer.assign(topicPartitionsToSSP.keySet());
```
dnishimura: Do you think it'll be beneficial to use the `ThreadSafeKafkaConsumer` wrapper class? Maybe as part of the follow up PR?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -265,28 +259,39 @@ public void stop() {
...
-      String exceptionMessage = String.format("KafkaSystemConsumer: %s had started. Registration of ssp: %s, startpoint: %s failed.", this, systemStreamPartition, startpoint);
+      String exceptionMessage = String.format("KafkaSystemConsumer: %s had started. Registration of ssp: %s, offset: %s failed.", this, systemStreamPartition, offset);
       throw new SamzaException(exceptionMessage);
     }
 
     if (!Objects.equals(systemStreamPartition.getSystem(), systemName)) {
       LOG.warn("{}: ignoring SSP {}, because this consumer's system doesn't match.", this, systemStreamPartition);
       return;
     }
+    LOG.info("{}: Registering ssp = {} with offset {}", this, systemStreamPartition, offset);
```
dnishimura: nit: `LOG.info("{}: Registering ssp: {} with offset: {}", this, systemStreamPartition, offset);`
for logging consistency

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -309,11 +314,11 @@ public String toString() {
     return super.poll(systemStreamPartitions, timeout);
   }
 
-  static TopicAndPartition toTopicAndPartition(TopicPartition tp) {
+  public static TopicAndPartition toTopicAndPartition(TopicPartition tp) {
```
dnishimura: Does this need to be public?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -309,11 +314,11 @@ public String toString() {
...
     return new TopicAndPartition(tp.topic(), tp.partition());
   }
 
-  static TopicPartition toTopicPartition(SystemStreamPartition ssp) {
+  public static TopicPartition toTopicPartition(SystemStreamPartition ssp) {
```
dnishimura: Does this need to be public?

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
```diff
@@ -164,4 +165,10 @@ default void deleteMessages(Map<SystemStreamPartition, String> offsets) {
     throw new UnsupportedOperationException();
   }
 
+  /**
+   * Builds and returns the {@link StartpointVisitor} for a system.
+   */
+  default StartpointVisitor getStartpointVisitor() {
```
dnishimura: Are you going to integrate this with the `OffsetManager` in a follow PR? I don't see it being used in the current PR.

samza-api/src/main/java/org/apache/samza/startpoint/StartpointVisitor.java
```diff
@@ -18,54 +18,53 @@
...
+ * A {@link SystemAdmin} implementation should implement this abstraction to support {@link Startpoint}'s.
  */
+@InterfaceStability.Evolving
 public interface StartpointVisitor {
 
   /**
-   * Seek to specific offset represented by {@link StartpointSpecific}
-   * @param systemStreamPartition The {@link SystemStreamPartition} to seek the offset to.
-   * @param startpointSpecific The {@link Startpoint} that represents the specific offset.
+   * Resolves the {@link StartpointSpecific} to a system specific offset.
```
cameronlee314: In the visitor pattern, the visitor interface is usually agnostic of what its implementations actually need to do. This keeps the visitor flexible in case we need it for something other than offset resolution in the future.
For example, another way to use the visitor could be for serializing `Startpoint`s into a store.

samza-api/src/main/java/org/apache/samza/system/SystemAdmin.java
cameronlee314: Would it be clearer to add a method called `resolveStartpoint`? Then the `SystemAdmin` impl can internally choose how to implement.

samza-core/src/test/java/org/apache/samza/startpoint/TestStartpointManager.java
```diff
@@ -170,21 +170,6 @@ public void testBasics() {
...
   }
 
-  @Test
-  public void testStaleStartpoints() {
-    SystemStreamPartition ssp = new SystemStreamPartition("mockSystem", "mockStream", new Partition(2));
-    TaskName taskName = new TaskName("MockTask");
-
-    startpointManager.start();
-    long staleTimestamp = Instant.now().toEpochMilli() - StartpointManager.DEFAULT_EXPIRATION_DURATION.toMillis() - 2;
-    MockStartpointCustom startpoint = new MockStartpointCustom("das boot", 42, staleTimestamp);
```
cameronlee314: Did this test specifically need `StartpointCustom`? It doesn't seem like it needs to be fully removed.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -386,12 +389,10 @@ private OffsetsMaps fetchTopicPartitionsMetadata(List<TopicPartition> topicParti
     Map<TopicPartition, Long> oldestOffsetsWithLong;
     Map<TopicPartition, Long> upcomingOffsetsWithLong;
 
-    synchronized (metadataConsumer) {
-      oldestOffsetsWithLong = metadataConsumer.beginningOffsets(topicPartitions);
-      LOG.debug("Kafka-fetched beginningOffsets: {}", oldestOffsetsWithLong);
-      upcomingOffsetsWithLong = metadataConsumer.endOffsets(topicPartitions);
-      LOG.debug("Kafka-fetched endOffsets: {}", upcomingOffsetsWithLong);
-    }
+    oldestOffsetsWithLong = threadSafeKafkaConsumer.execute(consumer -> consumer.beginningOffsets(topicPartitions));
```
cameronlee314: Here and below: can now combine var declaration with assignment.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -386,12 +389,10 @@ private OffsetsMaps fetchTopicPartitionsMetadata(List<TopicPartition> topicParti
...
+    LOG.debug("Kafka-fetched beginningOffsets: {}", oldestOffsetsWithLong);
+    upcomingOffsetsWithLong = threadSafeKafkaConsumer.execute(consumer -> consumer.endOffsets(topicPartitions));
```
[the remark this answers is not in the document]
cameronlee314: +1: It looks like this actually slightly changes functionality: Before, we could guarantee that `metadataConsumer` would not get used/modified in between the call to `beginningOffsets` and `endOffsets`. Now, it is possible to use `metadataConsumer` in between. Could that be a problem?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -639,11 +625,12 @@ protected Properties createAdminClientProperties() {
...
   @Override
   public Set<SystemStream> getAllSystemStreams() {
-    synchronized (metadataConsumer) {
-      return ((Set<String>) this.metadataConsumer.listTopics().keySet()).stream()
-          .map(x -> new SystemStream(systemName, x))
-          .collect(Collectors.toSet());
-    }
+    Map<String, List<PartitionInfo>> topicToPartitionInfoMap = threadSafeKafkaConsumer.execute(consumer -> consumer.listTopics());
+    Set<SystemStream> systemStreams = topicToPartitionInfoMap.keySet()
+                                                             .stream()
```
cameronlee314: Is this the conventional indentation?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -771,4 +763,112 @@ public Properties getKafkaProperties() {
       return kafkaProperties;
     }
   }
+
+  /**
+   *
```
cameronlee314: Empty comment block

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -771,4 +763,112 @@ public Properties getKafkaProperties() {
...
+
+      Map<TopicPartition, Long> topicPartitionToTimestamp = ImmutableMap.of(topicPartition, startpointTimestamp.getTimestampOffset());
+      LOG.info("Finding offset for timestamp: {} in topic partition: {}.", startpointTimestamp.getTimestampOffset(), topicPartition);
+      Map<TopicPartition, OffsetAndTimestamp> topicPartitionToOffsetTimestamps = threadSafeKafkaConsumer.execute(consumer -> consumer.offsetsForTimes(topicPartitionToTimestamp));
+
+      OffsetAndTimestamp offsetAndTimestamp = topicPartitionToOffsetTimestamps.get(topicPartition);
+      if (offsetAndTimestamp != null) {
+        return String.valueOf(offsetAndTimestamp.offset());
+      } else {
+        return getEndOffset(systemStreamPartition);
```
cameronlee314: Maybe a log message would be useful here?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -111,26 +99,12 @@ public KafkaSystemConsumer(Consumer<K, V> kafkaConsumer, String systemName, Conf
     fetchThresholdBytesEnabled = new KafkaConfig(config).isConsumerFetchThresholdBytesEnabled(systemName);
 
     // create a sink for passing the messages between the proxy and the consumer
-    this.messageSink = new KafkaConsumerMessageSink();
+    messageSink = new KafkaConsumerMessageSink();
 
     // Create the proxy to do the actual message reading.
+    String metricName = String.format("%s-%s", systemName, clientId);
```
cameronlee314: Unused?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -142,6 +116,7 @@ public KafkaSystemConsumer(Consumer<K, V> kafkaConsumer, String systemName, Conf
    * @return KafkaConsumer newly created kafka consumer object
    */
   public static <K, V> KafkaConsumer<K, V> createKafkaConsumerImpl(String systemName, HashMap<String, Object> kafkaConsumerConfig) {
+
```
cameronlee314: Extra whitespace?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -186,14 +160,30 @@ private void startSubscription() {
...
-    topicPartitionToStartpointMap.forEach((topicPartition, startpoint) -> {
-      Partition partition = new Partition(topicPartition.partition());
-      SystemStreamPartition systemStreamPartition = new SystemStreamPartition(systemName, topicPartition.topic(), partition);
-      startpoint.apply(systemStreamPartition, kafkaStartpointRegistrationHandler);
+    topicPartitionsToOffset.forEach((tp, startingOffsetString) -> {
+      long startingOffset = Long.valueOf(startingOffsetString);
+
+      try {
+        synchronized (kafkaConsumer) {
+          kafkaConsumer.seek(tp, startingOffset); // this value should already be the 'upcoming' value
```
cameronlee314: Comment doesn't seem to match functionality. The `startingOffset` might not be the "upcoming" offset.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -73,17 +61,17 @@
...
 
   // This proxy contains a separate thread, which reads kafka messages (with consumer.poll()) and populates
   // BlockingEnvelopMap's buffers.
-  private final KafkaConsumerProxy proxy;
+  private final KafkaConsumerProxy<K, V> proxy;
 
-  // Holds mapping from {@code TopicPartition} to registered {@code Startpoint}. This will be used in the start().
-  final Map<TopicPartition, Startpoint> topicPartitionToStartpointMap = new HashMap<>();
+  // keep registration data until the start - mapping between registered SSPs and topicPartitions, and their offsets
+  Map<TopicPartition, String> topicPartitionsToOffset = new HashMap<>();
```
cameronlee314: Instead of having two parallel maps, could you have a single map from `TopicPartition` to "ssp with offset"? It will help ensure they don't get out of sync.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -163,13 +146,13 @@ public void testFetchThresholdBytes() {
     consumer.start();
     consumer.messageSink.addMessage(ssp0, ime0);
     // queue for ssp0 should be full now, because we added message of size FETCH_THRESHOLD_MSGS/partitionsNum
-    Assert.assertFalse(consumer.messageSink.needsMoreMessages(ssp0));
```
cameronlee314: Why is this getting changed?

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemConsumer.java
```diff
@@ -206,13 +189,13 @@ public void testFetchThresholdBytesDiabled() {
     consumer.start();
     consumer.messageSink.addMessage(ssp0, ime0);
     // should be full by size, but not full by number of messages (1 of 2)
-    Assert.assertTrue(consumer.messageSink.needsMoreMessages(ssp0));
```
cameronlee314: Why is this getting changed?

samza-api/src/main/java/org/apache/samza/startpoint/Startpoint.java
```diff
@@ -50,12 +50,13 @@ public long getCreationTimestamp() {
...
-   * @param systemStreamPartition The {@link SystemStreamPartition} needed to register with the {@link StartpointVisitor}
-   * @param startpointVisitor The visitor to register with.
+   * Applies the {@link StartpointVisitor}'s visit methods to the {@link Startpoint}
+   * and resolves it to a system specific offset.
+   * @param systemStreamPartition the {@link SystemStreamPartition} of the startpoint.
+   * @param startpointVisitor the visitor of the startpoint.
+   * @return the resolved offset.
    */
-  public abstract void apply(SystemStreamPartition systemStreamPartition, StartpointVisitor startpointVisitor);
+  public abstract String apply(SystemStreamPartition systemStreamPartition, StartpointVisitor startpointVisitor);
```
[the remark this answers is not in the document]
cameronlee314: One of the core concepts of the visitor pattern is to make the visitor generic. The caller is always going to use a specific implementation of the visitor, so they do need to know which one they are using.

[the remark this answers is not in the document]
cameronlee314: To clarify further, when the caller uses a visitor, the caller is directly using the concrete implementation of the visitor. So that usage is what defines the functional contract and return type. The base `StartpointVisitor` does not need to define the functional contract and return type.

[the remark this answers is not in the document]
dnishimura: I think for now, generifying the inputs and outputs won't benefit much, but rather will make code readability a little harder. If anything, the output should be explicit as possible or else the interpretation is left to the caller. Also I don't see the input being something other than `SystemStreamPartition` in the future. 

I believe the major motivation for this API change is that other components of Samza that deal with offsets understand string/specific offsets only. IMO, if we want to generify this API, we should take a holistic approach and do it across all offset related components. It'll probably take more careful thought, so could we revisit this approach later?

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemAdminJava.java
```diff
@@ -337,4 +351,91 @@ public void testShouldAssembleMetadata () {
...
+
+    final KafkaStartpointToOffsetResolver kafkaStartpointToOffsetResolver = new KafkaStartpointToOffsetResolver(consumer);
+
+    final StartpointTimestamp startpointTimestamp = new StartpointTimestamp(testTimeStamp);
+    final Map<TopicPartition, OffsetAndTimestamp> offsetForTimesResult = ImmutableMap.of(
+        TEST_TOPIC_PARTITION, new OffsetAndTimestamp(Long.valueOf(TEST_OFFSET), testTimeStamp));
+
+    // Mock the consumer interactions.
+    Mockito.when(consumer.offsetsForTimes(ImmutableMap.of(TEST_TOPIC_PARTITION, testTimeStamp))).thenReturn(offsetForTimesResult);
+    Mockito.doNothing().when(consumer).seek(TEST_TOPIC_PARTITION, Long.valueOf(TEST_OFFSET));
```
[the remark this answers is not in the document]
shanthoosh: Not necessary. Removed.

samza-api/src/main/java/org/apache/samza/startpoint/Startpoint.java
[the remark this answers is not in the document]
shanthoosh: Discussed offline with @cameronlee314 @dnishimura .  Though there was some agreements on the points raised against generification in above comments, overall generifying the StartpointVisitor will allow disjoint processing control-flows to be integrated with startpoints in the future. We'd agreed to generify the `StartpointVisitor` contract.

samza-api/src/main/java/org/apache/samza/startpoint/Startpoint.java
```diff
@@ -56,7 +55,7 @@ public long getCreationTimestamp() {
    * @param startpointVisitor the visitor of the startpoint.
    * @return the resolved offset.
    */
-  public abstract String apply(SystemStreamPartition systemStreamPartition, StartpointVisitor startpointVisitor);
+  public abstract String apply(SystemStreamPartition systemStreamPartition, StartpointVisitor<SystemStreamPartition, String> startpointVisitor);
```
cameronlee314: This should also be generified. Otherwise, the `StartpointVisitor` can't be used generically.
e.g.:
`public abstract OUT apply(IN in, StartpointVisitor<IN, OUT> startpointVisitor)`

shanthoosh: I was quite aware that we'd to generify the startpoint POJO. Without doing that, generifying the startpoint visitor would be pointless. I was in the middle of making the change and you'd reviewed the patch when it was not ready yet.

PR #1018, opened 2019-05-02
Adding tests for legacy store offset file

rmatharu:
The store offset file and format was updated in Samza 1.1 (to consolidate with offset file for side-inputs), and the tests were updated accordingly in https://github.com/apache/samza/pull/915.

This PR makes the tests parameterized so that they run for both the legacy and the new format of the offset file. The actual change to ensure that apps can rollback from 1.1 to 1.0 without rebootstrapping was checked as part of https://github.com/apache/samza/pull/1005


samza-core/src/test/scala/org/apache/samza/storage/TestTaskStorageManager.scala
```diff
@@ -711,6 +739,17 @@ class TestTaskStorageManager extends MockitoSugar {
   }
 }
 
+object TestTaskStorageManager {
+
+  @Parameters def parameters: util.Collection[Array[String]] = {
+    val offsetFileNames = new util.ArrayList[Array[String]]()
+    offsetFileNames.add(Array(StorageManagerUtil.OFFSET_FILE_NAME_NEW))
```
cameronlee314: Is there a test which ensures that if both files exist, then only the NEW one is used?
You might already have a test for this somewhere, but wanted to double check.

PR #1021, opened 2019-05-03
SAMZA-2213: Adding writing of container.metadata file, and moving exception writing logic to DiagnosticsManager

rmatharu:
This PR
* Removes ListGauge from everywhere. 
* Adds logic to LocalContainerRunner to write metadata, that is useful when combined with java's -XX:OnOutOfMemoryError=string option. 

* Adds DiagnosticsManager to publish error info to diagnostics stream instead of metrics reporter. 
* Adds logic to SamzaContainer to write the DiagnosticsManager while re-using the SystemProducer from the MetricsReporter.

* Cleansup ContainerProcessManagerMetrics. 

* Adds DiagnosticsManager to ContainerProcessManager and the ClusterBasedJC


samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -65,13 +76,30 @@ public static void main(String[] args) throws Exception {
...
 
+    writeMetadataFile(jobName, jobId, containerName, System.getenv(ShellCommandConfig.ENV_EXECUTION_ENV_CONTAINER_ID()), config);
+
     ApplicationDescriptorImpl<? extends ApplicationDescriptor> appDesc =
         ApplicationDescriptorUtil.getAppDescriptor(ApplicationUtil.fromConfig(config), config);
 
     ContainerLaunchUtil.run(appDesc, containerId, jobModel);
   }
+
+  public static void writeMetadataFile(String jobName, String jobId, String containerName, String execEnvContainerId, Config config) throws Exception {
```
mynameborat: I suppose we need this logic in AM as well to react to OOM in the application master.
Would it be possible to have AM detect that container OOM'd in which case, we have this logic in one place and have the AM react for container/AM OOMing?

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -29,17 +31,25 @@
...
-import java.util.Random;
+
 
 /**
  * Launches and manages the lifecycle for {@link SamzaContainer}s in YARN.
  */
 public class LocalContainerRunner {
   private static final Logger log = LoggerFactory.getLogger(LocalContainerRunner.class);
+  private static final String CONTAINER_METADATA_FILENAME_FORMAT = "%s.metadata"; // Filename: containerID.metadata
+  private static final String CONTAINER_METADATA_DIRECTORY = System.getProperty("samza.log.dir");
```
prateekm: Would prefer moving this to JobConfig instead of referencing system properties directly (they're not as well documented). Should also handle the case where this is missing (e.g. don't do anything).

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
```diff
@@ -29,17 +31,25 @@
...
 import org.slf4j.MDC;
-import java.util.Random;
+
 
 /**
  * Launches and manages the lifecycle for {@link SamzaContainer}s in YARN.
  */
 public class LocalContainerRunner {
   private static final Logger log = LoggerFactory.getLogger(LocalContainerRunner.class);
+  private static final String CONTAINER_METADATA_FILENAME_FORMAT = "%s.metadata"; // Filename: containerID.metadata
```
prateekm: Maybe clarify that the containerID is the cluster manager container id.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -141,6 +143,9 @@ object SamzaContainer extends Logging {
     val systemConfig = new SystemConfig(config)
     val containerModel = jobModel.getContainers.get(containerId)
     val containerName = "samza-container-%s" format containerId
+    val jobName = config.getName.getOrElse(throw new ConfigException("Missing required config: job.name"))
```
prateekm: Does config.getName already check for app.name first? The exception message may also be misleading.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -141,6 +143,9 @@ object SamzaContainer extends Logging {
...
+    val jobId = config.getJobId
+    val execEnvContainerId = System.getenv(ShellCommandConfig.ENV_EXECUTION_ENV_CONTAINER_ID)
```
prateekm: Maybe pass as a constructor param? Would be good to remove system properties dependency in SamzaContainer. E.g., What happens in standalone?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -409,10 +414,36 @@ object SamzaContainer extends Logging {
...
+      val diagnosticsMetricsReporter = new MetricsSnapshotReporter(systemProducer, diagnosticsSystemStream, publishInterval, jobName, jobId,
+        containerName, Util.getTaskClassVersion(config), Util.getSamzaVersion(), Util.getLocalHost.getHostName,
+        new MetricsSnapshotSerdeV2, config.getMetricsSnapshotReporterBlacklist(diagnosticsReporterName))
+
+      reporters += (diagnosticsReporterName -> diagnosticsMetricsReporter)
+
+      new DiagnosticsManager(jobName, jobId, containerName, execEnvContainerId, Util.getTaskClassVersion(config),
+        Util.getSamzaVersion, Util.getLocalHost.getHostName, diagnosticsSystemStream, systemProducer)
+    }
+    else null
```
prateekm: Use options instead of null. nulls in Scala are unexpected and may cause accidental NPEs in future.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -409,10 +414,36 @@ object SamzaContainer extends Logging {
...
 
+    info("Setting up diagnostics.")
+
+    val diagnosticsManager = if (config.getDiagnosticsEnabled) {
+
+      val diagnosticsReporterName = MetricsConfig.METRICS_SNAPSHOT_REPORTER_NAME_FOR_DIAGNOSTICS
+
+      val publishInterval: Int = config.getMetricsSnapshotReporterInterval(diagnosticsReporterName).getOrElse("60").toInt
+
+      val diagnosticsSystemStream = StreamUtil.getSystemStreamFromNames(new MetricsConfig(config).
```
prateekm: Minor: Extract variables for parameters for readability.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -409,10 +414,36 @@ object SamzaContainer extends Logging {
...
     info("Setting up metrics reporters.")
 
-    val reporters = MetricsReporterLoader.getMetricsReporters(config, containerName).asScala.toMap ++ customReporters
+    var reporters = MetricsReporterLoader.getMetricsReporters(config, containerName).asScala.toMap ++ customReporters
 
     info("Got metrics reporters: %s" format reporters.keys)
 
+    info("Setting up diagnostics.")
+
+    val diagnosticsManager = if (config.getDiagnosticsEnabled) {
```
prateekm: This should be created in LCR / LAR and wired in to the container.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -690,7 +722,8 @@ class SamzaContainer(
   containerContext: ContainerContext,
   applicationContainerContextOption: Option[ApplicationContainerContext],
   externalContextOption: Option[ExternalContext],
-  containerStorageManager: ContainerStorageManager) extends Runnable with Logging {
+  containerStorageManager: ContainerStorageManager,
+  diagnosticsManager: DiagnosticsManager = null) extends Runnable with Logging {
```
prateekm: Use options instead of nulls.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -1077,6 +1095,12 @@ class SamzaContainer(
     offsetManager.stop
   }
 
+  def shutdownDiagnostics {
+    info("Shutting down diagnostics manager.")
+    if (diagnosticsManager != null)
+      diagnosticsManager.stop()
```
prateekm: Formatting: Braces for if always.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
```diff
@@ -0,0 +1,173 @@
...
+      String taskClassVersion, String samzaVersion, String hostname, SystemStream diagnosticSystemStream,
+      SystemProducer systemProducer) {
+    this.jobName = jobName;
+    this.jobId = jobId;
+    this.containerName = containerName;
+    this.executionEnvContainerId = executionEnvContainerId;
+    this.taskClassVersion = taskClassVersion;
+    this.samzaVersion = samzaVersion;
+    this.hostname = hostname;
+    resetTime = Instant.now();
```
prateekm: Is this still relevant? Maybe just do an Instant.now() on line 153?

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
```diff
@@ -0,0 +1,173 @@
...
+  public void start() {
+    this.scheduler.scheduleWithFixedDelay(new DiagnosticsStreamPublisher(), 0, DEFAULT_PUBLISH_PERIOD.getSeconds(),
+        TimeUnit.SECONDS);
+  }
+
+  public void stop() throws InterruptedException {
+    scheduler.shutdown();
+
+    // Allow any scheduled publishes to finish, and block for termination (for max 60 seconds)
+    scheduler.awaitTermination(60, TimeUnit.SECONDS);
```
prateekm: Shouldn't block for more than task shutdown ms, ideally much less. Should call shutdownNow() if timeout expires.

samza-core/src/main/scala/org/apache/samza/util/FileUtil.scala
```diff
@@ -51,11 +51,51 @@ object FileUtil {
...
+    * @param file The file handle to write to
+    * @param data The data to be written to the file
+    * @param append true for appending data to file, false otherwise
+    * */
+  def writeToTextFile(file: File, data: String, append: Boolean): Unit = {
+
+    val tmpFilePath = file.getAbsolutePath + ".tmp"
+    var fileWriter: FileWriter = null
+    val tmpFile = new File(tmpFilePath)
+
```
prateekm: Minor: remove extra newline.

samza-core/src/main/scala/org/apache/samza/util/FileUtil.scala
```diff
@@ -51,11 +51,51 @@ object FileUtil {
...
+
+    //atomic swap of tmp and real file if we need to append
+    if (append) {
+      swapFiles(file, tmpFile)
+    }
+
+    try {
+      System.out.println("--------" + tmpFile)
```
prateekm: Remove?

samza-core/src/main/scala/org/apache/samza/util/FileUtil.scala
```diff
@@ -51,11 +51,51 @@ object FileUtil {
...
+      fileWriter = new FileWriter(tmpFile, append)
+      fileWriter.write(data)
+    } catch {
+      case e: Exception =>
+        System.out.println(e)
```
prateekm: Also log details ("error writing data: {} to file: {} in mode: {} ... ")

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
rmatharu: Its possible, but shouldnt be a common occurrence. 
I can move this to a common place, and have the ClusterBasedJC also call it.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
rmatharu: After https://github.com/apache/samza/pull/920/files 
the planner enforces job and app name to be the same, 
so yes, by the time we land in Samza Container we would've already check both configs.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
rmatharu: Yes in MetricsHeader reset-time and time are different things. 
Reset-time = container startup timestamp. 
time = timestamp at time of emitting metric.

rmatharu:
@prateekm  Addressed all comments. 
Tested on local yarn.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -219,6 +220,11 @@ public void run() {
       //initialize JobCoordinator state
       LOG.info("Starting cluster based job coordinator");
 
+      // write the diagnostics metadata file
+      Optional<String> execEnvContainerId = Optional.ofNullable(System.getenv("CONTAINER_ID"));
```
rmatharu: Is there a better way of obtaining this?
In AM the yarn-containerID is in env variable CONTAINER_ID whereas on containers it is under 
EXEC_ENV_CONTAINER_ID.

samza-core/src/main/java/org/apache/samza/runtime/LocalContainerRunner.java
rmatharu: done

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -219,6 +220,11 @@ public void run() {
...
+      DiagnosticsUtil.writeMetadataFile(new JobConfig(config).getName().get(), new JobConfig(config).getJobId(),
```
prateekm: Nitpick: Extract fields for jobName, jobId.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -219,6 +220,11 @@ public void run() {
...
+          "ApplicationMaster", execEnvContainerId, config);
```
prateekm: Extract/reuse constant for string used across multiple files (METRICS_SOURCE_NAME in CPM)

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -106,7 +117,10 @@
   /**
    * Metrics for the {@link ContainerProcessManager}
    */
-  private final ContainerProcessManagerMetrics metrics;
+  private final static String METRICS_SOURCE_NAME = "ApplicationMaster";
```
prateekm: Nitpick: declaration order: statics before fields.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -23,16 +23,25 @@
...
+import org.apache.samza.util.DiagnosticsUtil;
+import org.apache.samza.util.MetricsReporterLoader;
 import org.apache.samza.util.Util;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import scala.Option;
+import scala.Tuple2;
```
prateekm: Prefer using a pair class instead of scala tuple2.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -135,23 +169,9 @@ public ContainerProcessManager(Config config, SamzaApplicationState state, Metri
...
-    this.metrics = new ContainerProcessManagerMetrics(config, state, registry);
-    this.containerAllocator = allocator;
-    this.allocatorThread = new Thread(this.containerAllocator, "Container Allocator Thread");
-    this.standbyContainerManager = Optional.empty();
-  }
-
   //package private, used only in tests
   ContainerProcessManager(Config config, SamzaApplicationState state, MetricsRegistryMap registry,
-      ClusterResourceManager resourceManager) {
+      ClusterResourceManager resourceManager, Optional<AbstractContainerAllocator> allocator) {
```
prateekm: Why is the allocator optional? Can a CPM even work without it? Would prefer to keep allocator creation logic in a single place.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -117,8 +131,28 @@ public ContainerProcessManager(Config config, SamzaApplicationState state, Metri
...
     this.clusterResourceManager = checkNotNull(factory.getClusterResourceManager(this, state));
-    this.metrics = new ContainerProcessManagerMetrics(config, state, registry);
 
+    // Initialize metrics
+    this.containerProcessManagerMetrics = new ContainerProcessManagerMetrics(config, state, registry);
+    this.jvmMetrics = new JvmMetrics(registry);
+    this.metricsReporters = MetricsReporterLoader.getMetricsReporters(new MetricsConfig(config), METRICS_SOURCE_NAME);
+
+    // Creating diagnostics manager and reporter, and wiring it respectively
+    Optional<String> execEnvContainerId = Optional.ofNullable(System.getenv("CONTAINER_ID"));
```
prateekm: Is there a constant for this env property?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -117,8 +131,28 @@ public ContainerProcessManager(Config config, SamzaApplicationState state, Metri
...
+    Optional<Tuple2<DiagnosticsManager, MetricsSnapshotReporter>> diagnosticsManagerReporterPair = DiagnosticsUtil.buildDiagnosticsManager(new JobConfig(config).getName().get(),
```
prateekm: Extract fields for jobname, id.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -186,7 +208,18 @@ public boolean shouldShutdown() {
 
   public void start() {
     log.info("Starting the container process manager");
-    metrics.start();
+
+    if (jvmMetrics != null) {
+      jvmMetrics.start();
```
prateekm: Why do 'metrics' need to be started?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -217,12 +250,28 @@ public void stop() {
...
+        log.error("InterruptedException while stopping diagnosticsManager", e);
+      }
+    }
+
+    if (containerProcessManagerMetrics != null) {
       try {
-        metrics.stop();
-        log.info("Stopped metrics reporters");
+
+        if (this.metricsReporters != null) {
```
prateekm: Should this be independent of whether CPMMetrics are defined or not?

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -330,11 +336,20 @@ SamzaContainer getContainer() {
 
   @VisibleForTesting
   SamzaContainer createSamzaContainer(String processorId, JobModel jobModel) {
+
+    // Creating diagnostics manager and reporter, and wiring it respectively
+    Optional<Tuple2<DiagnosticsManager, MetricsSnapshotReporter>> diagnosticsManagerReporterPair = DiagnosticsUtil.buildDiagnosticsManager(new ApplicationConfig(config).getAppName(), new ApplicationConfig(config).getAppId(), processorId, Optional.empty(), config);
```
prateekm: Why is this appName and the other ones jobName?
Nitpick: extract fields.

samza-core/src/main/java/org/apache/samza/runtime/ContainerLaunchUtil.java
```diff
@@ -90,15 +98,24 @@ private static void run(
...
       }
 
+      Map<String, MetricsReporter> metricsReporters = loadMetricsReporters(appDesc, containerId, config);
+
+      // Creating diagnostics manager and reporter, and wiring it respectively
+      Optional<Tuple2<DiagnosticsManager, MetricsSnapshotReporter>> diagnosticsManagerReporterPair = DiagnosticsUtil.buildDiagnosticsManager(jobName, jobId, containerId, execEnvContainerId, config);
+      Option<DiagnosticsManager> diagnosticsManager = Option.empty();
+      if (diagnosticsManagerReporterPair.isPresent()) {
+        diagnosticsManager = Option.apply(diagnosticsManagerReporterPair.get()._1());
+        metricsReporters.put(MetricsConfig.METRICS_SNAPSHOT_REPORTER_NAME_FOR_DIAGNOSTICS(), diagnosticsManagerReporterPair.get()._2());
```
prateekm: Nitpick: MetricsConfig.DIAGNOSTICS_SNAPSHOT_REPORTER_NAME?

samza-core/src/main/java/org/apache/samza/runtime/ContainerLaunchUtil.java
```diff
@@ -150,6 +167,7 @@ public void afterFailure(Throwable t) {
     }
   }
 
+
```
prateekm: Nitpick: remove extra newline.

samza-core/src/main/java/org/apache/samza/util/DiagnosticsUtil.java
```diff
@@ -0,0 +1,131 @@
...
+
+      StringBuilder metadata = new StringBuilder("Version: 1");
+      metadata.append(System.lineSeparator());
+      MetricsHeader metricsHeader =
+          new MetricsHeader(jobName, jobId, "samza-container-" + containerId, execEnvContainerId.orElse(""), LocalContainerRunner.class.getName(),
+              Util.getTaskClassVersion(config), Util.getSamzaVersion(), Util.getLocalHost().getHostName(),
+              System.currentTimeMillis(), System.currentTimeMillis());
+
+      MetricsSnapshot metricsSnapshot = new MetricsSnapshot(metricsHeader, new Metrics());
+      metadata.append("MetricsSnapshot: ");
```
prateekm: Should this be "MetricsHeader" or "ContainerMetadata" something? IIUC This is not a snapshot of the metrics themselves, and also not particularly related to the metrics. Also, can you serialize the header directly (e.g. using SamzaObjectMapper / ObjectMapper)?

samza-core/src/main/java/org/apache/samza/util/DiagnosticsUtil.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+  }
+
+
+  /**
+   * Create a pair of DiagnosticsManager and Reporter for the given jobName, jobId, containerId, and execEnvContainerId,
+   * if diagnostics is enabled.
+   * execEnvContainerId is the ID assigned to the container by the cluster manager (e.g., YARN).
+   */
+  public static Optional<Tuple2<DiagnosticsManager, MetricsSnapshotReporter>> buildDiagnosticsManager(String jobName, String jobId,
```
prateekm: Should return a java pair class. Can use from apache commons, or add a new one to util, it's generally useful.

samza-core/src/main/java/org/apache/samza/util/DiagnosticsUtil.java
```diff
@@ -0,0 +1,131 @@
...
+
+      if (diagnosticsReporterStreamName.isEmpty()) {
+        throw new ConfigException("Missing required config: " + String.format(MetricsConfig.METRICS_SNAPSHOT_REPORTER_STREAM(), diagnosticsReporterName));
+      }
+
+      SystemStream diagnosticsSystemStream = StreamUtil.getSystemStreamFromNames(diagnosticsReporterStreamName.get());
+
+      Optional<String> diagnosticsSystemFactoryName = new SystemConfig(config).getSystemFactory(diagnosticsSystemStream.getSystem());
+      if (!diagnosticsSystemFactoryName.isPresent()) {
+        throw new SamzaException("A stream uses system " + diagnosticsSystemStream.getSystem() + ", which is missing from the configuration.");
```
prateekm: Maybe "Missing factory class for diagnostic system: ..."?

samza-core/src/main/java/org/apache/samza/util/DiagnosticsUtil.java
```diff
@@ -0,0 +1,131 @@
...
+      // Create a systemProducer for giving to diagnostic-reporter and diagnosticsManager
+      SystemFactory systemFactory = Util.getObj(diagnosticsSystemFactoryName.get(), SystemFactory.class);
+      SystemProducer systemProducer = systemFactory.getProducer(diagnosticsSystemStream.getSystem(), config, new MetricsRegistryMap());
+      DiagnosticsManager diagnosticsManager = new DiagnosticsManager(jobName, jobId, containerId, execEnvContainerId.orElse(""), taskClassVersion,
+          samzaVersion, hostName, diagnosticsSystemStream, systemProducer, Duration.ofMillis(new TaskConfigJava(config).getShutdownMs()));
+
+      MetricsSnapshotReporter diagnosticsReporter =
+          new MetricsSnapshotReporter(systemProducer, diagnosticsSystemStream, publishInterval, jobName, jobId,
+              "samza-container-" + containerId, taskClassVersion, samzaVersion, hostName, new MetricsSnapshotSerdeV2(),
+              blacklist, new AbstractFunction0<Object>() {
```
prateekm: ScalaJavaUtil.toScalaFunction(() -> System.currentTimeMillis())

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -114,6 +114,10 @@ object JobConfig {
   val DEFAULT_STANDBY_TASKS_REPLICATION_FACTOR = 1
   val SYSTEM_STREAM_PARTITION_MAPPER_FACTORY = "job.system.stream.partition.mapper.factory"
 
+  // Naming format and directory for container.metadata file
+  private val CONTAINER_METADATA_FILENAME_FORMAT = "%s.metadata" // Filename: <containerID>.metadata
+  private val CONTAINER_METADATA_DIRECTORY_SYS_PROPERTY = "samza.log.dir"
```
prateekm: Is this introducing a new config or redefining the constant key for an existing? If the former, is this really required? Is this documented?

samza-core/src/main/scala/org/apache/samza/config/MetricsConfig.scala
```diff
@@ -49,7 +49,7 @@ class MetricsConfig(config: Config) extends ScalaMapConfig(config) {
 
   def getMetricsSnapshotReporterStream(name: String): Option[String] = getOption(MetricsConfig.METRICS_SNAPSHOT_REPORTER_STREAM format name)
 
-  def getMetricsSnapshotReporterInterval(name: String): Option[String] = getOption(MetricsConfig.METRICS_SNAPSHOT_REPORTER_INTERVAL format name)
+  def getMetricsSnapshotReporterInterval(name: String): String = getOption(MetricsConfig.METRICS_SNAPSHOT_REPORTER_INTERVAL format name).getOrElse("60")
```
prateekm: Should update config docs with default value.

samza-core/src/main/scala/org/apache/samza/config/TaskConfig.scala
```diff
@@ -96,11 +96,6 @@ class TaskConfig(config: Config) extends ScalaMapConfig(config) with Logging {
     case _ => TaskConfig.DEFAULT_COMMIT_MS
   }
 
-  def getShutdownMs: Option[Long] = getOption(TaskConfig.SHUTDOWN_MS) match {
```
prateekm: What happened to this? Present in the java version?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -1077,6 +1069,13 @@ class SamzaContainer(
     offsetManager.stop
   }
 
+  def shutdownDiagnostics {
+    info("Shutting down diagnostics manager.")
```
prateekm: Move inside isDefined

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
```diff
@@ -0,0 +1,180 @@
...
+import scala.Tuple2;
+import scala.collection.JavaConverters;
+
+
+/**
+ * Responsible for publishing data to the diagnostic stream.
+ * Currently emits exception/error events obtained using a customer-appender that attaches to the root-logger.
+ */
+public class DiagnosticsManager {
+
```
prateekm: Nitpick: remove extra newlines.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
```diff
@@ -0,0 +1,180 @@
...
+  private final String hostname;
+  private final Instant resetTime;
+
+  private SystemProducer systemProducer; // SystemProducer for writing diagnostics data
+  private BoundedList<DiagnosticsExceptionEvent> exceptions; // A BoundedList for storing DiagnosticExceptionEvent
+  private final ScheduledExecutorService scheduler; // Scheduler for pushing data to the diagnostic stream
+  private final Duration terminationDuration; // duration to wait when terminating the scheduler
+
+  private final SystemStream diagnosticSystemStream;
+
```
prateekm: Nitpick: remove extra newlines.

samza-core/src/main/scala/org/apache/samza/config/MetricsConfig.scala
prateekm: Maybe return int here directly instead of toInt later. Not sure why this was a string.

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -64,13 +65,37 @@ object Util extends Logging {
...
+        warn("Unable to find implementation samza version in jar's meta info. Defaulting to 0.0.1.")
+        "0.0.1"
+      })
+  }
+
+  def getTaskClassVersion(config: Config): String = {
+    try {
+      val taskClass = Option(new ApplicationConfig(config).getAppClass())
+        .orElse(new TaskConfig(config).getTaskClass).get
+      Option(Class.forName(taskClass).getPackage.getImplementationVersion).get
```
prateekm: Why wrap in option then get?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: Because we have tests that pass in their own allocator, and other tests that want the CPM to create the allocator, and we had 3 different constructors for these tests.

samza-core/src/main/scala/org/apache/samza/config/TaskConfig.scala
rmatharu: This was duplicated between TaskConfig.scala and TaskConfigJava.java 
(there is duplication between these, Im working parallely on removing)

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: org.apache.commons.lang3.tuple.Pair seems more useful than the rest.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
rmatharu: This is already existing. This is just defining the constant.

samza-core/src/main/java/org/apache/samza/util/DiagnosticsUtil.java
rmatharu: Thanks!

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: Because someone decided to have JVMetrics with their own executor pool

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: There are numerous tests that fail if i dont have this.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
rmatharu: I thought standalone would require appconfig, replaced with job-config

samza-core/src/main/java/org/apache/samza/runtime/ContainerLaunchUtil.java
rmatharu: Not changing right now since its used in li-samza-rewriter.

samza-core/src/main/java/org/apache/samza/util/DiagnosticsUtil.java
rmatharu: I prefer using the serde, which internally uses objectMapper with additional information for typing.

rmatharu:
Addressed all.

samza-core/src/main/java/org/apache/samza/util/DiagnosticsUtil.java
```diff
@@ -0,0 +1,126 @@
...
+      StringBuilder metadata = new StringBuilder("Version: 1");
+      metadata.append(System.lineSeparator());
+      MetricsHeader metricsHeader =
+          new MetricsHeader(jobName, jobId, "samza-container-" + containerId, execEnvContainerId.orElse(""), LocalContainerRunner.class.getName(),
+              Util.getTaskClassVersion(config), Util.getSamzaVersion(), Util.getLocalHost().getHostName(),
+              System.currentTimeMillis(), System.currentTimeMillis());
+
+      MetricsSnapshot metricsSnapshot = new MetricsSnapshot(metricsHeader, new Metrics());
+      metadata.append("ContainerMetadata: ");
+      metadata.append(new String(new MetricsSnapshotSerdeV2().toBytes(metricsSnapshot)));
```
shanthoosh: Just curious.

Here we use the string serde's  to store the entire data in the string form in separate lines(version, container-metadata) on disk. This would necessitate readers having to read individual lines and have the parsing logic on their end. 

Instead, if we use the JsonSerde(or avro or any serde) to serialize/deserialize the entire metadata-content, then  readers (or monitors) doesn't need to have a parsing logic from their front to deserialize the data. We would have a POJO describing the data-format clearly and promote tractable data-evolution. 

What do you think?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
prateekm: Bump ^

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
prateekm: Oh this is only for testing, it's fine then.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
prateekm: I mean this block can be outside of the `if (containerProcessManagerMetrics != null)` check by itself.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
prateekm: Why not use the existing constant then? It'll be simpler to find usages later.

samza-core/src/main/scala/org/apache/samza/config/JobConfig.scala
```diff
@@ -132,6 +136,18 @@ object JobConfig {
     }
     fwkPath
   }
+
+  // The metadata file is written in a <exec-env-container-id>.metadata file in the log-dir of the container.
```
prateekm: Minor: /** instead of // for method javadocs

samza-core/src/main/java/org/apache/samza/util/DiagnosticsUtil.java
rmatharu: Sure let me followup on that and give that a shot, should be fairly straightforward; just need to make sure samza-admin has everything to be able to read it.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: I defined one, but the principled way to fix this is to define an API in ClusterResourceManager called getContainerId() and have YarnClusterResourceManager (which is in samza-yarn) return it. 
I will do that change separately.

rmatharu:
All done

PR #1026, opened 2019-05-09
SAMZA-2189: Integrate startpoint resolution workflow with SamzaContainer startup sequence.

shanthoosh:

* Integrates the resolving startpoints to offset workflow to container startup sequence.
* Either if the startpoint does not exist for a (taskname, system-stream-partition) or the startpoint resolution fails, then the checkpointed offset is used as starting offset 
* Added unit-tests to validate the altered starting offset computation flow for a partition.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
+++ b/samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
@@ -23,14 +23,15 @@ package org.apache.samza.container
 import java.util.{Objects, Optional}
 import java.util.concurrent.ScheduledExecutorService
 
+import org.apache.commons.lang3.StringUtils
 import org.apache.samza.SamzaException
 import org.apache.samza.checkpoint.OffsetManager
 import org.apache.samza.config.Config
 import org.apache.samza.config.StreamConfig.Config2Stream
 import org.apache.samza.context._
 import org.apache.samza.job.model.{JobModel, TaskModel}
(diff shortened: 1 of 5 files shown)


shanthoosh:
@dnishimura 

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -141,15 +142,36 @@ class TaskInstance(
...
     systemStreamPartitions.foreach(systemStreamPartition => {
-      val startingOffset = getStartingOffset(systemStreamPartition)
-      val startpoint = offsetManager.getStartpoint(taskName, systemStreamPartition).getOrElse(null)
-      consumerMultiplexer.register(systemStreamPartition, startingOffset, startpoint)
-      metrics.addOffsetGauge(systemStreamPartition, () =>
-          offsetManager.getLastProcessedOffset(taskName, systemStreamPartition).orNull
-        )
+      var startingOffset: String = getStartingOffset(systemStreamPartition)
+      val startpoint: Option[Startpoint] = offsetManager.getStartpoint(taskName, systemStreamPartition)
+      if (startpoint != null && startpoint.isDefined) {
```
dnishimura: nit: Probably cleaner to use scala `match`:
```
startpoint match {
  case Some(sp) => ...
  case None => ...
}
```

shanthoosh: Done.

PR #1027, opened 2019-05-09
SAMZA-2046: Startpoint fan out implementation

dnishimura:
This PR implements the fan out logic. During fan out, Startpoints written to the metadata store keyed by SSP and/or SSP+TaskName are moved to a seperate namespace with structure: `Map<TaskName, Map<SSP, Startpoint>>`, similar to how checkpoints are structured. 

The intent is for the job coordinators perform the fan out upon startup. This allows for Startpoints to be written while the job is running and applied to the starting offsets only upon job start time. The fan out isolation allows for Startpoints to be written at anytime.

A follow on PR after this PR will integrate the fan out with the various job coordinators. Another follow on PR(#1026) that is in flight and is required for this to work is the integration of the Startpoint-to-starting offset resolver.

@shanthoosh @cameronlee314 @prateekm


samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -173,71 +226,151 @@ public void deleteStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
+      if (startpoint1.getCreationTimestamp() > startpoint2.getCreationTimestamp()) {
+        return startpoint1;
+      }
+      return startpoint2;
+    }
+    return startpoint1 != null ? startpoint1 : startpoint2;
+  }
+
+  private static String toReadWriteStoreKey(SystemStreamPartition ssp, TaskName taskName) {
     return new String(new JsonSerdeV2<>().toBytes(new StartpointKey(ssp, taskName)));
```
cameronlee314: Is `JsonSerdeV2` guaranteed to keep the same ordering of keys across different JVMs?
For example, {"ssp": "ssp_value", "taskName": "taskName_value"} is the same semantic key as {"taskName": "taskName_value", "ssp": "ssp_value"}, but the byte representation is different.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -356,20 +356,13 @@ class OffsetManager(
...
+      info("%d startpoint(s) for taskName: %s have been committed to the checkpoint." format (startpoints.get(taskName).size, taskName.getTaskName))
+      startpointManager.removeFanOutForTask(taskName)
       startpoints -= taskName
 
       if (startpoints.isEmpty) {
-        // Stop startpoint manager after last startpoint is deleted
-        startpointManager.stop()
-        info("No more startpoints left to consume. Stopped the startpoint manager.")
+        info("All outstanding startpoints have been committed to the checkpoint.")
+        startpointManager.stop
```
cameronlee314: This would have been better as a comment on the previous PR:
Should the `StartpointManager.stop` javadoc be updated to make sure it is idempotent (in case changes are made in the future)?

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -517,23 +510,25 @@ class OffsetManager(
...
-
-          if (!sspToStartpoint.isEmpty) {
-            startpoints += taskName -> sspToStartpoint
+          Option(startpointManager.getFanOutForTask(taskName)) match {
+            case Some(fanOut) => {
+              val filteredFanOut = fanOut.asScala
+                .filter(f => systemStreamPartitionSet.contains(f._1))
+                .toMap
+              if (!filteredFanOut.isEmpty) {
+                startpoints += taskName -> filteredFanOut
```
cameronlee314: Do you have a log somewhere which shows that there was a fanout entry?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -173,71 +226,151 @@ public void deleteStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
+    LOG.info("Fanning out to {} tasks", fanOuts.size());
+
+    StartpointFanOut.StartpointFanOutSerde startpointFanOutSerde = new StartpointFanOut.StartpointFanOutSerde();
+    for (TaskName taskName : fanOuts.keySet()) {
+      String fanOutKey = toFanOutStoreKey(taskName);
+      StartpointFanOut newFanOut = fanOuts.get(taskName);
+      byte[] fanOutFromStoreInBytes = fanOutStore.get(fanOutKey);
+
+      if (ArrayUtils.isNotEmpty(fanOutFromStoreInBytes)) {
+        // Merge new fan out with existing fan out. This is not a typical scenario but it can happen when the job
```
cameronlee314: Is there any concern here regarding non-atomic merge? Is there a case where the job coordinator could be merging a startpoint at the same time as a processing container deleting a startpoint?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -173,71 +226,151 @@ public void deleteStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
+        // restarts before all startpoints are committed by the first checkpoint commit and new startpoints were written
+        // to the readWriteStore after the previous fan out.
+        StartpointFanOut existingFanOut = startpointFanOutSerde.fromBytes(fanOutFromStoreInBytes);
+        existingFanOut.getFanOuts().putAll(newFanOut.getFanOuts());
+        newFanOut = existingFanOut;
+      }
+
+      fanOutStore.put(fanOutKey, startpointFanOutSerde.toBytes(newFanOut));
```
cameronlee314: The fanout is not transactional across all tasks, so what if there is a failure in the middle of writing all of the fanouts? Could that result in a corrupt state?

samza-core/src/test/scala/org/apache/samza/checkpoint/TestOffsetManager.scala
```diff
@@ -138,33 +142,35 @@ class TestOffsetManager {
...
     val offsetManager = OffsetManager(systemStreamMetadata, config, checkpointManager, getStartpointManager(), systemAdmins, Map(), new OffsetManagerMetrics)
     offsetManager.register(taskName, Set(systemStreamPartition))
 
     // Pre-populate startpoint
     startpointManager.writeStartpoint(systemStreamPartition, taskName, new StartpointOldest)
+    startpointManager.fanOut(asTaskToSSPMap(taskName, systemStreamPartition))
     offsetManager.start
     // Should get offset 45 back from the checkpoint manager, which is last processed, and system admin should return 46 as starting offset.
-    assertNotNull(startpointManager.readStartpoint(systemStreamPartition, taskName))
+    assertEquals(1, startpointManager.getFanOutForTask(taskName).size())
```
cameronlee314: Is this assertion necessary? Is it testing some part of `OffsetManager`? If so, then is it possible to check the actual result of `getFanOutForTask`?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: The fan out will happen around the time the job coordinator is computing the job model, so the containers should not be running at this point. It may happen on a rogue/orphan container

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -173,71 +226,151 @@ public void deleteStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
+      String fanOutKey = toFanOutStoreKey(taskName);
+      StartpointFanOut newFanOut = fanOuts.get(taskName);
+      byte[] fanOutFromStoreInBytes = fanOutStore.get(fanOutKey);
+
+      if (ArrayUtils.isNotEmpty(fanOutFromStoreInBytes)) {
+        // Merge new fan out with existing fan out. This is not a typical scenario but it can happen when the job
+        // restarts before all startpoints are committed by the first checkpoint commit and new startpoints were written
+        // to the readWriteStore after the previous fan out.
+        StartpointFanOut existingFanOut = startpointFanOutSerde.fromBytes(fanOutFromStoreInBytes);
+        existingFanOut.getFanOuts().putAll(newFanOut.getFanOuts());
```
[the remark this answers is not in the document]
dnishimura: It's guaranteed here, but I see where you're getting it at. I'll put a stronger guarantee in the fan out class itself and create a new entry as you suggested.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -173,71 +226,151 @@ public void deleteStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
+        newFanOut = existingFanOut;
+      }
+
+      fanOutStore.put(fanOutKey, startpointFanOutSerde.toBytes(newFanOut));
 
+      // Delete from readWriteStore after fan out.
+      for (SystemStreamPartition ssp : newFanOut.getFanOuts().keySet()) {
```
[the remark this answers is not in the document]
dnishimura: Ah it was based on some old code I got rid of but forgot to change the logic here. Will iterate through the keys and call delete.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -173,71 +226,151 @@ public void deleteStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
         }
       }
     }
 
-    // Delete SSP-only keys
-    sspsToDelete.forEach(ssp -> {
-        deleteStartpoint(ssp);
-        LOG.info("All Startpoints for SSP: {} have been grouped to the appropriate tasks and the SSP was deleted.");
-      });
+    return ImmutableSet.copyOf(fanOuts.keySet());
```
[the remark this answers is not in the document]
dnishimura: You're right probably not necessary. taskToSSPs only provides the task-->SSP mappings to use in the fan out. Originally I had the job model passed in here, but the only info I needed from the job model is the task-->SSP.
If there are no startpoints for an SSP, then the task mapped to that SSP will be excluded from the return value.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -173,71 +226,151 @@ public void deleteStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
+  }
+
+  /**
+   * Read the fanned out {@link Startpoint}s for the given {@link TaskName}
+   * @param taskName to read the fan out Startpoints for
+   * @return fanned out Startpoints
+   */
+  public Map<SystemStreamPartition, Startpoint> getFanOutForTask(TaskName taskName) {
```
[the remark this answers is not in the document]
dnishimura: I thought about this as well. The read/write to non-fanout is sort of the public facing API of startpoints and the read/write to fanout store is entirely internal. I agree that separating these two categories into 2 classes makes the responsibilities more explicit.
The counter argument in having it all in one place is that it gives a better bird's eye view of the full startpoint flow.
If I were to create two classes, any suggestions for naming the non-fanout class? The fanout class is obvious of course.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -173,71 +226,151 @@ public void deleteStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
+    Preconditions.checkState(!stopped, "Underlying metadata store not available");
+    Preconditions.checkNotNull(taskName, "TaskName cannot be null");
 
-    return ImmutableSet.copyOf(sspsToDelete);
+    StartpointFanOut.StartpointFanOutSerde startpointFanOutSerde = new StartpointFanOut.StartpointFanOutSerde();
+    byte[] fanOutBytes = fanOutStore.get(toFanOutStoreKey(taskName));
+    if (ArrayUtils.isEmpty(fanOutBytes)) {
+      return null;
```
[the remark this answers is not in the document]
dnishimura: Sure will do

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: Yes if the keys are json object itself are the same. However, makes sense to put stronger guarantees. I'll do that.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
dnishimura: `.stop` is no longer required if we completely remove the metadata store creation from the startpoint manager.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
dnishimura: Only when creating the fan outs. Will add a log line here.

samza-core/src/test/scala/org/apache/samza/checkpoint/TestOffsetManager.scala
dnishimura: Not necessary. Will remove.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: no exceptions are swallowed, so a failure should cause the caller (i.e. job coordinator) to restart. In that case all the fanouts will be recreated.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointFanOutKey.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.google.common.base.MoreObjects;
+import com.google.common.base.Objects;
+import com.google.common.base.Preconditions;
+import org.apache.samza.container.TaskName;
+import org.codehaus.jackson.map.annotate.JsonSerialize;
+
+
+@JsonSerialize(using = StartpointFanOutKeySerializer.class)
+class StartpointFanOutKey {
```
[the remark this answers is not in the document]
cameronlee314: That makes sense for the wrapper object. But will we need to use JSON to serialize then? Could it just be the task name as the key, since the store is namespaced?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -43,64 +45,115 @@
...
-   *  This is primarily used for testing.
+   *  This should only be used for testing.
    */
   @VisibleForTesting
   StartpointManager(MetadataStoreFactory metadataStoreFactory, Config config, MetricsRegistry metricsRegistry) {
     Preconditions.checkNotNull(metadataStoreFactory, "MetadataStoreFactory cannot be null");
     Preconditions.checkNotNull(config, "Config cannot be null");
     Preconditions.checkNotNull(metricsRegistry, "MetricsRegistry cannot be null");
 
+    this.manageMetadataStoreLifecyle = true;
```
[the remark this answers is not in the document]
cameronlee314: `StartpointManager` still creates `NamespaceAwareCoordinatorStreamStore`s, and `NamespaceAwareCoordinatorStreamStore` still has `init`/`close`, so I think you will still need to have `start`/`stop` here.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -43,64 +45,115 @@
...
+    this.fanOutStore = new NamespaceAwareCoordinatorStreamStore(metadataStore, NAMESPACE_FAN_OUT);
+    LOG.info("Lifecycle of metadata store: {} is managed externally", metadataStore.getClass().getCanonicalName());
+    LOG.info("Startpoints are written to namespace: {} and fanned out to namespace: {} in the metadata store", NAMESPACE,
+        NAMESPACE_FAN_OUT);
   }
 
+  /**
+   * Perform startup operations.
+   */
   public void start() {
```
cameronlee314: This is related to a reply on a different comment, but putting directly here too. I think you will need to manage the lifecycle of `fanOutStore` and `readWriteStore`, since you created those within this object.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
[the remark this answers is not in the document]
cameronlee314: Makes sense that it is nice to see the whole flow in one class, but it does still seem like there are at least 2 distinct flows: 1) someone creates a startpoint from outside Samza, 2) Samza needs to interpret and use the startpoint
The only naming suggestion that comes to mind is `StartpointClient`, or something like that. I'm not very tied to it though.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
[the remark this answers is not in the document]
cameronlee314: It'll depend on how you plan to use this method. If it is important to know the tasks which had startpoints, then it seems like it'll be useful to keep the return value.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
cameronlee314: Hm, good point about orphan containers. Seems like those could cause more problems than just this concurrency case too. More generally, is there a plan for how to handle those cases?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
cameronlee314: What if there is a continuous failure after a partial write, so the job coord is not able to finish writing the fanouts? The fanout store would be left in a partial state until the issue is fixed. Would this cause any issue? When the job comes back up, is there a case where the job coord would want to write different fanouts, so that the existing partial ones are no long valid?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointFanOutKey.java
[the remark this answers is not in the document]
dnishimura: True. For consistency, I'll make the fan out key non-JSON as well.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: I don't think there is a way other than detecting the orphan containers and killing them. The orphan containers would continue to checkpoint as well so it's broader issue.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: The partial state would be complete at the subsequent startup of the job coordinator. However, thinking about it more, we probably don't need to do the merge with the existing fan out because the readWriteStore entries would still exist and the fan out's complete state can be rewritten again on the subsequent restart. I'll remove the merge logic.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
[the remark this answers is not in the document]
dnishimura: Also if we were to separate into two classes, the fan out class would depend on the readWrite class to move the startpoints from the readWriteStore to the fanOutStore. Since this refactor would be significant, I'll do it in a follow on PR.

dnishimura:
@cameronlee314 I think I addressed most of your comments. I'm punting on splitting out the `StartpointManager` to a future PR.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -143,14 +171,18 @@ public Startpoint readStartpoint(SystemStreamPartition ssp, TaskName taskName) {
     Preconditions.checkState(!stopped, "Underlying metadata store not available");
```
shanthoosh: Would be a better to return Optional<Startpoint> here, rather than handling null at the caller-end.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -20,87 +20,115 @@
...
   }
 
   /**
-   * Writes a {@link Startpoint} that defines the start position for a {@link SystemStreamPartition}.
-   * @param ssp The {@link SystemStreamPartition} to map the {@link Startpoint} against.
-   * @param startpoint Reference to a Startpoint object.
+   * Perform teardown operations. Method is idempotent.
    */
+  public void stop() {
+    if (!stopped) {
```
shanthoosh: Is this check necessary? Is start() called after stop() in any legitimate use-cases?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -173,71 +205,148 @@ public void deleteStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
+    HashMultimap<SystemStreamPartition, TaskName> deleteKeys = HashMultimap.create();
+    HashMap<TaskName, StartpointFanOutPerTask> fanOuts = new HashMap<>();
+    for (TaskName taskName : taskToSSPs.keySet()) {
+      Set<SystemStreamPartition> ssps = taskToSSPs.get(taskName);
+      if (CollectionUtils.isEmpty(ssps)) {
+        LOG.warn("No SSPs are mapped to taskName: {}", taskName.getTaskName());
+        continue;
+      }
+      for (SystemStreamPartition ssp : ssps) {
+        Startpoint startpoint = readStartpoint(ssp); // Read SSP-only key
```
shanthoosh: Just curious.
Given that the metadata-store returns the last written value for a key, is the deletion of older value even necessary before writing the new value for a key? Wouldn't just over-writing the existing value suffice. Please share your thoughts.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -173,71 +205,148 @@ public void deleteStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
+   * to a "fan out" namespace.
    * This method is not atomic or thread-safe. The intent is for the Samza Processor's coordinator to use this
    * method to assign the Startpoints to the appropriate tasks.
-   * @param jobModel The {@link JobModel} is used to determine which {@link TaskName} each {@link SystemStreamPartition} maps to.
-   * @return The list of {@link SystemStreamPartition}s that were fanned out to SystemStreamPartition+TaskName.
+   * @param taskToSSPs Determines which {@link TaskName} each {@link SystemStreamPartition} maps to.
+   * @return The set of active {@link TaskName}s that were fanned out to.
    */
-  public Set<SystemStreamPartition> fanOutStartpointsToTasks(JobModel jobModel) {
+  public Set<TaskName> fanOut(Map<TaskName, Set<SystemStreamPartition>> taskToSSPs) throws IOException {
```
shanthoosh: Would be better to include the actual startpoint fanned-out for a task in the return value to give more context to the caller. What do you think?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -173,71 +205,148 @@ public void deleteStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
+    return fanOutStore;
+  }
+
+  @VisibleForTesting
+  ObjectMapper getObjectMapper() {
+    return objectMapper;
+  }
+
+  private static Startpoint resolveStartpointPrecendence(Startpoint startpoint1, Startpoint startpoint2) {
+    if (startpoint1 != null && startpoint2 != null) {
```
shanthoosh: Since we've capability to resolve startpoint to offset now, we can even support startpoint with broad-cast stream, by resolving the startpoint to offset and comparing the resolved offsets.  What do you think?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -173,71 +205,148 @@ public void deleteStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
+      if (startpoint1.getCreationTimestamp() > startpoint2.getCreationTimestamp()) {
+        return startpoint1;
+      }
+      return startpoint2;
+    }
+    return startpoint1 != null ? startpoint1 : startpoint2;
+  }
+
+  private static String toReadWriteStoreKey(SystemStreamPartition ssp, TaskName taskName) {
+    String storeKey = ssp.getSystem() + "." + ssp.getStream() + "." + String.valueOf(ssp.getPartition().getPartitionId());
```
shanthoosh: Just to be safe, would be better to add not-null checks to some of the fields before accessing them. What do you think?

samza-api/src/main/java/org/apache/samza/startpoint/Startpoint.java
```diff
@@ -22,11 +22,14 @@
 import com.google.common.base.Objects;
 import java.time.Instant;
 import org.apache.samza.annotation.InterfaceStability;
+import org.codehaus.jackson.annotate.JsonTypeInfo;
+
```
shanthoosh: nit: Minor. Please remove the empty line.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -20,87 +20,115 @@
...
+ * {@link org.apache.samza.checkpoint.OffsetManager} gets the fan outs to set the starting offsets per task and per
+ * {@link SystemStreamPartition}. The fan outs are deleted once the offsets are committed to the checkpoint.
+ *
+ * The read, write and delete methods are intended for external callers.
+ * The fan out methods are intended to be used within a job coordinator.
  */
 public class StartpointManager {
-  private static final Logger LOG = LoggerFactory.getLogger(StartpointManager.class);
-  public static final String NAMESPACE = "samza-startpoint-v1";
+  public static final Integer VERSION = 1;
```
shanthoosh: Does this have to be public. I don't see it used outside of this class.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -143,14 +171,18 @@ public Startpoint readStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
-      Startpoint startpoint = startpointSerde.fromBytes(startpointBytes);
-      if (Instant.now().minus(DEFAULT_EXPIRATION_DURATION).isBefore(Instant.ofEpochMilli(startpoint.getCreationTimestamp()))) {
-        return startpoint; // return if deserializable and if not stale
+    if (ArrayUtils.isNotEmpty(startpointBytes)) {
+      try {
+        Startpoint startpoint = objectMapper.readValue(startpointBytes, Startpoint.class);
+        if (Instant.now().minus(DEFAULT_EXPIRATION_DURATION).isBefore(Instant.ofEpochMilli(startpoint.getCreationTimestamp()))) {
+          return startpoint; // return if deserializable and if not stale
+        }
+        LOG.warn("Stale Startpoint: {} was read. Ignoring.", startpoint);
```
shanthoosh: nit: Minor. Would be better to include the expiration-duration and startpoint-timestamp in the log-message. 

```suggestion
        LOG.warn("Creation timestamp: {} of startpoint: {} has crossed the expiration duration: {}. Ignoring it", startpoint.getTimestamp(), startpoint, DEFAULT_EXPIRATION_DURATION);
```

samza-api/src/main/java/org/apache/samza/startpoint/Startpoint.java
dnishimura: Sure

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: No, will make it private.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: Possibly in an external tool or service. Any reason to disallow this?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: Sure

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: Will do

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: Doesn't hurt to. I'll add it.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: It's not just the older value, but it's marking the startpoints keyed by SSP and SSP+TaskName for deletion after they are moved to the fan out store. If we don't delete them, the fan out will happen again on subsequent restarts. Once the startpoints are moved to the fan out, they are no longer valid in the readWriteStore.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: I'm a little confused. Isn't the offset resolution done in the offset manager? The resolution here is the resolution between two startpoints, not a startpoint and the resolved offset. Also how would we support startpoint with broadcast here? 
In a follow on PR, I'll be integrating the fan out with the JobCoordinators to use the info in the job model to fan out the startpoints, which should take care of broadcast streams. Once the startpoints are fanned out, the actual offset resolution itself happens in the offset manager.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: Sure better safe than sorry.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
cameronlee314: At least it seems like an orphan container won't be able to steal a startpoint since it is already started up.

samza-core/src/test/java/org/apache/samza/startpoint/TestStartpointFanOutPerTaskSerde.java
```diff
@@ -0,0 +1,46 @@
...
+import java.io.IOException;
+import java.time.Instant;
+import org.apache.samza.Partition;
+import org.apache.samza.system.SystemStreamPartition;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.junit.Assert;
+import org.junit.Test;
+
+
+public class TestStartpointFanOutPerTaskSerde {
```
cameronlee314: Maybe name this `TestStartpointObjectMapper`?

samza-core/src/test/java/org/apache/samza/startpoint/TestStartpointSerde.java
```diff
@@ -18,45 +18,47 @@
  */
 package org.apache.samza.startpoint;
 
+import java.io.IOException;
+import org.codehaus.jackson.map.ObjectMapper;
 import org.junit.Assert;
 import org.junit.Test;
 
 public class TestStartpointSerde {
```
cameronlee314: Maybe combine this with the other startpoint object mapper tests, since there isn't an actual `StartpointSerde` class anymore.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointObjectMapper.java
```diff
@@ -39,12 +40,16 @@ static ObjectMapper getObjectMapper() {
...
     module.addDeserializer(Instant.class, new CustomInstantDeserializer());
-
     objectMapper.registerModule(module);
-    objectMapper.registerSubtypes(StartpointSpecific.class);
-    objectMapper.registerSubtypes(StartpointTimestamp.class);
-    objectMapper.registerSubtypes(StartpointUpcoming.class);
-    objectMapper.registerSubtypes(StartpointOldest.class);
+
+    // 1. To support polymorphism for serialization, the Startpoint subtypes must be registered here.
+    // 2. The NamedType container class provides a logical name as an external identifier so that the full canonical
```
cameronlee314: Just double checking: Does the "logical name" allow for easily adding new types in the future? Is there a certain way new types need to be added? For example, if the "logical name" is just a counter, then we would need to add a new subtype at the end of the list.

dnishimura: Not necessarily easier, but it seems to be a best practice based on the examples I've seen. Underneath the hood, it just uses the simple class name and not the full canonical name, so ordering doesn't matter. The `NamedType` container constructor also has a 2nd optional parameter where we can specify the logical name ourselves too.

cameronlee314: Sounds good. Just wanted to make sure there wasn't any gotcha with using the new way. I didn't need it to be easier than the old way. It's good that it makes the serialization more compact.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
shanthoosh: Generally `StartpointManager.stop` is supposed to be called during the end of lifecycle of compoent that manages the `StartpointManager`. Even if `StartpointManager` is stopped, the component that owns it could recreate a new-one rather than invoking start() again. The implicit assumption here is that the metastore implementations can be started after they're stopped, which might not be true.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -173,71 +207,148 @@ public void deleteStartpoint(SystemStreamPartition ssp, TaskName taskName) {
...
+    }
+
+    LOG.info("Fanning out to {} tasks", fanOuts.size());
 
-    return ImmutableSet.copyOf(sspsToDelete);
+    // Fan out to store
+    for (TaskName taskName : fanOuts.keySet()) {
+      String fanOutKey = toFanOutStoreKey(taskName);
+      StartpointFanOutPerTask newFanOut = fanOuts.get(taskName);
+      fanOutStore.put(fanOutKey, objectMapper.writeValueAsBytes(newFanOut));
```
shanthoosh: Consider this scenario.

1. Let's say a samza job consumes from a single stream that has 10000 partitions. This would make the `JobModel` of the job to have 10000 tasks .
2. Let's say user wants to consume all the `SystemStreamPartition's` from a specific timestamp after he restarts that particular samza job.
3. User publishes a startpoint with timestamp set at the stream level.
4. After reading the published `StartpointTimestamp` from `Namespace-1` of `MetadataStore`, we would be publishing around 10000 individual sequential writes to the `Namespace-2` to `MetadataStore`. This will entail 10000 RTT's to the storage layer. Rather than making one batch write, we're making N+1 writes(where N is the number of topic partitions consumed by the samza job). 
5. Since this fan-out of `SystemStreamPartition's` will happen in the samza `ApplicationMaster` startup sequence, this will incur significant delay in the allocation of containers by `ApplicationMaster` and increase the messages behind high-watermark. Higher the number of topic partitions consumed by the samza job, more the processing delay.

The existing metadata-store API doesn't support batch puts and hence you had to opt for individual fanned-out `SystemStreamPartition` writes. 

What do you think? Do you agree that this will cause performance issues?

dnishimura: Yes unfortunately the `CoordinatorMetadataStore` does a flush per entry write, so it will suffer a performance hit. However, it's a general problem. For example, `ApplicationMaster` sequence writes the container to task map entry per task:
```JobModelManager$ [INFO] Storing task: Partition 799 and container ID: 0 into metadata store```
as well as the SSP to task map entry per task per SSP:```JobModelManager$ [INFO] Storing ssp: SystemStreamPartition [queuing, StartpointTestInput, 799] and task: [Partition 799] into metadata store``` at every start sequence, not just the first. It's an issue that we should probably address on the metadata store via batch puts or some other mechanism.

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
dnishimura: No real use cases I can think of other than a long running service using the StartpointManager wanting to clean up resources when not in use, but I guess they can get around it by creating a new instance.
If we want to prevent the StartpointManager from being usable after a `stop()` then I can remove the `start()` method all together and do the initialization within the constructor only. Long-term, do you think it's ok to impose this restriction?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
shanthoosh: I will create a follow-up JIRA with ask for multiPut in metadata-store and we can add these use-cases in it.

shanthoosh:
@dnishimura Thanks for the changes. Merged to trunk.

PR #1029, opened 2019-05-10
Config logger

sborya:
This class logs a config using a separate logger line by line, limiting each line to a specified size.


samza-core/src/main/java/org/apache/samza/config/ConfigLogger.java
```diff
@@ -0,0 +1,71 @@
...
+    String msg = buildConfigString(header, maxLen, config);
+    log(logger, msg, level);
+  }
+
+  private static String buildConfigString(String header, int maxLen, Config config) {
+    StringBuilder bldr = new StringBuilder(header + "\n");
+    for (MapConfig.Entry e: config.entrySet()) {
+      String val;
+      if (e.getValue().toString().length() > maxLen) {
+        val = e.getValue().toString().substring(0, maxLen) + "<...>";
```
Sanil15: I am trying to understand the use case, does log4j does not automatically split lines that are too long?

If not maxLen can't default?

samza-core/src/main/java/org/apache/samza/config/ConfigLogger.java
```diff
@@ -0,0 +1,71 @@
...
+    switch (level) {
+      case INFO:
+        logger.info(msg); break;
+      case DEBUG:
+        logger.debug(msg); break;
+      case WARN:
+        logger.warn(msg); break;
+      case ERROR:
+        logger.error(msg); break;
+    }
```
rmatharu: Probably wanna add Default here (in case someone is trying to use trace or fatal).

samza-core/src/main/java/org/apache/samza/config/ConfigLogger.java
```diff
@@ -0,0 +1,71 @@
...
+
+  // static api
+  public static void logConfig(String loggerName, String header, Config config, int maxLen, Level level) {
+    Logger logger = LoggerFactory.getLogger("ConfigLogger." + loggerName);
+    String msg = buildConfigString(header, maxLen, config);
+    log(logger, msg, level);
+  }
+
+  private static String buildConfigString(String header, int maxLen, Config config) {
+    StringBuilder bldr = new StringBuilder(header + "\n");
```
rmatharu: System.lineSeparator()

samza-core/src/main/java/org/apache/samza/config/ConfigLogger.java
```diff
@@ -0,0 +1,71 @@
...
+   * @param level at what level {@ConfigLogger.Leevel}
+   */
+  public void logConfig(String header, Config config, Level level) {
+    String msg = buildConfigString(header, maxLen, config);
+    log(log, msg, level);
+  }
+
+
+  // static api
+  public static void logConfig(String loggerName, String header, Config config, int maxLen, Level level) {
```
rmatharu: Perhaps should also be a private method, called by logConfig(String header, Config config, Level level)

samza-core/src/main/java/org/apache/samza/config/ConfigLogger.java
```diff
@@ -0,0 +1,71 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.function.Consumer;
+
+/**
+ * This logger logs the config line by line, while shortening the lines which are too long (over maxLen)
+ * Logging preserves the original logger name and may be redirected to a different file
+ * by specifying a logger name that starts with ConfigLogger. in log4j configuration.
+ */
+public class ConfigLogger {
```
prateekm: What's the purpose of this class? Where is this used?

I'd strongly prefer to not do this. IMHO it'll be much better to log these things (config, job model, starting offsets) as JSON instead, so that the log representation is compact but we can copy and "pretty print" in the IDE / browser for explorability.

PR #1030, opened 2019-05-14
SAMZA-2192: Add StartpointVisitor implementation for EventHub.

shanthoosh:


shanthoosh:
@dnishimura 

samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
```diff
@@ -205,4 +234,75 @@ public static Integer compareOffsets(String offset1, String offset2) {
...
+
+        // 2. Create a partition receiver with event position defined by the timestamp.
+        partitionReceiver = eventHubClient.createReceiverSync(consumerGroup, partitionId, eventPosition);
+
+        // 3. Read a single message from the partition receiver.
+        Iterable<EventData> eventHubMessagesIterator = partitionReceiver.receiveSync(1);
+        ArrayList<EventData> eventHubMessageList = Lists.newArrayList(eventHubMessagesIterator);
+
+        // 4. Validate that a single message was fetched from the broker.
+        Preconditions.checkState(eventHubMessageList.size() == 1, "Failed to read messages from EventHub system.");
```
dnishimura: Does it make sense to fail if the partition is empty?

samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
```diff
@@ -205,4 +234,75 @@ public static Integer compareOffsets(String offset1, String offset2) {
...
+        Instant epochInMillisInstant = Instant.ofEpochMilli(startpointTimestamp.getTimestampOffset());
+        EventPosition eventPosition = EventPosition.fromEnqueuedTime(epochInMillisInstant);
+        String consumerGroup = eventHubConfig.getStreamConsumerGroup(systemStreamPartition.getSystem(), streamName);
+
+        // 2. Create a partition receiver with event position defined by the timestamp.
+        partitionReceiver = eventHubClient.createReceiverSync(consumerGroup, partitionId, eventPosition);
+
+        // 3. Read a single message from the partition receiver.
+        Iterable<EventData> eventHubMessagesIterator = partitionReceiver.receiveSync(1);
+        ArrayList<EventData> eventHubMessageList = Lists.newArrayList(eventHubMessagesIterator);
```
dnishimura: nit: Do you need to create a new array list or can you use the `.hasNext()` and `.next()` methods from the iterator since it'll either be one or none?

shanthoosh: There's no big performance cost associated with creating a array-list here. It makes the code necessary to handle the validations and message-access simple.

samza-azure/src/main/java/org/apache/samza/system/eventhub/admin/EventHubSystemAdmin.java
shanthoosh: 1. My initial idea was to fall back to either oldest or newest offset upon startpoint resolution failure. However, depending upon the timestamp that the user wants to seek to for a partition doing either of them could be incorrect. Let's say user  wants to seek to a timestamp closer to beginning-offset and offset for it doesn't exist. If we default it to newest, then it will be diametrically opposite to what the user desires(or vice-versa).
2. In case of startpoint resolution failure, we fall back to the checkpointed-offset in `OffsetManager` essentially ignoring the startpoint. 
3. Ideal solution should allow the user to define the default.offset policy in case of startpoint resolution failures and use it in the `OffsetManager`. But doing it before hand, until we have concrete use-cases or asks might be unfruitful. 

What do you think?

dnishimura: Cool it should be ok as long as it falls back to the checkpoint offset. Thanks.

PR #1038, opened 2019-05-17
SAMZA-2196: [Scala cleanup] Convert SerializerConfig to Java 

cameronlee314:
There are two commits:
1) Move all usage of SerializerConfig.scala to JavaSerializerConfig.java
2) Rename JavaSerializerConfig.java to SerializerConfig.java

It might be easier to initially look at just the first commit to see the more meaningful changes.


samza-core/src/main/java/org/apache/samza/config/SerializerConfig.java
```diff
@@ -0,0 +1,108 @@
...
+   * throws an exception.
+   */
+  public static String getSerdeFactoryName(String serdeName) {
+    String serdeFactoryName = doGetSerdeFactoryName(serdeName);
+    LOGGER.info(String.format("Using default serde %s for serde name %s", serdeFactoryName, serdeName));
+    return serdeFactoryName;
+  }
+
+  /**
+   * @param name name of the serde in the config
```
prateekm: Nitpick: Add description for javadoc.

samza-core/src/main/java/org/apache/samza/config/SerializerConfig.java
```diff
@@ -0,0 +1,108 @@
...
+   * @return serde factory class name for {@code name}; empty if no class was found in the config
+   */
+  public Optional<String> getSerdeClass(String name) {
+    return Optional.ofNullable(get(String.format(SERDE_FACTORY_CLASS, name), null));
+  }
+
+  /**
+   * Useful for getting individual serializers.
```
prateekm: "Gets all serializer names defined in configuration." (or can remove somewhat obvious javadocs)

samza-core/src/main/java/org/apache/samza/config/SerializerConfig.java
```diff
@@ -0,0 +1,108 @@
...
+   * @return a list of all serializer names from the config file
```
prateekm: s/config file/config/

samza-core/src/main/java/org/apache/samza/config/SerializerConfig.java
```diff
@@ -0,0 +1,108 @@
...
+        results.add(key.replace(".class", ""));
+      }
+    }
+    return results;
+  }
+
+  private static String doGetSerdeFactoryName(String serdeName) {
+    switch (serdeName) {
+      case "byte":
+        return ByteSerdeFactory.class.getCanonicalName();
```
prateekm: Use 'Class.getName()' for class names to be used for classloading. Canonical names may not be unique within the classloader. E.g., https://stackoverflow.com/questions/15202997/what-is-the-difference-between-canonical-name-simple-name-and-class-name-in-jav

samza-core/src/main/java/org/apache/samza/config/SerializerConfig.java
```diff
@@ -0,0 +1,108 @@
...
+    super(config);
+  }
+
+  /**
+   * Returns the pre-defined serde factory class name for the provided serde name. If no pre-defined factory exists,
+   * throws an exception.
+   */
+  public static String getSerdeFactoryName(String serdeName) {
+    String serdeFactoryName = doGetSerdeFactoryName(serdeName);
+    LOGGER.info(String.format("Using default serde %s for serde name %s", serdeFactoryName, serdeName));
```
prateekm: Nitpick: Use {} for logger instead of string format.

samza-core/src/main/java/org/apache/samza/config/SerializerConfig.java
```diff
@@ -0,0 +1,108 @@
...
+  public SerializerConfig(Config config) {
+    super(config);
+  }
+
+  /**
+   * Returns the pre-defined serde factory class name for the provided serde name. If no pre-defined factory exists,
+   * throws an exception.
+   */
+  public static String getSerdeFactoryName(String serdeName) {
+    String serdeFactoryName = doGetSerdeFactoryName(serdeName);
```
prateekm: This behavior/naming seems weird. Are we missing a method to get the serde factory name for a non-default serde? Is that what getSerdeClass below does? If so, maybe worth renaming one or both of these.

samza-core/src/main/java/org/apache/samza/config/SerializerConfig.java
cameronlee314: Sounds good. The previous Scala code used `getCanonicalName`, but seems reasonable to just make the change here.

samza-core/src/main/java/org/apache/samza/config/SerializerConfig.java
cameronlee314: I was just migrating existing method names. `getSerdeClass` does get the non-default factory names. I'll try to think of some better method names.

PR #1039, opened 2019-05-18
SAMZA-2197: [Scala cleanup] Convert TaskConfig to Java 

cameronlee314:
The first commit moves methods from TaskConfig.scala to TaskConfigJava.java.
The second commit moves constants from TaskConfig.scala to TaskConfigJava.java.
The third commit renames TaskConfigJava.java to TaskConfig.scala (it might be desirable to just look at the diffs for the 1st/2nd commit).


samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
```diff
@@ -0,0 +1,298 @@
...
+import org.apache.samza.system.SystemStream;
+import org.apache.samza.system.SystemStreamPartition;
+import org.apache.samza.util.StreamUtil;
+import org.apache.samza.util.Util;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Helper for accessing config values related to tasks.
```
prateekm: Can remove, this is not true since task scoped configs unfortunately contain many job related configs.

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
```diff
@@ -0,0 +1,298 @@
...
+  public static final String DROP_DESERIALIZATION_ERRORS = "task.drop.deserialization.errors";
+  // define whether drop the messages or not when serialization fails
+  public static final String DROP_SERIALIZATION_ERRORS = "task.drop.serialization.errors";
+  // whether to ignore producer errors and drop the messages that failed to send
+  public static final String DROP_PRODUCER_ERRORS = "task.drop.producer.errors";
+  // exceptions to ignore in process and window
+  public static final String IGNORED_EXCEPTIONS = "task.ignored.exceptions";
+  // class name for task grouper
+  public static final String GROUPER_FACTORY = "task.name.grouper.factory";
+  // max number of concurrent process for a AsyncStreamTask
```
prateekm: Actually applies to StreamTask as well. Maybe "max number of messages to process concurrently"

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
```diff
@@ -0,0 +1,298 @@
...
+  public static final String MAX_CONCURRENCY = "task.max.concurrency";
+  static final int DEFAULT_MAX_CONCURRENCY = 1;
+  // timeout period for triggering a callback
```
prateekm: s/period//

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
```diff
@@ -0,0 +1,298 @@
...
+  // how long to wait for a clean shutdown
+  public static final String TASK_SHUTDOWN_MS = "task.shutdown.ms";
+  static final long DEFAULT_TASK_SHUTDOWN_MS = 30000L;
+  // legacy config for specifying task class; replaced by SamzaApplication and app.class
+  public static final String TASK_CLASS = "task.class";
+  // command builder to use for launching a Samza job
+  public static final String COMMAND_BUILDER = "task.command.class";
+  // message chooser for controlling stream consumption
+  public static final String MESSAGE_CHOOSER_CLASS_NAME = "task.chooser.class";
+  // define whether drop the messages or not when deserialization fails
```
prateekm: Nitpick: whether to drop (same for next line)

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
```diff
@@ -0,0 +1,298 @@
...
+   * Decreasing this value means that empty SystemStreamPartitions are
+   * refreshed more frequently, thereby introducing less latency, but
+   * increasing CPU and network utilization.
+   */
+  public static final String POLL_INTERVAL_MS = "task.poll.interval.ms";
+  // broadcast streams consumed by all tasks. e.g. kafka.foo#1
+  public static final String BROADCAST_INPUT_STREAMS = "task.broadcast.inputs";
+  private static final String BROADCAST_STREAM_PATTERN = "^[\\d]+$";
+  private static final String BROADCAST_STREAM_RANGE_PATTERN = "^\\[[\\d]+\\-[\\d]+\\]$";
+  // class name to use when sending offset checkpoints
```
prateekm: Can remove.

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
```diff
@@ -0,0 +1,298 @@
...
+    return getLong(MAX_IDLE_MS, DEFAULT_MAX_IDLE_MS);
+  }
+
+  /**
+   * Create the checkpoint manager
+   *
+   * @param metricsRegistry Registry of metrics to use. Can be null if not using metrics.
+   * @return CheckpointManager object if checkpoint manager factory is configured, otherwise empty.
+   */
+  public Optional<CheckpointManager> getCheckpointManager(MetricsRegistry metricsRegistry) {
```
prateekm: Don't need to do right now, but what do you think about returning instances from config objects vs only returning classnames and letting the caller do the classloading? I'd prefer the latter.

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
```diff
@@ -0,0 +1,298 @@
...
+    // Initialize checkpoint streams during job coordination
```
prateekm: Incorrect comment, can remove.

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
```diff
@@ -0,0 +1,298 @@
...
+    }
+    return systemStreamPartitionSet;
+  }
+
+  /**
+   * Get the SystemStreams for the configured broadcast streams.
+   *
+   * @return the set of SystemStreams for which there are broadcast stream SSPs configured.
+   */
+  public Set<SystemStream> getBroadcastSystemStreams() {
```
prateekm: Up to you if you want to do this now, but I'd prefer to do this at the call sites and remove this method. I'd prefer Config classes to have one way of doing things, and ideally be logic-less. Multiple similar methods cause confusion.

prateekm: Case in point: getInputStreams vs getAllInputStreams below.

samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
```diff
@@ -173,13 +173,12 @@ class CheckpointTool(newOffsets: TaskNameToCheckpointMap, coordinatorStreamStore
...
+    val taskConfig = new TaskConfig(combinedConfig)
     // Instantiate the checkpoint manager with coordinator stream configuration.
-    val checkpointManager: CheckpointManager = combinedConfig.getCheckpointManagerFactory() match {
-      case Some(className) =>
-        Util.getObj(className, classOf[CheckpointManagerFactory])
-          .getCheckpointManager(combinedConfig, new MetricsRegistryMap)
-      case _ =>
-        throw new SamzaException("Configuration: task.checkpoint.factory is not defined.")
+    val checkpointManager: CheckpointManager =
+      JavaOptionals.toRichOptional(taskConfig.getCheckpointManager(new MetricsRegistryMap)).toOption match {
```
prateekm: Would prefer to remove this RichOptional / JavaOptionals class / transformation from ScalaJavaUtil. I don't think it's adding any significant functionality, and is adding yet another "Option" class to keep track of.

For this PR, it's OK to just not use it. It's also OK to use it if you think this will be easy to refactor away later.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -401,7 +399,8 @@ object SamzaContainer extends Logging {
 
     info("Setting up message chooser.")
 
-    val chooserFactoryClassName = config.getMessageChooserClass.getOrElse(classOf[RoundRobinChooserFactory].getName)
+    val taskConfig = new TaskConfig(config)
+    val chooserFactoryClassName = taskConfig.getMessageChooserClass.orElse(classOf[RoundRobinChooserFactory].getName)
```
prateekm: Minor: Maybe make this the default value in the class itself?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -422,11 +421,8 @@ object SamzaContainer extends Logging {
     }
     info("Got security manager: %s" format securityManager)
 
-    val checkpointManager = config.getCheckpointManagerFactory()
-      .filterNot(_.isEmpty)
-      .map(Util.getObj(_, classOf[CheckpointManagerFactory])
-        .getCheckpointManager(config, samzaContainerMetrics.registry))
-      .orNull
+    val checkpointManager =
```
prateekm: Does Optional#orElse(null) work?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -437,13 +433,10 @@ object SamzaContainer extends Logging {
...
-    val dropDeserializationError = config.getDropDeserializationErrors
-    val dropSerializationError = config.getDropSerializationErrors
+    val dropDeserializationError = taskConfig.getDropDeserializationErrors
+    val dropSerializationError = taskConfig.getDropSerializationErrors
 
-    val pollIntervalMs = config
-      .getPollIntervalMs
-      .getOrElse(SystemConsumers.DEFAULT_POLL_INTERVAL_MS.toString)
-      .toInt
+    val pollIntervalMs = taskConfig.getPollIntervalMs.orElse(SystemConsumers.DEFAULT_POLL_INTERVAL_MS)
```
prateekm: Minor: Move getOrElse(default) behavior to config class.

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
```diff
@@ -0,0 +1,298 @@
...
+  public static final String IGNORED_EXCEPTIONS = "task.ignored.exceptions";
+  // class name for task grouper
+  public static final String GROUPER_FACTORY = "task.name.grouper.factory";
+  // max number of concurrent process for a AsyncStreamTask
+  public static final String MAX_CONCURRENCY = "task.max.concurrency";
+  static final int DEFAULT_MAX_CONCURRENCY = 1;
+  // timeout period for triggering a callback
+  public static final String CALLBACK_TIMEOUT_MS = "task.callback.timeout.ms";
+  static final long DEFAULT_CALLBACK_TIMEOUT_MS = -1L;
+  // to enable async commit in a AsyncStreamTask
```
prateekm: s/to enable/enable/

I think this applies to StreamTask with task.max.concurrency as well? I can't find it in the config reference, maybe check with Shanthoosh and update.

samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
cameronlee314: Ah, yes, it can be removed here. I thought that Scala functions didn't work cleanly with Java classes, but it's not a problem here. I think there are still some places other where JavaOptionals might be useful, since sometimes it is nicer to use Scala syntax when in a Scala class.

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
cameronlee314: If `task.max.concurrency` works with `StreamTask`, then it looks like this would too. I'll update it.

samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
cameronlee314: Oops, I take that last comment back. I guess Scala functions aren't compatible with Java functions.
For example, `val supplier: Supplier[String] = () => "hello world"` doesn't actually compile (IDE doesn't show the red squiggle, but build actually will have an error).
So having an `Option` will allow for more natural syntax usage in Scala.

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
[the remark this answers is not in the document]
cameronlee314: There are currently a few different places where object instantiations are done in config objects. Maybe we could have a separate clean-up ticket to handle all of those at once.
In the cases where we have similar flows (e.g. 1) get checkpoint manager factory class, 2) create the checkpoint manager factory instance, 3) create checkpoint manager), would you prefer to have a utility helper method instead of using the config object directly?

samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
```diff
@@ -158,14 +159,12 @@ class CheckpointTool(newOffsets: TaskNameToCheckpointMap, coordinatorStreamStore
...
     // Instantiate the checkpoint manager with coordinator stream configuration.
-    val checkpointManager: CheckpointManager = combinedConfig.getCheckpointManagerFactory() match {
-      case Some(className) =>
-        ReflectionUtil.getObj(classLoader, className, classOf[CheckpointManagerFactory])
-          .getCheckpointManager(combinedConfig, new MetricsRegistryMap)
-      case _ =>
-        throw new SamzaException("Configuration: task.checkpoint.factory is not defined.")
-    }
+    val checkpointManager: CheckpointManager =
+      JavaOptionals.toRichOptional(taskConfig.getCheckpointManager(new MetricsRegistryMap, classLoader))
```
prateekm: Minor: Can we just call Optional#orElseThrow instead?

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -262,15 +262,14 @@ object JobModelManager extends Logging {
   }
 
   /**
-    * Computes the input system stream partitions of a samza job using the provided {@param config}
+    * Computes the input system stream partitions of a samza job using the provided {@param taskConfig}
```
prateekm: Revert change.

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -80,7 +81,7 @@ object Util extends Logging {
   def getTaskClassVersion(config: Config): String = {
     try {
       val taskClass = Option(new ApplicationConfig(config).getAppClass())
```
prateekm: An if-else would be a lot more readable :)

samza-test/src/test/scala/org/apache/samza/test/integration/StreamTaskTestUtil.scala
```diff
@@ -273,14 +274,13 @@ class StreamTaskTestUtil {
...
-    checkpointManager match {
-      case Some(checkpointManager) => {
-        checkpointManager.createResources
-        checkpointManager.stop
-      }
-      case _ => assert(checkpointManager != null, "No checkpoint manager factory configured")
+    val taskConfig = new TaskConfig(jobModel.getConfig)
+    JavaOptionals.toRichOptional(taskConfig.getCheckpointManager(new MetricsRegistryMap(), getClass.getClassLoader))
+      .toOption match {
+      case Some(checkpointManager) =>
```
prateekm: Nitpick: formatting / indentation.

samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
cameronlee314: I tried that, but the Scala syntax for a 0-arg method isn't compatible with Java `Supplier`.
For example, in Scala, `val supplier: Supplier[String] = () => "hello world"` is a "type mismatch" error.

PR #1040, opened 2019-05-18
SAMZA-2199: [Scala cleanup] Convert MetricsConfig to Java

cameronlee314:


prateekm:
@rmatharu can you take a look at your convenience.

samza-core/src/main/java/org/apache/samza/util/DiagnosticsUtil.java
```diff
@@ -122,6 +122,8 @@ public MetadataFileContents(String version, String metricsSnapshot) {
       DiagnosticsManager diagnosticsManager = new DiagnosticsManager(jobName, jobId, containerId, execEnvContainerId.orElse(""), taskClassVersion,
           samzaVersion, hostName, diagnosticsSystemStream, systemProducer, Duration.ofMillis(new TaskConfig(config).getShutdownMs()));
 
+      Option<String> blacklist = ScalaJavaUtil.JavaOptionals$.MODULE$.toRichOptional(
```
rmatharu: This can just be 
ScalaJavaUtil.JavaOptionals.toRichOptional(metricsConfig.getMetricsSnapshotReporterBlacklist(diagnosticsReporterName)).toOption()
right?

cameronlee314: That's what I tried initially, but the IDE complains that it can't find `JavaOptionals` when doing that. Looks like this `MODULE$` syntax is used in some other places too.

vjagadish1989: can this helper that converts `optional`s between java<->scala be added to one of our util classes?

samza-core/src/main/scala/org/apache/samza/metrics/reporter/MetricsSnapshotReporterFactory.scala
```diff
@@ -83,12 +83,11 @@ class MetricsSnapshotReporterFactory extends MetricsReporterFactory with Logging
...
     info("Got serde %s." format serde)
 
-    val pollingInterval: Int = config
-      .getMetricsSnapshotReporterInterval(name)
+    val pollingInterval: Int = metricsConfig.getMetricsSnapshotReporterInterval(name)
 
     info("Setting polling interval to %d" format pollingInterval)
 
-    val blacklist = config.getMetricsSnapshotReporterBlacklist(name)
+    val blacklist = JavaOptionals.toRichOptional(metricsConfig.getMetricsSnapshotReporterBlacklist(name)).toOption
```
vjagadish1989: possibly make this a util?

cameronlee314: `JavaOptionals` is already a util. I guess it just looks pretty verbose.

PR #1042, opened 2019-05-20
SAMZA-2201 : Startpoints - Integrate fan out with job coordinators

dnishimura:
This PR integrates the Startpoint fan out implemented in PR #1027.

Supported job coordinators:
- `ClusterBasedJobCoordinator`
- `ZkJobCoordinator`

Unsupported job coordinators:
- `PassthroughJobCoordinator` - Offset management is external to Samza.
- `AzureJobCoordinator` - Currently, the coordination logic is broken. Will link ticket here with explanation.

Next steps after this PR: Investigate if it makes sense to separate fan out logic from the `StartpointManager`

@shanthoosh @cameronlee314 - Please take a look when you get a chance.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
+++ b/samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
@@ -50,6 +50,7 @@ import org.apache.samza.job.model.TaskModel;
 import org.apache.samza.metrics.JmxServer;
 import org.apache.samza.metrics.MetricsRegistryMap;
 import org.apache.samza.serializers.model.SamzaObjectMapper;
+import org.apache.samza.startpoint.StartpointManager;
 import org.apache.samza.storage.ChangelogStreamManager;
 import org.apache.samza.system.StreamMetadataCache;
 import org.apache.samza.system.SystemAdmins;
@@ -194,7 +195,7 @@ public class ClusterBasedJobCoordinator {
     jobCoordinatorSleepInterval = clusterManagerConfig.getJobCoordinatorSleepInterval();
 
     // build a container process Manager
-    containerProcessManager = new ContainerProcessManager(config, state, metrics);
+    containerProcessManager = createContainerProcessManager();
   }
 
   /**
@@ -216,14 +217,23 @@ public class ClusterBasedJobCoordinator {
     }
 
     try {
-      //initialize JobCoordinator state
+      // initialize JobCoordinator state
       LOG.info("Starting cluster based job coordinator");
 
-      //create necessary checkpoint and changelog streams, if not created
+      // create necessary checkpoint and changelog streams, if not created
       JobModel jobModel = jobModelManager.jobModel();
       MetadataResourceUtil metadataResourceUtil =
           new MetadataResourceUtil(jobModel, metrics);
       metadataResourceUtil.createResources();
+
+      // fan out the startpoints
+      StartpointManager startpointManager = createStartpointManager();
+      startpointManager.start();
+      try {
+        startpointManager.fanOut(jobModel.getTaskToSystemStreamPartitions());
+      } finally {
+        startpointManager.stop();
+      }
 
       // Remap changelog partitions to tasks
       Map<TaskName, Integer> prevPartitionMappings = changelogStreamManager.readPartitionMapping();
@@ -374,6 +384,16 @@ public class ClusterBasedJobCoordinator {
   @VisibleForTesting
   StreamPartitionCountMonitor getPartitionMonitor() {
     return partitionMonitor.get();
+  }
+
+  @VisibleForTesting
+  StartpointManager createStartpointManager() {
+    return new StartpointManager(coordinatorStreamStore);
+  }
+
+  @VisibleForTesting
+  ContainerProcessManager createContainerProcessManager() {
+    return new ContainerProcessManager(config, state, metrics);
   }
 
   /**
--- a/samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
+++ b/samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
@@ -231,7 +231,7 @@ public class ClusterBasedJobCoordinator {
       StartpointManager startpointManager = createStartpointManager();
       startpointManager.start();
       try {
-        startpointManager.fanOut(new JobModelUtil(jobModel).getTaskToSystemStreamPartitions());
+        startpointManager.fanOut(JobModelUtil.getTaskToSystemStreamPartitions(jobModel));
       } finally {
         startpointManager.stop();
       }
--- a/samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
+++ b/samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
@@ -46,6 +46,7 @@ import org.apache.samza.coordinator.metadatastore.NamespaceAwareCoordinatorStrea
 import org.apache.samza.coordinator.stream.messages.SetChangelogMapping;
 import org.apache.samza.job.model.ContainerModel;
 import org.apache.samza.job.model.JobModel;
+import org.apache.samza.job.model.JobModelUtil;
 import org.apache.samza.job.model.TaskModel;
 import org.apache.samza.metrics.JmxServer;
 import org.apache.samza.metrics.MetricsRegistryMap;
@@ -230,7 +231,7 @@ public class ClusterBasedJobCoordinator {
       StartpointManager startpointManager = createStartpointManager();
       startpointManager.start();
       try {
-        startpointManager.fanOut(jobModel.getTaskToSystemStreamPartitions());
+        startpointManager.fanOut(new JobModelUtil(jobModel).getTaskToSystemStreamPartitions());
       } finally {
         startpointManager.stop();
       }
--- a/samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
+++ b/samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
@@ -19,12 +19,18 @@
 
 package org.apache.samza.job.model;
 
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 import org.apache.samza.config.Config;
 import org.apache.samza.container.LocalityManager;
+import org.apache.samza.container.TaskName;
 import org.apache.samza.coordinator.stream.messages.SetContainerHostMapping;
+import org.apache.samza.system.SystemStreamPartition;
+
 
 /**
  * <p>
@@ -117,6 +123,30 @@ public class JobModel {
       }
     }
     return allValues;
+  }
+
+  /**
+   * Extracts the map of {@link SystemStreamPartition}s to {@link TaskName} from the underlying {@link TaskModel}s
+   *
+   * @return the extracted map
+   */
+  public Map<TaskName, Set<SystemStreamPartition>> getTaskToSystemStreamPartitions() {
+    HashMap<TaskName, Set<SystemStreamPartition>> taskToSSPs = new HashMap<>();
+    for (String containerID : containers.keySet()) {
+      ContainerModel containerModel = containers.get(containerID);
+      for (TaskName taskName : containerModel.getTasks().keySet()) {
+        TaskModel taskModel = containerModel.getTasks().get(taskName);
+        if (taskModel.getTaskMode() != TaskMode.Active) {
+          // Avoid duplicate tasks
+          continue;
+        }
+        ImmutableSet.Builder<SystemStreamPartition> setOfSSPs = ImmutableSet.builder();
+        setOfSSPs.addAll(taskToSSPs.getOrDefault(taskName, ImmutableSet.of()));
+        setOfSSPs.addAll(taskModel.getSystemStreamPartitions());
+        taskToSSPs.put(taskName, setOfSSPs.build());
+      }
+    }
+    return ImmutableMap.copyOf(taskToSSPs);
   }
 
   private void populateContainerLocalityMappings() {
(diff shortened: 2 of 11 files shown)


samza-core/src/test/java/org/apache/samza/zk/TestZkJobCoordinator.java
```diff
@@ -200,4 +212,49 @@ public void testShouldStopPartitionCountMonitorWhenStoppingTheJobCoordinator() {
 
     Mockito.verify(monitor).stop();
   }
+
+  @Test
+  public void testStartpointFanOutOnBecomingLeader() throws IOException {
```
cameronlee314: Is there a test for the `loadMetadataResources` flow? Could the startpoint fanout flow be checked in that test?

samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
```diff
@@ -119,6 +125,30 @@ public String getContainerToHostValue(String containerId, String key) {
     return allValues;
   }
 
+  /**
+   * Extracts the map of {@link SystemStreamPartition}s to {@link TaskName} from the underlying {@link TaskModel}s
+   *
+   * @return the extracted map
+   */
+  public Map<TaskName, Set<SystemStreamPartition>> getTaskToSystemStreamPartitions() {
```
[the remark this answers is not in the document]
dnishimura: I considered that, but thought it'd be strange to have a util class with a single method. However, it makes more sense now that the plan is to move `JobModel` to `samza-api`.

samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
```diff
@@ -119,6 +125,30 @@ public String getContainerToHostValue(String containerId, String key) {
...
+    for (String containerID : containers.keySet()) {
+      ContainerModel containerModel = containers.get(containerID);
+      for (TaskName taskName : containerModel.getTasks().keySet()) {
+        TaskModel taskModel = containerModel.getTasks().get(taskName);
+        if (taskModel.getTaskMode() != TaskMode.Active) {
+          // Avoid duplicate tasks
+          continue;
+        }
+        ImmutableSet.Builder<SystemStreamPartition> setOfSSPs = ImmutableSet.builder();
+        setOfSSPs.addAll(taskToSSPs.getOrDefault(taskName, ImmutableSet.of()));
```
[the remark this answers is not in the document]
dnishimura: Typically it is, but not always. If you look at the `SingleContainerGrouperFactory`, it maps all tasks to all containers. Even though that grouper is supposed to only be used by the `PassthroughCoordinator` which can't support startpoints anyway, there is no strict enforcement on container to task mappings.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -304,6 +306,17 @@ private void loadMetadataResources(JobModel jobModel) {
...
         configStore.put(entry.getKey(), serializedValue);
       }
+
+      // fan out the startpoints
+      StartpointManager startpointManager = createStartpointManager(coordinatorStreamStore);
+      startpointManager.start();
+      try {
+        startpointManager.fanOut(jobModel.getTaskToSystemStreamPartitions());
+      } catch (IOException ex) {
+        throw new SamzaException(String.format("Error calling fan out from StartpointManager."), ex);
```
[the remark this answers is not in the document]
dnishimura: Sure, good catch (pun intended)

samza-core/src/test/java/org/apache/samza/clustermanager/TestClusterBasedJobCoordinator.java
```diff
@@ -119,4 +131,31 @@ public void testPartitionCountMonitorWithoutDurableStates() {
...
+    configMap.put("job.jmx.enabled", "false");
+    when(CoordinatorStreamUtil.readConfigFromCoordinatorStream(anyObject())).thenReturn(new MapConfig(configMap));
+    Config config = new MapConfig(configMap);
+    MockitoException stopException = new MockitoException("Stop");
+
+    ClusterBasedJobCoordinator clusterCoordinator = Mockito.spy(new ClusterBasedJobCoordinator(config));
+    ContainerProcessManager mockContainerProcessManager = mock(ContainerProcessManager.class);
+    doReturn(true).when(mockContainerProcessManager).shouldShutdown();
+    StartpointManager mockStartpointManager = mock(StartpointManager.class);
+    doThrow(stopException).when(mockStartpointManager).stop(); // Stop ClusterBasedJobCoordinator#run after stop() method.
```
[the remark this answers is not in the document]
dnishimura: It's a workaround to get the `ClusterBasedJobCoordinator#run` to not enter the event loop at the end of run. It'll take more refactoring on the `ClusterBasedJobCoordinator` for better mock support and I didn't want to do it as part of this PR. Here I just want to verify that `start`, `fanOut` and `stop` are called.

samza-core/src/test/java/org/apache/samza/zk/TestZkJobCoordinator.java
dnishimura: I'll look into it. Thanks.

samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
```diff
@@ -119,6 +125,30 @@ public String getContainerToHostValue(String containerId, String key) {
...
   }
 
+  /**
+   * Extracts the map of {@link SystemStreamPartition}s to {@link TaskName} from the underlying {@link TaskModel}s
+   *
+   * @return the extracted map
+   */
+  public Map<TaskName, Set<SystemStreamPartition>> getTaskToSystemStreamPartitions() {
+    HashMap<TaskName, Set<SystemStreamPartition>> taskToSSPs = new HashMap<>();
+    for (String containerID : containers.keySet()) {
```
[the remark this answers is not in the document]
dnishimura: It would work. Will change.

samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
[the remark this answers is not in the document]
cameronlee314: Thanks for clarifying.
The pattern you are using for building up the SSP sets (i.e. only using immutable sets and then creating new ones when they need to be updated) seems a little inconsistent with what I have seen more regularly in Java. Usually, the value is a mutable set, and then you just add to that set. The pattern you are using is more of a Scala pattern, since Scala better supports using immutable sets. If you stick with this pattern, could you please put a short note about why this works for what you need?

samza-core/src/test/java/org/apache/samza/clustermanager/TestClusterBasedJobCoordinator.java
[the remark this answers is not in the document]
cameronlee314: Could you please add this explanation as a doc to the test?

samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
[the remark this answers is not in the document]
dnishimura: Cool if it fits the Java pattern better, I'll switch to the mutable type. No strong requirement other than following the functional programming (Scala) pattern.

dnishimura:
@cameronlee314 - I addressed your comments. Please take a look when you get a chance.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -230,7 +231,7 @@ public void run() {
       StartpointManager startpointManager = createStartpointManager();
       startpointManager.start();
       try {
-        startpointManager.fanOut(jobModel.getTaskToSystemStreamPartitions());
+        startpointManager.fanOut(new JobModelUtil(jobModel).getTaskToSystemStreamPartitions());
```
cameronlee314: Consider making this a static helper. If there are any future util methods, then you won't have to build a new object to just use one of them.

samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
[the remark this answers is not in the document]
cameronlee314: If you wanted to, you could still wrap the end result in immutable structures. But not necessary.

samza-core/src/test/java/org/apache/samza/job/model/TestJobModelUtil.java
```diff
@@ -0,0 +1,70 @@
...
+        new SystemStreamPartition("system1", "stream1_1", new Partition(0)),
+        new SystemStreamPartition("system1", "stream1_2", new Partition(0)));
+    Set<SystemStreamPartition> ssps2 = ImmutableSet.of(
+        new SystemStreamPartition("system1", "stream2_1", new Partition(1)),
+        new SystemStreamPartition("system1", "stream2_2", new Partition(1)));
+    Set<SystemStreamPartition> ssps3 = ImmutableSet.of(
+        new SystemStreamPartition("system1", "stream3_1", new Partition(2)),
+        new SystemStreamPartition("system1", "stream3_2", new Partition(2)));
+    Map<TaskName, TaskModel> tasksForContainer1 = ImmutableMap.of(
+        new TaskName("t1"), new TaskModel(new TaskName("t1"), ssps1, new Partition(0)),
```
cameronlee314: Maybe map the same task to two different containers to test out that part?

samza-core/src/test/java/org/apache/samza/zk/TestZkJobCoordinator.java
```diff
@@ -214,47 +243,53 @@ public void testShouldStopPartitionCountMonitorWhenStoppingTheJobCoordinator() {
...
-    when(zkJobCoordinator.getPartitionCountMonitor()).thenReturn(monitor);
-
-    Set<SystemStreamPartition> ssps = ImmutableSet.of(
-        new SystemStreamPartition("system1", "stream1_1", new Partition(0)),
-        new SystemStreamPartition("system1", "stream1_2", new Partition(0)));
-    Map<TaskName, TaskModel> tasksForContainer = ImmutableMap.of(
-        new TaskName("t1"), new TaskModel(new TaskName("t1"), ssps, new Partition(0)));
-    ContainerModel containerModel = new ContainerModel("0", tasksForContainer);
-    doReturn(new JobModel(config, ImmutableMap.of("0", containerModel))).when(zkJobCoordinator).generateNewJobModel(any());
+    doReturn(mock(MetadataResourceUtil.class)).when(zkJobCoordinator).createMetadataResourceUtil(any(JobModel.class));
```
cameronlee314: Maybe also verify `createResources` gets called on the `MetadataResourceUtil`?

samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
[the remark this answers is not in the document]
dnishimura: I'll leave it mutable for local consistency.

samza-core/src/test/java/org/apache/samza/job/model/TestJobModelUtil.java
dnishimura: Good suggestion, will do.

samza-core/src/test/java/org/apache/samza/zk/TestZkJobCoordinator.java
dnishimura: Will do.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -304,6 +307,17 @@ private void loadMetadataResources(JobModel jobModel) {
...
+
+      // fan out the startpoints
+      StartpointManager startpointManager = createStartpointManager(coordinatorStreamStore);
+      startpointManager.start();
+      try {
+        startpointManager.fanOut(JobModelUtil.getTaskToSystemStreamPartitions(jobModel));
+      } finally {
+        startpointManager.stop();
+      }
+    } catch (IOException ex) {
```
shanthoosh: Minor: Why do we need to wrap the exception and throw-back. Wouldn't bluntly throwing back here would suffice?

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -216,15 +218,24 @@ public void run() {
...
       LOG.info("Starting cluster based job coordinator");
 
-      //create necessary checkpoint and changelog streams, if not created
+      // create necessary checkpoint and changelog streams, if not created
       JobModel jobModel = jobModelManager.jobModel();
       MetadataResourceUtil metadataResourceUtil =
           new MetadataResourceUtil(jobModel, metrics);
       metadataResourceUtil.createResources();
 
+      // fan out the startpoints
```
shanthoosh: For making this feature fully functional in open-source, there has to be startpoint publisher(tool/API) that writes the startpoints in fanout namespace of meta-store. Are we planning on extending CheckpointTool to support it or writing a new tool for it?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
dnishimura: Just to wrap it as a `RuntimeException` which `SamzaException` extends so the exception will propagate up as an unchecked exception and fail the job.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
dnishimura: Fan out is the responsibility of the job coordinator, which is the purpose of this PR. If there is a special use case where a tool/API needs to do fan out directly, the `fanOut` method will write to the namespace directly and the caller does not need to be aware of that namespace.

shanthoosh: Discussed offline with @dnishimura . 

Actual question was will we develop a tool similar to `CheckpointTool` to enable open source users to publish startpoints of their samza job. Daniel confirmed that the development of sucj a tool is in progress and will be made available soon.

PR #1043, opened 2019-05-21
SAMZA-2202: Create log compact topic with larger message size

lhaiesp:


shanthoosh:
Just curious.

Maximum message size accepted by a kafka broker is defined by the configuration: message.max.bytes. There're new asynchronous API's added in kafka-AdminClient which allows the users to query the broker configuration. Broker configurations can be accessed using `AdminClient.describeConfigs();` API

Rather than exposing a configuration to set maximum msg-size with defaults from samza's-end, would it be beneficial to query the configured value directly once from the broker during container-startup. By that, for different kind of brokers(with different supported max-message-size), users might not have to tune this configuration correctly at samza-end.

What do you think?

lhaiesp:
@shanthoosh  my understanding is that this is only for samza internal topics, not affecting user topics. Although we expose a config we don't really expect any users to set it. What happened is just that it seems like kafka's default max message size (1MB) doesn't fit our need, so we impose a larger size (again, on our own internal topics like changelog, checkpoint, coordinate stream). You can find more details here (note this is a internal link): https://jira01.corp.linkedin.com:8443/browse/LISAMZA-11704

lhaiesp:
Need to double check on the producer config "max.request.size" as well

samza-kafka/src/test/scala/org/apache/samza/config/TestKafkaConfig.scala
```diff
@@ -247,97 +252,115 @@ class TestKafkaConfig {
...
 
     val mapConfig = new MapConfig(props.asScala.asJava)
     val kafkaConfig = new KafkaConfig(mapConfig)
     val jobConfig = new JobConfig(mapConfig)
     assertEquals("kafka-system", jobConfig.getCoordinatorSystemNameOrNull)
     assertEquals("4", kafkaConfig.getCoordinatorReplicationFactor)
+    assertEquals("4096000", kafkaConfig.getCoordinatorMaxMessageByte)
     assertEquals("8675309", kafkaConfig.getCoordinatorSegmentBytes)
   }
+
```
shanthoosh: Minor: Please remove the empty lines.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -281,19 +336,22 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
...
-        "segment.bytes", String.valueOf(segmentBytes)))
+        "segment.bytes", String.valueOf(segmentBytes),
+        "max.message.bytes", String.valueOf(maxMessageBytes)))
     } else {
       properties.putAll(ImmutableMap.of(
         "cleanup.policy", "compact,delete",
         "retention.ms", String.valueOf(KafkaConfig.DEFAULT_RETENTION_MS_FOR_BATCH),
-        "segment.bytes", String.valueOf(segmentBytes)))
+        "segment.bytes", String.valueOf(segmentBytes),
+        "max.message.bytes", String.valueOf(maxMessageBytes)))
```
shanthoosh: Currently the checkpoint topic properties like `task.checkpoint.replication.factor`, `task.checkpoint.segment.bytes` are documented in samza [configuration-table](https://samza.apache.org/learn/documentation/latest/jobs/configuration-table.html). Also, if you agree it would be useful, then we can max.msg.bytes config with default to public configuration table.

lhaiesp:
Thank you. Will address the comments.

PR #1044, opened 2019-05-22
SAMZA-2204: Consolidate some reflection-related utility methods into a common class

cameronlee314:


build.gradle
```diff
@@ -326,6 +326,7 @@ project(":samza-sql_$scalaSuffix") {
 
   dependencies {
     compile project(':samza-api')
+    compile project(":samza-core_$scalaSuffix")
```
prateekm: Not sure if it's OK to introduce samza-core dependency on sql (for split deploys). Even if it is, I'd prefer to not do that for modularity. Can we move the util to API instead?

samza-core/src/main/java/org/apache/samza/util/ReflectionUtil.java
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * Create an instance of the specified class with constructor matching the argument array.
+   * @param className name of the class
+   * @param clazz type of object to return
+   * @param classLoader {@link ClassLoader} to use for loading the class
+   * @param args argument array
+   * @param <T> type fo the class
+   * @return instance of the class
+   */
+  public static <T> T getObjWithArgs(String className, Class<T> clazz, ClassLoader classLoader, Object... args) {
```
prateekm: Nitpick: move classloader param after args (makes more sense to keep class name and args together).

samza-core/src/main/java/org/apache/samza/util/ReflectionUtil.java
```diff
@@ -0,0 +1,96 @@
...
+    try {
+      //noinspection unchecked
+      Class<T> classObj = (Class<T>) Class.forName(className, true, classLoader);
+      Class<?>[] argTypes = new Class<?>[args.length];
+      IntStream.range(0, args.length).forEach(i -> argTypes[i] = args[i].getClass());
+      Constructor<T> ctor = classObj.getDeclaredConstructor(argTypes);
+      return ctor.newInstance(args);
+    } catch (Exception e) {
+      LOG.warn("Failed to create instance for: " + className, e);
+      throw new SamzaException(e);
```
prateekm: Nitpick: Include same message as the log in the new exception?

build.gradle
cameronlee314: samza-kafka depends on samza-core, and samza-sql depends on samza-kafka (although I'm not sure why), so samza-sql already depended on samza-core transitively. Just making it explicit.
Also, samza-sql was already using `DelegatingSystemDescriptor`, which is from samza-core.
Agreed that it would be nice for samza-sql to not depend on samza-core, but it also seems like this class doesn't belong in samza-api directly. Maybe our future plans to reorg the modules (e.g. introduce a samza-common) can move this over?

prateekm: Makes sense to make dependency explicit, we can clean up later.

samza-core/src/main/java/org/apache/samza/util/ReflectionUtil.java
cameronlee314: Java requires a varargs parameter to be the last parameter. I could move the `classLoader` argument earlier in the params list (even first?). Would that be worth it? If I do that, would it be worth it to make `getObj` consistent with the new parameter ordering too (hopefully the IDE can help me out with the re-ordering)?

samza-core/src/main/java/org/apache/samza/util/ReflectionUtil.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  /**
+   * Create an instance of the specified class with constructor matching the argument array.
+   * @param className name of the class
+   * @param clazz type of object to return
+   * @param classLoader {@link ClassLoader} to use for loading the class
+   * @param args argument array
+   * @param <T> type fo the class
```
dnishimura: typo: fo

samza-core/src/main/java/org/apache/samza/util/ReflectionUtil.java
```diff
@@ -0,0 +1,96 @@
...
+      throw new SamzaException(errorMessage, e);
+    }
+  }
+
+  /**
+   * Create an instance of the specified class with constructor matching the argument array.
+   * @param className name of the class
+   * @param clazz type of object to return
+   * @param classLoader {@link ClassLoader} to use for loading the class
+   * @param args argument array
```
dnishimura: Maybe state the args are for the ctor?

samza-core/src/main/java/org/apache/samza/util/ReflectionUtil.java
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * Instantiate an object of type T from a given className using the given classLoader.
+   *
+   * @param className full class name (including namespace)
+   * @param clazz type of object to return
+   * @param classLoader classloader to use to load className
+   * @return object of type T created from the empty constructor of the class corresponding to className
+   * @throws SamzaException if class instance cannot be created
+   */
+  public static <T> T getObj(String className, Class<T> clazz, ClassLoader classLoader) {
```
dnishimura: Maybe add some `Precondition` checks in each of the methods?

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -49,7 +49,10 @@ object Util extends Logging {
 
   /**
    * Instantiate an object of type T from a given className.
+   *
+   * Deprecated: Use [[ReflectionUtil.getObj(String, Class, ClassLoader)]] instead.
```
mynameborat: Consumers of this method are not aware of which class loader to use. Can you add a recommendation here as part of deprecated notes?

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlExecutionContext.java
```diff
@@ -72,7 +70,7 @@ public ScalarUdf createInstance(String clazz, String udfName) {
 
     // Configs should be same for all the UDF methods within a UDF. Hence taking the first one.
     Config udfConfig = udfMetadata.get(udfName).get(0).getUdfConfig();
-    ScalarUdf scalarUdf = ReflectionUtils.createInstance(clazz);
+    ScalarUdf scalarUdf = ReflectionUtil.createInstanceOrNull(clazz, ScalarUdf.class, getClass().getClassLoader());
```
dnishimura: How do we know which `ClassLoader` to use? I noticed sometimes the `ClassLoader` of the `this` object is used, sometimes the `ClassLoader` of another class is used (like in `SamzaSqlApplicationConfig`) and sometimes a custom `ClassLoader` is used (like `ArrayListOnlyClassLoader`).

samza-core/src/main/java/org/apache/samza/util/ReflectionUtil.java
```diff
@@ -0,0 +1,96 @@
...
+  private ReflectionUtil() {
+  }
+
+  /**
+   * Instantiate an object of type T from a given className using the given classLoader.
+   *
+   * @param className full class name (including namespace)
+   * @param clazz type of object to return
+   * @param classLoader classloader to use to load className
+   * @return object of type T created from the empty constructor of the class corresponding to className
```
mynameborat: nit: can we add the missing param documentation for T?

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlExecutionContext.java
cameronlee314: If the caller wants to use a custom classloader, then the caller will use that.
Otherwise, in general, I think `getClass().getClassLoader()` is preferable, but that is only accessible from outside of static contexts and constructor contexts.
The usage of the `.class` needs to be used when in a static method or a constructor, since `getClass()` is only an instance method.
Do you think it would be useful to doc this somewhere? (e.g. in `ReflectionUtil`)

samza-core/src/main/java/org/apache/samza/util/ReflectionUtil.java
prateekm: Ah makes sense. Yeah if it's a trivial refactor can move it to the first param, otherwise no big deal.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlExecutionContext.java
dnishimura: I think it'll be useful, even if it's not part of the official javadoc comments. Another developer searching for example usage might be confused with the various ways the `ClassLoader` is passed in. Thanks!

cameronlee314: Yep, it's good that you brought the question up. It'll be good to have some more context about this. I added some javadoc in `ReflectionUtil`.
Also, I wanted to correct something I said earlier: in a constructor scope, it is possible to use `getClass().getClassLoader`. It just can't be used in a static scope.

cameronlee314:
@prateekm @dnishimura @mynameborat thanks for the reviews. I have made some updates. Could you please take another look?

PR #1046, opened 2019-05-22
SAMZA-2211: [Cleanup] Move RunLoop to container package

mynameborat:
@prateekm @xinyuiscool please take a look


samza-core/src/main/java/org/apache/samza/container/RunLoop.java
```diff
@@ -50,7 +55,13 @@
...
 
 /**
- * The RunLoop supports multithreading execution of Samza {@link AsyncStreamTask}s.
+ * The run loop supports both single-threaded and multi-threaded execution models.
+ *    <p>
+ *      If job.container.thread.pool.size &gt; 1 (multi-threaded), operations like commit, window and timer for all tasks within a container
+ *      happens on a thread pool.
+ *      If job.container.thread.pool.size &lt; 1 (single-threaded), operations for all tasks are multiplexed onto one execution thread.
+ *    </p>.
+ *    Note: In both models, process for all tasks happen sequentially on the main execution thread.
```
prateekm: "process/processAsync for all tasks is called on the run loop thread"

samza-core/src/main/java/org/apache/samza/task/EpochTimeScheduler.java
```diff
@@ -81,7 +81,7 @@ private EpochTimeScheduler(ScheduledExecutorService executor) {
     }
   }
 
-  void registerListener(TimerListener listener) {
+  public void registerListener(TimerListener listener) {
```
prateekm: Should this class be in container instead of task as well?

mynameborat: I noticed we have a scheduler package and CallbackSchedulerImpl is the main consumer of this class. 
Ended up moving it there. Let me know if you think we consolidate scheduler and container package.

PR #987, opened 2019-04-05
SAMZA-2158: Remove the redunant coordinator stream reads in the ApplicationMaster startup sequence.

shanthoosh:
**Changes:**
Currently the input topic partitions assigned to a container of a samza job is stored in the coordinator stream(aka kafka topic).

In samza-yarn ApplicationMaster startup sequence, the JobModel from the previous run of the samza job is read from the coordinator stream.

JobModel is read multiple times from the same kafka topic with different connections. These redundant reads prolongs the launch of containers by the samza-yarn ApplicationMaster. This fix is to remove the inefficieny by reading the coordinator stream only once with one connection.

Please do note that the above two problems had slowed down ApplicationMaster startup and did not break functional correctness.

**Note:**
* In addition to the above problem,  KafkaSystemAdmin is created multiple times for same topic/system multiple times in ApplicationMaster/Container and it will be fixed in SAMZA-2157.
* Tested this patch with a test samza-yarn job inside linkedin and with a beam job.

PR #1048, opened 2019-05-23
SAMZA-2215 : StartpointManager fix for previous CoordinatorStreamStore refactor

dnishimura:
`StartpointManager` instantiation in the `ContainerLaunchUtil` depends on the `MetadataStoreFactory` config which is not currently be used now due to PR #987. This PR switches to using the directly created metadata store to construct the `StartpointManager` that puts it in parity with other managers that use the metadata store.
[SAMZA-2216](https://issues.apache.org/jira/browse/SAMZA-2216) tracks the follow on work to reintegrate the factory.


dnishimura:
@shanthoosh please take a look

PR #1049, opened 2019-05-24
SAMZA-2217: Refactoring of Couchbase table functions

weisong44:
Minor refactoring
1. Made java.util.concurrent.Timeout to be retriable
2. Allow CouchbaseTableWriteFunction.asyncWriteHelper() to return a CompletableFuture with value


samza-kv-couchbase/src/main/java/org/apache/samza/table/remote/couchbase/CouchbaseTableWriteFunction.java
```diff
@@ -108,4 +119,5 @@ public void onError(Throwable error) {
     });
     return future;
   }
+
```
lhaiesp: unnecessary empty line in the end?

PR #1050, opened 2019-05-24
SAMZA-2221: Use the KafkaStreamSpec instead of the StreamSpec 

dengpanyin:
When creating a new Kafka topic, currenlty StreamSpec is used other than the
KafkaStreamSpec. KafkaStreamSpec contains more properties, such as "retention.ms",
that StreamSpec is not aware of.

@sborya  @xinyuiscool  @weisong44

### DIFF
--- a/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
+++ b/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
@@ -461,24 +461,13 @@ public class KafkaSystemAdmin implements SystemAdmin {
   @Override
   public boolean createStream(StreamSpec streamSpec) {
     LOG.info("Creating Kafka topic: {} on system: {}", streamSpec.getPhysicalName(), streamSpec.getSystemName());
-    final String REPL_FACTOR = "replication.factor";
 
     KafkaStreamSpec kSpec = toKafkaSpec(streamSpec);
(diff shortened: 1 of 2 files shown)


samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -461,24 +461,13 @@ public Integer offsetComparator(String offset1, String offset2) {
...
-    Map<String, String> streamConfig = new HashMap<>(streamSpec.getConfig());
-    // HACK - replication.factor is invalid config for AdminClient.createTopics
-    if (streamConfig.containsKey(REPL_FACTOR)) {
-      String repl = streamConfig.get(REPL_FACTOR);
-      LOG.warn("Configuration {}={} for topic={} is invalid. Using kSpec repl factor {}",
-          REPL_FACTOR, repl, kSpec.getPhysicalName(), kSpec.getReplicationFactor());
-      streamConfig.remove(REPL_FACTOR);
-    }
-    newTopic.configs(new MapConfig(streamConfig));
+    newTopic.configs(KafkaStreamSpec.filterUnsupportedProperties(kSpec.getConfig()));
```
mynameborat: It should be sufficient to pass `kSpec.getConfig()` and you don't need apply filter again.
`toKafkaSpec(...)` takes care of filtering it unsupported properties.

PR #1051, opened 2019-05-24
SAMZA-2219: Add a dummy table read function

weisong44:
This is a convenience class for cases only writes are required.
Removed compilation warnings from various files.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/table/remote/DummyTableReadFunction.java
+++ b/samza-core/src/main/java/org/apache/samza/table/remote/DummyTableReadFunction.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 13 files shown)


samza-core/src/main/java/org/apache/samza/table/remote/DummyTableReadFunction.java
```diff
@@ -0,0 +1,40 @@
...
+package org.apache.samza.table.remote;
+
+import java.util.concurrent.CompletableFuture;
+import org.apache.samza.SamzaException;
+
+
+/**
+ * A dummy table read function
+ */
+public class DummyTableReadFunction<K, V> extends BaseTableFunction implements TableReadFunction<K, V> {
```
atoomula: How about renaming to NoOpTableReadFunction ?

weisong44: Yeah, this sounds better.

PR #1052, opened 2019-05-24
SAMZA-2222: Documentation for async API

mynameborat:
image


docs/learn/tutorials/versioned/samza-async-user-guide.md
```diff
@@ -107,16 +107,96 @@ task.callback.timeout.ms=5000
 
 **NOTE:** Samza also guarantees the in-order process of the messages within an AsyncStreamTask by default, meaning the next processAsync() of a task won't be called until the previous processAsync() callback has been triggered.
 
+### Asynchronous Process in High Level API
+
+If your application is asynchronous, e.g. making non-blocking remote IO calls, [AsyncFlatMapFunction](/learn/documentation/{{site.version}}/api/javadocs/org/apache/samza/operators/functions/AsyncFlatMapFunction.html) provides support for it. The following example illustrates an application that processes Wikipedia feed updates and invokes a remote service to standardize the updates and sends the standardized events to Wikipedia.
```
prateekm: "If your processing logic is asynchronous, e.g, it makes non-blocking remote calls, you can implement it using the AsyncFlatMapFunction".

docs/learn/tutorials/versioned/samza-async-user-guide.md
```diff
@@ -107,16 +107,96 @@ task.callback.timeout.ms=5000
...
+
+    @Override
+    public void close() {
+      client.close();
+    }
+  }
+}
+{% endhighlight %}
+
+In the above example, the messages are not sent to Wikipedia after `AsyncStandardizerFunction` returns. The framework keeps track of the future returned by the function and propagates the standardized result to downstream operator (in this case sendTo) only when the future completes. Samza has a timeout for the future to complete and can be configured through the following property:
```
prateekm: "The framework keeps track ..." etc is implementation detail. User guide should only talk about what the semantics are. E.g. , "In the above example, the results from the asynchronous ... function are propagated to downstream operators once the future is complete."

"Samza has a timeout for the future ..." makes it sound like we control when the future is completed or have an await. It'll be better to just say that "there is an overall timeout for each message to be processed and the processAsync callback to be invoked, and you can tune that using ..."

docs/learn/tutorials/versioned/samza-async-user-guide.md
```diff
@@ -107,16 +107,96 @@ task.callback.timeout.ms=5000
...
+
+{% highlight jproperties %}
+# Timeout for the future to complete. When the timeout elapses, the framework will throw a TaskCallbackTimeoutException and shut down the container.
```
prateekm: "will throw a ..." is implementation detail. Also, can just say "the container will be restarted".

docs/learn/tutorials/versioned/samza-async-user-guide.md
```diff
@@ -107,16 +107,96 @@ task.callback.timeout.ms=5000
...
+task.callback.timeout.ms
+{% endhighlight %}
+
+For remote IO libraries that don't support Futures/Promises and use callbacks, it should be straightforward to adapt the callbacks to Future pattern. The following code snippet illustrates it
```
prateekm: "If the I/O library accepts callbacks instead of returning a Future, the callback can be adapted to a Future in the following way: "

docs/learn/tutorials/versioned/samza-async-user-guide.md
```diff
@@ -107,16 +107,96 @@ task.callback.timeout.ms=5000
...
+
+{% highlight java %}
+
+  public CompletionStage<Collection<StandardizedWikipediaFeedEvent>> apply(WikipediaFeedEvent wikipediaFeedEvent) {
+    Request<StandardizerRequest> standardizedRequest = buildStandardizedRequest(wikipediaFeedEvent);
```
prateekm: s/standardizedRequest/standardizationRequest

docs/learn/tutorials/versioned/samza-async-user-guide.md
```diff
@@ -107,16 +107,96 @@ task.callback.timeout.ms=5000
...
+In all cases above, Samza supports in-order process by default. Further parallelism is also supported by allowing a task to process multiple outstanding messages in parallel. The following config allows one task to process at most 4 outstanding messages in parallel at a time:
 
 {% highlight jproperties %}
 # Max number of outstanding messages being processed per task at a time, applicable to both StreamTask and AsyncStreamTask.
 task.max.concurrency=4
 {% endhighlight %}
 
-**NOTE:** In case of AsyncStreamTask, processAsync() is still invoked in the order of the message arrivals, but the completion can go out of order. In case of StreamTask with multithreading, process() can run out-of-order since they are dispatched to a thread pool. This option should **NOT** be used when strict ordering of the output is required.
+**NOTE:** In case of AsyncStreamTask, processAsync() is still invoked in the order of the message arrivals, but the completion can go out of order. In case of StreamTask and StreamApplication with concurrency, process() can run out-of-order since they are dispatched to a thread pool. This option should **NOT** be used when strict ordering of the output is required.
```
prateekm: s/can go/can happen.

s/StreamApplication with concurrency/High Level API applications with task.max.concurrency > 1,

s/process() can run/delivery can be.

"dispatched to a thread pool" isn't relevant here.

docs/learn/tutorials/versioned/samza-async-user-guide.md
mynameborat: leaving out "the processAsync callback to be invoked" since it feels implementation detail too.

mynameborat:
@prateekm Addressed your comments. Some of the language used initially was done with the notion of keeping up consistency within the document. 
The page still has some implementation detail and we may want to make it consistent. Will file a follow up JIRA.

PR #1053, opened 2019-05-28
SAMZA-2220: Startpoints - Fully encapsulate resolution of starting offsets in OffsetManager

dnishimura:
In PR #1026, the startpoint resolution was integrated with only the system consumers, but not the task instances "starting offset caught up logic". Without the correct integration with the task instance caught up logic, the startpoint resolution from PR #1026 is ignored.

This PR encapsulates the startpoint resolution fully in the offset manager so that all components dependent on the starting offsets will get a consistent resolution of the offsets.


samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -193,6 +194,7 @@ class OffsetManager(
     stripResetStreams
     loadStartingOffsets
     loadStartpoints
+    resolveStartpointsToStartingOffsets
```
rmatharu: Would make sense to call resolveStartpointsToStartingOffsets  from loadStartpoints?

dnishimura: Initially had it after loading defaults. Makes better sense to put it in loadStartpoints. Good catch, thanks!

dnishimura:
@shanthoosh - please take a look when you get a chance.

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -535,9 +536,47 @@ class OffsetManager(
         startpoints
           .foreach(taskMap => taskMap._2
             .foreach(sspMap => info("Loaded startpoint: %s for SSP: %s and task: %s" format (sspMap._2, sspMap._1, taskMap._1))))
+        resolveStartpointsToStartingOffsets
```
shanthoosh: Minor: Would be better to update the java doc of the `OffsetManager` class and include that it accounts for startpoints into calculation of the starting offset.

dnishimura: Done. Thanks for the feedback.

shanthoosh:
@dnishimura 
Thanks for the patch. Merged to trunk.

PR #1055, opened 2019-05-29
SAMZA-2227 : Fix to ensure that new topics detected by regex-monitor are picked up during next deployment of app

rmatharu:


samza-api/src/main/java/org/apache/samza/job/model/ContainerModel.java
```diff
@@ -54,6 +58,15 @@ public String getId() {
     return tasks;
   }
 
+  /**
+   * Returns the {@link SystemStreamPartition}s that this container is responsible for consuming.
+   * @return {@link SystemStreamPartition}s for this container
+   */
+  public Set<SystemStreamPartition> getSystemStreamPartitions() {
```
dnishimura: For consistency with the `JobModel`, getters ignored in json serde should probably be moved to a util class similar to `JobModelUtil`

samza-core/src/main/java/org/apache/samza/job/model/JobModel.java
```diff
@@ -141,6 +145,15 @@ private void populateContainerLocalityMappings() {
     return containers;
   }
 
+  /**
+   * Returns the {@link SystemStreamPartition}s that this job is responsible for consuming.
+   * @return {@link SystemStreamPartition}s for this job
+   */
+  public Set<SystemStreamPartition> getSystemStreamPartitions() {
```
dnishimura: This can probably be moved to `JobModelUtil`.

samza-core/src/main/java/org/apache/samza/serializers/model/JsonContainerModelMixIn.java
```diff
@@ -49,5 +52,8 @@
 
   @JsonProperty(TASKS_KEY)
   abstract Map<TaskName, TaskModel> getTasks();
+
+  @JsonIgnore
+  abstract Set<SystemStreamPartition> getSystemStreamPartitions();
```
dnishimura: For consistency, please move JsonIgnore getters to a util class.

samza-core/src/main/java/org/apache/samza/serializers/model/JsonJobModelMixIn.java
```diff
@@ -40,4 +43,7 @@ public JsonJobModelMixIn(@JsonProperty("config") Config config, @JsonProperty("c
 
   @JsonProperty("containers")
   abstract Map<String, ContainerModel> getContainers();
+
+  @JsonIgnore
+  abstract Set<SystemStreamPartition> getSystemStreamPartitions();
```
dnishimura: Probably best to move this to `JobModelUtil`

rmatharu:
@jagadish-v0 Please take a look.

samza-core/src/main/scala/org/apache/samza/config/RegExTopicGenerator.scala
```diff
@@ -97,19 +95,13 @@ class RegExTopicGenerator extends ConfigRewriter with Logging {
     new MapConfig(((keysAndValsToAdd ++ config.asScala) += inputStreams).asJava)
   }
 
-  def getTopicsFromZK(rewriterName: String, config: Config): Seq[String] = {
-    val systemName = config
+  def getTopicsFromStreamMetadataCache(rewriterName: String, config: Config): Seq[String] = {
```
mynameborat: why are we changing this to use StreamMetadatatCache? 
Is this necessary for the fix?

rmatharu: Its neccessary because it allows the regex-expansion to be Kafka-agnostic, which allows the regexGenerator to be moved to samza-core (so that the job-model-manager can invoke it). 

Currently, the regexGenerator is in samza-kafka, so the job-model-manager cannot invoke the regexGenerator, without adding a dependency (circular) from samza-core to samza-kafka. 
Also, reading topic info from Zk without using a documented API, bypassing the broker, seems kinda rakish.

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -140,17 +140,26 @@ object Util extends Logging {
...
+      case Some(rewriters) => rewriters.split(",").foldLeft(config)(rewrite(_, _))
+      case _ => config
     }
+  }
 
+  private def rewrite(c: Config, rewriterName: String): Config = {
+    val rewriterClassName = c
+      .getConfigRewriterClass(rewriterName)
+      .getOrElse(throw new SamzaException("Unable to find class config for config rewriter %s." format rewriterName))
+    val rewriter = Util.getObj(rewriterClassName, classOf[ConfigRewriter])
```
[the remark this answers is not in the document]
rmatharu: done

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -148,17 +148,28 @@ object Util extends Logging {
...
+  private def rewrite(config: Config, rewriterName: String): Config = {
+    val rewriterClassName = config
+      .getConfigRewriterClass(rewriterName)
+      .getOrElse(throw new SamzaException("Unable to find class config for config rewriter %s." format rewriterName))
+    val rewriter = Util.getObj(rewriterClassName, classOf[ConfigRewriter])
+    info("Re-writing config with " + rewriter)
+    rewriter.rewrite(rewriterName, config)
+  }
+
+  def runRegexTopicGenerator(config: Config): Config = {
```
vjagadish1989: call this "invokeRegexTopicRewriter"?

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -148,17 +148,28 @@ object Util extends Logging {
...
-      val rewriter = Util.getObj(rewriterClassName, classOf[ConfigRewriter])
-      info("Re-writing config with " + rewriter)
-      rewriter.rewrite(rewriterName, c)
+    config.getConfigRewriters match {
+      case Some(rewriters) => rewriters.split(",").foldLeft(config)(rewrite(_, _))
+      case _ => config
     }
+  }
 
+  private def rewrite(config: Config, rewriterName: String): Config = {
```
vjagadish1989: "invokeRewriter"

vjagadish1989: order methods by significance - since it's a private method, move it to the bottom

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -148,17 +148,28 @@ object Util extends Logging {
...
+      .getOrElse(throw new SamzaException("Unable to find class config for config rewriter %s." format rewriterName))
+    val rewriter = Util.getObj(rewriterClassName, classOf[ConfigRewriter])
+    info("Re-writing config with " + rewriter)
+    rewriter.rewrite(rewriterName, config)
+  }
+
+  def runRegexTopicGenerator(config: Config): Config = {
     config.getConfigRewriters match {
-      case Some(rewriters) => rewriters.split(",").foldLeft(config)(rewrite(_, _))
+      case Some(rewriters) => rewriters.split(",").
```
vjagadish1989: is invoking rewrite on regex-config-rewriter idempotent?

samza-core/src/main/scala/org/apache/samza/config/RegExTopicGenerator.scala
prateekm: 1. Can we call `getAllSystemStreams` directly on the system admin?
2. SystemAdmin(s) need to started/stopped before/after use.

samza-core/src/main/scala/org/apache/samza/util/Util.scala
prateekm: Does this need to be a public util method? Just inline it in the JobModelManager? Would prefer to not make Util a dumping ground for helper methods again :)

samza-core/src/main/scala/org/apache/samza/util/Util.scala
rmatharu: Yes, it is.

rmatharu:
Addressed all comments.

samza-core/src/main/scala/org/apache/samza/config/RegExTopicGenerator.scala
```diff
@@ -97,19 +95,15 @@ class RegExTopicGenerator extends ConfigRewriter with Logging {
...
```
vjagadish1989: rename the method to reflect its current state

samza-core/src/main/scala/org/apache/samza/config/RegExTopicGenerator.scala
```diff
@@ -97,19 +95,15 @@ class RegExTopicGenerator extends ConfigRewriter with Logging {
...
 
-    try {
-      ZkUtils(zkClient, isZkSecurityEnabled = false).getAllTopics()
-    } finally {
-      zkClient.close()
-    }
+    val systemAdmin = new SystemConfig(config).getSystemAdmin(systemName)
+    systemAdmin.start()
+    val systemStreams = systemAdmin.getAllSystemStreams.asScala.map(systemStream => systemStream.getStream).toSeq
+    systemAdmin.stop()
```
vjagadish1989: wrap the stop() in a `finally` so that it's invoked no matter what

PR #1060, opened 2019-05-31
SAMZA-2230: Support for ANY type arguments in udfs that expect strongly typed arguments

srinipunuru:
Calcite and hence Samza SQL supports untyped fields of type SamzaSqlFieldType.ANY. Corresponding java type for such fields is Object. 

When an UDF expects a strongly typed field and when one of the arguments that is passed is untyped (i.e. ANY ) it fails. And calcite doesn't allow casting the ANY to a well defined type like java does.

To fix this we explicitly cast the untyped arguments to the type that the UDF expects as part of the UDF code generation. Also added a test case that fails without this change.


samza-sql/src/test/java/org/apache/samza/sql/util/MyTestObjUdf.java
```diff
@@ -0,0 +1,29 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Test UDF used by unit and integration tests.
+ */
+@SamzaSqlUdf(name = "MyTestObj", description = "Test UDF.")
+public class MyTestObjUdf implements ScalarUdf {
+
+  private static final Logger LOG = LoggerFactory.getLogger(MyTestUdf.class);
```
weiqingy: Why we use `MyTestUdf.class` here? What is the relationship between "MyTestObjUdf" class and `MyTestUdf` class? I saw the method `execute(Object value)` is deleted from `MyTestObj` class.

srinipunuru: This was a mistake, I will fix this. there is no relationship between MyTestUdf and MyTestObj. the deleted method in MyTestUDF was unused.

PR #1061, opened 2019-06-03
SAMZA-2232: Move StreamOperatorTaskFactory to samza-api

mynameborat:


samza-api/src/main/java/org/apache/samza/task/StreamOperatorTaskFactory.java
```diff
@@ -20,11 +20,11 @@
 
 
 /**
- * Build {@link StreamOperatorTask} instances.
+ * Build {@link AsyncStreamTask} instances.
  * <p>
- * Implementations should return a new instance of {@link StreamOperatorTask} for each {@link #createInstance()} invocation.
+ * Implementations should return a new instance of {@link AsyncStreamTask} for each {@link #createInstance()} invocation.
  * Note: It is not part of samza-api as it is a temporary hack introduced for SAMZA-2172. It will eventually
```
vjagadish1989: update java-docs to reflect current state accurately
```
"Note: It is not part of samza-api as it is a temporary hack introduced for SAMZA-2172"
```
For eg: it is part of samza-api in this PR

mynameborat: I updated the documentation

PR #1063, opened 2019-06-03
SAMZA-2234: Passing context to Samza SQL UDFs

srinipunuru:
This change provides ability for the UDFs to access the Samza context. Context is needed by the UDFs for various reasons
1. To create metrics
2. To access configs
3. Access any external contexts


samza-sql/src/main/java/org/apache/samza/sql/data/Expression.java
```diff
@@ -29,10 +30,10 @@
 public interface Expression {
   /**
    * This method is used to implement the expressions that takes in columns as input and returns multiple values.
-   * @param context the context
+   * @param context SamzaSqlExecution context
```
celiakung: nit: the variable name should be 'sqlContext'
and need to add another param on the next line for regular 'context'

PR #1065, opened 2019-06-04
SAMZA-2233: Allow overriding descriptor generated task.inputs using configs

prateekm:


prateekm:
@Sanil15 Can you please take a look for TestRunner changes.

samza-core/src/main/java/org/apache/samza/execution/JobPlanner.java
```diff
@@ -78,10 +78,8 @@ ExecutionPlan getExecutionPlan(String runId) {
     // TODO: This should all be consolidated with ExecutionPlanner after fixing SAMZA-1811
     // Don't generate any configurations for LegacyTaskApplications
     if (!LegacyTaskApplication.class.isAssignableFrom(appDesc.getAppClass())) {
-      if (userConfig.containsKey(TaskConfig.INPUT_STREAMS())) {
-        LOG.warn("SamzaApplications should not specify task.inputs in configuration. " +
-            "Specify them using InputDescriptors instead. Ignoring configured task.inputs value of " +
-            userConfig.get(TaskConfig.INPUT_STREAMS()));
+      // Don't allow overriding task.inputs to a blank string
+      if (StringUtils.isBlank(userConfig.get(TaskConfig.INPUT_STREAMS()))) {
```
rmatharu: Why is this check required?

prateekm: samza app-def sets task.inputs = "" by default, so we need to ignore that. It's a simple and useful enough check that I kept this in OSS instead of patching in LI.

PR #1069, opened 2019-06-05
SAMZA-2208: Migrate Util.rewriteConfig to inject classloader when doing reflection

cameronlee314:
Intention is for all functionality to stay the same. This PR is just for setting up a future PR to wire in a new classloader in the future.


cameronlee314:
No longer necessary to do this. Classloader propagation will be done in a more "automatic" way through a separate PR.

PR #1070, opened 2019-06-05
SAMZA-2226: Migrate Class.forName usage to inject classloader

cameronlee314:
Intention is for all functionality to stay the same. This PR is just for setting up a future PR to wire in a new classloader in the future.


samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -79,9 +79,20 @@ object Util extends Logging {
 
   def getTaskClassVersion(config: Config, classLoader: ClassLoader): String = {
     try {
-      val taskClass = Option(new ApplicationConfig(config).getAppClass())
-        .orElse(new TaskConfig(config).getTaskClass).get
-      Class.forName(taskClass, true, classLoader).getPackage.getImplementationVersion
+      val appClass = Option(new ApplicationConfig(config).getAppClass)
+      if (appClass.isDefined) {
+        Option.apply(Class.forName(appClass.get, true, classLoader).getPackage.getImplementationVersion)
```
prateekm: Nitpick: don't need to call apply, can just call Option(...)

samza-core/src/main/scala/org/apache/samza/util/Util.scala
```diff
@@ -79,9 +79,20 @@ object Util extends Logging {
...
+          .getOrElse(FALLBACK_VERSION)
+      } else {
+        val taskClass = new TaskConfig(config).getTaskClass
+        if (taskClass.isPresent) {
+          Option.apply(Class.forName(taskClass.get(), true, classLoader).getPackage.getImplementationVersion)
```
prateekm: Don't need option.apply

cameronlee314:
No longer necessary to do this. Classloader propagation will be done in a more "automatic" way through a separate PR.

PR #1078, opened 2019-06-11
SAMZA-2248: Fix AM bookkeeping on receiving dead container notifications

xinyuiscool:
Fixed the following issue:

1. AM gets extra containers from the RM which it saves for later use
2. When the container that we saved in step1 dies, the AM on receiving this callback does nothing about it.
3. Later, when we are looking for a container to use - we pick up the dead container that we saved and did not clean up (step 1&2) and launch a container. 
4. Now, if this launched container ever dies - the RM will never notify the AM about it since it see's it as a duplicate (step 2)
5. Job is left without the container rescheduled and will need to be restarted.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/clustermanager/ResourceRequestState.java
+++ b/samza-core/src/main/java/org/apache/samza/clustermanager/ResourceRequestState.java
@@ -18,6 +18,7 @@
  */
 package org.apache.samza.clustermanager;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


xinyuiscool:
@abhishekshivanna  and @vjagadish1989 : please help take a look at this pr. Thx.

abhishekshivanna:
LGTM. Thanks for the fix @xinyuiscool

vjagadish1989:
> (2) When the container that we saved in step1 dies, the AM on receiving this callback does nothing about it.

@xinyuiscool: 
what's the root-cause for (2)? the expected behavior to request resources for it from the ClusterManager and restart it

samza-core/src/test/java/org/apache/samza/clustermanager/TestContainerRequestState.java
```diff
@@ -193,4 +193,28 @@ public void testContainerAssignment() throws Exception {
 
   }
 
+  @Test
```
rmatharu: Could we also add a test for this in TestContainerAllocator and/or TestHostAwareContainerAllocator?

xinyuiscool: I think the scope of the change is quite limited so this test should cover it. The change does not affect the allocator.

samza-core/src/main/java/org/apache/samza/clustermanager/ResourceRequestState.java
```diff
@@ -229,6 +230,25 @@ public int releaseExtraResources() {
     }
   }
 
+  /**
+   * Releases a single resource based on containerId.
+   * @param containerId Yarn container ID
+   */
+  public void releaseResource(String containerId) {
+    if (StringUtils.isEmpty(containerId)) {
+      log.warn("ContainerId is not specified");
```
sborya: please add "ContainerId is not specified" 'in ClusterManager release resource'.
Without the context the message may be to vague.

PR #1079, opened 2019-06-11
SAMZA-2250: Support large job models in standalone.

shanthoosh:


shanthoosh:
Working on resolving the merge conflicts.

shanthoosh:
@sborya 

Can you please take a look when you get a chance.

prateekm:
@bharathkk to take a look and see if this will be addressed in the metadata store work. Batching large number of writes will be an issue with any implementation, so perhaps this should be addressed at the API layer, either by supporting batch puts, or by accepting the JobModel (and other large messages like configs) as a whole and chunking internally within an implementation. Also, when we have versioning in metadata store, we should also not need to store the job model in ZK.

@shanthoosh I don't like that we'e diverging in both the code path and key name for job models in ZK and YARN. The end state when metadata store is available should be a consistent code path to read and write job models from anywhere, including Samza dashboard. IMHO it'll be better to do this inside the ZKMetadataStore implementation since this is a temporary fix. WDYT?

@shanthoosh Also, do you need to store and read with a checksum for the multi-part read?

shanthoosh:
@prateekm

Thanks for the review. 

1. Segmenting the `JobModel` and storing it as independent byte array chunks will be necessary for any `MetadataStore` that doesn't support unlimited value size. This utility is not specific to `ZkMetadataStore` implementation alone and hence would be better to keep it outside of it.

2. Storing the checksum during writes and verifying it during the reads would prevent us from data corruption. I think it is a good suggestion, if acceptable, will be added in a immediate follow-up.

Adding points I discussed offline with you for using zookeeper for storing `JobModel`.

A. *Performance:* 

   * Number of `MetadataStore` writes: Current job model storage mechanism used for YARN is write intensive. JobModel is composed of different kind of metadata viz `TaskPartitionAssignment`, `TaskChangeLogAssignment`, `TaskContainerAssignment`, `Locality` etc and they are stored at task granularity independently. For a samza job that consumes from N input topic partitions, employing `YARN` metadata-storage-managers would result in 4 * N writes to the `MetadataStore` layer(For both `CoordinatorStreamStore`/`ZkMetadataStore`). Even after tuning kafka batch size, this is expensive for large standalone jobs(Example: Search standalone jobs consumes from 15K topic partitions on average).

   * JobModel write frequency: Currently in YARN, the JobModel is computed and stored in `MetadataStore` only once in `ApplicationMaster` startup. In standalone, JobModel can be computed and published multiple times as a part of a single deployment. 

   * Performance tests: I compared the time taken to publish JobModel with existing YARN approach vs segmenting JobModel by size through a series of perf-tests. From my performance tests, YARN approach to store JobModel is significantly slower(10-20 times slower with coordinator-stream implementation) when compared to segmenting the JobModel by size. Example: For a JobModel of size 5MB containing 10K input topic partitions, size based JobModel splitting approach would just perform 5 metadata-store-writes when compared to 40K metadata-store-writes by current YARN approach.



B. *Versioning:* JobModel versioning is not supported in coordinator stream currently. Let's say leader dies after publishing some portion of JobModel. From the coordinator stream alone, we cannot correctly compute the correct JobModel in standalone currently. This is required for jobmodel-lineage tracking and versioned JobModel read for functional correctness.

C. *Observability:* Currently YARN has dashboard support for viewing the JobModel. If we use coordinator stream to store JobModel in standalone then there's no tooling around JobModel visibility. This will make trouble shooting a standalone job harder.

dnishimura:
Nice summary @shanthoosh. Few questions below
> 
> * Number of `MetadataStore` writes: Current job model storage mechanism used for YARN is write intensive. JobModel is composed of different kind of metadata viz `TaskPartitionAssignment`, `TaskChangeLogAssignment`, `TaskContainerAssignment`, `Locality` etc and they are stored at task granularity independently. For a samza job that consumes from N input topic partitions, employing `YARN` metadata-storage-managers would result in 4 * N writes to the `MetadataStore` layer(For both `CoordinatorStreamStore`/`ZkMetadataStore`). Even after tuning kafka batch size, this is expensive for large standalone jobs(Example: Search standalone jobs consumes from 15K topic partitions on average).

When you mention "tuning kafka batch size" do you mean batch puts? (i.e. flush after M entries). Correct me if I'm wrong, but isn't the performance hit mainly because of the current way of "flush per entry"? Do we know if batch puts would increase performance by an order of magnitude or more or would the increase in performance be much less?
  
> * JobModel write frequency: Currently in YARN, the JobModel is computed and stored in `MetadataStore` only once in `ApplicationMaster` startup. In standalone, JobModel can be computed and published multiple times as a part of a single deployment.

This depends on the number of hosts and the configured debounce time?

> 
> B. _Versioning:_ JobModel versioning is not supported in coordinator stream currently. Let's say leader dies after publishing some portion of JobModel. From the coordinator stream alone, we cannot correctly compute the correct JobModel in standalone currently. This is required for jobmodel-lineage tracking and versioned JobModel read for functional correctness.

Seems like ZK is the only option then.

> 
> C. _Observability:_ Currently YARN has dashboard support for viewing the JobModel. If we use coordinator stream to store JobModel in standalone then there's no tooling around JobModel visibility. This will make trouble shooting a standalone job harder.

If we create a metadata store tool, this problem should be solved.

prateekm:
> This utility is not specific to ZkMetadataStore implementation alone and hence would be better to keep it outside of it.

prateekm:
@shanthoosh 
>  This utility is not specific to ZkMetadataStore implementation alone and hence would be better to keep it outside of it.

shanthoosh:
@prateekm

> After leader election the job model is always the latest version? Are there special requirements around versioning JobModels beyond what YARN already does?

Discussed with you offline. Adding details here for providing context to others.

Versioning solves race conditions in JobModel read/write work-flow. Let's say we update the `JobModel` at single location(and call it latest). Please consider the following execution sequence.

1. A leader(L1) of processor group publishes a new `JobModel` with version 1 and creates a barrier. 
2. All followers join the barrier and leader(L1) marks the barrier as done.
3. Before all the followers reads the `JobModel` with version 1 published by L1, the leader(L1) dies.
4. Leader failover kicks-in, new leader(L2) publishes a new `JobModel` with version 2. Depending upon when the followers read the `JobModel`(before or after L2 publishes `JobModel`), some followers would have read the `JobModel` published by L1 and some followers would have read `JobModel` published by L2.

Versioning solves the above  problem in standalone. 

> wouldn't it be sufficient to view the co-ordinator stream using kafka-tool?

JobModel is comprised of different kinds of metadata(task-partition-assignment, container-task-assignment etc) and each of them are stored as separate messages per task in coordinator stream. kafka-tool would dump the individual messages and user would have to combine the individual data to get whole `JobModel`.  Building a tool would be a good idea.

shanthoosh:
@dnishimura 

> Do we know if batch puts would increase performance by an order of magnitude

It was not a significant increase in performance. 

> This depends on the number of hosts and the configured debounce time?
Yes, you're right.

samza-core/src/main/java/org/apache/samza/job/model/JobModelUtil.java
```diff
@@ -57,6 +70,70 @@
...
+  /**
+   * Splits the larger {@link JobModel} into independent segments of 1 MB size and stores them in {@link MetadataStore}.
+   * @param jobModel the job model to store into {@link MetadataStore}.
+   * @param jobModelVersion the job model version.
+   * @param metadataStore the metadata store.
+   */
+  public static void writeJobModel(JobModel jobModel, String jobModelVersion, MetadataStore metadataStore) {
+    try {
+      List<byte[]> jobModelSegments = chunkJobModel(jobModel);
+      for (int jobModelSegmentIndex = 0; jobModelSegmentIndex < jobModelSegments.size(); jobModelSegmentIndex += 1) {
```
sborya: ++ ?

samza-core/src/main/java/org/apache/samza/job/model/JobModelUtil.java
```diff
@@ -57,6 +70,70 @@
...
+   * @return the job model read from the metadata store.
+   */
+  public static JobModel readJobModel(String jobModelVersion, MetadataStore metadataStore) {
+    try {
+      StringBuilder jobModelUnificationBuffer = new StringBuilder();
+      for (int jobModelSegmentIndex = 0;; ++jobModelSegmentIndex) {
+        String jobModelSegmentKey = getJobModelSegmentKey(jobModelVersion, jobModelSegmentIndex);
+        byte[] jobModelSegment = metadataStore.get(jobModelSegmentKey);
+        if (jobModelSegment == null) {
+          break;
```
sborya: we need some kind of verification it is not corrupted. I saw someone suggested checksum?

sborya:
I am still not clear on the two questions brought up by @prateekm : why we don't need checksum and why the invocation of the chunking should not be moved to the implementation of a store.

samza-core/src/main/java/org/apache/samza/job/model/JobModelUtil.java
shanthoosh: Done.

samza-core/src/main/java/org/apache/samza/job/model/JobModelUtil.java
```diff
@@ -19,20 +19,33 @@
...
+import org.codehaus.jackson.map.ObjectMapper;
 
 /**
  * Utility class for the {@link JobModel}
  */
 public class JobModelUtil {
 
+  private static final ObjectMapper MAPPER = SamzaObjectMapper.getObjectMapper();
+
+  private static final int JOB_MODEL_SEGMENT_SIZE_IN_BYTES = 1020 * 1020;
```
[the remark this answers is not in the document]
shanthoosh: Zookeeper-Client injects additional metadata on top of the value written to zookeeper node. This allows the space for those additional fields. If we set it to 1024 here, it would still hit the zookeeper-server 1 MB data node limit and fail the job. I was found it during my testing and hence lowered it.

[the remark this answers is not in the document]
prateekm: It'll be better to add this comment in the code instead of magic numbers.

shanthoosh:
@sborya 
1. I've moved the chunking logic into the zookeeper metadata store implementation.
2. I've already replied to checksum comment. I will add check-sum support in a immediate follow-up patch. Also, this check-sum for identifying data corruption is not required for `JobModel` alone in standalone. It is applicable to all kind of metadata stored in YARN and standalone.

dnishimura:
In its current implementation, there are `ZkMetadataStore` and `CoordinatorStreamStore` metadata store types used in the `ZkJobCoordinator` with the former used to store the job model and the latter used to store the current configs and startpoints. Just a heads up @bharathkk that the new metadata store design will have to address this separation.

prateekm:
@shanthoosh If checksum is useful (as we think it is), why does it need to be a separate patch? Should be trivial to implement here?

shanthoosh:
@prateekm 
Added checksum to the end of metadata-value in  `ZkMetadataStore`. It is used for verification during read and published to zookeeper during write.

shanthoosh:
@sborya 
I addressed both of your comments. Can you please take a look.

samza-core/src/main/java/org/apache/samza/job/model/JobModelUtil.java
```diff
@@ -57,6 +67,42 @@
...
   }
 
+
+  /**
+   * Converts the JobModel into a byte array into {@link MetadataStore}.
+   * @param jobModel the job model to store into {@link MetadataStore}.
+   * @param jobModelVersion the job model version.
+   * @param metadataStore the metadata store.
+   */
+  public static void writeJobModel(JobModel jobModel, String jobModelVersion, MetadataStore metadataStore) {
```
prateekm: This is currently ZK specific, right? 

Also, does this still need to be in a separate util class or can this be inlined / helper method in the caller?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -203,6 +206,7 @@ public void stop() {
           coordinatorListener.onCoordinatorStop();
         }
 
+        jobModelMetaStore.close();
```
prateekm: Don't need to call start/init? Or is that already called?

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -148,6 +150,7 @@
             LocationIdProviderFactory.class);
     LocationIdProvider locationIdProvider = locationIdProviderFactory.getLocationIdProvider(config);
     this.locationId = locationIdProvider.getLocationId();
+    this.jobModelMetaStore = jobModelMetaStore;
```
prateekm: Nitpick: jobModelMetadataStore everywhere.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -18,27 +18,32 @@
...
-import org.slf4j.LoggerFactory;
 
 /**
  * An implementation of the {@link MetadataStore} interface where the
  * metadata of the Samza job is stored in zookeeper.
  */
 public class ZkMetadataStore implements MetadataStore {
 
-  private static final Logger LOG = LoggerFactory.getLogger(ZkMetadataStore.class);
+  private static final int VALUE_SEGMENT_SIZE_IN_BYTES = 1020 * 1020;
```
prateekm: Please document why this particular number. It's also better to express it as 1024 * 1024 - XXXX and document that XXXX is headroom for zk stuff.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -64,25 +69,47 @@ public void init() {
...
+      byte[] zNodeValue = zkClient.readData(zkPath, true);
+      if (zNodeValue == null) {
+        break;
+      }
+      aggregatedZNodeValues = Bytes.concat(aggregatedZNodeValues, zNodeValue);
+    }
+    if (aggregatedZNodeValues.length > 0) {
+      byte[] value = ArrayUtils.subarray(aggregatedZNodeValues, 0, aggregatedZNodeValues.length - CHECKSUM_SIZE_IN_BYTES);
+      byte[] checkSum = ArrayUtils.subarray(aggregatedZNodeValues, aggregatedZNodeValues.length - CHECKSUM_SIZE_IN_BYTES, aggregatedZNodeValues.length);
+      byte[] expectedCheckSum = getCRCChecksum(value);
```
prateekm: Nitpick: 'expectedChecksum' (capitalization)

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -64,25 +69,47 @@ public void init() {
...
+      if (!Arrays.equals(checkSum, expectedCheckSum)) {
+        throw new IllegalStateException("Expected checksum of value did not match the actual checksum");
```
prateekm: Log contents for debugging?

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -125,7 +145,40 @@ public void close() {
...
+  }
+
+  /**
+   * Splits the input byte array value into independent byte array segments of 1 MB size.
+   * @param value the input byte array to split.
+   * @return the byte array splitted into independent byte array chunks.
+   */
+  private static List<byte[]> chunkMetadataStoreValue(byte[] value) {
+    try {
+      byte[] valueCrcAsBytes = getCRCChecksum(value);
```
prateekm: Nitpick: 'checksum'

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -125,7 +145,40 @@ public void close() {
...
+      value = ArrayUtils.addAll(value, valueCrcAsBytes);
```
prateekm: Nitpick: prefer new variable instead of reassigning input params.

samza-core/src/test/java/org/apache/samza/zk/TestZkJobCoordinator.java
```diff
@@ -62,12 +68,15 @@
 import static org.mockito.Mockito.verifyZeroInteractions;
 import static org.mockito.Mockito.when;
 
+@RunWith(PowerMockRunner.class)
```
prateekm: Why does this need powermock? Possible to refactor to not need this?

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
```diff
@@ -562,4 +569,33 @@ public void testCloseShouldTearDownZkConnectionOnInterruptedException() throws E
 
     Assert.assertTrue(field.getBoolean(zkClient));
   }
+
+  @Test
+  public void testShouldSupportReadingAndWritingLargeJobModels() {
```
prateekm: Why is this test in TestZkUtils? This should be for whichever class the job model is being read/written from (e.g. ZkJC)

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -64,25 +69,47 @@ public void init() {
    */
   @Override
   public byte[] get(String key) {
```
prateekm: Please add unit tests for large get/put/delete and helper methods in this class.

samza-core/src/main/java/org/apache/samza/job/model/JobModelUtil.java
```diff
@@ -57,6 +67,42 @@
...
+    try {
+      String jobModelSerializedAsString = MAPPER.writerWithDefaultPrettyPrinter().writeValueAsString(jobModel);
```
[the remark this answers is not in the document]
shanthoosh: Default pretty printer would indent the `JobModel` json properly. This will aid debugging when we dump the `JobModel` json in the logs of followers.

samza-core/src/main/java/org/apache/samza/job/model/JobModelUtil.java
shanthoosh: I think this is not specific to standalone. It will be applicable to YARN as well when it has support for versioning.

 I'm using these util functions in standalone integration test-cases to validate the `JobModel` published by a processors group.

samza-core/src/test/java/org/apache/samza/zk/TestZkUtils.java
shanthoosh: This test is not necessary anymore. This is covered by `TestZkMetadataStore`.

samza-core/src/test/java/org/apache/samza/zk/TestZkJobCoordinator.java
shanthoosh: This was required for mocking the static methods of `JobModelUtil` class in `ZkJobCoordinator` execution sequence.

dnishimura: Question for @prateekm and @cameronlee314 (unrelated to this specific PR, but related to this comment)
In prior code reviews, @cameronlee314 mentioned that private methods should be static whenever possible so we can test/validate such methods easier. However, the only way to verify static methods are called in a test is to use PowerMock. What should be the proper convention for testing these type of static methods if PowerMock is discouraged?

cameronlee314: Sorry, I think I might have been a little unclear previously. By "validate", I was referring to when someone was checking correctness while reading the code, or having the compiler help to enforce certain usage. The "private static" methods aren't going to help with unit testing.
That said, it should be unnecessary to test "private static" methods directly since tests should theoretically only test against the public APIs of a class (although there are lots of exceptions to this, such as package-private VisibleForTesting methods). In this case, it is a static method being mocked, so we aren't directly testing a static method here. As @prateekm mentioned, verifications can be done against the objects that the static method works on. Or maybe if the static method is very complicated, it shouldn't be a static method, and it should be an object which can then be passed as a dependency to the object-under-test.

dnishimura: Cool makes sense. Thanks for clarifying!

shanthoosh: > Prefer not using PowerMock and mocking statics

Done. Removed usages of `PowerMock` by mocking individual methods using `Mockito`.

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -64,25 +79,50 @@ public void init() {
...
+        break;
+      }
+      aggregatedZNodeValues = Bytes.concat(aggregatedZNodeValues, zNodeValue);
+    }
+    if (aggregatedZNodeValues.length > 0) {
+      byte[] value = ArrayUtils.subarray(aggregatedZNodeValues, 0, aggregatedZNodeValues.length - CHECKSUM_SIZE_IN_BYTES);
+      byte[] checkSum = ArrayUtils.subarray(aggregatedZNodeValues, aggregatedZNodeValues.length - CHECKSUM_SIZE_IN_BYTES, aggregatedZNodeValues.length);
+      byte[] expectedChecksum = getCRCChecksum(value);
+      if (!Arrays.equals(checkSum, expectedChecksum)) {
+        String exceptionMessage = String.format("Expected checksum: %s of value: %s did not match the actual checksum: %s",
```
prateekm: Nitpick: Since the value can be very large, put the two checksums together in the beginning. "Expected checksum: %s did not match actual checksum: %s for value: %s".

shanthoosh: Fixed the nitpick.

shanthoosh:
Travis build for this patch is green, just that it is not getting updated here. 
https://travis-ci.org/apache/samza/builds/549945381?utm_source=github_status&utm_medium=notification

PR #1080, opened 2019-06-11
SAMZA-2251: Minor diagnostics manager change to emit additional job details

rmatharu:


rmatharu:
@jagadish-v0 minor change to make the diagnostics manager emit additional data.

PR #1083, opened 2019-06-13
SAMZA-2251: Minor diagnostics manager change to emit additional job details

rmatharu:


rmatharu:
@prateekm can you please take a look, thanks

samza-core/src/test/java/org/apache/samza/serializers/model/serializers/TestMetricsSnapshotSerdeV2.java
```diff
@@ -53,17 +91,44 @@ public void testSerde() {
...
     MetricsSnapshot deserializedMetricsSnapshot = metricsSnapshotSerde.fromBytes(serializedBytes);
 
-    Assert.assertTrue("Headers map should be equal",
-        metricsSnapshot.getHeader().getAsMap().equals(deserializedMetricsSnapshot.getHeader().getAsMap()));
+    Assert.assertTrue(metricsSnapshot.getHeader().getAsMap().equals(deserializedMetricsSnapshot.getHeader().getAsMap()));
+  }
 
-    Assert.assertTrue("Metrics map should be equal",
-        metricsSnapshot.getMetrics().getAsMap().equals(deserializedMetricsSnapshot.getMetrics().getAsMap()));
+  HashMap convertMap(Map<String, Object> map) {
```
cameronlee314: 1) Can make this a "private static" method.
2) What is this useful for? It looks like it is copying all maps (incl. nested) into `HashMap`s in order to generate the serialized JSON for the test. But in the prod scenario, this doesn't get used, so is the test adding something extra that shouldn't be there?

samza-core/src/test/java/org/apache/samza/serializers/model/serializers/TestMetricsSnapshotSerdeV2.java
```diff
@@ -19,26 +19,64 @@
...
+    String samzaContainerMetricsGroupName = "org.apache.samza.container.SamzaContainerMetrics";
+    Map<String, Map<String, Object>> metricMessage = new HashMap<>();
+    metricMessage.put(samzaContainerMetricsGroupName, new HashMap<>());
+    metricMessage.get(samzaContainerMetricsGroupName).put("exceptions", boundedList.getValues());
+    metricMessage.get(samzaContainerMetricsGroupName).put("commit-calls", 0);
+    MetricsSnapshot metricsSnapshot = new MetricsSnapshot(metricsHeader, new Metrics(metricMessage));
+
+    ObjectMapper test = new ObjectMapper();
+    test.enableDefaultTyping(ObjectMapper.DefaultTyping.OBJECT_AND_NON_CONCRETE);
+    byte[] testBytes = test.writeValueAsString(convertMap(metricsSnapshot.getAsMap())).getBytes("UTF-8");
```
cameronlee314: It looks like you manually constructing the JSON string to deserialize, so it doesn't seem like you are testing the serialization flow. It does look like the serialization flow will exclude container model info after the first send, so it seems like that would be a useful flow to test directly as well.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
```diff
@@ -141,36 +171,83 @@ public void addExceptionEvent(DiagnosticsExceptionEvent diagnosticsExceptionEven
...
 
     @Override
     public void run() {
+
+      Map<String, Map<String, Object>> metricsMessage = new HashMap<>();
+
+      // Publish job-related params (if not already published)
+      if (!jobParamsEmitted) {
+        metricsMessage.putIfAbsent(GROUP_NAME_FOR_DIAGNOSTICS_MANAGER, new HashMap<>());
+        metricsMessage.get(GROUP_NAME_FOR_DIAGNOSTICS_MANAGER).put(CONTAINER_COUNT_METRIC_NAME, containerCount);
```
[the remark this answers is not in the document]
rmatharu: yes, since messages from container and AM can arrive in any order
publishing from both made the read logic simple
Since its only published once per container, i didn't think there was significant overhead.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
```diff
@@ -141,36 +171,83 @@ public void addExceptionEvent(DiagnosticsExceptionEvent diagnosticsExceptionEven
...
+  }
+
   private class DiagnosticsStreamPublisher implements Runnable {
 
     @Override
     public void run() {
+
+      Map<String, Map<String, Object>> metricsMessage = new HashMap<>();
+
+      // Publish job-related params (if not already published)
```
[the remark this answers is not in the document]
rmatharu: Yes, the reader assumes the data is time retention and the onus of reading consistently is on the reader.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
```diff
@@ -141,36 +171,83 @@ public void addExceptionEvent(DiagnosticsExceptionEvent diagnosticsExceptionEven
...
+      if (!jobParamsEmitted) {
+        metricsMessage.putIfAbsent(GROUP_NAME_FOR_DIAGNOSTICS_MANAGER, new HashMap<>());
```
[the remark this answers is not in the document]
rmatharu: Yup, totally agree. 
I've tried to add DiagnosticsStreamMessage pojo to define and enforce a contract on what does into the stream from the diagnostics-manager.

samza-core/src/test/java/org/apache/samza/serializers/model/serializers/TestMetricsSnapshotSerdeV2.java
rmatharu: Added

rmatharu:
@cameronlee314 Addressed all comments, defined a pojo for the stream contract, which simpified the change a fair bit. Please take a look.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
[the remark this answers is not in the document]
cameronlee314: What happens if there is a new reader that starts in the middle of a stream? It will be missing this "initial" info. Is that handled ok?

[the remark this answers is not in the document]
rmatharu: in case of the reader being diagnostics, it is handled. 
in case of autosizing being the reader, it will be handled.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
```diff
@@ -0,0 +1,305 @@
...
+  //   * @param metricsSnapshot
+  public static DiagnosticsStreamMessage convertToDiagnosticsStreamMessage(MetricsSnapshot metricsSnapshot) {
+    DiagnosticsStreamMessage diagnosticsStreamMessage =
+        new DiagnosticsStreamMessage(metricsSnapshot.getHeader().getJobName(), metricsSnapshot.getHeader().getJobId(),
+            metricsSnapshot.getHeader().getContainerName(), metricsSnapshot.getHeader().getExecEnvironmentContainerId(),
+            metricsSnapshot.getHeader().getVersion(), metricsSnapshot.getHeader().getSamzaVersion(),
+            metricsSnapshot.getHeader().getHost(), metricsSnapshot.getHeader().getTime(),
+            metricsSnapshot.getHeader().getResetTime());
+
+    if (metricsSnapshot.getMetrics().getAsMap().containsKey(GROUP_NAME_FOR_DIAGNOSTICS_MANAGER)) {
```
cameronlee314: Can you extract `metricsSnapshot.getMetrics().getAsMap()` so you don't create it multiple times?

cameronlee314: You could also get the GROUP_NAME_FOR_DIAGNOSTICS_MANAGER entry and reuse that. It might just be good to check if the GROUP_NAME_FOR_DIAGNOSTICS_MANAGER entry is null instead of checking `containsKey`, since a map can still "containsKey" and have a null value.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
```diff
@@ -0,0 +1,305 @@
...
+   * @param containerModelMap map of container models to serialize.
+   * @return
+   */
+  private static Map<String, String> serializeContainerModelMap(Map<String, ContainerModel> containerModelMap) {
+    Map<String, String> serializedContainerModelMap = new HashMap<>();
+    ObjectMapper samzaObjectMapper = SamzaObjectMapper.getObjectMapper();
+    String serializedContainerModel;
+
+    for (Map.Entry<String, ContainerModel> containerModelEntry : containerModelMap.entrySet()) {
+      serializedContainerModel = "";
```
cameronlee314: This empty string might get put into the map. Will Jackson read that correctly?

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
```diff
@@ -0,0 +1,305 @@
...
+    return diagnosticsStreamMessage;
+  }
+
+  /**
+   * Helper method to use {@link SamzaObjectMapper} to serialize {@link ContainerModel}s.
+   * {@link SamzaObjectMapper} provides several conventions and optimizations for serializing containerModels.
+   * @param containerModelMap map of container models to serialize.
+   * @return
+   */
+  private static Map<String, String> serializeContainerModelMap(Map<String, ContainerModel> containerModelMap) {
```
cameronlee314: Would it be simpler to just return a `String` here? Just serialize the whole `containerModelMap` using `SamzaObjectMapper`.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
```diff
@@ -0,0 +1,305 @@
...
+    return containerModelMap;
+  }
+
+  private void addToMetricsMessage(String groupName, String metricName, Object value) {
+    metricsMessage.putIfAbsent(groupName, new HashMap<>());
+    metricsMessage.get(groupName).put(metricName, value);
+  }
+
+  private Object getFromMetricsMessage(String groupName, String metricName) {
+    if (metricsMessage.containsKey(groupName)) {
```
cameronlee314: Be careful here too with map entries that have a value of null.

samza-core/src/test/java/org/apache/samza/diagnostics/TestDiagnosticsManager.java
```diff
@@ -0,0 +1,122 @@
...
+
+  @Before
+  public void setup() {
+
+    // Mocked system producer for publishing to diagnostics stream
+    mockSystemProducer = new MockCoordinatorStreamSystemFactory.MockSystemProducer("source");
+
+    // Mocked scheduled executor service which does a synchronous run() on scheduling
+    ScheduledExecutorService mockExecutorService = Mockito.mock(ScheduledExecutorService.class);
+    Mockito.when(mockExecutorService.scheduleWithFixedDelay(Mockito.argThat(new ArgumentMatcher<Runnable>() {
```
cameronlee314: Use `thenAnswer` instead of a matcher to execute arbitrary logic.

samza-core/src/test/java/org/apache/samza/diagnostics/TestDiagnosticsManager.java
```diff
@@ -0,0 +1,122 @@
...
+        new DiagnosticsManager(jobName, jobId, containerModels, containerMb, containerNumCores, numStoresWithChangelog,
+            "0", executionEnvContainerId, taskClassVersion, samzaVersion, hostname, diagnosticsSystemStream,
+            mockSystemProducer, Duration.ofSeconds(1), mockExecutorService);
+  }
+
+  @Test
+  public void testDiagnosticsStreamPublish() {
+    // invoking start will do a syncrhonous publish to the stream because of our mocked scheduled exec service
+    this.diagnosticsManager.start();
+    Assert.assertEquals("One message should have been published", 1, mockSystemProducer.getEnvelopes().size());
```
cameronlee314: Should you call it twice to make sure the container info doesn't get published for the second time?

samza-core/src/test/java/org/apache/samza/diagnostics/TestDiagnosticsManager.java
```diff
@@ -0,0 +1,122 @@
...
+    Assert.assertEquals(metricsSnapshot.getHeader().getSamzaVersion(), samzaVersion);
+    Assert.assertEquals(metricsSnapshot.getHeader().getHost(), hostname);
+    Assert.assertEquals(metricsSnapshot.getHeader().getSource(), DiagnosticsManager.class.getName());
+
+    // Validate the diagnostics stream message
+    DiagnosticsStreamMessage diagnosticsStreamMessage =
+        DiagnosticsStreamMessage.convertToDiagnosticsStreamMessage(metricsSnapshot);
+
+    Assert.assertEquals(containerMb, diagnosticsStreamMessage.getContainerMb());
+    Assert.assertTrue(diagnosticsStreamMessage.getExceptionEvents() == null);
```
cameronlee314: `Assert.assertNull`

samza-core/src/test/java/org/apache/samza/diagnostics/TestDiagnosticsManager.java
```diff
@@ -0,0 +1,122 @@
...
+    Assert.assertTrue(diagnosticsStreamMessage.getProcessorStopEvents() == null);
```
cameronlee314: Can you add a test for when these do exist?

samza-core/src/test/java/org/apache/samza/diagnostics/TestDiagnosticsStreamMessage.java
```diff
@@ -0,0 +1,133 @@
...
+    Set<SystemStreamPartition> sspsForTask2 = new HashSet<>();
+    sspsForTask2.add(new SystemStreamPartition("kafka", "test-stream", new Partition(1)));
+    tasks.put(new TaskName("Partition 1"), new TaskModel(new TaskName("Partition 1"), sspsForTask2, new Partition(1)));
+
+    containerModels.put("0", new ContainerModel("0", tasks));
+    return containerModels;
+  }
+
+  @Test
+  /**
```
cameronlee314: Could you please move the comment above the annotation?

samza-core/src/test/java/org/apache/samza/diagnostics/TestDiagnosticsStreamMessage.java
```diff
@@ -0,0 +1,133 @@
...
+    Assert.assertEquals(1024, diagnosticsStreamMessage.getContainerMb());
+    Assert.assertEquals(2, diagnosticsStreamMessage.getContainerNumCores());
+    Assert.assertEquals(3, diagnosticsStreamMessage.getNumStoresWithChangelog());
+    Assert.assertTrue(exceptionEventList.equals(diagnosticsStreamMessage.getExceptionEvents()));
+    Assert.assertEquals(getSampleContainerModels(), diagnosticsStreamMessage.getContainerModels());
+    Assert.assertEquals(diagnosticsStreamMessage.getProcessorStopEvents(), getProcessorStopEventList());
+  }
+
+  @Test
+  /**
```
cameronlee314: same as above about moving the comment

samza-core/src/test/java/org/apache/samza/diagnostics/TestDiagnosticsStreamMessage.java
```diff
@@ -0,0 +1,133 @@
...
+   * Tests serialization and deserialization of a {@link DiagnosticsStreamMessage} using {@link MetricsSnapshotSerdeV2}.
+   * This serde is currently used by {@link DiagnosticsManager} to serde messages to the diagnostics stream.
+   */ public void serdeTest() {
+    DiagnosticsStreamMessage diagnosticsStreamMessage = getDiagnosticsStreamMessage();
+    Collection<DiagnosticsExceptionEvent> exceptionEventList = getExceptionList();
+    diagnosticsStreamMessage.addDiagnosticsExceptionEvents(exceptionEventList);
+    diagnosticsStreamMessage.addProcessorStopEvents(getProcessorStopEventList());
+    diagnosticsStreamMessage.addContainerModels(getSampleContainerModels());
+
+    byte[] serializedBytes = new MetricsSnapshotSerdeV2().toBytes(diagnosticsStreamMessage.convertToMetricsSnapshot());
```
cameronlee314: The API for `DiagnosticsStreamMessage` is `convertToMetricsSnapshot` and `convertToDiagnosticsStreamMessage`, so you shouldn't need to call `MetricsSnapshotSerdeV2` here.
In my opinion, this test should have the following flow:
1) `convertToMetricsSnapshot`
2) check that the resulting `MetricsSnapshot` is correct
3) `convertToDiagnosticsStreamMessage`
4) check that the resulting `DiagnosticsStreamMessage` matches the original

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
rmatharu: yes, that gets handled; so that we can get the container count, even if the container-model doesnt get serialized for any reason.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
```diff
@@ -243,11 +219,33 @@ public void run() {
     public final String host;
```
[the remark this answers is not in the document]
rmatharu: Yes, to access them on the deserde path.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
rmatharu: I thought itd be the same since samza-object-mapper and the serde are jackson anyways, so i kept it as a map, where the key count = num of containers.

samza-core/src/test/java/org/apache/samza/diagnostics/TestDiagnosticsManager.java
rmatharu: Yup, added additional test.

rmatharu:
All done.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
cameronlee314: Looking at `deserializeContainerModelMap`, an entry that can't be deserialized seems to get skipped. So it doesn't look like you would be able to get the correct container count if some or all did not get put into the map.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
cameronlee314: I think it would be the same, but by returning the `String`, the implementation of `serializeContainerModelMap` just becomes `return SamzaObjectMapper.getObjectMapper().writeValueAsString(containerModelMap)`.

samza-core/src/test/java/org/apache/samza/diagnostics/TestDiagnosticsManager.java
```diff
@@ -52,44 +55,96 @@
...
+    outgoingMessageEnvelope = mockSystemProducer.getEnvelopeList().get(1);
+    validateMetricsHeader(outgoingMessageEnvelope);
+
+    // Validate the second message's body (should be all empty except for the processor-stop-event)
+    MetricsSnapshot metricsSnapshot =
+        new MetricsSnapshotSerdeV2().fromBytes((byte[]) outgoingMessageEnvelope.getMessage());
+    DiagnosticsStreamMessage diagnosticsStreamMessage =
+        DiagnosticsStreamMessage.convertToDiagnosticsStreamMessage(metricsSnapshot);
+
+    Assert.assertTrue(diagnosticsStreamMessage.getContainerMb() == null);
```
cameronlee314: Minor: `Assert.assertNull`

samza-core/src/test/java/org/apache/samza/diagnostics/TestDiagnosticsManager.java
```diff
@@ -103,20 +158,56 @@ public void testDiagnosticsStreamPublish() {
...
+        new MetricsSnapshotSerdeV2().fromBytes((byte[]) outgoingMessageEnvelope.getMessage());
+
     // Validate the diagnostics stream message
     DiagnosticsStreamMessage diagnosticsStreamMessage =
         DiagnosticsStreamMessage.convertToDiagnosticsStreamMessage(metricsSnapshot);
 
-    Assert.assertEquals(containerMb, diagnosticsStreamMessage.getContainerMb());
-    Assert.assertTrue(diagnosticsStreamMessage.getExceptionEvents() == null);
-    Assert.assertTrue(diagnosticsStreamMessage.getProcessorStopEvents() == null);
+    Assert.assertTrue(diagnosticsStreamMessage.getContainerMb().equals(containerMb));
```
cameronlee314: Minor: `Assert.assertEquals`

samza-core/src/test/java/org/apache/samza/diagnostics/TestDiagnosticsStreamMessage.java
```diff
@@ -93,41 +92,57 @@ private DiagnosticsStreamMessage getDiagnosticsStreamMessage() {
...
     DiagnosticsStreamMessage diagnosticsStreamMessage = getDiagnosticsStreamMessage();
     Collection<DiagnosticsExceptionEvent> exceptionEventList = getExceptionList();
     diagnosticsStreamMessage.addDiagnosticsExceptionEvents(exceptionEventList);
     diagnosticsStreamMessage.addProcessorStopEvents(getProcessorStopEventList());
     diagnosticsStreamMessage.addContainerModels(getSampleContainerModels());
 
-    Assert.assertEquals(1024, diagnosticsStreamMessage.getContainerMb());
-    Assert.assertEquals(2, diagnosticsStreamMessage.getContainerNumCores());
-    Assert.assertEquals(3, diagnosticsStreamMessage.getNumStoresWithChangelog());
+    Assert.assertTrue(diagnosticsStreamMessage.getContainerMb().equals(1024));
```
cameronlee314: Minor: `Assert.assertEquals`

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
rmatharu: agree, tried simplifying this.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
rmatharu: agree, tried simplifying this.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
```diff
@@ -236,23 +228,15 @@ public static DiagnosticsStreamMessage convertToDiagnosticsStreamMessage(Metrics
    * @param containerModelMap map of container models to serialize.
    * @return
    */
-  private static Map<String, String> serializeContainerModelMap(Map<String, ContainerModel> containerModelMap) {
-    Map<String, String> serializedContainerModelMap = new HashMap<>();
+  private static String serializeContainerModelMap(Map<String, ContainerModel> containerModelMap) {
```
cameronlee314: Do you have a javadoc somewhere explaining why you need do serialize explicitly instead of having Jackson do it on its own? (i.e. need to use default typing for backwards compatibility which doesn't work with some of the Samza data models like Partition since it doesn't have an empty constructor)
If you don't have that detail somewhere, could you please add it somewhere in here? This will be useful context for someone who is looking into this code in the future.

rmatharu: added

PR #1084, opened 2019-06-17
Samza-2254: Adding relevant calcite functions to Samza SQL

srinipunuru:
There were quite a few basic calcite sql functions that were not included in the original Samza SQL operators that are supported. 
1. Adding these functions to the operator table. 
2. Making the Samza SQL parser lenient to allow !=, %, - and row value operators
3. Adding SqlToRelConverterConfig explicitly so that it is clear that we have opportunity to override this in future if needed.


vjagadish1989:
for better tracking, Consider creating a JIRA and linking it in the PR title?

weiqingy:
Per the CI build, there is a test failure:
```
testEndToEndWithLike FAILED
    java.lang.AssertionError: expected:<20> but was:<0>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at org.apache.samza.test.samzasql.TestSamzaSqlEndToEnd.testEndToEndWithLike(TestSamzaSqlEndToEnd.java:371)
```

PR #1085, opened 2019-06-18
Samza-2253: Upgrading calcite to 1.19

srinipunuru:
Upgrading calcite to latest build to include bug fixes with compound boolean checks and others. Subqueries with flatten needed some change to get it working.


samza-test/src/test/java/org/apache/samza/test/samzasql/TestSamzaSqlEndToEnd.java
```diff
@@ -417,6 +416,8 @@ public void testEndToEndNestedRecord() {
     Assert.assertEquals(numMessages, outMessages.size());
   }
 
+  // SAMZA-2252 - Samza SQL subquery with flatten doesn't work after calcite 1.19
```
atoomula: This is required for Ambry use-case. What does it take to get it working ? Shall we wait on this change until the fix ?

srinipunuru: I fixed this issue. https://issues.apache.org/jira/browse/SAMZA-2252

atoomula: That's super quick fix. Thanks!

PR #1087, opened 2019-06-20
SAMZA-2258: Log thread dump when container stop times out during processor rebalancing

cameronlee314:


samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
```diff
@@ -388,8 +389,11 @@ private boolean stopSamzaContainer() {
     // we propagate to the application runner maybe overwritten by container failure cause in case of interleaved execution.
     // It is acceptable since container exception is much more useful compared to timeout exception.
     // We can infer from the logs about the fact that container shutdown timed out or not for additional inference.
-    if (!hasContainerShutdown && containerException == null) {
-      containerException = new TimeoutException("Container shutdown timed out after " + taskShutdownMs + " ms.");
+    if (!hasContainerShutdown) {
+      ThreadUtil.logThreadDump("Thread dump at failure for stopping container in stream processor");
```
cameronlee314: @vjagadish1989 do you think we should only do a thread dump when the processor is in certain states? The change here applies to rebalancing and shutting down. `SamzaContainer` does have its own thread dump when it doesn't shut down in time, so sometimes we would see a duplicate thread dump if we put this here.
An alternative would be to put this thread dump inside `onJobModelExpired` or check the state of the processor before initiating the thread dump. Are there cases other than just rebalancing where we would want this thread dump?

shanthoosh:
@cameronlee314 
There are checkstyle errors that fails the build. Can you please fix them.

cameronlee314:
> @cameronlee314
> There are checkstyle errors that fails the build. Can you please fix them.

Oops, missed that. Thanks for pointing it out. PR is updated.

samza-core/src/main/java/org/apache/samza/processor/StreamProcessor.java
vjagadish1989: > Are there cases other than just rebalancing

For now, we can stick to logging thread-dumps on failed rebalances, since the value is obvious. 

> so sometimes we would see a duplicate thread dump if we put this here

yeah, usually a thread-dump closest to the error (like in this PR for example) is the most useful. The other thread-dumps are a part of UncaughtExceptionHandler, so may or may not get logged depending on whether the process is shutting down

PR #1090, opened 2019-06-26
SAMZA-2260: Standalone - coordinator stream metadata store lifecycle

dnishimura:
**Motivation**
The current implementation of the coordinator stream metadata store's `init()` method is an expensive call as it bootstraps the full coordinator stream. We want to invoke init() only once during the entire lifecycle of a standalone application.

**Implementation**
This PR instantiates the coordinator stream metadata store from the scope of the top-level `LocalApplicationRunner`. This allows other components, such as `StreamProcessor`, `JobCoordinator`, `SamzaContainer`, `StartpointManager` and other managers, to use the coordinator stream metadata store freely without needing to manage its lifecycle.

**Note**
Currently there are split responsibilities between a ZK metadata store and the coordinator stream metadata store. The former is required for JobModel and run.id because of the versioning requirements that the coordinator stream metadata store cannot provide. The next-gen metadata store will address this requirement. The latter is the focus of this PR since its init() call is expensive.

Please take a look @shanthoosh 
CC: @bharathkk

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
@@ -34,7 +34,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.samza.SamzaException;
 import org.apache.samza.application.SamzaApplication;
@@ -84,7 +83,7 @@ public class LocalApplicationRunner implements ApplicationRunner {
   private final AtomicReference<Throwable> failure = new AtomicReference<>();
   private final boolean isAppModeBatch;
   private final Optional<CoordinationUtils> coordinationUtils;
-  private final MetadataStoreFactory coordinatorMetadataStoreFactory;
+  private final MetadataStoreFactory metadataStoreFactory;
   private Optional<String> runId = Optional.empty();
   private Optional<RunIdGenerator> runIdGenerator = Optional.empty();
 
@@ -111,8 +110,9 @@ public class LocalApplicationRunner implements ApplicationRunner {
     this.appDesc = ApplicationDescriptorUtil.getAppDescriptor(app, config);
     this.isAppModeBatch = new ApplicationConfig(config).getAppMode() == ApplicationConfig.ApplicationMode.BATCH;
     this.coordinationUtils = getCoordinationUtils(config, getClass().getClassLoader());
-    this.coordinatorMetadataStoreFactory = metadataStoreFactory;
+    this.metadataStoreFactory = metadataStoreFactory;
   }
+
   /**
    * Constructor only used in unit test to allow injection of {@link LocalJobPlanner}
    */
@@ -121,7 +121,7 @@ public class LocalApplicationRunner implements ApplicationRunner {
     this.appDesc = appDesc;
     this.isAppModeBatch = new ApplicationConfig(appDesc.getConfig()).getAppMode() == ApplicationConfig.ApplicationMode.BATCH;
     this.coordinationUtils = coordinationUtils;
-    this.coordinatorMetadataStoreFactory = new CoordinatorStreamMetadataStoreFactory();
+    this.metadataStoreFactory = new CoordinatorStreamMetadataStoreFactory();
   }
 
   private Optional<CoordinationUtils> getCoordinationUtils(Config config, ClassLoader classLoader) {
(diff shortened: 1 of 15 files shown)


samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -75,12 +78,13 @@
...
 
   private final ApplicationDescriptorImpl<? extends ApplicationDescriptor> appDesc;
-  private final Set<StreamProcessor> processors = ConcurrentHashMap.newKeySet();
+  private final Set<Pair<StreamProcessor, MetadataStore>> processors = ConcurrentHashMap.newKeySet();
   private final CountDownLatch shutdownLatch = new CountDownLatch(1);
   private final AtomicInteger numProcessorsToStart = new AtomicInteger();
   private final AtomicReference<Throwable> failure = new AtomicReference<>();
   private final boolean isAppModeBatch;
   private final Optional<CoordinationUtils> coordinationUtils;
+  private final MetadataStoreFactory coordinatorMetadataStoreFactory;
```
shanthoosh: nit: 
s/coordinatorMetadataStoreFactory/metadataStoreFactory

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -240,16 +261,26 @@ CountDownLatch getShutdownLatch() {
...
   @VisibleForTesting
-  StreamProcessor createStreamProcessor(Config config, ApplicationDescriptorImpl<? extends ApplicationDescriptor> appDesc,
+  MetadataStore createCoordinatorStreamStore(Config jobConfig) {
+    MetadataStore coordinatorStreamStore =
+        coordinatorMetadataStoreFactory.getMetadataStore("NoOp", jobConfig, new MetricsRegistryMap());
+    return coordinatorStreamStore;
+  }
+
+  @VisibleForTesting
+  Pair<StreamProcessor, MetadataStore> createStreamProcessor(Config config, ApplicationDescriptorImpl<? extends ApplicationDescriptor> appDesc,
```
shanthoosh: Just curious. 

1. Why are we passing in the metadata store as a argument to this API and returning it back to the caller of the API again? Would prefer to remove MetadataStore from the result of this API if it makes sense.
 2. Also, please note that LocalApplicationRunner at its current state can only manage a life-cycle of one stream processor. For more details, please look at how the exit of stream processors are handled in LocalApplicationRunner.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -308,38 +312,32 @@ JobModel readJobModelFromMetadataStore(String zkJobModelVersion) {
...
 
-      CoordinatorStreamValueSerde jsonSerde = new CoordinatorStreamValueSerde(SetConfig.TYPE);
-      NamespaceAwareCoordinatorStreamStore configStore =
-          new NamespaceAwareCoordinatorStreamStore(coordinatorStreamStore, SetConfig.TYPE);
-      for (Map.Entry<String, String> entry : config.entrySet()) {
-        byte[] serializedValue = jsonSerde.toBytes(entry.getValue());
-        configStore.put(entry.getKey(), serializedValue);
-      }
+      if (coordinatorStreamStore != null) {
+        CoordinatorStreamValueSerde jsonSerde = new CoordinatorStreamValueSerde(SetConfig.TYPE);
```
shanthoosh: We'd discussed  offline that publishing the heavy-weight configuration bag one-by-one through metadata-store API in leader-startup sequence had significantly slowed down the processor startup. Please create a JIRA ticket for it and add a TODO here.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorFactory.java
```diff
@@ -37,20 +38,18 @@
...
+  public JobCoordinator getJobCoordinator(String processorId, Config config, MetricsRegistry metricsRegistry,
+      MetadataStore coordinatorStreamStore) {
     // TODO: Separate JC related configs into a "ZkJobCoordinatorConfig"
     String jobCoordinatorZkBasePath = getJobCoordinationZkPath(config);
     ZkUtils zkUtils = getZkUtils(config, metricsRegistry, jobCoordinatorZkBasePath);
     LOG.debug("Creating ZkJobCoordinator with config: {}.", config);
-    ZkMetadataStore metadataStore = new ZkMetadataStore(zkUtils.getKeyBuilder().getRootPath(), config, metricsRegistry);
-    return new ZkJobCoordinator(processorId, config, metricsRegistry, zkUtils, metadataStore);
+
+    // TODO: This should be merged with coordinatorStreamStore
```
shanthoosh: Would be a good idea to create a JIRA for the above TODO and link it here. Please assign the JIRA to rightful owner.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
dnishimura: 1. Wanted to pair the store with the processor, but I can do the pairing outside of the result. Will change.
2. Where exacly should I look at? From the looks of the overall code, all the stream processors are kept in a `Set` and iterated over during lifecycle related calls like `kill()`. I don't see where the logic of "only one stream processor lifecycle is managed"

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
dnishimura: Will do. I'll link the ticket to move this logic in the timer scheduler.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinatorFactory.java
dnishimura: Sure will do.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
shanthoosh: >> From the looks of the overall code

Take a look at the life-cycle methods in `LocalApplicationRunner`. A `LocalApplicationRunner` instance can only manage one instance of `StreamApplication`. Launching multiple stream application through a single `LocalApplicationRunner` will lead to internal-state corruption in `LAR`.

 Currently samza only supports single stage jobs. Hence the `Set` you're referring to `LocalApplicationRunner` will be always of size `1`.  

Above point was the motivation behind finding out the rationale behind `pairing` store with a processor.

PR #1091, opened 2019-06-26
Minor: Increase Travis-CI timeout to 40 min.

dnishimura:
The builds are close to hitting the 30 min timeout limit. This increases the limit to 40 min.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ jdk:
 
 script:
   ## travis_wait increases build idle-wait time from 10 minutes to 30 minutes.
-  - travis_wait 30 ./gradlew clean build
+  - travis_wait 40 ./gradlew clean build
 #  - type sonar-scanner &>/dev/null; if [ $? -eq 0 ]; then sonar-scanner; else echo "Not running sonar"; fi
 
 before_cache:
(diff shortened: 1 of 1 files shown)


rmatharu:
Has there been a recent test addition/change that requires the increased timeout?

dnishimura:
> Has there been a recent test addition/change that requires the increased timeout?

@rmatharu I think just the slow accumulation of tests that uses timers.

dnishimura:
If we had admin access to the apache/samza org in travis-CI, we could probably move the integration tests (which take up majority of the build time) to a nightly travis cron job. It'll probably require Samza PMC intervention with the Apache org.

.travis.yml
```diff
@@ -31,7 +31,7 @@ jdk:
 
 script:
   ## travis_wait increases build idle-wait time from 10 minutes to 30 minutes.
```
prateekm: @dnishimura Let's update the comment.

prateekm: Nvm, tried editing directly on Github and it worked. Thanks for the fix.

prateekm:
@dnishimura Do you know if there is a travis dashboard somewhere to see which tests fail how often?

PR #1092, opened 2019-06-26
SAMZA-2262: Change container-startup-time to a gauge from a timer and to ms instead of ns

rmatharu:
This metric is the diff of System.nanoTime() which is a positive long (since the number of nano seconds for which a JVM a running shouldnt be negative). 
Therefore I am using a long for it.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
+++ b/samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
@@ -758,7 +758,7 @@ class SamzaContainer(
       if (containerListener != null) {
         containerListener.afterStart()
       }
-      metrics.containerStartupTime.update(System.nanoTime() - startTime)
+      metrics.containerStartupTime.set(System.nanoTime() - startTime)
       if (taskInstances.size > 0)
         runLoop.run
       else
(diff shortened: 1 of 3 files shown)


samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -758,7 +758,7 @@ class SamzaContainer(
       if (containerListener != null) {
         containerListener.afterStart()
       }
-      metrics.containerStartupTime.update(System.nanoTime() - startTime)
+      metrics.containerStartupTime.set(System.nanoTime() - startTime)
```
prateekm: Let's change this to use currentTimeMillis as well. I think we're pretty far from sub-ms startup times :)

vjagadish1989:
nit-pick: fix PR title

either: "SAMZA-1234: ..." or "Minor: ..."

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
rmatharu: done

rmatharu:
Done

vjagadish1989:
nit-nit-pick:  s/samza/SAMZA

PR #1095, opened 2019-06-27
SAMZA-2264 : Offset.default for intermediate streams

rmatharu:


samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
```diff
@@ -125,7 +126,12 @@ Config generateConfig() {
     if (isIntermediate()) {
       streamConfig.put(String.format(StreamConfig.IS_INTERMEDIATE_FOR_STREAM_ID(), spec.getId()), "true");
       streamConfig.put(String.format(StreamConfig.DELETE_COMMITTED_MESSAGES_FOR_STREAM_ID(), spec.getId()), "true");
-      streamConfig.put(String.format(StreamConfig.CONSUMER_OFFSET_DEFAULT_FOR_STREAM_ID(), spec.getId()), "oldest");
+
+      // Setting offset.default to oldest only if the job is running in batch mode
+      if (new ApplicationConfig(config).getAppMode().equals(ApplicationConfig.ApplicationMode.BATCH)) {
```
mynameborat: nit: I'd suggest to do this other way round to guard against getAppMode() returning potential nulls even though currently it seems to return STREAM by default.
`ApplicationConfig.ApplicationMode.BATCH.equals(...)`

samza-core/src/main/java/org/apache/samza/execution/StreamEdge.java
```diff
@@ -125,7 +126,12 @@ Config generateConfig() {
...
+        streamConfig.put(String.format(StreamConfig.CONSUMER_OFFSET_DEFAULT_FOR_STREAM_ID(), spec.getId()), "oldest");
```
mynameborat: nit: extract spec.getId() into a local variable?

PR #1103, opened 2019-07-11
SAMZA-2270: Fix ConsoleLoggingSystemFactory to handle null message value.

atoomula:


samza-tools/src/main/java/org/apache/samza/tools/ConsoleLoggingSystemFactory.java
```diff
@@ -85,7 +85,7 @@ public void register(String source) {
     @Override
     public void send(String source, OutgoingMessageEnvelope envelope) {
       String msg = String.format("OutputStream:%s Key:%s Value:%s", envelope.getSystemStream(), envelope.getKey(),
-          new String((byte[]) envelope.getMessage()));
+          envelope.getMessage() != null ? new String((byte[]) envelope.getMessage()) : null);
```
shanthoosh: Doesn't this non-null check on the message has to be done at other places as well in this class. For instance, in getFormattedValue implementation(line number: 100)

atoomula: Good catch. Fixed it here as well. Thanks!

PR #1104, opened 2019-07-11
SAMZA-2266: Introduce a backoff when there are repeated failures for host-affinity allocations

dnishimura:
**Motivation**
For host-affinity enabled jobs, a bad physical host may not immediately be marked as invalid by the Resource Manager (RM). As a result, when the `HostAwareContainerAllocator` requests preferred hosts, the RM generates the `onResourceCompleted` callback even though the host can't be allocated. The status error in the `onResourceCompleted` is equivalent to an application error and the retry logic kicks in to restart the failed container. Adding delays in the retry logic will prevent the job from failing prematurely (after 8 retries) before the bad host is marked invalid.

**Implementation notes**
Added an exponential back-off with a max delay. Container allocation requests are put in a priority queue with the priority determined by type and request timestamp. For retries that have a delay, I set the request timestamp in the future by time X where X is the calculated back-off.

**Testing**
Unit tests and tested a Samza job on a YARN cluster. I simulated the scenario by forcing an uncaught exception in a few containers to force the containers to fail.

@rmatharu @abhishekshivanna and others please take a look


samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
```diff
@@ -87,7 +88,10 @@
    */
   public static final String CONTAINER_RETRY_COUNT = "yarn.container.retry.count";
   public static final String CLUSTER_MANAGER_CONTAINER_RETRY_COUNT = "cluster-manager.container.retry.count";
-  private static final int DEFAULT_CONTAINER_RETRY_COUNT = 8;
+  public static final int DEFAULT_CONTAINER_RETRY_COUNT = 8;
+
+  public static final String CLUSTER_MANAGER_CONTAINER_RETRY_MAX_DELAY_MS = "cluster-manager.container.host-affinity-retry.max.delay.ms";
+  public static final long DEFAULT_CONTAINER_RETRY_MAX_DELAY_MS = Duration.ofSeconds(120).toMillis();
```
dnishimura: Debating if this should be a non-configurable constant. Thoughts?

samza-core/src/main/java/org/apache/samza/clustermanager/AbstractContainerAllocator.java
```diff
@@ -175,20 +176,23 @@ protected void runStreamProcessor(SamzaResourceRequest request, String preferred
...
   }
 
   /**
    * Retrieves, but does not remove, the next pending request in the queue.
    *
    * @return  the pending request or {@code null} if there is no pending request.
    */
-  protected final SamzaResourceRequest peekPendingRequest() {
-    return resourceRequestState.peekPendingRequest();
+  protected final SamzaResourceRequest peekReadyPendingRequest() {
```
rmatharu: Code nitpick:
Perhaps use Optional<SamzaResourceRequest>

samza-core/src/main/java/org/apache/samza/clustermanager/AbstractContainerAllocator.java
```diff
@@ -198,12 +202,20 @@ protected final SamzaResourceRequest peekPendingRequest() {
...
+    SamzaResourceRequest request = getResourceRequestWithDelay(processorId, preferredHost, delay);
     issueResourceRequest(request);
   }
 
   public final SamzaResourceRequest getResourceRequest(String processorId, String preferredHost) {
-    return new SamzaResourceRequest(this.containerNumCpuCores, this.containerMemoryMb, preferredHost, processorId);
+    return getResourceRequestWithDelay(processorId, preferredHost, Duration.ZERO);
+  }
+
+  public final SamzaResourceRequest getResourceRequestWithDelay(String processorId, String preferredHost, Duration delay) {
```
rmatharu: Code nitpick: 
Maybe add
/**
Create a resource request with a timestamp = current-timestamp + delay. 
*/

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -208,32 +210,45 @@ private static AbstractContainerAllocator buildContainerAllocator(boolean hostAf
...
       throw new SamzaException(exceptionOccurred);
     }
     return tooManyFailedContainers || state.completedProcessors.get() == state.processorCount.get() || !allocatorThread.isAlive();
   }
 
   public void start() {
-    log.info("Starting the container process manager");
+    LOG.info("Starting the container process manager");
+
+    int containerRetryCount = clusterManagerConfig.getContainerRetryCount();
```
rmatharu: Could this entire config handling logic be moved to a method in ClusterManagerConfig?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -524,10 +468,125 @@ public void onStreamProcessorLaunchFailure(SamzaResource resource, Throwable t)
...
   @Override
   public void onError(Throwable e) {
-    log.error("Exception occurred in callbacks in the Cluster Resource Manager", e);
+    LOG.error("Exception occurred in callbacks in the Cluster Resource Manager", e);
     exceptionOccurred = e;
   }
 
+  /**
+   * Called within {@link #onResourceCompleted(SamzaResourceStatus)} for unknown exit statuses. Usually these type of
+   * exit statuses are due to application errors causing the container resource to fail or for other unknown reasons.
```
rmatharu: Nitpick:
These exit statuses correspond to container completion other than container run-to-completion, abort or preemption, or disk failure (e.g., detected by YARN's NM healthchecks).

samza-core/src/main/java/org/apache/samza/clustermanager/AbstractContainerAllocator.java
```diff
@@ -198,12 +202,20 @@ protected final SamzaResourceRequest peekPendingRequest() {
    * @param preferredHost name of the host that you prefer to run the processor on
    */
   public final void requestResource(String processorId, String preferredHost) {
-    SamzaResourceRequest request = getResourceRequest(processorId, preferredHost);
+    requestResourceWithDelay(processorId, preferredHost, Duration.ZERO);
+  }
+
+  public final void requestResourceWithDelay(String processorId, String preferredHost, Duration delay) {
+    SamzaResourceRequest request = getResourceRequestWithDelay(processorId, preferredHost, delay);
```
rmatharu: Adding a delay to the timestamp on the request only delays the "matching/assignment" of the request to a resource. However the request will be sent to the RM immediately, since invoking 
`issueResourceRequest(request) `
will call 
`manager.requestResources(request)`
which will cause the AMRMClient to issue the request to the RM.

Now, at this instant, since the RM has not marked the NM-bearing-failed-container as unhealthy, the request will land at the RM and it may allocate a resource on the same NM again.
Whenenver this request is matched with this "bad" resource, it will fail again, and hopefully the second (or subsequent) retry of this request will be after the RM has marked the NM as unhealthy (due to exponential backoff). 

So either we should make the actual requests to the RM spaced out in time (exponentially), or at the very least document that the "matching" of requests to resource is spaced out exponentially.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -524,10 +468,125 @@ public void onStreamProcessorLaunchFailure(SamzaResource resource, Throwable t)
...
+      Instant lastFailureTime;
+      if (processorFailures.containsKey(processorId)) {
+        ProcessorFailure failure = processorFailures.get(processorId);
+        currentFailCount = failure.getCount() + 1;
+        lastFailureTime = failure.getLastFailure();
+      } else {
+        currentFailCount = 1;
+        lastFailureTime = Instant.now();
+      }
+      if (currentFailCount > maxRetryCount) {
```
rmatharu: currentFailCount >= maxRetryCount

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -524,10 +468,125 @@ public void onStreamProcessorLaunchFailure(SamzaResource resource, Throwable t)
...
+        Duration retryDelay = getHostRetryDelay(lastSeenOn, currentFailCount - 1);
+        long currentRetryWindowMs = Instant.now().toEpochMilli() - lastFailureTime.plus(retryDelay).toEpochMilli();
+
+        if (currentRetryWindowMs < maxRetryWindowMs) {
```
rmatharu: Should be currentRetryWindowMs <= maxRetryWindowMs

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -524,10 +468,125 @@ public void onStreamProcessorLaunchFailure(SamzaResource resource, Throwable t)
...
+      if (processorFailures.containsKey(processorId)) {
+        ProcessorFailure failure = processorFailures.get(processorId);
+        currentFailCount = failure.getCount() + 1;
+        lastFailureTime = failure.getLastFailure();
+      } else {
+        currentFailCount = 1;
+        lastFailureTime = Instant.now();
+      }
+      if (currentFailCount > maxRetryCount) {
+        Duration retryDelay = getHostRetryDelay(lastSeenOn, currentFailCount - 1);
```
rmatharu: For readability, can getHostRetryDelay be made to accept currentFailCount

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -524,10 +468,125 @@ public void onStreamProcessorLaunchFailure(SamzaResource resource, Throwable t)
...
+    }
+
+    long retryDelayMs = 0;
+    long delayMultiplier = Duration.ofSeconds(5).toMillis();
+
+    // add delay only after the first failure
+    if (failCount > 1) {
+      retryDelayMs = delayMultiplier * Math.round(Math.pow(2, failCount - 2));
+    }
+    retryDelayMs = Math.min(retryDelayMs, clusterManagerConfig.getContainerRetryMaxDelayMs());
```
rmatharu: Perhaps consider using org.apache.samza.util.ExponentialSleepStrategy or 
org.apache.commons.math3.distribution. ExponentialDistribution
to avoid dealing with backoff-match logic here.

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
rmatharu: This should be in greater than the max amount of time it takes for the RM to mark a dead NM unhealthy, 
which in YARN 2.9.2 is 5 minutes?
So perhaps not configurable and set to a higher value.

samza-core/src/main/java/org/apache/samza/clustermanager/SamzaResourceRequest.java
```diff
@@ -110,22 +118,31 @@ public String toString() {
...
    * Requests are ordered by the processor type and the time at which they were created.
-   * Active processors take precedence over standby processors, regardless of timestamp.
+   * Requests with timestamps in the future for retries take less precedence than timestamps in the past or current.
+   * Otherwise, active processors take precedence over standby processors, regardless of timestamp.
    * @param o the other
    */
   @Override
   public int compareTo(SamzaResourceRequest o) {
-    if (!StandbyTaskUtil.isStandbyContainer(this.processorId) && StandbyTaskUtil.isStandbyContainer(o.processorId)) {
+    if (!isInFuture() && o.isInFuture()) {
```
rmatharu: Couple of things
a. The isInFuture() check here is redundant since there is the timestamp based check below. 
b. Having this check here, above the check for standby vs. active is problematic. 
Imagine a scenario, where two containers one active (e.g., container 1) and one standby container (e.g., container 2-0) fail concurrently.
If the failure-callback for the standby arrives before the callback of the active, the timestamp for the standby will be less than the timestamp for the active. 
And at some point in time, the standby's request will be not in future, while the one for active will be in future. 
This scenario will violate the priority-queue assumption that active containers take precedence over standby containers.

samza-core/src/main/java/org/apache/samza/clustermanager/AbstractContainerAllocator.java
dnishimura: Ah thanks for the clarification. The logic makes more sense to me now. The documentation implies the requests are enqueued and not executed right away. The explanation of "issuing requests" vs "matching requests" makes the code more clear. I'll update the docs to reflect the true logic and I'll update the logic implementation to handle this.

samza-core/src/main/java/org/apache/samza/clustermanager/AbstractContainerAllocator.java
dnishimura: Will do thanks.

samza-core/src/main/java/org/apache/samza/clustermanager/AbstractContainerAllocator.java
dnishimura: Will separate the the delayed requests from the current `requestsQueue`

samza-core/src/main/java/org/apache/samza/clustermanager/AbstractContainerAllocator.java
dnishimura: Will change. thanks.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: got it

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: Removing the -1

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: got it, makes sense. 
1. 
The code used to have `currentFailCount >= retryCount` and we've changed it to 
`currentFailCount > maxRetryCount` which means without any change to the configparam, 
now the job would fail at 9 failures (1 initial failure + 8 failed retries). Perhaps we can add a lil comment to say that.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -524,10 +468,129 @@ public void onStreamProcessorLaunchFailure(SamzaResource resource, Throwable t)
...
+        ProcessorFailure failure = processorFailures.get(processorId);
+        currentFailCount = failure.getCount() + 1;
+        lastFailureTime = failure.getLastFailure();
+      } else {
+        currentFailCount = 1;
+        lastFailureTime = Instant.now();
+      }
+      if (currentFailCount > maxRetryCount) {
+        Duration retryDelay = getHostRetryDelay(lastSeenOn, currentFailCount - 1);
+        long currentRetryWindowMs = Instant.now().toEpochMilli() - lastFailureTime.plus(retryDelay).toEpochMilli();
```
rmatharu: As we discussed, the problem here is that, because now the lastFailureTime is set to currentTime, as opposed to 0 (in the current code). 
The currentRetryWindowMs value can now be < 0, as opposed to being the currTime value.

This means the check below on line 532 can erroneously pass, and will fail the job. 
In current code, in this case the currentRetryWindowMs (called lastFailureMsDiff) 
gets set = currentTime, which fails the check and falls into the else. 

One way to fix is to set lastFailureTime to 0 (as in current code), or set a flag (firstFailure) and use that to make a determination on the if branch (since there is no failure-window for the first fail).

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: Cool, thanks.

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
rmatharu: Yes maybe add 1 second to account for clock-skew and asynchrony. 
(typically clock skew with a dc is 100s of micros).

dnishimura: Sure will add

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
dnishimura: Not obvious, but the retryDelay is always zero on the first failure, so the scenario you mentioned wouldn't happen. I'll wrap this logic in a private method and document the logic to make it more clear. Thanks.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: Ah i see, 
But is it possible that the retryDelay is > 0 on a subsequent failure, and due to scheduling delays on the machine running this code
`Instant.now().toEpochMilli() - lastFailureTime.plus(retryDelay).toEpochMilli();` becomes < 0 because lastFailureTime is set to Instant.now() above. 
Could this happen even with retryDelay = 0?
Because if it becomes < 0 then the check below will fail.

dnishimura: `lastFailureTime` is only set to `Instant.now()` on the first failure. I'm I overlooking something?

rmatharu: Youre right, I was confusing the current code, which has 
`currentFailCount >= retryCount` with the `currentFailCount > maxRetryCount`
the equality could lead to a erroneous situation but with the change it shouldnt happen

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -366,86 +381,15 @@ public void onResourceCompleted(SamzaResourceStatus resourceStatus) {
...
-          if (processorFailures.containsKey(processorId)) {
-            ProcessorFailure failure = processorFailures.get(processorId);
-            currentFailCount = failure.getCount() + 1;
-            lastFailureTime = failure.getLastFailure();
-          } else {
-            currentFailCount = 1;
-            lastFailureTime = 0L;
-          }
-          if (currentFailCount >= retryCount) {
-            long lastFailureMsDiff = System.currentTimeMillis() - lastFailureTime;
```
rmatharu: Looks like we have a bug where the retry-window doesnt get recomputed upon each failure.

dnishimura: Yup will fix his existing bug in this PR since it's related. Thanks for pointing it out.

dnishimura:
PR on hold due to a bug that @rmatharu and I discovered. Will work on a PR for that bug and merge that in here afterwards once that bug fix is in master. Details of the bug here [SAMZA-2277](https://issues.apache.org/jira/browse/SAMZA-2277)

dnishimura:
@rmatharu please take a look at PR #1108 for the fix with the retry window. I'll merge master here once that PR is committed to master. Thanks.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: Summarizing the discussion, for anyone looking at this review, 
there was a bug where the "fail job if N failures in a M window" was only looking at the "last failure being in the M window"

dnishimura: Fixed in #1108

dnishimura:
@rmatharu please take a look at the latest changes. I merged in master to pull in the changes from PR #1108 that has the fix for the retry window. The rest of the commits where to clean up some things and to address your comments.
I'll be testing on an actual job, so please hold off from merging until I'm done, but feel free to review the code in the meantime. Thanks!

dnishimura:
@rmatharu I tested on an actual job and it seems to work well. Per our discussion, it's no longer doing an exponential backoff, but instead it adds a delay at the last retry for the container restart. The last delayed retry honors the `cluster-manager.container.retry.window.ms` by not including the delay as part of the window. Please take a look when you get a chance.

docs/learn/documentation/versioned/jobs/samza-configurations.md
```diff
@@ -292,6 +292,7 @@ Samza supports both standalone and clustered ([YARN](yarn-jobs.html)) [deploymen
...
 |cluster-manager.container.retry.window.ms|300000|This property determines how frequently a container is allowed to fail before we give up and fail the job. If the same container has failed more than `cluster-manager.container.retry.count` times, and the time between failures was less than this property `cluster-manager.container.retry.window.ms` (in milliseconds), then we fail the job. There is no limit to the number of times we will restart a container if the time between failures is greater than `cluster-manager.container.retry.window.ms`.|
+|cluster-manager.container.preferred-host.last.retry.delay.ms|360000|The delay of the last retry of `cluster-manager.container.retry.count` when `job.host-affinity.enabled` is true and the container is being requested to restart on a preferred host. This addresses the issue where there may be a delay when a preferred host is marked invalid and the container continuously attempts to restart and fail on the invalid preferred host. This property is useful to prevent the `cluster-manager.container.retry.count` from being exceeded too quickly for such scenarios.|
```
rmatharu: Nitpick: Perhaps the first sentence could be

The delay added to the last retry for a failing container after all but one of `cluster-manager.container.retry.count` retries have been exhausted.

dnishimura: Done. I pushed the change. Thanks.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -540,6 +476,114 @@ boolean getJobFailureCriteriaMet() {
...
+      LOG.error("Processor ID: {} (current Container ID: {}) failed, and retry count is set to 0, " +
+          "so shutting down the application master and marking the job as failed.", processorId, containerId);
+
+      jobFailureCriteriaMet = true;
+    } else if (retryCount > 0) {
+      long durationSinceLastRetryMs;
+      if (processorFailures.containsKey(processorId)) {
+        ProcessorFailure failure = processorFailures.get(processorId);
+        currentFailCount = failure.getCount() + 1;
+        Duration lastRetryDelay =
```
rmatharu: Nit: Could this `Duration lastRetryDelay =
            processorFailures.containsKey(processorId)
                ? processorFailures.get(processorId).getLastRetryDelay()
                : Duration.ZERO;` be extracted into a method, since it is used below on line 574

samza-core/src/main/java/org/apache/samza/clustermanager/ProcessorFailure.java
```diff
@@ -19,30 +19,44 @@
...
     this.lastFailure = lastFailure;
+    this.lastRetryDelay = lastRetryDelay;
   }
 
   public int getCount() {
     return count;
   }
 
-  public Long getLastFailure() {
+  public Instant getLastFailure() {
```
rmatharu: +1, thanks for this

samza-core/src/main/java/org/apache/samza/clustermanager/ResourceRequestState.java
```diff
@@ -205,6 +202,22 @@ public void updateStateAfterAssignment(SamzaResourceRequest request, String assi
     }
   }
 
+  /**
+   * Sends the {@link SamzaResourceRequest}s in the delayed requests queue that have expired.
+   * @return number of delayed requests sent.
+   */
+  public int sendExpiredDelayedResourceRequests() {
```
rmatharu: Could we use another name, e.g., `sendPendingDelayedRequests`
In this code "expired resource requests" is also used to refer to resource requests which were never answered with a resource by the RM, so could confuse a new dev looking at this.

samza-core/src/main/java/org/apache/samza/clustermanager/ResourceRequestState.java
```diff
@@ -46,16 +48,24 @@
...
   private final Map<String, List<SamzaResource>> allocatedResources = new HashMap<>();
+
   /**
    * Represents the queue of resource requests made by the {@link ContainerProcessManager}
    */
-  private final PriorityQueue<SamzaResourceRequest> requestsQueue = new PriorityQueue<SamzaResourceRequest>();
+  private final PriorityQueue<SamzaResourceRequest> requestsQueue = new PriorityQueue<>();
+
+  /**
+   * Represents the queue of delayed resource requests made by the {@link ContainerProcessManager}
```
rmatharu: maybe add to documentation: "The difference between requestsQueue and delayedRequestsQueue is that, any request present the requestsQueue has been sent out to the YARN-RM, 
while requests in the delayedRequestsQueue will be sent to the YARN-RM only when their delay reaches 0."

samza-core/src/main/java/org/apache/samza/clustermanager/ResourceRequestState.java
```diff
@@ -213,7 +226,7 @@ public void updateStateAfterAssignment(SamzaResourceRequest request, String assi
   public int releaseExtraResources() {
     synchronized (lock) {
       int numReleasedResources = 0;
-      if (requestsQueue.isEmpty()) {
+      if (requestsQueue.isEmpty() && delayedRequestsQueue.isEmpty()) {
```
rmatharu: We don't need to check the delayedRequestsQueue here, because a resource-request for anything in the delayedRequestsQueue will be sent out only when the requests are no-longer delayed.

The implication of this is that if requestsQueue is empty and delayedRequestsQueue is not, the CPM will continue to hold onto allocated resources, in the worst case for a period of 5 mins (default). 
Aggregated over num_containers, could be significant especially in resource-crunch scenarios.

Alternatively, we could release the resource only looking at the requestsQueue, and when a request in the delayedRequestsQueue "expires", we will send out the request to the YARN-RM and allocation flow shall resume.

dnishimura:
@rmatharu - I addressed your latest comments and the build is green. Please merge after your approve. Thanks for reviewing this PR.

rmatharu:
thanks, will  merge

dnishimura:
> thanks, will merge

Great! Thanks again for reviewing.

PR #1106, opened 2019-07-17
Making Samza-Sql-Shell commands pluggable

shenodaguirguis:
Adding CommandHandler andCommandType interfaces to make samza-sql-shell commands pluggable through conf


samza-sql-shell/conf/shell-defaults.conf
```diff
@@ -22,6 +22,8 @@
 # Example:
 shell.executor = org.apache.samza.sql.client.impl.SamzaExecutor
 samza.sql.output = compact
+# shell.commandhandler = com.linkedin.CustomeCommandHandler1
+# shell.commandhandler = com.linkedin.CustomeCommandHandler2
```
weiqingy: We may prefer a value list to multiple commandhandler entries, as multiple entries could pervade the configuration file and become hard to notice.

shenodaguirguis: made it a comma separated list

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/CommandType.java
```diff
@@ -0,0 +1,36 @@
...
+
+  static List<String> getAllCommands() { return new ArrayList<>(); }
+
+  String getCommandName();
+
+  String getDescription();
+
+  String getUsage();
+
+  boolean isNoArgsCommand();
```
weiqingy: The name of `isNorArgsCommand` could be confusing when a command that takes either parameters or no parameters is added in the future, like the `HELP`. By intuition you never know if that command shall be counted as a no args command or not. A dev may have to check how this flag is used in the code to decide which of `true` and `false` is appropriate.  Maybe a name like `mustHaveArgs`, `isNoArgsOK`, or 'canHaveNoArgs' would be clearer?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/CommandHandler.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.io.PrintWriter;
+import org.apache.samza.sql.client.cli.CliCommand;
+import org.apache.samza.sql.client.cli.CliEnvironment;
+import org.apache.samza.sql.client.cli.CliShell;
+import org.jline.terminal.Terminal;
+
+
+public interface CommandHandler {
+  void initCommandHandler(CliShell shell, CliEnvironment env, PrintWriter printWriter, Terminal terminal,
```
weiqingy: We can get the `printWriter` from `terminal.writer()`, so we can save a parameter here.

Also, consider remove the `shell` parameter. We may not want a `commandhandler` to access the Shell class directly. Encapsulate as much as we can, after all. I don't see code in the `commandhandler` implementation use it, but please correct me if I am mistaken.

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
```diff
@@ -173,11 +187,21 @@ public SqlExecutor getExecutor() {
     return executor;
   }
 
+  public List<CommandHandler> getCommandHandlers() { return commandHandlers; }
+
   private void createShellExecutor(String executorClassName) throws ExecutorException {
+    executor = (SqlExecutor) createInstance(executorClassName);
```
weiqingy: I just noticed the original code didn't handle the `ClassCastException`. Shell would crash if a user flirts with the configuration file, setting `executor` or `commandhandler` to some random class. Could you help fix it here?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/CommandType.java
shenodaguirguis: renamed to argsAreOptional (both member var and fun)

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
```diff
@@ -173,11 +187,21 @@ public SqlExecutor getExecutor() {
...
+  }
+
+  private void createCommandHandler(String handlerClassName) throws ExecutorException {
+    commandHandlers.add((CommandHandler) createInstance(handlerClassName));
```
weiqingy: Same as the above.

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
```diff
@@ -173,11 +187,21 @@ public SqlExecutor getExecutor() {
...
+  }
+
+  private Object createInstance(String className) throws ExecutorException {
```
weiqingy: It's fine to throw a `ExecutorException` for a method of `createShellExecutor`, but would be misleading for a method named `createInstance` as it creates `commandhandler`s as well.

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
```diff
@@ -173,11 +187,21 @@ public SqlExecutor getExecutor() {
     return executor;
   }
 
+  public List<CommandHandler> getCommandHandlers() { return commandHandlers; }
+
   private void createShellExecutor(String executorClassName) throws ExecutorException {
+    executor = (SqlExecutor) createInstance(executorClassName);
+  }
+
+  private void createCommandHandler(String handlerClassName) throws ExecutorException {
```
weiqingy: Throwing a `ExecutorException` here may not be appropriate.

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/CommandHandler.java
shenodaguirguis: removed printWriter, but need shell (used in commandSelect > view.open(shell, queryResult) at line 437), also encapsulation is preserved here since we only pass the shell to CliView...

weiqingy: Yeah... Thanks. I'll clean up CliView later. CliView actually only uses the terminal, executor and context.

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
shenodaguirguis: done

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
shenodaguirguis: done

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/CommandHandler.java
weiqingy: Just one trivia: `init` shall be sufficient, instead of `initCommandHandler`?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
shenodaguirguis: makes sense, I added CommandHandlerException class, CommandHandler now throws it, and createInstance here throws Exception, so that the createShellExecutor throws ExecutorException while createCommandHandler throws CommandHandlerException

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliEnvironment.java
shenodaguirguis: changed it to throw CommandHandlerException

samza-sql-shell/src/main/java/org/apache/samza/sql/client/interfaces/CommandHandler.java
shenodaguirguis: done

weiqingy:
Issues we have left untouched but maybe as a future work:

The environment variables (operated by the `SET` command). Shall a `commandhandler` have its own environment variables? Shall their environment variables start with `shell.`? At the moment we treat all environment variables starting with `shell.` as the Shell's, and pass everything else to the executor. The logic may no longer work. We may need to distinguish them, like `shell.` , `executor.`, `commandhandler1.`, `commandhandler2.`, etc.

The SET command and multiple `commandhandler`s. Just like we allow user to change the `executor` via the `SET` command, we allow user to change `commandhandler`s via the `SET` command. However, with multiple `commandhandler`s allowed, it's a bit tricky. Say a user has loaded `commandhandler1` and `commandhandler2`. He or she now wants to use `commandhandler3` as well. We wouldn't want them to issue a command like:
`SET commandhandler = commandhandler1, commandhandler2, commandhandler3`
What's in my mind is new commands `load` and `unload`. User may issue commands like:
```
load commandhandler1
load commandhandler2
```
We may also want to consider distinguish commands of different commandhandlers by adding a prefix on the command, like:
```
commandhandler1.commandA
commandhandler2.commandA
```
Let me know what you think.

shenodaguirguis:
my 2-mites:
+ the load/unload commands makes perfect sense, can be a future work
+ I am not a big fan of prefixing commands with handlers, not very user friendly? ... besides, with load/unload this won't be needed, but I don't anticipate users shuffling command handlers much, this feature is mainly to allow each user to plug-in their favorite handler.
+ We can discuss offline the env vars scope, I don't see them applicable per handler...

weiqingy:
Yeah. We'll see what we need to do according to the user feedback.

atoomula:
> my 2-mites:
> 
> * the load/unload commands makes perfect sense, can be a future work
> * I am not a big fan of prefixing commands with handlers, not very user friendly? ... besides, with load/unload this won't be needed, but I don't anticipate users shuffling command handlers much, this feature is mainly to allow each user to plug-in their favorite handler.
> * We can discuss offline the env vars scope, I don't see them applicable per handler...

atoomula:
> my 2-mites:
> 
> * the load/unload commands makes perfect sense, can be a future work
> * I am not a big fan of prefixing commands with handlers, not very user friendly? ... besides, with load/unload this won't be needed, but I don't anticipate users shuffling command handlers much, this feature is mainly to allow each user to plug-in their favorite handler.
> * We can discuss offline the env vars scope, I don't see them applicable per handler...

I think we need to differentiate between what environment variables users can load/set dynamically and what can be loaded at shell startup time. Command handler is one which sits in a bucket where it can be loaded only at shell startup time. It makes code also much simpler. I'm thinking about prefixing/name-spacing such variables so that it is easier to identify such env vars. It need not go as part of this PR but what do you guys think ?

weiqingy:
I agree.

IMO command handers should NOT be environment variables. They are EXTENSIONs. They shall NOT be loaded by SET command. I prefer a load command instead.

weiqingy:
@shenodaguirguis Forgot to remind of the Java document on interfaces and public methods. There'll be lots of build warnings otherwise.

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliShell.java
```diff
@@ -265,564 +215,40 @@ private void printVersion() {
...
-
-    // Footer
-    line = new StringBuilder(terminalWidth);
-    CliUtil.appendTo(line, longestLineCharNum - 1, LINE_SEP);
-    lines.add(line.toString());
-    return lines;
+    return new CommandAndHandler(parsedCommand, commandHandler);
   }
 
-  private String getFieldDisplayValue(SqlFieldSchema fieldSchema) {
```
atoomula: Shenoda - where did these changes go ? did we just delete them ?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliShell.java
```diff
@@ -265,564 +215,40 @@ private void printVersion() {
...
-        return String.format("ROW(%s)", rowDisplayValue);
-      default:
-        throw new UnsupportedOperationException("Unknown field type " + fieldType);
-    }
-  }
+  class CommandAndHandler {
+    CliCommand command;
+    CommandHandler handler;
 
-  private boolean isComplexField(SamzaSqlFieldType fieldtype) {
```
atoomula: How about this ?

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliShell.java
shenodaguirguis: looks like lost update issue, created PR to restore them: https://github.com/apache/samza/pull/1114

samza-sql-shell/src/main/java/org/apache/samza/sql/client/cli/CliShell.java
shenodaguirguis: same

PR #1108, opened 2019-07-19
SAMZA-2277: Semantics for cluster-manager.container.retry.window.ms not reflected in code

dnishimura:
In the current code, the window is only applied and checked on the last retry. However, the check should be done at all retries.

This was found during #1104 

@rmatharu please take a look

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
+++ b/samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
@@ -106,25 +106,25 @@ public class ContainerProcessManager implements ClusterResourceManager.Callback 
   private final ClusterResourceManager clusterResourceManager;
 
   /**
+   * If there are too many failed container failures (configured by job.container.retry.count) for a
+   * processor, the job exits.
+   */
+  private volatile boolean tooManyFailedContainers = false;
+
+  /**
    * Exception thrown in callbacks, such as {@code containerAllocator}
    */
   private volatile Throwable exceptionOccurred = null;
-
-  private ContainerProcessManagerMetrics containerProcessManagerMetrics;
-  private JvmMetrics jvmMetrics;
-  private Map<String, MetricsReporter> metricsReporters;
 
   /**
    * A map that keeps track of how many times each processor failed. The key is the processor ID, and the
    * value is the {@link ProcessorFailure} object that has a count of failures.
    */
-  final Map<String, ProcessorFailure> processorFailures = new HashMap<>();
+  private final Map<String, ProcessorFailure> processorFailures = new HashMap<>();
 
-  /**
-   * If there are too many failed container failures (configured by job.container.retry.count) for a
-   * processor, the job exits.
-   */
-  volatile boolean tooManyFailedContainers = false;
+  private ContainerProcessManagerMetrics containerProcessManagerMetrics;
+  private JvmMetrics jvmMetrics;
+  private Map<String, MetricsReporter> metricsReporters;
 
   public ContainerProcessManager(Config config, SamzaApplicationState state, MetricsRegistryMap registry,
       ClassLoader classLoader) {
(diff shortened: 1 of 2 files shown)


samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -104,25 +105,26 @@
...
-  private ContainerProcessManagerMetrics containerProcessManagerMetrics;
-  private JvmMetrics jvmMetrics;
-  private Map<String, MetricsReporter> metricsReporters;
+  final Map<String, ProcessorFailure> processorFailures = new HashMap<>();
+
+  /**
+   * If there are too many failed container failures (configured by job.container.retry.count) for a
+   * processor, the job exits.
+   */
+  volatile boolean tooManyFailedContainers = false;
```
rmatharu: Can we expose a getter instead for testing?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -409,32 +411,34 @@ public void onResourceCompleted(SamzaResourceStatus resourceStatus) {
...
-            } else {
-              log.info("Resetting failure count for Processor ID: {} back to 1, since last failure " +
-                  "(for Container ID: {}) was outside the bounds of the retry window.", processorId, containerId);
-
-              // Reset counter back to 1, since the last failure for this
-              // container happened outside the window boundary.
-              processorFailures.put(processorId, new ProcessorFailure(1, System.currentTimeMillis()));
-            }
+
+          long lastFailureMsDiff = Instant.now().toEpochMilli() - lastFailureTime;
```
rmatharu: Minor: Might be preferable to set lastFailureMsDiff in the if-condition above, to avoid time-skew in measurements.

dnishimura: I imagine it would be a sub-millisecond skew? Would that matter? Will change anyways to make the code cleaner.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
dnishimura: Will change.

dnishimura:
@rmatharu I addressed your minor comments. Please merge if you don't have any more comments. Thanks!

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: Typically, yes. In some cases, CPU preemption can cause the time to be relatively large, i've only seen it once when the host was so busy it was spending most of the time context-switching

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: Should we update the comment to "If there are more than job.container.retry.count failures of a container within a job.container.retry.window period, the CPM exits"?
and update the flag to jobFailureCriteriaMet.

samza-core/src/test/java/org/apache/samza/clustermanager/TestContainerProcessManager.java
```diff
@@ -338,6 +349,145 @@ public void testNewContainerRequestedOnFailureWithUnknownCode() throws Exception
...
+    SamzaResource container = new SamzaResource(1, 1024, "host1", "id0");
+    cpm.onResourceAllocated(container);
+
+    // Allow container to run and update state
+    if (!allocator.awaitContainersStart(1, 2, TimeUnit.SECONDS)) {
+      fail("timed out waiting for the containers to start");
+    }
+    cpm.onStreamProcessorLaunchSuccess(container);
+
+    // Mock 2nd failure exceeding retry window.
```
rmatharu: Nitpick: 
"1st failure" ?

samza-core/src/test/java/org/apache/samza/clustermanager/TestContainerProcessManager.java
```diff
@@ -338,6 +349,145 @@ public void testNewContainerRequestedOnFailureWithUnknownCode() throws Exception
     taskManager.stop();
   }
 
+  @Test
```
rmatharu: /** Test scenario where a container fails multiple times but failures are more than retryWindow apart.*/

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
dnishimura: Good suggestion, it makes it clearer. Will change.

dnishimura:
@rmatharu I made your suggested changes. I'm not a committer so I can't check in the changes. Please merge this PR when you get a chance. Thanks!

PR #1110, opened 2019-07-22
SAMZA-2278: Delete all for startpoints and fan outs.

dnishimura:
Added a "delete all" for startpoints so we have the ability to delete all in case we run into situations where the startpoints poses a problem for the job.


dnishimura:
Thanks please merge.

PR #1112, opened 2019-07-22
SAMZA-2276: Add Metdata store putAll API

dnishimura:
Add a putsAll API to the metadata store API. This allows for a delayed flush() call for underlying stores that require a flush() after write such as Kafka.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
+++ b/samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
@@ -50,7 +50,7 @@ public interface MetadataStore {
   void put(String key, byte[] value);
 
   /**
-   * Updates the mapping with the specified map.
+   * Updates the mapping with the specified map. This is not guaranteed to be atomic.
    *
    * @param entries mapping of key to values to write to the metadata store
    */
--- a/samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
+++ b/samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
@@ -18,8 +18,8 @@
  */
 package org.apache.samza.metadatastore;
 
-import org.apache.samza.annotation.InterfaceStability;
 import java.util.Map;
+import org.apache.samza.annotation.InterfaceStability;
 
 /**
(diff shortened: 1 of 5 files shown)


dnishimura:
> A few questions to understand the motivation of `PutAll`
> 
> 1. Does `Put` have to be synchronous? Since `flush()` is exposed as part of API, what do you think about callers handling it in two steps? Have we evaluated this?
> 2. What is the use of `flush()`? Looking at the implementation of `Put` and `PutAll`, both are synchronous. Do we intend to have different behavior for future MetadataStore implementation? If so, I suppose we should follow up this PR with another PR that details the guarantees and expectations of these APIs.

I believe the original intent of the metadata store is to have consistent behavior regardless of the underlying store. Since it's simpler to make async calls synchronous than vice-versa, the expected behavior is synchronous. The intent of the `putAll` API is to provide an optimization option at least for the short-term until the next-gen metadata store abstraction is designed. That being said, I think flush() should not be exposed as an external API.
All good points you have brought up that we can take into account for the next phase of the metadata store design.

dnishimura:
@mynameborat if you don't have any more comments, please approve and merge this commit. This is required to address an internal performance regression with job startup time. Thanks!
CC: @shanthoosh @prateekm

samza-api/src/main/java/org/apache/samza/metadatastore/MetadataStore.java
```diff
@@ -49,6 +49,17 @@
    */
   void put(String key, byte[] value);
 
+  /**
+   * Updates the mapping with the specified map.
+   *
+   * @param entries mapping of key to values to write to the metadata store
+   */
+  default void putAll(Map<String, byte[]> entries) {
```
mynameborat: Can you please add documentation around the new API? Especially around no guarantees on atomicity of the writes. If you'd like to address the documentation as part of a follow PR for the all the APIs, that is fine by me too.

dnishimura: No problem, will add.

dnishimura:
@mynameborat thanks for the review. I addressed your comment. This is ready to be merged.

PR #1113, opened 2019-07-24
SAMZA-2280: LocalApplicationRunner should use final, planned set of configs

bkonold:
LocalApplicationRunner should use final planned set of configs when creating stream processors. Prior to this change, LAR was relying on unplanned configs from the ApplicationDescriptor provided at construction time - this set of configs lacked system generated values coming from, for example, input and output descriptors defined in Application.describe()


bkonold:
Discussed offline. Though there is effectively no change in control flow, this PR brings the standalone case slightly closer to parity with YARN in that the ApplicationDescriptor will be constructed using the planned set of configs.

I've additionally (in the latest commit) removed some redundancy between config parameters and application descriptor, to make it more clear to the reader what configs are being used throughout.

bkonold:
Good point. In YARN we already do this, since the app descriptor is instantiated once during planning and again on container start up. So, at least there would be parity between standalone and YARN control flows. There are certainly cases where weird things could happen dependent on the idempotency of the user's code. Whether this breaks anything I think would be a question of what users are already doing in their describe methods.

bkonold:
Closing this PR as we decided to take a different approach in the short term. Some open questions need to be resolved before we bring this change into open source. Please refer to SAMZA-2312

PR #1115, opened 2019-07-25
SAMZA-2281: Make sql udfs case insensitive

atoomula:


shanthoosh:
@atoomula Do we still need this patch?

PR #1117, opened 2019-07-26
[SAMZA-2283] Samza Sql Diagnostics: use Nanosecond for logical operator processing time and query latency

shenodaguirguis:
1) Use nanoTime for SamzaSql specific latency metrics (i.e., logical operator processing time, and query latency)
2) remove some unused imports

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputTransformer.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputTransformer.java
@@ -19,7 +19,6 @@
 
 package org.apache.samza.sql;
 
-import java.time.Instant;
 import net.jodah.failsafe.internal.util.Assert;
 import org.apache.samza.operators.KV;
 import org.apache.samza.sql.data.SamzaSqlRelMsgMetadata;
@@ -40,9 +39,7 @@ public class SamzaSqlInputTransformer implements InputTransformer {
   public Object apply(IncomingMessageEnvelope ime) {
     Assert.notNull(ime, "ime is null");
     KV<Object, Object> keyAndMessageKV = KV.of(ime.getKey(), ime.getMessage());
-    SamzaSqlRelMsgMetadata metadata = new SamzaSqlRelMsgMetadata(
-        (ime.getEventTime() == 0) ? "" : Instant.ofEpochMilli(ime.getEventTime()).toString(),
-        (ime.getArrivalTime() == 0) ? "" : Instant.ofEpochMilli(ime.getArrivalTime()).toString(), null);
+    SamzaSqlRelMsgMetadata metadata = new SamzaSqlRelMsgMetadata(ime.getEventTime(), ime.getArrivalTime());
     SamzaSqlInputMessage samzaMsg = SamzaSqlInputMessage.of(keyAndMessageKV, metadata);
     return  samzaMsg;
   }
--- a/samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
@@ -20,7 +20,6 @@
 package org.apache.samza.sql.data;
 
 import java.io.Serializable;
-import java.time.Instant;
 import org.codehaus.jackson.annotate.JsonIgnore;
 import org.codehaus.jackson.annotate.JsonProperty;
 
@@ -56,60 +55,82 @@ public class SamzaSqlRelMsgMetadata implements Serializable {
    * TODO: copy eventTime through from source to RelMessage
    */
   @JsonProperty("eventTime")
-  private String eventTime;
+  private long eventTime = 0L;
 
   /**
    * the timestamp of when Samza App received the event
    * TODO: set arrivalTime during conversion from IME to SamzaMessage
    */
   @JsonProperty("arrivalTime")
-  private String arrivalTime;
+  private long arrivalTime = 0L;
 
   /**
-   * the timestamp when SamzaSQL query starts processing the event
-   * set by the SamzaSQL Scan operator
+   * the System.nanoTime when SamzaSQL query starts processing the event
+   * set by the SamzaSQL Scan operator, used by QueryTranslator to calculate
+   * the Query Latency
+   *
+   * Note: using JsonPeoperty("scanTim") for backward compatibility
    */
   @JsonProperty("scanTime")
-  private String scanTime;
+  private long scanTimeNanos = 0L;
 
-  public SamzaSqlRelMsgMetadata(@JsonProperty("eventTime") String eventTime, @JsonProperty("arrivalTime") String arrivalTime,
-      @JsonProperty("scanTime") String scanTime) {
+  /**
+   * the tiemstamp when SamzSQL wuery starts processing the event
+   * set by the SamzaSQL Scan oeprator, used by QueryTranslator to calculate
+   * the Queuing Latency
+   */
+  @JsonProperty("scanTimeMillis")
+  private long scanTimeMillis = 0L;
+
+  public SamzaSqlRelMsgMetadata(@JsonProperty("eventTime") long eventTime, @JsonProperty("arrivalTime") long arrivalTime,
+      @JsonProperty("scanTime") long scanTimeNanos, @JsonProperty("scanTimeMillis") long scanTimeMillis) {
     this.eventTime = eventTime;
     this.arrivalTime = arrivalTime;
-    this.scanTime = scanTime;
+    this.scanTimeNanos = scanTimeNanos;
+    this.scanTimeMillis = scanTimeMillis;
   }
 
-  public SamzaSqlRelMsgMetadata(String eventTime, String arrivalTime, String scanTime, boolean isNewInputMessage) {
-    this(eventTime, arrivalTime, scanTime);
+  public SamzaSqlRelMsgMetadata(long eventTime, long arrivalTime, long scanTimeNanos, long scanTimeMillis,
+      boolean isNewInputMessage) {
+    this(eventTime, arrivalTime, scanTimeNanos, scanTimeMillis);
     this.isNewInputMessage = isNewInputMessage;
   }
 
-  @JsonProperty("eventTime")
-  public String getEventTime() { return eventTime;}
+  public SamzaSqlRelMsgMetadata(long eventTime, long arrivalTime) {
+    this(eventTime, arrivalTime, 0L, 0L);
+  }
 
-  public void setEventTime(String eventTime) {
+
+      @JsonProperty("eventTime")
+  public long getEventTime() { return eventTime;}
+
+  public void setEventTime(long eventTime) {
     this.eventTime = eventTime;
   }
 
-  public boolean hasEventTime() { return eventTime != null && !eventTime.isEmpty(); }
+  public boolean hasEventTime() { return eventTime != 0L; }
 
   @JsonProperty("arrivalTime")
-  public String getarrivalTime() { return arrivalTime;}
+  public long getArrivalTime() { return arrivalTime;}
 
-  public void setArrivalTime(String arrivalTime) {
+  public void setArrivalTime(long arrivalTime) {
     this.arrivalTime = arrivalTime;
   }
 
-  public boolean hasArrivalTime() { return arrivalTime != null && !arrivalTime.isEmpty(); }
+  public boolean hasArrivalTime() { return arrivalTime != 0L; }
 
   @JsonProperty("scanTime")
-  public String getscanTime() { return scanTime;}
+  public long getScanTimeNanos() { return scanTimeNanos;}
 
-  public void setScanTime(String scanTime) {
-    this.scanTime = scanTime;
+  @JsonProperty("scanTimeMillis")
+  public long getScanTimeMillis() { return scanTimeMillis;}
+
+  public void setScanTime(long scanTimeNano, long scanTimeMillis) {
+    this.scanTimeNanos = scanTimeNano;
+    this.scanTimeMillis = scanTimeMillis;
   }
 
-  public boolean hasScanTime() { return scanTime != null && !scanTime.isEmpty(); }
+  public boolean hasScanTime() { return scanTimeNanos != 0L; }
 
   @JsonIgnore
   public  void setIsSystemMessage(boolean isSystemMessage) {
@@ -123,7 +144,7 @@ public class SamzaSqlRelMsgMetadata implements Serializable {
 
   @Override
   public String toString() {
-    return "[Metadata:{" + eventTime + " " + arrivalTime + " " + scanTime + "}]";
+    return "[Metadata:{" + eventTime + " " + arrivalTime + " " + scanTimeNanos + " " + scanTimeMillis + "}]";
   }
 
 }
--- a/samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
@@ -20,8 +20,6 @@
 package org.apache.samza.sql.translator;
 
 import com.google.common.annotations.VisibleForTesting;
-import java.time.Duration;
-import java.time.Instant;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
@@ -101,9 +99,9 @@ public class QueryTranslator {
     /* insert (SendToOutputStream) metrics */
     private transient SamzaHistogram insertProcessingTime;
     /* query metrics */
-    private transient SamzaHistogram totalLatency; // (if event time exists) = output time - event time (msec)
-    private transient SamzaHistogram queryLatency; // = output time - scan time (msec)
-    private transient SamzaHistogram queueingLatency; // = scan time - arrival time (msec)
+    private transient SamzaHistogram totalLatencyMs; // (if event time exists) = output time - event time (msec)
+    private transient SamzaHistogram queryLatencyNs; // = output time - scan time (msec)
+    private transient SamzaHistogram queuingLatencyMS; // = scan time - arrival time (msec)
     private transient Counter queryOutputEvents;
 
     private final String outputTopic;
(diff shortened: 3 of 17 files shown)


samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
```diff
@@ -71,7 +70,7 @@ void put(SystemStreamPartition ssp, Object key, Object message) {
     }
 
     IncomingMessageEnvelope messageEnvelope = new IncomingMessageEnvelope(ssp, offset, key, message,
-        0, 0L, Instant.now().toEpochMilli());
+        0, 0L, System.nanoTime());
```
prateekm: nanoTime != epoch millis in nanos.

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -92,21 +92,55 @@ public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, Stri
   }
 
   /**
-   * Getter for event time
+   * Getter for event time (Nanos)
    * @return this.eventTime
    */
-  public long getEventTime() {
+  public long getEventTimeNanos() {
```
prateekm: Have you seen scenarios where event time is in nanos? It's not common to have nanosecond resolution clocks.

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -92,21 +92,55 @@ public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, Stri
...
+   */
+  public long getEventTimeMillis() {
+    return eventTime / MILLIS_IN_NANO;
+  }
+
+  /**
+   * Getter for arrival time in Millis
+   * @return this.arrivalTime / 10^6
+   */
+  public long getArrivalTimeMillis() {
```
prateekm: I'd like to understand what we're trying to achieve here. Backwards compatibility is fine, but now there are 6 time related methods in this class which is not helpful.

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -81,8 +81,8 @@ public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, Stri
    * @param key A deserialized key received from the partition offset.
    * @param message A deserialized message received from the partition offset.
    * @param size size of the message and key in bytes.
-   * @param eventTime the timestamp (in epochMillis) of when this event happened
-   * @param arrivalTime the timestamp (in epochMillis) of when this event arrived to (i.e., was picked-up by) Samza
+   * @param eventTime the timestamp (in nanoTime) of when this event happened
+   * @param arrivalTime the timestamp (in nanoTime) of when this event arrived to (i.e., was picked-up by) Samza
    */
   public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, String offset,
```
cameronlee314: The constructor is still part of the public API, so it probably is best to add a new constructor as well for specifying nanos. If existing code is building an `IncomingMessageEnvelope` using this constructor, and then calling `getEventTime` and `getArrivalTime`, then they will be getting a value which is too low, since you will divide these values by 1000000.
Unfortunately, the same constructor signature is desired for both millis and nanos, so maybe you'll need a static method to build the envelope instead of using an actual constructor.
There may be places outside of Samza which are building `IncomingMessageEnvelope`s (e.g. Brooklin), so if you just change this constructor, then you will need to be very sure that you catch all of the existing usages of this constructor and update them. If you add a new syntax for building `IncomingMessageEnvelope`s for nano times, then existing usages will continue to work as before, and then you can update them to just pick up the more granular functionality.

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -92,21 +92,55 @@ public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, Stri
...
+  public long getEventTime() {
+    return getEventTimeMillis();
+  }
+
+  /**
+   * Getter for event time in Millis
+   * @return this.eventTime / 10^6
+   */
+  public long getEventTimeMillis() {
+    return eventTime / MILLIS_IN_NANO;
```
cameronlee314: Please use a time util for this and below: e.g. `java.util.concurrent.TimeUnit.NANOS.convert(millis, TimeUnit.MILLIS)`

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
shenodaguirguis: essentially we need to capture latencies in nano or micro seconds resolution, because millis report zero. On local machine testing, 0's for queryLatency becomes non-zero when we use nano. So that is why we switched to nanoTime. To make this change backward compatible, we needed the old get and the new getNano... the getMillis are the natural counterparts for getNano, and the old get uses them...

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
shenodaguirguis: it doesn't matter since what we want to achieve at the end is to calculate the duration (latency) so we will compare 2 nanoTime points. The only case where this matter is when we read the eventTime from the kafka record timestamp, which is EpochMilli ... any suggestions how to handle this?

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
shenodaguirguis: we need to use higher than Millis resolution to compute query latency in our case. 
One idea is to use Millis for eventTime, and keep its semantics, but use nanoArrivalTime so that we can calculate accurate queryLatency.

shenodaguirguis: will restore EpochMillis for eventTime, but will have arrivalSystemNanoTime to replace arrivalTime and it will be nanoTime

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
shenodaguirguis: this is no longer applicable as per our discussion

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
shenodaguirguis: agreed, and resolved as discussed offline by using Instant for eventTime and keep original semantics (use EpochMilli), while changing this constructor signature to take Instant for eventTime (this will allow to capture any missed usage by failing at compile time)

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
shenodaguirguis: this is resolved now, we have only 1 new getter (3 total)

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
shenodaguirguis: done

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
shenodaguirguis: this is resolved by reverting eventTime to its original semantics (i.e., epochMilli) and use nanoTime for arrivalTime (now renamed arrivalSystemNanoTime to be clear, and added javadoc)

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -37,10 +38,20 @@
...
   private final Object message;
   private final int size;
-  // the System's nanoTime when this event occurred, should be set by the event source, 0 means unassigned
-  private long eventTime = 0L;
-  // the System's nanoTime when this event is picked-up by samza, 0 means unassgined
-  private long arrivalTime;
+
+  /**
+   * the Instant when this event occurred, should be set by the event source,
+   * default null if not available or unassigned
```
cameronlee314: Instead of needing to deal with `null`, maybe we can just default to `Instant.ofEpochMilli(0)`? That is what is treated as the default anyways for `getEventTime`.

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -81,30 +92,41 @@ public IncomingMessageEnvelope(SystemStreamPartition systemStreamPartition, Stri
...
-    this.arrivalTime = arrivalTime;
+    this.eventTimeInstant = eventTimeInstant;
+    this.arrivalSystemNanoTime = arrivalSystemNanoTime;
+  }
+
+  /**
+   * Getter for eventTimeInstant
+   * @return Instant
+   */
+  public Instant getEventTimeInstant() {
```
cameronlee314: Do you need another getter for this?
Maybe you could recreate the `Instant` using `Instant.ofEpochMilli`, or mark `getEventTime` as deprecated to be replaced by `getEventTimeInstant`.

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -194,7 +216,7 @@ public boolean equals(Object obj) {
   @Override
   public String toString() {
     return "IncomingMessageEnvelope [systemStreamPartition=" + systemStreamPartition + ", offset=" + offset +
-        ", key=" + key + ", message=" + message + ", eventTime=" + eventTime +
-        ", arrivalTime=" + arrivalTime + "]";
+        ", key=" + key + ", message=" + message + ", eventTime=" + eventTimeInstant.toEpochMilli() +
```
cameronlee314: This will throw an NPE if `eventTimeInstant` is null.

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
```diff
@@ -194,7 +216,7 @@ public boolean equals(Object obj) {
   @Override
   public String toString() {
```
cameronlee314: Should `equals` and `hashCode` include the `eventTime` and `arrivalSystemNanoTime`? It seems like if two events with the same content came in at different times, then they might still be considered as different events.

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputTransformer.java
```diff
@@ -40,7 +39,8 @@
   public Object apply(IncomingMessageEnvelope ime) {
     Assert.notNull(ime, "ime is null");
     KV<Object, Object> keyAndMessageKV = KV.of(ime.getKey(), ime.getMessage());
-    SamzaSqlRelMsgMetadata metadata = new SamzaSqlRelMsgMetadata(ime.getEventTime(), ime.getArrivalTime(),0L);
+    SamzaSqlRelMsgMetadata metadata =
+        new SamzaSqlRelMsgMetadata(ime.getEventTime(), ime.getArrivalSystemNanoTime(),0L);
```
cameronlee314: Minor: Please add whitespace before the final argument.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
```diff
@@ -51,36 +50,37 @@
...
    */
   @JsonProperty("eventTime")
   private long eventTime;
 
   /**
    * the timestamp of when Samza App received the event
-   * TODO: set arrivalTime during conversion from IME to SamzaMessage
+   * TODO: set arrivalSystemNanoTime during conversion from IME to SamzaMessage
    */
-  @JsonProperty("arrivalTime")
```
cameronlee314: Does this JSON content get persisted anywhere across deployments? If so, then an upgraded system will not be able to read old data, since you renamed this JSON property.

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
```diff
@@ -165,19 +167,19 @@ private void updateMetrics(long beginProcessing, long endProcessing, SamzaSqlRel
...
-        long arrivalTime = metadata.getarrivalTime();
+      if (metadata.hasArrivalSystemNanoTime()) {
+        long arrivalTime = metadata.getArrivalSystemNanoTime();
         queueingLatency.update(TranslatorUtils.getMicroDurationFromNanoTimes(arrivalTime, scanTime));
       }
       /* since availability of eventTime depends on source, we need the following check */
       if (metadata.hasEventTime()) {
         long eventTime = metadata.getEventTime();
-        totalLatency.update(TranslatorUtils.getMicroDurationFromNanoTimes(eventTime, outputTime));
+        totalLatency.update(Duration.between(Instant.ofEpochMilli(eventTime), Instant.now()).toMillis());
```
cameronlee314: It looks like this was in micros before, and now you are changing to millis. Looks like this is intentional, but wanted to double check.

samza-sql/src/test/java/org/apache/samza/sql/serializers/TestSamzaSqlRelMessageSerde.java
```diff
@@ -38,12 +37,11 @@
 import org.apache.samza.sql.avro.schemas.StreetNumRecord;
 import org.apache.samza.sql.data.SamzaSqlRelMessage;
 import org.apache.samza.sql.data.SamzaSqlRelMsgMetadata;
-import org.apache.samza.sql.serializers.SamzaSqlRelMessageSerdeFactory;
 import org.apache.samza.system.SystemStream;
 import org.junit.Assert;
 import org.junit.Test;
 
-import static org.apache.samza.sql.serializers.SamzaSqlRelMessageSerdeFactory.SamzaSqlRelMessageSerde;
```
cameronlee314: We try to generally not use wildcards for static imports. Please revert back to the previous import statement.

samza-sql/src/test/java/org/apache/samza/sql/serializers/TestSamzaSqlRelRecordSerde.java
```diff
@@ -40,7 +38,7 @@
 import org.junit.Assert;
 import org.junit.Test;
 
-import static org.apache.samza.sql.serializers.SamzaSqlRelRecordSerdeFactory.SamzaSqlRelRecordSerde;
```
cameronlee314: Same about wildcard import

samza-sql/src/test/java/org/apache/samza/sql/translator/TestProjectTranslator.java
```diff
@@ -52,19 +46,14 @@
...
+
 import static org.junit.Assert.*;
-import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.eq;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Matchers.*;
```
cameronlee314: same about wildcard import

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
shenodaguirguis: makes sense. done.

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
shenodaguirguis: marked getEventTime as Deprecated

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
shenodaguirguis: different yes, but still equal, right? 
Given that eventTime which pre-existed before adding arrivalTime was not included in either `equals` or `hashcode`, if we include `eventTime` or `arrivalTime` that won't be a backward compatible change

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
shenodaguirguis: defaulted eventTimeInstant to epochMilli(0)

samza-sql/src/main/java/org/apache/samza/sql/SamzaSqlInputTransformer.java
shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/QueryTranslator.java
shenodaguirguis: yes, since we decided to revert back eventTime to EpochMilli

samza-sql/src/test/java/org/apache/samza/sql/serializers/TestSamzaSqlRelMessageSerde.java
shenodaguirguis: thanks! done!

samza-sql/src/test/java/org/apache/samza/sql/serializers/TestSamzaSqlRelRecordSerde.java
shenodaguirguis: done

samza-sql/src/test/java/org/apache/samza/sql/translator/TestProjectTranslator.java
shenodaguirguis: done

samza-api/src/main/java/org/apache/samza/system/IncomingMessageEnvelope.java
cameronlee314: I wasn't sure about what semantics we wanted. I think there is a reasonable argument either way. But you make a good point that `eventTime` was already not included, so `arrivalTime` can be not included as well.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
shenodaguirguis: I checked some Jackson tutorials and checked some code and it seems that Jackson's ObjectMapper uses org.codehaus.jackson.map, which doesn't seem to be persisted.. in the coming up update, arrivalTime won't be renamed, but scanTime is, however, I am using same JsonProperty value for backward compatibility

samza-sql/src/main/java/org/apache/samza/sql/translator/TranslatorConstants.java
```diff
@@ -20,10 +20,10 @@
 package org.apache.samza.sql.translator;
 
 public class TranslatorConstants {
-  public static final String PROCESSING_TIME_NAME = "processingTimeMs";
-  public static final String TOTAL_LATENCY_NAME = "totalLatencyMs";
-  public static final String QUERY_LATENCY_NAME = "queryLatencyMs";
-  public static final String QUEUEING_LATENCY_NAME = "queueingLatencyMs";
+  public static final String PROCESSING_TIME_NAME = "processingTimeMicros";
```
prateekm: Why is this micros? Same for the PR title.

If this timer is indeed in micros, I'd recommend making it nanos instead for consistency. Timers are usually assumed to be millis or nanos.

samza-sql/src/main/java/org/apache/samza/sql/data/SamzaSqlRelMsgMetadata.java
```diff
@@ -56,60 +55,82 @@
...
+  public void setArrivalTime(long arrivalTime) {
     this.arrivalTime = arrivalTime;
   }
 
-  public boolean hasArrivalTime() { return arrivalTime != null && !arrivalTime.isEmpty(); }
+  public boolean hasArrivalTime() { return arrivalTime != 0L; }
 
   @JsonProperty("scanTime")
-  public String getscanTime() { return scanTime;}
+  public long getscanTimeNanos() { return scanTimeNanos;}
```
prateekm: `getScanTimeNanos` (uppercase S)

shenodaguirguis: done

samza-sql/src/main/java/org/apache/samza/sql/translator/TranslatorConstants.java
shenodaguirguis: thanks for the catch. Done

PR #1119, opened 2019-07-30
SAMZA-2284: Remove redundant getStreamMetadata calls in SamzaContainer startup sequence.

shanthoosh:
**Problem:**

SamzaContainer startup sequence fetches the metadata of same input streams multiple times. Fetching the metadata of a stream entails making a expensive remote call to underlying messaging broker. This redundant fetch-input-stream-metadata API invocations incurred significant delays in the start of actual message processing in the samza job.

**Impact:**

1. With some samza jobs at LinkedIn, we observed that this fetch-input-stream-metadata loop took around 1.5 hrs to complete.
2. The redundant fetch-input-stream-metadata remote API calls will significantly increase the load on the underlying messaging broker.

**Fix:**

This patch fixes the unnecessary delay by fetching the metadata of the input streams only once in SamzaContainer(rather than once per SystemStreamPartition per task).

**Validation:**

1. Fix the existing unit-tests.
2. Verified this change with a sample samza-yarn job in LinkedIn and with a hello-samza job from samza-hello-world examples in OSS.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
+++ b/samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
@@ -22,8 +22,6 @@ package org.apache.samza.container
 
 import java.util.{Objects, Optional}
 import java.util.concurrent.ScheduledExecutorService
-
-import org.apache.commons.lang3.StringUtils
 import org.apache.samza.SamzaException
 import org.apache.samza.checkpoint.OffsetManager
 import org.apache.samza.config.Config
@@ -31,7 +29,6 @@ import org.apache.samza.config.StreamConfig.Config2Stream
 import org.apache.samza.context._
 import org.apache.samza.job.model.{JobModel, TaskModel}
 import org.apache.samza.scheduler.{CallbackSchedulerImpl, EpochTimeScheduler, ScheduledCallback}
-import org.apache.samza.startpoint.Startpoint
 import org.apache.samza.storage.kv.KeyValueStore
 import org.apache.samza.storage.TaskStorageManager
 import org.apache.samza.system._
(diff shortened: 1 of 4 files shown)


shanthoosh:
@prateekm 

samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -326,22 +324,22 @@ class TaskInstance(
...
-        val upcomingOffset = partitionMetadata.getUpcomingOffset
-        val startingOffset = offsetManager.getStartingOffset(taskName, ssp)
-          .getOrElse(throw new SamzaException("No offset defined for SystemStreamPartition: %s" format ssp))
-
-        // Mark ssp to be caught up if the starting offset is already the
-        // upcoming offset, meaning the task has consumed all the messages
-        // in this partition before and waiting for the future incoming messages.
-        if(Objects.equals(upcomingOffset, startingOffset)) {
-          ssp2CaughtupMapping(ssp) = true
+        if (inputStreamMetadata.contains(ssp.getSystemStream)) {
```
dnishimura: Can there be a scenario where the `inputStreamMetadata` does not contain the `SystemStream` of interest? Would you need to fallback to querying the stream metadata cache from the task context?
If not, it'd be good to at least log a warning in the else block in case we need to troubleshoot in the future.

shanthoosh: > inputStreamMetadata does not contain the SystemStream of interest

No, that scenario will not happen.  `inputStreamMetadata` map will contain metadata of all input streams in a SamzaContainer(Union of input streams assigned to all the `TaskInstance` in a container). So it will have the stream-metadata for any `SystemStreamPartition` assigned to a `TaskInstance`.

PR #1121, opened 2019-07-31
Samza-2287: Removing RegEx based matching while parsing the sql statement

srinipunuru:
SamzaSQLQueryParser uses a brittle regEx parsing to fetch the SQL statement from the SQL. Removing this Regex parsing and use the string generated from SqlSelect node.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/util/SamzaSqlQueryParser.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/util/SamzaSqlQueryParser.java
@@ -117,10 +117,16 @@ public class SamzaSqlQueryParser {
         LOG.info("Parsed select query {} from sql {}", selectQuery, sql);
         sources = getSourcesFromSelectQuery(sqlSelect);
       } else {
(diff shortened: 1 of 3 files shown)


samza-sql/src/main/java/org/apache/samza/sql/util/SamzaSqlQueryParser.java
```diff
@@ -117,7 +113,8 @@ public static QueryInfo parseQuery(String sql) {
       sink = sqlInsert.getTargetTable().toString();
       if (sqlInsert.getSource() instanceof SqlSelect) {
         SqlSelect sqlSelect = (SqlSelect) sqlInsert.getSource();
-        selectQuery = m.group(2);
+        selectQuery = sqlSelect.toString();
+        LOG.info("Parsed select query {} from sql {}", selectQuery, sql);
         sources = getSourcesFromSelectQuery(sqlSelect);
       } else {
         throw new SamzaException("Sql query is not of the expected format");
```
gauravrkulkarni: nit: not directly related to this PR, but might want to add what the expected format is in the exception message here and below.

PR #1126, opened 2019-08-05
Adding two additional params to DiagnosticsStreamMessage

rmatharu:


samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
```diff
@@ -80,20 +82,20 @@
   private final SystemStream diagnosticSystemStream;
 
   public DiagnosticsManager(String jobName, String jobId, Map<String, ContainerModel> containerModels,
-      Integer containerMemoryMb, Integer containerNumCores, Integer numStoresWithChangelog, String containerId,
-      String executionEnvContainerId, String taskClassVersion, String samzaVersion, String hostname,
+      Integer containerMemoryMb, Integer containerNumCores, Integer numStoresWithChangelog, Long maxHeapSizeBytes, Integer containerThreadPoolSize,
```
mynameborat: nit: can we add the new parameters at the end?

rmatharu: we could, i wanted to try and keep the sizing-related params together, container-id related together, and the remaining at the end.

samza-core/src/main/java/org/apache/samza/util/DiagnosticsUtil.java
```diff
@@ -101,6 +101,8 @@ public MetadataFileContents(String version, String metricsSnapshot) {
       ClusterManagerConfig clusterManagerConfig = new ClusterManagerConfig(config);
       int containerMemoryMb = clusterManagerConfig.getContainerMemoryMb();
       int containerNumCores = clusterManagerConfig.getNumCores();
+      long maxHeapSizeBytes = Runtime.getRuntime().maxMemory();
+      int containerThreadPoolSize = new JobConfig(config).getThreadPoolSize();
```
cameronlee314: There is a usage of `JobConfig` above on line 99. Could you please create one `JobConfig` and reuse it?

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
cameronlee314: Are these arguments nullable? If not, can you please use the primitive versions?

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
```diff
@@ -68,6 +68,8 @@
   private final Integer containerMemoryMb;
   private final Integer containerNumCores;
   private final Integer numStoresWithChangelog;
+  private final long maxHeapSizeBytes;
+  private final Integer containerThreadPoolSize;
```
cameronlee314: Is this nullable? If not, please use primitive.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
```diff
@@ -80,20 +82,20 @@
   private final SystemStream diagnosticSystemStream;
 
   public DiagnosticsManager(String jobName, String jobId, Map<String, ContainerModel> containerModels,
```
cameronlee314: For a constructor with this many args and a good chance for needing to add more, consider expanding to one line per argument. Then additions in the future will be easier to read, and it will be easier to keep the line lengths short.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
```diff
@@ -185,6 +203,14 @@ public Integer getNumStoresWithChangelog() {
         CONTAINER_NUM_STORES_WITH_CHANGELOG_METRIC_NAME);
   }
 
+  public Integer getContainerThreadPoolSize() {
+    return (Integer) getFromMetricsMessage(GROUP_NAME_FOR_DIAGNOSTICS_MANAGER, CONTAINER_THREAD_POOL_SIZE_METRIC_NAME);
+  }
+
+  public Long getMaxHeapSize() {
```
cameronlee314: Minor: can you keep the same ordering for methods? So could you move this one above `getContainerThreadPoolSize`

rmatharu:
Addressed all comments.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsManager.java
```diff
@@ -93,10 +105,22 @@ public DiagnosticsManager(String jobName, String jobId, Map<String, ContainerMod
...
-      int containerMemoryMb, int containerNumCores, int numStoresWithChangelog, Long maxHeapSizeBytes, Integer containerThreadPoolSize,
-      String containerId, String executionEnvContainerId, String taskClassVersion, String samzaVersion, String hostname,
-      SystemStream diagnosticSystemStream, SystemProducer systemProducer, Duration terminationDuration,
+  DiagnosticsManager(String jobName,
+      String jobId,
+      Map<String, ContainerModel> containerModels,
+      int containerMemoryMb,
+      int containerNumCores,
+      int numStoresWithChangelog,
+      Long maxHeapSizeBytes,
```
cameronlee314: Can these be primitives too?

rmatharu: yup, done

cameronlee314:
@rmatharu Please remember to include the JIRA ticket number in the commit message.

PR #1127, opened 2019-08-06
SAMZA-2291: Custom IME Put in InMemoryManager runs into Offset mismatch even if offsets are correctly set

Sanil15:
- Fixed the untested code path
- Fixed the bug in InMemoryManager

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
+++ b/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
@@ -19,6 +19,7 @@
 
 package org.apache.samza.system.inmemory;
 
+import com.google.common.base.Preconditions;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


Sanil15:
@bharathkk @shanthoosh for review

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
```diff
@@ -88,7 +88,7 @@ void put(SystemStreamPartition ssp, Object key, Object message) {
   void put(SystemStreamPartition ssp, IncomingMessageEnvelope envelope) {
     List<IncomingMessageEnvelope> messages = bufferedMessages.get(ssp);
     String offset = String.valueOf(messages.size());
-    if (envelope.getOffset().equals(offset)) {
+    if (!envelope.getOffset().equals(offset)) {
```
shanthoosh: Would be great to check if the envelope.getOffset() is not null.

Sanil15: Done

shanthoosh:
@Sanil15 
Merged the patch to trunk. Thanks for the bug-fix.

PR #1130, opened 2019-08-07
SAMZA-2294: LocalApplicationRunner support for missing coordinator stream

dnishimura:
Certain standalone configurations do not define a `job.coordinator.system` because a coordinator stream is not necessary. For example, in PassthroughJobCoordinator configured jobs or integration test jobs might not have this defined. This PR allows support for such scenarios.

Added additional unit tests and verified by @xinyuiscool with his testing.
@xinyuiscool - Please take a look when you get a chance.


shanthoosh:
@dnishimura 
Just curious. Please correct me if I'm wrong. 

Do we know why `TestRunner` integration tests did not fail before this fix, since TestRunner uses both `LocalApplicationRunner` and `PassThroughCoordinator`?

dnishimura:
> @dnishimura
> Just curious. Please correct me if I'm wrong.
> 
> Do we know why `TestRunner` integration tests did not fail before this fix, since TestRunner uses both `LocalApplicationRunner` and `PassThroughCoordinator`?
@shanthoosh 
I have to verify, but I think it had `job.coordinator.system` defined.

PR #1132, opened 2019-08-08
SAMZA-2296: Metadata streams not created in Standalone

dnishimura:
**Root cause**
In standalone, the configs are being read from the JobModel, but in the standalone scenario, the config in the JobModel is not populated. The checkpoint and changelog system configurations were attempted to be read from the job model rather than the internally constructed configs.

**Fix**
Use the constructed config rather than the configs from JobModel

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceUtil.java
+++ b/samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceUtil.java
@@ -21,6 +21,7 @@ package org.apache.samza.coordinator;
 
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.samza.checkpoint.CheckpointManager;
+import org.apache.samza.config.Config;
 import org.apache.samza.config.TaskConfig;
 import org.apache.samza.job.model.JobModel;
 import org.apache.samza.metrics.MetricsRegistry;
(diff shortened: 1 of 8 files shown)


dnishimura:
Please review, but hold off until I do some testing.

dnishimura:
Tested the fix on a backport to Samza 1.2 with a test standalone and test yarn job. I tested on both existing instances of the jobs and new instances of the jobs to trigger the creation of the new metadata streams. All verified to work. Please review.

samza-core/src/main/java/org/apache/samza/standalone/PassthroughJobCoordinator.java
```diff
@@ -86,7 +86,7 @@ public void start() {
     try {
       jobModel = getJobModel();
       // TODO metrics registry has been null here for a while; is it safe?
-      MetadataResourceUtil metadataResourceUtil = new MetadataResourceUtil(jobModel, null, getClass().getClassLoader());
+      MetadataResourceUtil metadataResourceUtil = new MetadataResourceUtil(jobModel, null, getClass().getClassLoader(), config);
       metadataResourceUtil.createResources();
```
Sanil15: Since TestRunner uses PassthroughJobCoordinator are the resources created (Changelog & Coordinator Stream) only when configs are present (job.coordinator.system & job.changelog.system with additional configs) ? Hence we need no changes on the TestRunner side?

dnishimura: This is for checkpoint and changelog, not coordinator. The coordinator issue is a separate issue that does not affect version 1.2 and will be addressed in a separate PR.

samza-core/src/main/java/org/apache/samza/coordinator/MetadataResourceUtil.java
```diff
@@ -62,7 +59,13 @@ public void createResources() {
     createChangelogStreams();
   }
 
+  @VisibleForTesting
   void createChangelogStreams() {
-    ChangelogStreamManager.createChangelogStreams(jobModel.getConfig(), jobModel.maxChangeLogStreamPartitions);
+    ChangelogStreamManager.createChangelogStreams(config, jobModel.maxChangeLogStreamPartitions);
```
shanthoosh: Would be great if we can log a message at appropriate level that we're creating the change-log and checkpoint stream. Currently it is not being logged anywhere, so it'll be useful during debugging.

dnishimura: They are being logged in the create calls themselves. That's how I was able to figure out these weren't being called.

PR #1133, opened 2019-08-08
SAMZA-2297: InMemorySystemAdmin offsets are off-by-one in some cases 

cameronlee314:
Added unit tests.
Ran a local build to make sure existing usages are still working.
Changed `TestSamzaSqlEndToEnd.testEndToEndStreamTableRightJoin` (which uses intermediate streams) to use in-memory system. Failed before this change, succeeded after this change.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
+++ b/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
@@ -179,9 +179,10 @@ class InMemoryManager {
             .entrySet()
             .stream()
             .collect(Collectors.toMap(entry -> entry.getKey().getPartition(), entry -> {
-                String oldestOffset = "0";
-                String newestOffset = String.valueOf(entry.getValue().size());
-                String upcomingOffset = String.valueOf(entry.getValue().size() + 1);
+                List<IncomingMessageEnvelope> messages = entry.getValue();
+                String oldestOffset = messages.isEmpty() ? null : "0";
+                String newestOffset = messages.isEmpty() ? null : String.valueOf(messages.size() - 1);
+                String upcomingOffset = String.valueOf(messages.size());
 
                 return new SystemStreamMetadata.SystemStreamPartitionMetadata(oldestOffset, newestOffset, upcomingOffset);
 
@@ -191,7 +192,7 @@ class InMemoryManager {
   }
 
   private List<IncomingMessageEnvelope> poll(SystemStreamPartition ssp, String offset) {
-    int startingOffset = Integer.parseInt(offset);
+    int startingOffset = (offset == null) ? 0 : Integer.parseInt(offset);
     List<IncomingMessageEnvelope> messageEnvelopesForSSP = bufferedMessages.getOrDefault(ssp, new LinkedList<>());
 
     if (startingOffset >= messageEnvelopesForSSP.size()) {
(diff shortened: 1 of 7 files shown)


samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemConsumer.java
```diff
@@ -77,8 +77,19 @@ public void stop() {
...
+      offsetToRegister = "0";
+    } else {
+      LOG.info("Registering ssp {} with starting offset {}", systemStreamPartition, offset);
+      offsetToRegister = offset;
+    }
+    /*
+     * A null offset is the same as the beginning of the stream. Can't use null directly since ConcurrentHashMap doesn't
+     * allow it.
+     */
+    sspToOffset.put(systemStreamPartition, (offset == null) ? "0" : offsetToRegister);
```
Sanil15: if, else above already compare for offset == null, why do it again?

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemConsumer.java
```diff
@@ -77,8 +77,19 @@ public void stop() {
...
+    String offsetToRegister;
+    if (offset == null) {
+      LOG.info("Registering ssp {} with starting offset null, overriding to 0", systemStreamPartition);
+      offsetToRegister = "0";
+    } else {
+      LOG.info("Registering ssp {} with starting offset {}", systemStreamPartition, offset);
+      offsetToRegister = offset;
+    }
+    /*
+     * A null offset is the same as the beginning of the stream. Can't use null directly since ConcurrentHashMap doesn't
```
Sanil15: can we also add a comment to function docs on assumption with "null" offset?

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemorySystemConsumer.java
cameronlee314: Good catch. I forgot to undo this.

samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
```diff
@@ -179,9 +179,10 @@ private SystemStreamMetadata constructSystemStreamMetadata(
             .entrySet()
             .stream()
             .collect(Collectors.toMap(entry -> entry.getKey().getPartition(), entry -> {
-                String oldestOffset = "0";
-                String newestOffset = String.valueOf(entry.getValue().size());
-                String upcomingOffset = String.valueOf(entry.getValue().size() + 1);
+                List<IncomingMessageEnvelope> messages = entry.getValue();
+                String oldestOffset = messages.isEmpty() ? null : "0";
```
prateekm: @cameronlee314 Any objection to defaulting this to "0" instead of null in case the SSP is empty? This is consistent with the behavior for Kafka as well where an empty SSP returns (0, null, 0) as (oldest, newest, upcoming) offsets. Will also remove the need to handle nulls in Consumer#register.

cameronlee314: `SystemStreamMetadata.SystemStreamPartitionMetadata.getOldestOffset` is documented with "A null value means the stream is empty". That's why I used `null` here. Does that mean that Kafka is not following the API?

prateekm: Seems like it. https://github.com/apache/samza/blob/master/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java#L403

I need to rely on this information during changelog restore for transactional state. It doesn't make sense to change the behavior for Kafka as part of the transactional state changes. Since InMemoryStore is used as a changelog for tests, I'll relax the assertion/validation for non-null starting offsets in the restore path. Thanks for the pointer.

PR #1134, opened 2019-08-09
Adding Ray to list

rmatharu:


vjagadish1989:
"Add rmatharu to committer's list"

PR #1135, opened 2019-08-09
SAMZA-2299: StreamTableJoin operator should invoke user join function only once

mynameborat:


samza-core/src/main/java/org/apache/samza/operators/impl/StreamTableJoinOperatorImpl.java
```diff
@@ -74,10 +74,11 @@ protected void handleInit(Context context) {
   }
 
   private Collection<JM> getJoinOutput(K key, Object value, M message) {
-    JM output = Optional.ofNullable(value)
+    R record = Optional.ofNullable(value)
```
dengpanyin: I didn't see how user join function could be invoked more than once before this change.

Can be simpler:
R record = value == null ? null : (R)  KV.of(key, value);

mynameborat: If the user join function returned null, due to the way optional works, we will invoke `orElseGet` and subsequently invoking the join function twice now with different parameters.

PR #1136, opened 2019-08-10
SAMZA-2298: Fix CoordinatorStreamStore creation for LocalApplicationRunner

dnishimura:
**Root Causes**
For standalone, to prevent a double bootstrap of the coordinator stream in the coordinator stream metadata store from the `StreamProcessor` and from the `ZkJobCoordinator`, the initialization of the metadata store was moved up to the `LocalApplicationRunner`. However, during the refactor, the creation of the underlying coordinator stream was left out.

The other root cause is that the assumption was made if a metadata store was not passed in to the `LocalApplicationRunner`, by default it will create a coordinator stream metadata store. However, if a PassthroughJobCoordinator is used or if the `job.coordinator.system` is not defined, the underlying coordinator stream cannot be created. These need to be accounted for.

**Fix**
Create the default coordinator stream metadata store only if using a `ZkJobCoordinator` and the coordinator system is defined, then create the underlying coordinator stream if a coordinator stream metadata store is used.

**Testing**
Tested against beam and the beam-runner examples which use the `PassthroughJobCoordinator`. Also tested against a `ZkJobCoordinator` job.

**Follow on**
The next phase of the metadata store abstraction [SAMZA-2271](https://issues.apache.org/jira/browse/SAMZA-2271) will clean up these work arounds.

@xinyuiscool @shanthoosh - please review when you get a chance.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
@@ -57,10 +57,15 @@ import org.apache.samza.metadatastore.MetadataStoreFactory;
 import org.apache.samza.metrics.MetricsRegistryMap;
 import org.apache.samza.metrics.MetricsReporter;
 import org.apache.samza.processor.StreamProcessor;
+import org.apache.samza.system.SystemAdmin;
+import org.apache.samza.system.SystemAdmins;
+import org.apache.samza.system.SystemStream;
 import org.apache.samza.task.TaskFactory;
 import org.apache.samza.task.TaskFactoryUtil;
+import org.apache.samza.util.CoordinatorStreamUtil;
 import org.apache.samza.util.ReflectionUtil;
 import org.apache.samza.util.Util;
+import org.apache.samza.zk.ZkJobCoordinatorFactory;
 import org.apache.samza.zk.ZkMetadataStoreFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -96,7 +101,7 @@ public class LocalApplicationRunner implements ApplicationRunner {
    * @param config configuration for the application
    */
   public LocalApplicationRunner(SamzaApplication app, Config config) {
-    this(app, config, getMetadataStoreFactory(new JobConfig(config)));
+    this(app, config, getDefaultCoordinatorStreamStoreFactory(new JobConfig(config)));
   }
 
   /**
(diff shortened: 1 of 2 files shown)


dnishimura:
@shanthoosh @xinyuiscool - please review when you get a chance. Thanks.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -272,15 +284,49 @@ CountDownLatch getShutdownLatch() {
...
+      if (metadataStoreFactory.get() instanceof CoordinatorStreamMetadataStoreFactory) {
+        if (createUnderlyingCoordinatorStream(config)) {
+          MetadataStore coordinatorStreamStore =
+              metadataStoreFactory.get().getMetadataStore("NoOp", config, new MetricsRegistryMap());
+          LOG.info("Created coordinator stream store of type: {}", coordinatorStreamStore.getClass().getSimpleName());
+          return coordinatorStreamStore;
+        }
+      } else {
+        MetadataStore otherMetadataStore =
+            metadataStoreFactory.get().getMetadataStore("NoOp", config, new MetricsRegistryMap());
```
shanthoosh: Why is the namespace set as NoOp?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -272,15 +284,49 @@ CountDownLatch getShutdownLatch() {
...
+        return otherMetadataStore;
+      }
     }
+
+    LOG.warn("No coordinator stream store created.");
     return null;
   }
 
+  @VisibleForTesting
+  boolean createUnderlyingCoordinatorStream(Config config) {
```
shanthoosh: Just curious. 
Every samza-standalone job has an associated  coordinator stream. Wouldn't this method return true for all standalone jobs?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -272,15 +284,49 @@ CountDownLatch getShutdownLatch() {
...
+    // TODO: This work around method is necessary due to SAMZA-2182 - Metadata store: disconnect between creation and usage of the underlying storage
+    //  and will be addressed in the next phase of metadata store abstraction
+    if (new JobConfig(config).getCoordinatorSystemNameOrNull() == null) {
+      LOG.warn("{} or {} not configured. Coordinator stream not created.",
+          JobConfig.JOB_COORDINATOR_SYSTEM, JobConfig.JOB_DEFAULT_SYSTEM);
+      return false;
+    }
+    SystemStream coordinatorSystemStream = CoordinatorStreamUtil.getCoordinatorSystemStream(config);
```
shanthoosh: Just curious. 

 It looks like the coordinator-stream-creation control-flow is idempotent. Why can't we house these coordinator-stream creation logic  in the `CoordinatorStreamMetadataStore.init()`. `MetadataStore.init()` contract allows us to do that (and `ZkMetadataStore` implementation already does something similar).

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
dnishimura: Recall that namespaces no longer have meaning to the pure metadata stores and that namespace handling has been moved up to the `NamespaceAwareMetadataStore` decorator class. We will need to address this in the next version of the metadata store abstraction.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
dnishimura: Not true. Coordinator streams were introduced to Standalone only recently. In the case of passthrough coordinator jobs, they may not have a default system or `job.coordinator.system` defined and therefore cannot create a coordinator stream.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
dnishimura: If you look a few lines up, there's a ticket in the comment that expresses the same concern: `SAMZA-2182 - Metadata store: disconnect between creation and usage of the underlying storage`. This is a general issue with coordinator stream metadata store and will need to be eventually addressed (but not in this PR though :) )

PR #1137, opened 2019-08-14
SAMZA-2301: Improve zookeeper metadata-store implementation

shanthoosh:
- We've observed corner cases where-in all the instances of an standalone application do not see the same state in zookeeper, i.e, some instance see the up-to-date JobModel state and some see an out-dated inconsistent state. There is an minuscule propagation delay(depending upon n/w bandwidth) between the leader of zookeeper quorum and the other servers in the ensemble. Consider the case where an follower undergoes the following execution sequence. 
   - Follower receives an jobModel version change notification from a zookeeper server.
   - Follower tries to read JobModel, but receives a session disconnect from the  up-to date zookeeper server. IOTec ZkClient library retries connecting to other servers in the ensemble and connection to an out-dated zookeeper server is established successfully. The read for JobModel from this out-dated zookeeper server would return null for the new JobModel zookeeper path and there by killing the standalone processor. 
    

This patch solves the problem,. by adding fixed retries in the zookeeper metadata-store read path.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
@@ -356,10 +356,12 @@ public class ZkJobCoordinator implements JobCoordinator {
     // If JobModel exists in zookeeper && cached JobModel version is unequal to JobModel version stored in zookeeper.
     if (zkJobModelVersion != null && !Objects.equals(cachedJobModelVersion, zkJobModelVersion)) {
       JobModel jobModel = readJobModelFromMetadataStore(zkJobModelVersion);
-      for (ContainerModel containerModel : jobModel.getContainers().values()) {
-        containerModel.getTasks().forEach((taskName, taskModel) -> changeLogPartitionMap.put(taskName, taskModel.getChangelogPartition().getPartitionId()));
+      if (jobModel != null) {
+        for (ContainerModel containerModel : jobModel.getContainers().values()) {
+          containerModel.getTasks().forEach((taskName, taskModel) -> changeLogPartitionMap.put(taskName, taskModel.getChangelogPartition().getPartitionId()));
+        }
+        cachedJobModelVersion = zkJobModelVersion;
       }
-      cachedJobModelVersion = zkJobModelVersion;
     }
 
     GrouperMetadata grouperMetadata = getGrouperMetadata(zkJobModelVersion, processorNodes);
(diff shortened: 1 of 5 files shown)


shanthoosh:
@rmatharu Please take a look when you get a chance.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -356,10 +356,12 @@ JobModel generateNewJobModel(List<ProcessorNode> processorNodes) {
     // If JobModel exists in zookeeper && cached JobModel version is unequal to JobModel version stored in zookeeper.
     if (zkJobModelVersion != null && !Objects.equals(cachedJobModelVersion, zkJobModelVersion)) {
```
rmatharu: Looks like TestZkLocalApplicationRunner.java needs to also be updated, there are some failed tests.

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -403,12 +405,14 @@ private GrouperMetadataImpl getGrouperMetadata(String jobModelVersion, List<Proc
...
-          taskToProcessorId.put(taskModel.getTaskName(), containerModel.getId());
-          for (SystemStreamPartition partition : taskModel.getSystemStreamPartitions()) {
-            taskToSSPs.computeIfAbsent(taskModel.getTaskName(), k -> new ArrayList<>());
-            taskToSSPs.get(taskModel.getTaskName()).add(partition);
+      if (jobModel != null) {
+        for (ContainerModel containerModel : jobModel.getContainers().values()) {
+          for (TaskModel taskModel : containerModel.getTasks().values()) {
+            taskToProcessorId.put(taskModel.getTaskName(), containerModel.getId());
+            for (SystemStreamPartition partition : taskModel.getSystemStreamPartitions()) {
+              taskToSSPs.computeIfAbsent(taskModel.getTaskName(), k -> new ArrayList<>());
```
rmatharu: Could this be 
`taskToSSPs.putIfAbsent(taskModel.getTaskName(), new ArrayList<>());` ?

samza-core/src/main/java/org/apache/samza/zk/ZkMetadataStore.java
```diff
@@ -103,6 +104,25 @@ public void init() {
     return null;
   }
 
+  private byte[] readFromZk(String zkPath) {
+    long retryTimeOutMs = zkConfig.getZkSessionTimeoutMs();
+    long startTimeMs = System.currentTimeMillis();
+    while (true) {
+      try {
+        byte[] value = zkClient.readData(zkPath);
+        LOG.info("Read for zkPath: {} returned: {}.", zkPath, value);
```
rmatharu: printing value will just print the reference right?
if this is encoded, could you print it as a string instead.

shanthoosh:
> If it is latter, when the new server catches up, it should trigger another data change notification which will be eventually picked up by the change listener.

Ignoring the data-change event by simple logging alone will result in data inconsistency issues. It and miss zookeeper events.

Consider the following sequence:

1. Follower(P1) receives the JobModelVersion change notification from server S1 in ensemble which had the majority quorum write.
2. Before Follower(P1) reads the JobModel, it receives a disconnect from zookeeper server S1 and successfully reconnects to zookeeper server S2. 
3. On re-connection, the follower(P1) will re-create all the data-node and child-node watches. 
4. If the watch is created by P1 after the JobModel version change write propogration had happened to zookeeper-server S2 and not the JobModel write, then the notification will not be propogated to the follower.

So logging alone wouldn't solve the problem. 

I preferred the retry since it can happen for all zk-writes.. Alternatively,  we can read the  `JobModel` for the agreed barrier version always before starting the SamzaContainer(and remove retry).

PR #1138, opened 2019-08-14
SAMZA-2302: Add equals, hashcode and toString to KV

xinyuiscool:
The current KV class misses these primitive methods so that it can be used correctly in comparison, map or hash. Adding these methods to fix it.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/operators/KV.java
+++ b/samza-api/src/main/java/org/apache/samza/operators/KV.java
@@ -59,6 +59,7 @@ public final class KV<K, V> {
       return false;
     }
     KV<?, ?> otherKv = (KV<?, ?>) other;
+    // deepEquals is used here for arrays. This impl comes from Beam.
     return Objects.deepEquals(this.key, otherKv.key)
         && Objects.deepEquals(this.value, otherKv.value);
   }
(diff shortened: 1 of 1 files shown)


xinyuiscool:
@cameronlee314 : please take a look. Thx!

samza-api/src/main/java/org/apache/samza/operators/KV.java
```diff
@@ -45,4 +49,30 @@ public K getKey() {
...
+  @Override
+  public boolean equals(Object other) {
+    if (this == other) {
+      return true;
+    }
+    if (!(other instanceof KV)) {
+      return false;
+    }
+    KV<?, ?> otherKv = (KV<?, ?>) other;
+    return Objects.deepEquals(this.key, otherKv.key)
```
mynameborat: Does this class assume key and value to be non null? If so, can we add a preconditions check in the constructor to ensure key and value are non null?

xinyuiscool: Objects.deepEquals takes care of null checks.

cameronlee314: Please add a comment about why we need `deepEquals` and that this implementation comes from Beam.

xinyuiscool: Added.

PR #1140, opened 2019-08-16
Adding documentation for the regex topic monitor config

rmatharu:


docs/learn/documentation/versioned/jobs/samza-configurations.md
```diff
@@ -67,6 +67,7 @@ These are the basic properties for setting up a Samza application.
...
+|job.coordinator.monitor-input-regex.frequency.ms| 300000| If using the `org.apache.samza.config.RegExTopicGenerator` as a rewriter (using `job.config.rewriter.**_rewriter-name_**.class`) in your Samza-YARN job, this configuration allows you to change the default monitoring frequency for detecting new topics matching the input regex. To disable monitoring for new topics set this value to < 0. |
```
vjagadish1989: reword: s/configuration/property so that the tone is consistent with the rest of the config docs

make docs concise: "This property sets the interval to poll for new topics matching the input-regex."

unless answered elsewhere, clarify these questions:

1.  what's the behavior when new topics matching the regex are discovered? 
2. why would/should a Samza customer care?

PR #1144, opened 2019-08-23
SAMZA-2309: Remove readFn requirement for remote tables

dxichen:
@dengpan-yin @dengpanyin @xinyuiscool 

Removed validation for read table requied

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
+++ b/samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
@@ -77,7 +77,7 @@ public class RemoteTableDescriptor<K, V> extends BaseTableDescriptor<K, V, Remot
   public static final String WRITE_RETRY_POLICY = "io.write.retry.policy";
   public static final String BATCH_PROVIDER = "io.batch.provider";
 
-  // Input support for a specific remote store (required)
+  // Input support for a specific remote store (optional)
   private TableReadFunction<K, V> readFn;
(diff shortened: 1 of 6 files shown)


prateekm:
Is there already validation that either read or write fn is present?

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
```diff
@@ -345,7 +346,6 @@ public String getProviderFactoryClassName() {
 
   @Override
   protected void validate() {
-    Preconditions.checkNotNull(readFn, "TableReadFunction is required.");
```
dengpanyin: Thanks for the change!

should we validate at least a readfn or writefn exist? 
Also, The constructor of RemoteTable and AsyncRemoteTable also has validations that readfn is required. Purely remove the validation here may not be enough. 

After this change, it's possible unit/regression test may fail, need to fix as well if any.

dxichen:
@dengpanyin @prateekm Added validations and tests for readFn and writeFn, tests for the new behaviour and updated AsyncRemoteTable and RemoteTable

samza-api/src/main/java/org/apache/samza/table/descriptors/RemoteTableDescriptor.java
dxichen: Good catch, added!

dengpanyin:
Looks good to me, thanks!

PR #1145, opened 2019-08-23
Emitting numPersistentStores instead of num stores with changelog from DiagnosticsManager

rmatharu:


samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
```diff
@@ -55,7 +55,7 @@
   private static final String STOP_EVENT_LIST_METRIC_NAME = "stopEvents";
   private static final String CONTAINER_MB_METRIC_NAME = "containerMemoryMb";
   private static final String CONTAINER_NUM_CORES_METRIC_NAME = "containerNumCores";
-  private static final String CONTAINER_NUM_STORES_WITH_CHANGELOG_METRIC_NAME = "numStoresWithChangelog";
```
cameronlee314: Can all readers of this data handle a missing entry for this? Just double checking on compatibility.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -225,10 +226,13 @@ public boolean hasDurableStores() {
...
+   * Helper method to get the number of persistent stores.
    */
-  public int getNumStoresWithChangelog() {
-    Config subConfig = subset(STORE_PREFIX, true);
-    return new Long(subConfig.keySet().stream().filter(key -> key.endsWith(CHANGELOG_SUFFIX)).count()).intValue();
+  public int getNumPersistentStores() {
+    return (int) getStoreNames().stream()
+        .map(storeName -> getStorageFactoryClassName(storeName))
+        .filter(factoryName -> factoryName.isPresent())
+        .filter(factoryName -> factoryName.get().equals(PERSISTENT_STORE_FACTORY))
```
cameronlee314: This is inconsistent with `StoreProperties.isPersistedToDisk`. Over there, it is "persisted" if it is not an in-memory store. Should we do the same here?

rmatharu: Ah, yes, I was looking for exactly that. 
Had to add a constant because of the module dependency.

samza-core/src/main/scala/org/apache/samza/diagnostics/DiagnosticsStreamMessage.java
rmatharu: Yup, only the asc prototype uses this currently.

PR #1147, opened 2019-08-27
Adding a container.maxheap.mb config to configure max heap size

rmatharu:


samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
```diff
@@ -75,6 +76,7 @@
   public static final String CONTAINER_MAX_MEMORY_MB = "yarn.container.memory.mb";
   public static final String CLUSTER_MANAGER_MEMORY_MB = "cluster-manager.container.memory.mb";
   private static final int DEFAULT_CONTAINER_MEM = 1024;
+  public static final String CLUSTER_MANAGER_MAX_HEAP_MB = "cluster-manager.container.maxheap.mb";
```
Sanil15: Please update the configuration reference doc with this config

rmatharu: It would be, yes. 
However I realized I dont need this open source change, so am closing out this PR.

PR #1148, opened 2019-08-30
SAMZA-2313: Adding validation for Samza Sql statements.

atoomula:


samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
```diff
@@ -0,0 +1,279 @@
...
+
+  protected void validate(RelRoot relRoot, SamzaSqlQueryParser.QueryInfo qinfo, SamzaSqlApplicationConfig sqlConfig)
+      throws SamzaSqlValidatorException {
+    // Validate select fields (including Udf return types) with output schema
+    validateOutput(relRoot, sqlConfig.getRelSchemaProviders().get(qinfo.getSink()));
+
+    // TODO:
+    //  1. Validate Udf arguments.
+    //  2. Validate operators. These are the operators that are supported by Calcite but not by Samza Sql.
+    //     Eg: LogicalAggregate is not supported by Samza Sql.
```
srinipunuru: I am not sure i understand this. Are you saying we should add support for validating aggregation  when samza SQL supports aggregations?

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
```diff
@@ -0,0 +1,279 @@
...
+  protected void validateOutputRecords(RelRecordType outputRecord, RelRecordType projectRecord)
+      throws SamzaSqlValidatorException {
+    Map<String, RelDataType> outputRecordMap = outputRecord.getFieldList().stream().collect(
+        Collectors.toMap(RelDataTypeField::getName, RelDataTypeField::getType));
+    Map<String, RelDataType> projectRecordMap = projectRecord.getFieldList().stream().collect(
+        Collectors.toMap(RelDataTypeField::getName, RelDataTypeField::getType));
+
+    // There could be default values for the output schema and hence fields in project schema could be a subset of
+    // fields in output schema.
+    // TODO: Validate that all non-default value fields in output schema are set in the projected fields.
```
srinipunuru: Can you open a bug for this one?

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
```diff
@@ -0,0 +1,279 @@
...
+   * SELECT `non_existing_name`, `name` AS `string_value`
+   * -------^^^^^^^^^^^^^^^^^^^--------------------------
+   * FROM `testavro`.`level1`.`level2`.`SIMPLE1` AS `s`
+   * WHERE `s`.`id` = 1
+   *
+   * @param query sql query
+   * @param e Exception returned by Calcite
+   * @return formatted error string
+   */
+  public static String formatErrorString(String query, Exception e) {
```
srinipunuru: This is going to be super useful! Thanks a lot for adding this capability :)  We just need to be careful to update this logic when When calcite is upgraded and they change the exception string.  Can we add a unit test to protect this logic when we upgrade calcite?

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
atoomula: Changed to "LogicalAggregate with sum function is not supported by Samza Sql."

samza-sql/src/main/java/org/apache/samza/sql/planner/QueryPlanner.java
```diff
@@ -87,49 +82,38 @@ public QueryPlanner(Map<String, RelSchemaProvider> relSchemaProviders,
...
-                sqlSchema.getFields().stream().map(SqlSchema.SqlField::getFieldSchema).collect(Collectors.toList()));
-
-            SqlSchema newSchema = new SqlSchema(fieldNames, fieldTypes);
-            RelDataType relationalSchema = relSchemaConverter.convertToRelSchema(newSchema);
-            previousLevelSchema.add(sourcePart, createTableFromRelSchema(relationalSchema));
-            break;
-          }
-        }
-      }
+      fetchSourceSchema(rootSchema);
```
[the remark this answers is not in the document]
atoomula: > Looks great overall. It would be good if you could add a UT to protect the string parsing logic.

That's a good idea. Added that test now.

PR #1149, opened 2019-09-04
SAMZA-2316: Validate that all non-default value fields in output schema are set in the projected fields.

atoomula:


samza-sql/src/main/java/org/apache/samza/sql/avro/AvroTypeFactoryImpl.java
```diff
@@ -60,62 +59,68 @@ private SqlSchema convertSchema(List<Schema.Field> fields) {
 
     SqlSchemaBuilder schemaBuilder = SqlSchemaBuilder.builder();
     for (Schema.Field field : fields) {
-      SqlFieldSchema fieldSchema = convertField(field.schema());
+      // Consider any field with default value as nullable. Is it the right assumption ?
```
srinipunuru: This is slightly tricky. Nullable fields in avro is of type union with one of the value being null. 
The fields with default value means they are optional during serializing and deserializing. 

The behavior on the consumption side. 
1. All avro fields should have values including nullable fields as well as optional fields(fields with default values)

Behavior on the producer side.
1. fields with default values are optional
2. Nullable fields may still need values if they are not optional

So nullability and optionality are slightly different. I am not sure it is a good idea to conflate them.

weiqingy:
Nit: copy the words in the description to the title?

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroTypeFactoryImpl.java
atoomula: Thanks for clarifying the behavior. Updated the logic by separating out nullable fields and fields with default values. Using default values for validation but not nullable fields.

samza-api/src/main/java/org/apache/samza/sql/schema/SqlFieldSchema.java
```diff
@@ -24,44 +24,52 @@
...
+  private final SamzaSqlFieldType fieldType;
+  private final SqlFieldSchema elementType;
+  private final SqlFieldSchema valueType;
+  private final SqlSchema rowSchema;
+  private final Boolean isNullable;
+  private final Boolean hasDefaultValue;
 
-  private SqlFieldSchema(SamzaSqlFieldType fieldType, SqlFieldSchema elementType, SqlFieldSchema valueType, SqlSchema rowSchema) {
+  private SqlFieldSchema(SamzaSqlFieldType fieldType, SqlFieldSchema elementType, SqlFieldSchema valueType,
+      SqlSchema rowSchema, boolean isNullable, boolean hasDefaultValue) {
```
srinipunuru: Do you think it makes sense to call the hasDefaultValue field as optionalField? I think it also is a good idea to document  isNullable and isOptional in sqlFieldSchema to make the differences clear?

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
```diff
@@ -98,32 +101,63 @@ protected void validate(RelRoot relRoot, SamzaSqlQueryParser.QueryInfo qinfo, Sa
   protected void validateOutput(RelRoot relRoot, RelSchemaProvider relSchemaProvider) throws SamzaSqlValidatorException {
     RelRecordType outputRecord = (RelRecordType) QueryPlanner.getSourceRelSchema(relSchemaProvider,
         new RelSchemaConverter());
+    // Get Samza Sql schema along with Calcite schema. The reason is that the Calcite schema does not have a way
+    // to represent fields with default values while Samza Sql schema can represent default value fields. This is
```
srinipunuru: Do you think it makes sense to call these as optional fields rather than fields with default values?

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
```diff
@@ -98,32 +101,63 @@ protected void validate(RelRoot relRoot, SamzaSqlQueryParser.QueryInfo qinfo, Sa
...
-    // fields in output schema.
-    // TODO: SAMZA-2316: Validate that all non-default value fields in output schema are set in the projected fields.
+    // Ensure that all non-default value fields in output schema are set in the projected fields and are of the
+    // same type.
+    for (Map.Entry<String, RelDataType> entry : outputRecordMap.entrySet()) {
+      RelDataType projectFieldType = projectRecordMap.get(entry.getKey());
+      SqlFieldSchema outputSqlFieldSchema = outputFieldSchemaMap.get(entry.getKey());
+
+      if (projectFieldType == null) {
+        if (entry.getKey().equals(SamzaSqlRelMessage.KEY_NAME) || outputSqlFieldSchema.hasDefaultValue()) {
```
srinipunuru: Can you add comments here on why we are special casing KEY_NAME field?

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
```diff
@@ -98,32 +101,63 @@ protected void validate(RelRoot relRoot, SamzaSqlQueryParser.QueryInfo qinfo, Sa
...
         Collectors.toMap(RelDataTypeField::getName, RelDataTypeField::getType));
+    Map<String, SqlFieldSchema> outputFieldSchemaMap = outputSqlSchema.getFields().stream().collect(
+        Collectors.toMap(SqlSchema.SqlField::getFieldName, SqlSchema.SqlField::getFieldSchema));
     Map<String, RelDataType> projectRecordMap = projectRecord.getFieldList().stream().collect(
         Collectors.toMap(RelDataTypeField::getName, RelDataTypeField::getType));
 
-    // There could be default values for the output schema and hence fields in project schema could be a subset of
-    // fields in output schema.
-    // TODO: SAMZA-2316: Validate that all non-default value fields in output schema are set in the projected fields.
+    // Ensure that all non-default value fields in output schema are set in the projected fields and are of the
```
srinipunuru: nit: rather than call projected fields, Can we just call them as fields from sql statement?

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
```diff
@@ -98,32 +101,63 @@ protected void validate(RelRoot relRoot, SamzaSqlQueryParser.QueryInfo qinfo, Sa
...
+        throw new SamzaSqlValidatorException(errMsg);
+      } else if (!compareFieldTypes(entry.getValue(), outputSqlFieldSchema, projectFieldType)) {
+        String errMsg = String.format("Field '%s' with type '%s' in output schema does not match the field type '%s' in"
+            + " projected fields.", entry.getKey(), entry.getValue(), projectFieldType);
+        LOG.error(errMsg);
+        throw new SamzaSqlValidatorException(errMsg);
+      }
+    }
+
+    // Ensure that all projected fields exist in the output schema and are of the same type.
```
srinipunuru: nit: Rather than call them as projected fields can we say fields from sql statement?.

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
```diff
@@ -98,32 +101,63 @@ protected void validate(RelRoot relRoot, SamzaSqlQueryParser.QueryInfo qinfo, Sa
...
     for (Map.Entry<String, RelDataType> entry : projectRecordMap.entrySet()) {
       RelDataType outputFieldType = outputRecordMap.get(entry.getKey());
+      SqlFieldSchema outputSqlFieldSchema = outputFieldSchemaMap.get(entry.getKey());
+
       if (outputFieldType == null) {
-        if (entry.getKey().equals(SamzaSqlRelMessage.OP_NAME)) {
+        if (entry.getKey().equals(SamzaSqlRelMessage.OP_NAME) || entry.getKey().equals(SamzaSqlRelMessage.KEY_NAME)) {
```
srinipunuru: Can you add comments on why we are special casing these fields?

atoomula: Made key as optional and nullable value as it is by default being set to null in SamzaSqlRelMessage.

PR #1150, opened 2019-09-11
SAMZA-2318: Empty config values from coordinator stream shouldn't be removed

dengpanyin:
In the previous PR: https://github.com/apache/samza/pull/1010/files
config with null/empty values will be filter out when reading the coordinator stream,
this will cause some problems when the user expects an empty string to be a value config.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/util/CoordinatorStreamUtil.scala
+++ b/samza-core/src/main/scala/org/apache/samza/util/CoordinatorStreamUtil.scala
@@ -131,8 +131,8 @@ object CoordinatorStreamUtil extends Logging {
       } else {
         val valueSerde: CoordinatorStreamValueSerde = new CoordinatorStreamValueSerde(SetConfig.TYPE)
         val valueAsString: String = valueSerde.fromBytes(valueAsBytes)
-        if (StringUtils.isBlank(valueAsString)) {
-          warn("Value for key: %s in config is empty or null. Ignoring it." format key)
+        if (valueAsString == null) {
+          warn("Value for key: %s in config is decoded to be null. Ignoring it." format key)
         } else {
           configMap.put(key, valueAsString)
         }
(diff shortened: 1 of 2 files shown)


samza-core/src/main/scala/org/apache/samza/util/CoordinatorStreamUtil.scala
```diff
@@ -131,8 +131,8 @@ object CoordinatorStreamUtil extends Logging {
       } else {
         val valueSerde: CoordinatorStreamValueSerde = new CoordinatorStreamValueSerde(SetConfig.TYPE)
         val valueAsString: String = valueSerde.fromBytes(valueAsBytes)
-        if (StringUtils.isBlank(valueAsString)) {
-          warn("Value for key: %s in config is empty or null. Ignoring it." format key)
+        if (valueAsString == null) {
+          warn("Value for key: %s in config is decoded to be null. Ignoring it." format key)
```
dnishimura: Might be useful to add a unit test to test that null values appear as non-entries and empty string values are entries with an empty string when loaded from the coordinator stream.

samza-core/src/test/scala/org/apache/samza/util/TestCoordinatorStreamUtil.scala
```diff
@@ -34,4 +39,31 @@ class TestCoordinatorStreamUtil {
...
   }
+
+  @Test
+  def testReadConfigFromCoordinatorStream {
+    val keyForNonBlankVal = "app.id"
+    val nonBlankVal = "1"
+    val keyForEmptyVal = "task.opt"
+    val emptyVal = ""
+    val keyForNullVal = "zk.server"
+    val nullVal = null
```
cameronlee314: This var is unused.
In order to test the `valueAsString == null` condition, can you please add `"null"` (as a String) to the config map? That should be the case where the bytes are non-null, but the `valueAsString` is null.

dengpanyin: Good catch. valueSerde.toBytes(nullVal) is able to correctly encode the null value. This function "readConfigFromCoordinatorStream" is using valueSerde.fromBytes(valueAsBytes) for deserialization, so we need to pass null value to the config map (other than a string of "null").

PR #1151, opened 2019-09-11
[SAMZA-2198] Fix deadlock in container signal handler

abhishekshivanna:
The fix removes join() on the main thread from the shutdown
hook which causes a deadlock when a signal is received.
The shutdown hook now polls periodically to check
the container status on shutdown.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/runtime/ClusterBasedProcessorLifecycleListener.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/ClusterBasedProcessorLifecycleListener.java
@@ -32,7 +32,7 @@ import org.slf4j.LoggerFactory;
 
 public class ClusterBasedProcessorLifecycleListener implements SamzaContainerListener {
   private static final Logger log = LoggerFactory.getLogger(ClusterBasedProcessorLifecycleListener.class);
-  private final Runnable onSignal;
+  private final Runnable shutdownHookCallback;
   private Thread shutdownHookThread = null;
   private CountDownLatch shutdownLatch = new CountDownLatch(1);
   private TaskConfig taskConfig;
@@ -40,10 +40,10 @@ public class ClusterBasedProcessorLifecycleListener implements SamzaContainerLis
   private volatile Throwable containerException = null;
 
   ClusterBasedProcessorLifecycleListener(Config config, ProcessorLifecycleListener processorLifecycleListener,
-      Runnable onSignal) {
+      Runnable shutdownHookCallback) {
     this.taskConfig = new TaskConfig(config);
     this.processorLifecycleListener = processorLifecycleListener;
-    this.onSignal = onSignal;
+    this.shutdownHookCallback = shutdownHookCallback;
   }
 
   @Override
@@ -81,13 +81,7 @@ public class ClusterBasedProcessorLifecycleListener implements SamzaContainerLis
   }
 
   private void removeShutdownHook() {
-    try {
-      if (shutdownHookThread != null) {
-        removeJVMShutdownHook();
-      }
-    } catch (IllegalStateException e) {
-      // Thrown when then JVM is already shutting down, so safe to ignore.
-    }
+    removeJVMShutdownHook(shutdownHookThread);
   }
 
   private void addShutdownHook() {
@@ -97,7 +91,7 @@ public class ClusterBasedProcessorLifecycleListener implements SamzaContainerLis
         long shutdownMs = taskConfig.getShutdownMs();
         log.info("Attempting to shutdown container from inside shutdownHook, will wait up to {} ms.", shutdownMs);
         try {
-          onSignal.run();
+          shutdownHookCallback.run();
           boolean hasShutdown = shutdownLatch.await(shutdownMs, TimeUnit.MILLISECONDS);
           if (hasShutdown) {
             log.info("Shutdown complete");
@@ -111,7 +105,7 @@ public class ClusterBasedProcessorLifecycleListener implements SamzaContainerLis
         }
       }
     };
-    addJVMShutdownHook();
+    addJVMShutdownHook(shutdownHookThread);
   }
 
   @VisibleForTesting
@@ -120,14 +114,22 @@ public class ClusterBasedProcessorLifecycleListener implements SamzaContainerLis
   }
 
   @VisibleForTesting
-  void addJVMShutdownHook() {
-    Runtime.getRuntime().addShutdownHook(shutdownHookThread);
-    log.info("Added Samza container shutdown hook");
+  void addJVMShutdownHook(Thread shutdownHookThread) {
+    if (shutdownHookThread != null) {
+      Runtime.getRuntime().addShutdownHook(shutdownHookThread);
+      log.info("Added Samza container shutdown hook");
+    }
   }
 
   @VisibleForTesting
-  void removeJVMShutdownHook() {
-    Runtime.getRuntime().removeShutdownHook(shutdownHookThread);
-    log.info("Removed Samza container shutdown hook");
+  void removeJVMShutdownHook(Thread shutdownHookThread) {
+    try {
+      if (shutdownHookThread != null) {
+        Runtime.getRuntime().removeShutdownHook(shutdownHookThread);
+        log.info("Removed Samza container shutdown hook");
+      }
+    } catch (IllegalStateException e) {
+      // Thrown when then JVM is already shutting down, so safe to ignore.
+    }
   }
 }
--- a/samza-core/src/main/java/org/apache/samza/runtime/ClusterBasedProcessorLifecycleListener.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/ClusterBasedProcessorLifecycleListener.java
@@ -1,0 +1,133 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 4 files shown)


samza-core/src/main/java/org/apache/samza/runtime/ContainerLaunchUtil.java
```diff
@@ -142,19 +150,59 @@ public void beforeStart() {
...
+                // Thrown when then JVM is already shutting down, so safe to ignore.
+              }
+            }
+
+            private void addShutdownHook() {
+              shutdownHookThread = new Thread("Samza Container Shutdown Hook Thread") {
+                @Override
+                public void run() {
+                  long shutdownMs = taskConfig.getShutdownMs();
+                  log.info("Attempting to shutdown container from inside shutdownHook, will wait upto {} ms.", shutdownMs);
```
cameronlee314: Minor: s/upto/up to

samza-core/src/main/java/org/apache/samza/runtime/ContainerLaunchUtil.java
```diff
@@ -142,19 +150,59 @@ public void beforeStart() {
...
+                    container.shutdown();
+                    boolean hasShutdown = shutdownLatch.await(shutdownMs, TimeUnit.MILLISECONDS);
+                    if (hasShutdown) {
+                      log.info("Shutdown complete");
+                    } else {
+                      log.error("Did not shut down within {} ms, exiting.", shutdownMs);
+                      ThreadUtil.logThreadDump("Thread dump from Samza Container Shutdown Hook.");
+                    }
+                  } catch (InterruptedException e) {
+                    e.printStackTrace();
```
cameronlee314: Please use the `log` here instead of `printStackTrace`. Maybe do the same thing here as if it didn't shut down in time.

samza-core/src/main/java/org/apache/samza/runtime/ContainerLaunchUtil.java
```diff
@@ -142,19 +150,59 @@ public void beforeStart() {
...
               listener.afterStart();
+              addShutdownHook();
             }
 
             @Override
             public void afterStop() {
               log.info("Container Stopped");
               listener.afterStop();
+              removeShutdownHook();
+              shutdownLatch.countDown();
```
cameronlee314: Can you please double check that either `afterStop` or `afterFailure` will always be called, even in failure cases? Ideally, the shutdown timeout is only hit because some part of the shutdown was taking a long time, but not because there was some flow which accidentally forgot to call the callback.
I think you are good here, but just double checking.

samza-core/src/main/java/org/apache/samza/runtime/ContainerLaunchUtil.java
```diff
@@ -132,6 +136,10 @@ private static void run(
 
       container.setContainerListener(
           new SamzaContainerListener() {
```
cameronlee314: Could we add some unit tests for this? This anonymous class is getting pretty complex. Maybe this is also a good time to pull it out into its own class.
You'll have to be a little careful since you don't actually want to add a real shutdown hook to the test process though. One way you could do it is to pull `Runtime.getRuntime().addShutdownHook` and `Runtime.getRuntime().removeShutdownHook` into internal helper methods and then use a spy to mock those internal methods. And then maybe you could capture the shutdown hook thread using Mockito and call it directly when necessary.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -976,41 +974,6 @@ class SamzaContainer(
     }
   }
 
-  def addShutdownHook {
-    val runLoopThread = Thread.currentThread()
-    shutdownHookThread = new Thread("Samza Container Shutdown Hook Thread") {
-      override def run() = {
-        info("Shutting down, will wait up to %s ms." format shutdownMs)
-        shutdownRunLoop()  //TODO: Pull out shutdown hook to LocalContainerRunner or SP
```
cameronlee314: It's great to see TODOs resolved :)

samza-core/src/main/java/org/apache/samza/runtime/ContainerLaunchUtil.java
abhishekshivanna: Look like either `afterStop` or `afterFailure` will always be called.
On graceful shutdown (via container.shutdown() or runLoop shutting down) -> `afterStop`
On Exception -> `afterFailure`

samza-core/src/main/java/org/apache/samza/runtime/ClusterBasedProcessorLifecycleListener.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  @VisibleForTesting
+  Thread getShutdownHookThread() {
+    return this.shutdownHookThread;
+  }
+
+  @VisibleForTesting
+  void addJVMShutdownHook() {
+    Runtime.getRuntime().addShutdownHook(shutdownHookThread);
```
cameronlee314: Since `shutdownHookThread` is a mutable instance variable, it makes it a little easier to accidentally call these before `shutdownHookThread` is actually set to a value (you currently have it right, but someone make change this code in the future).
Consider passing the `shutdownHookThread` as an argument to this method instead of using the instance var directly. Then the calling method is more aware that `shutdownHookThread` needs to be non-null.
Or you could check for null in this method.

samza-core/src/main/java/org/apache/samza/runtime/ClusterBasedProcessorLifecycleListener.java
```diff
@@ -0,0 +1,133 @@
...
+    log.info("Added Samza container shutdown hook");
+  }
+
+  @VisibleForTesting
+  void removeJVMShutdownHook() {
```
cameronlee314: Same as above about passing null `shutdownHookThread`.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -753,7 +753,6 @@ class SamzaContainer(
       startConsumers
       startSecurityManger
 
-      addShutdownHook
```
cameronlee314: I think there is a shutdown hook thread variable that is now unused.

samza-core/src/test/java/org/apache/samza/runtime/TestClusterBasedProcessorLifecycleListener.java
```diff
@@ -0,0 +1,98 @@
...
+  private RunLoop mockRunLoop;
+  private ClusterBasedProcessorLifecycleListener clusterBasedProcessorLifecycleListener;
+  private ProcessorLifecycleListener processorLifecycleListener;
+  private Runnable mockOnSignal;
+
+  @Before
+  public void setup() {
+    mockRunLoop = mock(RunLoop.class);
+    mockOnSignal = mock(Runnable.class);
+    container = StreamProcessorTestUtils.getDummyContainer(mockRunLoop, Mockito.mock(StreamTask.class));
```
cameronlee314: A unit test for `ClusterBasedProcessorLifecycleListener` should not require setting up a `SamzaContainer`. You just need to call the methods of the listener directly.
If you would like to test that the listener lifecycle methods are properly called, then those tests belong in the test for `SamzaContainer`.

samza-core/src/test/java/org/apache/samza/runtime/TestClusterBasedProcessorLifecycleListener.java
```diff
@@ -0,0 +1,98 @@
...
+    doNothing().when(clusterBasedProcessorLifecycleListener).addJVMShutdownHook();
+    doNothing().when(clusterBasedProcessorLifecycleListener).removeJVMShutdownHook();
+  }
+
+  @Test
+  public void testProcessorLifecycleListenerIsCalledOnContainerShutdown() {
+    container.run();
+    container.shutdown();
+
+    Mockito.verify(clusterBasedProcessorLifecycleListener, Mockito.times(1)).addJVMShutdownHook();
```
cameronlee314: For here and below: By default, `verify` will check `Mockito.times(1)`, so you don't need to include that. Also, instead of `Mockito.times(0)`, you can use `Mockito.never()`.

samza-core/src/main/java/org/apache/samza/runtime/ClusterBasedProcessorLifecycleListener.java
```diff
@@ -0,0 +1,133 @@
...
+  private static final Logger log = LoggerFactory.getLogger(ClusterBasedProcessorLifecycleListener.class);
+  private final Runnable onSignal;
+  private Thread shutdownHookThread = null;
+  private CountDownLatch shutdownLatch = new CountDownLatch(1);
+  private TaskConfig taskConfig;
+  private ProcessorLifecycleListener processorLifecycleListener;
+  private volatile Throwable containerException = null;
+
+  ClusterBasedProcessorLifecycleListener(Config config, ProcessorLifecycleListener processorLifecycleListener,
+      Runnable onSignal) {
```
cameronlee314: This is always the "shutdown container" method, right? Could you please name it to be a little more specific to that? `onSignal` seems a bit vague.

abhishekshivanna: Although we currently use this to only call `container.shutdown()`, the callback can be used to execute anything else. The naming here is based on the fact that the callback is invoked in the signal-handler/shutdown hook.
Do you think `shutdownHookCallback` is a better name ?

samza-core/src/test/java/org/apache/samza/runtime/TestClusterBasedProcessorLifecycleListener.java
abhishekshivanna: Good point! Seems like the tests for `SamzaContainer` does this check already. Fixed my tests.

samza-core/src/main/java/org/apache/samza/runtime/ClusterBasedProcessorLifecycleListener.java
abhishekshivanna: Makes sense. Fixed.

PR #1153, opened 2019-09-13
SAMZA-2320: Samza-sql: Refactor validation to cover more cases and make it more extensible.

atoomula:
Code changes involve the following:
1. AvroTypeFactoryImpl: Change API access modifiers from private to protected to be called from derived classes.
2. SamzaSqlValidator:
  2.1 Change order of validation of select query fields and output fields so that the error messages make more sense.
  2.2 Take care of the case where a field projected in select query is mentioned more than once and added tests for it. Eg: SELECT myUdf(id) as id, * from store.myTable. This type of pattern is typically followed when users want to just modify one field in the input table while keeping rest of the fields the same.
  2.3 Add a short term isOptional API to take care of cases where RelSchemaProviders have a complex mechanism to determine if a given output field is optional. Once the RelSchemaProviders are fixed to properly mark the fields as optional, this API will be removed.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/avro/AvroTypeFactoryImpl.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/avro/AvroTypeFactoryImpl.java
@@ -1,21 +1,21 @@
 /*
-* Licensed to the Apache Software Foundation (ASF) under one
-* or more contributor license agreements.  See the NOTICE file
-* distributed with this work for additional information
-* regarding copyright ownership.  The ASF licenses this file
-* to you under the Apache License, Version 2.0 (the
-* "License"); you may not use this file except in compliance
-* with the License.  You may obtain a copy of the License at
-*
-*   http://www.apache.org/licenses/LICENSE-2.0
-*
-* Unless required by applicable law or agreed to in writing,
-* software distributed under the License is distributed on an
-* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
-* KIND, either express or implied.  See the License for the
-* specific language governing permissions and limitations
-* under the License.
-*/
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
 
 package org.apache.samza.sql.avro;
 
@@ -31,9 +31,10 @@ import org.apache.samza.sql.schema.SqlSchemaBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+
 /**
- * Factory that creates {@link SqlSchema} from the Avro Schema. This is used by the
- * {@link AvroRelConverter} to convert Avro schema to Samza Sql schema.
+ * Factory that creates {@link SqlSchema} from the Avro Schema.
+ * TODO: This class will be renamed to AvroTypeFactoryImpl and copied to open source.
  */
 public class AvroTypeFactoryImpl extends SqlTypeFactoryImpl {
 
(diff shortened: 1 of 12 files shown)


srinipunuru:
Can you add description on 
1. What kind of refactoring is done
2. What are the more cases it covers?
3. How is it extensible in future?

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
```diff
@@ -86,33 +88,46 @@ public void validate(List<String> sqlStmts) throws SamzaSqlValidatorException {
...
+    LOG.info("Samza Sql Validation finished successfully.");
+  }
+
+  protected boolean skipOutputValidation(RelRoot relRoot, SamzaSqlQueryParser.QueryInfo qinfo,
+      SamzaSqlApplicationConfig sqlConfig) {
+    return false;
+  }
 
-    validateOutputRecords(outputRecord, outputSqlSchema, projetRecord);
+  protected boolean isOptional(RelSchemaProvider outputRelSchemaProvider, String outputFieldName,
```
weiqingy: Could you add some comments saying why isOptional and skipOutputValidation alway return false?

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
```diff
@@ -121,6 +136,47 @@ protected void validateOutputRecords(RelRecordType outputRecord, SqlSchema outpu
...
+      RelDataType outputFieldType = outputRecordMap.get(projectedFieldName);
+      SqlFieldSchema outputSqlFieldSchema = outputFieldSchemaMap.get(projectedFieldName);
+
+      if (outputFieldType == null) {
+        // If the field names are specified more than once in the select query, calcite appends 'n' as suffix to the
+        // dup fields based on the order they are specified, where 'n' starts from 0 for the first dup field.
+        // Take the following example: SELECT id as str, secondaryId as str, tertiaryId as str FROM store.myTable
+        //   Calcite renames the projected fieldNames in select query as str, str0, str1 respectively.
+        // Samza Sql allows a field name to be specified up to 2 times. Do the validation accordingly.
+        if (projectedFieldName.endsWith("0")) {
```
srinipunuru: Can we check whether there is another field in the fieldList without "0" to ensure that this is really a field that calcite appended "0" to and not a user specified field?

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
atoomula: Added a comment over this and a TODO

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroTypeFactoryImpl.java
```diff
@@ -52,15 +52,18 @@ public SqlSchema createType(Schema schema) {
       throw new SamzaException(msg);
     }
 
-    return convertSchema(schema.getFields());
+    return convertSchema(schema.getFields(), true);
   }
 
-  private SqlSchema convertSchema(List<Schema.Field> fields) {
+  protected boolean isOptional(Schema.Field field, boolean isTopLevel) {
+    return field.defaultValue() != null;
```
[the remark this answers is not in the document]
atoomula: Removed this altogether.

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroTypeFactoryImpl.java
```diff
@@ -52,15 +52,18 @@ public SqlSchema createType(Schema schema) {
...
+  }
+
+  private SqlSchema convertSchema(List<Schema.Field> fields, boolean isTopLevel) {
 
     SqlSchemaBuilder schemaBuilder = SqlSchemaBuilder.builder();
```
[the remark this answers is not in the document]
atoomula: Removed it.

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
```diff
@@ -86,33 +88,46 @@ public void validate(List<String> sqlStmts) throws SamzaSqlValidatorException {
...
-    SqlSchema outputSqlSchema = QueryPlanner.getSourceSqlSchema(relSchemaProvider);
+    SqlSchema outputSqlSchema = QueryPlanner.getSourceSqlSchema(outputRelSchemaProvider);
 
-    LogicalProject project = (LogicalProject) relRoot.rel;
-    RelRecordType projetRecord = (RelRecordType) project.getRowType();
+    validateOutputRecords(outputRecord, outputSqlSchema, projetRecord, outputRelSchemaProvider);
+    LOG.info("Samza Sql Validation finished successfully.");
+  }
+
+  protected boolean skipOutputValidation(RelRoot relRoot, SamzaSqlQueryParser.QueryInfo qinfo,
```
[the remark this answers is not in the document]
atoomula: Removed it.

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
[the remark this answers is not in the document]
atoomula: Added comments over it. This is short term fix.

samza-sql/src/main/java/org/apache/samza/sql/planner/SamzaSqlValidator.java
atoomula: It is already taking caring of this scenario. I now added a test case as well.

samza-sql/src/test/java/org/apache/samza/sql/planner/TestSamzaSqlValidator.java
```diff
@@ -59,6 +59,42 @@ public void testBasicValidation() throws SamzaSqlValidatorException {
...
+        "Insert into testavro.outputTopic select id, true as bool_value, false as bool_value, true as bool_value"
+            + " from testavro.level1.level2.SIMPLE1 as s where s.id = 1");
+    Config samzaConfig = SamzaSqlApplicationRunner.computeSamzaConfigs(true, new MapConfig(config));
+
+    List<String> sqlStmts = fetchSqlFromConfig(config);
+    new SamzaSqlValidator(samzaConfig).validate(sqlStmts);
+  }
+
+  @Test (expected = SamzaSqlValidatorException.class)
+  public void testFieldEndingInZeroValidation() throws SamzaSqlValidatorException {
```
srinipunuru: Can you add a test case where there is a legit field named "field0" in the schema?

samza-sql/src/test/java/org/apache/samza/sql/planner/TestSamzaSqlValidator.java
```diff
@@ -59,6 +59,42 @@ public void testBasicValidation() throws SamzaSqlValidatorException {
     new SamzaSqlValidator(samzaConfig).validate(sqlStmts);
   }
 
+  @Test
+  public void testRepeatedTwiceFieldsValidation() throws SamzaSqlValidatorException {
+    Map<String, String> config = SamzaSqlTestConfig.fetchStaticConfigsWithFactories(1);
+    config.put(SamzaSqlApplicationConfig.CFG_SQL_STMT,
+        "Insert into testavro.outputTopic select id, true as bool_value, false as bool_value"
```
srinipunuru: I think we need to add comments here on why we support a field repeated twice and not 3 times?

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroTypeFactoryImpl.java
```diff
@@ -67,11 +70,11 @@ private SqlSchema convertSchema(List<Schema.Field> fields) {
     return schemaBuilder.build();
   }
 
-  private SqlFieldSchema convertField(Schema fieldSchema) {
+  protected SqlFieldSchema convertField(Schema fieldSchema) {
```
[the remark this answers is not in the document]
atoomula: Removed this change.

samza-sql/src/main/java/org/apache/samza/sql/avro/AvroTypeFactoryImpl.java
```diff
@@ -44,37 +45,47 @@ public AvroTypeFactoryImpl() {
   }
 
   public SqlSchema createType(Schema schema) {
+    validateTopLevelAvroType(schema);
+    return convertSchema(schema.getFields(), true);
+  }
+
+  /**
+   * Given a schema field, determine if it is an optional field. There could be cases where a field (like audit header)
```
srinipunuru: auditHeader may confuse quite a few folks who read this code in open source. Please remove reference to auditHeader and provide a generic example on why this would be needed.

atoomula: Sure, removed the reference.

PR #1155, opened 2019-09-16
SAMZA-2322: Integrate startpoints with ZK Standalone integration tests.

dnishimura:
Added Startpoints to ZK Standalone integration tests.

### DIFF
--- a/samza-test/src/test/java/org/apache/samza/test/processor/TestStreamApplication.java
+++ b/samza-test/src/test/java/org/apache/samza/test/processor/TestStreamApplication.java
@@ -24,8 +24,9 @@ import java.io.ObjectInputStream;
 import java.io.Serializable;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
-import org.apache.samza.application.descriptors.StreamApplicationDescriptor;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.samza.application.StreamApplication;
+import org.apache.samza.application.descriptors.StreamApplicationDescriptor;
 import org.apache.samza.config.ApplicationConfig;
 import org.apache.samza.config.Config;
 import org.apache.samza.config.JobConfig;
@@ -141,7 +142,7 @@ public class TestStreamApplication implements StreamApplication {
     }
 
     static TestKafkaEvent fromString(String message) {
-      String[] messageComponents = message.split("|");
+      String[] messageComponents = StringUtils.split(message, "|");
       return new TestKafkaEvent(messageComponents[0], messageComponents[1]);
     }
   }
--- a/samza-test/src/test/java/org/apache/samza/test/processor/TestStreamApplication.java
+++ b/samza-test/src/test/java/org/apache/samza/test/processor/TestStreamApplication.java
@@ -21,10 +21,8 @@ package org.apache.samza.test.processor;
 
 import java.io.IOException;
 import java.io.ObjectInputStream;
-import java.io.Serializable;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.samza.application.StreamApplication;
 import org.apache.samza.application.descriptors.StreamApplicationDescriptor;
 import org.apache.samza.config.ApplicationConfig;
@@ -38,6 +36,7 @@ import org.apache.samza.serializers.StringSerde;
 import org.apache.samza.system.kafka.descriptors.KafkaInputDescriptor;
 import org.apache.samza.system.kafka.descriptors.KafkaOutputDescriptor;
 import org.apache.samza.system.kafka.descriptors.KafkaSystemDescriptor;
+import org.apache.samza.test.util.TestKafkaEvent;
 
 
 /**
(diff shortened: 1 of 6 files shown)


samza-test/src/test/java/org/apache/samza/test/processor/TestStreamApplication.java
```diff
@@ -141,7 +142,7 @@ public String toString() {
     }
 
     static TestKafkaEvent fromString(String message) {
-      String[] messageComponents = message.split("|");
+      String[] messageComponents = StringUtils.split(message, "|");
```
dnishimura: This was previously wrong. Calling `split` on the string directly expects a regex string and "|" is a regex character that needed to be escaped. Switched to `StringUtils.split` as it does what we thought was intended.

samza-test/src/test/java/org/apache/samza/test/processor/TestZkLocalApplicationRunner.java
```diff
@@ -1230,6 +1294,159 @@ public void testFirstProcessorDiesButSameRunIdForBatch() throws InterruptedExcep
...
   }
 
+  private void writeStartpoints(StartpointManager startpointManager, String streamName, Integer partitionCount, Startpoint startpoint) {
+    for (int p = 0; p < partitionCount; p++) {
+      startpointManager.writeStartpoint(new SystemStreamPartition(TEST_SYSTEM, streamName, new Partition(p)), startpoint);
+    }
+  }
+
+  @Test
+  public void testStartpoints() throws InterruptedException {
```
prateekm: Let's move this to a separate class instead of TestZkLAR.

dnishimura: @prateekm - Done. Reverted TestZkLAR (minus some minor refactoring) and moved this to its own test harness class.

samza-test/src/test/java/org/apache/samza/test/startpoint/StartpointTestHarness.java
```diff
@@ -0,0 +1,383 @@
...
+  @Override
+  public void tearDown() {
+    deleteTopics(
+        ImmutableList.of(inputKafkaTopic1, inputKafkaTopic2, inputKafkaTopic3, inputKafkaTopic4, outputKafkaTopic));
+    SharedContextFactories.clearAll();
+    super.tearDown();
+  }
+
+  @Test
+  public void testStartpoints() throws InterruptedException {
```
PawasChhokra: Can you break this method down into multiple functions: like one for startpoint specific, one for startpoint timestamp, etc. The code in its current state is too lengthy and not legible.

samza-test/src/test/java/org/apache/samza/test/startpoint/StartpointTestHarness.java
```diff
@@ -0,0 +1,383 @@
...
+    TreeMap<Integer, RecordMetadata> sentEvents1 =
```
PawasChhokra: Why are we using a TreeMap here specifically?

samza-test/src/test/java/org/apache/samza/test/processor/TestStreamApplication.java
PawasChhokra: Thanks for catching this!

samza-test/src/test/java/org/apache/samza/test/startpoint/StartpointTestHarness.java
dnishimura: Sure good suggestion.

samza-test/src/test/java/org/apache/samza/test/startpoint/StartpointTestHarness.java
dnishimura: I think previously I was iteratoring through the map (I forget why) and wanted to maintain ordering. Will change back to Map/HashMap

dnishimura:
@PawasChhokra @prateekm - I addressed your comments. Please review when you get a chance. Thanks.

samza-test/src/test/java/org/apache/samza/test/processor/TestTaskApplication.java
```diff
@@ -81,4 +107,8 @@ public void describe(TaskApplicationDescriptor appDescriptor) {
                  .withOutputStream(outputDescriptor)
                  .withTaskFactory((AsyncStreamTaskFactory) () -> new TestTaskImpl());
   }
+
+  public interface TaskApplicationCallback {
```
mynameborat: nit: prefer calling it `TaskApplicationProcessCallback` to be more explicit.

samza-test/src/test/java/org/apache/samza/test/startpoint/StartpointTestHarness.java
```diff
@@ -0,0 +1,468 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+
+
+public class StartpointTestHarness extends IntegrationTestHarness {
```
mynameborat: Is this a test harness? Prefer calling it `TestStartingpoint` otherwise.

samza-test/src/test/java/org/apache/samza/test/startpoint/StartpointTestHarness.java
```diff
@@ -0,0 +1,468 @@
...
+  private String outputKafkaTopic;
+  private ApplicationConfig applicationConfig1;
+  private ApplicationConfig applicationConfig2;
+  private ApplicationConfig applicationConfig3;
+  private ApplicationConfig applicationConfig4;
+  private String testStreamAppName;
+  private String testStreamAppId;
+
+  @Rule
+  public Timeout testTimeOutInMillis = new Timeout(150000, TimeUnit.MILLISECONDS);
```
mynameborat: Do we need such a long timeout? Is this inferred through some calculation from the debounce, barrier timeout etc? If so, can you add a doc?

samza-test/src/test/java/org/apache/samza/test/processor/TestTaskApplication.java
dnishimura: Will change.

samza-test/src/test/java/org/apache/samza/test/startpoint/StartpointTestHarness.java
dnishimura: Will rename. Thx.

samza-test/src/test/java/org/apache/samza/test/startpoint/StartpointTestHarness.java
dnishimura: Will figure out the right timeout and add doc. Good suggestion.

dnishimura:
@mynameborat - I addressed your comments. Please take a look.

PR #1156, opened 2019-09-16
SAMZA-2323: Provide option allow single containers to fail without failing the job

dnishimura:
This PR introduces a new config: `cluster-manager.container.fail.job.after.retries`. The default value (`true`) does not change existing behavior. However, if set false and a container fails after exhausting all its retries, the job will continue to run in a degraded state. 

This will allow healthy containers to continue while debugging the issue with the failed container.

@rmatharu @Sanil15 - Please take a look.
CC: @prateekm


docs/learn/documentation/versioned/jobs/samza-configurations.md
```diff
@@ -296,6 +296,7 @@ Samza supports both standalone and clustered ([YARN](yarn-jobs.html)) [deploymen
...
+|cluster-manager.container.fail.job.after.retries|true|This configuration sets the behavior of the job after all `cluster-manager.container.retry.count`s are exhausted on a single container. If set to true, the whole job will fail if any container fails after the last retry. If set to false, the job will continue to run without the failed container. The typical use cases of setting this to false is to aid in debugging the cluster manager in cases of unexpected failed containers and also to allow other healthy containers to continue to run so that lag does not accumulate across all containers. Samza job operators should diligent in monitoring the `job-healthy` and `failed-containers` metrics when setting this configuration to false. A full restart of the job is required if another attempt to restart the container is needed after failure.|
```
rmatharu: ...sets the behavior of the job after all `cluster-manager.container.retry.count` retries of a container are exhausted on a given container within the cluster-manager.container.retry.window.ms period .. is it?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -513,12 +509,23 @@ void onResourceCompletedWithUnknownStatus(SamzaResourceStatus resourceStatus, St
     int retryCount = clusterManagerConfig.getContainerRetryCount();
     int retryWindowMs = clusterManagerConfig.getContainerRetryWindowMs();
     int currentFailCount;
+    boolean retryContainerRequest = true;
 
     if (retryCount == 0) {
       LOG.error("Processor ID: {} (current Container ID: {}) failed, and retry count is set to 0, " +
```
rmatharu: Should this error message be removed?

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
```diff
@@ -192,6 +198,11 @@ else if (containsKey(CONTAINER_RETRY_COUNT)) {
     }
   }
 
+  public boolean getContainerFailJobAfterRetries() {
```
rmatharu: Should we add a javadoc here, 
and call it something like 
shouldFailJobAfterContainerRetries() ?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -513,12 +509,23 @@ void onResourceCompletedWithUnknownStatus(SamzaResourceStatus resourceStatus, St
...
-      jobFailureCriteriaMet = true;
+      // Failure criteria met only if failed containers can fail the job.
+      jobFailureCriteriaMet = clusterManagerConfig.getContainerFailJobAfterRetries();
+      if (jobFailureCriteriaMet) {
+        LOG.error("Processor ID: {} (current Container ID: {}) failed, and retry count is set to 0, " +
+            "so shutting down the application master and marking the job as failed.", processorId, containerId);
+      } else {
+        LOG.error("Processor ID: {} (current Container ID: {}) failed, and retry count is set to 0, " +
+            "marking the container as completed and will not retry request.", processorId, containerId);
+        incrementCompletedProcessorsAndUpdateState(state);
```
rmatharu: Perhaps we dont need the call here, since 
a. this will increment the "succeeded" container count regardless of the container's exit status. 
b. it increments the "suceeded" container count and matches it with the total-container-count. 
So in a 3-container job if container-0 fails 3 times, this will mark the job as "succeeded", perhaps not right.

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
dnishimura: Will add.

dnishimura:
@rmatharu - I addressed your comments. Please review when you get a chance.

PR #1157, opened 2019-09-16
SAMZA-2325 : Adding logic to read system config for repl-factor when creating a topic

rmatharu:


rmatharu:
@prateekm Could you please take a look

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -243,19 +244,35 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
...
+      val changelogSystemStream = new StorageConfig(config).getChangelogStream(storeName)
+      if (!changelogSystemStream.isPresent) {
+        throw new SamzaException("Changelog system not defined for store "+storeName)
+      }
+
+      val changelogSystem = StreamUtil.getSystemStreamFromNames(changelogSystemStream.get()).getSystem
+      changelogRF = Option.apply(getSystemDefaultReplicationFactor(changelogSystem, "2"))
+    }
+
+    changelogRF.get
```
cameronlee314: Would it be possible to have a flow such that you don't assume `changelogRF` has a value at the end? For example, maybe just do `changelogRF.getOrElse(DEFAULT_RF)` at the end?
Just in case someone changes this flow in the future, then `changelogRF` might accidentally be left empty at this point.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -268,7 +285,7 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
   def getChangelogStreamMaxMessageByte(name: String) = getOption(KafkaConfig.CHANGELOG_MAX_MESSAGE_BYTES format name) match {
     case Some(maxMessageBytes) => maxMessageBytes
     case _ =>
-      val changelogSystem = new StorageConfig(config).getChangelogSystem.orElse(null)
+      val changelogSystem = StreamUtil.getSystemStreamFromNames(new StorageConfig(config).getChangelogStream(name).get()).getSystem
```
cameronlee314: Should this do an `isPresent` check before calling `get` on the `Optional`?

samza-kafka/src/test/scala/org/apache/samza/config/TestKafkaConfig.scala
```diff
@@ -220,7 +220,6 @@ class TestKafkaConfig {
     val kafkaConfig = new KafkaConfig(mapConfig)
     assertEquals("3", kafkaConfig.getChangelogStreamReplicationFactor("store-with-override"))
     assertEquals("2", kafkaConfig.getChangelogStreamReplicationFactor("store-without-override"))
-    assertEquals("2", kafkaConfig.getDefaultChangelogStreamReplicationFactor)
```
cameronlee314: Can you please add unit tests for the new flows in `KafkaConfig` for RF?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -546,6 +546,19 @@ public KafkaStreamSpec toKafkaSpec(StreamSpec spec) {
       kafkaSpec = kafkaSpec.copyWithProperties(properties);
     } else {
       kafkaSpec = KafkaStreamSpec.fromSpec(spec);
+
```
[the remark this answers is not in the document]
rmatharu: yes, and for changelog not all configs were being respected (see getChangelogStreamReplicationFactor)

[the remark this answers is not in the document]
cameronlee314: Ah, yes, right. Although the fix for changelog was done in `KafkaConfig`, and then the fix for some of the other stream types was done here.
Thanks for clarifying.

samza-kafka/src/test/scala/org/apache/samza/config/TestKafkaConfig.scala
rmatharu: added in TestKafkaConfig and updated others

rmatharu:
@cameronlee314  Addressed all comments and feedback.

samza-core/src/test/java/org/apache/samza/config/TestStorageConfig.java
```diff
@@ -155,7 +154,7 @@ public void testGetStorageMsgSerde() {
   }
 
   @Test
-  public void testGetChangelogSystem() {
+  public void getChangelogStream() {
```
cameronlee314: There is already a test called `testGetChangelogStream`. Should this be merged with that?

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -285,7 +284,7 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
   def getChangelogStreamMaxMessageByte(name: String) = getOption(KafkaConfig.CHANGELOG_MAX_MESSAGE_BYTES format name) match {
     case Some(maxMessageBytes) => maxMessageBytes
     case _ =>
-      val changelogSystem = StreamUtil.getSystemStreamFromNames(new StorageConfig(config).getChangelogStream(name).get()).getSystem
+      val changelogSystem = StreamUtil.getSystemStreamFromNames(new StorageConfig(config).getChangelogStream(name).orElse(null)).getSystem
```
cameronlee314: It looks like this will throw an NPE if `getChangelogStream` returns empty.

samza-core/src/test/java/org/apache/samza/config/TestStorageConfig.java
rmatharu: done

samza-core/src/test/java/org/apache/samza/config/TestStorageConfig.java
```diff
@@ -54,9 +55,6 @@ public void testGetStoreNames() {
 
   @Test
   public void testGetChangelogStream() {
-    // empty config, so no changelog stream
-    assertEquals(Optional.empty(), new StorageConfig(new MapConfig()).getChangelogStream(STORE_NAME0));
```
cameronlee314: Did this need to be removed?

PR #1158, opened 2019-09-17
SAMZA-2324: Adding KV store metrics for rocksdb

bkonold:
1) size of store on disk (sst files; property provided by rocksdb)
2) maximum record size


prateekm:
@PawasChhokra can you take a look?

samza-kv/src/main/scala/org/apache/samza/storage/kv/SerializedKeyValueStore.scala
```diff
@@ -55,23 +55,27 @@ class SerializedKeyValueStore[K, V](
   }
 
   def put(key: K, value: V) {
-    metrics.puts.inc
     val keyBytes = toBytesOrNull(key, keySerde)
     val valBytes = toBytesOrNull(value, msgSerde)
     store.put(keyBytes, valBytes)
+    metrics.puts.inc
+    metrics.maxRecordSizeBytes.set(Math.max(metrics.maxRecordSizeBytes.getValue, valBytes.length))
```
cameronlee314: I think there might be a concurrency bug here for exact correctness. If two threads try to set the max at the same time, you might not always ultimately get the max. This is just metrics though, so I'm not sure how important that is to get exact correctness.

bkonold: You are correct. To avoid this I think we'd need to do a compareAndSet so each thread would only update if they actually are the new max value at the time the write happens.

InGraphs ends up diluting the granularity at most graph views anyway, so imo I wouldn't consider exactness important here. The intention here is more to track trend lines, than specific data points.

However, it may be useful to document the compromise we're making here if we don't decide to change this.

cameronlee314: From a metrics perspective, I agree that we don't need exactness (since this is open source, we shouldn't be referring to specific metrics display techs though). I think it is reasonable to document the known exactness issue and then go with this logic, as long as it won't get into a really bad state (e.g. access of incomplete object, significantly incorrect value).
However, do be wary that there is at least one place in Samza in which metrics is used for functionality (see "needed-containers" access in `ClientHelper.scala`; it probably shouldn't be used for functionality though).

bkonold: Ah, good point. From a more general perspective, since metrics in samza are designed in a "poll for the value" fashion, this already lends itself to lossy metrics data. Take MetricsSnapshotReporter for example, which will poll the registered metrics at some cadence and only see the value that is set at the particular polling interval. 

Nothing here should be getting to a bad state like you mention, we'd just be potentially recording values that aren't _actually_ the true maximum.

The problem may be exaggerated for jobs where message size is highly variable however. I think in general this problem would be best addressed by recording gauge metrics as a distribution rather than a single value, but that's out of scope for this PR.

bkonold: @cameronlee314 Thinking it might be better to document this concurrency issue in the javadoc of the Gauge.set() method, since anytime that's used in a multithreaded environment this would be a symptom.

Thoughts?

cameronlee314: It is this specific read-then-write pattern which is not threadsafe; `Gauge` itself is threadsafe. It can be argued that it is not `Gauge`'s responsibility to be able to handle a read-then-write pattern, so I don't think this issue should be called out there. `Gauge` isn't usually used to store values based on a previous value of the `Gauge`. I think if the user of `Gauge` wants to do "max", then the user needs to track the max separately (or use `Histogram` instead to track distribution, as you mentioned).

bkonold: Okay, you're right that it shouldn't be the responsibility of Gauge to warn about read then write thread safety. Easiest thing to do might be to replace with

`while(!compareAndSet(gauge.getValue,valSizeBytes);` or simply to use a lock.

Histogram feels like it is overkill since the distribution of record size may not be interesting... Max is interesting because it tells us how close we are to the kafka limit. Record size otherwise (median, 75th, 99th, etc) is interesting but will it be useful to customers in a meaningful way?

cameronlee314: I think other percentiles might give a useful insight into what applications are storing. Max can be influenced by single "out of ordinary" cases, but other percentiles give a more "normal" view into store usage.

prateekm: Max would be sufficient for now. Percentiles are more useful but also more expensive, we can add them if we get a request for them. The original request was for a metric to set alerts on if record sizes are increasing over time (e.g. for list values) and getting closer to the max logged store size limit. Re: single outlier, that's still relevant since a single large message will cause the produce to fail.

cameronlee314: @prateekm do you think we need to get "exact" (i.e. threadsafe) max or "approx" (non-threadsafe) max. If "approx" max, how do we communicate that it might not be 100% accurate?

prateekm: Exact max seems cheap enough to do that I'd prefer that. I don't think the overhead for CAS would be large enough to meaningfully affect throughput.

bkonold: Adjusted to use compareAndSet in a loop. LMK if there are alternative suggestions on formatting.

samza-kv/src/main/scala/org/apache/samza/storage/kv/SerializedKeyValueStore.scala
```diff
@@ -59,23 +59,23 @@ class SerializedKeyValueStore[K, V](
     val valBytes = toBytesOrNull(value, msgSerde)
     store.put(keyBytes, valBytes)
     metrics.puts.inc
-    metrics.maxRecordSizeBytes.set(Math.max(metrics.maxRecordSizeBytes.getValue, valBytes.length))
+    while({val max = metrics.maxRecordSizeBytes.getValue; !metrics.maxRecordSizeBytes.compareAndSet(max, Math.max(max, valBytes.length))}){}
```
cameronlee314: Although this format is compact, IMO the syntax looks a little weird. Maybe just write a more basic loop that looks more like Java syntax. Also, could you pull this into a helper method to share with `putAll`?

cameronlee314: `compareAndSet` probably doesn't have much overhead, but it might help to just check if the new `valBytes` is bigger than the current max (if not, then you don't even need to try the while loop). Most of the time, `valBytes` won't be bigger, so you won't have to set the value.

bkonold: sounds good

bkonold: good idea, will update this

prateekm:
@bkonold If this is ready, can you get this merged? Let's check why the tests failed before merging.

bkonold:
> @bkonold If this is ready, can you get this merged? Let's check why the tests failed before merging.

Dumbly mishandled some nulls. Should be fixed with the latest commit (tests all pass locally)

bkonold:
@prateekm @cameronlee314 assuming the latest changes are still ok (added a couple null checks), this PR is ready to be merged

PR #1159, opened 2019-09-17
Fix ClusterBasedProcessorLifecycleListener test to exit without waiting for timeout

abhishekshivanna:


samza-core/src/test/java/org/apache/samza/runtime/TestClusterBasedProcessorLifecycleListener.java
```diff
@@ -79,11 +79,14 @@ public void testLifecycleListenerAfterFailure() {
...
 
-    // setup shutdown hook
+    // call beforeStart to setup shutdownHook
     clusterBasedProcessorLifecycleListener.beforeStart();
 
     // Simulating shutdown hook invocation by JVM
     clusterBasedProcessorLifecycleListener.getShutdownHookThread().run();
+
+    // call afterStop to countdown shutdownLatch
+    clusterBasedProcessorLifecycleListener.afterStop();
```
bkonold: Just to clarify, this is to address the following line always waiting until the timeout? https://github.com/apache/samza/blob/9b3a9e68bca8cfef8a87542143212015b5263cea/samza-core/src/main/java/org/apache/samza/runtime/ClusterBasedProcessorLifecycleListener.java#L95

samza-core/src/test/java/org/apache/samza/runtime/TestClusterBasedProcessorLifecycleListener.java
```diff
@@ -79,11 +79,14 @@ public void testLifecycleListenerAfterFailure() {
   public void testShutdownHookInvokesShutdownHookCallback() {
     doNothing().when(mockShutdownHookCallback).run();
 
-    // setup shutdown hook
+    // call beforeStart to setup shutdownHook
```
cameronlee314: I think the timeout is actually set to 1ms in the setup, so this still might trigger the timeout. Could you set the timeout to something bigger? Then at least some could manually see the log to verify that the timeout did not get hit.

samza-core/src/test/java/org/apache/samza/runtime/TestClusterBasedProcessorLifecycleListener.java
abhishekshivanna: Yes, that is correct. Sorry I didn't make it clear.

samza-core/src/test/java/org/apache/samza/runtime/TestClusterBasedProcessorLifecycleListener.java
abhishekshivanna: Ah your right ! I moved the `afterStop()` into the callback so we can verify the invocation.

PR #1160, opened 2019-09-17
Transactional State [1/5]: Added Store Checkpoint API

prateekm:
This PR adds a new 'checkpoint' API to KeyValueStore that allows taking persistent (on-disk) snapshots of the local store state during commits for implementations that support it.

In RocksDB, this creates a new [Checkpoint](https://github.com/facebook/rocksdb/wiki/Checkpoints) directory. 

For supporting transactional state, during a commit, the newest changelog SSP offset is stored in the OFFSET file in this checkpoint directory, as well as in the checkpoint topic. During container start / store restore, the checkpoint directory corresponding to the changelog offset in the checkpoint topic is selected (if any) and used as the basis for restore.

Thanks to @xinyuiscool for contributing this.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/storage/kv/KeyValueStore.java
+++ b/samza-api/src/main/java/org/apache/samza/storage/kv/KeyValueStore.java
@@ -19,9 +19,12 @@
 
 package org.apache.samza.storage.kv;
 
+import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+
 
 /**
  * A key-value store that supports put, get, delete, and range queries.
@@ -141,4 +144,9 @@ public interface KeyValueStore<K, V> {
    * Flushes this key-value store, if applicable.
    */
   void flush();
+
+  /**
+   * Checkpoint the store snapshot.
+   */
+  Optional<Path> checkpoint(String id);
 }
--- a/samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
+++ b/samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
@@ -20,13 +20,16 @@
 package org.apache.samza.storage.kv
 
 import java.io.File
+import java.nio.file.{Path, Paths}
 import java.util
-import java.util.Comparator
+import java.util.{Comparator, Optional}
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.locks.ReentrantReadWriteLock
 
-import org.apache.samza.SamzaException
+import org.apache.commons.io.FileUtils
+import org.apache.samza.{SamzaException, checkpoint}
 import org.apache.samza.config.Config
+import org.apache.samza.serializers.CheckpointSerde
 import org.apache.samza.util.Logging
 import org.rocksdb.{TtlDB, _}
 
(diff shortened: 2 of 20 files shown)


samza-api/src/main/java/org/apache/samza/storage/kv/KeyValueStore.java
```diff
@@ -141,4 +144,9 @@ default void deleteAll(List<K> keys) {
    * Flushes this key-value store, if applicable.
    */
   void flush();
+
+  /**
+   * Checkpoint the store snapshot.
+   */
+  Optional<Path> checkpoint(String id);
```
bharathkumarasubramanian: 1. Does absence of `Path` here denote a no-op / not-implemented checkpoint or could it be the case where implementers choose to return empty optional on some failure scenario? Does that distinction matter upstream?
2. Should we consider returning a complex object w/ status, result and exception since this might be running on a separate thread (for async commit) and exception handling and bubbling exception will be easier and callers can make that choice. what do you think?

samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
```diff
@@ -236,6 +239,13 @@ class RocksDbKeyValueStore(
     trace("Flushed store: %s" format storeName)
   }
 
+  override def checkpoint(id: String): Optional[Path] = {
+    val checkpoint = Checkpoint.create(db)
+    val checkpointPath = dir.getPath + "-" + id
```
bharathkumarasubramanian: What happens if the path already exists? Do snapshots get overwritten or does it create unique checkpoint file name every time checkpoint is created?

prateekm: RocksDB checkpoint dir is required to be unique. Otherwise RocksDB will throw: `org.rocksdb.RocksDBException: Directory exists`

In part 5, id will be created using (current time millis + last 3 digit of nanotime), so the directory will typically be unique.

samza-api/src/main/java/org/apache/samza/storage/kv/KeyValueStore.java
prateekm: 1. If None, changelog offsets will not be written to the store (e.g. for persistent stores that don't support checkpoints) (PR #1164 , TransactionalStateTaskStorageManager). If exception, container will shut down. I'll update the javadocs in a follow up documentation PR after #1164. I'll also add a test in #1164 to ensure that any exception creating store checkpoint is propagated in TransactionalStateTaskStorageManager. There's already a test that ensure any exception thrown in TaskInstance#commit causes is propagated.

2. Transactional state is currently not supported for async commits. I'll mark this method as unstable so that we can evolve it if necessary. The behavior is: if checkpoints supported, return a path, else return None. On errors, throw an exception. Not sure if separate status, result and exception will help here?

prateekm: Ideally we'd make this a default method with the default impl returning an empty optional. However because of the way store wrapping works (cached store, logged store etc.), the correct default behavior for wrapper stores is to call the underlying store. Given the risk of accidentally forgetting to call the underlying store checkpoint in any new wrapper implementations if we make this a default method, I think it'll be better to keep this as a new required method.

mynameborat: Sure, please you make sure we throw an exception if transactional state is enabled w/ async commits.

prateekm: Created SAMZA-2336 to track.

PR #1161, opened 2019-09-17
Transactional State [2/5]: Added a ChangelogSSPIterator API

prateekm:
This PR changes the KeyValueStorageEngine restore API and implementation to take a new ChangelogSSPIterator instead of a plain Iterator.

The new ChangelogSSPIterator is similar to the existing SystemStreamPartitionIterator with the following differences: 
1. This new iterator has 2 modes: `Restore` and `Trim`. 
2. It takes an `endingOffset` during construction. The mode changes from `Restore` to `Trim` during iteration if the current message offset is greater than the ending offset (and trim mode is enabled).
3. Does not check for end of stream since it isn't applicable to changelog topics.

For supporting transactional state, we only restore changelog messages up to the changelog SSP offset in the checkpoint topic. Any messages after this 'checkpointed changelog offset' are trimmed by overwriting them with the current store value. When used in conjunction with an appropriate 'min.compaction.lag.ms' configuration for the Kafka changelog topic, this ensures that on container restart any store contents are consistent with the last input checkpoints and do not reflect any newer changes.

### DIFF
--- a/samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
+++ b/samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
@@ -20,16 +20,14 @@
 package org.apache.samza.storage.kv
 
 import java.io.File
-import java.nio.file.Path
-import java.util.Optional
 
 import org.apache.samza.util.Logging
 import org.apache.samza.storage.{StorageEngine, StoreProperties}
-import org.apache.samza.system.{IncomingMessageEnvelope, SystemStreamPartition}
+import org.apache.samza.system.{ChangelogSSPIterator, OutgoingMessageEnvelope, SystemStreamPartition}
 import org.apache.samza.task.MessageCollector
 import org.apache.samza.util.TimerUtil
-
-import scala.collection.JavaConverters._
+import java.nio.file.Path
+import java.util.Optional
 
 /**
  * A key value store.
@@ -47,8 +45,6 @@ class KeyValueStorageEngine[K, V](
   metrics: KeyValueStorageEngineMetrics = new KeyValueStorageEngineMetrics,
   batchSize: Int = 500,
   val clock: () => Long = { System.nanoTime }) extends StorageEngine with KeyValueStore[K, V] with TimerUtil with Logging {
-
-  var count = 0
 
   /* delegate to underlying store */
   def get(key: K): V = {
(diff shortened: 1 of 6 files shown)


samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
```diff
@@ -109,42 +105,89 @@ class KeyValueStorageEngine[K, V](
...
-        doPutAll(rawStore, batch)
-        batch.clear()
-      }
-
-      if (valBytes != null) {
-        metrics.restoredBytesGauge.set(metrics.restoredBytesGauge.getValue + valBytes.length)
+      val mode = iterator.getMode
+
+      if (mode.equals(ChangelogSSPIterator.Mode.RESTORE)) {
+        batch.add(new Entry(keyBytes, valBytes))
```
bharathkumarasubramanian: Should we throw an exception here if the iterator transitioned to `RESTORE` after being in `TRIM` or is it something we can guarantee as part of the `ChangelogIterator`?

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
```diff
@@ -109,42 +105,89 @@ class KeyValueStorageEngine[K, V](
...
+        // log progress every million messages
+        if (restoredMessages % 1000000 == 0) {
+          info(restoredMessages + " entries restored for store: " + storeName + " in directory: " + storeDir.toString + "...")
+        }
+      } else {
+        // first write any open restore batches to store
+        if (!lastBatchFlushed) {
+          info(restoredMessages + " total entries restored for store: " + storeName + " in directory: " + storeDir.toString + ".")
+          if (batch.size > 0) {
+            doPutAll(rawStore, batch)
```
bharathkumarasubramanian: do we need to clear the batch here just to be safe?

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
```diff
@@ -109,42 +105,89 @@ class KeyValueStorageEngine[K, V](
...
+      lastBatchFlushed = true
     }
+    info(restoredMessages + " entries trimmed for store: " + storeName + " in directory: " + storeDir.toString + ".")
 
-    info(count + " total entries restored for store: " + storeName + " in directory: " + storeDir.toString + ".")
-
-    if (batch.size > 0) {
-      doPutAll(rawStore, batch)
-    }
+    // flush the store and the changelog producer
```
bharathkumarasubramanian: This isn't required for old code path. However, should this be a `blocker` instead of `high` for new code path?

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestKeyValueStorageEngine.scala
```diff
@@ -142,12 +142,22 @@ class TestKeyValueStorageEngine {
   @Test
   def testRestoreMetrics(): Unit = {
```
bharathkumarasubramanian: maybe add a test case to ensure the transition from `TRIM` to `RESTORE` doesn't happen.

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
prateekm: We do a task commit after the restore is complete and before we start processing. That will flush all producers (incl. changelog producer). So not a blocker.

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
prateekm: Will fix in PR 5/5.

samza-kv/src/main/scala/org/apache/samza/storage/kv/KeyValueStorageEngine.scala
prateekm: Will update in PR 5.

samza-kv/src/test/scala/org/apache/samza/storage/kv/TestKeyValueStorageEngine.scala
prateekm: Will add in PR 5.

PR #1164, opened 2019-09-17
Transactional State [5/5]: Added implementations for transactional state checkpoints and restore

prateekm:
Adds implementation and tests for transactional state checkpoint and restore paths.

Changes the existing Task commit path to create and clean up store checkpoint directories, and write the newest changelog offset to the checkpoint.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/storage/TransactionalStateTaskRestoreManager.java
+++ b/samza-core/src/main/java/org/apache/samza/storage/TransactionalStateTaskRestoreManager.java
@@ -406,7 +406,7 @@ public class TransactionalStateTaskRestoreManager implements TaskRestoreManager 
             loggedStoreBaseDirectory, storeName, taskName, taskMode);
         LOG.info("Moving logged store checkpoint directory: {} for store: {} in task: {} to current directory: {}",
             storeDirsToRetain.toString(), storeName, taskName, currentDir);
-        storageManagerUtil.moveCheckpointFiles(storeDirToRetain, currentDir);
+        storageManagerUtil.restoreCheckpointFiles(storeDirToRetain, currentDir);
         // do not remove the checkpoint directory yet. in case commit fails and container restarts,
         // we can retry the move. if we delete the checkpoint, the current dir will be deleted as well on
         // restart, and we will have to do a full restore.
--- a/samza-core/src/main/java/org/apache/samza/storage/TransactionalStateTaskRestoreManager.java
+++ b/samza-core/src/main/java/org/apache/samza/storage/TransactionalStateTaskRestoreManager.java
@@ -216,6 +216,8 @@ public class TransactionalStateTaskRestoreManager implements TaskRestoreManager 
             !storageEngine.getStoreProperties().isLoggedStore()) {
           File currentDir = storageManagerUtil.getTaskStoreDir(
               nonLoggedStoreBaseDirectory, storeName, taskName, taskMode);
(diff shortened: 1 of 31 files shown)


samza-core/src/main/java/org/apache/samza/storage/TransactionalStateTaskRestoreManager.java
```diff
@@ -0,0 +1,525 @@
...
+            // delete all non-retained checkpoint directories
+            for (File checkpointDir: checkpointDirsOptional.get()) {
+              if (storeDirToRetain.get(storeName) == null ||
+                  !storeDirToRetain.get(storeName).equals(checkpointDir)) {
+                storeDirsToDelete.put(storeName, checkpointDir);
+              }
+            }
+
+            // if the store had not valid checkpoint dirs to retain, restore from changelog
+            if (!hasValidCheckpointDir) {
```
xinyuiscool: So for the job which just turned on this feature, there will not be any checkpoint store folders. Then what's the behavior? Will it bootstrap the changelog?

samza-core/src/main/java/org/apache/samza/storage/TransactionalStateTaskRestoreManager.java
```diff
@@ -0,0 +1,525 @@
...
+
+        LOG.info("For store: {} in task: {} got current dir: {}, checkpoint dirs: {}, checkpointed changelog offset: {}",
+            storeName, taskName, currentDirOptional, checkpointDirsOptional, checkpointedOffset);
+
+        // TODO BLOCKER pmaheshw: will do full restore from changelog even if retain existing state == true
+        // always delete current logged store dir for persistent stores.
+        currentDirOptional.ifPresent(currentDir -> storeDirsToDelete.put(storeName, currentDir));
+
+        // first check if checkpointed offset is invalid (i.e., out of range of current offsets, or null)
+        if (checkpointedOffset == null && oldestOffset != null) {
```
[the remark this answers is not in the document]
prateekm: Just FYI, current KafkaSystemAdmin does not follow the SSPMetadata interface, and returns oldest offset == 0 for an empty topic.

To your point, if you change the changelog topic for a store to an empty topic, this is the behavior:
1. Checkpointed offset will be null, since the checkpoint will not have an offset for the new changelog SSP.
2. Oldest offset may be non-zero, signifying that topic contains data, or null / 0, signifying that the topic is empty.
3. If oldest is non-zero, behavior depends on the retain.existing.changelog.state config. I renamed the config to clarify that if true, this will overwrite local state with the changelog topic state. This is also the case when you turn on the config for the first time, without writing changelog offests in checkpoints first.
4. If oldest is null or 0, it will fall through to the 'happy path', delete all local state, and do a full restore, which in this case will do nothing, which is the desired behavior.

I'll add an explicit unit test to cover the null oldest offset scenario.

samza-core/src/main/java/org/apache/samza/storage/TransactionalStateTaskRestoreManager.java
prateekm: As discussed offline, we will make the following changes:
1. In the next release, stores will create the checkpoint directories and write the changelog offset in the checkpoint regardless of whether the transactional state feature is turned on or not.
2. We will recommend that users first upgrade and deploy the new version with the feature turned off, and then after some time (at least one commit), turn the feature on.
This way we will have a checkpointed changelog offset + a store checkpoint directory available when the feature is turned on.
3. We will backport the change to handle null changelog offsets in the checkpoint (in CheckpointSerde) to version 1.2, to support rolling back to a previous version.

PR #1168, opened 2019-09-24
Samza-2331: Fix java doc of table descriptor for default rate limiter

lhaiesp:
Existing java doc is confusing about what the actual behavior is for the default rate limiter. We should explicitly note that it is for the entire job and per task credit needs to divide that by number of tasks.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/util/EmbeddedTaggedRateLimiter.java
+++ b/samza-core/src/main/java/org/apache/samza/util/EmbeddedTaggedRateLimiter.java
@@ -115,6 +115,8 @@ public class EmbeddedTaggedRateLimiter implements RateLimiter {
             int numTasks = jobModel.getContainers().values().stream()
                 .mapToInt(cm -> cm.getTasks().size())
                 .sum();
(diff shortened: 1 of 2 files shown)


samza-core/src/main/java/org/apache/samza/util/EmbeddedTaggedRateLimiter.java
```diff
@@ -115,6 +115,8 @@ public void init(Context context) {
             int numTasks = jobModel.getContainers().values().stream()
                 .mapToInt(cm -> cm.getTasks().size())
                 .sum();
+            Preconditions.checkArgument(e.getValue() >= numTasks,
```
dengpanyin: Shall we just warn the user instead of throw exception for this case?
Also change line 120 to be:
double effectiveRate = (double) e.getValue() / numTasks;

lhaiesp: Make sense. Turning the calculation into double is the real fix.

dengpanyin:
LGTM 
Thanks for the change!

PR #1171, opened 2019-09-26
SAMZA-2328: StreamConfig from scala to Java

sborya:
Convert StreamConfig from scala to java and update throughout the code.


samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -292,7 +292,8 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
   // The method returns a map of storenames to changelog topic names, which are configured to use kafka as the changelog stream
   def getKafkaChangelogEnabledStores() = {
```
cameronlee314: Could you add a return value to this method to clarify that it is returning a `util.HashMap` (as opposed to a scala `Map`)?

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.java
```diff
@@ -0,0 +1,394 @@
...
+  public static final String CONSUMER_RESET_OFFSET = SAMZA_PROPERTY + "reset.offset";
+  public static final String CONSUMER_OFFSET_DEFAULT = SAMZA_PROPERTY + "offset.default";
+  public static final String BOOTSTRAP = SAMZA_PROPERTY + "bootstrap";
+  public static final String PRIORITY = SAMZA_PROPERTY + "priority";
+  public static final String IS_INTERMEDIATE = SAMZA_PROPERTY + "intermediate";
+  public static final String DELETE_COMMITTED_MESSAGES = SAMZA_PROPERTY + "delete.committed.messages";
+  public static final String IS_BOUNDED = SAMZA_PROPERTY + "bounded";
+  public static final String BROADCAST = SAMZA_PROPERTY + "broadcast";
+
+  // We don't want any external dependencies on these patterns while both exist. Use getProperty to ensure proper values.
```
cameronlee314: Looks like this was an existing comment, but `getProperty` isn't actually a method anymore in this class. Maybe change to `getSamzaProperty`?

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.java
```diff
@@ -0,0 +1,394 @@
...
+  public static final String IS_INTERMEDIATE_FOR_STREAM_ID = STREAM_ID_PREFIX + IS_INTERMEDIATE;
+  public static final String DELETE_COMMITTED_MESSAGES_FOR_STREAM_ID = STREAM_ID_PREFIX + DELETE_COMMITTED_MESSAGES;
+  public static final String IS_BOUNDED_FOR_STREAM_ID = STREAM_ID_PREFIX + IS_BOUNDED;
+  public static final String PRIORITY_FOR_STREAM_ID = STREAM_ID_PREFIX + PRIORITY;
+  public static final String CONSUMER_OFFSET_DEFAULT_FOR_STREAM_ID = STREAM_ID_PREFIX + CONSUMER_OFFSET_DEFAULT;
+  public static final String BOOTSTRAP_FOR_STREAM_ID = STREAM_ID_PREFIX + BOOTSTRAP;
+  public static final String BROADCAST_FOR_STREAM_ID = STREAM_ID_PREFIX + BROADCAST;
+
+  /**
+   * Helper for accessing configs related to stream properties.
```
cameronlee314: Do you think this javadoc could go at the class level instead of inside here? Or do you think this is more of an internal implementation javadoc?

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.java
```diff
@@ -0,0 +1,394 @@
...
+   * This finds the stream id that is mapped to the system in systemStream through the config and that has a physical
+   * name (the physical name might be the stream id itself if there is no explicit mapping) that matches the stream in
+   * systemStream.
+   * Note: If the stream in the systemStream is a stream id which is mapped to a physical stream, then that stream won't
+   * be returned as a stream id here, since the stream in systemStream doesn't match the physical stream name.
+   *
+   * @param systemStream system stream to map to stream id
+   * @return             stream id corresponding to the system stream
+   */
+  public String systemStreamToStreamId(SystemStream systemStream) {
```
cameronlee314: Can this be private? It was private before.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.java
```diff
@@ -0,0 +1,394 @@
...
+  /**
+   * Gets the System associated with the specified streamId.
+   * It first looks for the property
+   * streams.{streamId}.system
+   * <p>
+   * If no value was provided, it uses
+   * job.default.system
+   *
+   * @param streamId the identifier for the stream in the config.
+   * @return the system name associated with the stream.
```
cameronlee314: Could you update this javadoc to specify that it might also return null?
I actually think there are some callers of `getSystem` which don't really check for null cleanly, but probably not in scope to change that in this PR.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.java
```diff
@@ -0,0 +1,394 @@
...
+   */
+  public String getSystem(String streamId) {
+    return getOrDefault(String.format(SYSTEM_FOR_STREAM_ID, streamId),
+      new JobConfig(this).getDefaultSystem().orElse(null));
```
cameronlee314: This is going to always compute `new JobConfig(this).getDefaultSystem().orElse(null)`, even if the value isn't used. Not a big deal, but the `JobConfig` constructor does make a new copy of the config map, so maybe would be nicer to lazily evaluate `new JobConfig(this).getDefaultSystem().orElse(null)`.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.java
```diff
@@ -0,0 +1,394 @@
...
+   * 2) Otherwise, look for "systems.{systemName}.streams.{streamName}.{property}" in which the systemName is the system
+   * mapped to the stream id and the streamName is the physical stream name for the stream id.
+   * 3) Otherwise, look for "systems.{systemName}.default.stream.{property}" in which the systemName is the system
+   * mapped to the stream id.
+   * If a stream id was not found or no property could be found using the above keys:
+   * 1) Look for "systems.{systemName}.streams.{streamName}.{property}" in which the systemName is the system in the
+   * input systemStream and the streamName is the stream from the input systemStream.
+   * 2) Otherwise, look for "systems.{systemName}.default.stream.{property}" in which the systemName is the system
+   * in the input systemStream.
+   *
```
cameronlee314: I forgot to add that this can return `null`. Could you please add that?

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.java
```diff
@@ -0,0 +1,394 @@
...
+   * Gets a Samza property, with a default value used if no property value is found.
+   * See getSamzaProperty(SystemStream, String).
+   *
+   * @param systemStream the SystemStream for which the property value will be retrieved.
+   * @param property the samza property key (including the "samza." prefix); for example, for both
+   *                 "streams.streamId.samza.prop.key" and "systems.system.streams.streamName.samza.prop.key", this
+   *                 argument should have the value "samza.prop.key"
+   * @param defaultValue the default value to use if the property value is not found
+   */
+  protected String getSamzaProperty(SystemStream systemStream, String property, String defaultValue) {
```
cameronlee314: Can `getSamzaProperty` (here and below) be private?
Could you please move these helpers further below in the class file so that they are not in between some of the public methods?

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.java
```diff
@@ -0,0 +1,394 @@
...
+  public boolean isDefaultStreamOffsetConfigured(SystemStream systemStream) {
+    return containsSamzaProperty(systemStream, CONSUMER_OFFSET_DEFAULT);
+  }
+
+  public boolean getBootstrapEnabled(SystemStream systemStream) {
+    return Boolean.parseBoolean(getSamzaProperty(systemStream, BOOTSTRAP));
+  }
+
+  public boolean getBroadcastEnabled(SystemStream systemStream) {
+    return java.lang.Boolean.parseBoolean(getSamzaProperty(systemStream, BROADCAST));
```
cameronlee314: Minor: `java.lang.` unnecessary here

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.java
sborya: changed the comment to 'use the corresponding get*() method'

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
sborya: he he. Intelij shows it :) in grayed out script. I will add. Makes sense.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.java
sborya: I feel this is more of implementation thing.

cameronlee314: This first line in the javadoc block might be slightly confusing then. Maybe just replace with "Implementation notes" (or something like that).
Also, can you replace the `/**` with `/*`? With the `/**`, this javadoc is associated with the method below (i.e. `nonEmptyOption`), which would be confusing.

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.java
sborya: you mean in the comment? As 3) return null.

cameronlee314: Yep

samza-core/src/main/scala/org/apache/samza/config/StreamConfig.java
sborya: agree. Done.

PR #1172, opened 2019-09-27
SAMZA-2332: [AM isolation] YarnJob should pass new command and additional environment variables for AM deployment

cameronlee314:


samza-yarn/src/test/java/org/apache/samza/job/yarn/TestYarnJob.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.samza.config.MapConfig;
+import org.apache.samza.config.ShellCommandConfig;
+import org.apache.samza.config.YarnConfig;
+import org.apache.samza.serializers.model.SamzaObjectMapper;
+import org.apache.samza.util.CoordinatorStreamUtil;
+import org.apache.samza.util.Util;
+import org.junit.Test;
+import scala.collection.JavaConverters;
+
+import static org.junit.Assert.*;
```
cameronlee314: This should not use wildcard.

samza-yarn/src/main/scala/org/apache/samza/job/yarn/YarnJob.scala
```diff
@@ -200,3 +167,63 @@ class YarnJob(config: Config, hadoopConfig: Configuration) extends StreamJob {
...
+  }
+
+  /**
+    * Build the command for the application master execution.
+    * Passing multiple separate config objects so that they can be reused in other places.
+    */
+  @VisibleForTesting
+  private[yarn] def buildAmCmd(config: Config, jobConfig: JobConfig): String = {
+    // figure out if we have framework is deployed into a separate location
+    val fwkPath = config.get(JobConfig.SAMZA_FWK_PATH, "")
```
prateekm: Is this code still relevant? IIRC, this is from the previous (abandoned) attempt at split deployment.

samza-core/src/main/java/org/apache/samza/classloader/IsolationUtils.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class IsolationUtils {
+  /**
+   * TODO make this configurable or taken from an environment variable
+   */
+  public static final String APPLICATION_MASTER_API_DIRECTORY = "__samzaFrameworkApi";
+  /**
+   * TODO make this configurable or taken from an environment variable
+   */
+  public static final String APPLICATION_MASTER_INFRASTRUCTURE_DIRECTORY = "__samzaFrameworkInfrastructure";
```
prateekm: Can you clarify (in docs) what these three directories are?

samza-core/src/main/scala/org/apache/samza/config/ShellCommandConfig.scala
```diff
@@ -50,6 +50,17 @@ object ShellCommandConfig {
...
   val ENV_EXECUTION_ENV_CONTAINER_ID = "EXECUTION_ENV_CONTAINER_ID"
 
+  /**
+   * Set to "true" if application master isolation is enabled. Otherwise, will be considered false.
+   */
+  val ENV_APPLICATION_MASTER_ISOLATION_ENABLED = "APPLICATION_MASTER_ISOLATION_ENABLED"
+
+  /**
+   * Can be set to a directory which is different from the base lib directory which contains some additional
+   * application-specific resources.
```
prateekm: Can you clarify (in docs) what this env variable is used for? E.g., what's a 'base lib directory', when would you set it to a non-'base lib directory', and what additional application-specific resources? Also, why would you do this?

samza-yarn/src/main/scala/org/apache/samza/job/yarn/YarnJob.scala
cameronlee314: It is not relevant. I will put out a separate PR to remove the old code and then apply this change on top of that.

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
```diff
@@ -128,6 +128,9 @@
   public static final String COORDINATOR_STREAM_FACTORY = "job.coordinatorstream.config.factory";
   public static final String DEFAULT_COORDINATOR_STREAM_CONFIG_FACTORY = "org.apache.samza.util.DefaultCoordinatorStreamConfigFactory";
 
+  // enables application master isolation
+  public static final String SAMZA_APPLICATION_MASTER_ISOLATION_ENABLED = "samza.application.master.isolation.enabled";
```
abhishekshivanna: Maybe we should include the word `dependency` in the config name while referring to isolation ?
Also do you think we should maybe rename this config to be more generic like
`samza.cluster.based.job.coordinator.dependency.isolation.enabled` ?
If not, maybe move it to `YarnConfig.java` since this is YARN specific ?

samza-core/src/main/scala/org/apache/samza/config/ShellCommandConfig.scala
```diff
@@ -50,6 +50,17 @@ object ShellCommandConfig {
     */
   val ENV_EXECUTION_ENV_CONTAINER_ID = "EXECUTION_ENV_CONTAINER_ID"
 
+  /**
+   * Set to "true" if application master isolation is enabled. Otherwise, will be considered false.
+   */
+  val ENV_APPLICATION_MASTER_ISOLATION_ENABLED = "APPLICATION_MASTER_ISOLATION_ENABLED"
```
abhishekshivanna: Similar to above comment regarding making this less YARN specific.

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
cameronlee314: Good point. I'll use the "cluster based job coordinator dependency" terminology.

prateekm:
@shanthoosh FYI

cameronlee314:
> @shanthoosh FYI

Just to make sure we are on the same page: The changes for SQL aren't going to immediately leverage general split deployment, because SQL does not want to make general split deployment a dependency right now. Eventually, SQL would converge to using general split deployment, but not right now.

prateekm:
@cameronlee314 Yup. @shanthoosh had a question about whether AM split deploy will continue working when we move planning to the AM, or if there would be further changes required.

PR #1173, opened 2019-09-27
SAMZA-2333: [AM isolation] Use cytodynamics classloader to launch job coordinator

cameronlee314:
This change uses cytodynamics to load Samza framework code from a different package than the Samza application. This allows dependencies to be isolated between framework code and application code. This flow is only enabled by setting the config `samza.cluster.based.job.coordinator.dependency.isolation.enabled`.

Made the following changes to samza-hello-samza to test this:
1) Exclude log4j, depend on log4j2 (and add log4j2.xml)
2) Explicitly specify YARN resource path to locally built artifact for Samza framework code
3) Add custom log4j2 appender which writes logs to stdout with a prefix


gradle/dependency-versions.gradle
```diff
@@ -25,6 +25,7 @@
   commonsCollectionVersion = "3.2.1"
   commonsHttpClientVersion = "3.1"
   commonsLang3Version = "3.4"
+  cytodynamicsVersion = "0.1.1-SNAPSHOT"
```
cameronlee314: Leaving a reminder: This needs to be replaced by a real version of cytodynamics before checking this in.

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
```diff
@@ -0,0 +1,195 @@
...
+        .withClasspath(getClasspathAsURIs(infrastructureLibDirectory))
+        // getClasspathAsURIs should already satisfy this, but doing it to be safe
+        .withOriginRestriction(OriginRestriction.denyByDefault().allowingDirectory(infrastructureLibDirectory, false))
+        .withParentRelationship(buildApiParentRelationship(apiLibDirectory, apiClassLoader))
+        // need to be able to fall back to application for loading certain classes
+        .addFallbackDelegate(buildFallbackDelegateRelationship(applicationClassLoader))
+        .build();
+  }
+
+  private static DelegateRelationship buildApiParentRelationship(File apiLibDirectory, ClassLoader apiClassLoader) {
```
prateekm: Can you add javadoc explaining what this method does. E.g., `Sets up classloader relationships for the API classloader such that ... bootstrap classes (e.g. java.lang Classes) are loaded from ... and api classes (as specified by classes listed in file ....) are loaded from ...` or something to that effect.

prateekm: Same for the buildApplicationClassLoader method and buildInfrastructureClassLoader method. Would be good to explain what a `parentRelationship` entails, e.g., who is parent of whom, and what are the implications for preference when loading classes. The API naming doesn't make it clear enough.

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
```diff
@@ -0,0 +1,195 @@
...
+        // getClasspathAsURIs should already satisfy this, but doing it to be safe
+        .withOriginRestriction(OriginRestriction.denyByDefault().allowingDirectory(applicationLibDirectory, false))
+        .withParentRelationship(buildApiParentRelationship(apiLibDirectory, apiClassLoader))
+        .build();
+  }
+
+  private static ClassLoader buildInfrastructureClassLoader(File infrastructureLibDirectory, File apiLibDirectory,
+      ClassLoader apiClassLoader, ClassLoader applicationClassLoader) {
+    return LoaderBuilder.anIsolatingLoader()
+        // look in application directory for JARs
```
prateekm: Did you mean 'in infrastructure lib directory'?

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
prateekm: Would be even better to include an ascii diagram explaining the relationship (or a link to one in the SEP).

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
```diff
@@ -0,0 +1,195 @@
...
+        .withClasspath(getClasspathAsURIs(infrastructureLibDirectory))
+        // getClasspathAsURIs should already satisfy this, but doing it to be safe
+        .withOriginRestriction(OriginRestriction.denyByDefault().allowingDirectory(infrastructureLibDirectory, false))
+        .withParentRelationship(buildApiParentRelationship(apiLibDirectory, apiClassLoader))
+        // need to be able to fall back to application for loading certain classes
```
prateekm: Why is this required / for which classes?

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
```diff
@@ -0,0 +1,195 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Use this to build an isolating classloader for running Samza.
+ */
+public class IsolatingClassLoaderFactory {
+  private static final Logger LOG = LoggerFactory.getLogger(IsolatingClassLoaderFactory.class);
+
+  public ClassLoader buildClassLoader() {
```
prateekm: Do you need to create a singleton or is it OK to call this multiple times?

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -406,10 +409,72 @@ ContainerProcessManager createContainerProcessManager(ClassLoader classLoader) {
...
+      isolatedClusterBasedJobCoordinatorClass =
+          isolatedClassLoader.loadClass(ClusterBasedJobCoordinator.class.getName());
+    } catch (ClassNotFoundException e) {
+      throw new SamzaException(
+          "Isolation was enabled, but unable to find ClusterBasedJobCoordinator in isolated classloader", e);
+    }
+
+    // save the current context classloader so it can be reset after finishing the call to runClusterBasedJobCoordinator
+    ClassLoader previousContextClassLoader = Thread.currentThread().getContextClassLoader();
+    // this is needed because certain libraries use the context classloader
```
prateekm: Any examples of these libraries?

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -406,10 +409,72 @@ ContainerProcessManager createContainerProcessManager(ClassLoader classLoader) {
...
+    Thread.currentThread().setContextClassLoader(isolatedClassLoader);
+
+    try {
+      executeRunClusterBasedJobCoordinatorForClass(isolatedClusterBasedJobCoordinatorClass, args);
+    } finally {
+      // reset the context class loader
```
prateekm: Just wondering, why is this required? Best practice or something breaks / can break if you don't do this?

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -406,10 +409,72 @@ ContainerProcessManager createContainerProcessManager(ClassLoader classLoader) {
...
+      Thread.currentThread().setContextClassLoader(previousContextClassLoader);
+    }
+  }
+
+  private static void executeRunClusterBasedJobCoordinatorForClass(Class<?> isolatedClusterBasedJobCoordinatorClass,
+      String[] args) {
+    Method runClusterBasedJobCoordinatorMethod;
+    try {
+      runClusterBasedJobCoordinatorMethod =
+          isolatedClusterBasedJobCoordinatorClass.getDeclaredMethod("runClusterBasedJobCoordinator", String[].class);
```
prateekm: Would be good to add a unit test that does this too, so that it'll break if someone changes the method name but not this.

samza-shell/src/main/bash/run-class.sh
```diff
@@ -128,11 +137,28 @@ function check_and_enable_64_bit_mode {
...
-    [[ $JAVA_OPTS != *-Dlog4j.configurationFile* ]] && export JAVA_OPTS="$JAVA_OPTS -Dlog4j.configurationFile=file:$DEFAULT_LOG4J2_FILE"
-elif [[ -n $(find "$base_dir/lib" -regex ".*samza-log4j.*.jar*") ]]; then
-    [[ $JAVA_OPTS != *-Dlog4j.configuration* ]] && export JAVA_OPTS="$JAVA_OPTS -Dlog4j.configuration=file:$DEFAULT_LOG4J_FILE"
+# Check if log4j configuration is specified; if not, look for a configuration file:
+# 1) Check if using log4j or log4j2
+# 2) Check if configuration file system property is already set
+# 3) If not, then look in $APPLICATION_LIB_DIR for configuration file (remember that $APPLICATION_LIB_DIR can be same or
+#    different from $BASE_LIB_DIR).
+# 4) If still can't find it, fall back to default (from $BASE_LIB_DIR).
+if [[ -n $(find "$BASE_LIB_DIR" -regex ".*samza-log4j2.*.jar*") ]]; then
```
prateekm: I'll let @abhishekshivanna review this.

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
```diff
@@ -0,0 +1,195 @@
...
+    return DelegateRelationshipBuilder.builder()
+        .withDelegateClassLoader(fallbackClassLoader)
+        // want to load any class from the fallback
+        .withIsolationLevel(IsolationLevel.NONE)
+        .build();
+  }
+
+  @VisibleForTesting
+  static List<String> getApiClasses(File directoryWithClassList) {
+    File parentPreferredFile = new File(directoryWithClassList, "application-master-samza-api-class-list.txt");
```
prateekm: What is this file name? Maybe extract constant (to IsolationUtil) and document how/when this is generated.

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
```diff
@@ -0,0 +1,195 @@
...
+    validateCanAccess(parentPreferredFile);
+    try {
+      return Files.readAllLines(Paths.get(parentPreferredFile.toURI()), StandardCharsets.UTF_8);
+    } catch (IOException e) {
+      throw new SamzaException("Error while reading samza-api class list", e);
+    }
+  }
+
+  /**
+   * @param jarsLocation directory where JARs are located
```
prateekm: Nitpick: Need description for javadocs.

gradle/dependency-versions.gradle
prateekm: I think you'll also need to add it to the LICENSE file.

samza-shell/src/main/bash/run-class.sh
abhishekshivanna: LGTM

gradle/dependency-versions.gradle
cameronlee314: We don't directly bundle anything from cytodynamics (BSD2 license) inside samza. We just depend on it. Do we need to add anything? I looked at the LICENSE file, and it seems to just include certain bundled components (I think for the website?). However, it doesn't seem to have anything about libraries (e.g. kafka, apache commons, guava, etc.).

prateekm:
@shanthoosh FYI

samza-core/src/main/java/org/apache/samza/classloader/DependencyIsolationUtils.java
```diff
@@ -25,16 +25,24 @@
...
    * placed, for usage in dependency isolation for the cluster-based job coordinator.
    * TODO make this configurable or taken from an environment variable
    */
   public static final String APPLICATION_DIRECTORY = "__package";
+
+  /**
+   * Name of the file which contains the class names (or globs) which should be loaded from the framework API
+   * classloader.
+   */
+  public static final String FRAMEWORK_API_CLASS_LIST_FILE_NAME = "samza-framework-api-class-list.txt";
```
prateekm: Nitpick: maybe samza-framework-api-classes.txt and same for the variable.

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
```diff
@@ -0,0 +1,316 @@
...
+ * Use this to build a classloader for running Samza which isolates the Samza framework code/dependencies from the
+ * application code/dependencies.
+ */
+public class IsolatingClassLoaderFactory {
+  private static final Logger LOG = LoggerFactory.getLogger(IsolatingClassLoaderFactory.class);
+
+  private static final String LIB_DIRECTORY = "lib";
+
+  /**
+   * Build a classloader which will isolate Samza framework code from application code. Samza framework classes and
```
prateekm: Nice, thanks for the thorough documentation.

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
prateekm: Is there a use case for having multiple instances of this classloader instead of a singleton? If not, wondering if having a singleton instance will be safer, to avoid someone accidentally creating an independent classloader. If this is not a likely risk, that's fine too, documenting the expectation might be sufficient.

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
```diff
@@ -0,0 +1,316 @@
...
+        .withClasspath(getClasspathAsURIs(applicationLibDirectory))
+        // getClasspathAsURIs should only return JARs within applicationLibDirectory anyways, but doing it to be safe
+        .withOriginRestriction(OriginRestriction.denyByDefault().allowingDirectory(applicationLibDirectory, false))
+        // delegate to the api classloader for API classes
+        .withParentRelationship(buildApiParentRelationship(apiLibDirectory, apiClassLoader))
+        .build();
+  }
+
+  /**
+   * Build the {@link ClassLoader} which can load Samza framework core classes.
```
prateekm: To clarify documentation, this can/will load application classes as well, right?

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
```diff
@@ -0,0 +1,316 @@
...
+    // bootstrap classes need to be loaded from a common classloader
+    apiParentRelationshipBuilder.addDelegatePreferredClassPredicate(new BootstrapClassPredicate());
+    // the classes which are Samza framework API classes are added here
+    getFrameworkApiClassGlobs(apiLibDirectory).forEach(
+        apiClassName -> apiParentRelationshipBuilder.addDelegatePreferredClassPredicate(new GlobMatcher(apiClassName)));
+    return apiParentRelationshipBuilder.build();
+  }
+
+  /**
+   * @param directoryWithClassList Directory in which
```
prateekm: Nitpick: Add description.

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
```diff
@@ -0,0 +1,316 @@
...
+    if (filesInJarsLocation == null) {
+      throw new SamzaException(
+          String.format("Could not find any files inside %s, probably because it is not a directory",
+              jarsLocation.getPath()));
+    }
+    URL[] urls = Stream.of(filesInJarsLocation)
+        .filter(file -> file.getName().endsWith(".jar") || file.getName().endsWith(".war"))
+        .map(IsolatingClassLoaderFactory::fileURL)
+        .toArray(URL[]::new);
+    LOG.info("Found {} items to load into classpath from {}", urls.length, jarsLocation);
```
prateekm: Might be useful  for debugging to log actual directory contents, either here or in buildClassLoader().

PR #1174, opened 2019-09-30
SAMZA-2335: Remove classloader argument from ReflectionUtil

cameronlee314:
SAMZA-2204 added a ReflectionUtil class for some reflection-based helpers, and SAMZA-2210 started to replace the usage of Util.getObj with those ReflectionUtil methods.
The intended purpose of those changes was to be able to explicitly choose a classloader for loading pluggable classes, but it is no longer necessary to explicitly pass it in. This ticket cleans up that explicit classloader argument.
We are able to use the default classloader propagation to properly pass around the classloader to use.


cameronlee314:
> A clarification question. Looks good otherwise.
> Does this mean there is no longer a need to have the ability to choose custom class loader or is the framework going to take care of making that choice? If it is former, did something change?

There is just no longer a need to pass the classloader explicitly. There is a way to use the default classloader propagation in Java to do what we need for using a custom classloader.

PR #1177, opened 2019-10-03
SAMZA-2339 : Update KafkaSystemAdmin to adhere to the checkpoint replication factor

VartulAgrawal:
As part of this PR: https://github.com/apache/samza/pull/789/files#diff-52d8a92e9c8e3d26c4e595919566326aR519 the replicationFactor value defined in the config was dropped since with the kafka client update it cannot be passed to the configs. 

To preserve this config it should be passed to the checkpoint topic spec; this PR adds the change to preserve the value set by the config `task.checkpoint.replication.factor` in the checkpoint kafka topic spec.

### DIFF
--- a/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
+++ b/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
@@ -540,6 +540,9 @@ public class KafkaSystemAdmin implements SystemAdmin {
       kafkaSpec =
           new KafkaStreamSpec(spec.getId(), spec.getPhysicalName(), systemName, 1, coordinatorStreamReplicationFactor,
               coordinatorStreamProperties);
+    } else if (spec.isCheckpointStream()) {
+      kafkaSpec = KafkaStreamSpec.fromSpec(StreamSpec.createCheckpointStreamSpec(spec.getPhysicalName(), systemName))
+              .copyWithReplicationFactor(Integer.parseInt(new KafkaConfig(config).getCheckpointReplicationFactor().get()));
     } else if (intermediateStreamProperties.containsKey(spec.getId())) {
       kafkaSpec = KafkaStreamSpec.fromSpec(spec);
       Properties properties = kafkaSpec.getProperties();
--- a/samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
+++ b/samza-api/src/main/java/org/apache/samza/system/StreamSpec.java
@@ -226,6 +226,10 @@ public class StreamSpec implements Serializable {
     return id.equals(COORDINATOR_STREAM_ID);
   }
 
+  public boolean isCheckpointStream() {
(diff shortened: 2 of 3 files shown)


mynameborat:
Thank you for your contribution :)
Can you please create a JIRA ticket and attach to this PR as per https://samza.apache.org/contribute/contributors-corner.html

VartulAgrawal:
Hi @mynameborat, I have already created this ticket and linked the PR to it: https://issues.apache.org/jira/browse/SAMZA-2339

mynameborat:
> Hi @mynameborat, I have already created this ticket and linked the PR to it: https://issues.apache.org/jira/browse/SAMZA-2339

oh I must missed the ticket then. Can you update the PR title to include ticket so that it automatically links and you don't have to manually do it?

VartulAgrawal:
> > Hi @mynameborat, I have already created this ticket and linked the PR to it: https://issues.apache.org/jira/browse/SAMZA-2339
> 
> oh I must missed the ticket then. Can you update the PR title to include ticket so that it automatically links and you don't have to manually do it?

Thank you for pointing that out :)
Updated the PR title 👍

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -540,6 +540,9 @@ public KafkaStreamSpec toKafkaSpec(StreamSpec spec) {
       kafkaSpec =
           new KafkaStreamSpec(spec.getId(), spec.getPhysicalName(), systemName, 1, coordinatorStreamReplicationFactor,
               coordinatorStreamProperties);
+    } else if (spec.isCheckpointStream()) {
+      kafkaSpec = KafkaStreamSpec.fromSpec(StreamSpec.createCheckpointStreamSpec(spec.getPhysicalName(), systemName))
+              .copyWithReplicationFactor(Integer.parseInt(new KafkaConfig(config).getCheckpointReplicationFactor().get()));
```
rmatharu: This makes sense, however, KafkaCheckpointManagerFactory already adheres to these configs when creating the spec. 
Please see KafkaCheckpointManagerFactory Line 52-55
```val checkpointSpec = KafkaStreamSpec.fromSpec(StreamSpec.createCheckpointStreamSpec(checkpointTopic, checkpointSystemName))
        .copyWithReplicationFactor(kafkaConfig.getCheckpointReplicationFactor.get.toInt)
        .copyWithProperties(kafkaConfig.getCheckpointTopicProperties)```

VartulAgrawal: The KafkaCheckpointManagerFactory is only created after the KafkaSystemAdmin is initialised so by the time KafkaCheckpointManagerFactory is called the replicationFactor is already reset to 2 (default for any general topic) and hence it needs to be based to the spec here. Tested this in our current setup too :)

prateekm:
@rmatharu @bharathkk Is this ready to merge? Anything else we need for this?

PR #1178, opened 2019-10-04
SAMZA-2334: SSPGrouperProxy should check if job is stateful, not if host-affinity is enabled.

dnishimura:
The current `SSPGrouperProxy` provides the capability for input partition counts to expand or contract by a multiple of the previous count. This is only need by stateful jobs. Currently the check for stateful jobs is to see if `host-affinity` is enabled. However, that is the wrong check because `host-affinity` does not imply a job is stateful. 

The fix here is to check to see if the job has any local state stores rather than checking if `host-affinity` is enabled. The other part of the fix is to allow disabling the `SSPGrouperProxy` in the situation where the job cannot expand or contract the input partitions by an appropriate factor.

@prateekm @shanthoosh @dxichen


dnishimura:
Closed to figure out the failing test.

PR #1179, opened 2019-10-04
SAMZA-2334: SSPGrouperProxy should check if job is stateful, not if host-affinity is enabled

dnishimura:
The current `SSPGrouperProxy` provides the capability for input partition counts to expand or contract by a multiple of the previous count. This is only need by stateful jobs. Currently the check for stateful jobs is to see if `host-affinity` is enabled. However, that is the wrong check because `host-affinity` does not imply a job is stateful. 

The fix here is to check to see if the job has any local state stores rather than checking if `host-affinity` is enabled. The other part of the fix is to allow disabling the `SSPGrouperProxy` in the situation where the job cannot expand or contract the input partitions by an appropriate factor.


dnishimura:
@prateekm @shanthoosh @dxichen - please review when you get a chance.

docs/learn/documentation/versioned/jobs/samza-configurations.md
```diff
@@ -77,9 +77,10 @@ These are the basic properties for setting up a Samza application.
...
+|job.systemstreampartition.<br>grouper.proxy.enabled|true|When enabled, this provides a proxy layer on top of the grouper configured by `job.systemstreampartition.grouper.factory` that allows stateful jobs to expand or contract their partition count by a multiple of the previous count so that events from an input stream partition are not stored in the local state of a different task. This will prevent erroneous results. If this configuration is disabled or if the job is stateless, the proxy will call the grouper directly and skip the expansion or contraction logic. See [SEP-5](https://cwiki.apache.org/confluence/display/SAMZA/SEP-5%3A+Enable+partition+expansion+of+input+streams) for more details.|
```
prateekm: Can we rename this config to reflect the feature (partition count increase) instead of the implementation (grouper proxy)? Same for the docs.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/SSPGrouperProxy.java
```diff
@@ -42,21 +44,21 @@
 /**
  * Provides a stream expansion aware task to partition assignments on top of a custom implementation
  * of the {@link SystemStreamPartitionGrouper}.
+ * @see <a href="https://cwiki.apache.org/confluence/display/SAMZA/SEP-5%3A+Enable+partition+expansion+of+input+streams">SEP-5</a>
```
prateekm: FYI, the Apache URL shortener: https://s.apache.org

docs/learn/documentation/versioned/jobs/samza-configurations.md
dnishimura: Sure will change.

samza-core/src/main/java/org/apache/samza/container/grouper/stream/SSPGrouperProxy.java
dnishimura: The url shortener requires an apache.org password. My apache JIRA or confluence passwords don't work. I guess you have to be a committer or PMC to have the credentials.

dnishimura:
@prateekm - Please take a look at the update and responses to your comments. Thanks!

dnishimura:
@shanthoosh - Can you please take a look as well? Thanks.

prateekm:
@dnishimura That's fine then. Let's merge this.

dnishimura:
> @dnishimura That's fine then. Let's merge this.

Thanks! @prateekm or @shanthoosh can you merge it? I don't have commit powers.

shanthoosh:
@dnishimura Done. Merged to trunk.

PR #1182, opened 2019-10-09
SAMZA-2343: Remove original split deployment code

cameronlee314:
Going to add new code for new split deployment strategy soon, so removing the original unused split deployment code.


cameronlee314:
@sborya can you please take a look?

samza-core/src/test/scala/org/apache/samza/job/TestShellCommandBuilder.scala
```diff
@@ -44,7 +44,7 @@ class TestShellCommandBuilder {
 
   // if cmdPath is specified, the full path to the command should be adjusted
   @Test
-  def testCommandWithFwkPath {
+  def testBuildCommandWithCommandPath {
```
prateekm: Did you mean 'testBuildCommandWithPackagePath'?

samza-shell/src/main/bash/run-class.sh
```diff
@@ -42,44 +42,16 @@ GC_LOG_ROTATION_OPTS="-XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=10 -XX:GC
...
-else
-  # default behavior, all the jars need to be appended on newlines
-  # to ensure line argument length of 72 bytes is not violated
-  for file in $BASE_LIB_DIR/*.[jw]ar;
-  do
-    CLASSPATH=$CLASSPATH" $file \n"
-  done
-  echo generated from BASE_LIB_DIR CLASSPATH=$CLASSPATH
-fi
+# all the jars need to be appended on newlines to ensure line argument length of 72 bytes is not violated
```
prateekm: @abhishekshivanna for this too.

samza-yarn/src/main/scala/org/apache/samza/job/yarn/YarnJob.scala
```diff
@@ -88,25 +85,7 @@ class YarnJob(config: Config, hadoopConfig: Configuration) extends StreamJob {
...
-    if (!fwkPath.isEmpty()) {
-      // if we have framework installed as a separate package - use it
-      cmdExec = fwkPath + "/" + fwkVersion + "/bin/run-jc.sh"
-
-      logger.info("Using FWK path: " + "export SAMZA_LOG_DIR=%s && ln -sfn %s logs && exec %s 1>logs/%s 2>logs/%s".
-             format(ApplicationConstants.LOG_DIR_EXPANSION_VAR, ApplicationConstants.LOG_DIR_EXPANSION_VAR, cmdExec,
-                    ApplicationConstants.STDOUT, ApplicationConstants.STDERR))
-
-    }
+    val cmdExec = "./__package/bin/run-jc.sh" // default location
```
prateekm: Minor: Maybe inline this method if not used in multiple places. If shared, can also just make this a class level val (constant).

cameronlee314:
> Looks good to me, thanks for the cleanup. Let's get a review from @sborya as well since he has more context.

He is busy with oncall this week. If you think it is important to wait for a review from him, then I can wait.

samza-yarn/src/main/scala/org/apache/samza/job/yarn/YarnJob.scala
cameronlee314: I noticed this too, but I will actually be adding more logic to this method in https://github.com/apache/samza/pull/1172, so I left it as is.

samza-core/src/test/scala/org/apache/samza/job/TestShellCommandBuilder.scala
cameronlee314: I intended to use "command path". There is a method `ShellCommandBuilder.setCommandPath`, so I made the test method consistent with that.

prateekm:
@cameronlee314 If Boris is OK with the overall change (removing the code + docs) then that's fine, we can commit.

cameronlee314:
Thanks for confirming @srinipunuru.
Boris also confirmed to me that he had no high-level concerns about merging this.

samza-shell/src/main/bash/run-class.sh
```diff
@@ -42,44 +42,16 @@ GC_LOG_ROTATION_OPTS="-XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=10 -XX:GC
...
+for file in $BASE_LIB_DIR/*.[jw]ar;
+do
+  CLASSPATH=$CLASSPATH" $file \n"
+done
+echo generated from BASE_LIB_DIR CLASSPATH=$CLASSPATH
```
abhishekshivanna: Minor:
I'm assuming the reverse sorting by version number is done so that if conflicting jar versions are somehow included in the build - they are loaded in a deterministic order ?
I think that changing this behavior may prevent these builds (with conflicts) from deploying deterministically.
Although, I'm not sure if its the Samza framework's responsibility to address this..

cameronlee314: My understanding is that this code is only triggered when `JOB_LIB_DIR` is different from `BASE_LIB_DIR`. With the removal of the old split deployment code, I don't see anywhere in which `JOB_LIB_DIR` is set, so it should always default to be the same as `BASE_LIB_DIR`. So this code block never will get executed anyways.
I think the default flow (`for file in $BASE_LIB_DIR/*.[jw]ar`) does also iterate in a deterministic (alphabetical) order anyways.

abhishekshivanna: Makes sense. LGTM !

PR #1184, opened 2019-10-10
SAMZA-2346: Fixing Test timeouts for Allocator testing

Sanil15:
Some of these tests were failing due to expiry timeouts on some machines and being flaky. To fix this we have just increased the thread sleep time to ensure expiry of requests on allocator thread gets triggered.


Sanil15:
@cameronlee314 @rmatharu for review

PR #1187, opened 2019-10-14
SAMZA-2348 : Notifying BEM about kafkaConsumerProxy failure to prevent BEM poll-blockin in case of proxy failures

rmatharu:
Notifying BEM about kafkaConsumerProxy failure via kafkaConsumer to prevent BEM poll blocking in case of proxy failures

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/util/BlockingEnvelopeMap.java
+++ b/samza-api/src/main/java/org/apache/samza/util/BlockingEnvelopeMap.java
@@ -30,6 +30,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.samza.SamzaException;
 import org.apache.samza.metrics.Counter;
 import org.apache.samza.metrics.Gauge;
 import org.apache.samza.metrics.MetricsRegistry;
@@ -68,6 +69,7 @@ public abstract class BlockingEnvelopeMap implements SystemConsumer {
   private final ConcurrentHashMap<SystemStreamPartition, AtomicLong> bufferedMessagesSize;  // size in bytes per SystemStreamPartition
   private final Map<SystemStreamPartition, Boolean> noMoreMessage;
   private final Clock clock;
+  private volatile Throwable failureCause = null;
 
   public BlockingEnvelopeMap() {
     this(new NoOpMetricsRegistry());
@@ -141,7 +143,7 @@ public abstract class BlockingEnvelopeMap implements SystemConsumer {
         if (timeout == SystemConsumer.BLOCK_ON_OUTSTANDING_MESSAGES) {
           // Block until we get at least one message, or until we catch up to
           // the head of the stream.
-          while (envelope == null && !isAtHead(systemStreamPartition)) {
+          while (envelope == null && !isAtHead(systemStreamPartition) && !isConsumerFailed()) {
             metrics.incBlockingPoll(systemStreamPartition);
             envelope = queue.poll(1000, TimeUnit.MILLISECONDS);
           }
@@ -239,6 +241,19 @@ public abstract class BlockingEnvelopeMap implements SystemConsumer {
     Boolean isAtHead = noMoreMessage.get(systemStreamPartition);
 
     return getNumMessagesInQueue(systemStreamPartition) == 0 && isAtHead != null && isAtHead.equals(true);
+  }
+
+  protected void setFailureCause(Throwable throwable) {
+    this.failureCause = throwable;
+  }
+
+  protected boolean isConsumerFailed() {
+    if (this.failureCause != null) {
+      String message = String.format("%s: Consumer has stopped.", this);
+      throw new SamzaException(message, this.failureCause);
+    } else {
+      return false;
+    }
   }
 
   public class BlockingEnvelopeMapMetrics {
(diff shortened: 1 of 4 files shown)


samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -204,6 +204,7 @@ private Runnable createProxyThreadRunnable() {
         // KafkaSystemConsumer uses the failureCause to propagate the throwable to the container
         failureCause = throwable;
         isRunning = false;
+        ((KafkaSystemConsumer) this.kafkaConsumer).setProxyFailureCause(this.failureCause);
```
prateekm: Should be set before isRunning = false?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -219,6 +219,14 @@ private void setFetchThresholds() {
         this, fetchThresholdBytes, fetchThreshold, numPartitions, perPartitionFetchThreshold, perPartitionFetchThresholdBytes);
   }
 
+  /**
+   * Invoked by {@link KafkaConsumerProxy} to notify the consumer of failure, so it can relay and stop the BEM polling.
+   * @param throwable
+   */
+  public void setProxyFailureCause(Throwable throwable) {
```
prateekm: setFailureCause

samza-api/src/main/java/org/apache/samza/util/BlockingEnvelopeMap.java
```diff
@@ -141,7 +143,7 @@ private void initializeInternalStateForSSP(SystemStreamPartition systemStreamPar
         if (timeout == SystemConsumer.BLOCK_ON_OUTSTANDING_MESSAGES) {
           // Block until we get at least one message, or until we catch up to
           // the head of the stream.
-          while (envelope == null && !isAtHead(systemStreamPartition)) {
+          while (envelope == null && !isAtHead(systemStreamPartition) && !isConsumerFailed()) {
```
prateekm: Nitpick: Reverse order of isConsumerFailed and isAtHead

samza-api/src/main/java/org/apache/samza/util/BlockingEnvelopeMap.java
```diff
@@ -241,6 +243,19 @@ protected boolean isAtHead(SystemStreamPartition systemStreamPartition) {
     return getNumMessagesInQueue(systemStreamPartition) == 0 && isAtHead != null && isAtHead.equals(true);
   }
 
+  protected void setFailureCause(Throwable throwable) {
+    this.failureCause = throwable;
+  }
+
+  protected boolean isConsumerFailed() {
```
prateekm: `boolean` return type methods shouldn't throw an exception. Maybe check for exception directly and rethrow.

samza-kafka/src/main/scala/org/apache/samza/system/kafka/KafkaConsumerProxy.java
prateekm: Does this work? Isn't kafkaConsumer the actual Kafka Consumer, not the SystemConsumer?

rmatharu: it doesnt. need some way of giving KafkaConsumer proxy a handle to the SystemConsumer.

samza-api/src/main/java/org/apache/samza/util/BlockingEnvelopeMap.java
```diff
@@ -144,6 +146,12 @@ private void initializeInternalStateForSSP(SystemStreamPartition systemStreamPar
           while (envelope == null && !isAtHead(systemStreamPartition)) {
             metrics.incBlockingPoll(systemStreamPartition);
             envelope = queue.poll(1000, TimeUnit.MILLISECONDS);
+
+            // Check for consumerFailure and throw exception
+            if (this.failureCause != null) {
```
prateekm: Minor: Can move before the metrics increment and poll (first thing in the loop).

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemConsumer.java
```diff
@@ -219,6 +219,15 @@ private void setFetchThresholds() {
         this, fetchThresholdBytes, fetchThreshold, numPartitions, perPartitionFetchThreshold, perPartitionFetchThresholdBytes);
   }
 
+  /**
+   * Invoked by {@link KafkaConsumerProxy} to notify the consumer of failure, so it can relay and stop the BEM polling.
+   * @param throwable
```
prateekm: Add documentation for @param, otherwise checkstyle will complain.

prateekm:
Code looks good top me. Can you clarify the JIRA / commit description to explain the symptoms we saw (Container startup blocks indefinitely during Changelog / Checkpoint / Coordinator stream read),
the cause: (SSPIterator keeps blocking on the blocking poll even when KafkaConsumerProxy stops due to exceptions thrown from KafkaConsumer), and the fix (KafkaConsumerProxy reports exceptions to KafkaSystemConsumer, which then reports it to BEM. BEM checks if there are any exceptions when doing a poll and re-throws.

PR #1190, opened 2019-10-15
SAMZA-2352: Use min.compaction.lag.ms to avoid compacting the Kafka changelog topic

xinyuiscool:
To enable transactional state, we will need to avoid the head of the changelog (the ones after last checkpoint) not being log-compacted so we can trim the rest of the updates. We use min.compaction.log.ms to control the compaction time.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
+++ b/samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
@@ -52,6 +52,9 @@ public class StorageConfig extends MapConfig {
   public static final boolean DEFAULT_DISALLOW_LARGE_MESSAGES = false;
   public static final String DROP_LARGE_MESSAGES = STORE_PREFIX + "%s.drop.large.messages";
   public static final boolean DEFAULT_DROP_LARGE_MESSAGES = false;
+  // The log compaction lag time for transactional state change log
+  public static final String CHANGELOG_MIN_COMPACTION_LAG_MS = STORE_PREFIX + "%s.kafka.changelog.replication.factor";
+  public static final long DEFAULT_CHANGELOG_MIN_COMPACTION_LAG_MS = TimeUnit.HOURS.toMillis(4);
 
   static final String CHANGELOG_SYSTEM = "job.changelog.system";
   static final String CHANGELOG_DELETE_RETENTION_MS = STORE_PREFIX + "%s.changelog.delete.retention.ms";
@@ -205,6 +208,10 @@ public class StorageConfig extends MapConfig {
 
   public boolean getDropLargeMessages(String storeName) {
     return getBoolean(String.format(DROP_LARGE_MESSAGES, storeName), DEFAULT_DROP_LARGE_MESSAGES);
+  }
+
+  public long getChangelogMinCompactionLagMs(String storeName) {
+    return getLong(String.format(CHANGELOG_MIN_COMPACTION_LAG_MS, storeName), DEFAULT_CHANGELOG_MIN_COMPACTION_LAG_MS);
   }
 
   /**
--- a/samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
+++ b/samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
@@ -53,7 +53,7 @@ public class StorageConfig extends MapConfig {
   public static final String DROP_LARGE_MESSAGES = STORE_PREFIX + "%s.drop.large.messages";
   public static final boolean DEFAULT_DROP_LARGE_MESSAGES = false;
   // The log compaction lag time for transactional state change log
-  public static final String CHANGELOG_MIN_COMPACTION_LAG_MS = STORE_PREFIX + "%s.kafka.changelog.replication.factor";
+  public static final String CHANGELOG_MIN_COMPACTION_LAG_MS = STORE_PREFIX + "%s.changelog.kafka.min.compaction.lag.ms";
   public static final long DEFAULT_CHANGELOG_MIN_COMPACTION_LAG_MS = TimeUnit.HOURS.toMillis(4);
 
   static final String CHANGELOG_SYSTEM = "job.changelog.system";
--- a/samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
+++ b/samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
@@ -31,6 +31,8 @@ import org.apache.samza.SamzaException;
 import org.apache.samza.execution.StreamManager;
 import org.apache.samza.util.StreamUtil;
 
+import static com.google.common.base.Preconditions.*;
+
 
 /**
  * Config helper methods related to storage.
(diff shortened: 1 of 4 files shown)


xinyuiscool:
@prateekm : please take a look. thx.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -338,6 +340,15 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
 
     kafkaChangeLogProperties.setProperty("segment.bytes", KafkaConfig.CHANGELOG_DEFAULT_SEGMENT_SIZE)
     kafkaChangeLogProperties.setProperty("delete.retention.ms", String.valueOf(new StorageConfig(config).getChangeLogDeleteRetentionInMs(name)))
+
+    // To enable transactional state, we will need to avoid the head of the changelog
+    // (the messages after last checkpoint) being log-compacted so we can trim the rest of the updates.
+    // We use min.compaction.log.ms to control the compaction time.
+    if (new TaskConfig(this).getTransactionalStateRestoreEnabled) {
+      kafkaChangeLogProperties.setProperty("min.compaction.lag.ms",
```
prateekm: Nitpick: extract constant for config key.

xinyuiscool:
@prateekm : the current logic already allows user to override the value for each store, using something like ""stores.test2.changelog.kafka.min.compaction.lag.ms". The unit test I added verified it. Currently all the other changelog related configs live in KafkaConfig, like replication factor and max message bytes. So the new config also went in there. KafkaConfig also takes care of the overrides too.

Let me add a getter in KafkaConfig for getting this value.

prateekm:
@xinyuiscool Got it, thanks. Another issue with adding this in KafkaConfig is that KafkaConfig is in samza-kafka. To read this during restore, we'll have to add a dependency on samza-kafka and kafka in samza-core, which we should avoid.

xinyuiscool:
@prateekm : by looking at the other config, e.g. replication factor and max message byte, looks like the StorageConfig provides a getter for those configs. I can do the same. The current logic in KafkaConfig will remain there as that's the central place to create changelog configs, similar to the other configs. Sounds good?

prateekm:
@xinyuiscool Sounds good, thanks.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -52,6 +52,9 @@
   public static final boolean DEFAULT_DISALLOW_LARGE_MESSAGES = false;
   public static final String DROP_LARGE_MESSAGES = STORE_PREFIX + "%s.drop.large.messages";
   public static final boolean DEFAULT_DROP_LARGE_MESSAGES = false;
+  // The log compaction lag time for transactional state change log
+  public static final String CHANGELOG_MIN_COMPACTION_LAG_MS = STORE_PREFIX + "%s.kafka.changelog.replication.factor";
```
prateekm: Typo in config name (still says replication.factor)

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -88,6 +88,8 @@ object KafkaConfig {
 
   val DEFAULT_RETENTION_MS_FOR_BATCH = TimeUnit.DAYS.toMillis(1)
 
+  val MIN_COMPACTION_LOG_MS = "min.compaction.lag.ms"
```
prateekm: Typo in variable name: LAG_MS instead of LOG_MS

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
xinyuiscool: Good catch! Let me fix it.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
xinyuiscool: Fixed.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -52,6 +52,9 @@
   public static final boolean DEFAULT_DISALLOW_LARGE_MESSAGES = false;
   public static final String DROP_LARGE_MESSAGES = STORE_PREFIX + "%s.drop.large.messages";
   public static final boolean DEFAULT_DROP_LARGE_MESSAGES = false;
+  // The log compaction lag time for transactional state change log
+  public static final String CHANGELOG_MIN_COMPACTION_LAG_MS = STORE_PREFIX + "%s.changelog.kafka.min.compaction.lag.ms";
```
prateekm: Should this be 'changelog.min.compaction.lag.ms' to match the other Kafka related properties above (replication.factor and max.message.size.bytes)?

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -338,6 +340,15 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
 
     kafkaChangeLogProperties.setProperty("segment.bytes", KafkaConfig.CHANGELOG_DEFAULT_SEGMENT_SIZE)
     kafkaChangeLogProperties.setProperty("delete.retention.ms", String.valueOf(new StorageConfig(config).getChangeLogDeleteRetentionInMs(name)))
+
+    // To enable transactional state, we will need to avoid the head of the changelog
+    // (the messages after last checkpoint) being log-compacted so we can trim the rest of the updates.
+    // We use min.compaction.log.ms to control the compaction time.
+    if (new TaskConfig(this).getTransactionalStateRestoreEnabled) {
+      kafkaChangeLogProperties.setProperty(KafkaConfig.MIN_COMPACTION_LAG_MS,
+        String.valueOf(StorageConfig.DEFAULT_CHANGELOG_MIN_COMPACTION_LAG_MS))
```
prateekm: How does the override work? Do you need to get this using the StorageConfig getter instead of the default?

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
xinyuiscool: I think we are in an inconsistent position for this config. On one hand, the logic in KafkaConfig.getChangelogKafkaProperties() already allows overriding any kafka properties by using stores.x.changelog.kafka..... So the user can already override things like replication.factor using stores.x.changelog.kafka.replication.factor. The changelog configs in StorageConfig make a second way to configure it, which is problematic since the user can use the original kafka override to do it and StorageConfig code getters won't return the correct value. Seems like an existing bug to me already.

Not sure how many use cases are using those two configs, but it's problematic in the long run.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
xinyuiscool: no, the override is already happening by appending all the properties specified after stores.x.changelog.kafka.xxxxx

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
prateekm: Oh I see, that's quite confusing. Even the configuration reference documents the changelog.kafka.* properties incorrectly (provided example is incorrect) and does not clarify its precedence. Looks like the convention is: if it's a known property that the framework cares about, set it explicitly using the changelog.xyz convention. If it's an arbitrary Kafka property, set it using changelog.kafka.xyz convention. 

Following this convention, and for consistency with existing behavior class, I'd still prefer to rename this to changelog.min.compaction.lag.ms. 

Seems like long term fix would be to give higher precedence to the explicitly specified properties over the kafka.* specified properties, although that's backwards incompatible. We can create a ticket to track that separately.

PR #1192, opened 2019-10-15
SAMZA-2354: Improve UDF discovery in samza-sql.

shanthoosh:
This patch is comprised of following changes: 

1. Currently to plugin a udf implementation into samza-sql engine, users have to set the fully qualified class name of their UDF implementation in `samza.sql.udfClasses` configuration.  Any user error made in defining the fully-qualified name of udf-class configuration can only be identified  at run-time. This patch addresses this problem by using reflection to find all the samza-sql UDF implementations available in the class-path of the samza-sql application.
2. Removes the ConfigBasedUdfResolver class and it's associated usages. 
3. Adds  unit-tests for the newly introduced `UDFResolver` implementation.


shanthoosh:
@atoomula @srinipunuru 

samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedUdfResolver.java
```diff
@@ -1,119 +0,0 @@
-/*
```
atoomula: Please don't delete this file.

samza-sql/src/main/java/org/apache/samza/sql/udf/ReflectionBasedUdfResolver.java
```diff
@@ -0,0 +1,97 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(ReflectionBasedUdfResolver.class);
+
+  private static final String CONFIG_PACKAGE_PREFIX = "samza.sql.udf.resolver.package.prefix";
+
+  private final Set<UdfMetadata> udfs  = new HashSet<>();
+
+  public ReflectionBasedUdfResolver(Config udfConfig) {
+    // Searching the entire classpath to discover the subtypes of SamzaSqlUdf is expensive. To reduce the search space,
+    // the search is limited to the set of package prefixes defined in the configuration.
+    // Within Linkedin this configuration will be overridden to ["com.linkedin.samza", "org.apache.samza", "com.linkedin.samza.sql.shade.prefix"].
```
atoomula: Please remove references to Linkedin from the code.

samza-sql/src/main/java/org/apache/samza/sql/udf/ReflectionBasedUdfResolver.java
```diff
@@ -0,0 +1,97 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * An UDF resolver implementation that uses reflection to discover the subtypes
+ * of the {@link SamzaSqlUdf} from the classpath. Performs the validation to
+ * ensure that all subtypes of {@link SamzaSqlUdf} extend and implement the
+ * method annotated with {@link SamzaSqlUdfMethod}.
+ */
+public class ReflectionBasedUdfResolver implements UdfResolver {
```
atoomula: Just curious, how much of a hit are we taking with OSS jar to load the UDFs via reflection ?

shanthoosh: By selectively searching for UDF's  only in selected package-prefixes from class-path for UDF's, it's a very minimal over-head.

samza-sql/src/main/java/org/apache/samza/sql/udf/ReflectionBasedUdfResolver.java
shanthoosh: Removed the LinkedIn comment.

samza-sql/src/main/java/org/apache/samza/sql/impl/ConfigBasedUdfResolver.java
shanthoosh: Sure, reverted this deletion.

samza-sql/src/main/java/org/apache/samza/sql/runner/SamzaSqlApplicationConfig.java
```diff
@@ -214,7 +214,7 @@ private UdfResolver createUdfResolver(Map<String, String> config) {
     Properties props = new Properties();
     props.putAll(domainConfig);
     HashMap<String, String> udfConfig = getDomainProperties(config, CFG_UDF_CONFIG_DOMAIN, false);
-    return new ConfigBasedUdfResolver(props, new MapConfig(udfConfig));
+    return new ReflectionBasedUdfResolver(new MapConfig(udfConfig));
```
atoomula: Sorry, I didn't realize that this is hard-coded here. Could we make this pluggable ? I think calling initializePlugin like above should do it with the udf resolver based arguments.

shanthoosh: I agree that this is a useful thing to do and discussed it offline. Created a follow-up ticket(SAMZA-2355) for making the UDFResolver pluggable in samza-sql. It will be done in an immediate follow-up patch.

PR #1196, opened 2019-10-19
SAMZA-2356: [Transactional State] Do not trim changelog if time since last checkpoint is greater than min.compaction.lag.

prateekm:
In the current implementation for transactional state, if the checkpointed changelog offset gets compacted, (e.g. if the processor was down for more than min compaction lag), and the checkpointed changelog offset is b/w the oldest and newest post compaction, we will end up deleting part of the current state (potentially everything after checkpointed offset). We should not do this.

Instead, we should write the checkpoint timestamp along with the the checkpointed changelog offset, and not do a trim during restore if time since last checkpoint is older than (or close to) min.compaction.lag. We should instead restore to latest and log a warning that transactional state guarantee doesn't hold (i.e., same as the current behavior). The fix for users would then be to increase the min.compaction.lag.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/checkpoint/CheckpointId.java
+++ b/samza-api/src/main/java/org/apache/samza/checkpoint/CheckpointId.java
@@ -18,6 +18,7 @@
  */
 package org.apache.samza.checkpoint;
 
+import java.util.Objects;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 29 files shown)


prateekm:
@bharathkk FYI

prateekm:
@xinyuiscool gave me a verbal approval for this, but is having issues with his Github 2FA. Merging this commit for now.

PR #1200, opened 2019-10-22
SAMZA-2360 : Adding config stores.init.clean to clean all store-dirs on init, regardless of contents

rmatharu:
Adding config stores.init.clean to clean all store-dirs on init, regardless of contents


prateekm:
@rmatharu Please add a JIRA and describe why we're making this change.

samza-core/src/main/java/org/apache/samza/storage/NonTransactionalStateTaskRestoreManager.java
```diff
@@ -110,20 +110,21 @@
    */
   @Override
   public void init(Map<SystemStreamPartition, String> checkpointedChangelogSSPOffsets) {
-    cleanBaseDirsAndReadOffsetFiles();
+    cleanBaseDirsAndReadOffsetFiles(new StorageConfig(config).getCleanLoggedStoreDirsOnInit());
```
prateekm: What about the other restore manager?

Extract variables.

Add tests.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -73,6 +73,8 @@
   static final String INMEMORY_KV_STORAGE_ENGINE_FACTORY =
       "org.apache.samza.storage.kv.inmemory.InMemoryKeyValueStorageEngineFactory";
 
+  public static final String CLEAN_LOGGED_STOREDIRS_ON_INIT = STORE_PREFIX + "init.clean";
```
prateekm: Shouldn't be init, should say container start or something similar.

samza-core/src/main/java/org/apache/samza/storage/NonTransactionalStateTaskRestoreManager.java
rmatharu: The TransactionalStateTaskRestoreManager already seems to delete storeDirs on start after reading?

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
rmatharu: changed it to stores.container.start.clean 
how does that sound

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -73,6 +73,8 @@
   static final String INMEMORY_KV_STORAGE_ENGINE_FACTORY =
       "org.apache.samza.storage.kv.inmemory.InMemoryKeyValueStorageEngineFactory";
 
+  public static final String CLEAN_LOGGED_STOREDIRS_ON_START = STORE_PREFIX + "container.start.clean";
```
bharathkumarasubramanian: Is this supposed to be an internal config? If not, can you add a configuration doc for this?

rmatharu: Yes this is purely internal -- intended for benchmarking and validation use.

prateekm: Would still be good to add docs (comments) explaining what this config does.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
prateekm: Maybe also make it per store.
`stores.<store-name>.clear.on.container.start`?

samza-core/src/main/java/org/apache/samza/storage/NonTransactionalStateTaskRestoreManager.java
prateekm: It always deletes the `current` directory, but may restore from a store snapshot/checkpoint. So you'll either need to disable restore completely, or clear the restored store at the end.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
rmatharu: done

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
rmatharu: done

samza-core/src/main/java/org/apache/samza/storage/NonTransactionalStateTaskRestoreManager.java
rmatharu: done

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
mynameborat: +1. `clearn.on.container.start` sounds more readable.

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
mynameborat: If you do plan to open it up, you can leave it as is.
Otherwise, I noticed we also insert `internal` as part of the config name to call out its internal. e.g. `samza.internal.execution.plan`

PR #1170, opened 2019-09-25
Samza-2330: Handle expired resource request for Container allocator when host affinity is disabled	

Sanil15:
**Summary of the problem**: Today when host affinity is not enabled & Cluster Manager does not issue the requested resources, the allocator thread keeps on looping forever, we want to add an expiry check for such requests to re-issue another ANY_HOST just in case if capacity is later added

This PR introduces expiry check for container allocator when host affinity is disabled. Expiry timeout is defined by the existing config: `cluster-manager.container.request.timeout.ms` with a default of 5 mins. This will prevent samza users to prevent to restart jobs in case of cluster manager being not responsive. 

**Testing**: Added unit tests for simulating and verifying the behavior

PR #1201, opened 2019-10-24
SAMZA-2361: Revert "Samza-2330: Handle expired resource request for Container allocator when host affinity is disabled"

Sanil15:
**Symptom:** PR #1170 introduced expired request checking for Host Affinity disabled cases, however, this leads to a few problems that we might face in future:
1. In the scenarios of a cluster bounce, this might result in jobs (in either case of host affinity enabled & disabled cases) requesting containers from the ClusterResourceManager every 5 (default request expiry timeout) seconds if requests can't be fulfilled. Before PR #1170 this only used to happen for host affinity disabled jobs. This might potentially result in bringing down the ClusterManager. 
2. Unless we have some performance benchmarking on RM (for Yarn as a cluster balancer), having a default expiry timeout of 5 seconds for all container requests is not justified.

**Cause:** PR #1170

**Fix:** We are reverting this behavior change since we don't have Performance benchmarking for Yarn

**Tests:** Refactored test to ensure, expired requests are only checked with host affinity enabled jobs

**API Changes:** N/A

**Upgrade Instructions:** N/A


Sanil15:
@rmatharu please have a look!

prateekm:
@Sanil15 Is this a revert of a previous commit? If so, instead of reverting on master, you should revert the commit on the release branch. @lhaiesp can help you with that.

Sanil15:
@prateekm let's revert it on master too since before enabling this behavior we need more benchmarking, after that we can cherry-pick this change to release branches?

prateekm:
Sure, revert it from both master and the 1.3 branch in that case.

PR #1209, opened 2019-11-06
SAMZA-2324: Adding KV store metrics for rocksdb

bkonold:
**Design:** Adding a metric to track the maximum serialized value size written to rocksdb. This provides value to customers using logged stores backed by kafka. As kafka has configurable message size limits, the addition of this metric will allow users to understand how their records are approaching that limit.
**Changes:** Adds a new metric "max-record-size-bytes" to SerializedKeyValueStoreMetrics, which is tracked in SerializedKeyValueStore.
**Tests:** Built a version of samza locally which included this commit, then ran a test job that wrote consistently sized records to rocksdb and verified that the metric emitted matched that size.

**API Changes:** none
**Upgrade Instructions:** No manual steps other than upgrading to the samza version containing this commit. Once upgraded the new record size metric will be emitted.
**Usage Instructions:** None. New metric will be emitted automatically on upgrading version.

### DIFF
--- a/samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
+++ b/samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
@@ -25,7 +25,6 @@ import java.util.concurrent.TimeUnit
 import java.util.concurrent.locks.ReentrantReadWriteLock
 import java.util.{Comparator, Optional}
 
-import org.apache.commons.io.FileUtils
 import org.apache.samza.SamzaException
 import org.apache.samza.checkpoint.CheckpointId
 import org.apache.samza.config.Config
@@ -101,8 +100,6 @@ object RocksDbKeyValueStore extends Logging {
             "0"
           }
         ))
-
-      metrics.newGauge("rocksdb.size-on-disk", () => FileUtils.sizeOfDirectory(new File(dir.getAbsolutePath)))
 
       rocksDb
     } catch {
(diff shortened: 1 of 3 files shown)


samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
```diff
@@ -101,6 +102,8 @@ object RocksDbKeyValueStore extends Logging {
           }
         ))
 
+      metrics.newGauge("rocksdb.size-on-disk", () => FileUtils.sizeOfDirectory(new File(dir.getAbsolutePath)))
```
prateekm: 1. Why not use RocksDB internal metric?
2. Regardless of 1, should not re-use the rocksdb internal metrics naming convention to report custom metrics.

bkonold: 1. Two reasons: first, the behavior I recently reported of the metrics being broken (or at the very least, not predictably populated by all jobs running the latest release). Second, I was unable to determine whether the "total-sst-files-size" would include files nested within checkpoint dirs; calculating the size of the directory this way would recursively include that as well. If that is undesirable however, this should be discarded in favor of another approach.
2. Fair enough, I will rename to something else. Feel free to comment if you have suggestions.

prateekm: For 1, we should debug and fix that behavior if metrics aren't being emitted correctly. Would also be good to verify if total-sst-files-size includes files retained due to checkpoints, should be simple to verify locally, I can help set that up. I'd prefer to try these out before making this change.

bkonold: Ya you're right. Let's remove that from this PR then and circle back to it - I can't commit time immediately to debug and fix.

Created SAMZA-2374 to track.

bkonold:
@prateekm is there anything else you'd like addressed in this PR / others to take a look?

PR #1210, opened 2019-11-06
SAMZA-2372:Null pointer exception in LocalApplicationRunner

IgorDurovic:
Null pointer exception is thrown when calling LocalApplicationRunner.kill() without setting a coordinator stream. My guess that the bug is a result of https://issues.apache.org/jira/browse/SAMZA-2294 forgetting to add the null check everywhere it's required. I added a test that exposes the exception.


IgorDurovic:
@dnishimura

PR #1215, opened 2019-11-09
SAMZA-2377: Batch mode should be computed based on generated configs

xinyuiscool:
Currently the batch mode in job planner is computed based on the userconfig. This should be changed after the Samza InputDescriptor change. We should compute that based on the generated configs now.

Also add the logic to remove host affinity config for batch mode.


xinyuiscool:
@mynameborat : pls help take a look.

PR #1217, opened 2019-11-12
SAMZA-2381: Refactor JobRunner#run and split into multiple util methods.

kw2542:
Issues:
  JobRunner#run is composed of three parts, persisting config in coordinator stream, creation of diagnostics stream when applicable and submit the job. It is not easy to read when all of them are combined in a single method and it cannot be reused when a subset of the functionality is needed elsewhere. For example, when we move planning to AM, we will only need to submit job on runner but the rest on AM.

Changes:
  1. Move coordinator stream related logic to CoordinatorStreamUtil#writeConfigToCoordinatorStream
  2. Move diagnostics stream related logic to DiagnosticsUtil#createDiagnosticsStream
  3. Move job submission related logic to JobRunner#submit()

Tests:
  1. Add unit tests to test CoordinatorStreamUtil#writeConfigToCoordinatorStream

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
+++ b/samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
@@ -66,17 +66,33 @@ object JobRunner extends Logging {
 class JobRunner(config: Config) extends Logging {
 
   /**
-   * This function submits the samza job.
-   * @param resetJobConfig This flag indicates whether or not to reset the job configurations when submitting the job.
+   * This function persist config in coordinator stream, create diagnostics stream if applicable and
+   * then submits the samza job.
+   * @param resetJobConfig This flag indicates whether or not to reset the job configurations in coordinator stream
+   *                       when submitting the job.
    *                       If this value is set to true, all previously written configs to coordinator stream will be
    *                       deleted, and only the configs in the input config file will have an affect. Otherwise, any
    *                       config that is not deleted will have an affect.
    *                       By default this value is set to true.
    * @return The job submitted
    */
-  def run(resetJobConfig: Boolean = true) = {
+  def run(resetJobConfig: Boolean = true): StreamJob = {
+    persist(resetJobConfig)
+    submit()
+  }
+
+  /**
+   * This function persist config in coordinator stream.
+   *
+   * @param resetJobConfig This flag indicates whether or not to reset the job configurations in coordinator stream
+   *                       when submitting the job.
+   *                       If this value is set to true, all previously written configs to coordinator stream will be
+   *                       deleted, and only the configs in the input config file will have an affect. Otherwise, any
+   *                       config that is not deleted will have an affect.
+   *                       By default this value is set to true.
+   */
+  def persist(resetJobConfig: Boolean = true): Unit = {
     debug("config: %s" format (config))
-    val jobFactory: StreamJobFactory = getJobFactory
     val coordinatorSystemConsumer = new CoordinatorStreamSystemConsumer(config, new MetricsRegistryMap)
     val coordinatorSystemProducer = new CoordinatorStreamSystemProducer(config, new MetricsRegistryMap)
     val systemAdmins = new SystemAdmins(config)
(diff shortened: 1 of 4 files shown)


samza-core/src/main/scala/org/apache/samza/job/JobRunner.scala
```diff
@@ -66,17 +66,33 @@ object JobRunner extends Logging {
...
    * @return The job submitted
    */
-  def run(resetJobConfig: Boolean = true) = {
+  def run(resetJobConfig: Boolean = true): StreamJob = {
+    persist(resetJobConfig)
+    submit()
+  }
+
+  /**
+   * This function persist config in coordinator stream.
```
lhaiesp: Besides creating and writing into coordinator stream, this method also creates diagnostics streams and potentially other stuff in the future. Suggest we provide more details in the java doc here. Also suggest a more generic method name such as "prepare()"

samza-core/src/main/scala/org/apache/samza/util/CoordinatorStreamUtil.scala
```diff
@@ -37,13 +39,12 @@ object CoordinatorStreamUtil extends Logging {
     * Given a job's full config object, build a subset config which includes
     * only the job name, job id, and system config for the coordinator stream.
     */
-  def buildCoordinatorStreamConfig(config: Config) = {
+  def buildCoordinatorStreamConfig(config: Config): MapConfig = {
     val jobConfig = new JobConfig(config)
-    val buildConfigFactory = jobConfig.getCoordinatorStreamFactory();
+    val buildConfigFactory = jobConfig.getCoordinatorStreamFactory
```
mynameborat: nit: I'd recommend leaving the parentheses as is for convention. all java method call inside scala has explicit parentheses even in the absence of parameters.

samza-core/src/test/scala/org/apache/samza/util/TestCoordinatorStreamUtil.scala
```diff
@@ -84,4 +84,19 @@ class TestCoordinatorStreamUtil {
...
+    addConfig.put("job.name", "test-job-name")
+    addConfig.put("job.id", "i001")
+    addConfig.put("job.coordinator.system", "samzatest")
+    addConfig.put("systems.samzatest.test","test")
+    addConfig.put("test.only","nothing")
+    addConfig.put("systems.samzatest.samza.factory", "org.apache.samza.system.MockSystemFactory")
+    val config = new MapConfig(addConfig)
+    val configMap = CoordinatorStreamUtil.buildCoordinatorStreamConfig(config)
+
+    CoordinatorStreamUtil.writeConfigToCoordinatorStream(configMap)
```
mynameborat: Can we verify if the configs are piped to the mock system producer? 
Now that you are adding tests, maybe add one for resetJobConfig to make sure the configs not part of the old configs get deleted?

mynameborat:
Can you also fix the checkstyle issue
```
> Task :samza-core_2.11:checkstyleMain
[ant:checkstyle] /home/travis/build/apache/samza/samza-core/src/main/java/org/apache/samza/util/DiagnosticsUtil.java:163:12: error: Name 'DIAGNOSTICS_STREAM_ID' must match pattern '^[a-z][a-zA-Z0-9]*$'.
> Task :samza-core_2.11:checkstyleMain FAILED
```

samza-core/src/main/scala/org/apache/samza/util/CoordinatorStreamUtil.scala
kw2542: reverted the change to keep consistency.

samza-core/src/test/scala/org/apache/samza/util/TestCoordinatorStreamUtil.scala
kw2542: As discussed, in order for us to mock and verify, we need to mock multiple nested data, it is not worth the effort to do it for now, as the code are simply copied from existing code.

PR #1218, opened 2019-11-12
SAMZA-2382: User specified side input processor should take precedence over the identity processor

rmatharu:
User specified side input processor should take precedence over the identity processor

Symptom: Standby containers on jobs using side-inputs will run into deserialization exceptions when side-inputs use a custom serde (while changelog uses byte serde).

Cause: If the changelog doesnt use byte serde, the custom serde and processor should be respected.

Fix: Added logic that enables user specified side input processor to take precedence over the identity processor.

Tests: Samza test job that uses standby containers with side-inputs, with side-inputs using a custom serde.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/storage/ContainerStorageManager.java
+++ b/samza-core/src/main/scala/org/apache/samza/storage/ContainerStorageManager.java
@@ -554,7 +554,7 @@ public class ContainerStorageManager {
             // if this is a standby-task and the store is a non-side-input changelog store
             // we creating identity sideInputProcessor for stores of standbyTasks
             // have to use the right serde because the sideInput stores are created
-            
+
             Serde keySerde = serdes.get(config.getStorageKeySerde(storeName)
                 .orElseThrow(() -> new SamzaException("Could not find storage key serde for store: " + storeName)));
             Serde msgSerde = serdes.get(config.getStorageMsgSerde(storeName)
@@ -572,6 +572,7 @@ public class ContainerStorageManager {
                 }
               }
             };
+            LOG.info("Using identitty side-inputs-processor for store: {}, task: {}", storeName, taskName);
           }
 
           sideInputStoresToProcessors.get(taskName).put(storeName, sideInputsProcessor);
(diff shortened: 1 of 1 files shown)


samza-core/src/main/scala/org/apache/samza/storage/ContainerStorageManager.java
```diff
@@ -526,51 +526,57 @@ private StorageEngine createStore(String storeName, TaskName taskName, TaskModel
...
-        }
-      });
+            sideInputsProcessor = sideInputsProcessorFactory.getSideInputsProcessor(config, taskInstanceMetrics.get(taskName).registry());
+            LOG.info("Using side-inputs-processor from factory: {} for store: {}, task: {}", config.getSideInputsProcessorFactory(storeName).get(), storeName, taskName);
 
-    // creating identity sideInputProcessor for stores of standbyTasks
-    getTasks(containerModel, TaskMode.Standby).forEach((taskName, taskModel) -> {
-        sideInputStoresToProcessors.put(taskName, new HashMap<>());
-        for (String storeName : sideInputSystemStreams.keySet()) {
+          } else if (!config.getSideInputsProcessorFactory(storeName).isPresent() && taskMode.equals(TaskMode.Active)) {
```
mynameborat: Ideally all side inputs should have side inputs processor irrespective of what task mode is. At this point, we blur the line between actual side inputs vs changelog SSPs and that only allows us to validate for task mode active since we can't differentiate the actual side inputs vs changelog SSPs as side inputs.

I suggest we remove this validation since we have a validation upstream to check if the side inputs are specified along with the processor factory or processor lambda.

mynameborat:
Ignore my comment on the description. Looks like you already did. I have one minor comment on the validation. Looks good otherwise.

samza-core/src/main/scala/org/apache/samza/storage/ContainerStorageManager.java
rmatharu: done

rmatharu:
done

prateekm:
Looks like something easily unit testable. Can we add one for this?

PR #1219, opened 2019-11-12
SAMZA-2373: [SEP-22] Container Placement Service (core functionality) for container move and restart

Sanil15:
**Design:** [SEP-22: Container Placements in Samza](https://cwiki.apache.org/confluence/display/SAMZA/SEP-22%3A+Container+Placements+in+Samza)

**Changes:** : Following abilities are added as a part of this PR:

1. Container placement service core functionality to the Job Coordinator code
2. Ability to move containers to any / specific host for host affinity enabled jobs
3. Ability to move containers to any host only for host affinity disabled jobs
4. Standby enabled jobs are not supported in the PR (will follow with the stacked PR)

**Tests:** 
1. Integ test & Unit tests are added to moc and end to end control action and behavior
2. Manual testing matrix: [Container Placement Test Plan]( https://docs.google.com/spreadsheets/d/1v-fw0pHxKRobGkALDCno4FuPCsBhdepQ86vIGLHWu54/edit#gid=0)
 
**API Changes:**: 
1. This PR introduces behavior changes when a control action is active on job coordinator in such scenario the JC will try to reserve resources for the request, if it gets the resources it it attempts to stop the active container then attempts to start on the allocated resouces
2. All the state realted to control actions are mainted in memory in the JC, followed up by reporting & cleanup as soon as the control action finished
3. Optionally every control action can specifiy a request expiry timeout which when defined takes precedence over `cluster-manager.container.request.timeout.ms` which acts as a timeout for waiting to get resources or getting an ack on stop of a container from ResourceManager 

**Usage Instructions:** The ability to access these apis will be added as a part of follow up PR


samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
```diff
@@ -415,20 +407,34 @@ public void stop() {
...
     }
     return requestExpired;
   }
 
+  /**
+   * Control actions like move, restarts can optionally override request expiry timeout. Otherwise it defaults to
+   * { @code requestExpiryTimeout }
+   */
+  private Long getRequestTimeout(SamzaResourceRequest request) {
+    Optional<Long> controlActionRequestExpirytimeout = containerManager.getActionExpiryTimeout(request.getProcessorId());
```
Sanil15: @rmatharu this is the way I offer flexibility to do per-request expiry timeout, let me know your thoughts. If the timeout is present that takes higher precedence than one defined by `cluster-manager.container.
request.timeout.ms`

This timeout is not only used as a wait time when active container stop is issued

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
```diff
@@ -110,7 +110,7 @@
    */
   protected final ResourceRequestState resourceRequestState;
   /**
-   * Tracks the expiration of a request for resources.
+   * Tracks the default expiration of a request for resources.
```
rmatharu: the default value is in ClusterManagerConfig: DEFAULT_CONTAINER_REQUEST_TIMEOUT_MS, this tracks the expiration-threshold for a resource request.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
```diff
@@ -395,6 +386,7 @@ private CommandBuilder getCommandBuilder(String processorId) {
    * @param samzaResource returned by the ContainerManager
    */
   public final void addResource(SamzaResource samzaResource) {
+    //resourceRequestState.addResource(samzaResource, containerManager.isResourceRequestedForControlAction(samzaResource));
```
rmatharu: LOG.info("Container ID: {} allocated from RM on host: {}", resource.getContainerId(), resource.getHost());

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
```diff
@@ -415,20 +407,34 @@ public void stop() {
 
 
   /**
-   * Checks if a request has expired.
+   * Checks if a request has expired. If this request is due to a control action, then expiry timeout can be overridden
+   * otherwise defaults to
```
rmatharu: defaults to ... 
{@link ClusterManagerConfig.CLUSTER_MANAGER_REQUEST_TIMEOUT_MS} if specified or {@link ClusterManagerConfig.DEFAULT_CONTAINER_REQUEST_TIMEOUT_MS}

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
rmatharu: Do we really need this override-able timeout? Or is it a nice to have that we've thought of?
I'd suggest keeping this out for now and just using the configured value.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
```diff
@@ -415,20 +407,34 @@ public void stop() {
...
           request.getProcessorId(), request.getPreferredHost(), request.getRequestTimestamp(), currTime, requestExpiryTimeout);
     }
     return requestExpired;
   }
 
+  /**
+   * Control actions like move, restarts can optionally override request expiry timeout. Otherwise it defaults to
+   * { @code requestExpiryTimeout }
+   */
+  private Long getRequestTimeout(SamzaResourceRequest request) {
```
rmatharu: private long getRequestTimeout

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -161,4 +237,140 @@ void handleExpiredRequestWithHostAffinityEnabled(String processorId, String pref
...
+  private void handleControlActionFailure(String processorId, ContainerPlacementMetadata metaData, String failureMessage) {
+    metaData.setActionStatus(ContainerPlacementStatus.StatusCode.FAILED, failureMessage);
+    LOG.info("Control Action failed with metadata {}", metaData);
+    this.actions.remove(processorId);
+  }
+
+  /**
+   * Check if a activeContainerResource has control-action-metadata associated with it
+   */
+  private Optional<ContainerPlacementMetadata> getControlActionMetadata(String processorId) {
```
mynameborat: I'd suggest not wrapping this up in optional and instead have a method to determine if a processorId is associated with a control action and use that check everywhere this method is used and directly fetch the `ContainerPlacementMetadata`. 

IMO, the fact this is wrapped in `Optional` and we still end up doing `.isPresent`, `.get` and not have any functional programming usage defeats the brevity and the point of having optional.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -161,4 +237,140 @@ void handleExpiredRequestWithHostAffinityEnabled(String processorId, String pref
...
+   * Case 1. If active container fails to stop mark the container placement action failed.
+   * Case 2. Otherwise once active container shuts down then issue a start for the container on the preferred host.
+   *
+   * This method is invoked by the allocator thread which waits on notify from the thread issuing
+   * callbacks i.e {@link ContainerProcessManager}
+   */
+  private void handleContainerAllocationForExistingControlAction(String processorId, ContainerAllocator allocator,
+      ResourceRequestState state, SamzaResourceRequest request, String preferredhost) {
+    // check if container is already dead without issuing a stop here, fail the move request
+    ContainerPlacementMetadata actionMetaData = getControlActionMetadata(processorId).get();
```
mynameborat: invoking `.get()` without `.isPresent()` check

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -161,4 +237,140 @@ void handleExpiredRequestWithHostAffinityEnabled(String processorId, String pref
...
+    if (samzaApplicationState.runningProcessors.containsKey(processorId)) {
+      LOG.info("Requesting running container to shutdown due to existing container placement action {}",
+          getControlActionMetadata(processorId).get());
```
mynameborat: can reuse `actionMetadata`

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -161,4 +237,140 @@ void handleExpiredRequestWithHostAffinityEnabled(String processorId, String pref
...
+    actions.put(processorId, actionMetaData);
+    // note this also updates state.preferredHost count
+    containerAllocator.issueResourceRequest(resourceRequest);
+    LOG.info("Control action with metadata {} and issued a request for resources in progress", actionMetaData);
+    return actionStatus;
+  }
+
+  public Optional<Long> getActionExpiryTimeout(String processorId) {
+    return this.actions.containsKey(processorId) ? Optional.of(
+        this.actions.get(processorId).getRequestActionExpiryTimeout()) : Optional.empty();
```
mynameborat: why do we return `empty` here instead of the default timeout for action defined above? Consider using Duration and also remove `Optional` if possible since there is a default value defined.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -35,23 +37,43 @@
...
 public class ContainerManager {
 
   private static final Logger LOG = LoggerFactory.getLogger(ContainerManager.class);
+  private static final String ANY_HOST = ResourceRequestState.ANY_HOST;
 
+  /**
+   * Default timeout for request to the cluster manager for stopping an active container & checking expiry for
+   * requested resources
+   */
+  private static final Long DEFAULT_CONTROL_ACTION_EXPIRY = Duration.ofSeconds(10).toMillis();
```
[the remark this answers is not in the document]
rmatharu: +1

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -72,7 +96,10 @@ public ContainerManager(SamzaApplicationState samzaApplicationState, ClusterReso
    */
   void handleContainerLaunch(SamzaResourceRequest request, String preferredHost, SamzaResource allocatedResource,
       ResourceRequestState resourceRequestState, ContainerAllocator allocator) {
-    if (this.standbyContainerManager.isPresent()) {
+    if (getControlActionMetadata(request.getProcessorId()).isPresent()) {
```
rmatharu: expose a suitable method e.g., isControlActionRequest(request) (perhaps a wrapper on getControlActionMetadata), 
It'd make the logic more understanbable and later if you need to add additional validations (e.g., timestamp based) one can add it there.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -97,7 +126,13 @@ void handleContainerLaunch(SamzaResourceRequest request, String preferredHost, S
    */
   void handleContainerStop(String processorId, String containerId, String preferredHost, int exitStatus,
       Duration preferredHostRetryDelay, ContainerAllocator containerAllocator) {
-    if (standbyContainerManager.isPresent()) {
+    if (getControlActionMetadata(processorId).isPresent()) {
+      LOG.info("Setting the container with processorId {} stopped to be true because of existing container placement action", processorId);
+      Object activeContainerStopped = getControlActionMetadata(processorId).get().getActiveContainerStopped();
+      synchronized (activeContainerStopped) {
+        activeContainerStopped.notifyAll();
```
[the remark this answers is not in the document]
rmatharu: +1 expose single synchronized methods for reads and writes to the control-action-metadata.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -72,7 +96,10 @@ public ContainerManager(SamzaApplicationState samzaApplicationState, ClusterReso
    */
   void handleContainerLaunch(SamzaResourceRequest request, String preferredHost, SamzaResource allocatedResource,
       ResourceRequestState resourceRequestState, ContainerAllocator allocator) {
-    if (this.standbyContainerManager.isPresent()) {
+    if (getControlActionMetadata(request.getProcessorId()).isPresent()) {
+      handleContainerAllocationForExistingControlAction(request.getProcessorId(), allocator, resourceRequestState,
```
rmatharu: Is this simply checkControlActionAndRunStreamProcessor ?
similar to checkStandbyConstraintsAndRunStreamProcessor.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
rmatharu: What are we trying to accomplish by having this notify here?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -97,7 +126,13 @@ void handleContainerLaunch(SamzaResourceRequest request, String preferredHost, S
    */
   void handleContainerStop(String processorId, String containerId, String preferredHost, int exitStatus,
       Duration preferredHostRetryDelay, ContainerAllocator containerAllocator) {
-    if (standbyContainerManager.isPresent()) {
+    if (getControlActionMetadata(processorId).isPresent()) {
+      LOG.info("Setting the container with processorId {} stopped to be true because of existing container placement action", processorId);
+      Object activeContainerStopped = getControlActionMetadata(processorId).get().getActiveContainerStopped();
```
rmatharu: boolean?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -109,17 +144,26 @@ void handleContainerStop(String processorId, String containerId, String preferre
...
+   * @param processorId logical id of the container eg 1,2,3
    * @param containerId last known id of the container deployed
    * @param preferredHost host on which container is requested to be deployed
    * @param containerAllocator allocator for requesting resources
    */
   void handleContainerLaunchFail(String processorId, String containerId, String preferredHost,
       ContainerAllocator containerAllocator) {
-    if (processorId != null && standbyContainerManager.isPresent()) {
+    if (processorId != null && getControlActionMetadata(processorId).isPresent()) {
+      ContainerPlacementMetadata metaData = getControlActionMetadata(processorId).get();
```
rmatharu: See above comment on wrapper function isControlActionRequest(), 
also, can move  Optional/<ContainerPlacementMetadata> metaData = getControlActionMetadata(processorId) outside the if

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -35,23 +37,43 @@
...
    * Resource-manager, used to stop containers
    */
   private final ClusterResourceManager clusterResourceManager;
   private final SamzaApplicationState samzaApplicationState;
+  private final boolean hostAffinityEnabled;
+
+  /**
+   * Map maintaining active container placement action meta data indexed by containers processorId eg 0, 1, 2
+   */
+  private final ConcurrentHashMap<String, ContainerPlacementMetadata> actions;
```
rmatharu: Please add comments here to indicate, why it is indexed by processorId and not containerID?
Is it to ensure there is a single request inflight per processorId at all times?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -161,4 +237,140 @@ void handleExpiredRequestWithHostAffinityEnabled(String processorId, String pref
...
+  /**
+   * A valid container placement action is only issued for a running processor with a valid processor id
+   */
+  private ContainerPlacementStatus checkValidControlAction(String processorId, String destinationHost) {
+    String errorMessagePrefix =
+        String.format("ControlAction to move or restart container with processor id %s to host %s is rejected due to",
+            processorId, destinationHost);
+    Boolean invalidAction = false;
+    String errorMessage = null;
+    if (standbyContainerManager.isPresent()) {
```
rmatharu: if currentHost == destinationHost, is that also not a bad request?

samza-core/src/main/java/org/apache/samza/clustermanager/container/placements/ContainerPlacementMetadata.java
```diff
@@ -0,0 +1,89 @@
...
+package org.apache.samza.clustermanager.container.placements;
+
+import java.util.HashSet;
+import java.util.Set;
+import org.apache.samza.clustermanager.SamzaResourceRequest;
+
+/**
+ * Maintains the state for any control action like move or restart container issued externally
+ */
+public class ContainerPlacementMetadata {
```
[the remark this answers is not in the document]
rmatharu: +1, 
@Sanil15  it'd be much more readable and understandable, if all accesses to metadata were to synchronized and read-written from a specific single methods.

samza-core/src/main/java/org/apache/samza/clustermanager/container/placements/ContainerPlacementMetadata.java
```diff
@@ -0,0 +1,89 @@
...
+  // Logical container id 0,1,2,3,
+  private final String processorId;
```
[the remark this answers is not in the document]
rmatharu: I think we migrated the rest of the AM code to call 0,1,2, as processorID (inline with standalone i believe), and container_e01_123123123_121321243123 as containerId.

samza-core/src/main/java/org/apache/samza/clustermanager/container/placements/ContainerPlacementStatus.java
```diff
@@ -0,0 +1,47 @@
...
+package org.apache.samza.clustermanager.container.placements;
+
+
+/**
+ * Encapsulates state related to progress of a control action which should be passed between the
+ * {@link org.apache.samza.clustermanager.ContainerManager} and any external controllers for reporting state of an action
+ */
+public class ContainerPlacementStatus {
+  public enum StatusCode {
+    CREATED, BAD_REQUEST, ACCEPTED, IN_PROGRESS, SUCCEEDED, FAILED
```
rmatharu: Can you add meaning for each of these states?
Created = request has been recorded but no action has been taken.
Bad-request = request has been recorded, processed, and deemed a bad due to ......
Accepted = request has been recorded, and a resource request for it has been made??
In-progress = resource request has been met ??
succeeded = complete?
failed = ?????
Do we need all of these?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
Sanil15: ContainerAllocator thread used to wait and get notified by the ClusterResourceManager callback thread on stop notification receiving, but we are gonna change to do busy looping with a sleep timeout so when later we parallelize container allocation, the allocation threads would not need to be blocked by this change.

samza-api/src/main/java/org/apache/samza/container/placements/ContainerPlacementMessage.java
```diff
@@ -0,0 +1,145 @@
...
+  // Logical container Id 0, 1, 2
+  protected final String processorId;
+  // Destination host where container is desired to be moved
+  protected final String destinationHost;
+  // Optional request expiry which acts as a timeout for any resource request to
+  protected final Long requestExpiry;
+  // Status of the current request
+  protected final StatusCode statusCode;
+
+  protected ContainerPlacementMessage(UUID uuid, String applicationId, String processorId, String destinationHost,
```
Sanil15: Note to reviewers: I have copied these classes from an independent PR: https://github.com/apache/samza/pull/1227, feel free to do a review there and skip it here

samza-api/src/main/java/org/apache/samza/container/placements/ContainerPlacementRequestMessage.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.UUID;
+
+
+/**
+ * Encapsulates the request sent from the external controller to the JobCoordinator to take a container placement action
+ */
+public class ContainerPlacementRequestMessage extends ContainerPlacementMessage {
+
+  public ContainerPlacementRequestMessage(UUID uuid, String applicationId, String processorId, String destinationHost, Long requestExpiry) {
```
rmatharu: requestExpirty vs. requestTimeout vs. requestExpiryTimeout
Please pick one and use the same throughput this PR.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
```diff
@@ -110,7 +110,8 @@
    */
   protected final ResourceRequestState resourceRequestState;
   /**
-   * Tracks the expiration of a request for resources.
+   * Tracks the configured expiration of a resource request defaults to {@code ClusterManagerConfig#CLUSTER_MANAGER_REQUEST_TIMEOUT_MS}
+   * is specified or {@code ClusterManagerConfig#DEFAULT_CONTAINER_REQUEST_TIMEOUT_MS} otherwise
    */
```
rmatharu: and 
```requestExpiryTimeout for a ContainerPlacement request can be specified in the request itself```

File: samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java [hunk 24]
```diff
@@ -174,13 +175,7 @@ public void run() {
...
    *
    * If the requested host is not available, the thread checks to see if the request has expired. If it has expired
-   * then two cases are handled separately
-   *
-   * Case 1: host-affinity is enabled, looks for allocated resouces on ANY_HOST and issues a container start if available,
-   *         otherwise issues an ANY_HOST request
-   * Case 2: host-affinity is disabled, expired requests are not handled, allocator waits for cluster manager to issue
-   *         resources
-   *         TODO: SAMZA-2330 Hadle expired request for host affinity disabled case
+   * please refer to {@code containerManager#handleContainerLaunch} for constraints to handle expired requests
```
rmatharu: Please update comment to provide details on what this method is doing, no forward pointers.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
```diff
@@ -195,31 +190,31 @@ void assignResourceRequests() {
...
           state.matchedResourceRequests.incrementAndGet();
         }
 
-        containerManager.handleContainerLaunch(request, preferredHost, peekAllocatedResource(preferredHost), resourceRequestState, this);
+        boolean waitForNextAction =
+            containerManager.handleContainerLaunch(request, preferredHost, peekAllocatedResource(preferredHost),
+                resourceRequestState, this);
+
+        if (waitForNextAction) {
+          break;
```
rmatharu: dont you need a "continue;" here?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -62,33 +83,67 @@ public ContainerManager(SamzaApplicationState samzaApplicationState, ClusterReso
...
+   * TODO: SAMZA-2399: Investigate & configure a timeout for container stop if needed
    *
    * @param request pending request for the preferred host
    * @param preferredHost preferred host to start the container
    * @param allocatedResource resource allocated from {@link ClusterResourceManager}
    * @param resourceRequestState state of request in {@link ContainerAllocator}
    * @param allocator to request resources from @{@link ClusterResourceManager}
+   *
+   * @return true if the allocator thread supposed to sleep for sometime before checking the stop of active container
+   *         for an undergoing container placement action, false otherwise
```
rmatharu: This return semantics is kinda confusing.
Could this method instead, 
return true if the launch is complete, false otherwise?

The caller can then use the fact that launch is not-complete and check again (however it wants).

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
```diff
@@ -195,31 +190,31 @@ void assignResourceRequests() {
...
-        // Found allocated container on preferredHost
-        LOG.info("Found an available container for Processor ID: {} on the host: {}", processorId, preferredHost);
-
         // Needs to be only updated when host affinity is enabled
         if (hostAffinityEnabled) {
           state.matchedResourceRequests.incrementAndGet();
         }
 
-        containerManager.handleContainerLaunch(request, preferredHost, peekAllocatedResource(preferredHost), resourceRequestState, this);
+        boolean waitForNextAction =
```
rmatharu: Consider
```boolean containerLaunchComplete = ```
```// if the launch is not complete (e.g., if the launch is for moving a container but the previous container incarnation is not completely stopped yet), we sleep for ?? and try again.```

File: samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java (hunk 25 above)
Comment (rmatharu) [inline samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java]: comments

File: samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java [hunk 28]
```diff
@@ -416,19 +411,32 @@ public void stop() {
...
     }
     return requestExpired;
   }
 
+  /**
+   * Determines the request expiry timeout. Container placement actions like move, restarts can optionally override
+   * request expiry timeout. Otherwise it defaults to { @code requestExpiryTimeout }
+   */
+  private long getRequestTimeout(SamzaResourceRequest request) {
+    Optional<Duration> controlActionRequestExpirytimeout = containerManager.getActionExpiryTimeout(request.getProcessorId());
```
rmatharu: Consider using 
```private Duration getRequestTimeout(SamzaResourceRequest request) {```

File: samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java [hunk 29]
```diff
@@ -161,4 +262,140 @@ void handleExpiredRequestWithHostAffinityEnabled(String processorId, String pref
...
+    if (hasActiveContainerPlacementAction(requestMessage.getProcessorId())) {
+      return getPlacementActionMetadata(requestMessage.getProcessorId()).get();
+    }
+    return null;
+  }
+
+  public Optional<Duration> getActionExpiryTimeout(String processorId) {
+    return hasActiveContainerPlacementAction(processorId) ?
+        getPlacementActionMetadata(processorId).get().getRequestActionExpiryTimeout() : Optional.empty();
+  }
```
rmatharu: This method is really confusing, it takes as input a processorId then checks if there is action-metadata present on that processorId (which btw is no-longer to be deleted?), then gets the timeout on that timeout. 

Instead maybe the caller can just use getContainerPlacementAction(processorId) to check if there is an inflight action for the SamzaResourceRequest (which is identified by a unique requestId)?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -62,33 +83,67 @@ public ContainerManager(SamzaApplicationState samzaApplicationState, ClusterReso
...
    * @param allocator to request resources from @{@link ClusterResourceManager}
+   *
+   * @return true if the allocator thread supposed to sleep for sometime before checking the stop of active container
+   *         for an undergoing container placement action, false otherwise
    */
-  void handleContainerLaunch(SamzaResourceRequest request, String preferredHost, SamzaResource allocatedResource,
+  boolean handleContainerLaunch(SamzaResourceRequest request, String preferredHost, SamzaResource allocatedResource,
       ResourceRequestState resourceRequestState, ContainerAllocator allocator) {
+
+    if (hasActiveContainerPlacementAction(request.getProcessorId())) {
```
rmatharu: Replace with 
```hasActiveContainerPlacementAction(request.getRequestId())```

File: samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java [hunk 31]
```diff
@@ -62,33 +83,67 @@ public ContainerManager(SamzaApplicationState samzaApplicationState, ClusterReso
...
+      String processorId = request.getProcessorId();
+      ContainerPlacementMetadata actionMetaData = getPlacementActionMetadata(processorId).get();
```
rmatharu: Status status = actionMetaData.getContainerStatus()

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -62,33 +83,67 @@ public ContainerManager(SamzaApplicationState samzaApplicationState, ClusterReso
...
   /**
-   * Handles the action to be taken after the container has been stopped.
+   * Handles the action to be taken after the container has been stopped. If stop was issued due to existing control
+   * action, mark the container as stopped, otherwise
+   *
    * Case 1. When standby is enabled, refer to {@link StandbyContainerManager#handleContainerStop} to check constraints.
    * Case 2. When standby is disabled there are two cases according to host-affinity being enabled
-   *    Case 2.1. When host-affinity is enabled resources are requested on host where container was last seen
+   *    Case 1.1. When host-affinity is enabled resources are requested on host where container was last seen
    *    Case 2.2. When host-affinity is disabled resources are requested for ANY_HOST
```
rmatharu: 1.2

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -97,7 +152,10 @@ void handleContainerLaunch(SamzaResourceRequest request, String preferredHost, S
    */
   void handleContainerStop(String processorId, String containerId, String preferredHost, int exitStatus,
       Duration preferredHostRetryDelay, ContainerAllocator containerAllocator) {
-    if (standbyContainerManager.isPresent()) {
+    if (hasActiveContainerPlacementAction(processorId)) {
```
rmatharu: ```hasActiveContainerPlacementAction()``` should be defined either on the containerId or on requestId, since a processor can have multiple container-placement-actions during its lifetime.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -109,17 +167,28 @@ void handleContainerStop(String processorId, String containerId, String preferre
...
   void handleContainerLaunchFail(String processorId, String containerId, String preferredHost,
       ContainerAllocator containerAllocator) {
-    if (processorId != null && standbyContainerManager.isPresent()) {
+    if (processorId != null && hasActiveContainerPlacementAction(processorId)) {
+      ContainerPlacementMetadata metaData = getPlacementActionMetadata(processorId).get();
+      // Issue a request to start the container on source host
+      String sourceHost = hostAffinityEnabled ? metaData.getSourceHost() : ResourceRequestState.ANY_HOST;
+      handleControlActionFailure(processorId, metaData,
+          String.format("failed to start container on destination host %s, attempting to start on source host %s",
+              preferredHost, sourceHost));
```
rmatharu: Update container-placement-metadata before making a resourceRequest, so that the allocator knows what to do with the resource?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -132,9 +201,27 @@ void handleContainerLaunchFail(String processorId, String containerId, String pr
...
+   * @param processorId logical processor id of container 0,1,2
+   */
+  void handleContainerLaunchSuccess(String processorId) {
+    if (hasActiveContainerPlacementAction(processorId)) {
+      ContainerPlacementMetadata metadata = getPlacementActionMetadata(processorId).get();
+      metadata.setActionStatus(ContainerPlacementMessage.StatusCode.SUCCEEDED,
+              "Successfully completed the container placement action");
+      LOG.info("Marking the container placement action success {}", metadata);
+    }
+  }
```
rmatharu: This method is called from a single place, please inline with caller in ContainerProcessManager

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -143,9 +230,23 @@ void handleContainerLaunchFail(String processorId, String containerId, String pr
...
    */
   @VisibleForTesting
-  void handleExpiredRequestWithHostAffinityEnabled(String processorId, String preferredHost,
+  void handleExpiredRequestForControlActionOrHostAffinityEnabled(String processorId, String preferredHost,
       SamzaResourceRequest request, ContainerAllocator allocator, ResourceRequestState resourceRequestState) {
     boolean resourceAvailableOnAnyHost = allocator.hasAllocatedResource(ResourceRequestState.ANY_HOST);
+
+    if (hasActiveContainerPlacementAction(processorId)) {
+      resourceRequestState.cancelResourceRequest(request);
+      handleControlActionFailure(processorId, getPlacementActionMetadata(processorId).get(),
```
rmatharu: The handleControlActionFailure should be called markControlActionFailed, all it does is update the state.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -143,9 +230,23 @@ void handleContainerLaunchFail(String processorId, String containerId, String pr
...
+          "failed the Control action because request for resources to ClusterManager expired");
+      return;
+    }
+
+    // Host affinity disabled wait for Resource manager to return resources
+    if (!hostAffinityEnabled) {
+      return;
+    }
```
rmatharu: If host affinity is not enabled, why would this method be called?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -161,4 +262,140 @@ void handleExpiredRequestWithHostAffinityEnabled(String processorId, String pref
...
+   *                       is sent to resource manager for any host)
+   * @param containerAllocator to request physical resources
+   */
+  public void registerContainerPlacementAction(ContainerPlacementRequestMessage requestMessage, ContainerAllocator containerAllocator) {
+    LOG.info("Received ControlAction request to move or restart container with processor id {} to host {}",
+        requestMessage.getProcessorId(), requestMessage.getDestinationHost());
+    String processorId = requestMessage.getProcessorId();
+    String destinationHost = requestMessage.getDestinationHost();
+    Pair<ContainerPlacementMessage.StatusCode, String> actionStatus =
+        checkValidControlAction(processorId, destinationHost, requestMessage.getUuid());
```
rmatharu: Why does this return a pair, when actionStatus.getValue is never used?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -161,4 +262,140 @@ void handleExpiredRequestWithHostAffinityEnabled(String processorId, String pref
...
+
+    // Request is bad just update the response on message & return
+    if (actionStatus.getKey() == ContainerPlacementMessage.StatusCode.BAD_REQUEST) {
+      return;
+    }
+
+    SamzaResource currentResource = samzaApplicationState.runningProcessors.get(processorId);
+    LOG.info("Processor ID: {} matched a active container with deployment ID: {} running on host: {}", processorId,
```
rmatharu: matched an active container

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -161,4 +262,140 @@ void handleExpiredRequestWithHostAffinityEnabled(String processorId, String pref
...
+        currentResource.getContainerId(), currentResource.getHost());
+
+    if (destinationHost.equals(ANY_HOST) || !hostAffinityEnabled) {
```
rmatharu: Should just be 
```if (!hostAffinityEnabled) {```

File: samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java [hunk 41]
```diff
@@ -161,4 +262,140 @@ void handleExpiredRequestWithHostAffinityEnabled(String processorId, String pref
...
+          ResourceRequestState.ANY_HOST);
+      destinationHost = ANY_HOST;
+    }
+
+    SamzaResourceRequest resourceRequest = containerAllocator.getResourceRequest(processorId, destinationHost);
+    Optional<Duration> requestExpiry =
+        requestMessage.getRequestExpiry() != null ? Optional.of(Duration.ofMillis(requestMessage.getRequestExpiry())) : Optional.empty();
+    ContainerPlacementMetadata actionMetaData =
+        new ContainerPlacementMetadata(requestMessage.getUuid(), processorId, currentResource.getContainerId(),
+            currentResource.getHost(), destinationHost, requestExpiry);
```
rmatharu: Use Optional.ofNullable

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -161,4 +262,140 @@ void handleExpiredRequestWithHostAffinityEnabled(String processorId, String pref
...
+    // Record the resource request for monitoring
+    actionMetaData.setActionStatus(ContainerPlacementMessage.StatusCode.IN_PROGRESS, "Preferred Resources requested");
+    actionMetaData.recordResourceRequest(resourceRequest);
+    actions.put(processorId, actionMetaData);
+    containerAllocator.issueResourceRequest(resourceRequest);
+    LOG.info("Container Placement action with metadata {} and issued a request for resources in progress",
+        actionMetaData);
+  }
+
+  @VisibleForTesting
```
rmatharu: javadoc?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -161,4 +262,140 @@ void handleExpiredRequestWithHostAffinityEnabled(String processorId, String pref
...
+  /**
+   * A ContainerPlacementAction is only active if it is either CREATED, ACCEPTED or IN_PROGRESS
+   */
+  private Boolean hasActiveContainerPlacementAction(String processorId) {
+    Optional<ContainerPlacementMetadata> metadata = getPlacementActionMetadata(processorId);
+    if (metadata.isPresent()) {
+      ContainerPlacementMessage.StatusCode actionStatus = metadata.get().getActionStatus();
+      return (actionStatus == ContainerPlacementMessage.StatusCode.CREATED) ||
+          (actionStatus == ContainerPlacementMessage.StatusCode.ACCEPTED) ||
+          (actionStatus == ContainerPlacementMessage.StatusCode.IN_PROGRESS);
```
rmatharu: switch

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -612,4 +613,14 @@ private String getPendingProcessorId(String containerId) {
...
     containerManager.handleContainerStop(processorId, containerId, preferredHost, exitStatus, preferredHostRetryDelay, containerAllocator);
   }
+
+  public Map<String, SamzaResource> getCurrentRunningContainers() {
+    return state.runningProcessors;
+  }
+
+  // processorId 0,1,2 = current yarn container id not samza container id
+  // preferredHost = fully qualified hostname
+  public void requestMoveContainer(String processorId, String preferredHost) {
```
rmatharu: containerPlacement vs. moveContainer vs. ContainerMovement 
pick one and use throughout this PR

samza-core/src/test/java/org/apache/samza/clustermanager/TestContainerProcessManager.java
```diff
@@ -625,7 +631,8 @@ public void testRerequestOnAnyHostIfContainerStartFails() throws Exception {
     configMap.putAll(getConfig());
     MockClusterResourceManagerCallback callback = new MockClusterResourceManagerCallback();
     MockClusterResourceManager clusterResourceManager = new MockClusterResourceManager(callback, state);
-    ContainerManager containerManager = new ContainerManager(state, clusterResourceManager, false);
+    ContainerManager containerManager = new ContainerManager(state, clusterResourceManager,
+        Boolean.valueOf(config.get(ClusterManagerConfig.HOST_AFFINITY_ENABLED)), false);
```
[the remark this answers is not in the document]
mynameborat: `getBoolean` has an overload that takes a default to be returned in case of null

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
```diff
@@ -203,23 +213,31 @@ void assignResourceRequests() {
...
+            containerManager.handleContainerLaunch(request, preferredHost, peekAllocatedResource(preferredHost),
+                resourceRequestState, this);
+
+        /**
+         * Some Container launch requests are due to Container Placement actions like move, restarts. Under those
+         * circumstances a container launch needs to wait for the previous container incarnation to stop. In this scenario
+         * the allocator thread needs to sleep and recheck the stop again
+         */
+        if (!containerLaunchComplete) {
+          break;
```
rmatharu: Do you want to log this outcome?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -62,33 +83,67 @@ public ContainerManager(SamzaApplicationState samzaApplicationState, ClusterReso
...
+   *         placement actions
    */
-  void handleContainerLaunch(SamzaResourceRequest request, String preferredHost, SamzaResource allocatedResource,
+  boolean handleContainerLaunch(SamzaResourceRequest request, String preferredHost, SamzaResource allocatedResource,
       ResourceRequestState resourceRequestState, ContainerAllocator allocator) {
+    LOG.info("Found an available container for Processor ID: {} on the host: {}", request.getProcessorId(), preferredHost);
+    if (hasActiveContainerPlacementAction(request.getProcessorId())) {
+      String processorId = request.getProcessorId();
+      ContainerPlacementMetadata actionMetaData = getPlacementActionMetadata(processorId).get();
+      ContainerPlacementMetadata.ContainerStatus actionStatus = actionMetaData.getContainerStatus();
```
rmatharu: Consider switch?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -97,7 +152,10 @@ void handleContainerLaunch(SamzaResourceRequest request, String preferredHost, S
    */
   void handleContainerStop(String processorId, String containerId, String preferredHost, int exitStatus,
       Duration preferredHostRetryDelay, ContainerAllocator containerAllocator) {
-    if (standbyContainerManager.isPresent()) {
+    if (hasActiveContainerPlacementAction(processorId)) {
+      LOG.info("Setting the for container state with processorId {} to be stopped because of existing container placement action", processorId);
+      getPlacementActionMetadata(processorId).get().setContainerStatus(ContainerPlacementMetadata.ContainerStatus.STOPPED);
```
rmatharu: is there any scenario where getPlacementActionMetadata(processorId).get() fails?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -143,9 +233,26 @@ void handleContainerLaunchFail(String processorId, String containerId, String pr
...
+      resourceRequestState.cancelResourceRequest(request);
+      markContainerPlacementActionFailed(getPlacementActionMetadata(processorId).get(),
+          "failed the Control action because request for resources to ClusterManager expired");
+      return;
+    }
+
+    // Case 2. When host affinity is disabled wait for cluster resource manager return resources
+    if (!hostAffinityEnabled) {
+      return;
+    }
```
rmatharu: Please add a comment to indicate that this is only to maintain parity with current-behavior, and TODO/SAMZA-XXXX: Enable any-host-retry for host-affinity-disabled scenarios.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
Sanil15: This will cause log spam when allocator thread wakes up to check if the active container has stopped.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
Sanil15: because of && in the first if I cannot use it for all three cases, maybe for the other two if its worth it?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
Sanil15: The metadata for a previously completed action is not deleted until a new action is accepted, new is only accepted when an existing action is complete, so this should not fail.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
Sanil15: added!

samza-api/src/main/java/org/apache/samza/container/placements/ContainerPlacementMessage.java
```diff
@@ -0,0 +1,147 @@
...
+  // Destination host where container is desired to be moved
+  protected final String destinationHost;
+  // Optional request expiry which acts as a timeout for any resource request to cluster manager
+  protected final Optional<Duration> requestExpiryTimeout;
+  // Status of the current request
+  protected final StatusCode statusCode;
+
+  protected ContainerPlacementMessage(UUID uuid, String applicationId, String processorId, String destinationHost,
+      Duration requestExpiryTimeout, StatusCode statusCode) {
+    Preconditions.checkNotNull(uuid, "uuid cannot be null");
```
kw2542: Use java.util.Objects.requireNonNull

samza-api/src/main/java/org/apache/samza/container/placements/ContainerPlacementMessage.java
```diff
@@ -0,0 +1,147 @@
...
+    Preconditions.checkNotNull(applicationId, "applicationId cannot be null");
+    Preconditions.checkNotNull(processorId, "processorId cannot be null");
+    Preconditions.checkNotNull(destinationHost, "destinationHost cannot be null");
+    Preconditions.checkNotNull(statusCode, "statusCode cannot be null");
+    this.uuid = uuid;
```
kw2542: Can be simplified to:

this.uuid = Preconditions.checkNotNull(uuid, "")

or 

this.uuid = Objects.requireNonNull(uuid, "")

samza-api/src/main/java/org/apache/samza/container/placements/ContainerPlacementMessage.java
```diff
@@ -0,0 +1,147 @@
...
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    ContainerPlacementMessage that = (ContainerPlacementMessage) o;
+    return uuid.equals(that.uuid) && getApplicationId().equals(that.getApplicationId()) && getProcessorId().equals(
+        that.getProcessorId()) && getDestinationHost().equals(that.getDestinationHost())
+        && getStatusCode() == that.getStatusCode();
```
kw2542: potential NPEs, we may use Objects.equals(uuid, that.uuid), which handles null checks.

samza-api/src/main/java/org/apache/samza/container/placements/ContainerPlacementRequestMessage.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public ContainerPlacementRequestMessage(UUID uuid, String applicationId, String processorId, String destinationHost) {
+    super(uuid, applicationId, processorId, destinationHost, StatusCode.CREATED);
+  }
+
+  @Override
+  public String toString() {
+    return "ContainerPlacementRequestMessage{" + "uuid=" + uuid + ", applicationId='" + applicationId + '\''
+        + ", processorId='" + processorId + '\'' + ", destinationHost='" + destinationHost + '\'' + ", requestExpiryTimeout="
+        + requestExpiryTimeout + ", statusCode=" + statusCode + "} " + super.toString();
```
kw2542: it will be easier to read to using String.format()

samza-api/src/main/java/org/apache/samza/container/placements/ContainerPlacementRequestMessage.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Override
+  public int hashCode() {
+    return super.hashCode();
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    return super.equals(obj);
+  }
```
kw2542: Do we need to override if they just calls super?

samza-api/src/main/java/org/apache/samza/container/placements/ContainerPlacementMessage.java
Sanil15: The constructor does not allow nulls, then I don't need this write

samza-api/src/main/java/org/apache/samza/container/placements/ContainerPlacementRequestMessage.java
Sanil15: Already used here: https://github.com/apache/samza/pull/1227

samza-api/src/main/java/org/apache/samza/container/placements/ContainerPlacementRequestMessage.java
Sanil15: removing it here: https://github.com/apache/samza/pull/1227

samza-api/src/main/java/org/apache/samza/container/placements/ContainerPlacementMessage.java
Sanil15: Sure but for consistency with other code I prefer to use Preconditions, https://stackoverflow.com/questions/34646554/java-util-objects-requirenonnull-vs-preconditions-checknotnull

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
```diff
@@ -202,12 +202,9 @@ void assignResourceRequests() {
       String preferredHost = hostAffinityEnabled ? request.getPreferredHost() : ResourceRequestState.ANY_HOST;
       Instant requestCreationTime = request.getRequestTimestamp();
 
-      LOG.info("Handling assignment request for Processor ID: {} on host: {}.", processorId, preferredHost);
+      LOG.info("Handling assignment for request {}", request);
       if (hasAllocatedResource(preferredHost)) {
 
-        // Found allocated container on preferredHost
-        LOG.info("Found an available container for Processor ID: {} on the host: {}", processorId, preferredHost);
```
Sanil15: note: hasAllocatedResource(...) already logs this line so this is not userful

Sanil15:
Completed the manual testing, here is the link to test-spec: https://docs.google.com/spreadsheets/d/1v-fw0pHxKRobGkALDCno4FuPCsBhdepQ86vIGLHWu54/edit#gid=0&range=16:16

@rmatharu please merge once you have verified testing results

PR #1220, opened 2019-11-18
SAMZA-2386: Get store names should return correct store names in the presence of side inputs

mynameborat:
**Symptom**: Enabling side inputs for low level jobs using configuration will throw an exception during job start up.
**Cause**: `getStoreNames()` uses pattern matching to infer store names from the config. The pattern matching conflicts with `side.inputs.processor.factory` and returns incorrect store names.
**Fix**: Handle side inputs processors factory prefix correctly within `getStoreNames()`
**Tests**: Added unit test for `getStoreNames()`
**API Change**: None
**Usage Instructions**: None
**Upgrade Instructions**: None


prateekm:
@mynameborat  Thanks for the quick fix. @rmatharu Can you review and merge?

PR #1222, opened 2019-11-21
SAMZA-2396: Support dynamic addition of jars in ReflectionUdfResolver.

shanthoosh:


samza-sql/src/main/java/org/apache/samza/sql/udf/ReflectionBasedUdfResolver.java
```diff
@@ -49,20 +54,26 @@
   private static final Logger LOG = LoggerFactory.getLogger(ReflectionBasedUdfResolver.class);
 
   private static final String CONFIG_PACKAGE_PREFIX = "samza.sql.udf.resolver.package.prefix";
```
atoomula: remove this ?

shanthoosh: Discussed offline. We'd planned to proceed with having both the configurations for now.

PR #1223, opened 2019-11-22
SAMZA-2397: Updating gauge-val function on newGauge on same metric name

rmatharu:
Updating gauge-val function on newGauge on same metric name.

**API Change**: MetricRegistryMap. newGauge will no overwrite any existing gauge, if one exists for the same group and gauge.name.

**Symptom**: RocksDB-specific metric value like estimate-table-readers-mem, cur-size-active-mem-table, etc, are not being emitted after Samza 1.1; only a 0 value is emitted instead.

**Cause**: Samza 1.1 uses an optimization where a RocksDBStore is created first in BulkLoad mode for optimizing bootstrap write, after bootstrapping is complete the store is closed (to trigger compaction), and a new store is created in ReadWrite mode. The rocksdb metrics are calculated using a function which checks if the store is open (see https://github.com/apache/samza/pull/588 by @xinyuiscool , SAMZA-1778), otherwise it emits 0.
See RocksdbKeyValueStore#95.
However, because the metric-gauge-object is not updated after the store is recreated, the gauge keeps checking the first (closed) store and emits 0.

**Fix**: Added logic that ensures that the second store's metric overrides the previous gauge for the same name, and logs a message.

**Tests**: Added unit test, test with a samza job that emits non-zero RocksDB metrics.


samza-core/src/main/scala/org/apache/samza/metrics/MetricsRegistryMap.scala
```diff
@@ -51,7 +51,10 @@ class MetricsRegistryMap(val name: String) extends ReadableMetricsRegistry with
 
   def newGauge[T](group: String, gauge: Gauge[T]) = {
     debug("Adding new gauge %s %s %s." format (group, gauge.getName, gauge))
-    putAndGetGroup(group).putIfAbsent(gauge.getName, gauge)
+    if (putAndGetGroup(group).containsKey(gauge.getName)) {
+      info("Updating existing gauge %s %s %s" format (group, gauge.getName, gauge))
+    }
+    putAndGetGroup(group).put(gauge.getName, gauge)
```
rmatharu: Should we do this for all other metric types -- timers, etc, and methods to create them. 
Last metric overrides -- Is this not the desired semantics?

prateekm:
@bkonold can you review?

samza-api/src/main/java/org/apache/samza/metrics/MetricsRegistry.java
```diff
@@ -64,6 +64,17 @@
...
 
+  /**
+   * Register an existing {@link org.apache.samza.metrics.Gauge}
+   * @param group Group for this Gauge
+   * @param value Initial value for the Gauge
+   * @param <T> Type the Gauge will be wrapping
+   * @param overrideExistingGauge Overwrite any existing gauges present for the same group and gauge
+   * @return Gauge was registered
+   */
+  <T> Gauge<T> newGauge(String group, Gauge<T> value, Boolean overrideExistingGauge);
```
bkonold: Are we defining a new method here with the additional parameter to avoid a large code change in adding a boolean to the definition above?

Wondering whether the two methods will coexist long term or if we see a way to unify eventually.

rmatharu: Yes, making overriding to be the default behavior requires testing to ensure that it doesnt break existing metrics; I can work on that separately. 
Although I tried with a test-job and everything looked ok.

bkonold: If leaving them separate is what is needed to ensure the P2 timeline and we plan for the two living along side one another to be temporary, maybe we can open a ticket and indicate in a comment when / under what conditions clean up can happen?

samza-core/src/main/java/org/apache/samza/metrics/MetricGroup.java
```diff
@@ -62,6 +62,15 @@ public T getValue() {
     });
   }
 
+  public <T> Gauge<T> newGauge(String name, final ValueFunction<T> valueFunc, Boolean overwrite) {
```
bkonold: There seems to be a mixing of terminology here vs MetricsRegistry (overwrite vs overrideExistingGauge). All we're doing is passing the arg here on to the registry; is the difference in naming intentional?

samza-api/src/main/java/org/apache/samza/metrics/MetricsRegistry.java
rmatharu: Done, added details in https://issues.apache.org/jira/browse/SAMZA-2417

samza-core/src/main/java/org/apache/samza/metrics/MetricGroup.java
rmatharu: done, made naming consistent.

samza-kv-rocksdb/src/main/scala/org/apache/samza/storage/kv/RocksDbKeyValueStore.scala
```diff
@@ -98,7 +98,7 @@ object RocksDbKeyValueStore extends Logging {
             rocksDb.getProperty(property)
           } else {
             "0"
-          }
+          }, true // overrite-gauge = true, to accomodate for store-close and re-open (after bulk-load)
```
bkonold: nit: typo

"// override gauge to accomodate..." ?

samza-core/src/main/scala/org/apache/samza/metrics/MetricsHelper.scala
```diff
@@ -51,6 +51,12 @@ trait MetricsHelper {
     })
   }
 
+  def newGauge[T](name: String, value: () => T, overwrite : Boolean) = {
```
bkonold: let's have this param name consistent with the others as well?

samza-api/src/main/java/org/apache/samza/metrics/MetricsRegistry.java
prateekm: I'd prefer to not add a new variant here and update the existing behavior instead. It seems to fall into "fixing undefined behavior" territory, in that I don't think a reasonable user expectation would be to retain the old metric. We can fix this here and call it out in the "backwards incompatible / API changes" section of the PR description.

@cameronlee314 What do you think?

rmatharu: I was debating doing that, but I'm not sure if it'd break some existing metrics-subsystem which relies on us retaining the first object rather than the last?

rmatharu: @prateekm I rollbacked all most of the changes; we now just retain the newest metric. 
can you approve it if it looks ok, so we can merge.

samza-core/src/main/scala/org/apache/samza/metrics/MetricsRegistryMap.scala
```diff
@@ -50,8 +52,10 @@ class MetricsRegistryMap(val name: String) extends ReadableMetricsRegistry with
   }
 
   def newGauge[T](group: String, gauge: Gauge[T]) = {
-    debug("Adding new gauge %s %s %s." format (group, gauge.getName, gauge))
-    putAndGetGroup(group).putIfAbsent(gauge.getName, gauge)
+    if (putAndGetGroup(group).containsKey(gauge.getName)) {
+      info("Updating existing gauge %s %s %s" format (group, gauge.getName, gauge))
```
prateekm: Should be debug.

samza-core/src/main/scala/org/apache/samza/metrics/MetricsRegistryMap.scala
```diff
@@ -19,6 +19,8 @@
 
 package org.apache.samza.metrics
 
+import java.lang
```
prateekm: Remove.

PR #1226, opened 2019-11-27
SAMZA-2406: Add ConfigLoader interface and default PropertiesConfigLo…

kw2542:
Add ConfigLoader interface and default PropertiesConfigLoader implementation, which will be used on AM to fetch configs.

The existing ConfigFactory interface will be deprecated and removed as we will not fetch configs on start up script run-app.sh anymore.

Design:

Changes:
  1. Add ConfigLoader interface which will be used in job coordinator to retrieve config from.
  2. Add PropertiesConfigLoader which is a default implementation of ConfigLoader that reads a local properties file given the file path.

API Changes:
Upgrade Instructions:
Usage Instructions:

Tests:
  1. Add TestPropertiesConfigLoader to unit test PropertiesConfigLoader.


prateekm:
@bharathkk and @abhishekshivanna offered to look at this.

@kw2542 Can you update usage/upgrade instructions in the description with changes required for using this new feature? This information will be collected for the release notes and changelog for the next release.

Let's also fix the commit title (wrapping).

samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
```diff
@@ -0,0 +1,35 @@
...
+
+package org.apache.samza.config;
+
+import java.util.Map;
+
+
+/**
+ * Build a {@link org.apache.samza.config.Config}
+ */
+public interface ConfigLoader {
```
mynameborat: Can we add more descriptive java docs? 
1. It will be helpful if you can talk about how and when this interface is used.
2. What if `ConfigLoader` implementors want to handle dependency injection? Do we need to provide a factory for the config loader implementors?

samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
```diff
@@ -0,0 +1,35 @@
...
+  /**
+   * Build a specific Config.
+   * @param properties Map containing necessary information for this loader to get config.
+   * @return Newly constructed Config.
+   */
+  Config getConfig(Map<String, String> properties);
```
mynameborat: It will better to have APIs that have strong and well-defined contracts. Taking in a Map<String, String> seems very fragile and makes it harder for the implementors. 

Additionally, what do you think about passing this information during construction as opposed to method parameter since these are fundamental properties for the config loaders to work in the first place. With that, we can fail fast during construction and it is also easy to reason about the failures in getConfig as operational failures instead of validation checks.

samza-core/src/main/scala/org/apache/samza/config/loaders/PropertiesConfigLoader.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import scala.collection.JavaConverters._
+import org.apache.samza.SamzaException
+import org.apache.samza.config.{Config, ConfigLoader, MapConfig}
+import org.apache.samza.util.Logging
+
+/**
+ * Load config from a properties file with given path. "path" will be given in the input properties map.
+ */
+class PropertiesConfigLoader extends ConfigLoader with Logging {
```
mynameborat: Since this is a new class, can you write this in Java instead of Scala?

samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
mynameborat: Do we need an init or other lifecycle method for loaders that are potentially heavy weight (e.g. KafkaConfigLoader)?
It will offload the `getConfig()` method and also makes it cheap to be invoked multiple times if need be.

samza-core/src/main/scala/org/apache/samza/config/loaders/PropertiesConfigLoader.scala
mynameborat: Where is this class used? Can you point me to some examples of how this will be used if you have implemented it? I noticed in your design the branching in presence of loader class, but didn't see any examples or pseudo code on how where config loader is instantiated and how it is used to get the config.

samza-core/src/test/scala/org/apache/samza/config/loaders/TestPropertiesConfigLoader.scala
```diff
@@ -0,0 +1,49 @@
...
+package org.apache.samza.config.loaders
+
+import java.util.Collections
+
+import org.apache.samza.SamzaException
+import org.apache.samza.config.MapConfig
+import org.junit.Assert._
+import org.junit.Test
+
+class TestPropertiesConfigLoader {
```
mynameborat: Same as above. Please rewrite new classes in Java instead of Scala.

mynameborat:
> @bharathkk and @abhishekshivanna offered to look at this.
> 
> @kw2542 Can you update usage/upgrade instructions in the description with changes required for using this new feature? This information will be collected for the release notes and changelog for the next release.
> 
> Let's also fix the commit title (wrapping).

@kw2542 Can you also update the API changes in the description and add more details about this API (internal vs external) etc

abhishekshivanna:
@kw2542 , could you also clarify as part of this patch the reason we need this new implementation over the existing `config-factory` (eg: PropertiesConfigFactory) and `config-path` implementation ? Is this meant to be a replacement for the way we do config loading today ? If so, should we deprecate the old way ?

samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
kw2542: Added more descriptions.

samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
kw2542: As discussed, switched to Config as the input and move it to constructor.

samza-core/src/main/scala/org/apache/samza/config/loaders/PropertiesConfigLoader.scala
kw2542: Moved to Java.

samza-core/src/test/scala/org/apache/samza/config/loaders/TestPropertiesConfigLoader.scala
kw2542: Moved to Java.

samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
kw2542: During the life cycle of a job, config loader is supposed to invoked once during start up, the final config will be persisted in AM's memory as well as coordinator stream.

samza-core/src/main/scala/org/apache/samza/config/loaders/PropertiesConfigLoader.scala
kw2542: Please see the design doc for how it will be used: https://cwiki.apache.org/confluence/display/SAMZA/SEP-23%3A+Simplify+Job+Runner

kw2542:
> @kw2542 , could you also clarify as part of this patch the reason we need this new implementation over the existing `config-factory` (eg: PropertiesConfigFactory) and `config-path` implementation ? Is this meant to be a replacement for the way we do config loading today ? If so, should we deprecate the old way ?

kw2542:
> @kw2542 , could you also clarify as part of this patch the reason we need this new implementation over the existing `config-factory` (eg: PropertiesConfigFactory) and `config-path` implementation ? Is this meant to be a replacement for the way we do config loading today ? If so, should we deprecate the old way ?

Updated both the PR description as well as SEP itself.

ConfigLoader will replace ConfigFactory interface, which will be deleted in the following PRs.

kw2542:
Discard to start a new PR clean.

PR #1227, opened 2019-11-27
SAMZA-2404: [SEP-22] Container Placement Handler for dispatching container placement messages between metastore and JobCoordinator

Sanil15:
**Design:** [SEP-22: Container Placements in Samza](https://cwiki.apache.org/confluence/display/SAMZA/SEP-22%3A+Container+Placements+in+Samza)

**Changes:** : Following abilities are added as a part of this PR:

1. Container placement handler which is a stateless handler that periodically dispatches messages between the Metadata store and the Container Placement Service APIs running as a part of Job Coordinator
2. ContainerPlacementRequestMessage & ContainerPlacementResponseMessage are two POJO used to encapsulate request & response messages for container placement actions
3. Custom ObjectMapper for serializing and deserializing these messages is added as a part of ContainerPlacementMessageObjectMapper

**Tests:** 
1. Unit tests are added to test message serialization & deserialization
2. Unit tests to test ContainerPlacementHandler
2. Manual testing has been done to see reads / writes from meta-store

**API Changes:**: 
1. Adds two new public exposed API class: ContainerPlacementRequestMessage & ContainerPlacementResponseMessage in samza-api
2. ContainerPlacementRequestMessage are messages used to issue container placement actions, one can write these messages to meta-store using Meta-store writer 
3. ContainerPlacementResponseMessage are messages used as a response to container placement actions, Job Coordinator writes these messages to Meta-store so external controllers can read response to the current actions

**Upgrade Instructions:** Existing jobs with a persisted metastore will be able to ignore these newly populated messages because they have a different namespace

**Usage Instructions:** An external controller / tool (ex: CoordinatorStreamWriter) would be able to issue such actions by writing to metastore (currently: CoordinatorStream)


Sanil15:
@dnishimura @bharathkk @shanthoosh @rmatharu Please have a look

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -97,6 +100,8 @@
 
   // The ContainerManager manages control actions for both active & standby containers
   private final ContainerManager containerManager;
+  private ContainerPlacementHandler containerPlacementHandler;
+  private Thread containerPlacementHandlerThread;
```
rmatharu: Please use ExecutorService executorService = Executors.newSingleThreadExecutor(

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -185,18 +196,22 @@ public ContainerProcessManager(Config config, SamzaApplicationState state, Metri
...
     this.allocatorThread = new Thread(this.containerAllocator, "Container Allocator Thread");
+    this.containerPlacementHandler = new ContainerPlacementHandler(metadataStore, this, clusterManagerConfig.getClusterManagerContainerPlacementHandlerSleepMs());
+    this.containerPlacementHandlerThread = new Thread(containerPlacementHandler, "Container Placement Handler Thread");
     LOG.info("Finished container process manager initialization");
   }
 
   public boolean shouldShutdown() {
-    LOG.debug("ContainerProcessManager state: Completed containers: {}, Configured containers: {}, Are there too many failed containers: {}, Is allocator thread alive: {}",
-      state.completedProcessors.get(), state.processorCount, jobFailureCriteriaMet ? "yes" : "no", allocatorThread.isAlive() ? "yes" : "no");
+    LOG.debug(
```
rmatharu: Maybe this should be  LOG.info

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementHandler.java
```diff
@@ -0,0 +1,287 @@
...
+          String.format("ContainerPlacementRequestMessage might have been not written to metastore %s", message), ex);
+    }
+  }
+
+  /**
+   * Writes a {@link ContainerPlacementResponseMessage} to the underlying metastore. This method should be used by external controllers
+   * to issue a request to JobCoordinator
+   * @param message
+   */
+  public void writeContainerPlacementResponseMessage(ContainerPlacementResponseMessage message) {
```
rmatharu: Is this intended for use by the client?

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementHandler.java
```diff
@@ -0,0 +1,287 @@
...
+  /**
+   * Reads a {@link ContainerPlacementRequestMessage} from the underlying metastore
+   * @param processorId key of the message, logical processor id of a samza container 0,1,2
+   * @return ContainerPlacementRequestMessage is its present
+   */
+  public Optional<ContainerPlacementRequestMessage> readContainerPlacementRequestMessage(String processorId) {
+    Preconditions.checkState(isRunning, "Underlying metadata store not available");
+    Preconditions.checkNotNull(processorId, "processorId cannot be null");
+
+    byte[] messageBytes = requestStore.get(processorId);
```
rmatharu: Do you know what would the actual key for a request/response message look like?

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementMessageObjectMapper.java
```diff
@@ -0,0 +1,113 @@
...
+      }
+      ContainerPlacementMessage message = null;
+      if (subType.equals(ContainerPlacementRequestMessage.class.getName())) {
+        message = new ContainerPlacementRequestMessage(uuid, applicationId, processorId, destinationHost, requestExpiry);
+      } else if (subType.equals(ContainerPlacementResponseMessage.class.getName())) {
+        String responseMessage = node.get("responseMessage").getTextValue();
+        message = new ContainerPlacementResponseMessage(uuid, applicationId, processorId, destinationHost, requestExpiry,
+            statusCode, responseMessage);
+      }
+      return message;
```
rmatharu: Could we add this to SamzaObjectMapper and use SamzaObjectMapper instead?

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
```diff
@@ -113,6 +113,12 @@
   public static final String CLUSTER_MANAGER_SLEEP_MS = "cluster-manager.jobcoordinator.sleep.interval.ms";
   private static final int DEFAULT_CLUSTER_MANAGER_SLEEP_MS = 1000;
 
+  /**
+   * The container placement handler sleeps for a configurable time before checking for now ContainerPlacement messages
+   */
+  public static final String CLUSTER_MANAGER_CONTAINER_PLACEMENT_HANDLER_SLEEP_MS = "cluster-manager.container-placement-handler.sleep.interval.ms";
+  private static final int DEFAULT_CLUSTER_MANAGER_CONTAINER_PLACEMENT_HANDLER_SLEEP_MS = 5000;
```
rmatharu: Do we expect different jobs to tune this timeout differently?
If not, i'd suggest removing this config

Sanil15: No for the first pass we can read defaults!

samza-api/src/main/java/org/apache/samza/container/placement/ContainerPlacementMessage.java
```diff
@@ -0,0 +1,144 @@
...
+
+  public String getDestinationHost() {
+    return destinationHost;
+  }
+
+  public StatusCode getStatusCode() {
+    return statusCode;
+  }
+
+  public Long getRequestExpiry() {
```
dnishimura: Please document the time unit (seconds or millis)

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -242,6 +259,8 @@ public void stop() {
 
     // Shutdown allocator thread
     containerAllocator.stop();
+    // Shutdown container placement handler thread
+    containerPlacementHandler.stop();
```
dnishimura: Would you need the similar thread join pattern for the container placement handler thread , similar to the container allocator thread?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
```diff
@@ -562,6 +582,17 @@ void onResourceCompletedWithUnknownStatus(SamzaResourceStatus resourceStatus, St
...
 
+  /**
+   * Registers a ContainerPlacement action, this method is invoked by ContainerPlacementHandler. {@link ContainerProcessManager}
+   * needs to intercept container placement actions between ContainerPlacementHandler and {@link ContainerManager} to avoid
+   * cyclic dependency between {@link ContainerManager} and {@link ContainerAllocator} on each other
+   *
+   * @param requestMessage request containing details of the desited container placement action
+   */
+  public void registerContainerPlacementAction(ContainerPlacementRequestMessage requestMessage) {
+    // Call the ContainerManager#registerContainerPlacementAction
```
dnishimura: Is this a `TODO:` for a follow on PR?

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementMessageObjectMapper.java
dnishimura: Yup SamzaObjectMapper is super bloated. IMO, decoupling is good, especially in this specific domain.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java
rmatharu: final

rmatharu: Maybe we can use the 
```afterExecute(Runnable r, Throwable t)``` method of ExecutorService?

File: samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementMessageObjectMapper.java (hunk 5 above)
Comment (rmatharu) [reply samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementMessageObjectMapper.java]: The reason its bloated is because the Mixins and custom serializers and deserializers have been defined in the same file. 
You can define yours in separate files but link them in SamzaObjectMapper since this is a part of Samza.

File: samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java (hunk 1 above)
Comment (Sanil15) [reply samza-core/src/main/java/org/apache/samza/clustermanager/ContainerProcessManager.java]: nope I have to use the semaphore for propagating the exception from CPH thread to main thread

**Comment by Sanil15** [comment]
Addressing @prateekm feedback & adding the following changes
1. Changed how ContainerPlacementHandler wires in JC, wiring CPH with ClusterBasedJobCoordinator
2. Refactored meta-store read/write utils from ContainerPlacementHandler to a separate entity ContainerPlacementUtil, wired it to ClusterBasedJobCoordinator
3. ContainerPlacementUtil is gonna be wired into ContainerPlacementService in a follow-up PR and can be used by outside clients to request container placement actions.
4. Metastore key for Container placement actions is UUID + Message type (req/res) and both request & response messages now belong to the same Namespace in metastore
5. Documentation & other improvements

File: samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementMessageObjectMapper.java (hunk 5 above)
Comment (Sanil15) [reply samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementMessageObjectMapper.java]: "You can define yours in separate files but link them in SamzaObjectMapper since this is a part of Samza."

SamzaObjectMapper class is only used for JobModel serde, I think maybe that is the reason @dnishimura also did not link Startpoint object mappers with it.

File: samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java [hunk 10]
```diff
@@ -316,6 +345,19 @@ private void onShutDown() {
...
 
+  private void shutDownContainerPlacementHandlerAndUtils() {
+    // Shutdown container placement handler
+    containerPlacementHandler.stop();
+    try {
+      containerPlacementHandlerThread.join();
+      LOG.info("Stopped container placement handler thread");
+      containerPlacementUtil.stop();
+    } catch (InterruptedException ie) {
+      LOG.error("Container Placement handler thread join threw an interrupted exception", ie);
```
dnishimura: This error log assumes `containerPlacementUtil.stop()` will never throw an exception. Is that true?

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementMessageObjectMapper.java
```diff
@@ -0,0 +1,119 @@
...
+    public void serialize(ContainerPlacementMessage value, JsonGenerator jsonGenerator, SerializerProvider provider)
+        throws IOException {
+      Map<String, Object> containerPlacementMessageMap = new HashMap<String, Object>();
+      if (value instanceof ContainerPlacementRequestMessage) {
+        containerPlacementMessageMap.put("subType", ContainerPlacementRequestMessage.class.getSimpleName());
+      } else if (value instanceof ContainerPlacementResponseMessage) {
+        containerPlacementMessageMap.put("subType", ContainerPlacementResponseMessage.class.getSimpleName());
+        containerPlacementMessageMap.put("responseMessage",
+            ((ContainerPlacementResponseMessage) value).getResponseMessage());
+      }
```
dnishimura: Would you want to add an else check for unexpected subtypes of `ContainerPlacementMessage`? Also may apply to the `deserialize` method below.

samza-core/src/test/java/org/apache/samza/clustermanager/container/placement/TestContainerPlacementObjectMapper.java
```diff
@@ -0,0 +1,75 @@
...
+import java.time.Duration;
+import java.util.Random;
+import java.util.UUID;
+import org.apache.samza.container.placement.ContainerPlacementMessage;
+import org.apache.samza.container.placement.ContainerPlacementRequestMessage;
+import org.apache.samza.container.placement.ContainerPlacementResponseMessage;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
dnishimura: nit: use non-wildcard imports.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -150,6 +152,13 @@
    */
   private final Optional<StreamRegexMonitor> inputStreamRegexMonitor;
 
+  /**
+   * Container placement action dispatcher and util
+   */
+  private final ContainerPlacementUtil containerPlacementUtil;
```
rmatharu: Either rename this to ContainerPlacementMetastoreReader or make all its methods static so that it is really a Util. 
A Util should not be instantiated like this and have a complete lifecycle.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -150,6 +152,13 @@
...
+  private final ContainerPlacementHandler containerPlacementHandler;
+  private final Thread containerPlacementHandlerThread;
```
rmatharu: // containerPlacement thread that reads and writes placement-requests and responses to the metastore

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -264,9 +279,15 @@ public void run() {
       partitionMonitor.ifPresent(StreamPartitionCountMonitor::start);
       inputStreamRegexMonitor.ifPresent(StreamRegexMonitor::start);
 
+      // ContainerPlacementHandler thread has to start after the cpm is started
+      LOG.info("Starting the container placement handler thread");
+      containerPlacementUtil.start();
```
rmatharu: see comment above

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -264,9 +279,15 @@ public void run() {
...
+      containerPlacementHandlerThread.start();
+
       boolean isInterrupted = false;
```
rmatharu: should this be volatile?

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
rmatharu: +1, perhaps use finally to do a clean guaranteed cleanup?

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementHandler.java
```diff
@@ -0,0 +1,84 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(ContainerPlacementHandler.class);
+  private static final int DEFAULT_CLUSTER_MANAGER_CONTAINER_PLACEMENT_HANDLER_SLEEP_MS = 5000;
+
+  /**
+   * {@link ContainerProcessManager} needs to intercept container placement actions between ContainerPlacementHandler and
+   * {@link org.apache.samza.clustermanager.ContainerManager} to avoid cyclic dependency between
+   * {@link org.apache.samza.clustermanager.ContainerManager} and {@link org.apache.samza.clustermanager.ContainerAllocator}
+   */
+  private final ContainerProcessManager containerProcessManager;
+  private final ContainerPlacementUtil containerPlacementUtil;
```
rmatharu: Dont quite understand why both ClusterBasedJC and this class hold on to ContainerPlacementUtil

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementHandler.java
```diff
@@ -0,0 +1,84 @@
...
+          containerProcessManager.registerContainerPlacementAction(message);
+        }
+        Thread.sleep(DEFAULT_CLUSTER_MANAGER_CONTAINER_PLACEMENT_HANDLER_SLEEP_MS);
+      } catch (InterruptedException e) {
+        LOG.warn("Got InterruptedException in ContainerPlacementHandler thread.", e);
+        Thread.currentThread().interrupt();
+      } catch (Exception e) {
+        LOG.error(
+            "Got unknown Exception while registering ContainerPlacement actions in ContainerPlacementHandler thread.",
+            e);
```
rmatharu: indent

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementUtil.java
```diff
@@ -0,0 +1,273 @@
...
+   */
+  public void deleteAllContainerPlacementMessages(UUID uuid) {
+    deleteContainerPlacementRequestMessage(uuid);
+    deleteContainerPlacementResponseMessage(uuid);
+  }
+
+  /**
+   * Deletes all {@link ContainerPlacementMessage}
+   */
+  public void deleteAllContainerPlacementMessages() {
```
rmatharu: This method is not invoked from anywhere, then why do we need it?

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementUtil.java
```diff
@@ -0,0 +1,273 @@
...
+
+  /**
+   * Deletes a {@link ContainerPlacementResponseMessage} if present identified by the key {@code processorId}
+   * @param uuid uuid of the request
+   */
+  public void deleteContainerPlacementResponseMessage(UUID uuid) {
+    Preconditions.checkState(!stopped, "Underlying metadata store not available");
+    Preconditions.checkNotNull(uuid, "uuid cannot be null");
+    containerPlacementMessageStore.delete(toContainerPlacementMessageKey(uuid, ContainerPlacementResponseMessage.class));
+  }
```
rmatharu: If these two are invoked from within the class, should they be private?

samza-core/src/test/java/org/apache/samza/clustermanager/TestContainerAllocatorWithoutHostAffinity.java
```diff
@@ -54,7 +54,8 @@
   private final JobModelManager jobModelManager = JobModelManagerTestUtil.getJobModelManager(config, 1,
       new MockHttpServer("/", 7777, null, new ServletHolder(DefaultServlet.class)));
 
-  private final SamzaApplicationState state = new SamzaApplicationState(jobModelManager);
+  private final SamzaApplicationState state =
+      new SamzaApplicationState(jobModelManager);
```
rmatharu: undo

samza-api/src/main/java/org/apache/samza/container/placement/ContainerPlacementMessage.java
```diff
@@ -0,0 +1,152 @@
...
+
+  protected ContainerPlacementMessage(UUID uuid, String deploymentId, String processorId, String destinationHost,
+      Duration requestExpiry, StatusCode statusCode, long timestamp) {
+    Preconditions.checkNotNull(uuid, "uuid cannot be null");
+    Preconditions.checkNotNull(deploymentId, "deploymentId cannot be null");
+    Preconditions.checkNotNull(processorId, "processorId cannot be null");
+    Preconditions.checkNotNull(destinationHost, "destinationHost cannot be null");
+    Preconditions.checkNotNull(statusCode, "statusCode cannot be null");
+    Preconditions.checkNotNull(timestamp, "Timestamp of the message cannot be null");
+    this.uuid = uuid;
```
kw2542: We can do null check and assignment in one line:

this.uuid = Preconditions.checkNotNull(uuid, "uuid cannot be null");

same for others.

samza-api/src/main/java/org/apache/samza/container/placement/ContainerPlacementMessage.java
```diff
@@ -0,0 +1,152 @@
...
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    ContainerPlacementMessage message = (ContainerPlacementMessage) o;
+    return getUuid().equals(message.getUuid()) && getDeploymentId().equals(message.getDeploymentId())
+        && getProcessorId().equals(message.getProcessorId()) && getDestinationHost().equals(
+        message.getDestinationHost());
```
kw2542: is it expected that we are ignoring requestExpiry, statusCode and timestamp when comparing? Same for other classes.

samza-api/src/main/java/org/apache/samza/container/placement/ContainerPlacementRequestMessage.java
```diff
@@ -0,0 +1,62 @@
...
+  public String toString() {
+    final StringBuilder sb = new StringBuilder("ContainerPlacementRequestMessage{");
+    sb.append(" UUID: ").append(uuid);
+    sb.append(", Processor ID: ").append(processorId);
+    sb.append(", deploymentId='").append(deploymentId).append('\'');
+    sb.append(", destinationHost='").append(destinationHost).append('\'');
+    sb.append(", requestExpiry=").append(requestExpiry);
+    sb.append(", statusCode=").append(statusCode);
+    sb.append(", timestamp=").append(timestamp);
+    sb.append('}');
```
kw2542: How about we have a static class variable for this format, and toString() just do

String.format(FORMAT, uuid, processorId, deploymentId .....)

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
kw2542: +1

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -264,9 +279,15 @@ public void run() {
...
 
-      while (!containerProcessManager.shouldShutdown() && !checkAndThrowException() && !isInterrupted) {
+      while (!containerProcessManager.shouldShutdown() && !checkAndThrowException() && !isInterrupted
+          && checkContainerPlacementHandlerThreadIsAlive()) {
```
kw2542: nit: I personally prefer to have a new line for each condition if they can not fit in one line:

!containerProcessManager.shouldShutdown()
&& !checkAndThrowException()
&& !isInterrupted
&& checkContainerPlacementHandlerThreadIsAlive()

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementMessageObjectMapper.java
```diff
@@ -0,0 +1,119 @@
...
+      long timestamp = node.get("timestamp").getLongValue();
+      Duration requestExpiry =
+          node.get("requestExpiry") == null ? null : Duration.ofMillis(node.get("requestExpiry").getLongValue());
+      ContainerPlacementMessage.StatusCode statusCode = null;
+      UUID uuid = UUID.fromString(node.get("uuid").getTextValue());
+      for (ContainerPlacementMessage.StatusCode code : ContainerPlacementMessage.StatusCode.values()) {
+        if (code.name().equals(node.get("statusCode").getTextValue())) {
+          statusCode = code;
+        }
+      }
```
kw2542: It is not easy to read/understand, can we do:

statusCode = ContainerPlacementMessage.StatusCode.valueOf(node.get("statusCode").getTextValue())

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementUtil.java
```diff
@@ -0,0 +1,273 @@
...
+   * @param timestamp timestamp of the request
+   * @return uuid generated for the request
+   */
+  public UUID writeContainerPlacementRequestMessage(String deploymentId, String processorId, String destinationHost,
+      Duration requestExpiry, long timestamp) {
+    Preconditions.checkState(!stopped, "Underlying metadata store not available");
+    Preconditions.checkNotNull(deploymentId, "deploymentId should not be null");
+    Preconditions.checkNotNull(processorId, "processorId not be null");
+    Preconditions.checkNotNull(destinationHost, "destinationHost should not be null");
+    Preconditions.checkNotNull(timestamp, "timestamp should not be null");
```
kw2542: Constructor of ContainerPlacementRequestMessage handles null checks, do we need to do another null check here?

samza-api/src/main/java/org/apache/samza/container/placement/ContainerPlacementMessage.java
kw2542: We can use Preconditions.checkNotNull but in one line, e.g.

this.uuid = Preconditions.checkNotNull(uuid, "uuid cannot be null");

samza-api/src/main/java/org/apache/samza/container/placement/ContainerPlacementMessage.java
kw2542: I understand that in reality, each message will have its own UUID, but in terms of comparing java object itself, should we compare all fields? It may be very confusing if two ContainerPlacementMessage equals but their status code is different.

samza-api/src/main/java/org/apache/samza/container/placement/ContainerPlacementRequestMessage.java
kw2542: I will leave this to your judgement, both are fine to me.

samza-api/src/main/java/org/apache/samza/container/placement/ContainerPlacementMessage.java
Sanil15: I misread this, sure changing it

samza-api/src/main/java/org/apache/samza/container/placement/ContainerPlacementMessage.java
Sanil15: I see your point will change it

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementMessageObjectMapper.java
```diff
@@ -0,0 +1,119 @@
...
+    SimpleModule module = new SimpleModule("ContainerPlacementModule", new Version(1, 0, 0, ""));
+    module.addSerializer(ContainerPlacementMessage.class,
+        new ContainerPlacementMessageObjectMapper.ContainerPlacementMessageSerializer());
+    module.addDeserializer(ContainerPlacementMessage.class,
+        new ContainerPlacementMessageObjectMapper.ContainerPlacementMessageDeserializer());
+    objectMapper.registerModule(module);
+    objectMapper.registerSubtypes(new NamedType(ContainerPlacementResponseMessage.class));
+    objectMapper.registerSubtypes(new NamedType(ContainerPlacementRequestMessage.class));
+    return objectMapper;
+  }
```
kw2542: Can we have a static ObjectMapper? It is thread safe and we are not changing its configuration either.

Sanil15: I suppose you mean to have a static object Mapper declared here: https://github.com/apache/samza/pull/1227/files#diff-20fafbc969e1be36e38d63017cbfc77aR58?

kw2542: Yes and no, we can change that to be a static field or getObjectMapper() can return a static singleton as well.

Since getObjectMapper() takes no argument, so it is returning a object mapper with the same configuration, we could make getObjectMapper() to return the same instance of ObjectMapper instead of relying the caller to have a static variable.

Sanil15: I see sure since we are not changing the configuration it can be a singleton

samza-api/src/main/java/org/apache/samza/container/placement/ContainerPlacementMessage.java
rmatharu: +1, Please make sure all POJO fields are in the equals and hashCode if they can be.

samza-core/src/test/java/org/apache/samza/clustermanager/container/placement/TestContainerPlacementObjectMapper.java
```diff
@@ -0,0 +1,76 @@
...
+
+  private void testContainerPlacementRequestMessage(ContainerPlacementRequestMessage requestMessage)
+      throws IOException {
+    ObjectMapper objectMapper = ContainerPlacementMessageObjectMapper.getObjectMapper();
+    ContainerPlacementMessage message =
+        objectMapper.readValue(objectMapper.writeValueAsString(requestMessage), ContainerPlacementMessage.class);
+    assertTrue(message instanceof ContainerPlacementRequestMessage);
+    ContainerPlacementRequestMessage deserializedRequest = (ContainerPlacementRequestMessage) message;
+    assertEquals(requestMessage, deserializedRequest);
+    assertEquals(deserializedRequest.getStatusCode(), ContainerPlacementMessage.StatusCode.CREATED);
```
rmatharu: Since ContainerPlacementMessage.equals  includes statusCode, this step is not needed.

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
```diff
@@ -227,7 +227,6 @@ public int getContainerRetryWindowMs() {
     }
   }
 
-
```
rmatharu: Remove change?

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementRequestAllocator.java
```diff
@@ -0,0 +1,84 @@
...
+          LOG.info("Received a container placement message {}", message);
+          containerProcessManager.registerContainerPlacementAction(message);
+        }
+        Thread.sleep(DEFAULT_CLUSTER_MANAGER_CONTAINER_PLACEMENT_HANDLER_SLEEP_MS);
+      } catch (InterruptedException e) {
+        LOG.warn("Got InterruptedException in ContainerPlacementRequestAllocator thread.", e);
+        Thread.currentThread().interrupt();
+      } catch (Exception e) {
+        LOG.error(
+            "Got unknown Exception while reading ContainerPlacementRequestMessage in ContainerPlacementRequestAllocator thread",
```
rmatharu: Well, if we know the Exception so its not really "unknown" :-)

samza-core/src/test/java/org/apache/samza/clustermanager/container/placement/TestContainerPlacementObjectMapper.java
Sanil15: sure

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
Sanil15: sure

samza-core/src/main/java/org/apache/samza/clustermanager/container/placement/ContainerPlacementRequestAllocator.java
Sanil15: removed "unknown"

PR #1228, opened 2019-12-02
Adding metric that is 0 when container-starting, 1 when started.

rmatharu:
Adding metric that is 0 when container-starting, 1 when started.

Symptom: External monitoring and control (e.g., autosizing) require a way to know when a container has started (e.g., state-restore has completed). Currently, they have to rely on restore-time or container-startup-time metrics, which are a bit tricky to use.

Cause: None

Fix: Added a new metric, a counter, which is 0 and remains at 0 after the container has started. A Gauge on the other hand will go to 0 if its not updated.

Tests: Samza test job with and without state.

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
+++ b/samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
@@ -764,6 +764,7 @@ class SamzaContainer(
         containerListener.afterStart()
       }
       metrics.containerStartupTime.set((System.nanoTime() - startTime)/1000000)
+      metrics.containerStarted.set(1)
       if (taskInstances.size > 0)
         runLoop.run
       else
--- a/samza-core/src/main/scala/org/apache/samza/container/SamzaContainerMetrics.scala
+++ b/samza-core/src/main/scala/org/apache/samza/container/SamzaContainerMetrics.scala
@@ -46,6 +46,7 @@ class SamzaContainerMetrics(
   val diskQuotaBytes = newGauge("disk-quota-bytes", Long.MaxValue)
   val executorWorkFactor = newGauge("executor-work-factor", 1.0)
   val physicalMemoryMb = newGauge[Double]("physical-memory-mb", 0.0F)
+  val containerStarted = newCounter("container-started")
 
   val taskStoreRestorationMetrics: util.Map[TaskName, Gauge[Long]] = new util.HashMap[TaskName, Gauge[Long]]()
 


prateekm:
You should use a gauge, not counter. In inGraphs, a counter is displayed as a rate (v2-v1/t2-t1).
1. That's probably not what you want.
2. Not sure what will happen if v2 < v1.

A gauge should not reset to 0 if not updated, it should retain its last updated value.

rmatharu:
> You should use a gauge, not counter. In inGraphs, a counter is displayed as a rate (v2-v1/t2-t1).
> 
> 1. That's probably not what you want.
> 2. Not sure what will happen if v2 < v1.
> 
> A gauge should not reset to 0 if not updated, it should retain its last updated value.

Fair enough, I was confused between Timers and Gauges.

prateekm:
@rmatharu Let's also add this to the metrics docs.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -764,6 +764,7 @@ class SamzaContainer(
         containerListener.afterStart()
       }
       metrics.containerStartupTime.set((System.nanoTime() - startTime)/1000000)
```
mynameborat: Can we use (`container-startup-time`) metric instead to infer if the container has started?

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
```diff
@@ -764,6 +764,7 @@ class SamzaContainer(
...
+      metrics.containerStarted.set(1)
```
mynameborat: Not related to this PR, but curious what will be the behavior of the auto sizing controller if this metric is broken. Wondering if it can differentiate between metrics broken vs container not started.

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
rmatharu: we can but its relatively harder to automate since we need to look for changes

samza-core/src/main/scala/org/apache/samza/container/SamzaContainer.scala
rmatharu: Without this (or another similar metric) the controller has no way of figuring out that a container has finished state-restore, upto which point any input-lag increase does not warrant scale-up

PR #1229, opened 2019-12-05
Adding internal config to retain non-logged stores on container start

rmatharu:


samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -74,9 +74,12 @@
   static final String INMEMORY_KV_STORAGE_ENGINE_FACTORY =
       "org.apache.samza.storage.kv.inmemory.InMemoryKeyValueStorageEngineFactory";
 
-  // Internal config to clean storeDirs of a store on container start. This is used to benchmark bootstrap performance.
+  // Internal config to clean storeDirs of a logged store on container start. This is used to benchmark bootstrap performance.
   static final String CLEAN_LOGGED_STOREDIRS_ON_START = STORE_PREFIX + "%s.clean.on.container.start";
 
+  // Internal config to clean storeDirs of a logged store on container start. This is used to benchmark bootstrap performance.
```
abhishekshivanna: /s/logged/non-logged/

samza-core/src/main/java/org/apache/samza/config/StorageConfig.java
```diff
@@ -74,9 +74,12 @@
...
+  static final String RETAIN_NONLOGGED_STOREDIRS_ON_START = STORE_PREFIX + "%s.retain.on.container.start";
```
abhishekshivanna: should we name this config to reflect the non-logged only case ? `%s.retain.nonlogged.on.container.start`

samza-core/src/main/java/org/apache/samza/storage/NonTransactionalStateTaskRestoreManager.java
```diff
@@ -132,7 +132,7 @@ private void cleanBaseDirsAndReadOffsetFiles() {
               storageManagerUtil.getTaskStoreDir(nonLoggedStoreBaseDirectory, storeName, taskModel.getTaskName(), taskModel.getTaskMode());
           LOG.info("Got non logged storage partition directory as " + nonLoggedStorePartitionDir.toPath().toString());
 
-          if (nonLoggedStorePartitionDir.exists()) {
+          if (nonLoggedStorePartitionDir.exists() || !storageConfig.getRetainNonloggedStoreDirsOnStart(storeName)) {
```
abhishekshivanna: update doc string to indicate that deletion is conditional.

prateekm:
@rmatharu please use PR description format.

mynameborat:
I'd prefer to not introduce force cleaning optional internal config for more combination of our stores properties (we already have one for logged store and now we have one for non-logged store). It makes it hard to follow up all the potential branching. 

If we still need this, can we consolidate all the flags to one configuration? Also, do we have other use cases for this outside experimental purpose?

samza-core/src/main/java/org/apache/samza/storage/NonTransactionalStateTaskRestoreManager.java
cameronlee314: Should this be `&&` instead of `||`?

mynameborat:
Is this PR still needed? Lets close it if its not.

PR #1234, opened 2019-12-10
SAMZA-2416 : Adding null-check before incrementing metrics for bytesSerialized

rmatharu:
Adding null-check before incrementing metrics for bytesSerialized

Symptom: User serde-serialization logic can choose to serialize non-null objects to a null, which implies delete the value. In this case, current code tries to increment the metrics using the size of the null and fails with an NPE.

Cause: Current code tries to increment the bytesSerialized metrics using the size of the null and fails with an NPE.

Fix: Added logic to skip metric value update if serde returns null.


rmatharu:
@bkonold Samza provided serdes dont run into this, user-defined serdes though can map non-null objects to null.

PR #1236, opened 2019-12-17
SAMZA-2419: removing duplicate cleaning of stores

bkonold:
**Issues:** Code block for cleaning logged stores according to config value is duplicated and references incorrect & incomplete pathnames.
 
**Changes:** Removing duplicate code block, and correcting the directories marked for deletion. When the clean.on.container.start config is true for a store and transactional state is enabled, the current _and_ checkpoint directories should be deleted. Previously only the current directory was marked for deletion, but at the wrong path.
 
**Tests:** Added unit test to confirm deletion of appropriate store directories

**API Changes:** None
**Usage instructions:** None

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/storage/TransactionalStateTaskRestoreManager.java
+++ b/samza-core/src/main/java/org/apache/samza/storage/TransactionalStateTaskRestoreManager.java
@@ -247,8 +247,9 @@ public class TransactionalStateTaskRestoreManager implements TaskRestoreManager 
         // if the clean.store.start config is set, delete the currentDir, restore from oldest offset to checkpointed
         if (storageEngine.getStoreProperties().isPersistedToDisk() && new StorageConfig(
           config).getCleanLoggedStoreDirsOnStart(storeName)) {
-          File currentDir = storageManagerUtil.getTaskStoreDir(nonLoggedStoreBaseDirectory, storeName, taskName, taskMode);
-          LOG.info("Marking current directory: {} for store: {} in task: {}.", currentDir, storeName, taskName);
+          File currentDir = storageManagerUtil.getTaskStoreDir(loggedStoreBaseDirectory, storeName, taskName, taskMode);
+          LOG.info("Marking current directory: {} for store: {} in task: {} for deletion due to clean.on.container.start config.",
+              currentDir, storeName, taskName);
           storeDirsToDelete.put(storeName, currentDir);
           LOG.info("Marking restore offsets for store: {} in task: {} to {}, {} ", storeName, taskName, oldestOffset, checkpointedOffset);
           storesToRestore.put(storeName, new RestoreOffsets(oldestOffset, checkpointedOffset));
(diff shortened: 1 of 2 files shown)


samza-core/src/main/java/org/apache/samza/storage/TransactionalStateTaskRestoreManager.java
```diff
@@ -243,17 +243,6 @@ static StoreActions getStoreActions(
           timeSinceLastCheckpointInMs = System.currentTimeMillis() -
               checkpointedChangelogOffset.getCheckpointId().getMillis();
         }
-      
-        // if the clean.store.start config is set, delete the currentDir, restore from oldest offset to checkpointed
-        if (storageEngine.getStoreProperties().isPersistedToDisk() && new StorageConfig(
-          config).getCleanLoggedStoreDirsOnStart(storeName)) {
-          File currentDir = storageManagerUtil.getTaskStoreDir(nonLoggedStoreBaseDirectory, storeName, taskName, taskMode);
```
bkonold: @prateekm @rmatharu shouldn't we be cleaning from loggedStoreBaseDirectory here (in the non-deleted copy of this code block)?

prateekm:
@rmatharu @bharathkk can you take a look?

samza-core/src/main/java/org/apache/samza/storage/TransactionalStateTaskRestoreManager.java
mynameborat: Good catch! I suppose the duplicate code block should for the `loggedStore` and should use the `loggedStoreBaseDirectory`. 
Do you mind updating the patch with the changes?

bkonold: This is what I suspected, just wanted to confirm with another set of eyes. I'll update the patch.

samza-core/src/main/java/org/apache/samza/storage/TransactionalStateTaskRestoreManager.java
```diff
@@ -243,30 +243,19 @@ static StoreActions getStoreActions(
...
-          LOG.info("Marking current directory: {} for store: {} in task: {}.", currentDir, storeName, taskName);
-          storeDirsToDelete.put(storeName, currentDir);
-          LOG.info("Marking restore offsets for store: {} in task: {} to {}, {} ", storeName, taskName, oldestOffset, checkpointedOffset);
-          storesToRestore.put(storeName, new RestoreOffsets(oldestOffset, checkpointedOffset));
-          return;
-        }
 
         // if the clean.store.start config is set, delete the currentDir, restore from oldest offset to checkpointed
         if (storageEngine.getStoreProperties().isPersistedToDisk() && new StorageConfig(
```
mynameborat: I'd suggest you also check if the store is logged store explicitly. It is possible logged store directory and non-logged store directory happens to be the same directory and we will by-mistake clear the stores that are non-logged.

bkonold: good catch.

@rmatharu was your intention in PR https://github.com/apache/samza/pull/1200 for both logged and non-logged stores to be clear based on this config?

bkonold: to recap: @mynameborat 's question was with regard to side inputs, which are persisted and non-logged but are not to be deleted. this is a non-issue since side inputs are not hooked in through TransactionalStateTaskRestoreManager but instead TaskSideInputStorageManager. @mynameborat feel to comment if i am misunderstanding our discussion.

mynameborat:
Looks good to me. Can you add a unit test?

bkonold:
@mynameborat done

bkonold:
> Will merge it once travis and checks pass.

@mynameborat must have been a transient failure before; i didn't see any specific test failures, and the next run passed.

PR #1237, opened 2019-12-18
SAMZA-2406: Add ConfigLoader, ConfigLoaderFactory interface and default PropertiesConfigLoader as well as PropertiesConfigLoaderFactory implementations.

kw2542:
Add ConfigLoader, ConfigLoaderFactory interface and default PropertiesConfigLoader as well as PropertiesConfigLoaderFactory implementations, which will be used on AM to fetch configs. It will also be used on LocalApplicationRunner to fetch configs as well.

The existing ConfigFactory interface will be deprecated and removed as we do not fetch configs on start up script run-app.sh anymore.

Design:

Changes:

1. Add ConfigLoader interface which will be used in job coordinator to retrieve config from.
2. Add ConfigLoaderFactory interface which will be used to create an instance of ConfigLoader.
3. Add PropertiesConfigLoader which is a default implementation of ConfigLoader that reads a local properties file given the file path.
4. Add PropertiesConfigLoaderFactory which is a default implementation of ConfigLoaderFactory that reads properties file path from start up config and creates a PropertiesConfigLoader with the provided path.

API Changes:

Upgrade Instructions:
Usage Instructions:

Tests:

Add TestPropertiesConfigLoader to unit test PropertiesConfigLoader.

### DIFF
--- a/samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
+++ b/samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
@@ -1,0 +1,27 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.config;
+
+/**
+ * The primary means of fetching {@link org.apache.samza.config.Config} on Application Master during start up.
+ */
+public interface ConfigLoader {
+  Config getConfig();
+}
--- a/samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
+++ b/samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
@@ -20,7 +20,8 @@
 package org.apache.samza.config;
 
 /**
- * The primary means of fetching full job {@link org.apache.samza.config.Config} on cluster based job coordinator during start up.
+ * The primary means of fetching full job {@link org.apache.samza.config.Config} on
+ * LocalApplicationRunner and ClusterBasedJobCoordinator during start up.
  */
 public interface ConfigLoader {
   Config getConfig();
--- a/samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
+++ b/samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
@@ -20,7 +20,7 @@
 package org.apache.samza.config;
 
 /**
- * The primary means of fetching {@link org.apache.samza.config.Config} on Application Master during start up.
+ * The primary means of fetching full job {@link org.apache.samza.config.Config} on cluster based job coordinator during start up.
  */
 public interface ConfigLoader {
   Config getConfig();
--- a/samza-api/src/main/java/org/apache/samza/config/ConfigLoaderFactory.java
+++ b/samza-api/src/main/java/org/apache/samza/config/ConfigLoaderFactory.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.config;
+
+/**
+ * The factory for an {@link ConfigLoader} instance to loader job config.
+ */
+public interface ConfigLoaderFactory {
+  /**
+   * Get an instance of {@link ConfigLoader}
+   *
+   * @param config start up config which specifies properties needed for the loader to load config.
+   * @return a config loader instance.
+   */
+  ConfigLoader getLoader(Config config);
+}
--- a/samza-api/src/main/java/org/apache/samza/config/ConfigLoaderFactory.java
+++ b/samza-api/src/main/java/org/apache/samza/config/ConfigLoaderFactory.java
@@ -20,13 +20,13 @@
 package org.apache.samza.config;
 
 /**
- * The factory for an {@link ConfigLoader} instance to loader job config.
+ * The factory for an {@link ConfigLoader} instance to load full job config.
  */
 public interface ConfigLoaderFactory {
   /**
    * Get an instance of {@link ConfigLoader}
    *
-   * @param config start up config which specifies properties needed for the loader to load config.
+   * @param config start up config which specifies properties needed for the loader to load full job config.
    * @return a config loader instance.
    */
   ConfigLoader getLoader(Config config);
--- a/samza-api/src/main/java/org/apache/samza/config/MapConfig.java
+++ b/samza-api/src/main/java/org/apache/samza/config/MapConfig.java
@@ -25,7 +25,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Properties;
 import java.util.Set;
 
 /**
--- a/samza-api/src/main/java/org/apache/samza/config/MapConfig.java
+++ b/samza-api/src/main/java/org/apache/samza/config/MapConfig.java
@@ -42,17 +42,6 @@ public class MapConfig extends Config {
     this(Collections.singletonList(map));
   }
 
-  /**
-   * Build a {@link MapConfig} from {@link Properties}
-   *
-   * @param properties to build MapConfig
-   */
-  public MapConfig(Properties properties) {
-    this.map = new HashMap<>();
-    // Per Properties JavaDoc, all its keys and values are of type String
-    properties.forEach((key, value) -> this.map.put(key.toString(), value.toString()));
-  }
-
   public MapConfig(List<Map<String, String>> maps) {
     this.map = new HashMap<>();
     for (Map<String, String> m: maps) {
--- a/samza-api/src/main/java/org/apache/samza/config/MapConfig.java
+++ b/samza-api/src/main/java/org/apache/samza/config/MapConfig.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 import java.util.Set;
 
 /**
@@ -39,6 +40,17 @@ public class MapConfig extends Config {
 
   public MapConfig(Map<String, String> map) {
     this(Collections.singletonList(map));
+  }
+
+  /**
+   * Build a {@link MapConfig} from {@link Properties}
+   *
+   * @param properties to build MapConfig
+   */
+  public MapConfig(Properties properties) {
+    this.map = new HashMap<>();
+    // Per Properties JavaDoc, all its keys and values are of type String
+    properties.forEach((key, value) -> this.map.put(key.toString(), value.toString()));
   }
 
   public MapConfig(List<Map<String, String>> maps) {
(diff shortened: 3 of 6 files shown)


samza-core/src/main/java/org/apache/samza/config/loaders/PropertiesConfigLoader.java
```diff
@@ -0,0 +1,61 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+
+public class PropertiesConfigLoader implements ConfigLoader {
+  private static final Logger LOG = LoggerFactory.getLogger(PropertiesConfigLoader.class);
+
+  private final String path;
+
+  public PropertiesConfigLoader(String path) {
```
cameronlee314: Instead of using `String` for the type, consider using something like `URI`. It provides a bit more structure/validation. The existing `ConfigFactory` uses `URI`, and it seems like that is something useful to keep.

samza-core/src/test/java/org/apache/samza/config/loaders/TestPropertiesConfigLoader.java
```diff
@@ -0,0 +1,52 @@
...
+    assertEquals("bar", config.get("foo"));
+  }
+
+  @Test
+  public void testCanNotReadWithoutPath() {
+    try {
+      ConfigLoader loader = new PropertiesConfigLoaderFactory().getLoader(new MapConfig());
+      loader.getConfig();
+      fail("should have gotten a samza exception");
+    } catch (SamzaException e) {
```
cameronlee314: Since you don't need to check anything in the exception, you can use `@Test(expected = SamzaException.class)` and then you don't need the try-catch.

samza-core/src/test/java/org/apache/samza/config/loaders/TestPropertiesConfigLoader.java
```diff
@@ -0,0 +1,52 @@
...
+package org.apache.samza.config.loaders;
+
+import java.util.Collections;
+import org.apache.samza.SamzaException;
+import org.apache.samza.config.Config;
+import org.apache.samza.config.ConfigLoader;
+import org.apache.samza.config.MapConfig;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
cameronlee314: One of the conventions in Samza code is to avoid using wildcard imports. Could you please do individual imports?

samza-core/src/main/java/org/apache/samza/config/loaders/PropertiesConfigLoader.java
```diff
@@ -0,0 +1,61 @@
...
+      Properties props = new Properties();
+
+      props.load(in);
+      in.close();
+
+      LOG.debug("got config {} from config {}", props, path);
+
+      return new MapConfig(props);
+    } catch (IOException e) {
+      throw new SamzaException("Failed to read from");
```
cameronlee314: Was this the full exception message you intended? It looks like you meant to include the path or something in there.

samza-core/src/main/java/org/apache/samza/config/loaders/PropertiesConfigLoader.java
```diff
@@ -0,0 +1,61 @@
...
+  @Override
+  public Config getConfig() {
+    try {
+      InputStream in = new FileInputStream(path);
+      Properties props = new Properties();
+
+      props.load(in);
+      in.close();
+
+      LOG.debug("got config {} from config {}", props, path);
```
cameronlee314: Did you mean `got config {} from path {}`?

samza-api/src/main/java/org/apache/samza/config/MapConfig.java
```diff
@@ -41,6 +42,17 @@ public MapConfig(Map<String, String> map) {
     this(Collections.singletonList(map));
   }
 
+  /**
+   * Build a {@link MapConfig} from {@link Properties}
+   *
+   * @param properties to build MapConfig
+   */
+  public MapConfig(Properties properties) {
```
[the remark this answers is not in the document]
kw2542: Updated

samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
```diff
@@ -0,0 +1,27 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.config;
+
+/**
+ * The primary means of fetching {@link org.apache.samza.config.Config} on Application Master during start up.
```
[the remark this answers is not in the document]
kw2542: Updated

samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+public interface ConfigLoader {
+  Config getConfig();
```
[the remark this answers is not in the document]
kw2542: You are right, in the original SEP, getConfig takes in a config argument. In this PR, it is removed as we introduced ConfigLoaderFactory, which takes the config instead. 

I agree that we are not expecting Config to change between calling ConfigLoaderFactory.getLoader and calling ConfigLoader.getConfig, in which case, if exists, users should call getLoader again.

samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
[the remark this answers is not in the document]
kw2542: Clarified this is for full job config.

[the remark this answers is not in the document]
kw2542: Updated

samza-api/src/main/java/org/apache/samza/config/MapConfig.java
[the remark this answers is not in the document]
kw2542: Updated

samza-core/src/main/java/org/apache/samza/config/loaders/PropertiesConfigLoader.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.samza.SamzaException;
+import org.apache.samza.config.Config;
+import org.apache.samza.config.ConfigLoader;
+import org.apache.samza.config.MapConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static java.util.Objects.requireNonNull;
+
+
```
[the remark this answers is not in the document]
kw2542: Added

samza-core/src/test/java/org/apache/samza/config/loaders/TestPropertiesConfigLoader.java
kw2542: Updated

samza-core/src/main/java/org/apache/samza/config/loaders/PropertiesConfigLoader.java
kw2542: Nice catch.

samza-core/src/main/java/org/apache/samza/config/loaders/PropertiesConfigLoader.java
kw2542: Nice catch.

samza-core/src/test/java/org/apache/samza/config/loaders/TestPropertiesConfigLoader.java
kw2542: Updated.

samza-core/src/main/java/org/apache/samza/config/loaders/PropertiesConfigLoader.java
kw2542: Good question, I am putting String here as we need a relative path on cluster based job coordinator to read configs, URI with file schema does not seem to have a clean support of relative path.

samza-core/src/test/java/org/apache/samza/config/loaders/TestPropertiesConfigLoader.java
bkonold: if we expect various exceptions to ever be thrown from this method, it may be worthwhile to verify the message as well to disambiguate

samza-core/src/main/java/org/apache/samza/config/loaders/PropertiesConfigLoaderFactory.java
```diff
@@ -0,0 +1,41 @@
...
+package org.apache.samza.config.loaders;
+
+import org.apache.samza.SamzaException;
+import org.apache.samza.config.Config;
+import org.apache.samza.config.ConfigLoader;
+import org.apache.samza.config.ConfigLoaderFactory;
+
+
+public class PropertiesConfigLoaderFactory implements ConfigLoaderFactory {
+  private static final String KEY = "path";
```
bkonold: How will customers know this config key is required? Just this runtime check, or is this documented elsewhere?

kw2542: it is documented in this factory class that "ConfigLoaderFactory which initialize {@link PropertiesConfigLoader} with properties file path specified in job.config.loader.properties.path"

bkonold: I'm sorry, where is this documentation located? I don't see it anywhere in this PR

kw2542: I must forget to update the PR before.

Now, you should be able to see it:

/**
 * ConfigLoaderFactory which initialize {@link PropertiesConfigLoader} with properties file path specified in
 * job.config.loader.properties.path.
 */
public class PropertiesConfigLoaderFactory implements ConfigLoaderFactory

samza-core/src/main/java/org/apache/samza/config/loaders/PropertiesConfigLoader.java
cameronlee314: Is it not possible to use absolute path? The existing `PropertiesConfigFactory` uses `URI`, so it can access a config file from anywhere. It seems like your `PropertiesConfigLoader` should be consistent with that.
For example, https://github.com/apache/samza-hello-samza uses absolute path.

samza-core/src/main/java/org/apache/samza/config/loaders/PropertiesConfigLoaderFactory.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.samza.config.ConfigLoader;
+import org.apache.samza.config.ConfigLoaderFactory;
+
+
+/**
+ * ConfigLoaderFactory which initialize {@link PropertiesConfigLoader} with properties file path specified in
+ * job.config.loader.properties.path.
+ */
+public class PropertiesConfigLoaderFactory implements ConfigLoaderFactory {
+  private static final String KEY = "path";
```
cameronlee314: minor: Can you please use a bit more descriptive name for the variable? (e.g. `PATH_KEY`)

samza-core/src/main/java/org/apache/samza/config/loaders/PropertiesConfigLoader.java
kw2542: PropertiesConfigFactory is able to use absolute path because it is running during start up and we can specify the absolute path when launching the job.

PropertiesConfigLoader, on the other hand, is being invoked on AM, where the config file is located under the deployment folder, which means, the absolute path contains the deployment folder name, such as "application_1576531941601_0894", which we don't know before submitting the job. So, during job submission, we only specify the relative path to tell where AM should be looking for the config file.

samza-api/src/main/java/org/apache/samza/config/ConfigLoader.java
```diff
@@ -0,0 +1,27 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.config;
+
+/**
+ * The primary means of fetching full job {@link org.apache.samza.config.Config} on cluster based job coordinator during start up.
```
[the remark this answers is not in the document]
kw2542: Good question, you are right, it will be used in LocalApplicationRunner for standalone jobs. Let me update it.

samza-api/src/main/java/org/apache/samza/config/ConfigLoaderFactory.java
```diff
@@ -0,0 +1,33 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.config;
+
+/**
+ * The factory for an {@link ConfigLoader} instance to loader job config.
```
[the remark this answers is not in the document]
kw2542: Good catch.

samza-api/src/main/java/org/apache/samza/config/ConfigLoaderFactory.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public interface ConfigLoaderFactory {
+  /**
+   * Get an instance of {@link ConfigLoader}
+   *
+   * @param config start up config which specifies properties needed for the loader to load config.
```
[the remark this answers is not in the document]
kw2542: Updated

samza-core/src/main/java/org/apache/samza/config/loaders/PropertiesConfigLoaderFactory.java
kw2542: Updated

samza-core/src/main/java/org/apache/samza/config/loaders/PropertiesConfigLoader.java
cameronlee314: Ah, right, the configs will be loaded on the job coordinator. Thanks for clarifying.
Can you please add some more documentation about how to properly specify the path? It's unclear from the code what "path" means and what the format is, since it is just a String.

kw2542: Added more instruction using Hello Samza as an example.

PR #1238, opened 2019-12-18
SAMZA-2407: Add job.config.loader.* configs in JobConfig & corresponding env variables in ShellCommandConfig

kw2542:
Design:

Changes:

1. Add job.config.loader.factory to specify ConfigLoaderFactory.
2. Add job.config.loader.properties prefix to specify properties needed by ConfigLoader to fetch config.
3. Add ENV_SUBMISSION_CONFIG enviroment variable to pass job submission configs to Yarn.

API Changes:

Upgrade Instructions:
Usage Instructions:

Tests:
N/A.


samza-core/src/main/java/org/apache/samza/config/JobConfig.java
```diff
@@ -47,6 +47,15 @@
   public static final String CONFIG_REWRITERS = "job.config.rewriters";
   public static final String CONFIG_REWRITER_CLASS = "job.config.rewriter.%s.class";
 
+  /**
+   * job.config.loader.factory specifies {@link ConfigLoaderFactory} to get {@link ConfigLoader}
+   */
+  public static final String CONFIG_LOADER_FACTORY = "job.config.loader.factory";
+  /**
+   * job.config.loader.properties prefix wraps properties needed for {@link ConfigLoader} to loader config.
```
cameronlee314: Do you think it would be helpful to put a small example here about how the prefix will be used?

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
```diff
@@ -125,6 +134,16 @@
...
+
+  // Auto-sizing related configs tthat ake precedence over respective sizing confings job.container.count, etc,
+  // *only* when job.autosizing.enabled is true. Otherwise current behavior is maintained.
+  public static final String JOB_AUTOSIZING_ENABLED = "job.autosizing.enabled";
+  public static final String JOB_AUTOSIZING_CONTAINER_COUNT = "job.autosizing.container.count";
+  public static final String JOB_AUTOSIZING_CONTAINER_THREAD_POOL_SIZE = "job.autosizing.container.thread.pool.size";
+  public static final String JOB_AUTOSIZING_CONTAINER_MAX_HEAP_MB = "job.autosizing.container.maxheap.mb";
+  public static final String JOB_AUTOSIZING_CONTAINER_MEMORY_MB = "job.autosizing.container.memory.mb";
+  public static final String JOB_AUTOSIZING_CONTAINER_MAX_CORES = "job.autosizing.container.cpu.cores";
+
```
cameronlee314: I think @rmatharu was working on merging some of this code in separately. You may want to double check with him on that. We should avoid merging the autosizing code with this PR, just to keep the changes isolated.

samza-core/src/main/scala/org/apache/samza/config/ShellCommandConfig.scala
```diff
@@ -25,6 +25,11 @@ object ShellCommandConfig {
    */
   val ENV_COORDINATOR_SYSTEM_CONFIG = "SAMZA_COORDINATOR_SYSTEM_CONFIG"
 
+  /**
+   * This enviorment variable is used to pass configs provided during job submission.
```
cameronlee314: 1. s/enviorment/environment
2. Is this a JSON serialized map like `ENV_COORDINATOR_SYSTEM_CONFIG`? If so, could you please specify that in the javadoc?

bkonold: / whatever the format of the config is, it should be indicated in the javadoc

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
kw2542: Good point, this change is pulled in accidentally.

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
kw2542: Added

samza-core/src/main/scala/org/apache/samza/config/ShellCommandConfig.scala
kw2542: Updated the doc to reflect that this is a JSON serialized map.

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
```diff
@@ -195,20 +205,20 @@ public int getMonitorPartitionChangeFrequency() {
...
+        if (existingPatternForSystem == null) {
+          newPatternForSystem = Pattern.compile(rewriterRegex.get());
+        } else {
+          newPatternForSystem =
+              Pattern.compile(String.join("|", existingPatternForSystem.pattern(), rewriterRegex.get()));
         }
-      });
+        inputRegexesToMonitor.put(rewriterSystem.get(), newPatternForSystem);
+      }
+    });
```
cameronlee314: Is this an indentation change only? If so, can you please undo it so your diff is just your changes?
Your name would go into the commit history for changing this piece of code, so it would be good to avoid that given that your change is unrelated to cleaning up indentation.

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
```diff
@@ -348,4 +358,11 @@ public String getCoordinatorStreamFactory() {
     return get(COORDINATOR_STREAM_FACTORY, DEFAULT_COORDINATOR_STREAM_CONFIG_FACTORY);
   }
 
+  /**
+   * Get config loader factory according to the configs
+   * @return full qualified name of {@link ConfigLoaderFactory}
+   */
+  public String getConfigLoaderFactory() {
+    return get(CONFIG_LOADER_FACTORY);
```
cameronlee314: Is this an optional config, or is it required? If it is optional, can you please have it return `Optional<String>`, or add a default value? If required, can you please throw a `ConfigException`?

samza-core/src/main/scala/org/apache/samza/config/ShellCommandConfig.scala
```diff
@@ -25,6 +25,11 @@ object ShellCommandConfig {
    */
   val ENV_COORDINATOR_SYSTEM_CONFIG = "SAMZA_COORDINATOR_SYSTEM_CONFIG"
 
+  /**
+   * This enviorment variable is used to pass a JSON serialized map of configs provided during job submission.
```
cameronlee314: "environment" is spelled incorrectly

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
kw2542: Reverted. It was automatically corrected by IDEA.

samza-core/src/main/scala/org/apache/samza/config/ShellCommandConfig.scala
kw2542: Fixed. Nice catch.

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
kw2542: It will eventually be required but optional before we fully deprecate the old path where we read config from coordinator stream. 

I will use Optional<String> for now, and will change to required with ConfigException when we clean up the old flow.

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
```diff
@@ -47,6 +47,17 @@
...
+  /**
+   * job.config.loader.factory specifies {@link ConfigLoaderFactory} to get {@link ConfigLoader}
+   */
+  public static final String CONFIG_LOADER_FACTORY = "job.config.loader.factory";
+  /**
+   * job.config.loader.properties prefix wraps properties needed for {@link ConfigLoader} to loader config.
+   * e.g. {@link org.apache.samza.config.loaders.PropertiesConfigLoaderFactory}
+   * will read job.config.loader.properties.path to locate the config file location
+   */
+  public static final String CONFIG_LOADER_PROPERTIES_PREFIX = "job.config.loader.properties.";
```
sborya: where is it used? if it used only in the Loader, should't it be defined there too?

samza-core/src/main/scala/org/apache/samza/config/ShellCommandConfig.scala
```diff
@@ -25,6 +25,11 @@ object ShellCommandConfig {
    */
   val ENV_COORDINATOR_SYSTEM_CONFIG = "SAMZA_COORDINATOR_SYSTEM_CONFIG"
 
+  /**
+   * This environment variable is used to pass a JSON serialized map of configs provided during job submission.
+   */
+  val ENV_SUBMISSION_CONFIG = "SAMZA_SUBMISSION_CONFIG"
```
sborya: where is it used?

kw2542: This will be used in YarnJob when submitting the job to Yarn. This environment variable will encapsulate config overrides users would like to override on AM.

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
kw2542: Yes, it is supposed to be only used by config loader, which specifies properties it needed to retrieve full job config. 

Are you suggesting we move this to ConfigLoader instead?

sborya: yea, because if it is not used here - there is no point keeping it here.

samza-core/src/main/scala/org/apache/samza/config/ShellCommandConfig.scala
sborya: So code using it will be added later? then it is fine. In general it is easier to read PR if all relevant code is in. But this is fine too.

kw2542: Good point, yes, it will be added later. I am splitting them to make PRs relatively small.

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
kw2542: Moved to ConfigLoaderFactory

PR #1239, opened 2019-12-19
SAMZA-2421: Add SystemProducer for Azure Blob Storage

lakshmi-manasa-g:
**Feature:** [SEP-26: Azure Blob System Producer](https://cwiki.apache.org/confluence/display/SAMZA/SEP-26%3A+Azure+Blob+Storage+Producer)

**Changes:** New "azureblob" package added that contains the new AzureBlobSystemProducer and its components. This new AzureBlobSystemProducer implements the existing SystemProducer interface. There are no changes made to existing API and existing implementations in any manner.

**Tests:** Unit tests added for all components. A sample job in hello-samza was successfully configured to use this producer and create blobs on Azure Blob Storage.

**Upgrade instructions:** Backwards compatible change as its a completely new component and hence no upgrade needed.

**Usage instructions:** Configure a Samza job to use this producer by setting the following configs
`systems.<system-name/azure-container-name>.samza.factory=org.apache.samza.system.azureblob.AzureBlobSystemFactory
systems.<system-name/azure-container-name>azureblob.account.name=<azure-account-name>
systems.<system-name/azure-container-name>.azureblob.account.key=<azure-account-key>`

Samza job might also need to pull in Jackson dependencies (com.fasterxml.jackson.core:jackson-core/jackson-databind/jackson-annotations:2.10.0)

### DIFF
--- a/samza-azure/src/main/java/org/apache/samza/system/azureblob/AzureBlobConfig.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/azureblob/AzureBlobConfig.java
@@ -1,0 +1,189 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.samza.system.azureblob;
+
+import org.apache.samza.system.azureblob.compression.CompressionType;
+import java.time.Duration;
+import org.apache.samza.config.Config;
+import org.apache.samza.config.ConfigException;
+import org.apache.samza.config.MapConfig;
+
+public class AzureBlobConfig extends MapConfig {
+  private static final String SYSTEM_AZUREBLOB_PREFIX = "systems.%s.azureblob.";
+  //Server Instance Level Property
+
+  // The duration after which an Azure request will be logged as a warning.
+  public static final String AZURE_BLOB_LOG_SLOW_REQUESTS_MS = "samza.azureblob.log.slowRequestMs";
+  private static final long AZURE_BLOB_LOG_SLOW_REQUESTS_MS_DEFAULT = Duration.ofSeconds(30).toMillis();
+
+  // system Level Properties.
+  // fully qualified class name of the AzureBlobWriter impl for the producer system
+  public static final String SYSTEM_WRITER_FACTORY_CLASS_NAME = "systems.%s.producer.azureblob.writer.factory.class";
+  public static final String SYSTEM_WRITER_FACTORY_CLASS_NAME_DEFAULT = "org.apache.samza.system.azureblob.avro.AzureBlobAvroWriterFactory";
+
+  // Azure Storage Account name under which the Azure container representing this system is.
+  // System name = Azure container name (https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names)
+  public static final String SYSTEM_AZURE_ACCOUNT_NAME = SYSTEM_AZUREBLOB_PREFIX + "account.name";
+
+  // Azure Storage Account key associated with the Azure Storage Account
+  public static final String SYSTEM_AZURE_ACCOUNT_KEY  = SYSTEM_AZUREBLOB_PREFIX + "account.key";
+
+  // Whether to use proxy while connecting to Azure Storage
+  public static final String SYSTEM_AZURE_USE_PROXY  = SYSTEM_AZUREBLOB_PREFIX + "proxy.use";
+  public static final boolean SYSTEM_AZURE_USE_PROXY_DEFAULT = false;
+
+  // name of the host to be used as proxy
+  public static final String SYSTEM_AZURE_PROXY_HOSTNAME = SYSTEM_AZUREBLOB_PREFIX + "proxy.hostname";
+
+  // port in the proxy host to be used
+  public static final String SYSTEM_AZURE_PROXY_PORT = SYSTEM_AZUREBLOB_PREFIX + "proxy.port";
+
+  // type of compression to be used before uploading blocks : “none” or “gzip”
+  public static final String SYSTEM_COMPRESSION_TYPE = SYSTEM_AZUREBLOB_PREFIX + "compression.type";
+  private static final CompressionType SYSTEM_COMPRESSION_TYPE_DEFAULT = CompressionType.NONE;
+
+  // maximum size of uncompressed block in bytes
+  public static final String SYSTEM_MAX_FLUSH_THRESHOLD_SIZE = SYSTEM_AZUREBLOB_PREFIX + "maxFlushThresholdSize";
+  private static final int SYSTEM_MAX_FLUSH_THRESHOLD_SIZE_DEFAULT = 10485760;
+
+  // maximum size of uncompressed blob in bytes
+  public static final String SYSTEM_MAX_BLOB_SIZE = SYSTEM_AZUREBLOB_PREFIX + "maxBlobSize";
+  private static final long SYSTEM_MAX_BLOB_SIZE_DEFAULT = Long.MAX_VALUE; // unlimited
+
+  // maximum number of messages in a blob
+  public static final String SYSTEM_MAX_MESSAGES_PER_BLOB = SYSTEM_AZUREBLOB_PREFIX + "maxMessagesPerBlob";
+  private static final long SYSTEM_MAX_MESSAGES_PER_BLOB_DEFAULT = Long.MAX_VALUE; // unlimited
+
+  // number of threads to asynchronously upload blocks
+  public static final String SYSTEM_THREAD_POOL_COUNT = SYSTEM_AZUREBLOB_PREFIX + "threadPoolCount";
+  private static final int SYSTEM_THREAD_POOL_COUNT_DEFAULT = 2;
+
+  // size of the queue to hold blocks ready to be uploaded by asynchronous threads.
+  // If all threads are busy uploading then blocks are queued and if queue is full then main thread will start uploading
+  // which will block processing of incoming messages
+  // Default - Thread Pool Count * 2
+  public static final String SYSTEM_BLOCKING_QUEUE_SIZE = SYSTEM_AZUREBLOB_PREFIX + "blockingQueueSize";
+
+  // timeout to finish uploading all blocks before committing a blob
+  public static final String SYSTEM_FLUSH_TIMEOUT_MS = SYSTEM_AZUREBLOB_PREFIX + "flushTimeoutMs";
+  private static final long SYSTEM_FLUSH_TIMEOUT_MS_DEFAULT = Duration.ofMinutes(3).toMillis();
+
+  // timeout to finish committing all the blobs currently being written to. This does not include the flush timeout per blob
+  public static final String SYSTEM_CLOSE_TIMEOUT_MS = SYSTEM_AZUREBLOB_PREFIX + "closeTimeoutMs";
+  private static final long SYSTEM_CLOSE_TIMEOUT_MS_DEFAULT = Duration.ofMinutes(5).toMillis();
+
+  // if true, a random string of 8 chars is suffixed to the blob name to prevent name collision
+  // when more than one Samza tasks are writing to the same SSP.
+  // It is advisable to set this to true
+  public static final String SYSTEM_SUFFIX_RANDOM_STRING_TO_BLOB_NAME = SYSTEM_AZUREBLOB_PREFIX + "suffixRandomStringToBlobName";
+  private static final boolean SYSTEM_SUFFIX_RANDOM_STRING_TO_BLOB_NAME_DEFAULT = false;
+
+  public AzureBlobConfig(Config config) {
+    super(config);
+  }
+
+  public String getAzureAccountKey(String systemName) {
+    String accountKey = get(String.format(SYSTEM_AZURE_ACCOUNT_KEY, systemName));
+    if (accountKey == null) {
+      throw new ConfigException("Azure account key is required.");
+    }
+    return accountKey;
+  }
+
+  public String getAzureAccountName(String systemName) {
+    String accountName = get(String.format(SYSTEM_AZURE_ACCOUNT_NAME, systemName));
+    if (accountName == null) {
+      throw new ConfigException("Azure account name is required.");
+    }
+    return accountName;
+  }
+
+  public boolean getUseProxy(String systemName) {
+    return getBoolean(String.format(SYSTEM_AZURE_USE_PROXY, systemName), SYSTEM_AZURE_USE_PROXY_DEFAULT);
+  }
+
+  public String getAzureProxyHostname(String systemName) {
+    String hostname = get(String.format(SYSTEM_AZURE_PROXY_HOSTNAME, systemName));
+    if (hostname == null) {
+      throw new ConfigException("Azure proxy host name is required.");
+    }
+    return hostname;
+  }
+
+  public int getAzureProxyPort(String systemName) {
+    return getInt(String.format(SYSTEM_AZURE_PROXY_PORT, systemName));
+  }
+
+  public CompressionType getCompressionType(String systemName) {
+    return CompressionType.valueOf(get(String.format(SYSTEM_COMPRESSION_TYPE, systemName), SYSTEM_COMPRESSION_TYPE_DEFAULT.name()).toUpperCase());
+  }
+
+  public String getAzureBlobWriterFactoryClassName(String systemName) {
+    return get(String.format(SYSTEM_WRITER_FACTORY_CLASS_NAME, systemName), SYSTEM_WRITER_FACTORY_CLASS_NAME_DEFAULT);
+  }
+
+  public int getMaxFlushThresholdSize(String systemName) {
+    return getInt(String.format(SYSTEM_MAX_FLUSH_THRESHOLD_SIZE, systemName), SYSTEM_MAX_FLUSH_THRESHOLD_SIZE_DEFAULT);
+  }
+
+  public int getAzureBlobThreadPoolCount(String systemName) {
+    return getInt(String.format(SYSTEM_THREAD_POOL_COUNT, systemName), SYSTEM_THREAD_POOL_COUNT_DEFAULT);
+  }
+
+  public int getBlockingQueueSizeOrDefault(String systemName, int defaultQueueSize) {
+    return getInt(String.format(SYSTEM_BLOCKING_QUEUE_SIZE, systemName), defaultQueueSize);
+  }
+
+  public long getFlushTimeoutMs(String systemName) {
+    long timeout = getLong(String.format(SYSTEM_FLUSH_TIMEOUT_MS, systemName), SYSTEM_FLUSH_TIMEOUT_MS_DEFAULT);
+    if (timeout <= 0) {
+      throw new ConfigException("Azure Blob flush timeout can not be <= 0");
+    }
+    return timeout;
+  }
+
+  public long getCloseTimeoutMs(String systemName) {
+    long timeout = getLong(String.format(SYSTEM_CLOSE_TIMEOUT_MS, systemName), SYSTEM_CLOSE_TIMEOUT_MS_DEFAULT);
+    if (timeout <= 0) {
+      throw new ConfigException("Azure Blob close timeout can not be <= 0");
+    }
+    return timeout;
+  }
+
+  public long getLogSlowRequestsMs() {
+    long duration = getLong(AZURE_BLOB_LOG_SLOW_REQUESTS_MS, AZURE_BLOB_LOG_SLOW_REQUESTS_MS_DEFAULT);
+    if (duration <= 0) {
+      throw new ConfigException("Azure blob duration to log slow requests can not be <=0.");
+    }
+    return duration;
+  }
+
+  public boolean getSuffixRandomStringToBlobName(String systemName) {
+    return getBoolean(String.format(SYSTEM_SUFFIX_RANDOM_STRING_TO_BLOB_NAME, systemName), SYSTEM_SUFFIX_RANDOM_STRING_TO_BLOB_NAME_DEFAULT);
+  }
+
+  public long getMaxBlobSize(String systemName) {
+    return getLong(String.format(SYSTEM_MAX_BLOB_SIZE, systemName), SYSTEM_MAX_BLOB_SIZE_DEFAULT);
+  }
+
+  public long getMaxMessagesPerBlob(String systemName) {
+    return getLong(String.format(SYSTEM_MAX_MESSAGES_PER_BLOB, systemName), SYSTEM_MAX_MESSAGES_PER_BLOB_DEFAULT);
+  }
+}
--- a/samza-azure/src/main/java/org/apache/samza/system/azureblob/AzureBlobConfig.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/azureblob/AzureBlobConfig.java
@@ -35,7 +35,7 @@ public class AzureBlobConfig extends MapConfig {
 
   // system Level Properties.
   // fully qualified class name of the AzureBlobWriter impl for the producer system
-  public static final String SYSTEM_WRITER_FACTORY_CLASS_NAME = "systems.%s.producer.azureblob.writer.factory.class";
+  public static final String SYSTEM_WRITER_FACTORY_CLASS_NAME = SYSTEM_AZUREBLOB_PREFIX + "writer.factory.class";
   public static final String SYSTEM_WRITER_FACTORY_CLASS_NAME_DEFAULT = "org.apache.samza.system.azureblob.avro.AzureBlobAvroWriterFactory";
 
   // Azure Storage Account name under which the Azure container representing this system is.
@@ -73,7 +73,7 @@ public class AzureBlobConfig extends MapConfig {
 
   // number of threads to asynchronously upload blocks
   public static final String SYSTEM_THREAD_POOL_COUNT = SYSTEM_AZUREBLOB_PREFIX + "threadPoolCount";
-  private static final int SYSTEM_THREAD_POOL_COUNT_DEFAULT = 2;
+  private static final int SYSTEM_THREAD_POOL_COUNT_DEFAULT = 1;
 
   // size of the queue to hold blocks ready to be uploaded by asynchronous threads.
   // If all threads are busy uploading then blocks are queued and if queue is full then main thread will start uploading
@@ -91,9 +91,8 @@ public class AzureBlobConfig extends MapConfig {
 
   // if true, a random string of 8 chars is suffixed to the blob name to prevent name collision
   // when more than one Samza tasks are writing to the same SSP.
-  // It is advisable to set this to true
   public static final String SYSTEM_SUFFIX_RANDOM_STRING_TO_BLOB_NAME = SYSTEM_AZUREBLOB_PREFIX + "suffixRandomStringToBlobName";
-  private static final boolean SYSTEM_SUFFIX_RANDOM_STRING_TO_BLOB_NAME_DEFAULT = false;
+  private static final boolean SYSTEM_SUFFIX_RANDOM_STRING_TO_BLOB_NAME_DEFAULT = true;
 
   public AzureBlobConfig(Config config) {
     super(config);
@@ -147,8 +146,8 @@ public class AzureBlobConfig extends MapConfig {
     return getInt(String.format(SYSTEM_THREAD_POOL_COUNT, systemName), SYSTEM_THREAD_POOL_COUNT_DEFAULT);
   }
 
-  public int getBlockingQueueSizeOrDefault(String systemName, int defaultQueueSize) {
-    return getInt(String.format(SYSTEM_BLOCKING_QUEUE_SIZE, systemName), defaultQueueSize);
+  public int getBlockingQueueSize(String systemName) {
+    return getInt(String.format(SYSTEM_BLOCKING_QUEUE_SIZE, systemName), 2 * getAzureBlobThreadPoolCount(systemName));
   }
 
   public long getFlushTimeoutMs(String systemName) {
--- a/samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobAvroWriter.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobAvroWriter.java
@@ -157,9 +157,11 @@ public class AzureBlobAvroWriter implements AzureBlobWriter {
     if (ome.getMessage() instanceof IndexedRecord) {
       optionalIndexedRecord = Optional.of((IndexedRecord) ome.getMessage());
       encodedRecord = encodeRecord((IndexedRecord) ome.getMessage());
-    } else {
+    } else if (ome.getMessage() instanceof byte[]) {
       optionalIndexedRecord = Optional.empty();
       encodedRecord = (byte[]) ome.getMessage();
+    } else {
+      throw new IllegalArgumentException("AzureBlobAvroWriter only supports IndexedRecord and byte[].");
     }
 
     synchronized (currentDataFileWriterLock) {
@@ -270,11 +272,12 @@ public class AzureBlobAvroWriter implements AzureBlobWriter {
       LOG.info("Closing the blob: {}", blockBlobAsyncClient.getBlobUrl().toString());
       // dataFileWriter.close calls close of the azureBlobOutputStream associated with it.
       dataFileWriter.close();
-    } finally {
+    } catch (Exception e) {
       // ensure that close is called even if dataFileWriter.close fails.
       // This is to avoid loss of all the blocks uploaded for the blob
       // as commitBlockList happens in close of azureBlobOutputStream.
       azureBlobOutputStream.close();
+      throw e;
     }
   }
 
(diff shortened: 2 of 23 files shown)


samza-azure/src/main/java/org/apache/samza/system/azureblob/AzureBlobConfig.java
```diff
@@ -0,0 +1,189 @@
...
+  private static final long SYSTEM_MAX_MESSAGES_PER_BLOB_DEFAULT = Long.MAX_VALUE; // unlimited
+
+  // number of threads to asynchronously upload blocks
+  public static final String SYSTEM_THREAD_POOL_COUNT = SYSTEM_AZUREBLOB_PREFIX + "threadPoolCount";
+  private static final int SYSTEM_THREAD_POOL_COUNT_DEFAULT = 2;
+
+  // size of the queue to hold blocks ready to be uploaded by asynchronous threads.
+  // If all threads are busy uploading then blocks are queued and if queue is full then main thread will start uploading
+  // which will block processing of incoming messages
+  // Default - Thread Pool Count * 2
```
cameronlee314: This comment seems out of scope for this class, since it doesn't actually set a default.

samza-azure/src/main/java/org/apache/samza/system/azureblob/producer/AzureBlobSystemProducer.java
```diff
@@ -0,0 +1,505 @@
...
+
+  private static final int PREMIUM_MAX_BLOCK_SIZE = 100 * 1024 * 1024; // 100MB
+  private static final int STANDARD_MAX_BLOCK_SIZE = 4 * 1024 * 1024; // 4MB
+
+  private BlobContainerAsyncClient containerAsyncClient;
+  private final String systemName;
+  private final AzureBlobConfig config;
+
+  // Map of writers indexed first by sourceName and then by (streamName, partitionName) or just streamName if partition key does not exist.
+  private Map<String, Map<String, AzureBlobWriter>> writerMap;
```
cameronlee314: Can this be final?

samza-azure/src/main/java/org/apache/samza/system/azureblob/producer/AzureBlobSystemProducer.java
```diff
@@ -0,0 +1,505 @@
...
+      throw new SystemProducerException("Could not create writer factory with name " + writerFactoryClassName, e);
+    }
+    this.flushTimeoutMs = this.config.getFlushTimeoutMs(this.systemName);
+    this.closeTimeout = this.config.getCloseTimeoutMs(this.systemName);
+    this.blockFlushThresholdSize = this.config.getMaxFlushThresholdSize(this.systemName);
+    int asyncBlobThreadPoolCount = this.config.getAzureBlobThreadPoolCount(this.systemName);
+    int blockingQueueSize = this.config.getBlockingQueueSizeOrDefault(this.systemName, asyncBlobThreadPoolCount * 2);
+
+    LOG.info("SystemName: {} block flush size:{}", systemName, this.blockFlushThresholdSize);
+    LOG.info("SystemName: {} thread count:{}", systemName, asyncBlobThreadPoolCount);
```
cameronlee314: Are these needed?

samza-azure/src/main/java/org/apache/samza/system/azureblob/producer/AzureBlobSystemProducer.java
```diff
@@ -0,0 +1,505 @@
...
+   */
+  @Override
+  public synchronized void stop() {
+    if (!isStarted) {
+      throw new SystemProducerException("Attempting to stop a producer that was not started.");
+    }
+
+    if (isStopped) {
+      throw new SystemProducerException("Attempting to stop an already stopped producer.");
+    }
```
cameronlee314: Should this throw an exception in these cases? What do other `SystemProducer` impls do?

samza-azure/src/main/java/org/apache/samza/system/azureblob/producer/AzureBlobSystemProducer.java
```diff
@@ -0,0 +1,505 @@
...
+      SkuName accountType = storageClient.getAccountInfo().block().getSkuName();
+      long flushThresholdSize = config.getMaxFlushThresholdSize(systemName);
+      boolean isPremiumAccount = SkuName.PREMIUM_LRS == accountType;
+      if (isPremiumAccount && flushThresholdSize > PREMIUM_MAX_BLOCK_SIZE) { // 100 MB
+        throw new SystemProducerException("Azure storage account with name: " + accountName
+            + " is a premium account and can only handle upto 100MB threshold size. Given flush threshold size is "
+            + flushThresholdSize);
+      } else if (!isPremiumAccount && flushThresholdSize > STANDARD_MAX_BLOCK_SIZE) { // STANDARD account
+        throw new SystemProducerException("Azure storage account with name: " + accountName
+            + " is a standard account and can only handle upto 4MB threshold size. Given flush threshold size is "
```
cameronlee314: Should the `100MB` and `4MB` be changed to use `PREMIUM_MAX_BLOCK_SIZE` and `STANDARD_MAX_BLOCK_SIZE`? Then, if someone changes the max block size constants, the log messages will be updated automatically.

samza-azure/src/main/java/org/apache/samza/system/azureblob/producer/AzureBlobSystemProducer.java
```diff
@@ -0,0 +1,505 @@
...
+      writerMapKey = messageEnvelope.getSystemStream().getStream() + "/" + partitionKey;
+      blobURLPrefix = String.format(BLOB_NAME_PARTITION_PREFIX, messageEnvelope.getSystemStream().getStream(), partitionKey);
+    }
+    Map<String, AzureBlobWriter> sourceWriterMap = writerMap.get(source);
+    if (sourceWriterMap == null) {
+      throw new SystemProducerException("Attempting to send to source: " + source + " but it is not registered");
+    }
+    AzureBlobWriter writer = sourceWriterMap.get(writerMapKey);
+    if (writer == null) {
+      synchronized (sourceWriterCreationLockMap.get(source)) {
```
cameronlee314: Should these two lines be switched? Otherwise, two different threads could still create a new writer.

samza-azure/src/main/java/org/apache/samza/system/azureblob/producer/AzureBlobSystemProducer.java
```diff
@@ -0,0 +1,505 @@
...
+      pendingClose.clear();
+    }
+  }
+
+  @VisibleForTesting
+  AzureBlobWriter createNewWriter(String blobURL, AzureBlobWriterMetrics writerMetrics) {
+    try {
+      return writerFactory.getWriterInstance(containerAsyncClient, blobURL, asyncBlobThreadPool, writerMetrics,
+          blockFlushThresholdSize, flushTimeoutMs,
+          new CompressionFactory().getCompression(config.getCompressionType(systemName)),
```
cameronlee314: Can you make `CompressionFactory` into a singleton?

samza-azure/src/test/java/org/apache/samza/system/azureblob/avro/TestAzureBlobAvroWriter.java
```diff
@@ -0,0 +1,452 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+import static org.mockito.Matchers.*;
+import static org.mockito.Mockito.*;
```
cameronlee314: Please avoid using wildcard imports.

samza-azure/src/test/java/org/apache/samza/system/azureblob/avro/TestAzureBlobAvroWriter.java
```diff
@@ -0,0 +1,452 @@
...
+    BlobContainerAsyncClient mockContainerClient = PowerMockito.mock(BlobContainerAsyncClient.class);
+    azureBlobAvroWriter = spy(new AzureBlobAvroWriter(mockContainerClient,
+        mock(AzureBlobWriterMetrics.class), threadPool, THRESHOLD, 60000, blobUrlPrefix,
+        mockDataFileWriter, mockAzureBlobOutputStream, mockBlockBlobAsyncClient, maxBlobSize, maxRecordsPerBlob));
+
+    DataFileWriter<IndexedRecord> mockDataFileWriter2 = mock(DataFileWriter.class);
+    AzureBlobOutputStream mockAzureBlobOutputStream2 = mock(AzureBlobOutputStream.class);
+
+    when(mockAzureBlobOutputStream.getSize()).thenReturn((long) 1);
+    when(mockAzureBlobOutputStream.getSize()).thenReturn((long) 1);
```
cameronlee314: These two lines are the same. Is that intentional?

samza-azure/src/test/java/org/apache/samza/system/azureblob/avro/TestAzureBlobAvroWriter.java
```diff
@@ -0,0 +1,452 @@
...
+    String blobUrlPrefix = "test";
+    long maxBlobSize = 1000;
+    BlobContainerAsyncClient mockContainerClient = PowerMockito.mock(BlobContainerAsyncClient.class);
+    azureBlobAvroWriter = spy(new AzureBlobAvroWriter(mockContainerClient,
+        mock(AzureBlobWriterMetrics.class), threadPool, THRESHOLD, 60000, blobUrlPrefix,
+        null, null, null, maxBlobSize, 10));
+
+    when(mockAzureBlobOutputStream.getSize()).thenReturn((long) maxBlobSize - 1);
+    PowerMockito.whenNew(AzureBlobOutputStream.class).withAnyArguments().thenReturn(mockAzureBlobOutputStream);
+    PowerMockito.whenNew(DataFileWriter.class).withAnyArguments().thenReturn(mockDataFileWriter);
```
cameronlee314: It looks like this currently just tests that multiple blob clients are created.
Would it be useful/feasible to test that the data gets appended to the correct places?

samza-azure/src/test/java/org/apache/samza/system/azureblob/avro/TestAzureBlobAvroWriter.java
```diff
@@ -0,0 +1,452 @@
...
+    long maxBlobSize = AzureBlobAvroWriter.DATAFILEWRITER_OVERHEAD + 1000;
+    long maxRecordsPerBlob = 10;
+    BlobContainerAsyncClient mockContainerClient = PowerMockito.mock(BlobContainerAsyncClient.class);
+    azureBlobAvroWriter = spy(new AzureBlobAvroWriter(mockContainerClient,
+        mock(AzureBlobWriterMetrics.class), threadPool, THRESHOLD, 60000, blobUrlPrefix,
+        null, null, null, maxBlobSize, maxRecordsPerBlob));
+
+    when(mockAzureBlobOutputStream.getSize()).thenReturn((long) 1);
+    PowerMockito.whenNew(AzureBlobOutputStream.class).withAnyArguments().thenReturn(mockAzureBlobOutputStream);
+    PowerMockito.whenNew(DataFileWriter.class).withAnyArguments().thenReturn(mockDataFileWriter);
```
cameronlee314: Same as above about verifying where the data gets appended.

samza-azure/src/test/java/org/apache/samza/system/azureblob/avro/TestAzureBlobOutputStream.java
```diff
@@ -0,0 +1,347 @@
...
+        Assert.assertEquals(ByteBuffer.wrap(compressB), flux.blockFirst());
+      });
+    verify(mockMetrics).updateWriteByteMetrics(2 * THRESHOLD);
+    verify(mockMetrics, times(2)).updateAzureUploadMetrics();
+  }
+
+  @Test
+  public void testWriteLargeRecordWithSmallRecordInBuffer() {
+    byte[] halfBlock = new byte[THRESHOLD / 2];
+    byte[] fullBlock = new byte[THRESHOLD];
```
cameronlee314: Would it be good to fill in these blocks with some data to be able to verify that the expected data gets written? The test currently just checks the size of the blocks through the byte array comparisons, but doesn't verify that the expected data is written as expected.

samza-azure/src/test/java/org/apache/samza/system/azureblob/avro/TestAzureBlobOutputStream.java
```diff
@@ -0,0 +1,347 @@
...
+    int blockNum1 = 1;
+    String blockId1 = String.format("%05d", blockNum1);
+    String blockIdEncoded1 = Base64.getEncoder().encodeToString(blockId1.getBytes());
+
+    PowerMockito.doAnswer(invocation -> {
+        ArrayList<String> blockListArg = (ArrayList<String>) invocation.getArguments()[0];
+        String blockIdArg = (String) blockListArg.toArray()[0];
+        String blockIdArg1 = (String) blockListArg.toArray()[1];
+        Assert.assertEquals(blockIdEncoded, blockIdArg);
+        Assert.assertEquals(blockIdEncoded1, blockIdArg1);
```
cameronlee314: Can this be replaced with `Assert.assertEquals(Arrays.asList(blockIdEncoded, blockIdEncoded1), blockListArg)`?

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobAvroWriter.java
```diff
@@ -0,0 +1,358 @@
...
+
+  // currentBlobWriterComponents == null only for the first blob immediately after this AzureBlobAvroWriter object has been created.
+  // rest of this object's lifecycle, currentBlobWriterComponents is not null.
+  private BlobWriterComponents currentBlobWriterComponents = null;
+  private final List<BlobWriterComponents> allBlobWriterComponents = new ArrayList<>();
+  private Schema schema = null;
+  // datumWriter == null only for the first blob immediately after this AzureBlobAvroWriter object has been created.
+  // It is created from the schema taken from the first OutgoingMessageEnvelope. Hence the first OME has to be a decoded avro record.
+  // For rest of this object's lifecycle, datumWriter is not null.
+  private DatumWriter<IndexedRecord> datumWriter = null;
```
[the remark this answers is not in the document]
lakshmi-manasa-g: Reason these were not marked volatile is  because these are not accessed by multiple threads at a time. The currentDataFileWriterLock ensures only one thread can access these at a time.

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobOutputStream.java
```diff
@@ -0,0 +1,314 @@
...
+public class AzureBlobOutputStream extends OutputStream {
+
+  private static final Logger LOG = LoggerFactory.getLogger(AzureBlobOutputStream.class);
+  private static final int MAX_ATTEMPT = 3;
+  private static final int MAX_BLOCKS_IN_AZURE_BLOB = 50000;
+  public static final String BLOB_RAW_SIZE_BYTES_METADATA = "rawSizeBytes";
+  private final long flushTimeoutMs;
+  private final BlockBlobAsyncClient blobAsyncClient;
+  private final Executor blobThreadPool;
+  private Optional<ByteArrayOutputStream> byteArrayOutputStream;
```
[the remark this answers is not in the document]
lakshmi-manasa-g: All the methods are synchronized => only one thread can access these variables at a time. Hence did not mark them volatile. Is there another reason for asking them to be marked volatile?

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobOutputStream.java
```diff
@@ -0,0 +1,314 @@
...
+  // even if stageBlock is a blocking call.
+  private final ArrayList<String> blockList;
+  private final Set<CompletableFuture<Void>> pendingUpload = ConcurrentHashMap.newKeySet();
+  private final int maxBlockFlushThresholdSize;
+  private final AzureBlobWriterMetrics metrics;
+  private final Compression compression;
+
+  private volatile boolean isClosed = false;
+  private long totalUploadedBlockSize = 0;
+  private int blockNum;
```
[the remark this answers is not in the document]
lakshmi-manasa-g: All the methods are synchronized => only one thread can access these variables at a time. Hence did not mark them volatile. Is there another reason for asking them to be marked volatile?

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobOutputStream.java
```diff
@@ -0,0 +1,314 @@
...
+   * @throws RuntimeException when
+   *       - byteArrayOutputStream.close fails or
+   *       - any of the pending uploads fails or
+   *       - blob's commitBlockList fails
+   */
+  @Override
+  public synchronized void close() {
+
+    if (isClosed) {
+      return;
```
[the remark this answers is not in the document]
lakshmi-manasa-g: This method is going to be called twice in the event of no failures - once by DataFileWriter.close and in the finally block of AzureBlobAvroWriter.close - This was done to avoid loss of blocks if DataFilwWriter.close failed. Hence, logging something would unnecessarily confuse the users

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobAvroWriter.java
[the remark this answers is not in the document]
cameronlee314: Ah yes, if you are protecting all access (read and write) with locks (which implement `Lock`), then looks like you are good.
https://stackoverflow.com/questions/1570589/is-the-volatile-keyword-required-for-fields-accessed-via-a-reentrantlock

samza-azure/src/main/java/org/apache/samza/system/azureblob/producer/AzureBlobSystemProducer.java
lakshmi-manasa-g: the writer creation is wrapped inside another check if writer == null at line 410 and hence 2 thread will not create the writer.
On the other hand if we did switch the statements, then every thread every time will be forced to acquire the lock even if the writer is not null

here is the problematic scenario and why it wont lead to two writer creation 
at t = 0, two threads T1 and T2 both hit line 407 and both see writer = null
at t = 1, T1 and T2 both hit line 408 and try to get lock
at t = 2, T1 gets the lock and T2 enters wait
at t = 3, T1 hits line 409 and gets the writer from the map again and it is still null
at t =4 , T1 hits line 410 and checks if writer is null
at t = 5, T1 creates writer in line 411,412 and 413 
at t = 6, T1 adds the writer to the map
at t= 7, T1 exits critical section and gives up lock and  T2 gets it. 
at t = 8, T2 hits 409 and gets writer from map
at t = 9, T2 checks if writer is null and it is not
at t = 10, T2 exits the critical section and returns the writer

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobOutputStream.java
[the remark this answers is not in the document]
cameronlee314: Hm, generally it is not the case that a `close` is expected to be called twice (and then ignored the second time). Is it difficult to only have it called once in the happy path case?

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobOutputStream.java
```diff
@@ -0,0 +1,314 @@
...
+      Map<String, String> blobMetadata = Collections.singletonMap(BLOB_RAW_SIZE_BYTES_METADATA, Long.toString(totalUploadedBlockSize));
+      blobAsyncClient.commitBlockListWithResponse(blockList, null, blobMetadata, null, null).block();
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      String msg = String.format("Close blob %s failed with exception. Total pending sends %d",
+          blobAsyncClient.getBlobUrl().toString(), pendingUpload.size());
+      throw new RuntimeException(msg, e);
+    } catch (Exception e) {
+      String msg = String.format("Close blob %s failed with exception. Resetting the system producer. %s",
+          blobAsyncClient.getBlobUrl().toString(), e.getLocalizedMessage());
+      throw new RuntimeException(msg, e);
```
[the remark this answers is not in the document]
cameronlee314: Specific exception types can be caught within code. It can make logs slightly more readable or searchable to have this more specific type as well.
Do you have a concern with using a more specific exception type?

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobOutputStream.java
```diff
@@ -0,0 +1,314 @@
...
+      return;
+    }
+    long size = byteArrayOutputStream.get().size();
+    if (size == 0) {
+      return;
+    }
+    LOG.info("Blob: {} uploadBlock. Size:{}", blobAsyncClient.getBlobUrl().toString(), size);
+
+    String blockId = String.format("%05d", blockNum);
+    String blockIdEncoded = Base64.getEncoder().encodeToString(blockId.getBytes());
```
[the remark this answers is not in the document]
cameronlee314: Can you please leave a comment about this in the code? It is not obvious that this is a requirement from the azure SDK.

samza-azure/src/test/java/org/apache/samza/system/azureblob/avro/TestAzureBlobOutputStream.java
lakshmi-manasa-g: `compressB = randomString.substring(0, THRESHOLD / 2).getBytes();
..
Assert.assertEquals(ByteBuffer.wrap(compressB), flux.blockFirst());`

data in compressB is compared to the argument passed to stageBlock. This does check the actual data passed to stageBlock. Hence it is checking that the expected data is written as expected.

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobOutputStream.java
```diff
@@ -0,0 +1,314 @@
...
+            if (retryCount == MAX_ATTEMPT) {
+              throw new RuntimeException("Exceeded number of retries. Max attempts is: " + MAX_ATTEMPT, e);
+            }
+          }
+        }
+      }
+    }, blobThreadPool);
+
+    pendingUpload.add(future);
+    future.handle((aVoid, throwable) -> {
```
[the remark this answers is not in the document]
lakshmi-manasa-g: updated to add to pendingUpload only if not yet done

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobOutputStream.java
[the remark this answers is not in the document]
cameronlee314: You maybe you could have just one new exception type for any azure blob issue. Any new writers could use that one new exception type. Also note that there is a `SystemProducerException`. Maybe that could be useful in some places (not sure you would use it in this class, but maybe you could use it in your producer class?).

samza-azure/src/main/java/org/apache/samza/system/azureblob/producer/AzureBlobSystemProducer.java
cameronlee314: My question is if you should throw an exception or just ignore the call. The interface docs don't seem to indicate if a failure should actually be thrown or not when calling `stop` twice.

samza-azure/src/main/java/org/apache/samza/system/azureblob/producer/AzureBlobSystemProducer.java
cameronlee314: Aha, I missed that you check for `writer == null` again inside the synchronized block.

samza-azure/src/test/java/org/apache/samza/system/azureblob/avro/TestAzureBlobOutputStream.java
```diff
@@ -0,0 +1,347 @@
...
+@PrepareForTest({BlockBlobAsyncClient.class})
+public class TestAzureBlobOutputStream {
+  private ThreadPoolExecutor threadPool;
+  private ByteArrayOutputStream mockByteArrayOutputStream;
+  private static final int THRESHOLD = 100;
+  private BlockBlobAsyncClient mockBlobAsyncClient;
+  private AzureBlobOutputStream azureBlobOutputStream;
+  private static String randomString = "roZzozzLiR7GCEjcB0UsRUNgBAip8cSLGXQSo3RQvbIDoxOaaRs4hrec2s5rMPWgTPRY4UnE959worEtyhRjwUFnRnVuNFZ554yuPQCbI69qFkQX7MmrB4blmpSnFeGjWKjFjIRLFNVSsQBYMkr5jT4T83uVtuGumsjACVrpcilihdd194H8Y71rQcrXZoTQtw5OvmPicbwptawpHoRNzHihyaDVYgAs0dQbvVEu1gitKpamzYdMLFtc5h8PFZSVEB";
+  private static byte[] bytes = randomString.substring(0, THRESHOLD).getBytes();
+  private static byte[] compressedBytes = randomString.substring(0, THRESHOLD / 2).getBytes();
```
cameronlee314: Can you please use `final` and full capitalization for constant variables?

samza-azure/src/test/java/org/apache/samza/system/azureblob/avro/TestAzureBlobOutputStream.java
cameronlee314: It looks like you use the bytes in `testWriteLargerThanThreshold`, but not in `testWriteLargeRecordWithSmallRecordInBuffer`. If you believe that `testWriteLargeRecordWithSmallRecordInBuffer` does not need to use bytes due to already using bytes in `testWriteLargerThanThreshold`, please add a comment to `testWriteLargeRecordWithSmallRecordInBuffer`.

samza-azure/src/test/java/org/apache/samza/system/azureblob/avro/TestAzureBlobOutputStream.java
```diff
@@ -0,0 +1,347 @@
...
+  @Test
+  public void testWriteLargerThanThreshold() {
+    byte[] largeRecord = randomString.substring(0, 2 * THRESHOLD).getBytes();
+    byte[] largeRecordFirstHalf = randomString.substring(0, THRESHOLD).getBytes();
+    byte[] largeRecordSecondHalf = randomString.substring(THRESHOLD, 2 * THRESHOLD).getBytes();
+
+    byte[] compressB = randomString.substring(0, THRESHOLD / 2).getBytes();
+
+    doReturn(compressB).when(mockCompression).compress(largeRecordFirstHalf);
+    doReturn(compressB).when(mockCompression).compress(largeRecordSecondHalf);
```
cameronlee314: Please consider returning different "compressed bytes" for the first half and second half. It will make the test more complete to ensure that the correct bytes are used for each half.

samza-azure/src/test/java/org/apache/samza/system/azureblob/avro/TestAzureBlobOutputStream.java
```diff
@@ -124,11 +126,10 @@ public void testWriteLargerThanThreshold() {
...
     ArgumentCaptor<Flux> argument = ArgumentCaptor.forClass(Flux.class);
-    verify(mockBlobAsyncClient).stageBlock(eq(blockIdEncoded(0)), argument.capture(), eq((long) compressB.length));
-    verify(mockBlobAsyncClient).stageBlock(eq(blockIdEncoded(1)), argument.capture(), eq((long) compressB.length));
-    argument.getAllValues().forEach(flux -> {
-        Assert.assertEquals(ByteBuffer.wrap(compressB), flux.blockFirst());
-      });
+    verify(mockBlobAsyncClient).stageBlock(eq(blockIdEncoded(0)), argument.capture(), eq((long) compressB1.length));
+    verify(mockBlobAsyncClient).stageBlock(eq(blockIdEncoded(1)), argument.capture(), eq((long) compressB2.length));
+    Assert.assertEquals(ByteBuffer.wrap(compressB1), argument.getAllValues().get(0).blockFirst());
+    Assert.assertEquals(ByteBuffer.wrap(compressB2), argument.getAllValues().get(1).blockFirst());
```
cameronlee314: It is my understanding that `stageBlock` is part of the async flow. Therefore, it might be possible that `stageBlock` for `blockIdEncoded(1)` is called before `blockIdEncoded(0)`. I think that would mean that the values in the argument captor could be in the other order. To be more stable, you can have a separate argument captor for block 0 and block 1, and each captor will just have a single value.

samza-azure/src/main/java/org/apache/samza/system/azureblob/compression/CompressionFactory.java
```diff
@@ -25,6 +25,16 @@
  * based on the {@link org.apache.samza.system.azureblob.compression.CompressionType}.
  */
 public class CompressionFactory {
+  private static CompressionFactory compressionFactoryInstance = null;
+  private CompressionFactory() {}
+
+  public static CompressionFactory getInstance() {
```
cameronlee314: If you would like to do lazy initialization for the singleton, then you will need to be careful about concurrent calls to `getInstance`.
This class doesn't need lazy initialization, so you could just initialize directly when declaring the static variable. Then, you won't have to worry about concurrent calls.

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobAvroWriter.java
```diff
@@ -0,0 +1,360 @@
...
+      blobURL = String.format(BLOB_NAME_RANDOM_STRING_AVRO, blobURLPrefix,
+          UTC_FORMATTER.format(System.currentTimeMillis()), blobNumber, UUID.randomUUID().toString().substring(0, 8),
+          compression.getFileExtension());
+    } else {
+      blobURL = String.format(BLOB_NAME_AVRO, blobURLPrefix,
+          UTC_FORMATTER.format(System.currentTimeMillis()), blobNumber, compression.getFileExtension());
+    }
+    LOG.info("Creating new blob: {}", blobURL);
+    BlockBlobAsyncClient blockBlobAsyncClient = containerAsyncClient.getBlobAsyncClient(blobURL).getBlockBlobAsyncClient();
+    blobNumber++;
```
cameronlee314: Could you please describe what the `blobNumber` is used for?

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobOutputStream.java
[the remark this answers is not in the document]
lakshmi-manasa-g: Verified that `future.handle` is invoked even if its chained after `future` completes.

Test code used to verify
```
CompletableFuture<Void> future = CompletableFuture.runAsync(new Runnable() {
      @Override
      public void run() {
        System.out.println("Executor thread sleeping for X min");
        Thread.sleep(**X**);
        System.out.println("inside runAsync");
      }
    }, threadPool);

    pendingUpload.add(future);

    System.out.println("added future, now main thread sleeping for Y min");
    Thread.sleep(**Y**);
    System.out.println("Main thread woke up");

    future.handle((aVoid,throwable) -> {
      System.out.println("inside handle");
      pendingUpload.remove(future);
      return aVoid;
    });
```
Works for all values of X, Y (X=Y, X=Y=0, X\<Y and X\>Y)

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobOutputStream.java
```diff
@@ -277,24 +277,19 @@ public void run() {
                 + " failed for blockid: " + blockId + " due to exception. RetryCount: " + attemptCount;
```
cameronlee314: minor: Could you please change from using "Retry" here too?

samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobAvroWriter.java
```diff
@@ -108,6 +108,8 @@
   private final long maxRecordsPerBlob;
   private final boolean useRandomStringInBlobName;
   private final Object currentDataFileWriterLock = new Object();
+  // when either maxRecordsPerBlob or maxBlobSize is exceeded then a new blob is created
+  // with the same blobURLPrefix and blobNumber is suffixed to the blobURL to differentiate between the blobs
```
cameronlee314: I thought you used timestamp and a random string to differentiate between blobs. Do you still need `blobNumber`?

lakshmi-manasa-g: thanks for pointing out. `blobNumber` was introduced at a time when both timestamp and random string were generated inside AzureBlobSystemProducer. Have removed it now.

samza-azure/src/test/java/org/apache/samza/system/azureblob/avro/TestAzureBlobAvroWriter.java
```diff
@@ -0,0 +1,525 @@
...
+    BlobContainerAsyncClient mockContainerClient = PowerMockito.mock(BlobContainerAsyncClient.class);
+    azureBlobAvroWriter = spy(new AzureBlobAvroWriter(mockContainerClient,
+        mockMetrics, threadPool, THRESHOLD, 60000, blobUrlPrefix,
+        null, null, null, maxBlobSize, maxRecordsPerBlob, mockCompression, true));
+
+    DataFileWriter mockDataFileWriter1 = mock(DataFileWriter.class);
+    PowerMockito.whenNew(DataFileWriter.class).withAnyArguments().thenReturn(mockDataFileWriter1);
+
+    BlobAsyncClient mockBlobAsyncClient1 = mock(BlobAsyncClient.class);
+    doReturn(mockBlobAsyncClient1).when(mockContainerClient).getBlobAsyncClient(anyString());
```
cameronlee314: Would it be possible to check at least some part of the blob URL that you use to create the blob async client? For example, it should match some regex something like `test/[0-9]+/[0-9]+/[0-9]+/[0-9]+/[0-9]+-[0-9]+-.*.avro.gz`. This could verify some of the expected format and that there is some random string in there.

PR #1240, opened 2019-12-20
SAMZA-2423: Heartbeat failure causes incorrect container shutdown

abhishekshivanna:
**Symptom:** 
When a container heartbeat fails, the container shutdown
sequence is triggered and the Container is never restarted.

**Cause:**
When a container heartbeat fails, the container shutdown
sequence exists the Container with an exit code of `0` which
marks the container as `Completed` - preventing the JobCoordinator
from restarting the container.
The bug is caused by `containerException` overwritten with the value
returned by `listener.getContainerException` without checking if 
`containerException` was already set by the heartbeat monitor

**Changes:** 
The container can shutdown exceptionally in the following two ways:
1) Exception in the container
2) Heartbeat Expired
In both paths the ContainerLaunchUtil previously expected a
shared static variable to hold the exception. The change introduced
gets rid of the static variable and checks each path explicitly
and exits with code `1` in both cases.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatMonitor.java
+++ b/samza-core/src/main/java/org/apache/samza/container/ContainerHeartbeatMonitor.java
@@ -39,6 +39,7 @@ public class ContainerHeartbeatMonitor {
   private final Runnable onContainerExpired;
   private final ContainerHeartbeatClient containerHeartbeatClient;
   private boolean started = false;
+  private boolean heartbeatExpired = false;
 
   public ContainerHeartbeatMonitor(Runnable onContainerExpired, ContainerHeartbeatClient containerHeartbeatClient) {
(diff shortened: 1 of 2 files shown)


mynameborat:
Seems like the bug is we end up overwriting the `containerException` with `listener.getContainerException` without checking if it has already been populated. Due to this, it is possible that container listener doesn't have any exception because it was requested to shut itself down but the request was due to expired heartbeat. 

Is that correct?

abhishekshivanna:
@mynameborat Correct !

mynameborat:
can we add a unit test for this?

mynameborat:
> can we add a unit test for this?

abhishekshivanna:
I agree, looks like there is no way to mock objects that `ContainerLaunchUtil` instantiates in order to test this.

PR #1241, opened 2019-12-20
SAMZA-2424: AM should cache and serve serialized Job Model to containers

prateekm:
Symptom: AM runs out of memory or gets stuck in GC loop during job deployment.
 
Cause: During startup, containers fetch the JobModel and configs from the job coordinator http endpoint. For jobs with a large number of containers this can mean many concurrent requests to the JC Servlet for the large job model. The servlet currently serializes the JobModel to JSON for every incoming request. This can lead to excessive memory utilization and GC, leading to slow response times, AM running out of memory, or AM getting stuck in GC loop.
 
Changes: JobModelManager now serializes the JobModel when creating the JobServlet. JobServlet now serves this serialized JobModel instead of serializing every time.
 
Tests: Local performance test results using Apache Bench and monitoring memory utilization using YourKit: https://docs.google.com/document/d/1Qy4ScC9aLNZaJQuUBvtyoXdilUhkVtYjb5Mzwzv-DV8/edit?usp=sharing

API Changes: None
 
Upgrade Instructions: None
 
Usage Instructions: None


shanthoosh:
Please correct me if i'm wrong.


JobModel currently contains the container-to-host-locality information, accessible via the high-level field `all-container-locality`.

In a typical run of a samza-job, containers could move around different NM hosts due to transient h/w failures (or containers could be explicitly moved around by using our powerful container placement API)


So if we cache the JobModel and after that the container-locality changes, then subsequent AM  HTTP reads of JobModel could potentially be serving inconsistent JobModel. 

Sample JobModel taken from a samza job 

```bash
[svenkata@host ~]$ cat ~/job-model  | jq . | head -n 10 
{
  "all-container-locality": {
    "68-0": "host-1",
    "135-0": "host-2",
    "173-0": "host-3",
    "158-0": "host-4",
    "45-0": "host-5",
    "112-0": "host-6",
    "109": "host-7",
    "108": "host-8",
```

In the above sample JobModel, the active container '68-0' runs in host-1. 

Currently the ApplicationMaster process is used as the source of truth for JobModel. Some of our internal samza-admin monitors uses it to garbage-collect the left-over local RocksDB state-stores from disk(to reclaim disk-space). Just curious, if the above API change would affect our state-store GC-monitor.

samza-core/src/main/scala/org/apache/samza/coordinator/JobModelManager.scala
```diff
@@ -93,15 +94,17 @@ object JobModelManager extends Logging {
       val streamMetadataCache = new StreamMetadataCache(systemAdmins, 0)
       val grouperMetadata: GrouperMetadata = getGrouperMetadata(config, localityManager, taskAssignmentManager, taskPartitionAssignmentManager)
 
-      val jobModel: JobModel = readJobModel(config, changelogPartitionMapping, streamMetadataCache, grouperMetadata)
-      jobModelRef.set(new JobModel(jobModel.getConfig, jobModel.getContainers, localityManager))
+      val jobModel = readJobModel(config, changelogPartitionMapping, streamMetadataCache, grouperMetadata)
+      val jobModelToServe = new JobModel(jobModel.getConfig, jobModel.getContainers, localityManager)
+      val serializedJobModelToServe = SamzaObjectMapper.getObjectMapper().writeValueAsBytes(jobModelToServe)
+      jobModelRef.set(serializedJobModelToServe)
```
rmatharu: Maybe serializedJobModelRef ?

samza-core/src/main/scala/org/apache/samza/coordinator/server/JobServlet.scala
```diff
@@ -22,19 +22,22 @@ package org.apache.samza.coordinator.server
...
+import javax.servlet.http.{HttpServlet, HttpServletRequest, HttpServletResponse}
 
 /**
- * A servlet that dumps the job model for a Samza job.
+ * Serves the JSON serialized job model for the job.
  */
-class JobServlet(jobModelRef: AtomicReference[JobModel]) extends ServletBase with Logging {
-  protected def getObjectToWrite() = {
+class JobServlet(jobModelRef: AtomicReference[Array[Byte]]) extends HttpServlet {
+  override protected def doGet(request: HttpServletRequest, response: HttpServletResponse) {
```
rmatharu: Should we update the 
doGet implementation in ServletBase ?

rmatharu:
As @shanthoosh  pointed out, state-store-GC-monitor uses containers' locality mapping to determine the tasks on a given host (if the app is currently running), with this change the monitor could end up using a previous app-attempt's locality mappings.

An alternative way may be to expose separate calls for cached vs. non-cached JobModel?

samza-core/src/main/scala/org/apache/samza/coordinator/server/JobServlet.scala
prateekm: Good catch. Meant to delete `ServletBase`, it's no longer used.

prateekm:
@shanthoosh @rmatharu Thanks for the review.

This PR does not affect LocalStoreMonitor:
1. This PR updates the only place where jobModelRef is set to store the serialized job model instead of unserialized job model. So this shouldn't affect any usages of jobModelRef. Btw, I don't think this field needs to be an AtomicReference since its only updated once. @shanthoosh do you know if JobModelManager#apply is called more than once in any scenario (YARN or Standalone)? If not, I can change it to a regular field.

2. LocalStoreMonitor doesn't use the JobModel served from the AM. It reads the coordinator stream and creates the JobModel itself, in a rather roundabout way. The code path is: `LocalStoreMonitor -> JobsClient#getTasks ---HTTP Request---> TasksResource#getTasks -> SamzaTaskProxy#getTasks -> CoordinatorStreamSystemConsumer/LocalityManger read`

shanthoosh:
@prateekm

1. LocalStoreMonitor in OSS is not used internally within LI. We'd re-implemented the same logic internally in C for GDPR purposes. The re-implementation queries the AM, AFAIK, to get the JobModel. 
2. Last I checked, the samza dashboard also queries the AM endpoint to get the JobModel. Might be better to check if it would break any existing dashboard functionality with this caching change(if we return inconsistent container-to-host mapping).
3. I think you're right in that we do not call the JobModelManager.apply is not invoked multiple times.

prateekm:
@shanthoosh See point 1. If those tools are querying the AM for container locality, they're already affected since the job model served by AM is not currently updated with locality information after the initial read. This PR does not change that behavior. 

Also, I think Samza dashboard uses the Yarn RM APIs for finding container location, not the AM job model.

prateekm:
@shanthoosh clarified that the JobModel object isn't immutable, it re-reads the container locality from the metadata store every time getAllContainerLocality() is called (e.g.during serialization). Let me see what we can do to handle that, it might require a bigger change to how the JobModel is created.

mynameborat:
@prateekm Now that the locality [refactor PR](https://github.com/apache/samza/commit/f7f9f3c7905b047f262383bcc64ebb01ab73f421) is in, can you pick this up when you get a chance?

mynameborat (approved):
Looks good to me; Feel free to merge after the checks have passed.

PR #1242, opened 2020-01-02
SAMZA-2425: Add support for sub-query in joins

atoomula:
**Symptom**: Using sub-query with a join fails today with an unsupported exception. 
**Cause**: We have never supported sub-query in joins due to the way we have been parsing sql statements which was fixed recently.
 **Fix**: A simple code enhancement to handle logical filter while traversing the calcite logical plan to get the table IO. 
**Tests**: Added a Sql End-To-End test for sub-query and join.

### DIFF
--- a/samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
+++ b/samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
@@ -28,6 +28,8 @@ import org.apache.calcite.adapter.enumerable.EnumerableTableScan;
 import org.apache.calcite.plan.RelOptUtil;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.core.JoinRelType;
+import org.apache.calcite.rel.core.TableScan;
+import org.apache.calcite.rel.logical.LogicalFilter;
 import org.apache.calcite.rel.logical.LogicalJoin;
 import org.apache.calcite.rel.logical.LogicalProject;
 import org.apache.calcite.rex.RexCall;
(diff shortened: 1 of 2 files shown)


samza-sql/src/main/java/org/apache/samza/sql/translator/JoinTranslator.java
```diff
@@ -292,17 +294,27 @@ private String dumpRelPlanForNode(RelNode relNode) {
...
-    // We are returning the sourceConfig for the table as null when the table is in another join rather than an output
-    // table, that's because the output of stream-table join is considered a stream.
-    if (relNode.getInputs().size() > 1) {
+    if (relNode instanceof LogicalFilter) {
+      return resolveSQlIOForTable(((LogicalFilter) relNode).getInput(), context);
+    }
+
+    // We return null for table IO as the table seems to be involved in another join. The output of stream-table join
+    // is considered a stream. Hence, we return null for the table.
+    if (relNode instanceof LogicalJoin && relNode.getInputs().size() > 1) {
```
shanthoosh: 1. There're other kinds of `RelNode` which are not covered by `instanceOf` checks in recursive function. Do we need to handle other kinds of `RelNode` here? If not, why?
2. Can you please add a unit-test in JoinTranslator for these changes? 
3. Just curious. Would the schema validations be applied on the result of the sub-queries within joins?

atoomula: Very good questions.
1. In Samza Sql, we currently support only filters, projection and joins along with scan. Hence, only they are handled in this function.
2. There are end-to-end tests for each of these if conditions and for the new addition.
3. Schema validations for joins is not yet implemented.

PR #1246, opened 2020-01-10
SAMZA-2409: Update YarnJob to construct job submission env varible

kw2542:
Design:

Changes:
1. Depending the on the existence of job.config.loader.factory, YarnJob will alternatively wraps job submission configs to Yarn.
2. In a separate PR, ClusterBasedJobCoordinator will be updated to read full job config from loader or coordinator stream based on the existence of submission config environment variable.

API Changes:

Upgrade Instructions:

Usage Instructions:

Tests:
Unit Tests

### DIFF
--- a/samza-yarn/src/main/scala/org/apache/samza/job/yarn/YarnJob.scala
+++ b/samza-yarn/src/main/scala/org/apache/samza/job/yarn/YarnJob.scala
@@ -173,10 +173,16 @@ object YarnJob extends Logging {
   @VisibleForTesting
   private[yarn] def buildEnvironment(config: Config, yarnConfig: YarnConfig,
     jobConfig: JobConfig): Map[String, String] = {
-    val coordinatorSystemConfig = CoordinatorStreamUtil.buildCoordinatorStreamConfig(config)
     val envMapBuilder = Map.newBuilder[String, String]
-    envMapBuilder += ShellCommandConfig.ENV_COORDINATOR_SYSTEM_CONFIG ->
-      Util.envVarEscape(SamzaObjectMapper.getObjectMapper.writeValueAsString(coordinatorSystemConfig))
+    if (jobConfig.getConfigLoaderFactory.isPresent) {
+      envMapBuilder += ShellCommandConfig.ENV_SUBMISSION_CONFIG ->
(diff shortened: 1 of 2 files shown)


samza-yarn/src/main/scala/org/apache/samza/job/yarn/YarnJob.scala
```diff
@@ -173,10 +173,16 @@ object YarnJob extends Logging {
   @VisibleForTesting
   private[yarn] def buildEnvironment(config: Config, yarnConfig: YarnConfig,
     jobConfig: JobConfig): Map[String, String] = {
-    val coordinatorSystemConfig = CoordinatorStreamUtil.buildCoordinatorStreamConfig(config)
     val envMapBuilder = Map.newBuilder[String, String]
-    envMapBuilder += ShellCommandConfig.ENV_COORDINATOR_SYSTEM_CONFIG ->
-      Util.envVarEscape(SamzaObjectMapper.getObjectMapper.writeValueAsString(coordinatorSystemConfig))
+    if (jobConfig.getConfigLoaderFactory.isPresent) {
+      envMapBuilder += ShellCommandConfig.ENV_SUBMISSION_CONFIG ->
+        Util.envVarEscape(SamzaObjectMapper.getObjectMapper.writeValueAsString(config))
```
bkonold: Is "config" here intended to be only the subset of configs required for submissions (the subset you outlined in SEP-23)?

I may be misunderstanding the flow somewhere, but won't ApplicationRunnerMain end up loading the full (pre-planning) config from the properties file which will end up getting passed through to here?

kw2542: Very good question, in the final approved SEP-23, we will not read any properties/config file during submission anymore. All submission related configs will be provided explicitly using --config.

ApplicationRunnerMain will be updated in a separate PR to reflect this change as well. Full job config will only be read on AM, and then apply the overridden values we get from ApplicationRunnerMain will be passed to AM to override the value we load.

bkonold: Sounds good. Thanks for clarifying.

PR #1247, opened 2020-01-10
SAMZA-2408: Update RemoteApplicationRunner to submit job only

kw2542:
Design:

Changes:

1. Depending on the existence of job.config.loader.factory, RemoteApplicationRunner will alternatively only submit the job, without planning.
2. Late initialize RemoteJobPlanner only when needed to avoid executing user code in submitting only mode.

API Changes:

Upgrade Instructions:

Usage Instructions:

Tests:
End to end tested with Hello Samza job.


samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -57,13 +55,21 @@
...
 
   @Override
   public void run(ExternalContext externalContext) {
+    if (new JobConfig(config).getConfigLoaderFactory().isPresent()) {
+      JobRunner runner = new JobRunner(config);
+      runner.submit();
+      return;
+    }
+
+    // TODO: Clean this up once SAMZA-2405 is completed when legacy flow is removed.
```
bkonold: Would it be helpful to create a ticket for this cleanup work (for all changes related to the SEP) so we don't lose track of it?

kw2542: Good point, I will create a ticket for it.

bkonold:
Change LGTM. Can you include a unit test for the behavior?

samza-core/src/test/java/org/apache/samza/runtime/TestRemoteApplicationRunner.java
```diff
@@ -75,9 +75,20 @@ public void testWaitForFinishTimesout() {
     assertFalse("Application finished before the timeout.", finished);
   }
 
+  @Test
+  public void testRunWithConfigLoaderFactoryPresent() {
```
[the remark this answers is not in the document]
kw2542: Haha, I figured that too. In general, I prefer not to mess with production code for the sake of unit tests.

The current unit actually test the new flow, I passed null SamzaApplication to RemoteApplicationRunner's constructor. In RemoteApplicationRunner#run, if it goes the legacy flow, it will throw exception when constructing RemoteJobPlanner where SamzaApplication is required. Do you think this is enough?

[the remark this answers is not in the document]
bkonold: There is some precedence elsewhere in samza code; ZkJobCoordinator.readJobModelFromMetadataStore for example uses a similar method structure + spy in unit test.

You are correct that the test written now does ensure that the legacy flow isn't triggered, but it does little to verify the behavior of the new flow. I won't block the PR on this though.

samza-core/src/main/java/org/apache/samza/runtime/RemoteApplicationRunner.java
```diff
@@ -85,7 +91,7 @@ public void kill() {
     // since currently we only support single actual remote job, we can get its status without
     // building the execution plan.
     try {
-      JobConfig jc = new JobConfig(JobPlanner.generateSingleJobConfig(appDesc.getConfig()));
+      JobConfig jc = new JobConfig(JobPlanner.generateSingleJobConfig(config));
```
cameronlee314: This is a question for `kill`, `status`, and `waitForFinish`:
Should you be calling `JobPlanner.generateSingleJobConfig` even when config loader factory is specified? It seems like you would want to call `JobRunner`, similar to how you changed it for `run`.

kw2542: `JobPlanner.generateSingleJobConfig` is a util method which overrides `job.name` and `job.id` based on `app.name` and `app.id`, which I suppose is safe to keep. 

`kill`, `status` and `waitForFinish` are still using calling JobRunner.

cameronlee314: But in `run`, you don't call `JobPlanner.generateSingleJobConfig`. For consistency, it seems like you should call `generateSingleJobConfig` for all cases or for no cases.

kw2542: Got you, `JobPlanner.generateSingleJobConfig` is not being used in `run()` in the legacy code path neither, do you think we should explore this separately from this PR?

cameronlee314: It looks like `ExecutionPlanner.createJobGraph` (part of planning) uses `JobPlanner.generateSingleJobConfig`, so it looks like `JobPlanner.generateSingleJobConfig` is used in `run` in the legacy path. Can you please check if I am reading the code correctly here?

kw2542: Updated.

kw2542: Thanks for pointing this out!

samza-core/src/test/java/org/apache/samza/runtime/TestRemoteApplicationRunner.java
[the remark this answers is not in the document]
cameronlee314: +1 that there is a bit of missing test coverage here. It would be nice if this class was set up better for improved testing, but I also won't block the PR on this.

PR #1248, opened 2020-01-10
SAMZA-2410: Update ClusterBasedJobCoordinator config retrieval logic from loader.

kw2542:
Design:

Changes:

1. Make ClusterBasedJobCoordinator's constructor to be private
2. Add static methods createFromMetadataStore() and createFromConfigLoader() to initialize ClusterBasedJobCoordinator instead.
3. Update CoordinatorStreamStore to its interface MetadataStore to be more generic.
4. Based on the existence of ENV_SUBMISSION_CONFIG, ClusterBasedJobCoordinator will alternatively deserilize ENV_SUBMISSION_CONFIG and load full job config from config loader factory.
5. Execute planning, create diagnostics stream and persist full job config back to coordinator stream when loading full job config from config loader.

API Changes:

Upgrade Instructions:

Usage Instructions:

Tests:
End to end tested with Hello Samza job.


samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -178,9 +185,35 @@
...
+
+    if (jobConfig.getConfigLoaderFactory().isPresent()) {
+      // load full job config with ConfigLoader
+      Config originalConfig = ConfigUtil.loadConfig(coordinatorSystemConfig);
+
+      // Execute planning
+      ApplicationDescriptorImpl<? extends ApplicationDescriptor>
+          appDesc = ApplicationDescriptorUtil.getAppDescriptor(ApplicationUtil.fromConfig(originalConfig), originalConfig);
+      RemoteJobPlanner planner = new RemoteJobPlanner(appDesc);
+      List<JobConfig> jobConfigs = planner.prepareJobs();
```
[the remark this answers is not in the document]
kw2542: Good question, both RemoteApplicationRunner and LocalJobPlanner themselves only works for single-job applications anyway, we may update them, but I think it is better to do it separately. It also relies on our plan to support multi job applications too.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -480,20 +512,41 @@ private static void executeRunClusterBasedJobCoordinatorForClass(Class<?> cluste
...
+    final String submissionEnv = System.getenv(ShellCommandConfig.ENV_SUBMISSION_CONFIG());
+
+    if (submissionEnv != null) {
+      Config submissionConfig;
+      try {
+        //Read and parse the coordinator system config.
+        LOG.info("Parsing submission config {}", submissionEnv);
+        submissionConfig =
+            new MapConfig(SamzaObjectMapper.getObjectMapper().readValue(submissionEnv, Config.class));
+        LOG.info("Using the submission config: {}.", submissionEnv);
```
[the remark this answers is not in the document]
kw2542: Fixed.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -178,9 +185,35 @@
...
+          appDesc = ApplicationDescriptorUtil.getAppDescriptor(ApplicationUtil.fromConfig(originalConfig), originalConfig);
+      RemoteJobPlanner planner = new RemoteJobPlanner(appDesc);
+      List<JobConfig> jobConfigs = planner.prepareJobs();
+
+      if (jobConfigs.size() != 1) {
+        throw new SamzaException("Only support single remote job is supported.");
+      }
+
+      // Merge with default coordinator stream config
+      config = ConfigUtil.override(jobConfigs.get(0), CoordinatorStreamUtil.buildCoordinatorStreamConfig(jobConfigs.get(0)));
```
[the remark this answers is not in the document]
kw2542: Good question, you are right, we should not need to merge it with the full job config, we only need to build it from the full job config and initialize CoordinatorStreamStore.

samza-core/src/main/java/org/apache/samza/util/ConfigUtil.java
```diff
@@ -67,4 +72,42 @@ public static Config applyRewriter(Config config, String rewriterName) {
...
+  public static Config loadConfig(Config original) {
+    JobConfig jobConfig = new JobConfig(original);
+    Config fullConfig = original;
+
+    if (jobConfig.getConfigLoaderFactory().isPresent()) {
+      ConfigLoaderFactory factory = ReflectionUtil.getObj(jobConfig.getConfigLoaderFactory().get(), ConfigLoaderFactory.class);
+      ConfigLoader loader = factory.getLoader(original.subset(ConfigLoaderFactory.CONFIG_LOADER_PROPERTIES_PREFIX));
+      // overrides config loaded with original config, which may contain overridden values.
+      fullConfig = override(ConfigUtil.rewriteConfig(loader.getConfig()), original);
+    }
```
bkonold: Is it an exceptional case for ConfigLoaderFactory to not be present here?

kw2542: Eventually Yes, but during the migration, there will be cases where jobs are still following the legacy submission flow, but still goes to this new method in code.

For example, this util will be called by LocalApplicationRunner too, which will first support both legacy and new flow.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -178,9 +185,33 @@
   public ClusterBasedJobCoordinator(Config coordinatorSystemConfig) {
```
bkonold: Should we rename this to something more general now that it won't always be just the coordinator system configs?

kw2542: I was planning to rename this variable when cleaning up the legacy flow, but we can choose a generic name in the mid now as well, what do you think?

bkonold: I think changing it now would be less misleading in the interim. "jobCoordinatorConfig"?

kw2542: Renamed to jobCoordinatorConfig and updated the javadoc as well.

samza-core/src/main/java/org/apache/samza/util/ConfigUtil.java
cameronlee314: The method name `loadConfig` is a bit misleading if it just might return the config that is passed in. If you still prefer to allow this to succeed when `ConfigLoaderFactory` is not present, then maybe at least rename the method to help clarify what it might do, so that someone doesn't accidentally call it in an unexpected way.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -480,20 +510,41 @@ private static void executeRunClusterBasedJobCoordinatorForClass(Class<?> cluste
...
-      LOG.info("Parsing coordinator system config {}", coordinatorSystemEnv);
-      coordinatorSystemConfig =
-          new MapConfig(SamzaObjectMapper.getObjectMapper().readValue(coordinatorSystemEnv, Config.class));
-      LOG.info("Using the coordinator system config: {}.", coordinatorSystemConfig);
-    } catch (IOException e) {
-      LOG.error("Exception while reading coordinator stream config", e);
-      throw new SamzaException(e);
+    final String submissionEnv = System.getenv(ShellCommandConfig.ENV_SUBMISSION_CONFIG());
+
+    if (submissionEnv != null) {
```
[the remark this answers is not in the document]
kw2542: Good catch. Updated.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -172,15 +179,39 @@
...
+
+      if (jobConfigs.size() != 1) {
+        throw new SamzaException("Only support single remote job is supported.");
+      }
+
+      config = jobConfigs.get(0);
+      coordinatorStreamStore = new CoordinatorStreamStore(CoordinatorStreamUtil.buildCoordinatorStreamConfig(config), metrics);
+      coordinatorStreamStore.init();
+      CoordinatorStreamUtil.writeConfigToCoordinatorStream(config, true);
+      DiagnosticsUtil.createDiagnosticsStream(config);
```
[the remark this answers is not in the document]
kw2542: Added a comment

samza-core/src/main/java/org/apache/samza/util/ConfigUtil.java
```diff
@@ -67,4 +72,42 @@ public static Config applyRewriter(Config config, String rewriterName) {
...
+   */
+  public static Config loadConfig(Config original) {
+    JobConfig jobConfig = new JobConfig(original);
+    Config fullConfig = original;
+
+    if (jobConfig.getConfigLoaderFactory().isPresent()) {
+      ConfigLoaderFactory factory = ReflectionUtil.getObj(jobConfig.getConfigLoaderFactory().get(), ConfigLoaderFactory.class);
+      ConfigLoader loader = factory.getLoader(original.subset(ConfigLoaderFactory.CONFIG_LOADER_PROPERTIES_PREFIX));
+      // overrides config loaded with original config, which may contain overridden values.
+      fullConfig = override(ConfigUtil.rewriteConfig(loader.getConfig()), original);
```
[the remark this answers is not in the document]
kw2542: Good point, we should override the original config with provided overrides and then invoke config rewriters. 

Updated.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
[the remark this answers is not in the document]
cameronlee314: The comment was resolved, but it looks like you are still just checking for null. Did you intend to check for blank string too?

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -480,20 +485,101 @@ private static void executeRunClusterBasedJobCoordinatorForClass(Class<?> cluste
...
+   *
+   * @param metadataStoreConfig to initialize {@link MetadataStore}
+   * @return {@link ClusterBasedJobCoordinator}
+   */
+  // TODO SAMZA-2432: Clean this up once SAMZA-2405 is completed when legacy flow is removed.
+  public static ClusterBasedJobCoordinator createFromMetadataStore(Config metadataStoreConfig) {
+    MetricsRegistryMap metrics = new MetricsRegistryMap();
+
+    CoordinatorStreamStore coordinatorStreamStore = new CoordinatorStreamStore(metadataStoreConfig, metrics);
+    coordinatorStreamStore.init();
```
cameronlee314: It looks like you already call `init` in the shared constructor. Please call it only once.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -480,20 +485,101 @@ private static void executeRunClusterBasedJobCoordinatorForClass(Class<?> cluste
...
+
+  /**
+   * Initialize {@link ClusterBasedJobCoordinator} with coordinator stream config, full job config will be fetched from
+   * coordinator stream.
+   *
+   * @param metadataStoreConfig to initialize {@link MetadataStore}
+   * @return {@link ClusterBasedJobCoordinator}
+   */
+  // TODO SAMZA-2432: Clean this up once SAMZA-2405 is completed when legacy flow is removed.
+  public static ClusterBasedJobCoordinator createFromMetadataStore(Config metadataStoreConfig) {
```
cameronlee314: Does this need to be public? If you need in tests, could you make it package private and mark as `VisibleForTesting`?

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -480,20 +485,101 @@ private static void executeRunClusterBasedJobCoordinatorForClass(Class<?> cluste
...
+  }
+
+  /**
+   * Initialize {@link ClusterBasedJobCoordinator} with submission config, full job config will be fetched using
+   * specified {@link org.apache.samza.config.ConfigLoaderFactory}
+   *
+   * @param submissionConfig specifies {@link org.apache.samza.config.ConfigLoaderFactory}
+   * @return {@link ClusterBasedJobCoordinator}
+   */
+  public static ClusterBasedJobCoordinator createFromConfigLoader(Config submissionConfig) {
```
cameronlee314: Does this need to be public? If you need in tests, could you make it package private and mark as `VisibleForTesting`?

samza-core/src/main/java/org/apache/samza/util/ConfigUtil.java
cameronlee314: Is there something you can do to clarify that this may not actually load any additional configs?

samza-core/src/test/java/org/apache/samza/clustermanager/TestClusterBasedJobCoordinator.java
```diff
@@ -198,4 +203,31 @@ public void testRunWithClassLoader() throws Exception {
...
+
+    PowerMockito.doAnswer(invocation -> invocation.getArgumentAt(0, Config.class))
+        .when(ClusterBasedJobCoordinator.class, "prepareJob", any());
+    PowerMockito.whenNew(ClusterBasedJobCoordinator.class).withAnyArguments().thenReturn(mock(ClusterBasedJobCoordinator.class));
+    PowerMockito.whenNew(CoordinatorStreamStore.class).withAnyArguments().thenReturn(mock(CoordinatorStreamStore.class));
+
+    ClusterBasedJobCoordinator.createFromConfigLoader(new MapConfig(config));
+
+    verifyPrivate(ClusterBasedJobCoordinator.class).invoke("prepareJob", any());
+    verifyNew(ClusterBasedJobCoordinator.class).withArguments(any(), any(), any());
```
cameronlee314: Is it practical to do any more-specific validation on the arguments here?

samza-core/src/test/java/org/apache/samza/util/TestConfigUtil.java
```diff
@@ -161,6 +162,31 @@ public void testApplyRewriterClassDoesNotExist() {
...
 
+  @Test
+  public void testLoadConfigWithoutLoader() {
+    Map<String, String> config = new HashMap<>();
+    config.put(JobConfig.JOB_NAME, "new-test-job");
+
+    Config actual = ConfigUtil.loadConfig(new MapConfig(config));
+
+    assertEquals(config.size(), actual.size());
+    assertEquals("new-test-job", actual.get(JobConfig.JOB_NAME));
```
cameronlee314: Can this just be `assertEquals(config, actual)`?

samza-core/src/test/java/org/apache/samza/util/TestConfigUtil.java
```diff
@@ -161,6 +162,31 @@ public void testApplyRewriterClassDoesNotExist() {
...
+  }
+
+  @Test
+  public void testLoadConfigWithLoader() {
```
cameronlee314: Should this also test config rewriting and overriding?

samza-core/src/main/java/org/apache/samza/util/ConfigUtil.java
```diff
@@ -67,4 +72,42 @@ public static Config applyRewriter(Config config, String rewriterName) {
...
+
+  /**
+   * Overrides original config with overridden values.
+   *
+   * @param original config to be overridden.
+   * @param overrides overridden values.
+   * @return the overridden config.
+   */
+  @SafeVarargs
+  public static Config override(Config original, Map<String, String>... overrides) {
```
cameronlee314: If this is public, can you please add a unit test for it? Or you could make it private.
You could also simplify it to only take in one extra argument since that is currently the only way it is used.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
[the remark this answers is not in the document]
kw2542: My bad, changes are lost during updates. Updated again.

samza-core/src/main/java/org/apache/samza/util/ConfigUtil.java
kw2542: Updated to disallow the case when ConfigLoaderFactory is not present.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
kw2542: Update to call outside constructor as we need it to be initialized before we can read full job config from it in the legacy flow.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
kw2542: Update to private.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
kw2542: Update to package private and mark as `VisibleForTesting`

samza-core/src/test/java/org/apache/samza/util/TestConfigUtil.java
kw2542: Update the unit test to check ConfigException instead as we disallow config w/o loader.

samza-core/src/main/java/org/apache/samza/util/ConfigUtil.java
kw2542: Updated to private

samza-core/src/test/java/org/apache/samza/util/TestConfigUtil.java
kw2542: Updated to include both overrides and rewrites.

samza-core/src/test/java/org/apache/samza/clustermanager/TestClusterBasedJobCoordinator.java
kw2542: Updated the unit test have to explicit validation on arguments.

PR #1249, opened 2020-01-10
Refactor LocalApplicationRunner

kw2542:
Issues: LocalApplicationRunner has multiple constructors and follows different initialization logic.

Changes:

1. Refactor all constructors to end up with the same private constructor.
2. Align "static final" and "final static" to "static final"
3. Group class variales based on accessibility
4. Minor fixs on inconsistent styling, return etc.

API Changes:
None

Upgrade Instructions: 
None
 
Usage Instructions:
None

Tests:
Unit Tests


prateekm:
@kw2542 Please document any public API changes in the description.

PR #1250, opened 2020-01-11
Clean up unused org.apache.samza.autoscaling module

kw2542:
Issues: samza-autoscaling module is not used.

Changes: remove the unused module.

API Changes:
None

Upgrade Instructions:
None

Usage Instructions:
None

Tests: build


shanthoosh:
@kw2542 Thanks for the changes. Merged the patch to trunk.

PR #1251, opened 2020-01-14
SAMZA-2431: Fix the checkpoint and changelog topic auto-creation.

shanthoosh:
Symptom: Checkpoint and changelog kafka topics of a samza job may be created with cleanup.policy set to 'delete' instead of 'compact' for certain cases.

Cause: 
- *Checkpoint:* The control-flow in KafkaStreamSpec to build checkpoint spec swallows the essential kafka-topic configuration rather it passes empty configuration bag to kafka-broker. This behavior was introduced in SAMZA-2339. 
- *Changelog:* The change-log topic configurations were incorrectly generated when the RocksDB store TTL is set to -1 by the user. This behavior was introduced in SAMZA-1929.

Changes: Fix the topic-creation control-flow for the metadata topics and generate the correct topic-configurations.

Tests: Added unit tests to validate that the expected topic configuration bag was generated for both checkpoint and changelog topics.

API Changes: None

Upgrade Instructions: None

Usage Instructions: None

### DIFF
--- a/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
+++ b/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
@@ -541,8 +541,8 @@ public class KafkaSystemAdmin implements SystemAdmin {
           new KafkaStreamSpec(spec.getId(), spec.getPhysicalName(), systemName, 1, coordinatorStreamReplicationFactor,
               coordinatorStreamProperties);
     } else if (spec.isCheckpointStream()) {
-      kafkaSpec = KafkaStreamSpec.fromSpec(StreamSpec.createCheckpointStreamSpec(spec.getPhysicalName(), systemName))
-              .copyWithReplicationFactor(Integer.parseInt(new KafkaConfig(config).getCheckpointReplicationFactor().get()));
+      kafkaSpec = KafkaStreamSpec.fromSpec(spec)
+                                 .copyWithReplicationFactor(Integer.parseInt(new KafkaConfig(config).getCheckpointReplicationFactor().get()));
     } else if (intermediateStreamProperties.containsKey(spec.getId())) {
       kafkaSpec = KafkaStreamSpec.fromSpec(spec);
       Properties properties = kafkaSpec.getProperties();
--- a/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
+++ b/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
@@ -541,8 +541,11 @@ public class KafkaSystemAdmin implements SystemAdmin {
           new KafkaStreamSpec(spec.getId(), spec.getPhysicalName(), systemName, 1, coordinatorStreamReplicationFactor,
               coordinatorStreamProperties);
     } else if (spec.isCheckpointStream()) {
-      kafkaSpec = KafkaStreamSpec.fromSpec(spec)
-                                 .copyWithReplicationFactor(Integer.parseInt(new KafkaConfig(config).getCheckpointReplicationFactor().get()));
+      Properties checkpointTopicProperties = new Properties();
+      checkpointTopicProperties.putAll(spec.getConfig());
+      kafkaSpec = KafkaStreamSpec.fromSpec(StreamSpec.createCheckpointStreamSpec(spec.getPhysicalName(), spec.getSystemName()))
+              .copyWithReplicationFactor(Integer.parseInt(new KafkaConfig(config).getCheckpointReplicationFactor().get()))
+              .copyWithProperties(checkpointTopicProperties);
     } else if (intermediateStreamProperties.containsKey(spec.getId())) {
       kafkaSpec = KafkaStreamSpec.fromSpec(spec);
       Properties properties = kafkaSpec.getProperties();
--- a/samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
+++ b/samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
@@ -322,8 +322,7 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
     // Adjust changelog topic setting, when TTL is set on a RocksDB store
     //  - Disable log compaction on Kafka changelog topic
     //  - Set topic TTL to be the same as RocksDB TTL
-    val storeTTLkey = "stores.%s.rocksdb.ttl.ms" format name
-    Option(config.get(storeTTLkey)) match {
+    Option(config.get("stores.%s.rocksdb.ttl.ms" format name)) match {
       case Some(rocksDbTtl) =>
         if (!rocksDbTtl.isEmpty && rocksDbTtl.toInt < 0) {
           kafkaChangeLogProperties.setProperty("cleanup.policy", "compact")
--- a/samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
+++ b/samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
@@ -325,8 +325,9 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
     val storeTTLkey = "stores.%s.rocksdb.ttl.ms" format name
     Option(config.get(storeTTLkey)) match {
       case Some(rocksDbTtl) =>
-        if (config.getInt(storeTTLkey, 0) < 0) {
+        if (!rocksDbTtl.isEmpty && rocksDbTtl.toInt < 0) {
           kafkaChangeLogProperties.setProperty("cleanup.policy", "compact")
+          kafkaChangeLogProperties.setProperty("max.message.bytes", getChangelogStreamMaxMessageByte(name))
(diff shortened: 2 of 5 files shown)


shanthoosh:
@prateekm Can you please take a look at this patch.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -541,8 +541,8 @@ public KafkaStreamSpec toKafkaSpec(StreamSpec spec) {
           new KafkaStreamSpec(spec.getId(), spec.getPhysicalName(), systemName, 1, coordinatorStreamReplicationFactor,
               coordinatorStreamProperties);
     } else if (spec.isCheckpointStream()) {
-      kafkaSpec = KafkaStreamSpec.fromSpec(StreamSpec.createCheckpointStreamSpec(spec.getPhysicalName(), systemName))
-              .copyWithReplicationFactor(Integer.parseInt(new KafkaConfig(config).getCheckpointReplicationFactor().get()));
+      kafkaSpec = KafkaStreamSpec.fromSpec(spec)
```
prateekm: Minor: Fix indentation.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -316,16 +316,18 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
...
 
     // Adjust changelog topic setting, when TTL is set on a RocksDB store
     //  - Disable log compaction on Kafka changelog topic
     //  - Set topic TTL to be the same as RocksDB TTL
-    Option(config.get("stores.%s.rocksdb.ttl.ms" format name)) match {
+    val storeTTLkey = "stores.%s.rocksdb.ttl.ms" format name
+    Option(config.get(storeTTLkey)) match {
       case Some(rocksDbTtl) =>
-        if (!config.containsKey("stores.%s.changelog.kafka.cleanup.policy" format name)) {
+        if (config.getInt(storeTTLkey, 0) < 0) {
```
prateekm: Minor: Can use 'rocksDbTtl' here instead of getting it again.

prateekm: Set the max.message.bytes property too (line 338)

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -316,16 +316,18 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
...
+          kafkaChangeLogProperties.setProperty("cleanup.policy", "compact")
+        } else if (!config.containsKey("stores.%s.changelog.kafka.cleanup.policy" format name)) {
```
prateekm: Do you need this containsKey check in the new condition as well?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
prateekm: Here's how this is done in KafkaCheckpointManager:
```
    val checkpointSpec = KafkaStreamSpec.fromSpec(StreamSpec.createCheckpointStreamSpec(checkpointTopic, checkpointSystemName))
        .copyWithReplicationFactor(kafkaConfig.getCheckpointReplicationFactor.get.toInt)
        .copyWithProperties(kafkaConfig.getCheckpointTopicProperties)
```

Can we do that here for consistency?

shanthoosh: 1. Fixed the indentation. 
2. Updated the checkpoint spec building to be consistent with `KafkaCheckpointManager`.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
shanthoosh: Done.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
shanthoosh: Yes.  At the end of this function, the changelog properties if defined for the store in user-config are picked-up and populated in the resultant change-log-topic-properties. So this additional cleanup.policy is still necessary. We could simplify the code, but I prefer not to couple clean-up with a critical-fix.  What do you think?

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -541,8 +541,11 @@ public KafkaStreamSpec toKafkaSpec(StreamSpec spec) {
           new KafkaStreamSpec(spec.getId(), spec.getPhysicalName(), systemName, 1, coordinatorStreamReplicationFactor,
               coordinatorStreamProperties);
     } else if (spec.isCheckpointStream()) {
-      kafkaSpec = KafkaStreamSpec.fromSpec(StreamSpec.createCheckpointStreamSpec(spec.getPhysicalName(), systemName))
-              .copyWithReplicationFactor(Integer.parseInt(new KafkaConfig(config).getCheckpointReplicationFactor().get()));
+      Properties checkpointTopicProperties = new Properties();
+      checkpointTopicProperties.putAll(spec.getConfig());
+      kafkaSpec = KafkaStreamSpec.fromSpec(StreamSpec.createCheckpointStreamSpec(spec.getPhysicalName(), spec.getSystemName()))
+              .copyWithReplicationFactor(Integer.parseInt(new KafkaConfig(config).getCheckpointReplicationFactor().get()))
+              .copyWithProperties(checkpointTopicProperties);
```
prateekm: Why not kafkaConfig.getCheckpointTopicProperties?

shanthoosh: The control flow to create the checkpoint stream is the following:

1. Build the checkpoint spec in `KafkaCheckpointManagerFactory`, which uses kafkaConfig.getCheckpointProperties to populate the config in `CheckpointSpec`.
2. `KafkaCheckpointManager.createStream` uses the checkpoint spec that was built from 1 and `KafkaSystemAdmin.createStream` in-turn invokes `KafkaSystemAdmin.toKafkaSpec`.
3. `KafkaSystemAdmin.toKafkaSpec(CheckpointSpec)` above is used to convert the incoming spec of type `StreamSpec` to `KafkaStreamSpec` .
 

The config in incoming checkpoint `StreamSpec` is already built using   `kafkaConfig.getCheckpointTopicProperties` in `KafkaCheckpointManagerFactory`. So using `kafkaConfig.getCheckpointProperties` here again would be redundant, unnecessary and might return incorrect config-bag.

What do you think?

prateekm: Got it, so the spec.getConfig here is already the checkpoint topic related subset. In that case, LGTM.

kw2542: We can do it separately, can we add .copyWithConfig or .copyWithMap which takes Config or Map<String, String> ?

copyWithProperties is not user friendly, as we are converting a Map/Config to a properties, and this method basically convert it back.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
```diff
@@ -316,16 +316,19 @@ class KafkaConfig(config: Config) extends ScalaMapConfig(config) {
...
-    val appConfig = new ApplicationConfig(config)
     // SAMZA-1600: do not use the combination of "compact,delete" as cleanup policy until we pick up Kafka broker 0.11.0.57,
     // 1.0.2, or 1.1.0 (see KAFKA-6568)
 
     // Adjust changelog topic setting, when TTL is set on a RocksDB store
     //  - Disable log compaction on Kafka changelog topic
     //  - Set topic TTL to be the same as RocksDB TTL
-    Option(config.get("stores.%s.rocksdb.ttl.ms" format name)) match {
+    val storeTTLkey = "stores.%s.rocksdb.ttl.ms" format name
+    Option(config.get(storeTTLkey)) match {
```
kw2542: any specific reason to split this line into two? val storeTTLkey does not seem to be used elsewhere.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
shanthoosh: Yes, adding the `copyWithConfig` to KafkaStreamSpec might be good convenience API. We've additional cleanup planned with this control-flow in the near-future. If acceptable, then I can couple this change with that. Since it's a critical bug-fix, I would prefer to keep the change minimal by containing it to the fix alone and do the clean-up later.

samza-kafka/src/main/scala/org/apache/samza/config/KafkaConfig.scala
shanthoosh: Removed.

PR #1252, opened 2020-01-15
SAMZA-2434: Fix the coordinator steam creation workflow

shanthoosh:
Symptom: Currently the coordinator kafka topic of a samza job is created with cleanup.policy set to 'delete' instead of 'compact' for certain cases. This potentially leads to losing all the metadata(JobModel/Config) stored in coordinator stream for some samza jobs. 

Cause: The control-flow in KafkaSystemAdmin.createStream to build the coordinator stream spec swallows the essential kafka-topic configurations and passes along the empty configuration bag to the kafka-broker when creating the coordinator stream. This issue was introduced in [SAMZA-1868](https://issues.apache.org/jira/browse/SAMZA-1868). 

Changes: Fix the topic-creation control-flow for the coordinator topics and generate the correct topic-configurations.

Tests: Added unit tests to validate that the expected topic configuration bag was generated for the coordinator stream topics.

API Changes: None

Upgrade Instructions: None

Usage Instructions: None


shanthoosh:
@prateekm Can you please take a look at this patch.

PR #1255, opened 2020-01-17
SAMZA-2429: Update LocalApplicatoinRunner to load full job config from config loader when present.

kw2542:
Design:

Changes:
1. Update unit test only public constructor of LocalApplicationRunner to take SamzaApplication and Config as well to be consistent with other constructors.
2. Update private constructor of LocalApplicationRunner to take SamzaApplication and Config.
3. Update private constructor of LocalApplicationRunner to load full job config when config loader is present.

API Changes:
LocalApplicationRunner's constructor now allows initial config/job submission config only instead of full job config, this is backward compatible as full job config is still supported.

Upgrade Instructions:
This is part of a series PRs, detailed information will be provided in the last/main PR.

Usage Instructions:
This is part of a series PRs, detailed information will be provided in the last/main PR.

Tests
Unit tests

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
@@ -133,7 +133,7 @@ public class LocalApplicationRunner implements ApplicationRunner {
     this.isAppModeBatch = isAppModeBatch(config);
     this.coordinationUtils = context.coordinationUtils.isPresent()
         ? context.coordinationUtils
-        : getCoordinationUtils(appDesc.getConfig());
+        : getCoordinationUtils(config);
     this.metadataStoreFactory = context.metadataStoreFactory.isPresent()
         ? context.metadataStoreFactory
         : getDefaultCoordinatorStreamStoreFactory(config);
--- a/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
@@ -62,6 +62,7 @@ import org.apache.samza.system.SystemAdmins;
 import org.apache.samza.system.SystemStream;
 import org.apache.samza.task.TaskFactory;
 import org.apache.samza.task.TaskFactoryUtil;
+import org.apache.samza.util.ConfigUtil;
 import org.apache.samza.util.CoordinatorStreamUtil;
 import org.apache.samza.util.ReflectionUtil;
 import org.apache.samza.zk.ZkJobCoordinatorFactory;
@@ -111,25 +112,33 @@ public class LocalApplicationRunner implements ApplicationRunner {
    * @param metadataStoreFactory the instance of {@link MetadataStoreFactory} to read and write to coordinator stream.
    */
   public LocalApplicationRunner(SamzaApplication app, Config config, MetadataStoreFactory metadataStoreFactory) {
-    this(ApplicationDescriptorUtil.getAppDescriptor(app, config), getCoordinationUtils(config), metadataStoreFactory);
+    this(app, config, getCoordinationUtils(config), metadataStoreFactory);
   }
 
   /**
    * Constructor only used in unit test to allow injection of {@link LocalJobPlanner}
    */
   @VisibleForTesting
-  LocalApplicationRunner(ApplicationDescriptorImpl<? extends ApplicationDescriptor> appDesc, Optional<CoordinationUtils> coordinationUtils) {
-    this(appDesc, coordinationUtils, getDefaultCoordinatorStreamStoreFactory(new JobConfig(appDesc.getConfig())));
+  LocalApplicationRunner(SamzaApplication app, Config config, Optional<CoordinationUtils> coordinationUtils) {
+    this(app, config, coordinationUtils, getDefaultCoordinatorStreamStoreFactory(new JobConfig(config)));
   }
 
   private LocalApplicationRunner(
-      ApplicationDescriptorImpl<? extends ApplicationDescriptor> appDesc,
+      SamzaApplication app,
+      Config config,
       Optional<CoordinationUtils> coordinationUtils,
       MetadataStoreFactory metadataStoreFactory) {
-    this.appDesc = appDesc;
+    this.appDesc = getApplicationDescriptor(app, config);
     this.isAppModeBatch = isAppModeBatch(appDesc.getConfig());
     this.coordinationUtils = coordinationUtils;
     this.metadataStoreFactory = Optional.ofNullable(metadataStoreFactory);
+  }
+
+  @VisibleForTesting
+  static ApplicationDescriptorImpl<? extends ApplicationDescriptor> getApplicationDescriptor(SamzaApplication app, Config config) {
+    return new JobConfig(config).getConfigLoaderFactory().isPresent()
+        ? ApplicationDescriptorUtil.getAppDescriptor(app, ConfigUtil.loadConfig(config))
+        : ApplicationDescriptorUtil.getAppDescriptor(app, config);
   }
 
   @VisibleForTesting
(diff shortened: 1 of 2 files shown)


samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -111,27 +112,35 @@ public LocalApplicationRunner(SamzaApplication app, Config config) {
    * @param metadataStoreFactory the instance of {@link MetadataStoreFactory} to read and write to coordinator stream.
    */
   public LocalApplicationRunner(SamzaApplication app, Config config, MetadataStoreFactory metadataStoreFactory) {
-    this(ApplicationDescriptorUtil.getAppDescriptor(app, config), getCoordinationUtils(config), metadataStoreFactory);
+    this(app, config, getCoordinationUtils(config), metadataStoreFactory);
```
cameronlee314: In the old flow, `getCoordinationUtils` seemed to be called with the full config. In the new flow, `getCoordinationUtils` will get called with the initial config, not the full config. Do you need to load the config before calling `getCoordinationUtils`?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -111,27 +112,35 @@ public LocalApplicationRunner(SamzaApplication app, Config config) {
...
   }
 
   /**
    * Constructor only used in unit test to allow injection of {@link LocalJobPlanner}
    */
   @VisibleForTesting
-  LocalApplicationRunner(ApplicationDescriptorImpl<? extends ApplicationDescriptor> appDesc, Optional<CoordinationUtils> coordinationUtils) {
-    this(appDesc, coordinationUtils, getDefaultCoordinatorStreamStoreFactory(new JobConfig(appDesc.getConfig())));
+  LocalApplicationRunner(SamzaApplication app, Config config, Optional<CoordinationUtils> coordinationUtils) {
+    this(app, config, coordinationUtils, getDefaultCoordinatorStreamStoreFactory(new JobConfig(config)));
```
cameronlee314: I guess this constructor is just for testing, but it does have a similar change as above where `config` is now just the initial config, but before it was the full config.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -111,27 +112,35 @@ public LocalApplicationRunner(SamzaApplication app, Config config) {
...
   }
 
   private LocalApplicationRunner(
```
cameronlee314: Is this considered a semantic API change? Before, it seems like `config` was always the full config, but now you are allowing the initial config with only config loader properties.
It's ok to have this be an API change, but just wanted to make sure you note that where necessary.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
kw2542: Very good question, and nice catch. You are right, both getCoordinationUtils as well as getDefaultCoordinatorStreamStoreFactory need to be updated to load full config first if applicable.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
kw2542: Updated.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
kw2542: Good question, I would consider changes in samza-api module to be public facing API changes, and changes in samza-core to be internal API changes, which can be less risky. 

I will update the PR description to explicitly call out the constructor changes, also this is a backward compatible API change.

cameronlee314: `LocalApplicationRunner` is kind of in that middle layer: it's not in samza-api and not quite a direct API, but it also can get used by users running standalone and it can be specified in configs as the "app.runner.class".

cameronlee314: In my opinion, `LocalApplicationRunner` doesn't truly belong in samza-core, but it had to be somewhere...

kw2542: Got you, in addition to note this change in PR's description, anywhere else we should notify this change?

cameronlee314: Someone will need to remember to add this as a release note, and if we have information in the user guide about this flow, then that will need to be updated.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -111,29 +112,41 @@ public LocalApplicationRunner(SamzaApplication app, Config config) {
...
+    return new JobConfig(config).getConfigLoaderFactory().isPresent()
+        ? ApplicationDescriptorUtil.getAppDescriptor(app, ConfigUtil.loadConfig(config))
+        : ApplicationDescriptorUtil.getAppDescriptor(app, config);
+  }
+
   @VisibleForTesting
   static MetadataStoreFactory getDefaultCoordinatorStreamStoreFactory(JobConfig jobConfig) {
+    if (jobConfig.getConfigLoaderFactory().isPresent()) {
+      jobConfig = new JobConfig(ConfigUtil.loadConfig(jobConfig));
+    }
```
cameronlee314: Should we avoid multiple calls to `ConfigUtil.loadConfig`? You already call it in `getApplicationDescriptor`.
If the config loader logic is complex/heavy (e.g. makes remote calls), then that is going to result in inefficient calls.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
```diff
@@ -150,6 +163,10 @@ static MetadataStoreFactory getDefaultCoordinatorStreamStoreFactory(JobConfig jo
   }
 
   private static Optional<CoordinationUtils> getCoordinationUtils(Config config) {
+    if (new JobConfig(config).getConfigLoaderFactory().isPresent()) {
+      config = ConfigUtil.loadConfig(config);
+    }
```
cameronlee314: Same as above about multiple calls to `loadConfig`.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
kw2542: Ideally yes, in order to achieve this, each constructor needs to have some if checks, thus it cannot invoke another constructor anymore. This will return in all constructors follows its own logic instead of merging into a single private constructor. Do you think this is a good trade off?

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
kw2542: I will make sure this to be added in the OSS release when it is out and I will go through our user guide to see if we have any thing about this flow.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
cameronlee314: I do think it is reasonable trade-off.
One of the challenges is that the `LocalApplicationRunner(SamzaApplication app, Config config)` constructor is overloaded semantically. In one case, `config` is already the full config, and in another case `config` is a partial config which needs to be loaded. And then it may seem like we can't use a single common constructor due to the need to use the full config in the constructor arguments. I do still think there is a way to share a common constructor: instead of passing individual arguments, create a `LocalApplicationRunnerContext` which contains all of the arguments, and then you can build that other object based on what is given to you.

kw2542: I like the idea of LocalApplicationRunnerContext, will update the PR.

samza-core/src/main/java/org/apache/samza/runtime/LocalApplicationRunner.java
kw2542: Updated

PR #1256, opened 2020-01-22
SAMZA-2420: Update CommandLine to use config loader for local config file

kw2542:
This is supposed to be included only in Samza 1.4

Design:

Changes:
1. Update CommandLine to use config loader instead of config factory
2. Removed common.properties and adds its values to each job config
3. Update CheckpointTool to read new offsets from local file only.

API Changes:
1. Add config-loader-factory and config-loader-properties in CommandLine to support specifying ConfigLoaderFactory and its properties needed to load config.
2. Remove config-factory and config-path in CommandLine to discountinue the usage of ConfigFactory
3. Update CheckpointTool to read new offsets from local file only, i.e. --new-offsets only supports local file URI now.

Upgrade Instructions:

All usages in CommandLine and its subclasses will switch from --config-factory & --config-path to --config-loader-factory & --config-loader-properties, including job launch.

Usage Instructions

For example,

--config-factory=org.apache.samza.config.factories.PropertiesConfigFactory --config-path=file:///location/file.txt

will be changed to

--config-loader-factory=org.apache.samza.config.loaders.PropertiesConfigLoaderFactory --config-loader-properties path=file:///location/file.txt

Tests

1. Unit tests
2. Test against Hello Samza example with

  deploy/samza/bin/run-app.sh --config-loader-factory=org.apache.samza.config.loaders.PropertiesConfigLoaderFactory --config-loader-properties path=$PWD/deploy/samza/config/wikipedia-feed.properties


samza-core/src/main/java/org/apache/samza/storage/StorageRecovery.java
```diff
@@ -63,11 +62,11 @@
 
 
 /**
- * Recovers the state storages from the changelog streams and store the storages
+ * Recovers the state storage from the changelog streams and store the storages
```
mynameborat: nit: might as well fix the entire doc. "Recovers the state store from the changelog stream and stores in the directory provided...."

samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
```diff
@@ -117,10 +118,15 @@ object CheckpointTool {
         .mapValues(m => m.reduce( _ ++ _))  // Merge all the maps of SSPs->Offset into one for the whole taskname
     }
 
-    override def loadConfig(options: OptionSet): MapConfig = {
+    override def loadConfig(options: OptionSet): Config = {
       val config = super.loadConfig(options)
       if (options.has(newOffsetsOpt)) {
-        val properties = configFactory.getConfig(options.valueOf(newOffsetsOpt))
+        val in = new FileInputStream(options.valueOf(newOffsetsOpt).getPath)
```
mynameborat: nit: rename `in` to something meaningful?

samza-test/src/main/config/join/joiner.samza
```diff
@@ -16,6 +16,33 @@
 # specific language governing permissions and limitations
 # under the License.
 
+####################
+### UPDATE THIS! ###
```
mynameborat: Can you add more details here? Not sure what update this refers to?

kw2542: These unit tests are reading two files, common.properties and joiner.samza, as the input config, after updating to use ConfigLoader, we do not want to read from multiple files anymore, so I am deleting common.properties file and move all its properties in each job config file instead.

samza-core/src/main/scala/org/apache/samza/checkpoint/CheckpointTool.scala
kw2542: Renamed to newOffsetsInputStream

samza-core/src/main/java/org/apache/samza/storage/StorageRecovery.java
```diff
@@ -63,11 +62,11 @@
 
 
 /**
- * Recovers the state storages from the changelog streams and store the storages
+ * Recovers the state storage from both the changelog streams and stores
```
cameronlee314: Can you please clarify this documentation change? The usage of the term "both" is unclear to me.
Is it supposed to be "Recovers the state storage from the changelog streams and stores the state in the directory provided by the users."?

samza-core/src/main/java/org/apache/samza/storage/StorageRecovery.java
```diff
@@ -158,6 +155,7 @@ private void getContainerModels() {
    * Get the changelog streams and the storage factories from the config file
    * and put them into the maps
    */
+  @SuppressWarnings("unchecked")
```
cameronlee314: Instead of doing a blanket suppress for the whole method, could you please consider doing a suppression for the single line that needs to be suppressed? Then, if someone accidentally makes a change which should not be suppressed, then there will still be a warning for it.

samza-core/src/main/java/org/apache/samza/storage/StorageRecovery.java
```diff
@@ -195,6 +191,7 @@ private void getChangeLogMaxPartitionNumber() {
     maxPartitionNumber = maxPartitionId + 1;
   }
 
+  @SuppressWarnings("unchecked")
```
cameronlee314: Same as above about suppression

samza-core/src/main/scala/org/apache/samza/util/CommandLine.scala
```diff
@@ -19,55 +19,60 @@
...
 /**
  * Defines a basic set of command-line options for Samza tasks. Tools can use this
  * class directly, or subclass it to add their own options.
  */
 class CommandLine {
   val parser = new OptionParser()
-  val configFactoryOpt =
-    parser.accepts("config-factory", "The config factory to use to read your config file.")
+  val configLoaderFactoryOpt: ArgumentAcceptingOptionSpec[String] =
+    parser.accepts("config-loader-factory", "The config factory to use to read your config file.")
```
cameronlee314: The "help" string seems out-of-date now.

samza-core/src/main/scala/org/apache/samza/util/CommandLine.scala
```diff
@@ -19,55 +19,60 @@
...
-  val configPathOpt =
-    parser.accepts("config-path", "URI location to a config file (e.g. file:///some/local/path.properties). " +
+          .defaultsTo(classOf[PropertiesConfigLoaderFactory].getName)
+  val configLoaderPropertiesOpt: ArgumentAcceptingOptionSpec[KeyValuePair] =
+    parser.accepts("config-loader-properties", "URI location to a config file (e.g. file:///some/local/path.properties). " +
                                   "If multiple files are given they are all used with later files overriding any values that appear in earlier files.")
           .withRequiredArg
-          .ofType(classOf[URI])
+          .ofType(classOf[KeyValuePair])
           .describedAs("path")
```
cameronlee314: This seems out-of-date.

samza-core/src/main/scala/org/apache/samza/util/CommandLine.scala
```diff
@@ -19,55 +19,60 @@
...
           .describedAs("key=value")
 
-  var configFactory: ConfigFactory = null
+  var configLoaderFactory: ConfigLoaderFactory = _
 
-  def loadConfig(options: OptionSet) = {
+  def loadConfig(options: OptionSet): Config = {
     // Verify legitimate parameters.
-    if (!options.has(configPathOpt)) {
+    if (!options.has(configLoaderPropertiesOpt)) {
```
cameronlee314: Should this check for `configLoaderFactoryOpt` instead? Theoretically, the config loader factory won't need additional properties. You might also need to somehow validate or catch that not all config loader properties were specified for the config loader.

samza-test/src/main/java/org/apache/samza/test/integration/LocalApplicationRunnerMain.java
```diff
@@ -53,14 +53,14 @@ public static void main(String[] args) throws Exception {
...
-      runner.run(buildExternalContext(config).orElse(null));
+      runner.run(buildExternalContext().orElse(null));
       runner.waitForFinish();
     } catch (Exception e) {
       LOGGER.error("Exception occurred when running application: {}.", app, e);
     }
   }
 
-  private static Optional<ExternalContext> buildExternalContext(Config config) {
+  private static Optional<ExternalContext> buildExternalContext() {
```
cameronlee314: The intention to have the "unused" `config` was so that a custom fork could use the `config` if necessary (see comment inside `buildExternalContext`. Could we please keep the `config`?

samza-core/src/main/scala/org/apache/samza/util/CommandLine.scala
```diff
@@ -19,55 +19,60 @@
...
+    parser.accepts("config-loader-factory", "The config factory to use to read your config file.")
           .withRequiredArg
           .ofType(classOf[java.lang.String])
           .describedAs("com.foo.bar.ClassName")
-          .defaultsTo(classOf[PropertiesConfigFactory].getName)
-  val configPathOpt =
-    parser.accepts("config-path", "URI location to a config file (e.g. file:///some/local/path.properties). " +
+          .defaultsTo(classOf[PropertiesConfigLoaderFactory].getName)
+  val configLoaderPropertiesOpt: ArgumentAcceptingOptionSpec[KeyValuePair] =
+    parser.accepts("config-loader-properties", "URI location to a config file (e.g. file:///some/local/path.properties). " +
```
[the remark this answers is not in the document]
kw2542: Updated.

samza-core/src/main/scala/org/apache/samza/util/CommandLine.scala
kw2542: Updated.

samza-core/src/main/scala/org/apache/samza/util/CommandLine.scala
kw2542: Updated

samza-core/src/main/scala/org/apache/samza/util/CommandLine.scala
kw2542: configLoaderFactoryOpt has a default value of PropertiesConfigLoaderFactory.

This check is to be consistent with the previous check where we require a config path (URI) to pass to the legacy ConfigFactory, which in theory may not need URI either.

We may remove this check completely too, if we want to support the case where config loader factory does not need any properties at all.

samza-test/src/main/java/org/apache/samza/test/integration/LocalApplicationRunnerMain.java
kw2542: Added back.

samza-core/src/main/java/org/apache/samza/storage/StorageRecovery.java
kw2542: I misunderstood the previous document, updated.

samza-core/src/main/java/org/apache/samza/storage/StorageRecovery.java
kw2542: Updated to suppress the single line.

samza-core/src/main/java/org/apache/samza/storage/StorageRecovery.java
kw2542: Updated to suppress the single line.

samza-core/src/main/scala/org/apache/samza/util/CommandLine.scala
cameronlee314: In the previous case (i.e. `ConfigFactory`), the `configURI` was a required argument for the interface, and I don't think you can have an empty `URI`. For `ConfigLoaderFactory` interface, it is required to pass a `Config`, but it is possible to have an empty `Config`, so I think you could just remove the check to handle the general case.
The additional complexity that the `ConfigLoaderFactory` brings is that each concrete type requires different configs to be in `Config`, so if there is an error due to a missing config value, then that's maybe where you need to do the `parser.printHelpOn(System.err); System.exit(-1);`.

kw2542: Make sense, removed the check.

The default concrete `PropertiesConfigLoaderFactory` is checking the properties itself too.

PR #1258, opened 2020-01-23
SAMZA-2275: Azure Blob System Producer: fixes for high level jobs

lakshmi-manasa-g:
**Symptom**: 
1. Samza High level job using this SystemProducer gets UnsupportedOperationException from AzureBlobSystemAdmin.
2. Azure account details of the samza job are logged. 

**Cause**:
1. getSystemStreamMetadata throws unsupported exception. 
2. Account details configs are not marked sensitive.

**Changes**
1. getSystemStreamMetadata returns an empty map.
2. "sensitive" prefixed to the account details configs.

**Tests**: Sample High level job in hello-samza (https://github.com/apache/samza-hello-samza/pull/71) works correctly with these changes.

### DIFF
--- a/samza-azure/src/main/java/org/apache/samza/system/azureblob/AzureBlobConfig.java
+++ b/samza-azure/src/main/java/org/apache/samza/system/azureblob/AzureBlobConfig.java
@@ -40,10 +40,10 @@ public class AzureBlobConfig extends MapConfig {
 
   // Azure Storage Account name under which the Azure container representing this system is.
   // System name = Azure container name (https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names)
-  public static final String SYSTEM_AZURE_ACCOUNT_NAME = SYSTEM_AZUREBLOB_PREFIX + "account.name";
+  public static final String SYSTEM_AZURE_ACCOUNT_NAME = Config.SENSITIVE_PREFIX + SYSTEM_AZUREBLOB_PREFIX + "account.name";
 
   // Azure Storage Account key associated with the Azure Storage Account
-  public static final String SYSTEM_AZURE_ACCOUNT_KEY  = SYSTEM_AZUREBLOB_PREFIX + "account.key";
+  public static final String SYSTEM_AZURE_ACCOUNT_KEY  = Config.SENSITIVE_PREFIX + SYSTEM_AZUREBLOB_PREFIX + "account.key";
 
   // Whether to use proxy while connecting to Azure Storage
   public static final String SYSTEM_AZURE_USE_PROXY  = SYSTEM_AZUREBLOB_PREFIX + "proxy.use";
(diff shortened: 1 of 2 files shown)


lakshmi-manasa-g:
> How does `getSystemStreamMetadata` get used? Based on your description, it sounds like `getSystemStreamMetadata` gets called, so does that mean something needs to use the results from it? If you return an empty map, could that be misleading to the logic that uses `getSystemStreamMetadata`?

**The exact stack trace of this method being called for the high-level sample job**
`at org.apache.samza.system.azureblob.AzureBlobSystemAdmin.getSystemStreamMetadata(AzureBlobSystemAdmin.java:44)
at org.apache.samza.execution.StreamManager.getStreamPartitionCounts(StreamManager.java:93)
at org.apache.samza.execution.ExecutionPlanner.setInputAndOutputStreamPartitionCount(ExecutionPlanner.java:197)
 at org.apache.samza.execution.ExecutionPlanner.plan(ExecutionPlanner.java:86)
at org.apache.samza.execution.JobPlanner.getExecutionPlan(JobPlanner.java:101)
at org.apache.samza.execution.RemoteJobPlanner.prepareJobs(RemoteJobPlanner.java:57)
at org.apache.samza.runtime.RemoteApplicationRunner.run(RemoteApplicationRunner.java:73)
at org.apache.samza.runtime.ApplicationRunnerUtil.invoke(ApplicationRunnerUtil.java:49)
at org.apache.samza.runtime.ApplicationRunnerMain.main(ApplicationRunnerMain.java:53)`

**Usage of `getSystemStreamMetadata` in this stack trace:**
In the stack trace above, this method is used to get the number of partitions (=size of the map) of a stream in the job graph (all input and output streams). The partition counts are added to the job graph and then used to validate joins and calculate partitions for intermediate streams. Hence, returning an empty map for AzureBlobSystem will result in the partition count being 0 which is does not lead to any incorrect behavior.

**Other usages (directly or transitively) of `getSystemStreamMetadata`:**

1. Direct uses: CoordinatorStreamStore, CoordinatorStreamSystemConsumer, StreamManager.getPartitionCounts, StreamMetadataCache, KafkaSystemAdmin, KafkaCheckpointManager, TestRunner, SytemConsumerBench - of these only the StreamManager is relevant cause Azure Blob can not be as CoordinatorStream (as it has no consumer).
StreamManager usage is described above.

2. Transitive uses (via SystemAdmin.getSSPMetadata and SystemAdmin.getSystemStreamPartitionCounts) : NonTransactionalStateTaskStorageManager.getNewestChangelogSSPOffsets, TransactionalStateTaskStorageManager.getNewestChangelogSSPOffsets and SSPMetadataCache. None of these are relevant for Azure Blob.

**Azure blob Storage as a system in samza:**
1. There is no concept of a partition in Azure Blob Storage - it is mimicked through virtual folders (basically “/“) in the blob name. 
2. Samza's SystemStreamMetadata consists of StreamName and 3 offsets (oldest, newest and upcoming) which are not relevant for an Azure blob.

For all the above stated reasons, I believe it is acceptable to return an empty map for this method in AzureBlobSystemAdmin.

cameronlee314:
Thanks for the detailed explanation. Can you please summarize your response into a comment for `AzureBlobSystemAdmin.getSystemStreamMetadata`?

PR #1125, opened 2019-08-01
SAMZA-2286: Integrate MetadataStore#putAll for improved startup time.

dnishimura:
This PR integrates `MetadataStore#putAll` with `TaskAssignmentManager`, `TaskPartitionAssignmentManager`, `ChangelogStreamManager` and `ZkJobCoordinator` when writing assignments to the coordinator stream. For a Kafka-based coordinator stream, the `putAll` method allows multiple messages to be produced before calling flush. We have observed in preliminary benchmark tests that this provides an order of magnitude improvement when calling `putAll` vs. `put` for each entry. The impact is faster startup times.

@shanthoosh - please take a look since you have the most context.

PR #1259, opened 2020-01-23
SAMZA-2444: JobModel save in CoordinatorStreamStore resulting flush for each message

alnzng:
## Symptom

When Samza's job creates lots of tasks/partitions, it can take over a long time for AM to save the metadata in a run which may cause timeout exception.
We observed if the Samza's job has over 37k tasks/partitions, it takes over 10 min for AM to save it in a run.

## Cause

JobModelManager uses CoordinatorStreamStore.put() to save job metadata information which does flush for each message, and the flush operation is heavy especially when the remote server suffering the performance issues. 

## Changes

- [x] Separate `flush` from `put/putAll/delete` functions in `CoordinatorStreamStore`.
- [x] Explicity call `flush` after call `put/putAll/delete` in related classes
- [x] Batch write task partition assignments information to metadata store.
- [x] Batch write task container information to metadata store.

## Tests

## API Changes

- [x] Replace `writeTaskPartitionAssignment` with new batch write method `writeTaskPartitionAssignments` in `TaskPartitionAssignmentManager`
- [x] Replace `writeTaskContainerMapping` with  new batch write method `writeTaskContainerMappings` in `TaskAssignmentManager`.

## Upgrade Instructions
None

## Usage Instructions
None

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
@@ -96,28 +96,32 @@ public class TaskAssignmentManager {
   }
 
   /**
-   * Method to write task container info to {@link MetadataStore}.
-   *
-   * @param taskName    the task name
-   * @param containerId the SamzaContainer ID or {@code null} to delete the mapping
-   * @param taskMode the mode of the task
+   * Method to batch write task container info to {@link MetadataStore}.
+   * @param mappings the task and container mappings: (ContainerId, (TaskName, TaskMode))
    */
-  public void writeTaskContainerMapping(String taskName, String containerId, TaskMode taskMode) {
-    String existingContainerId = taskNameToContainerId.get(taskName);
-    if (existingContainerId != null && !existingContainerId.equals(containerId)) {
-      LOG.info("Task \"{}\" in mode {} moved from container {} to container {}", new Object[]{taskName, taskMode, existingContainerId, containerId});
-    } else {
-      LOG.debug("Task \"{}\" in mode {} assigned to container {}", taskName, taskMode, containerId);
-    }
+  public void writeTaskContainerMappings(Map<String, Map<String, TaskMode>> mappings) {
+    for (String containerId : mappings.keySet()) {
+      Map<String, TaskMode> tasks = mappings.get(containerId);
+      for (String taskName : tasks.keySet()) {
+        TaskMode taskMode = tasks.get(taskName);
+        LOG.info("Storing task: {} and container ID: {} into metadata store", taskName, containerId);
+        String existingContainerId = taskNameToContainerId.get(taskName);
+        if (existingContainerId != null && !existingContainerId.equals(containerId)) {
+          LOG.info("Task \"{}\" in mode {} moved from container {} to container {}", new Object[]{taskName, taskMode, existingContainerId, containerId});
+        } else {
+          LOG.debug("Task \"{}\" in mode {} assigned to container {}", taskName, taskMode, containerId);
+        }
 
-    if (containerId == null) {
-      taskContainerMappingMetadataStore.delete(taskName);
-      taskModeMappingMetadataStore.delete(taskName);
-      taskNameToContainerId.remove(taskName);
-    } else {
-      taskContainerMappingMetadataStore.put(taskName, containerIdSerde.toBytes(containerId));
-      taskModeMappingMetadataStore.put(taskName, taskModeSerde.toBytes(taskMode.toString()));
-      taskNameToContainerId.put(taskName, containerId);
+        if (containerId == null) {
+          taskContainerMappingMetadataStore.delete(taskName);
+          taskModeMappingMetadataStore.delete(taskName);
+          taskNameToContainerId.remove(taskName);
+        } else {
+          taskContainerMappingMetadataStore.put(taskName, containerIdSerde.toBytes(containerId));
+          taskModeMappingMetadataStore.put(taskName, taskModeSerde.toBytes(taskMode.toString()));
+          taskNameToContainerId.put(taskName, containerId);
+        }
+      }
     }
     taskContainerMappingMetadataStore.flush();
     taskModeMappingMetadataStore.flush();
--- a/samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
+++ b/samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
@@ -119,6 +119,8 @@ public class TaskAssignmentManager {
       taskModeMappingMetadataStore.put(taskName, taskModeSerde.toBytes(taskMode.toString()));
       taskNameToContainerId.put(taskName, containerId);
     }
+    taskContainerMappingMetadataStore.flush();
+    taskModeMappingMetadataStore.flush();
   }
 
   /**
@@ -132,6 +134,8 @@ public class TaskAssignmentManager {
       taskModeMappingMetadataStore.delete(taskName);
       taskNameToContainerId.remove(taskName);
     }
+    taskContainerMappingMetadataStore.flush();
+    taskModeMappingMetadataStore.flush();
   }
 
   public void close() {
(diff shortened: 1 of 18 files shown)


prateekm:
@bharathkk and @lakshmi-manasa-g can you take a look?

xinyuiscool:
Now I looked at the coodinator stream impl for metadatstore, it seems we shouldn't couple flush with put/putall/delete/deleteall. It should be done as a separate call at the end of updates. The current impl is hideous and can cause further perf problems down the road.

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskAssignmentManager.java
```diff
@@ -103,21 +106,53 @@ public TaskAssignmentManager(MetadataStore taskContainerMappingMetadataStore, Me
    * @param taskMode the mode of the task
    */
   public void writeTaskContainerMapping(String taskName, String containerId, TaskMode taskMode) {
-    String existingContainerId = taskNameToContainerId.get(taskName);
-    if (existingContainerId != null && !existingContainerId.equals(containerId)) {
-      LOG.info("Task \"{}\" in mode {} moved from container {} to container {}", new Object[]{taskName, taskMode, existingContainerId, containerId});
-    } else {
-      LOG.debug("Task \"{}\" in mode {} assigned to container {}", taskName, taskMode, containerId);
-    }
+    writeTaskContainerMappings(ImmutableMap.of(containerId, ImmutableMap.of(taskName, taskMode)));
```
lakshmi-manasa-g: would it be good to replace `wirteTaskContainerMapping` with the new `wirteTaskContainerMappings` altogether?
there seems to be only one usage of this method and that has been replaced to use the new one in JobModelManager. 
is there a reason to keep this old method?

alnzng:
> Now I looked at the coodinator stream impl for metadatstore, it seems we shouldn't couple flush with put/putall/delete/deleteall. It should be done as a separate call at the end of updates. The current impl is hideous and can cause further perf problems down the road.

@xinyuiscool Do you mean we should remove `flush` calls from `put/putall/delete/deleteall`? And let the callers of those functions decide when to do `flush` operation?

xinyuiscool:
Yes, I think that's a typical store api. Not sure why it was doing all the flushes before. @bharathkk  might have some context about it, but he is on vacation.

alnzng:
@lakshmi-manasa-g Thanks for the review. Had one discussion with @xinyuiscool and @prateekm , we decided to change with another solution which will move the `flush` out of all update methods in `CoordinatorStreamStore`.  So above codes with your comments will be removed. I will let you know when the new changes are done.

prateekm:
Alan, this PR seems to be solving a similar problem. https://github.com/apache/samza/pull/1125

Can you check how the problem / solution in that PR relates to this one?

alnzng:
> Alan, this PR seems to be solving a similar problem. #1125
> 
> Can you check how the problem / solution in that PR relates to this one?

@prateekm 
Just checked the PR: #1125. I believe it is solving exactly the same performance issue with this one.
 But it is doing with batch updates way, and current PR is doing the way to decouple `flush` and update methods.

If we solve the issue with this PR, later we can close it.

alnzng:
The changes are done and updated the details in the PR's description, please help review, thanks!
 @xinyuiscool @prateekm @lakshmi-manasa-g

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskPartitionAssignmentManager.java
```diff
@@ -66,28 +66,33 @@ public TaskPartitionAssignmentManager(MetadataStore metadataStore) {
...
+      LOG.info("Storing ssp: {} and task: {} into metadata store", partition, taskNames);
+      // For broadcast streams, a input system stream partition will be mapped to more than one tasks in a
+      // SamzaContainer. Rather than storing taskName to list of SystemStreamPartitions in metadata store, here
+      // systemStreamPartition to list of taskNames is stored. This was done due to 1 MB limit on value size in kafka.
+      String serializedSSPAsJson = serializeSSPToJson(partition);
+      if (taskNames == null || taskNames.isEmpty()) {
+        LOG.info("Deleting the key: {} from the metadata store.", partition);
+        metadataStore.delete(serializedSSPAsJson);
+      } else {
+        try {
```
lakshmi-manasa-g: why do we catch only exceptions from `metadatstore.put` and not from its `delete` or `flush`?

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStore.java
```diff
@@ -149,6 +136,13 @@ private void putWithoutFlush(String namespacedKey, byte[] value) {
     systemProducer.send(SOURCE, envelope);
   }
 
+  @Override
+  public void putAll(Map<String, byte[]> entries) {
```
lakshmi-manasa-g: do we need this impl of `putAll` because `MetadataStore.putAll` default impl in the interface `MetadatStore` is the same?

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskPartitionAssignmentManager.java
```diff
@@ -66,28 +66,33 @@ public TaskPartitionAssignmentManager(MetadataStore metadataStore) {
...
-        String taskNamesAsString = taskNamesMapper.writeValueAsString(taskNames);
-        byte[] taskNamesAsBytes = valueSerde.toBytes(taskNamesAsString);
-        LOG.info("Storing the partition: {} and taskNames: {} into the metadata store.", serializedSSPAsJson, taskNames);
-        metadataStore.put(serializedSSPAsJson, taskNamesAsBytes);
-      } catch (Exception e) {
-        throw new SamzaException("Exception occurred when writing task to partition assignment.", e);
+  public void writeTaskPartitionAssignments(Map<SystemStreamPartition, List<String>> sspToTaskNameMapping) {
+    for (SystemStreamPartition partition: sspToTaskNameMapping.keySet()) {
+      List<String> taskNames = sspToTaskNameMapping.get(partition);
+      LOG.info("Storing ssp: {} and task: {} into metadata store", partition, taskNames);
```
lakshmi-manasa-g: minor: is this log statement really needed? Because it will be followed by another log statement immediately that says either the `key is being deleted` or `partition is being stored` (only that the SSP is a JSON in the `storing` log statement).

samza-core/src/main/java/org/apache/samza/coordinator/metadatastore/CoordinatorStreamStore.java
alnzng: Good catch @lakshmi-manasa-g . This function is duplicated with the one in `MetadataStore`, I will delete it.

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskPartitionAssignmentManager.java
alnzng: In order to not break anything here, I just move the whole code block to the new function. 
The statement `String taskNamesAsString = taskNamesMapper.writeValueAsString(taskNames)` throws exceptions explicitly, so exception need to be catched explicitly. I think original idea is not for `metadatstore.put` method.

lakshmi-manasa-g: ah, got it. thanks!

samza-core/src/main/java/org/apache/samza/container/grouper/task/TaskPartitionAssignmentManager.java
alnzng: Hmm, I think you are right. Let me remove this log line. Thanks.

mynameborat:
> Yes, I think that's a typical store api. Not sure why it was doing all the flushes before. @bharathkk might have some context about it, but he is on vacation.

@alnzng here is the PR for context -  https://github.com/apache/samza/pull/1112
+1 to splitting flush and the API. Refer to the comments in the above PR regarding separating out flush and put APIs

alnzng:
@xinyuiscool @prateekm 
Can any of you help check and merge the PR if no new comments? Thanks.

PR #1260, opened 2020-01-27
SAMZA-2446: Invoke onCheckpoint only for registered SSPs

bkonold:
**Symptom:** Since introducing changelog SSP checkpointing in 1.3, onCheckpoint will be invoked for changelog SSPs if job.changelog.system is configured to be the same as an input system which implements CheckpointListener. This can be problematic because the types of checkpoints being written for changelog SSPs and input SSPs are different.
 
**Cause:** OffsetManager.writeCheckpoint is called with a Checkpoint parameter that includes changelog SSP checkpoints.
 
**Changes:** OffsetManager will now only invoke onCheckpoint for SSPs that are registered with it.
 
**Tests:** Wrote a test which attempts to write a Checkpoint with OffsetManager.writeCheckpoint which includes an SSP which was not registered with the OffsetManager. The test verifies that the callback is invoked only for those SSPs that were registered via OffsetManager.register. This test fails without the patch and passes with it.

**API Changes:** None
 
**Usage Instructions:** None

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
+++ b/samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
@@ -353,13 +353,16 @@ class OffsetManager(
         }
       }
 
-      // invoke checkpoint listeners
-      checkpoint.getOffsets.asScala.groupBy { case (ssp, _) => ssp.getSystem }.foreach {
-        case (systemName:String, offsets: Map[SystemStreamPartition, String]) => {
-          // Option is empty if there is no checkpointListener for this systemName
-          checkpointListeners.get(systemName).foreach(_.onCheckpoint(offsets.asJava))
+      // invoke checkpoint listeners only for SSPs that are registered with the OffsetManager
+      val registeredSSPs = systemStreamPartitions.getOrElse(taskName, immutable.Set[SystemStreamPartition]())
+      checkpoint.getOffsets.asScala
+        .filterKeys(registeredSSPs.contains)
+        .groupBy { case (ssp, _) => ssp.getSystem }.foreach {
+          case (systemName:String, offsets: Map[SystemStreamPartition, String]) => {
+            // Option is empty if there is no checkpointListener for this systemName
+            checkpointListeners.get(systemName).foreach(_.onCheckpoint(offsets.asJava))
+          }
         }
-      }
     }
 
     // delete corresponding startpoints after checkpoint is supposed to be committed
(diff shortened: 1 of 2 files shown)


samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -353,13 +353,16 @@ class OffsetManager(
         }
       }
 
-      // invoke checkpoint listeners
-      checkpoint.getOffsets.asScala.groupBy { case (ssp, _) => ssp.getSystem }.foreach {
-        case (systemName:String, offsets: Map[SystemStreamPartition, String]) => {
-          // Option is empty if there is no checkpointListener for this systemName
-          checkpointListeners.get(systemName).foreach(_.onCheckpoint(offsets.asJava))
+      // invoke checkpoint listeners only for SSPs that are registered with the OffsetManager
+      val registeredSSPs = systemStreamPartitions.getOrElse(taskName, immutable.Set[SystemStreamPartition]())
```
prateekm: Minor: Can we do the following?
`val registeredSSPs = systemStreamPartitions.getOrElse(taskName, Set())`

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
```diff
@@ -353,13 +353,16 @@ class OffsetManager(
         }
       }
 
-      // invoke checkpoint listeners
-      checkpoint.getOffsets.asScala.groupBy { case (ssp, _) => ssp.getSystem }.foreach {
-        case (systemName:String, offsets: Map[SystemStreamPartition, String]) => {
-          // Option is empty if there is no checkpointListener for this systemName
-          checkpointListeners.get(systemName).foreach(_.onCheckpoint(offsets.asJava))
+      // invoke checkpoint listeners only for SSPs that are registered with the OffsetManager
```
prateekm: Let's add a comment explaining when this will not be the case.

prateekm:
@bkonold Can you check why the travis build failed?

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
bkonold: We need to indicate the type somewhere or the compiler throws up. E.g.
`val registeredSSPs = systemStreamPartitions.getOrElse(taskName, Set[SystemStreamPartition]()`

samza-core/src/main/scala/org/apache/samza/checkpoint/OffsetManager.scala
bkonold: Good point, I'll add an example.

bkonold:
> @bkonold Can you check why the travis build failed?

Looks like it was a transient failure, something azure related.

PR #1261, opened 2020-01-28
SAMZA-2447: Checkpoint dir removal should only search in valid store dirs

bkonold:
**Symptom:** NPE when attempting to remove old checkpoint directories.
 
**Cause:** Removal of old checkpoints assumes every file in the logged store base directory is also a directory, which may not be true if the user has configured it to a path that is not exclusive to store contents.
 
**Changes:** Check that the result of `listFiles` is non-null before attempting to delete.
 
**Tests:** Wrote a test to mimic the behavior of a non-directory file in the logged store base directory. Test fails without the patch, passes with it.

**API Changes:** None
 
**Usage Instructions:** None

### DIFF
--- a/samza-core/src/test/scala/org/apache/samza/storage/TestTransactionalStateTaskStorageManager.java
+++ b/samza-core/src/test/scala/org/apache/samza/storage/TestTransactionalStateTaskStorageManager.java
@@ -21,6 +21,8 @@ package org.apache.samza.storage;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import java.io.FileFilter;
+import org.apache.commons.io.filefilter.WildcardFileFilter;
 import scala.Option;
 import scala.collection.immutable.Map;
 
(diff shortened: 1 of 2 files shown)


samza-core/src/test/scala/org/apache/samza/storage/TestTransactionalStateTaskStorageManager.java
```diff
@@ -492,6 +493,32 @@ public void testWriteChangelogOffsetFilesWithNoChangelogOffset() throws IOExcept
     fail("Should have thrown an exception if no changelog offset found for checkpointed store");
   }
 
+  @Test
+  public void testRemoveOldCheckpoints() {
```
prateekm: Looks like this is testing the specific sub-case when listFiles returns null. If so, let's rename method to reflect that (e.g. testRemoveOldCheckpointsWhenXYZ

bkonold: done

PR #1263, opened 2020-01-29
SAMZA-2448: [Scala cleanup] Clean up and convert some parts of samza-kafka to java

cameronlee314:
Issues: Most of samza-kafka is now in Java, so it would be good to convert more classes from Scala to Java. Some Java classes are within the src/main/scala directory.

Changes:
1. Moved some classes around so that they are in the proper location.
2. Rewrote KafkaUtil.scala into KafkaUtil.java
3. Removed some unused classes

Testing: ./gradlew build

API changes: None
Upgrade/usage instructions: None


samza-kafka/src/main/java/org/apache/samza/util/KafkaUtil.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.samza.execution.StreamManager;
+import org.apache.samza.system.OutgoingMessageEnvelope;
+import org.apache.samza.system.SystemStreamPartition;
+
+
+public class KafkaUtil {
+  private static final int CHECKPOINT_LOG_VERSION_NUMBER = 1;
+
+  public static String getCheckpointTopic(String jobName, String jobId, Config config) {
+    String checkpointTopic = String.format("__samza_checkpoint_ver_%d_for_%s_%s", CHECKPOINT_LOG_VERSION_NUMBER,
```
Sanil15: minor: define: "__samza_checkpoint_ver_%d_for_%s_%s" and static final?

samza-kafka/src/main/java/org/apache/samza/util/KafkaUtil.java
```diff
@@ -0,0 +1,56 @@
...
+        jobName.replaceAll("_", "-"), jobId.replaceAll("_", "-"));
+    return StreamManager.createUniqueNameForBatch(checkpointTopic, config);
+  }
+
+  /**
+   * Partition key in the envelope must not be null.
+   */
+  public static Integer getIntegerPartitionKey(OutgoingMessageEnvelope envelope, List<PartitionInfo> partitions) {
```
mynameborat: Does it have to be boxed? Can we replace this w/ primitive?

PR #1265, opened 2020-01-31
SAMZA-2441: Update ApplicationRunnerMain#ApplicationRunnerCommandLine not to load local file

kw2542:
Design:

Changes:
1. Override ApplicationRunnerCommandLine#loadConfig not to invoke config loader as ApplicationRunners handles initial config and full job config themselves.
2. Add JobRunnerCommandLine to load full job config for JobRunner to keep consistent with previous behavior as it still needs full job config for special use cases.

API Changes:
None

Upgrade Instructions:
None

Usage Instructions:
None

Tests:
1. Unit Tests

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
+++ b/samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
@@ -19,9 +19,15 @@
 
 package org.apache.samza.runtime;
 
+import java.util.Collections;
+import java.util.Map;
+import java.util.stream.Collectors;
 import joptsimple.OptionSet;
 import joptsimple.OptionSpec;
 import org.apache.samza.config.Config;
+import org.apache.samza.config.ConfigLoaderFactory;
+import org.apache.samza.config.JobConfig;
+import org.apache.samza.config.MapConfig;
 import org.apache.samza.util.CommandLine;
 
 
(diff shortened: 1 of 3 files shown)


samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
```diff
@@ -43,9 +49,33 @@ public ApplicationRunnerOperation getOperation(OptionSet options) {
       String rawOp = options.valueOf(operationOpt);
       return ApplicationRunnerOperation.fromString(rawOp);
     }
+
+    @Override
+    public Config loadConfig(OptionSet options) {
```
cameronlee314: Do you feel that the "base" use case for `CommandLine` is to need the full config? To me, it looks a little awkward to need to override `loadConfig` here with an implementation which is almost the same as the base `loadConfig`.
I suppose one alternative would be to make this impl here the base impl, and then have each use case call `ConfigUtil.loadConfig` if is needs it. What do you see as the tradeoffs?

samza-core/src/test/java/org/apache/samza/runtime/TestApplicationRunnerMain.java
```diff
@@ -79,6 +82,28 @@ public void TestStatusOperation() throws Exception {
...
+
+    Config actual = cmdLine.loadConfig(options);
+
+    assertEquals(4, actual.size());
+    assertEquals("org.apache.samza.config.loaders.PropertiesConfigLoaderFactory", actual.get(JobConfig.CONFIG_LOADER_FACTORY));
+    assertEquals(
+        getClass().getResource("/test.properties").getPath(),
+        actual.get(ConfigLoaderFactory.CONFIG_LOADER_PROPERTIES_PREFIX + "path"));
+    assertEquals(MockStreamApplication.class.getName(), actual.get(ApplicationConfig.APP_CLASS));
+    assertEquals(TestApplicationRunnerInvocationCounts.class.getName(), actual.get("app.runner.class"));
```
cameronlee314: Minor suggestion: Instead of needing to check the size and each individual entry, could you build the expected config map and then do an equals check?

kw2542: Good point, updated.

samza-core/src/main/java/org/apache/samza/runtime/ApplicationRunnerMain.java
kw2542: Good question, in all 13 direct usage of CommandLine, and 5 sub classes of CommandLine, they are all expecting full job configs to work properly, e.g. RocksDbReadingTool extends CommandLine to read full job config before being able to read data from Rocks DB, which is being invoked by CLI.

If we change CommandLine to the other way, that means we need to invoke ConfigUtil.loadConfig explicitly in 13 places, and override loadConfigs in 5 sub classes to invoke ConfigUtil.loadConfig too, compared to overriding once in ApplicationRunnerCommandLine only.

I would prefer the current way to keep the base class to serve majority use case.

PR #1266, opened 2020-02-03
SAMZA-2452 : Adding internal autosizing related configs

rmatharu:
Adding internal autosizing related configs, can be used by an external controller. 
These sizing-related configs (e.g., container-count, etc) take precedence when job.autosizing.enabled is set, otherwise existing behavior is retained.

Symptom: Sizing configs set by a controller need to be different from user-facing configs.

Cause: --
Fix: Added internal configs for sizing related configs. Added tests. 
The enabled config is also needs to be emitted by the DiagnosticsManager.
API Changes: Adding internal autosizing related configs (job.autosizing.enabled, job.autosizing.container.count, job.autosizing.container.thread.pool.size, job.autosizing.maxheap.mb, job.autosizing.cpu.cores) , can be used by an external controller. These sizing-related configs (e.g., container-count, etc) take precedence when job.autosizing.enabled is set, otherwise existing behavior is retained.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/config/JobConfig.java
+++ b/samza-core/src/main/java/org/apache/samza/config/JobConfig.java
@@ -130,6 +130,15 @@ public class JobConfig extends MapConfig {
   public static final String CONTAINER_METADATA_FILENAME_FORMAT = "%s.metadata"; // Filename: <containerID>.metadata
   public static final String CONTAINER_METADATA_DIRECTORY_SYS_PROPERTY = "samza.log.dir";
 
+  // Auto-sizing related configs tthat ake precedence over respective sizing confings job.container.count, etc,
+  // *only* when job.autosizing.enabled is true. Otherwise current behavior is maintained.
+  public static final String JOB_AUTOSIZING_ENABLED = "job.autosizing.enabled";
+  public static final String JOB_AUTOSIZING_CONTAINER_COUNT = "job.autosizing.container.count";
+  public static final String JOB_AUTOSIZING_CONTAINER_THREAD_POOL_SIZE = "job.autosizing.container.thread.pool.size";
+  public static final String JOB_AUTOSIZING_CONTAINER_MAX_HEAP_MB = "job.autosizing.container.maxheap.mb";
+  public static final String JOB_AUTOSIZING_CONTAINER_MEMORY_MB = "job.autosizing.container.memory.mb";
+  public static final String JOB_AUTOSIZING_CONTAINER_MAX_CORES = "job.autosizing.container.cpu.cores";
+
   public static final String COORDINATOR_STREAM_FACTORY = "job.coordinatorstream.config.factory";
   public static final String DEFAULT_COORDINATOR_STREAM_CONFIG_FACTORY = "org.apache.samza.util.DefaultCoordinatorStreamConfigFactory";
 
@@ -166,8 +175,12 @@ public class JobConfig extends MapConfig {
   }
 
   public int getContainerCount() {
+    Optional<String> autoscalingContainerCountValue = Optional.ofNullable(get(JOB_AUTOSIZING_CONTAINER_COUNT));
     Optional<String> jobContainerCountValue = Optional.ofNullable(get(JOB_CONTAINER_COUNT));
-    if (jobContainerCountValue.isPresent()) {
+
+    if (getAutosizingEnabled() && autoscalingContainerCountValue.isPresent()) {
+      return Integer.parseInt(autoscalingContainerCountValue.get());
+    } else if (jobContainerCountValue.isPresent()) {
       return Integer.parseInt(jobContainerCountValue.get());
     } else {
       // To maintain backwards compatibility, honor yarn.container.count for now.
@@ -287,7 +300,13 @@ public class JobConfig extends MapConfig {
   }
 
   public int getThreadPoolSize() {
-    return getInt(JOB_CONTAINER_THREAD_POOL_SIZE, 0);
+    Optional<String> autoscalingContainerThreadPoolSize = Optional.ofNullable(get(
+        JOB_AUTOSIZING_CONTAINER_THREAD_POOL_SIZE));
+    if (getAutosizingEnabled() && autoscalingContainerThreadPoolSize.isPresent()) {
+      return Integer.parseInt(autoscalingContainerThreadPoolSize.get());
+    } else {
+      return getInt(JOB_CONTAINER_THREAD_POOL_SIZE, 0);
+    }
   }
 
   public int getDebounceTimeMs() {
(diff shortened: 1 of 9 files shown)


prateekm:
@cameronlee314 @abhishekshivanna can you review?

samza-core/src/main/scala/org/apache/samza/util/CoordinatorStreamUtil.scala
```diff
@@ -163,7 +163,14 @@ object CoordinatorStreamUtil extends Logging {
 
     val oldConfig = coordinatorSystemConsumer.getConfig
     if (resetJobConfig) {
-      val keysToRemove = oldConfig.keySet.asScala.toSet.diff(config.keySet.asScala)
+      var keysToRemove = oldConfig.keySet.asScala.toSet.diff(config.keySet.asScala)
+
+      val jobConfig = new JobConfig(config)
+      if (jobConfig.getAutosizingEnabled) {
+        // If autosizing is enabled, we retain auto-sizing related configs
+        keysToRemove = keysToRemove.filter(configKey => !jobConfig.isAutosizingConfig(configKey))
```
cameronlee314: Is there a case where you want to intentionally delete an autosizing config? For example, for cleaning up after getting a stable set of parameters.

samza-core/src/main/scala/org/apache/samza/config/ShellCommandConfig.scala
```diff
@@ -116,7 +116,25 @@ object ShellCommandConfig {
 class ShellCommandConfig(config: Config) extends ScalaMapConfig(config) {
   def getCommand = getOption(ShellCommandConfig.COMMAND_SHELL_EXECUTE).getOrElse("bin/run-container.sh")
 
-  def getTaskOpts = getOption(ShellCommandConfig.TASK_JVM_OPTS)
+  def getTaskOpts = {
+    var jvmOpts = getOption(ShellCommandConfig.TASK_JVM_OPTS)
+    val jobConfig = new JobConfig(config)
+
+    if (jobConfig.getAutosizingEnabled && getOption(JobConfig.JOB_AUTOSIZING_CONTAINER_MAX_HEAP_MB).isDefined) {
```
cameronlee314: Could you please add a unit test for this?

cameronlee314:
Can you please add a JIRA ticket to the PR and add an "API changes" section to the description (see https://cwiki.apache.org/confluence/display/SAMZA/SEP-25%3A+PR+Title+And+Description+Guidelines)?

samza-core/src/main/scala/org/apache/samza/util/CoordinatorStreamUtil.scala
rmatharu: The expectation there is that the user copies the internal config values to respective user configs (e.g., using a tool), and disables autosizing (which will delete these internal configs).

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
```diff
@@ -310,6 +339,23 @@ public boolean getDiagnosticsEnabled() {
...
+    return getBoolean(JOB_AUTOSIZING_ENABLED, false);
+  }
+
+  public boolean isAutosizingConfig(String configParam) {
+    switch (configParam) {
+      case JOB_AUTOSIZING_CONTAINER_COUNT:
+      case JOB_AUTOSIZING_CONTAINER_MAX_CORES:
+      case JOB_AUTOSIZING_CONTAINER_MAX_HEAP_MB:
+      case JOB_AUTOSIZING_CONTAINER_MEMORY_MB:
+      case JOB_AUTOSIZING_CONTAINER_THREAD_POOL_SIZE:
```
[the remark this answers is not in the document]
rmatharu: Chose 3, and added comments and a config prefix variable.

rmatharu:
Addressed all comments, please take a look

samza-core/src/test/java/org/apache/samza/config/TestJobConfig.java
```diff
@@ -608,4 +604,42 @@ public void testAutosizingConfig() {
...
+    shellCommandConfig = new ShellCommandConfig(new MapConfig(
+        ImmutableMap.of(JobConfig.JOB_AUTOSIZING_ENABLED, "true", JobConfig.JOB_AUTOSIZING_CONTAINER_MAX_HEAP_MB,
+            "1024")));
+    assertEquals(Option.apply("-Xmx1024m"), shellCommandConfig.getTaskOpts());
+
+    // opts set without -Xmx, autosizing max heap set
+    shellCommandConfig = new ShellCommandConfig(new MapConfig(
+        ImmutableMap.of(JobConfig.JOB_AUTOSIZING_ENABLED, "true", JobConfig.JOB_AUTOSIZING_CONTAINER_MAX_HEAP_MB,
+            "1024")));
+    assertEquals(Option.apply("-Xmx1024m"), shellCommandConfig.getTaskOpts());
```
cameronlee314: I think this section is the same as the block above. Seems like the code doesn't match the comment.

samza-core/src/test/java/org/apache/samza/config/TestJobConfig.java
```diff
@@ -610,11 +610,6 @@ public void testGetTaskOptsAutosizingDisabled() {
     ShellCommandConfig shellCommandConfig =
         new ShellCommandConfig(new MapConfig(ImmutableMap.of(JobConfig.JOB_AUTOSIZING_ENABLED, "false")));
     assertEquals(Option.empty(), shellCommandConfig.getTaskOpts());
-
-    String taskOpts = "-Dproperty=value";
-    shellCommandConfig = new ShellCommandConfig(new MapConfig(
-        ImmutableMap.of(ShellCommandConfig.TASK_JVM_OPTS(), taskOpts, JobConfig.JOB_AUTOSIZING_ENABLED, "false")));
-    assertEquals(Option.apply(taskOpts), shellCommandConfig.getTaskOpts());
```
cameronlee314: I don't think this should be removed. It is actually a different test case (i.e. autosizing is disabled).

samza-core/src/main/java/org/apache/samza/config/JobConfig.java
```diff
@@ -130,6 +130,15 @@
   public static final String CONTAINER_METADATA_FILENAME_FORMAT = "%s.metadata"; // Filename: <containerID>.metadata
   public static final String CONTAINER_METADATA_DIRECTORY_SYS_PROPERTY = "samza.log.dir";
 
+  // Auto-sizing related configs tthat ake precedence over respective sizing confings job.container.count, etc,
```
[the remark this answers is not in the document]
cameronlee314: It looks like you still have a typo ("that ake") after your last change.

samza-core/src/test/java/org/apache/samza/config/TestJobConfig.java
cameronlee314: 1. I saw your change, but I meant to refer to the lines immediately above this. I think this code block needs to be changed (not removed) to have a non-empty opts that doesn't have an -Xmx param.
2. I would also suggest adding a test in which opts has -Xmx but needs to be replaced by the autosizing config.

PR #1269, opened 2020-02-07
SAMZA-2455: Validate the argument types in SamzaSQL UDF on execution planning phase

shanthoosh:
Issues: Incompatible types between UDF and schema is a irrecoverable error and it would kill the SamzaSQL job. Argument type validations in SamzaSQL UDF happens at runtime after the containers are launched. This causes operational inconvenience since these UDF type-errors causes a sql-job to fail at runtime.

Changes: This patch adds UDF validations to execution planning phase of SQL application. Operand checker in SamzaSQL is extended to house the udf argument type validation. This will ensure that the deployment of error-prone Samza SQL UDF will fail.

API changes: None

Upgrade/usage instructions: None


shanthoosh:
@weiqingy Can you please take a look at this PR?

srinipunuru:
@santhoshvenkat1988  Couple of things that you could look at in future

1. If adding validations with UDF polymorphism is hard, We could disable/remove it.
2. May be an end to end planning test UDf type validation test would be great.

shanthoosh:
Thanks for the review @srinipunuru 
- I've updated the PR description.
- I have added the end-to-end integration test for mismatched SamzaSQL UDF type.
- Created a follow-up ticket to disallow the UDF polymorphism.

PR #1279, opened 2020-02-18
SAMZA-2462 : Adding metric for container thread pool size

rmatharu:
Adding metric for container thread pool size.

Symptom: Currently it is not possible to track the container-thread-pool size using metrics, even though its possible to configure it similar to container-memory-mb, vcore, etc.
Cause: No metric for this configurable
Fix: Added metric for container thread pool size, updated metrics reference sheet.
API changes: Added new metric for container-thread-pool-size.
Upgrade Instructions: None
Usage Instructions: See the metrics reference sheet for details on the metric.


prateekm:
Please include API changes etc. sections for the PR description. This way it's easy to collect them in the changelog during a release.

PR #1280, opened 2020-02-18
SAMZA-2461: Fix Concurrent Modification Exception in InMemorySystem

Zhangyx39:
Symptom: There is a chance to run into ConcurrentModificationException using InMemorySystem.

Cause: The returned list of InMemoryManager#poll is a sublist of the original list. While iterating the sublist, if the original list is modified, ConcurrentModificationException will be thrown.

Changes: We need to make the returned list of InMemoryManager#poll to be a new copy.

Tests: Tested with a samza beam job. Didn't see the exception any more.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
+++ b/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
@@ -20,6 +20,7 @@
 package org.apache.samza.system.inmemory;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -199,6 +200,6 @@ class InMemoryManager {
       return new ArrayList<>();
     }
 
-    return messageEnvelopesForSSP.subList(startingOffset, messageEnvelopesForSSP.size());
+    return ImmutableList.copyOf(messageEnvelopesForSSP.subList(startingOffset, messageEnvelopesForSSP.size()));
   }
 }


prateekm:
Please follow https://cwiki.apache.org/confluence/display/SAMZA/SEP-25%3A+PR+Title+And+Description+Guidelines for the PR description format.

mynameborat:
Looks good to me. Can you add a unit test?

Zhangyx39:
> Looks good to me. Can you add a unit test?

A unit test I can think of is to call poll twice to get two lists. Then try to modify one list and verity that the other list won't be affected. But since we are using immutable list here. The test turns to be testing the immutability of the returned list (expected method not supported exception)...
Does it sound good to you? Do you have a better idea on this? @mynameborat

mynameborat:
> > Looks good to me. Can you add a unit test?
> 
> A unit test I can think of is to call poll twice to get two lists. Then try to modify one list and verity that the other list won't be affected. But since we are using immutable list here. The test turns to be testing the immutability of the returned list (expected method not supported exception)...
> Does it sound good to you? Do you have a better idea on this? @mynameborat

Can you create a follow up patch for the unit tests? I will get this patch in since its needed for 1.4 release.

Zhangyx39:
> > > Looks good to me. Can you add a unit test?
> > 
> > 
> > A unit test I can think of is to call poll twice to get two lists. Then try to modify one list and verity that the other list won't be affected. But since we are using immutable list here. The test turns to be testing the immutability of the returned list (expected method not supported exception)...
> > Does it sound good to you? Do you have a better idea on this? @mynameborat
> 
> Can you create a follow up patch for the unit tests? I will get this patch in since its needed for 1.4 release.

Thanks. I have create a ticket and will add the test soon. https://issues.apache.org/jira/browse/SAMZA-2467

PR #1281, opened 2020-02-18
SAMZA-2378: Container Placements support for Standby containers enabled jobs

Sanil15:
**Design:** SEP-22: Container Placements in Samza

**Changes:** Following abilities are added as a part of this PR:
1. Ability to issue container placement actions for standby enabled jobs on active & standby containers
2. Ability to queue actions if there are any in-flight actions on active or its standby or vice-versa
3. Validate placement actions when standby containers are enabled to ensure standby constraints (active container & standby container are not on the same host) are not violated

**Tests:**
Integ test: End-to-end test for standby enabled job is added 
Manual testing matrix: Container Placement Test Plan

**API Changes:** 
1. This PR introduces behavior changes when a container placement action is written to metastore and is read by JobCoordinator (JC), it will try to reserve resources for the request, if it gets the resources it attempts to stop the active container then attempts to start the container on the allocated resources.
2. Each placement request is identified by a UUID which is generated and given to a client on issuing a control command
3. Responses for each placement action can be queried by directly queueing meta-store by UUID

**Usage Instructions:** Instantiate ContainerPlacementMetadataStore to write Container placement messages to Metastore, query it by the UUID generated


samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -48,8 +48,7 @@
  * ContainerManager encapsulates logic and state related to container placement actions like move, restarts for active container
  * if issued externally.
  *
- * TODO SAMZA-2378: Container Placements for Standby containers enabled jobs
- *      SAMZA-2379: Container Placements for job running in degraded state
+ * TODO SAMZA-2379: Container Placements for job running in degraded state
```
rmatharu: We should probably prioritize this sooner than later

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -136,7 +135,22 @@ boolean handleContainerLaunch(SamzaResourceRequest request, String preferredHost
         LOG.info("Waiting for running container to shutdown due to existing ContainerPlacement action {}", actionMetaData);
         return false;
       } else if (actionStatus == ContainerPlacementMetadata.ContainerStatus.STOPPED) {
-        allocator.runStreamProcessor(request, preferredHost);
+        // If the job has standby containers enabled, always check standby constraints before issuing a start on container
```
rmatharu: Can the lines above 
`
if (hasActiveContainerPlacementAction(request.getProcessorId())) {
      String processorId = request.getProcessorId();	      String processorId = request.getProcessorId();
      ContainerPlacementMetadata actionMetaData = getPlacementActionMetadata(processorId).get();	      ContainerPlacementMetadata actionMetaData = getPlacementActionMetadata(processorId).get();
`

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -136,7 +135,22 @@ boolean handleContainerLaunch(SamzaResourceRequest request, String preferredHost
...
+        if (standbyContainerManager.isPresent() && !standbyContainerManager.get().checkStandbyConstraints(request.getProcessorId(), allocatedResource.getHost())) {
+          LOG.info(
+              "Starting container {} on host {} does not meet standby constraints, falling back to source host placement metadata: {}",
+              request.getProcessorId(), preferredHost, actionMetaData);
+          resourceRequestState.releaseUnstartableContainer(allocatedResource, preferredHost);
+          resourceRequestState.cancelResourceRequest(request);
+          // Fallback to source host since the new allocated resource does not meet standby constraints
+          allocator.requestResource(processorId, actionMetaData.getSourceHost());
+          markContainerPlacementActionFailed(actionMetaData,
+              String.format("allocated resource %s does not meet standby constraints now, falling back to source host", allocatedResource));
```
rmatharu: Update failedStandbyAllocations metric?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -136,7 +135,22 @@ boolean handleContainerLaunch(SamzaResourceRequest request, String preferredHost
...
+        } else {
+          LOG.info("Status updated for ContainerPlacement action: ", actionMetaData);
```
rmatharu: Status wasnt updated here?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
Sanil15: hasActiveContainerPlacementAction checks the metadata of the action to be either in ACCEPTED or IN_PROGRESS, it not a check for the presence of metadata

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -412,16 +429,23 @@ private void writeContainerPlacementResponseMessage(ContainerPlacementRequestMes
...
    * @param requestMessage container placement request message
    * @return true if action should be taken right now, false if it needs to wait to be taken in future
    */
   private boolean deQueueAction(ContainerPlacementRequestMessage requestMessage) {
     // Do not dequeue action wait for the in-flight action to complete
-    if (hasActiveContainerPlacementAction(requestMessage.getProcessorId())) {
-      LOG.info("ContainerPlacement request: {} is en-queued because container has an in-progress placement action", requestMessage);
+    if (hasActiveContainerPlacementAction(requestMessage.getProcessorId())
+        || checkStandbyOrActiveContainerHasActivePlacementAction(requestMessage.getProcessorId())) {
+      if (standbyContainerManager.isPresent()) {
```
rmatharu: This logic seems twisted, 
we check if there there is an active-action on the container-id, or if there is an active-action on its active/standby counterparts, 
after that we check if there is a standby-container-manager present?

Would it be possible to 
a. first check if a standby-container-manager is present?
Or
b. can hasActiveContainerPlacementAction encapsulate the logic of checking a and checking with if there is an active-action on its active/standby counterparts of the given processor-id.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -435,32 +459,59 @@ private boolean deQueueAction(ContainerPlacementRequestMessage requestMessage) {
...
   /**
    * A valid container placement action needs a valid processor id. Duplicate actions are handled by de-duping on uuid.
+   * If standby containers are enabled destination host requested must meet standby constraints
    *
    * @param requestMessage container placement request message
    * @return Pair<ContainerPlacementMessage.StatusCode, String> which is status code & response suggesting if the request is valid
    */
   private Pair<ContainerPlacementMessage.StatusCode, String> validatePlacementAction(ContainerPlacementRequestMessage requestMessage) {
-    String errorMessagePrefix = String.format("ContainerPlacement request: %s is rejected due to", requestMessage);
+    String errorMessagePrefix = ContainerPlacementMessage.StatusCode.BAD_REQUEST + " reason: ";
```
rmatharu: Instead can we define errorMessage here 
as ContainerPlacementMessage.StatusCode.BAD_REQUEST + " reason: %s"

and later String.format(errorMessage, reason);

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -435,32 +459,59 @@ private boolean deQueueAction(ContainerPlacementRequestMessage requestMessage) {
...
+      if (!StandbyTaskUtil.isStandbyContainer(processorId)) {
+        for (String standby : standbyContainerManager.get().getStandbyList(processorId)) {
+          if (hasActiveContainerPlacementAction(standby)) {
+            return true;
+          }
+        }
+      }
+    }
+    return false;
+  }
```
rmatharu: This method can be simplified/inlined with hasActiveAction?

samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java
```diff
@@ -425,6 +423,15 @@ public void handleExpiredResourceRequest(String containerID, SamzaResourceReques
...
   }
 
+  /**
+   * Fetches a list of standby container for an active container
+   * @param activeContainerId logical id of the container ex: 0,1,2
+   * @return list of standby containers ex: for active container 0: {0-0, 0-1}
+   */
+  List<String> getStandbyList(String activeContainerId) {
+    return this.standbyContainerConstraints.get(activeContainerId);
+  }
```
rmatharu: see comment above on simplifying,
also make public?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
Sanil15: done

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
Sanil15: I see your point, let me refactor it

lakshmi-manasa-g:
yes, the failing test (at org.apache.samza.system.azureblob.avro.TestAzureBlobOutputStream.testClose) in the last build here is being fixed in  https://github.com/apache/samza/pull/1298

For now, do an empty commit to trigger another build and it should mostly pass. this test is flaky and wont fail in all Travis builds.

PR #1283, opened 2020-02-20
SAMZA-2464: Container shuts down when task fails to remove old state checkpoint dirs

bkonold:
**Symptom:** Container shuts down with exception on invocation of TaskStorageManager.removeOldCheckpoints from TaskInstance.commit
 
**Cause:** Concurrent modification of checkpoint directories by other processes / threads may cause FileNotFoundException to be thrown and shutdown the container. IOException may be thrown for other miscellaneous failures; these should not cause the container to shutdown but be logged and allow processing to continue.
 
**Tests:** Added a unit test which fails when exception from removeOldCheckpoints is not caught.

**API Changes:** None
**Usage instructions:** None

### DIFF
--- a/samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
+++ b/samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
@@ -277,7 +277,7 @@ class TaskInstance(
       try {
         storageManager.removeOldCheckpoints(checkpointId)
       } catch {
-        case e: Exception => error("Failed to remove old checkpoints", e)
+        case e: Exception => error("Failed to remove old checkpoints for task: %s. Current checkpointId: %s" format (taskName, checkpointId), e)
       }
     }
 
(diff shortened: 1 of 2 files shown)


samza-core/src/main/scala/org/apache/samza/container/TaskInstance.scala
```diff
@@ -274,7 +274,11 @@ class TaskInstance(
 
     if (storageManager != null) {
       trace("Remove old checkpoint stores for taskName: %s" format taskName)
-      storageManager.removeOldCheckpoints(checkpointId)
+      try {
+        storageManager.removeOldCheckpoints(checkpointId)
+      } catch {
+        case e: Exception => error("Failed to remove old checkpoints", e)
```
prateekm: Minor: include current checkpoint id and task name in message.

bkonold: ack, i'll add those

PR #1292, opened 2020-02-25
SAMZA-2469: Fix out of order events throwing Duplicate key registration with same timer exception

kathulasandeep:
https://issues.apache.org/jira/browse/SAMZA-2469

When using sessions with certain gap and with triggers in beam code with samza runner, for out of order events we are seeing duplicate key registration for the same timer exception.

The exception is thrown in two cases that happen within context of one trigger:
  1.  Two events for same key have same timestamp 
  2.  Events come out of order

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/scheduler/EpochTimeScheduler.java
+++ b/samza-core/src/main/java/org/apache/samza/scheduler/EpochTimeScheduler.java
@@ -19,14 +19,12 @@
 
 package org.apache.samza.scheduler;
 
-import java.util.Map;
-import java.util.TreeMap;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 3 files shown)


prateekm:
Thanks for the contribution @kathulasandeep.
@bharathkk @xinyuiscool can you take a look at this? Do we need this in 1.4?

mynameborat:
@kathulasandeep Sorry for not getting back earlier. The problem outlined is fixed as part of https://issues.apache.org/jira/browse/SAMZA-2519. Making the timer API callbacks multi-valued doesn't align with timer contracts on Beam and other APIs. To be consistent with semantics of the Timer API (Beam, Samza) and the semantics  of timers themselves across runners, we fixed this issue to cancel the previously registered callback and overwrite it with the new callback provided previous callback can be successfully unregistered.

mynameborat:
I am closing this PR since this is already fixed.

PR #1293, opened 2020-02-26
SAMZA-2472: Use runtime-framework-resources-pathing.jar to specify part of the runtime classpath and leverage it in IsolatingClassLoaderFactory

cameronlee314:
**Symptom**: When running in YARN and job coordinator dependency isolation is enabled, the framework infrastructure is unable to access the correct YARN configurations. This means that when it tries to interact with YARN, it is not using the correct configurations, and that results in issues. The observed issue was that the YARN client from the Samza side did not follow the RM HA config correctly, and so there was an issue in connecting to the RM.
**Cause**: The YARN client reads configurations from certain resource files on the classpath. In `run-class.sh`, the `HADOOP_CONF_DIR` is directly used and added to the classpath. In job coordinator dependency isolation mode, the classpath in `run-class.sh` is not used for the framework infrastructure, since `IsolatingClassLoaderFactory` builds its own classpath. Therefore, the `HADOOP_CONF_DIR` is not on the framework infrastructure classpath.
**Changes**: Refactored `run-class.sh` to create an additional "runtime framework resources" pathing JAR (`runtime-framework-resources-pathing.jar`). The `HADOOP_CONF_DIR` is included in this new pathing JAR. `runtime-framework-resources-pathing.jar` is added to the regular classpath in order to keep backwards compatibility, and it is also detected and used in `IsolatingClassLoaderFactory` when isolation is enabled.
**Tests**: Deployed a job with isolation enabled in `samza-hello-samza` and verified using a log that `yarn.resourcemanager.scheduler.class` is properly set (the default of CapacityScheduler is overridden to FifoScheduler in the `samza-hello-samza` set-up). I also deployed the job without the fix and verified that the `yarn.resourcemanager.scheduler.class` property was not properly overridden.
**API changes**: None
**Upgrade/usage instructions**: None

An additional benefit of this is that we can add other runtime resources in the future to this pathing JAR (e.g. `ADDITIONAL_CLASSPATH_DIR`) and not need to worry about argument limits.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/classloader/DependencyIsolationUtils.java
+++ b/samza-core/src/main/java/org/apache/samza/classloader/DependencyIsolationUtils.java
@@ -45,4 +45,6 @@ public class DependencyIsolationUtils {
    * classloader.
    */
   public static final String FRAMEWORK_API_CLASS_LIST_FILE_NAME = "samza-framework-api-classes.txt";
+
+  public static final String RUNTIME_FRAMEWORK_RESOURCES_PATHING_JAR_NAME = "runtime-framework-resources-pathing.jar";
 }
--- a/samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
+++ b/samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
@@ -35,6 +35,8 @@ import java.net.URLClassLoader;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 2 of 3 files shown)


prateekm:
@cameronlee314 
"The pathing JAR is added to the regular classpath in order to keep backwards compatibility" 
Doesn't look like we're doing this anymore, since base-lib-pathing.jar and the -cp args don't contain HADOOP_CONF_DIR now. Is that intentional? 

Otherwise LGTM, thanks for the thorough explanation.

cameronlee314:
> @cameronlee314
> "The pathing JAR is added to the regular classpath in order to keep backwards compatibility"
> Doesn't look like we're doing this anymore, since base-lib-pathing.jar and the -cp args don't contain HADOOP_CONF_DIR now. Is that intentional?
> 
> Otherwise LGTM, thanks for the thorough explanation.

`HADOOP_CONF_DIR` is placed into `runtime-framework-resources-pathing.jar` and `runtime-framework-resources-pathing.jar` is put into the `-cp`. So `HADOOP_CONF_DIR` is still on the classpath through `runtime-framework-resources-pathing.jar`; it is intentionally removed from the direct `-cp` argument, but the functionality should still be the same. Hopefully that addresses your question.

prateekm:
I assumed by backwards compatible you meant both pathing jars (base-lib and runtime-framework-resources) will have the conf dir. Is this only used within the runtime-framework-resources classloader?

cameronlee314:
> I assumed by backwards compatible you meant both pathing jars (base-lib and runtime-framework-resources) will have the conf dir. Is this only used within the runtime-framework-resources classloader?

By backwards compatible, I meant that when using `run-class.sh` to run a class, the same resources will be available on the classpath. I think the actual name/contents of the pathing JARs are internal details, so as long as the final classpath is the same, the details of the pathing JAR(s) can change without being considered as a backwards incompatible change.
`HADOOP_CONF_DIR` is used in both non-isolation mode and isolation mode. In non-isolation mode, it is accessed through `runtime-framework-resources-pathing.jar` which is passed through the `-cp` argument. In isolation mode, `IsolatingClassLoaderFactory` will add `runtime-framework-resources-pathing.jar` to the classpath of the infrastructure classloader.

PR #1296, opened 2020-02-28
SAMZA-2475: Add a allocated resource expiry timeout in samza yarn type of apps

Sanil15:
**Changes:** Today if samza apps are not able to use an allocated resource within
yarn.resourcemanager.rm.container-allocation.expiry-interval-ms
start of the container fails with this exception, this can be avoided by just setting an allocated resource timeout check before issuing a start

**API Changes:** Added isResourceExpired to ClusterResourceManager

**Usage Instructions:** None

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/clustermanager/ClusterResourceManager.java
+++ b/samza-core/src/main/java/org/apache/samza/clustermanager/ClusterResourceManager.java
@@ -132,6 +132,15 @@ public abstract class ClusterResourceManager {
 
   public abstract void stop(SamzaApplicationState.SamzaAppStatus status);
 
+  /**
+   * Checks if the allocated resource is expired. If the {@link ClusterResourceManager} does not have a
+   * concept of expired allocated resource we assume allocated resources never expire
+   * @param resource allocated resource
+   * @return if the allocated resource is expired
+   */
+  public boolean isResourceExpired(SamzaResource resource) {
+    return false;
+  }
 
   /***
    * Defines a callback interface for interacting with notifications from a ClusterResourceManager
--- a/samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
+++ b/samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
@@ -261,6 +261,14 @@ public class ContainerAllocator implements Runnable {
       throw new SamzaException("Expected resource for Processor ID: " + request.getProcessorId() + " was unavailable on host: " + preferredHost);
     }
 
+    /**
+     * If the allocated resource has expired then release the expired resource and re-request the resources from {@link ClusterResourceManager}
+     */
+    if (clusterResourceManager.isResourceExpired(resource)) {
+      containerManager.handleExpiredResource(request, resource, preferredHost, resourceRequestState, this);
+      return;
+    }
+
     // Update state
     resourceRequestState.updateStateAfterAssignment(request, preferredHost, resource);
     String processorId = request.getProcessorId();
--- a/samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
+++ b/samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
@@ -305,12 +305,12 @@ public class ContainerManager {
         resource, request.getProcessorId(), request);
     resourceRequestState.releaseUnstartableContainer(resource, preferredHost);
     resourceRequestState.cancelResourceRequest(request);
-    SamzaResourceRequest resourceRequest = allocator.getResourceRequest(request.getProcessorId(), request.getPreferredHost());
-    if (hasActiveContainerPlacementAction(request.getProcessorId())) {
+    SamzaResourceRequest newResourceRequest = allocator.getResourceRequest(request.getProcessorId(), request.getPreferredHost());
+    if (hasActiveContainerPlacementAction(newResourceRequest.getProcessorId())) {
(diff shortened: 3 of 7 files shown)


samza-core/src/main/java/org/apache/samza/clustermanager/ClusterResourceManager.java
```diff
@@ -132,6 +132,9 @@ public ClusterResourceManager(Callback callback) {
 
   public abstract void stop(SamzaApplicationState.SamzaAppStatus status);
 
+  public boolean isResourceExpired(SamzaResource resource) {
```
rmatharu: javadoc please
also, this method likely belongs in SamzaResource?
(if it really needs to be here, needs to be abstract)

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
```diff
@@ -261,6 +261,14 @@ protected void runStreamProcessor(SamzaResourceRequest request, String preferred
       throw new SamzaException("Expected resource for Processor ID: " + request.getProcessorId() + " was unavailable on host: " + preferredHost);
     }
 
+    if(clusterResourceManager.isResourceExpired(resource)) {
```
rmatharu: resource.isExpired()

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
```diff
@@ -261,6 +261,14 @@ protected void runStreamProcessor(SamzaResourceRequest request, String preferred
...
+      LOG.info("Lease for the resource has been expired for resource: {} request: {}", resource, request);
```
rmatharu: Resource has expired, re-requesting resource for ..... processorId: 
Having the processorId: pattern in the logs makes for easy searcheability when we're debugging

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnClusterResourceManager.java
```diff
@@ -575,6 +576,15 @@ public void onStopContainerError(ContainerId containerId, Throwable t) {
     }
   }
 
+  @Override
+  public boolean isResourceExpired(SamzaResource resource) {
+    /**
+     * Time from which resource was allocated + 1 min (to account for clock skew) > Yarn Expiry Timeout
+     */
+    return System.currentTimeMillis() - resource.getTimestamp() + Duration.ofMinutes(1).toMillis()
+        > YarnConfiguration.DEFAULT_RM_CONTAINER_ALLOC_EXPIRY_INTERVAL_MS;
```
rmatharu: Could DEFAULT_RM_CONTAINER_ALLOC_EXPIRY_INTERVAL_MS be a Duration?

Duration.between(resource.getTS, Instant.now()).compareTo(DEFAULT_RM_CONTAINER_ALLOC_EXPIRY_INTERVAL_MS) > 0;

Sanil15:
This was just a sample PR to get some thoughts on possible patches, closing this, will fix in another PR

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterResourceManager.java
Sanil15: Kubernetes does not have a concept of expired allocated resource so this method needs a default implementation, it cannot belong to SamzaResource because that is yarn agnostic class and timeout is read from Yarn config and not a samza config.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerAllocator.java
Sanil15: I came up with this PR to gather some early thoughts on approach, will clean it up

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnClusterResourceManager.java
Sanil15: will make this cleaner

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -290,6 +290,29 @@ void handleExpiredRequest(String processorId, String preferredHost,
...
+   *
+   * @param request pending request for the preferred host
+   * @param resource resource allocated from {@link ClusterResourceManager} which has expired
+   * @param preferredHost host on which container is requested to be deployed
+   * @param resourceRequestState state of request in {@link ContainerAllocator}
+   * @param allocator allocator for requesting resources
+   */
+  void handleExpiredResource(SamzaResourceRequest request, SamzaResource resource, String preferredHost,
+      ResourceRequestState resourceRequestState, ContainerAllocator allocator) {
+    LOG.info("Allocated resource {} has been expired for Processor ID: {} request: {} re-requesting resource again",
```
prateekm: "has expired"  "{}. Re-requesting again"

samza-yarn/src/main/java/org/apache/samza/job/yarn/YarnClusterResourceManager.java
```diff
@@ -575,6 +576,17 @@ public void onStopContainerError(ContainerId containerId, Throwable t) {
     }
   }
 
+  @Override
+  public boolean isResourceExpired(SamzaResource resource) {
+    /**
```
prateekm: Use `//` for comments inside methods, not `/*`. IntelliJ warns about dangling javadocs for these.

PR #1297, opened 2020-02-28
SAMZA-2379: Support Container Placements for job running in degraded state 

Sanil15:
**Design**: [SEP-22: Container Placements in Samza](https://cwiki.apache.org/confluence/display/SAMZA/SEP-22:+Container+Placements+in+Samza)

**Changes:**: Today if the config "cluster-manager.container.fail.job.after.retries" is set to false that means after exhausting all the retry on the containers, Job Coordinator does not fail the job but keeps running the job without the failed container. In such scenarios, this container can be started back on the host it was last running on or any other host using container placements APIs 

**API Changes:** None

**Usage Instructions:** Instantiate ContainerPlacementMetadataStore to write Container placement messages to Metastore, query it by the UUID generated


Sanil15:
Note: this is stacked PR on (#1281 ), please review changes from this commit: https://github.com/apache/samza/pull/1297/commits/9d8df53c2315b73cb21d099eb95f2bd309d6334d

samza-core/src/test/java/org/apache/samza/clustermanager/MockClusterResourceManager.java
```diff
@@ -102,12 +102,17 @@ public void launchStreamProcessor(SamzaResource resource, CommandBuilder builder
 
   @Override
   public void stopStreamProcessor(SamzaResource resource) {
-    SamzaResourceStatus status = new SamzaResourceStatus(resource.getContainerId(), "diagnostics", SamzaResourceStatus.PREEMPTED);
+    stopStreamProcessor(resource, SamzaResourceStatus.PREEMPTED);
+  }
+
+  void stopStreamProcessor(SamzaResource resource, int exitCode) {
```
rmatharu: Is this required only for testing?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -429,6 +429,25 @@ private void writeContainerPlacementResponseMessage(ContainerPlacementRequestMes
...
+    String processorId = requestMessage.getProcessorId();
+    if (samzaApplicationState.runningProcessors.containsKey(processorId)) {
+      SamzaResource currentResource = samzaApplicationState.runningProcessors.get(processorId);
+      LOG.info("Processor ID: {} matched a running container with containerId ID: {} is running on host: {} for ContainerPlacement action: {}",
+          processorId, currentResource.getContainerId(), currentResource.getHost(), requestMessage);
+      sourceHost = currentResource.getHost();
+    } else {
+      sourceHost = samzaApplicationState.jobModelManager.jobModel().getContainerToHostValue(processorId, SetContainerHostMapping.HOST_KEY);
+      LOG.info("Processor ID: {} is not running and was last seen on host: {} for ContainerPlacement action: {}",
+          processorId, sourceHost, requestMessage);
```
rmatharu: CPM does a similar thing at 
```
String lastSeenOn = state.jobModelManager.jobModel().getContainerToHostValue(processorId, SetContainerHostMapping.HOST_KEY);
```
Is it possible to consolidate these multiple job-model-manager accesses?

maybe call it just getHostForContainer?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -443,6 +462,11 @@ private boolean deQueueAction(ContainerPlacementRequestMessage requestMessage) {
     if (checkIfActiveOrStandbyContainerHasActivePlacementAction(requestMessage)) {
       return false;
     }
+
+    if (samzaApplicationState.failedProcessors.containsKey(requestMessage.getProcessorId())) {
+      return true;
```
rmatharu: Log statements needed?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -464,7 +488,8 @@ private boolean deQueueAction(ContainerPlacementRequestMessage requestMessage) {
     Boolean invalidAction = false;
     String errorMessage = null;
     if (!samzaApplicationState.runningProcessors.containsKey(requestMessage.getProcessorId()) &&
-        !samzaApplicationState.pendingProcessors.containsKey(requestMessage.getProcessorId())
+        !samzaApplicationState.pendingProcessors.containsKey(requestMessage.getProcessorId()) &&
+        !samzaApplicationState.failedProcessors.containsKey(requestMessage.getProcessorId())
```
rmatharu: To make this readable, maybe define variables for each of the lookups and 
define the if to be 
if (inRunning || inPending || inFailed) {//do blah}
else {
this
}

samza-core/src/main/java/org/apache/samza/clustermanager/SamzaApplicationState.java
```diff
@@ -110,7 +110,17 @@
...
-   /**
+  /**
+   *  Map of the failed Samza processor ID to resource status of the last attempted of the container.
+   *  This map is only used when {@link org.apache.samza.config.ClusterManagerConfig#CLUSTER_MANAGER_CONTAINER_FAIL_JOB_AFTER_RETRIES}
+   *  is set to false, this map tracks the containers which have exhausted all retires for restart and JobCoordinator is
+   *  no longer attempting to restart this container
+   *
+   *  Modified by both the AMRMCallbackThread and the ContainerAllocator thread
+   */
+  public final ConcurrentHashMap<String, SamzaResourceStatus> failedProcessors = new ConcurrentHashMap<>(0);
```
rmatharu: This class already has a map called 
```
/**
   * ContainerStatuses of failed containers.	   * ContainerStatuses of failed containers.
   */	 
  public final ConcurrentMap<String, SamzaResourceStatus> failedContainersStatus = new ConcurrentHashMap<>();	 
```

Could we please reuse that?

Also the variable type can just be a Map<,> doesnt need to be ConcurrentHashMap

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
rmatharu: maybe link jira for that here?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -486,9 +512,12 @@ private boolean deQueueAction(ContainerPlacementRequestMessage requestMessage) {
...
-    if (!samzaApplicationState.runningProcessors.containsKey(requestMessage.getProcessorId()) &&
-        !samzaApplicationState.pendingProcessors.containsKey(requestMessage.getProcessorId())
-    ) {
+
+    boolean isRunning = samzaApplicationState.runningProcessors.containsKey(requestMessage.getProcessorId());
+    boolean isPending = samzaApplicationState.pendingProcessors.containsKey(requestMessage.getProcessorId());
+    boolean isFailed = samzaApplicationState.failedProcessors.containsKey(requestMessage.getProcessorId());
+
+    if (!isRunning && !isPending && !isFailed) {
       errorMessage = String.format(errorMessagePrefix, "invalid processor id neither in running or pending processors");
```
rmatharu: "neither in running, pending or failed state"

samza-core/src/main/java/org/apache/samza/clustermanager/SamzaApplicationState.java
rmatharu: As long as we use one of the two. 
Its also used by YarnClusterResourceManager for updating "completed" containers, 
which are containers that exit after finishing gracefully.

samza-core/src/test/java/org/apache/samza/clustermanager/MockClusterResourceManager.java
rmatharu: Maybe @visible for testing or 
// package private for testing?

samza-core/src/test/java/org/apache/samza/clustermanager/TestContainerPlacementActions.java
```diff
@@ -528,6 +528,124 @@ public Void answer(InvocationOnMock invocation) {
     assertFalse(containerPlacementMetadataStore.readContainerPlacementRequestMessage(requestMessage.getUuid()).isPresent());
   }
 
+
+  @Test(timeout = 20000)
```
rmatharu: This test is littered with thread.sleep and timeouts -- a typical recipe for flaky jarvis/test, and increasing build times.
Would it be possible to reduce/consolidate them?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
Sanil15: sure

samza-core/src/main/java/org/apache/samza/clustermanager/SamzaApplicationState.java
Sanil15: thats YarnAppState not SamzaApplicationState

samza-core/src/test/java/org/apache/samza/clustermanager/MockClusterResourceManager.java
Sanil15: this class is a Mock itself but sure Ill add it

samza-core/src/test/java/org/apache/samza/clustermanager/TestContainerPlacementActions.java
Sanil15: I have used semaphores where it was easy to inject (awaiting container launch), this timeout configured on tests is generous enough, each callback spawns a different thread to mimic AMRMClientAsync behavior, so if I were to do all of these with using semaphores instead of thread.sleep for state change, I would need some significant refactors

Sanil15: These are integ tests and it takes 35 seconds to run all of them serially, we can also make them run in parallel to further reduce build times

mynameborat:
Committing the patch based on - https://travis-ci.org/Sanil15/samza/builds/660300111

PR #1298, opened 2020-02-28
SAMZA-2476:Fix flaky tests in TestAzureBlobOutputStream

lakshmi-manasa-g:
Symptom:
    Travis builds occasionally fail with org.mockito.exceptions.misusing.UnfinishedStubbingException for org.apache.samza.system.azureblob.avro.TestAzureBlobOutputStream.testCloseMultipleBlocks. 

Cause: 
Seems to be due to using PowerMockito.doAnswer for mocking com.azure.storage.blob.specialized.BlockBlobAsyncClient.commitBlockListWithResponse

Changes:
pull call to commitBlockListWithResponse into a package private method of AzureBlobOutputStream and verify that method.

Tests: 
Rerun of TestAzureBlobOutputStream in original code cause 240 failures in 10K runs
Rerun of TestAzureBlobOutputStream with this fix causes 0 failures in 10K runs


samza-azure/src/main/java/org/apache/samza/system/azureblob/avro/AzureBlobOutputStream.java
```diff
@@ -227,6 +227,11 @@ public synchronized void releaseBuffer() throws IOException {
     this.compression = compression;
   }
 
+  @VisibleForTesting
```
rmatharu: maybe simple javadoc?

PR #1301, opened 2020-03-04
Samza-2478: Add new metrics to track key and value size of records written to RocksDb

xiefan46:
**Feature:** Add new metrics to track key and value size of records written to RocksDb. 

**Changes:** Add three new metrics **recordKeySizeBytes** , **recordValueSizeBytes** and **maxRecordKeySizeBytes** in **SerializedKeyValueStoreMetrics** class to record the percentiles and maximum key and value sizes written to RocksDB. 

**Tests:** Added a unit test call **TestKeyValueSizeHistogramMetric** to verify that those metrics are written to SamzaHistogram correctly and can be access by MetricVisitor.  Also verified that those metrics can be displayed correctly by deploying a Samza job in a private cluster and drawing those metrics into graphs. 

**API Changes:** None
**Upgrade instructions:** 
**Usage instructions:** Users can monitor P50, P90, P99, P99.9 percentiles as well as the maximum value of key and value sizes written to RocksDB via MetricReporter. Please refer to the metrics reference sheet for details on the metric.


samza-test/src/test/scala/org/apache/samza/storage/kv/TestKeyValueStores.scala
```diff
@@ -555,6 +555,7 @@ class TestKeyValueStores(typeOfStore: String, storeConfig: String) {
 
     store.flush()
   }
+
```
rmatharu: Try and remove/avoid such changes.

samza-test/src/test/java/org/apache/samza/storage/kv/TestKeyValueSizeHistogramMetric.java
```diff
@@ -0,0 +1,107 @@
...
+  @Before
+  public void setup() {
+    KeyValueStore<byte[], byte[]> kvStore = new InMemoryKeyValueStore(keyValueStoreMetrics);
+    KeyValueStore<String, String> serializedStore =
+        new SerializedKeyValueStore<>(kvStore, stringSerde, stringSerde, serializedKeyValueStoreMetrics);
+    store = new NullSafeKeyValueStore<>(serializedStore);
+  }
+
+  @Test
+  public void testHistogramMetric() {
```
rmatharu: Maybe add comments to describe what this test is doing?

samza-kv/src/main/scala/org/apache/samza/storage/kv/SerializedKeyValueStoreMetrics.scala
```diff
@@ -36,6 +35,10 @@ class SerializedKeyValueStoreMetrics(
   val bytesSerialized = newCounter("bytes-serialized")
   val bytesDeserialized = newCounter("bytes-deserialized")
   val maxRecordSizeBytes = newGauge("max-record-size-bytes", 0L)
+  val record_key_size_percentiles = java.util.Arrays.asList[java.lang.Double](10D, 50D, 90D, 99D)
+  val record_value_size_percentiles = java.util.Arrays.asList[java.lang.Double](10D, 50D, 90D, 99D)
+  val recordKeySizeBytes = newHistogram("key-size-bytes-histogram",record_key_size_percentiles )
+  val recordValueSizeBytes = newHistogram("value-size-bytes-histogram", record_value_size_percentiles)
```
rmatharu: Maybe add javadocs to explain how this works, the histogram stores counts for each percentile range? 
and the count for each range is emitted as a gauge?

samza-core/src/main/java/org/apache/samza/metrics/MetricGroup.java
```diff
@@ -65,4 +68,8 @@ public T getValue() {
   public Timer newTimer(String name) {
     return registry.newTimer(groupName, (prefix + name).toLowerCase());
   }
+
```
rmatharu: javadoc?

samza-api/src/main/java/org/apache/samza/metrics/SamzaHistogram.java
```diff
@@ -60,6 +60,11 @@ public void updateGaugeValues(double percentile) {
     gauges.get(percentile).set(values.getValue(percentile / 100));
   }
 
+  //for debug
+  public Map<Double, Gauge<Double>> getGauges() {
+    return gauges;
+  }
```
rmatharu: Is this only for testing?
In that case you can make it package-private and use @visiblefortesting flag.

xiefan46: I will modify it and use a MetricsVisitor to get access to the Histogram instead of adding a getter method.

xiefan46: I deleted this method and used a MetricsVisitor instead.

samza-kv/src/main/scala/org/apache/samza/storage/kv/SerializedKeyValueStore.scala
```diff
@@ -62,7 +62,10 @@ class SerializedKeyValueStore[K, V](
     val keyBytes = toBytesOrNull(key, keySerde)
     val valBytes = toBytesOrNull(value, msgSerde)
     store.put(keyBytes, valBytes)
+    val keySizeBytes = if (keyBytes == null) 0 else keyBytes.length
     val valSizeBytes = if (valBytes == null) 0 else valBytes.length
+    metrics.recordKeySizeBytes.update(keySizeBytes)
+    metrics.recordValueSizeBytes.update(valSizeBytes)
     updatePutMetrics(1, valSizeBytes)
```
bkonold: Is there a way we can roll together recording maximum record size into the histogram tracking? When I implemented the metric for recording max I had considered modifying histogram to record maximum in addition to percentiles, but decided against it since the memory footprint of the histogram was unjustifiable at the time. Now that we'd like those metrics anyway, maybe it is justifiable now.

samza-kv/src/main/scala/org/apache/samza/storage/kv/SerializedKeyValueStoreMetrics.scala
```diff
@@ -36,6 +35,25 @@ class SerializedKeyValueStoreMetrics(
   val bytesSerialized = newCounter("bytes-serialized")
   val bytesDeserialized = newCounter("bytes-deserialized")
   val maxRecordSizeBytes = newGauge("max-record-size-bytes", 0L)
+  val record_key_size_percentiles = java.util.Arrays.asList[java.lang.Double](10D, 50D, 90D, 99D)
```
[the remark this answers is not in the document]
rmatharu: +1, 
50, 90, 99 maybe good enough? or 999 in addition perhaps.

xiefan46:
**Feature:** Add new metrics to track key and value size of records written to RocksDb

**Changes:** Add two SamzaHistogram metrics **recordKeySizeBytes** and **recordValueSizeBytes** in **SerializedKeyValueStoreMetrics** class

**Tests:** Added a unit test 

**API Changes:** None
**Usage instructions:** Every time a record is written to KeyValue store via SerializedKeyValueStore, its size will be recorded in those histograms. A metrics reporter can use MetricVisitor to get access to this two metrics.

prateekm:
Thanks for the updated desciption @xiefan46! Minor suggestion: the "Usage instructions" section is meant for Samza users, not samza devs. I.e., we should document what users need to do to leverage this new feature. For example, for this feature you might want to clarify what new metrics can users look at?  

Similarly, let's update the documentation with the new metrics as well. @rmatharu knows where that is.

samza-test/src/test/java/org/apache/samza/storage/kv/TestKeyValueSizeHistogramMetric.java
```diff
@@ -0,0 +1,129 @@
...
+   * Make sure that the histograms can record the key value size and we can use a
+   * {@link MetricsVisitor} to get access to the value store in the histograms
+   */
+  @Test
+  public void testHistogramMetric() {
+
+    List<String> keys = new ArrayList<>();
+    List<String> values = new ArrayList<>();
+
+    for (int i = 0; i < 1000; i++) {
```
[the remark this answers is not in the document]
xiefan46: Use 300 instead of 1000 in this for loop and also reduce the max length of random string to 100. This test takes less than 1s now.

samza-test/src/test/java/org/apache/samza/storage/kv/TestKeyValueSizeHistogramMetric.java
```diff
@@ -0,0 +1,129 @@
...
+
+    Collections.shuffle(keys);
+    Collections.shuffle(values);
+
+    for (int i = 0; i < keys.size(); i++) {
+      store.put(keys.get(i), values.get(i));
+    }
+
+    metricsRegistry.getGroups().forEach(group -> metricsRegistry.getGroup(group.toString()).forEach((name, metric) -> {
+        if (name.contains("size-bytes-histogram")) {
```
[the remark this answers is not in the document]
xiefan46: Modified

samza-test/src/test/java/org/apache/samza/storage/kv/TestKeyValueSizeHistogramMetric.java
```diff
@@ -0,0 +1,129 @@
...
+    List<String> values = new ArrayList<>();
+
+    for (int i = 0; i < 1000; i++) {
+      keys.add(getRandomString());
+      values.add(getRandomString());
+    }
+
+    Collections.shuffle(keys);
+    Collections.shuffle(values);
+
```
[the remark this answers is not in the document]
xiefan46: At the beginning I didn't use a random way to generate keys and values so I had to shuffle them. I forgot to delete the shuffle code later on. I modified the test case and deleted them now.

samza-kv/src/main/scala/org/apache/samza/storage/kv/SerializedKeyValueStoreMetrics.scala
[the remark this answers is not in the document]
xiefan46: Removed P10 and added 999 instead.

samza-kv/src/main/scala/org/apache/samza/storage/kv/SerializedKeyValueStore.scala
xiefan46: I thought about that before. But maybe there are some people using the maximum record size metric and deleting it may caused some trouble. So I choose to keep this metric and do not combine it with the histogram.

bkonold: I see. We wouldn't be able to preserve the name if we emitted a maximum metric from the histogram?

xiefan46: I am not very sure about that because I am not very familiar with the code yet. I will keep that in mind and try to optimize it in the future if possible.

samza-test/src/test/java/org/apache/samza/storage/kv/TestKeyValueSizeHistogramMetric.java
```diff
@@ -0,0 +1,153 @@
...
+      names.add(storeName + "-" + keyPrefix + "_" + p);
+    }
+
+    for (Double p : serializedKeyValueStoreMetrics.record_value_size_percentiles()) {
+      names.add(storeName + "-" + valuePrefix + "_" + p);
+    }
+
+    metricsRegistry.getGroups().forEach(group -> metricsRegistry.getGroup(group.toString()).forEach((name, metric) -> {
+        if (names.contains(name)) {
+          //System.out.println(name);
```
mynameborat: nit: remove

samza-core/src/main/java/org/apache/samza/metrics/MetricGroup.java
mynameborat: +1

PR #1309, opened 2020-03-10
SAMZA-2453: Update ClusterBasedJobCoordinator to support Beam jobs

kw2542:
Design:

Changes:

1. Invoke app.main.class with app.main.args when present instead of invoking ClusterBasedJobCoordinator#run

API Changes:

N/A

Upgrade Instructions:

For beam jobs, beam's main class and pipeline options need to passed with app.main.class and app.main.args configs.

Usage Instructions:

Take https://github.com/apache/samza-beam-examples as an example:

1. run-beam-yarn.sh#line 36 will be updated to

exec $(dirname $0)/run-class.sh $1 --config app.main.class=$1 --config app.main.args="--runner=org.apache.beam.runners.samza.SamzaRunner --configOverride=\"$override\" \"${@:2}\""

Tests:
Unit tests


samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -551,8 +553,23 @@ private static void runClusterBasedJobCoordinator(String[] args) {
         throw new SamzaException(e);
       }
 
-      ClusterBasedJobCoordinator jc = createFromConfigLoader(submissionConfig);
-      jc.run();
+      ApplicationConfig appConfig = new ApplicationConfig(submissionConfig);
+
+      if (appConfig.getAppMainClass().isPresent()) {
```
xinyuiscool: Please add some comments about the use cases here for Beam. what's the expected main/args being passed in and what are we going to do with them.

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
```diff
@@ -634,4 +651,20 @@ static ClusterBasedJobCoordinator createFromConfigLoader(Config submissionConfig
...
   }
+
+  @VisibleForTesting
+  static String[] toArgs(ApplicationConfig config) {
+    List<String> args = new ArrayList<>(config.size() * 2);
+
+    config.forEach((key, value) -> {
+      if (key.equals(ApplicationConfig.APP_MAIN_ARGS)) {
+        args.addAll(Arrays.asList(value.split("\\s")));
+      } else {
```
xinyuiscool: is there else part being used right now?

kw2542: Yes,

if block is to pass native Beam pipeline options such as 
--runner SamzaRunner --maxSourceParallelism=1024

else block is to change other native samza configs with --config key=value, such as 

--config job.name=beam-test

samza-core/src/main/java/org/apache/samza/clustermanager/ClusterBasedJobCoordinator.java
kw2542: Added

PR #1314, opened 2020-03-12
SAMZA-2484: Fix flaky tests for Container Placements

Sanil15:
**Symptom:**
Noticed a few runs being flaky with:

org.apache.samza.clustermanager.TestContainerPlacementActions > testAlwaysMoveToAnyHostForHostAffinityDisabled FAILED
    java.lang.AssertionError: expected:<IN_PROGRESS> but was:<SUCCEEDED>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:118)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at org.apache.samza.clustermanager.TestContainerPlacementActions.testAlwaysMoveToAnyHostForHostAffinityDisabled(TestContainerPlacementActions.java:746)
 
**Cause:**
A race between main thread expects to read a success message from metastore and am-rm thread to write a success messages  

Changes:
Make the main thread busy loop on metastore success message, since a semaphore is hard to inject

### DIFF
--- a/samza-core/src/test/java/org/apache/samza/clustermanager/TestContainerPlacementActions.java
+++ b/samza-core/src/test/java/org/apache/samza/clustermanager/TestContainerPlacementActions.java
@@ -94,7 +94,7 @@ public class TestContainerPlacementActions {
   private CoordinatorStreamStore coordinatorStreamStore;
   private ContainerPlacementMetadataStore containerPlacementMetadataStore;
 
-  private SamzaApplicationState state;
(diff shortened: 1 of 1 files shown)


rmatharu:
Is there a way to write such tests based on invariants, so that they dont rely on thread-sleeps and fail in case of scheduling delays?

PR #1318, opened 2020-03-18
SAMZA-2488: Add JobCoordinatorLaunchUtil to handle common logic when launching job coordiantor.

kw2542:
Issues: High/Low and Beam are sharing the same logic when launching ClusterBasedJobCoordinator and it is ideal to have them in single place.

Changes: Add JobCoordinatorLaunchUtil which contains the common logic when launching ClusterBasedJobCoordinator.

Tests: Unit tests + ./gradlew build

API Changes: None
Upgrade Instructions: None
Usage Instructions: None


samza-core/src/main/java/org/apache/samza/clustermanager/JobCoordinatorLaunchUtil.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.samza.config.JobConfig;
+import org.apache.samza.coordinator.metadatastore.CoordinatorStreamStore;
+import org.apache.samza.execution.RemoteJobPlanner;
+import org.apache.samza.metadatastore.MetadataStore;
+import org.apache.samza.metrics.MetricsRegistryMap;
+import org.apache.samza.util.CoordinatorStreamUtil;
+import org.apache.samza.util.DiagnosticsUtil;
+
+
+public class JobCoordinatorLaunchUtil {
```
xinyuiscool: Please add javadoc to this class, including a note about this will be invoked in both Samza api as well as beam api, so any change needs to be tested in both.

samza-core/src/main/java/org/apache/samza/clustermanager/JobCoordinatorLaunchUtil.java
```diff
@@ -0,0 +1,74 @@
...
+
+    Config finalConfig = jobConfigs.get(0);
+
+    // This needs to be consistent with RemoteApplicationRunner#run where JobRunner#submit to be called instead of JobRunner#run
+    CoordinatorStreamUtil.writeConfigToCoordinatorStream(finalConfig, true);
+    DiagnosticsUtil.createDiagnosticsStream(finalConfig);
+    MetricsRegistryMap metrics = new MetricsRegistryMap();
+    MetadataStore
+        metadataStore = new CoordinatorStreamStore(CoordinatorStreamUtil.buildCoordinatorStreamConfig(finalConfig), metrics);
+    metadataStore.init();
```
xinyuiscool: Could you make a comment about where the metasdatastore will be closed?

samza-core/src/main/java/org/apache/samza/clustermanager/JobCoordinatorLaunchUtil.java
kw2542: Updated

samza-core/src/main/java/org/apache/samza/clustermanager/JobCoordinatorLaunchUtil.java
kw2542: Updated

PR #1320, opened 2020-03-18
SAMZA-2490: AsyncFlatmapOperatorImpl#handleClose should call transformFn.close()

Zhangyx39:
Symptom: In beam samza runner, Op#close is not invoked during shutdown, which is not expected.

Cause: AsyncFlatmapOperatorImpl#handleClose() doesn't call transformFn.close(). Therefore the underlying transform function will never be closed.

Changes: Just invoke transformFn.close inside handleClose method should be enough.

Tests: Tested with a samza beam job. After the change, Op#close is able to function properly.


xinyuiscool:
Thanks for finding this bug. Can we add a unit test to prevent this happening again?

PR #1326, opened 2020-03-23
SAMZA-2492: Add new deserialize function for JobGraphJson and change the scope of related classes as public

alnzng:
#### Issues
In Samza core, `JobGraphJsonGenerator` provides the function to serialize `JobGraphJson` as plan JSON and later the config `samza.internal.execution.plan` will use this string as value.

However, it doesn't provide the deserialize function to help generate `JobGraphJson` from plan JSON string. This function is useful when you need to parse information from the config `samza.internal.execution.plan`.

#### Changes
In this PR, there are two changes made for supporting `JobGraphJson` deserialization:

1. Add new deserialize function `toJobGraphJson()`
1. Change the package scope of related classes as `public`

#### Tests

#### API Changes
* Change package scope as `public` for the following classes:
   * `JobGraphJsonGenerator` class
   * All inner classes inside `JobGraphJsonGenerator`.
* Provide new deserialize function `toJobGraphJson `.

#### Upgrade Instructions
None

#### Usage Instructions
None

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
+++ b/samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
@@ -39,7 +39,6 @@ import org.apache.samza.operators.spec.OutputStreamImpl;
 import org.apache.samza.operators.spec.PartitionByOperatorSpec;
 import org.apache.samza.operators.spec.SendToTableOperatorSpec;
 import org.apache.samza.operators.spec.StreamTableJoinOperatorSpec;
-import org.apache.samza.system.StreamSpec;
 import org.apache.samza.table.descriptors.BaseTableDescriptor;
 import org.apache.samza.table.descriptors.TableDescriptor;
 import org.codehaus.jackson.annotate.JsonProperty;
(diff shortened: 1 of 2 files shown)


alnzng:
@MabelYC Seems you are looking for the same function provided in this PR, can you please help take a look at it?

samza-core/src/main/java/org/apache/samza/execution/JobGraphJsonGenerator.java
```diff
@@ -44,9 +45,9 @@
 /**
  * This class generates the JSON representation of the {@link JobGraph}.
  */
-/* package private */ class JobGraphJsonGenerator {
+public class JobGraphJsonGenerator {
```
Sanil15, line 48: Since we are making them public, lets add some java docs for all the Pojo's and there use

alnzng: Thanks for the review! Added the java docs for those inner classes.

samza-core/src/test/java/org/apache/samza/execution/TestJobGraphJsonGenerator.java
```diff
@@ -368,6 +368,31 @@ public void testLegacyTaskApplication() throws Exception {
     assertEquals(0, jsonObject.jobs.size());
   }
 
+  @Test
+  public void testToJobGraphJson() throws Exception {
```
PanTheMan, line 372: Could we also add more tests to test the other changed classes like JobNodeJson? The test here only tests JobGraphJson

alnzng: Thanks for the review @PanTheMan . Had the discussion with @cameronlee314 , we need to figure out one better to way to expose `JobGraphJson`. So I will close this ticket for now.

PR #1327, opened 2020-03-25
SAMZA-2493: Keep checkpoint manager consumer open for repeated polling

bkonold:
**Feature:** Allow control of whether the underlying consumer backing a checkpoint manager will be left open (until stopped), or closed after the initial read of checkpoints
 
**Changes:** Introduces an internal config, injected by SamzaContainer, that is injected dependent on whether the container is standby or not, to control whether the container's checkpoint manager should have it's underlying consumer stopped after initial read or not.
 
 **Tests:** Added a unit test each for the config set to true / false

**API Changes:** None
**Usage Instructions:** None

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
+++ b/samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
@@ -105,7 +105,7 @@ public class TaskConfig extends MapConfig {
   private static final String BROADCAST_STREAM_PATTERN = "^[\\d]+$";
   private static final String BROADCAST_STREAM_RANGE_PATTERN = "^\\[[\\d]+\\-[\\d]+\\]$";
   public static final String CHECKPOINT_MANAGER_FACTORY = "task.checkpoint.factory";
-  public static final String CHECKPOINT_MANAGER_CONSUMER_STOP_AFTER_FIRST_READ = "task.checkpoint.consumer.stop.after.first.read";
+  public static final String INTERNAL_CHECKPOINT_MANAGER_CONSUMER_STOP_AFTER_FIRST_READ = "samza.internal.task.checkpoint.consumer.stop.after.first.read";
 
   public static final String TRANSACTIONAL_STATE_CHECKPOINT_ENABLED = "task.transactional.state.checkpoint.enabled";
   private static final boolean DEFAULT_TRANSACTIONAL_STATE_CHECKPOINT_ENABLED = true;
@@ -215,11 +215,11 @@ public class TaskConfig extends MapConfig {
   }
 
   /**
-   * Internal config to indicate whether the SystemConsumer underlying a CheckpointManager should be left open after
+   * Internal config to indicate whether the SystemConsumer underlying a CheckpointManager should be stopped after
    * initial read of checkpoints.
    */
   public boolean getCheckpointManagerConsumerStopAfterFirstRead() {
-    return getBoolean(CHECKPOINT_MANAGER_CONSUMER_STOP_AFTER_FIRST_READ, true);
+    return getBoolean(INTERNAL_CHECKPOINT_MANAGER_CONSUMER_STOP_AFTER_FIRST_READ, true);
   }
 
   /**
--- a/samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
+++ b/samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
@@ -105,6 +105,7 @@ public class TaskConfig extends MapConfig {
   private static final String BROADCAST_STREAM_PATTERN = "^[\\d]+$";
   private static final String BROADCAST_STREAM_RANGE_PATTERN = "^\\[[\\d]+\\-[\\d]+\\]$";
   public static final String CHECKPOINT_MANAGER_FACTORY = "task.checkpoint.factory";
+  public static final String CHECKPOINT_MANAGER_CONSUMER_STOP_AFTER_FIRST_READ = "task.checkpoint.consumer.stop.after.first.read";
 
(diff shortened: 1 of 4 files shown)


bkonold:
@bharathkk @xinyuiscool can you take a look over this?

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
```diff
@@ -105,6 +105,7 @@
   private static final String BROADCAST_STREAM_PATTERN = "^[\\d]+$";
   private static final String BROADCAST_STREAM_RANGE_PATTERN = "^\\[[\\d]+\\-[\\d]+\\]$";
   public static final String CHECKPOINT_MANAGER_FACTORY = "task.checkpoint.factory";
+  public static final String CHECKPOINT_MANAGER_CONSUMER_STOP_AFTER_FIRST_READ = "task.checkpoint.consumer.stop.after.first.read";
```
prateekm: Why do this as a config instead of a parameter?
If internal config, let's use samza.internal prefix.

bkonold: @prateekm The rationale between @mynameborat and I is that the internal config was a lighter-weight change.

Passing this via construction parameter would mean changing the `CheckpointManagerFactory` API and underlying implementations. If we're ok with that, I'm open to either approach.

mynameborat: +1 to internal prefix. We were debating between passing the exposing task mode through parameter vs config. The config route seemed less invasive and also doesn't introduce breaking API changes. 
IIRC, we want to know if the container is standby or not. Given KafkaCheckpointManager is a singleton for the container, is there a way to infer on the fly from the existing configurations if it is standby or active? for e.g. task.inputs is empty?

bkonold: I spoke with @rmatharu and there is nothing in config we can use to infer standby vs not. If we're all ok with the internal config, I will change the prefix and we can go with that.

May also be worth pointing out that this PR doesn't do much to lock us into anything - if we later want to treat this information as more a first class citizen we can go ahead and change the factory API.

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
```diff
@@ -105,6 +105,7 @@
   private static final String BROADCAST_STREAM_PATTERN = "^[\\d]+$";
   private static final String BROADCAST_STREAM_RANGE_PATTERN = "^\\[[\\d]+\\-[\\d]+\\]$";
   public static final String CHECKPOINT_MANAGER_FACTORY = "task.checkpoint.factory";
+  public static final String INTERNAL_CHECKPOINT_MANAGER_CONSUMER_STOP_AFTER_FIRST_READ = "samza.internal.task.checkpoint.consumer.stop.after.first.read";
```
rmatharu, line 108: Lets maybe add a one-line description of what this config does?
Also if it'd be possible to shrink the name somewhat.

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
rmatharu: Is adding a parameter like JobModel or containermodel, etc to CheckpointManagerFactory  a big deal?

mynameborat: I agree. For now, the config approach is minimally invasive and backward compatible. We can always reevaluate this if there is a stronger justification to have container model exposed to the checkpoint manager. 
Since it is not a user config, we should be able to deprecate it without much effort.

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
```diff
@@ -76,6 +76,8 @@ class KafkaCheckpointManager(checkpointSpec: KafkaStreamSpec,
   val producerRef: AtomicReference[SystemProducer] = new AtomicReference[SystemProducer](getSystemProducer())
   val producerCreationLock: Object = new Object
 
+  val stopConsumerAfterFirstRead: Boolean = new TaskConfig(config).getCheckpointManagerConsumerStopAfterFirstRead
```
mynameborat, line 79: can we put in a comment on when this config is set vs not?

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
rmatharu: makes sense

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
bkonold: I initially thought of something like `samza.internal.task.checkpoint.consumer.keep.open` but thought that the name in the PR currently is more expressive of behavior.

Thoughts?

rmatharu: sure lets go with that one then

samza-kafka/src/main/scala/org/apache/samza/checkpoint/kafka/KafkaCheckpointManager.scala
bkonold: Done

samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
bkonold: Closing this thread then. Thanks.

mynameborat:
Do you plan to add tests? I noticed TBD in the description.

bkonold:
> Do you plan to add tests? I noticed TBD in the description.

@mynameborat yeah, added them

PR #1332, opened 2020-03-26
SAMZA-2442: Update doc/tutorials from config factory to config loader

kw2542:
Update documentations and left over scripts from config factory to config loader.
Tests:
1. Test samza rest following instructions on https://samza.apache.org/learn/documentation/latest/rest/overview.html

2. 
API Changes:
None
Upgrade Instructions:
None
Usage Instructions:
None


samza-test/src/main/python/deployment.py
```diff
@@ -76,7 +76,7 @@ def setup_suite():
 
   # Setup Samza job deployer.
   samza_job_deployer = SamzaJobYarnDeployer({
-    'config_factory': c('samza_config_factory'),
+    'config_factory': c('samza_config_loader_factory'),
```
cameronlee314, line 79: Does the key here need to be `config_loader_factory`? You changed `samza_job_yarn_deployer.py` to use `config_loader_factory`.

samza-rest/src/main/java/org/apache/samza/rest/proxy/job/SimpleYarnJobProxy.java
```diff
@@ -40,8 +40,8 @@
   private static final String START_SCRIPT_NAME = "run-job.sh";
```
cameronlee314, line 40: Does this need to use `run-app.sh`? It looks like you changed everything else to `run-app.sh`. I'm not sure what this is used for, but would this also need testing?

kw2542: Updated to run-app.sh as well. 

Tested with `curl -X PUT 'localhost:9139/v1/jobs/wikipedia-feed/1?status=start'`

samza-test/src/main/python/deployment.py
kw2542: Nice catch

kw2542:
> Can you please update the PR description with the testing that you did? For example, please confirm that the integration tests passed, and you mentioned you did something to test `SimpleYarnJobProxy`.
> I will merge the PR once that is done.

I am seeing partial integration tests now, let me fix them.

PR #1333, opened 2020-03-27
SAMZA-2497: Close MetadataResourceUtil when job is stopped

PanTheMan:
**Symptoms**: Error message about a Kafka consumer allocated but not closed appears in our log messages
**Cause**: The MetadataResouceUtil has a Kafka checkpoint manager that is never closed in our code. A LifeCycleAwareConsumer sees this and complains about it.
**Fix:**: Add a stop in MetadataResourceUtil that will close the checkpoint manager in it. Then have MetadataResourceUtil.stop called everywhere it's used when the job is stopped
**Tests**: Any job that uses ProcessJob or PassthroughJobCoordinator will see errors in the log files about a Kafka consumer allocated but not closed. With this fix in, that error shouldn't appear

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
+++ b/samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
@@ -114,6 +114,7 @@ public class ZkJobCoordinator implements JobCoordinator {
   private JobModel newJobModel;
   private boolean hasLoadedMetadataResources = false;
   private String cachedJobModelVersion = null;
+  private MetadataResourceUtil metadataResourceUtil;
 
   @VisibleForTesting
   ZkSessionMetrics zkSessionMetrics;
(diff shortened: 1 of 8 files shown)


PanTheMan:
@xinyuiscool if you could review this rb

samza-core/src/main/java/org/apache/samza/zk/ZkJobCoordinator.java
```diff
@@ -312,7 +313,7 @@ JobModel readJobModelFromMetadataStore(String zkJobModelVersion) {
   @VisibleForTesting
   void loadMetadataResources(JobModel jobModel) {
     try {
-      MetadataResourceUtil metadataResourceUtil = createMetadataResourceUtil(jobModel, config);
+      metadataResourceUtil = createMetadataResourceUtil(jobModel, config);
```
Sanil15, line 316: Where is `metadataResourceUtil#stop` is being stopped for ZkJobCoordinator?

PR #1335, opened 2020-03-31
SAMZA-2501 : Optimizing startpoint manager to not make successive bootstrapMessage calls to coordinator-store

rmatharu:
Symptom: Currently the startpoint manager queries metastore for task-ssps and for each ssp queries coordinator-store, issuing a re-read on the coordinator store. This causes increased AM startup time and depending on the number of input SSPs to a job (thousands in case of regex), YARN may timeout the AM.
Cause: Above.
Fix: This change reduces to number of metastore reads to 2; one for startpoints keyed by ssp and the other for startpoints keyed by ssp+taskname.
API changes: None
Upgrade Instructions: None

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
+++ b/samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
@@ -127,7 +127,6 @@ public class ClusterManagerConfig extends MapConfig {
    */
   private static final String AM_JMX_ENABLED = "yarn.am.jmx.enabled";
   private static final String CLUSTER_MANAGER_JMX_ENABLED = "cluster-manager.jobcoordinator.jmx.enabled";
-  private static final String CLUSTER_MANAGER_STARTPOINT_FANOUT_ENABLED = "job.startpoint.fanout.enabled";
 
   public ClusterManagerConfig(Config config) {
       super(config);
@@ -263,9 +262,5 @@ public class ClusterManagerConfig extends MapConfig {
     } else {
       return true;
     }
-  }
-
-  public boolean getStartpointFanoutEnabled() {
-    return getBoolean(CLUSTER_MANAGER_STARTPOINT_FANOUT_ENABLED, true);
   }
 }
--- a/samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
+++ b/samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
@@ -127,7 +127,7 @@ public class ClusterManagerConfig extends MapConfig {
    */
   private static final String AM_JMX_ENABLED = "yarn.am.jmx.enabled";
   private static final String CLUSTER_MANAGER_JMX_ENABLED = "cluster-manager.jobcoordinator.jmx.enabled";
-  private static final String CLUSTER_MANAGER_STARTPOINT_FANOUT_DISABLED = "job.startpoint.fanout.disabled";
+  private static final String CLUSTER_MANAGER_STARTPOINT_FANOUT_ENABLED = "job.startpoint.fanout.enabled";
 
   public ClusterManagerConfig(Config config) {
       super(config);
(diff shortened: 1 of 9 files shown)


samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -158,8 +158,25 @@ public void writeStartpoint(SystemStreamPartition ssp, TaskName taskName, Startp
    * @return {@link Optional} of {@link Startpoint} for the {@link SystemStreamPartition}.
    *         It is empty if it does not exist or if it is too stale.
    */
+  @VisibleForTesting
   public Optional<Startpoint> readStartpoint(SystemStreamPartition ssp) {
-    return readStartpoint(ssp, null);
```
Sanil15, line 162: Revert the change on line line 162 calling `readStartpoint(ssp, null);` will still do the intended, do not need line 163-165

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
```diff
@@ -158,8 +158,25 @@ public void writeStartpoint(SystemStreamPartition ssp, TaskName taskName, Startp
...
+   * Returns the last {@link Startpoint} that defines the start position for a {@link SystemStreamPartition} and {@link TaskName}.
+   * @param ssp The {@link SystemStreamPartition} to fetch the {@link Startpoint} for.
+   * @param taskName the {@link TaskName} to fetch the {@link Startpoint} for.
+   * @return {@link Optional} of {@link Startpoint} for the {@link SystemStreamPartition}.
+   *         It is empty if it does not exist or if it is too stale.
+   */
+  @VisibleForTesting
+  public Optional<Startpoint> readStartpoint(SystemStreamPartition ssp, TaskName taskName) {
+    Map<String, byte[]> startpointBytes = readWriteStore.all();
+    // there is no task-name to use as key for the startpoint in this case (only the ssp), so we use a null task-name
```
Sanil15, line 178: Please correct the comment, task-name can non-null

rmatharu:
+1, simplified and addressed both comments.

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
```diff
@@ -127,6 +127,7 @@
    */
   private static final String AM_JMX_ENABLED = "yarn.am.jmx.enabled";
   private static final String CLUSTER_MANAGER_JMX_ENABLED = "cluster-manager.jobcoordinator.jmx.enabled";
+  private static final String CLUSTER_MANAGER_STARTPOINT_FANOUT_ENABLED = "job.startpoint.fanout.enabled";
```
Sanil15, line 130: If this config is called `job.startpoint.fanout.enabled`, isn't it better to be a part of JobConfig than ClusterManagerConfig?

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
rmatharu: Is used in numerous tests, so need it.

samza-core/src/main/java/org/apache/samza/config/ClusterManagerConfig.java
Sanil15: Once you decide where it fits, since this is user-facing, please add docs to doc markdown. Thanks!

samza-core/src/main/java/org/apache/samza/startpoint/StartpointManager.java
Sanil15: You are not changing the function signature, so the test should still work when you invoke `readStartpoint(ssp, null)` from here what am I missing?

rmatharu: The fact that null as task-name maps to "only-ssp" is an internal of the class, i.e., the unit. 
Unit-tests test the unit not internal nitty gritties of the unit.

PR #1337, opened 2020-04-01
SAMZA-2498: [Job coordinator isolation] Classloader isolation for SamzaApplication.describe execution in job coordinator

cameronlee314:
Feature: https://cwiki.apache.org/confluence/display/SAMZA/SEP-24%3A+Cluster-based+Job+Coordinator+Dependency+Isolation#SEP-24:Cluster-basedJobCoordinatorDependencyIsolation-HandlingSamzaApplication.describe

Changes:
1. Add a method `IsolatingClassLoaderFactory.buildSamzaApplicationClassLoader` to construct a classloader for loading the `SamzaApplication` implementation for an app.
2. If isolation is enabled, then call `buildSamzaApplicationClassLoader` to load the `SamzaApplication` implementation in the job coordinator.

Tests:
Deployed `wikipedia-feed` application (uses low-level API, has custom input descriptor, uses Kafka output descriptor) and `wikipedia-parser` application (uses low-level API, Kafka descriptors) in split deployment mode with some debug logs, and verified the following:
1. `TaskApplicationDescriptorImpl` and `KafkaSystemDescriptor` came from the framework infrastructure classloader.
2. `WikipediaSystemDescriptor` (custom descriptor) came from the application classloader.
3. `WikipediaFeedTaskApplication` (impl of `SamzaApplication`) came from the newly built application describe classloader.
4. Lambda which was used for the `TaskFactory` in `WikipediaFeedTaskApplication` came from the newly built application describe classloader, and it implemented the `StreamTaskFactory` interface which was loaded by the framework API classloader.
5. Verified expected system-stream metadata from custom system (`WikipediaSystemFactory`)
Deployed `wikipedia-parser` application (uses low-level API, Kafka descriptors) and verified the classloaders using debug logs.

API/usage changes: None

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/application/ApplicationUtil.java
+++ b/samza-core/src/main/java/org/apache/samza/application/ApplicationUtil.java
@@ -33,14 +33,28 @@ import org.apache.samza.config.TaskConfig;
  */
 public class ApplicationUtil {
 
+  /**
+   * Creates the {@link SamzaApplication} object from the task or application class name specified in {@code config}.
+   *
+   * This will use the current classloader to load the {@link SamzaApplication}.
+   *
+   * @param config the configuration of the application
+   * @return the {@link SamzaApplication} object
+   */
   public static SamzaApplication fromConfig(Config config) {
     return fromConfig(config, false);
   }
 
   /**
-   * Creates the {@link SamzaApplication} object from the task or application class name specified in {@code config}
+   * Creates the {@link SamzaApplication} object from the task or application class name specified in {@code config}.
+   * If {@code canApplyFrameworkIsolationIfEnabled} is true and isolation is enabled (in the config), then this will
+   * build the {@link SamzaApplication} using a special isolating classloader. Otherwise, this will use the current
+   * classloader to load the {@link SamzaApplication}.
    *
    * @param config the configuration of the application
+   * @param canApplyFrameworkIsolationIfEnabled if this is enabled and isolation is enabled in the config, then this
+   *                                            will build the {@link SamzaApplication} using a special isolating
+   *                                            classloader
    * @return the {@link SamzaApplication} object
    */
   public static SamzaApplication fromConfig(Config config, boolean canApplyFrameworkIsolationIfEnabled) {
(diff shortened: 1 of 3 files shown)


samza-core/src/main/java/org/apache/samza/application/ApplicationUtil.java
```diff
@@ -32,17 +34,37 @@
...
    * @return the {@link SamzaApplication} object
    */
   public static SamzaApplication fromConfig(Config config) {
+    return fromConfig(config, false);
+  }
+
+  /**
+   * Creates the {@link SamzaApplication} object from the task or application class name specified in {@code config}.
+   * If {@code canApplyFrameworkIsolationIfEnabled} is true and isolation is enabled (in the config), then this will
+   * build the {@link SamzaApplication} using a special isolating classloader. Otherwise, this will use the current
```
mynameborat, line 51: why do we need two levels (this boolean & the config) to determine the behavior? Can we not infer from the config and then use the appropriate loader?

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
```diff
@@ -121,31 +121,72 @@
...
+        // SamzaApplication is in the application classpath
+        .withClasspath(getClasspathAsURIs(applicationLibDirectory))
+        // getClasspathAsURIs should only return JARs within applicationLibDirectory anyways, but doing it to be safe
+        .withOriginRestriction(OriginRestriction.denyByDefault().allowingDirectory(applicationLibDirectory, false))
+        .withParentRelationship(DelegateRelationshipBuilder.builder()
+            .withDelegateClassLoader(parentClassLoader)
+            // almost everything should come from the parent classloader
+            .addDelegatePreferredClassPredicate(new GlobMatcher("*"))
+            // blacklisting the SamzaApplication class from the parent so that it gets loaded by this classloader
+            .addBlacklistedClassPredicate(new GlobMatcher(samzaApplicationClassName))
```
mynameborat, line 179: Is it possible that by only blacklisting `samzaApplicationClass`, we end up class loading some of the dependencies of user code through parent class loader assuming both parent and application class loader are capable of classing loading that class?

samza-core/src/main/java/org/apache/samza/application/ApplicationUtil.java
cameronlee314: Currently, we only support using this classloader on the job coordinator. So we can't use this new flow on the processing containers nor `RemoteApplicationRunner` yet. Once we support split deployment on the processing containers and remove the need from `RemoteApplicationRunner`, then we can remove this flag. I will update the doc with that clarification.

samza-core/src/main/java/org/apache/samza/classloader/IsolatingClassLoaderFactory.java
cameronlee314: Ah yes, that is a good catch. I need to think through what the impact of that is.

cameronlee314:
This approach does not properly satisfy the isolation requirements. I will open up a separate PR with a different solution.

PR #1338, opened 2020-04-02
SAMZA-2502: Byte array keys be partitioned based on array contents in…

Zhangyx39:
Issue:
InMemorySystemProducer uses the hashCode of the partition key to decide to which partition the message goes. This works well when the key is an object whose hashCode method can be override. But in the case when the partition key is serialized as a byte[], the message can go to any partition. It turns out that the hash code of a byte array is based on the address in memory but not the content. Therefore, even though two messages may have same key, they can be sent to different partitions after their keys are serialized into byte[] whose hash code is kind of random.

 
Fix:
We want to be able to partition messages based on the contents of the partition keys. An easy fix would be: in the case of byte array, we calculate the hash code with Arrays.hashCode(byte[] input). This allows us to calculate the hash code of the byte array by its contents.

Test:
Added a unit test in TestInMemorySystemProducer.

### DIFF
--- a/samza-core/src/test/java/org/apache/samza/system/inmemory/TestInMemorySystemProducer.java
+++ b/samza-core/src/test/java/org/apache/samza/system/inmemory/TestInMemorySystemProducer.java
@@ -1,0 +1,65 @@
+package org.apache.samza.system.inmemory;
+
+import org.apache.samza.system.OutgoingMessageEnvelope;
+import org.apache.samza.system.SystemStream;
+import org.apache.samza.system.SystemStreamPartition;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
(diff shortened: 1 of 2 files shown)


samza-core/src/test/java/org/apache/samza/system/inmemory/TestInMemorySystemProducer.java
```diff
@@ -0,0 +1,65 @@
+package org.apache.samza.system.inmemory;
```
mynameborat, line 1: Can you fix the rat check and add copyright to this new file?

Zhangyx39: Done

PR #1339, opened 2020-04-03
SAMZA-2503 : Adding retry when getting ssp metadata from kafka

MabelYC:
Currently, we only fetch ssp metadata from kafka only once, ideally we should add retry:
- to be consistent with function getSystemStreamMetadata()
- avoid any unexpected failure due to transient errors.

Issue:
Currently Samza will only fetch ssp metadata from kafka once. So some transient errors may cause the job dies.
Fix:
Adding retry to getSSPMetadata function to avoid issue happens.
Test:
Unit test.

### DIFF
--- a/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
+++ b/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
@@ -344,6 +344,7 @@ public class KafkaSystemAdmin implements SystemAdmin {
             return sspToSSPMetadata;
           }
         };
+
     Function2<Exception, ExponentialSleepStrategy.RetryLoop, BoxedUnit> onExceptionRetryOperation =
         new AbstractFunction2<Exception, ExponentialSleepStrategy.RetryLoop, BoxedUnit>() {
           @Override
--- a/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
+++ b/samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
@@ -313,7 +313,7 @@ public class KafkaSystemAdmin implements SystemAdmin {
    * @param retryBackoff retry backoff strategy
    * @return a map from ssp to sspMetadata which has offsets
    */
-  private Map<SystemStreamPartition, SystemStreamMetadata.SystemStreamPartitionMetadata> getSSPMetadata(
+  Map<SystemStreamPartition, SystemStreamMetadata.SystemStreamPartitionMetadata> getSSPMetadata(
       Set<SystemStreamPartition> ssps, ExponentialSleepStrategy retryBackoff) {
 
     LOG.info("Fetching SSP metadata for: {}", ssps);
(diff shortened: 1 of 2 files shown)


samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -298,24 +298,78 @@ public BoxedUnit apply(Exception exception, ExponentialSleepStrategy.RetryLoop l
...
+   * Given a set of SystemStreamPartition, fetch metadata from Kafka for each
+   * of them, and return a map from ssp to SystemStreamPartitionMetadata for
+   * each of them. This method will return null for oldest and newest offsets
+   * if a given SystemStreamPartition is empty. This method will block and
+   * retry indefinitely until it gets a successful response from Kafka.
+   * @param ssps a set of strings of SSP
+   * @param retryBackoff retry backoff strategy
+   * @return a map from ssp to sspMetadata which has offsets
+   */
+  public Map<SystemStreamPartition, SystemStreamMetadata.SystemStreamPartitionMetadata> getSSPMetadata(
```
mynameborat, line 316: Can this be package private? Also, add a unit tests as well.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
```diff
@@ -599,7 +653,7 @@ public void deleteMessages(Map<SystemStreamPartition, String> offsets) {
       Map<TopicPartition, RecordsToDelete> recordsToDelete = offsets.entrySet()
           .stream()
           .collect(Collectors.toMap(entry ->
-              new TopicPartition(entry.getKey().getStream(), entry.getKey().getPartition().getPartitionId()),
+                  new TopicPartition(entry.getKey().getStream(), entry.getKey().getPartition().getPartitionId()),
```
mynameborat, line 656: nit: revert

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
MabelYC: Modified it to a private one and added unite test to test retry.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaSystemAdmin.java
MabelYC: Done

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaSystemAdminWithMock.java
```diff
@@ -308,4 +309,46 @@ public void testGetSystemStreamPartitionCountsShouldTerminateAfterFiniteRetriesO
...
+        kafkaSystemAdmin.getSSPMetadata(ssps);
+
+    assertEquals("metadata should return for 2 topics", sspMetadata.size(), 2);
+
+    // retried twice because the first fails and the second succeeds
+    Mockito.verify(mockKafkaConsumer, Mockito.times(2)).beginningOffsets(topicPartitions);
+  }
+
+  @Test(expected = SamzaException.class)
+  public void testGetSSPMetadataShouldTerminateAfterFiniteRetriesOnException() throws Exception{
```
mynameborat, line 337: The reason why I suggested to make the retry method package private and write tests against it is because you can override the policy for tests. With the current setup, this test is going to always take ~ 16 seconds regardless.

Given our builds are already impacted due to tests, I'd suggest to override defaults to smaller multipliers and values for tests.

PR #1340, opened 2020-04-07
SAMZA-2336: Disable transactional state restore when not supported

bkonold:
**Feature:** Disable transactional state for use cases that are not yet supported
 
**Changes:** Modifies the logic to accessing the `task.transactional.state.restore.enabled` by returning false if standby or async commit are enabled.

 **Tests:** Added a unit test each for each permutation of configuration.

**API Changes:** None
**Usage Instructions:** None


bkonold:
@mynameborat @prateekm can you take a look at this?

PR #1342, opened 2020-04-10
SAMZA-2509: New config name to represent if split deployment feature is enabled

alnzng:
#### Issues

Currently, we are using `samza.cluster.based.job.coordinator.dependency.isolation.enabled` to show if Samza application enables split deployment on job coordinator.
However, we also need one similar config for the job container with the same purpose. 
So we'd better consolidate them to one generic config `job.split.deployment.enabled`.

#### Changes
1. Renamed `ShellCommandConfig.ENV_CLUSTER_BASED_JOB_COORDINATOR_DEPENDENCY_ISOLATION_ENABLED` to `ShellCommandConfig. ENV_SPLIT_DEPLOYMENT_ENABLED `
2. Renamed `JobConfig. CLUSTER_BASED_JOB_COORDINATOR_DEPENDENCY_ISOLATION_ENABLED ` to `JobConfig. JOB_SPLIT_DEPLOYMENT_ENABLED `

#### Tests

#### API Changes
None

#### Upgrade Instructions
None

#### Usage Instructions
None

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/config/ShellCommandConfig.java
+++ b/samza-core/src/main/java/org/apache/samza/config/ShellCommandConfig.java
@@ -61,21 +61,22 @@ public class ShellCommandConfig extends MapConfig {
    * Set to "true" if split deployment feature is enabled. Otherwise, will be considered false.
    *
    * The launch process for the cluster-based job coordinator and job container depends on the value of this, since it
-   * needs to be known if the cluster-based job coordinator and job container should be launched in an isolated mode.
+   * needs to be known if the cluster-based job coordinator and job container should be launched in a split deployment
+   * mode.
    * This needs to be an environment variable, because the value needs to be known before the full configs can be read
    * from the metadata store (full configs are only read after launch is complete).
    */
   public static final String ENV_SPLIT_DEPLOYMENT_ENABLED = "ENV_SPLIT_DEPLOYMENT_ENABLED";
 
   /**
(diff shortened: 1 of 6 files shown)


samza-core/src/main/java/org/apache/samza/config/ShellCommandConfig.java
```diff
@@ -58,16 +58,14 @@
...
-   * Set to "true" if cluster-based job coordinator dependency isolation is enabled. Otherwise, will be considered
-   * false.
+   * Set to "true" if split deployment feature is enabled. Otherwise, will be considered false.
    *
-   * The launch process for the cluster-based job coordinator depends on the value of this, since it needs to be known
-   * if the cluster-based job coordinator should be launched in an isolated mode. This needs to be an environment
-   * variable, because the value needs to be known before the full configs can be read from the metadata store (full
-   * configs are only read after launch is complete).
+   * The launch process for the cluster-based job coordinator and job container depends on the value of this, since it
+   * needs to be known if the cluster-based job coordinator and job container should be launched in an isolated mode.
```
cameronlee314, line 64: Instead of "isolated", could you please use the term "split deployment". I think the term isolation will generally be replaced by the term "split deployment", so it might get confusing what "isolation" refers to in the future.

alnzng: sure, thanks for pointing out.

alnzng: The related terms are fixed.

PR #1344, opened 2020-04-14
SAMZA-2510: Incorrect shutdown status due to race between runloop and process callback thread

mynameborat:
**Symptom**: A race between the process callback thread and the runloop thread can result in incorrect shutdown status for shutdown triggered by end of stream.
**Issue**: Currently runloop performs message choosing and dispatching in a loop indefinitely until interrupted by a shutdown request or an exception during processing. In async mode, message completion is notified by the process callback thread using `onComplete()` and `onFailure()` to represent corresponding success and failure. The failure callback updates the exception code within the runloop to notify processing failures. Similarly, shutdown can be requested by user code or end of stream messages through various scopes(task level, container level). These requests are notified to the runloop through shutdownNow flag.
Currently as long as shutdownNow flag is not set, the exception code is promptly seen by the runloop and propagated upstream correctly. However, if a shutdown triggered by end of stream is requested, runloop doesn't check the exception code before exiting.
**Changes**: Make sure we check the exception code before we exit. Also, set the exception code on process failure before updating the state to done to promptly signal runloop about failure.
**Tests**: Added unit tests to check if the exception is thrown even if the shutdown is requested ahead of callback failure.
**API Changes**: None
**Usage Instructions**: None
**Upgrade Instructions**: None

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/container/RunLoop.java
+++ b/samza-core/src/main/java/org/apache/samza/container/RunLoop.java
@@ -19,6 +19,7 @@
 
 package org.apache.samza.container;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -181,6 +182,12 @@ public class RunLoop implements Runnable, Throttleable {
         }
       }
 
+      /*
+       * The current semantics of external shutdown request (RunLoop.shutdown()) is loosely defined and run loop doesn't
+       * wait for inflight messages to complete and triggers shutdown as soon as it notices the shutdown request.
+       * Hence, it is possible that the exception may or may not propagated based on order of execution
+       * between process callback and run loop thread.
+       */
       if (throwable != null) {
         log.error("Caught throwable and stopping run loop", throwable);
         throw new SamzaException(throwable);
@@ -205,6 +212,11 @@ public class RunLoop implements Runnable, Throttleable {
   public void shutdown() {
     shutdownNow = true;
     resume();
+  }
+
+  @VisibleForTesting
+  Throwable getThrowable() {
+    return throwable;
   }
 
   /**
--- a/samza-core/src/main/java/org/apache/samza/container/RunLoop.java
+++ b/samza-core/src/main/java/org/apache/samza/container/RunLoop.java
@@ -648,6 +648,8 @@ public class RunLoop implements Runnable, Throttleable {
     @Override
     public void onFailure(TaskCallback callback, Throwable t) {
       try {
+        // set the exception code ahead of marking the message as processed to make sure the exception
+        // is visible to the run loop thread promptly. Refer SAMZA-2510 for more details.
         abort(t);
         state.doneProcess();
         // update pending count, but not offset
--- a/samza-core/src/main/java/org/apache/samza/container/RunLoop.java
+++ b/samza-core/src/main/java/org/apache/samza/container/RunLoop.java
@@ -19,7 +19,6 @@
 
 package org.apache.samza.container;
 
-import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


samza-core/src/test/java/org/apache/samza/container/TestRunLoop.java
```diff
@@ -780,4 +782,27 @@ public void testProcessBehaviourWhenAsyncCommitIsEnabled() throws InterruptedExc
...
+
+    Map<TaskName, TaskInstance> tasks = ImmutableMap.of(taskName0, t0);
+
+    int maxMessagesInFlight = 2;
+    RunLoop runLoop = new RunLoop(tasks, executor, consumerMultiplexer, maxMessagesInFlight, windowMs, commitMs,
+        callbackTimeoutMs, maxThrottlingDelayMs, maxIdleMs, containerMetrics,
+        () -> 0L, false);
+    when(consumerMultiplexer.choose(false))
+        .thenReturn(envelope0)
+        .thenReturn(ssp0EndOfStream)
```
cameronlee314, line 803: Will this test just trigger the `throwable != null` condition (and not the `shutdownNow` condition) for exiting the main while loop? It seems like the `envelope0` gets submitted and then processing starts, so `throwable` could get set before `ssp0EndOfStream` even gets seen. It seems like it depends on when the context switch happens.
If it is non-deterministic what the test is actually covering, then maybe just turn this into a `testExceptionIsPropagated` test (I think this class might need a test for that anyways).
If you think it's worth maintaining a test for this race, maybe you could put a latch into `TestTask` processing somewhere, so you could make sure `shutdownNow` gets set (you could use `RunLoop.shutdown` to set the flag) before the `throwable` gets set.

samza-core/src/main/java/org/apache/samza/container/RunLoop.java
```diff
@@ -648,8 +648,8 @@ public void run() {
     @Override
     public void onFailure(TaskCallback callback, Throwable t) {
       try {
-        state.doneProcess();
         abort(t);
+        state.doneProcess();
```
cameronlee314, line 652: Could you please put a comment about why `abort` has to go before `doneProcess`?

samza-core/src/main/java/org/apache/samza/container/RunLoop.java
```diff
@@ -185,6 +180,11 @@ public void run() {
           containerMetrics.utilization().set(((double) activeNs) / totalNs);
         }
       }
+
+      if (throwable != null) {
```
cameronlee314, line 184: Is the following scenario possible?
1) Async processing has a bad message in flight
2) Manual call to `RunLoop.shutdown`
3) Exit main while loop
4) `throwable` is still null since async processing is incomplete
5) Async processing throws exception
6) Container will exit with "success"

samza-core/src/test/java/org/apache/samza/container/TestRunLoop.java
mynameborat: Good point and let me apply your naming suggestion. As far as testing deterministically that exceptions which get set after `shutdownNow` is set gets propagated, let me do that in a different test by leveraging the custom hook for tests in `TestTask`

samza-core/src/main/java/org/apache/samza/container/RunLoop.java
mynameborat: That is correct. The semantics of manual call to `RunLoop.shutdown` is loosely defined in the sense that we don't wait for things in flight to finish and trigger shutdown as soon as runloop thread sees the request. So this would be in line with that and may or may not propagate the exception.

mynameborat:
Looking into the travis failure.

samza-core/src/main/java/org/apache/samza/container/RunLoop.java
cameronlee314: Could you please add a note somewhere (maybe javadoc and/or PR description) about this scenario? The PR title/description seems to refer to the general shutdown case (which would include `RunLoop.shutdown`), but it looks like you are focusing on the end-of-stream shutdown case in this PR.

mynameborat: Added a java doc, updated the PR description and also added some details to the JIRA ticket to outline what is this PR targeting.

PR #1345, opened 2020-04-14
SAMZA-2506: Inconsistent end of stream semantics in SystemStreamPartitionMetadata

bkonold:
I found this when refactoring side input restore to use RunLoop (https://github.com/apache/samza/pull/1343); this patch is necessary to prevent integration tests from failing after that refactor as marking side inputs as caught up relies on newest offset.

**Symptom:** When an end of stream message is present in an in memory SSP, the new and upcoming offset are reported as 1 higher than they should be.

Example:
| E0, offset: 0 | E1, offset: 1 | ... | E3, offset: 3 | EndOfStream |

In the above case, InMemoryManager will report a newestOffset of 4 and an upcomingOffset of 5; they should be 3 and 4 respectively.

See the jira ticket for a more in depth analysis of semantics of end of stream as it relates to offset metadata.
 
**Cause:** InMemoryManager treats end of stream as any other envelope with a numeric offset even though it is not.
 
**Changes:** Check in `InMemoryManager.constructSystemStreamMetadata` whether the last message in the stream is an EndOfStream, and adjust offsets accordingly.
 
**Tests:** Added a unit test to verify that offsets are adjusted by 1 in the case of EndOfStream.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
+++ b/samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
@@ -181,21 +181,26 @@ class InMemoryManager {
             .stream()
             .collect(Collectors.toMap(entry -> entry.getKey().getPartition(), entry -> {
                 List<IncomingMessageEnvelope> messages = entry.getValue();
-                String oldestOffset = messages.isEmpty() ? null : "0";
-                String upcomingOffset = String.valueOf(messages.size());
-                String newestOffset;
-                if (messages.isEmpty()) {
-                  newestOffset = null;
-                } else if (messages.get(messages.size() - 1).isEndOfStream()) {
-                  newestOffset = messages.size() > 1 ? String.valueOf(messages.size() - 2) : null;
-                  upcomingOffset = String.valueOf(messages.size() - 1);
-                } else {
-                  newestOffset = String.valueOf(messages.size() - 1);
-                  upcomingOffset = String.valueOf(messages.size());
+                Integer oldestOffset = null;
+                Integer newestOffset = null;
+                int upcomingOffset = 0;
+
+                if (!messages.isEmpty() && !messages.get(messages.size() - 1).isEndOfStream()) {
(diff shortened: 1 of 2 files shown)


samza-core/src/main/java/org/apache/samza/system/inmemory/InMemoryManager.java
```diff
@@ -181,12 +181,26 @@ private SystemStreamMetadata constructSystemStreamMetadata(
...
                 List<IncomingMessageEnvelope> messages = entry.getValue();
-                String oldestOffset = messages.isEmpty() ? null : "0";
-                String newestOffset = messages.isEmpty() ? null : String.valueOf(messages.size() - 1);
-                String upcomingOffset = String.valueOf(messages.size());
-
-                return new SystemStreamMetadata.SystemStreamPartitionMetadata(oldestOffset, newestOffset, upcomingOffset);
-
+                Integer oldestOffset = null;
+                Integer newestOffset = null;
+                int upcomingOffset = 0;
```
mynameborat, line 186: maybe simplify this to have it default to original values `0, messages.size() - 1, messages.size` and only modify in case of the last message being end of stream?

bkonold: IMO the definition of oldest and newest offset is such that they should be valid pointers to data. In the case where `messages` is empty or has only a single end of stream message, I think those values would be a misrepresentation.

`upcomingOffset` I am less sure of semantically but looking back again at the documentation In SystemStreamPartitionMetadata suggests it should be null as well in the case of an empty stream. It also doesn't really make sense in the context of bounded streams either (in HdfsSystemConsumer, it is returned as null).

I suggest we init upcomingOffset to null, remove it from being set where there is an end of stream message, and leave the other two as-is.

Thoughts?

bkonold: @mynameborat is this resolved after the latest diff?

PR #1346, opened 2020-04-14
Setting transactional state restore to default enabled

bkonold:
**Feature:** Turning transactional state restore on by default in order to roll out at-least-once fixes.
 
**Changes:** Changes the default value of `task.transactional.state.restore.enabled` from `false` to `true`.
 
 **Tests:** Wrote a test job that writes two sets of data, with a commit in between. Verified that on restart, only the first set of data (which is committed) gets restored.

**API Changes:** None
 
**Upgrade Instructions:** Please reference the [samza configuration page](https://samza.apache.org/learn/documentation/latest/jobs/samza-configurations.html#advanced-storage-configurations) for notes on configuring compaction lag for store changelogs that are backed by Kafka. For existing stores, the store specific configuration (`stores.store-name.changelog.min.compaction.lag.ms`) must be set to match the existing properties of the store's changelog topic. For newly created stores, samza will reference this value when creating the changelog topic.

With transactional state restore enabled, this value represents the maximum time duration a job can be down, or otherwise not have performed a checkpoint, while retaining its ability to revert to a store checkpoint transactionally. Once the latest checkpoint (C1) is older than this value, messages (if any) written _after_ C1 will become eligible for compaction and messages written before C1 with the same keys may be erased from the changelog. This prevents the application from accurately reverting to the state of C1 when it resumes.

Please refer to the configuration documentation for tradeoffs to consider on application startup time and changelog topic size.

Once these configurations have been set, upgrade to the version containing this commit, and re-deploy. Subsequent deploys will now revert to state checkpoints transactionally.
 
**Usage Instructions:** None

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
+++ b/samza-core/src/main/java/org/apache/samza/config/TaskConfig.java
@@ -109,7 +109,7 @@ public class TaskConfig extends MapConfig {
   public static final String TRANSACTIONAL_STATE_CHECKPOINT_ENABLED = "task.transactional.state.checkpoint.enabled";
   private static final boolean DEFAULT_TRANSACTIONAL_STATE_CHECKPOINT_ENABLED = true;
   public static final String TRANSACTIONAL_STATE_RESTORE_ENABLED = "task.transactional.state.restore.enabled";
-  private static final boolean DEFAULT_TRANSACTIONAL_STATE_RESTORE_ENABLED = false;
+  private static final boolean DEFAULT_TRANSACTIONAL_STATE_RESTORE_ENABLED = true;
   public static final String TRANSACTIONAL_STATE_RETAIN_EXISTING_STATE =
       "task.transactional.state.retain.existing.state";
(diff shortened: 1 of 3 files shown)


mynameborat:
Aren't there other instructions around min compaction configs and stuff that you need to be aware of when enabling transactional state? Can you add all of transactional upgrade notes to this section or link it so that it is clear and explicit.

bkonold:
> Aren't there other instructions around min compaction configs and stuff that you need to be aware of when enabling transactional state? Can you add all of transactional upgrade notes to this section or link it so that it is clear and explicit.

AFAIK all the necessary configs have defaults but good point that customers will need to tweak at least some of them (compaction like you said) to match their existing settings. I'll take a look back at anything else required and add it to this PR.

bkonold:
> Code changes looks good to me. Feel free to merge once you resolve the comment.

I've added instructions. I'll work with whoever is doing the next release to make sure these get included in the next release's instructions as well.

bkonold:
Travis shows some errors; will update the PR once fixed but this shouldn't merge before then.

PR #1347, opened 2020-04-15
SAMZA-2511 : Adding logic to handle container stop fail 

rmatharu:
Problem: The standby container manager does not handle container-stop-failures.
These events can happen as a result of certificate/authentication issue during the execution of the container-stop. The problem is the standby-container-failover flow relies on a stop-container succeeding and in this case does not complete the failover. This means the active container, for which a failover was initiated, is never started again.
In case of a container-placement action, that runs into container-stop-fail, the action is declared as failed. 
Cause: Above.
Fix: The fix is for standby-container-manager to intercept and handle these events by continuing the failover by either selecting another standby container (if one is present i.e., rf > 2) or using a standby host or using any-host.
API changes: None
Upgrade Instructions: None

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
+++ b/samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
@@ -249,6 +249,8 @@ public class ContainerManager {
       ContainerPlacementMetadata metaData = getPlacementActionMetadata(processorId).get();
       markContainerPlacementActionFailed(metaData,
           String.format("failed to stop container on current host %s", metaData.getSourceHost()));
+
+      metaData.getResourceRequests().forEach(request -> containerAllocator.cancelResourceRequest(request));
     } else if (processorId != null && standbyContainerManager.isPresent()) {
       standbyContainerManager.get().handleContainerStopFail(processorId, containerId, containerAllocator);
     } else {
--- a/samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
+++ b/samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
@@ -245,6 +245,7 @@ public class ContainerManager {
    * @param containerId last known id of the container deployed
    * @param containerHost host on which container is requested to be deployed
    * @param containerAllocator allocator for requesting resources
+   * TODO: SAMZA-2512 Add integ test for handleContainerStopFail
    */
   void handleContainerStopFail(String processorId, String containerId, String containerHost,
       ContainerAllocator containerAllocator) {
--- a/samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
+++ b/samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
@@ -134,6 +134,12 @@ public class ContainerManager {
       } else if (actionStatus == ContainerPlacementMetadata.ContainerStatus.STOP_IN_PROGRESS) {
         LOG.info("Waiting for running container to shutdown due to existing ContainerPlacement action {}", actionMetaData);
         return false;
+      } else if (actionStatus == ContainerPlacementMetadata.ContainerStatus.STOP_FAILED) {
+        LOG.info("Shutdown on running container failed for action {}", actionMetaData);
+        markContainerPlacementActionFailed(actionMetaData,
+            String.format("failed to stop container on current host %s", actionMetaData.getSourceHost()));
+        resourceRequestState.cancelResourceRequest(request);
+        return true;
       } else if (actionStatus == ContainerPlacementMetadata.ContainerStatus.STOPPED) {
         // If the job has standby containers enabled, always check standby constraints before issuing a start on container
         // Note: Always check constraints against allocated resource, since preferred host can be ANY_HOST as well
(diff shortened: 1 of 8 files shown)


samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -233,6 +233,30 @@ void handleContainerLaunchFail(String processorId, String containerId, String pr
...
+   * Handle the container stop failure for active containers and standby (if enabled).
+   * @param processorId logical id of the container eg 1,2,3
+   * @param containerId last known id of the container deployed
+   * @param containerHost host on which container is requested to be deployed
+   * @param containerAllocator allocator for requesting resources
+   */
+  void handleContainerStopFail(String processorId, String containerId, String containerHost,
+      ContainerAllocator containerAllocator) {
+    if (processorId != null && hasActiveContainerPlacementAction(processorId)) {
+      // Assuming resource acquired on destination host will be relinquished by the containerAllocator,
```
Sanil15, line 246: ContainerAllocator only relinquishes resources when there are no requests in the queue. It will be worth manually releasing the resource in this case.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -233,6 +233,30 @@ void handleContainerLaunchFail(String processorId, String containerId, String pr
...
+      // we mark the placement action as failed, and return.
```
Sanil15, line 247: In the case of ContainerPlacement actions, request in the allocator is waiting for active container to stop and since the active container stop fails here, you also want to cancel the resource request tied to the placement action

resourceRequestState.cancelResourceRequest(request);

I believe ContainerPlacementMetadata maintains a set the resource-request that we can read, but only need to cancel the latest one from that set (since that is the one that is active)

rmatharu: Isnt the container-stop on the active, issued after the resource has already been acquired?
In that case, if the resource has already been acquired, the resource-request for it has already been met, so it should have already been cancelled?

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
rmatharu: There is nothing in the ContainerPlacementMetadata to determine the container-id (not processor-id) of the resources already acquired. 
The containerid in this callback is of the container whose stop failed. 

Given that there are likely many placement actions in the queue (not just one), the resource acquired may be useful for handling subsequent placement actions that are in the queue. 
Stop-fail is rare occurrence, so relying on the allocator to relinquish is perhaps not that bad.

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -233,6 +239,28 @@ void handleContainerLaunchFail(String processorId, String containerId, String pr
...
+   * @param containerId last known id of the container deployed
+   * @param containerHost host on which container is requested to be deployed
+   * @param containerAllocator allocator for requesting resources
+   */
+  void handleContainerStopFail(String processorId, String containerId, String containerHost,
+      ContainerAllocator containerAllocator) {
+    if (processorId != null && hasActiveContainerPlacementAction(processorId)) {
+      // Assuming resource acquired on destination host will be relinquished by the containerAllocator,
+      // We mark the placement action as failed, and return.
+      ContainerPlacementMetadata metaData = getPlacementActionMetadata(processorId).get();
```
Sanil15, line 254: AMRMAsync thread will do a container status update here, allocator thread marks the actions as failed. Please remove the comments.

samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java
```diff
@@ -126,6 +126,30 @@ public void handleContainerLaunchFail(String containerID, String resourceID,
     }
   }
 
+  /**
+   *  Handle the failed stop for a container, based on
+   *  Case 1. If it is standby container, continue the failover
+   *  Case 2. If it is an active container, then log a warning and return.
```
Sanil15, line 132: Please correct the comment here, Case 2. If this is an active container, throw an exception restart the AM.

samza-core/src/main/java/org/apache/samza/clustermanager/StandbyContainerManager.java
```diff
@@ -126,6 +126,30 @@ public void handleContainerLaunchFail(String containerID, String resourceID,
...
+   * @param containerID the ID of the container that has failed
```
Sanil15, line 133: nit: @param containerID logical id ex 0,1,2 of the failed samza container
       @param resourceID physical id of the failed container

samza-core/src/main/java/org/apache/samza/clustermanager/ContainerManager.java
```diff
@@ -233,6 +239,28 @@ void handleContainerLaunchFail(String processorId, String containerId, String pr
...
   }
 
+  /**
+   * Handle the container stop failure for active containers and standby (if enabled).
+   * @param processorId logical id of the container eg 1,2,3
+   * @param containerId last known id of the container deployed
+   * @param containerHost host on which container is requested to be deployed
+   * @param containerAllocator allocator for requesting resources
+   */
+  void handleContainerStopFail(String processorId, String containerId, String containerHost,
```
Sanil15, line 249: Please add a todo for adding an integ test for this, I can pick that up later!

PR #1349, opened 2020-04-23
SAMZA-2514: Refactor codes to make runWithClassLoader method more generic

alnzng:
#### Issues
Currently, we define the `runWithClassLoader` method in `ClusterBasedJobCoordinator` to support split deployment. However, the job container needs a similar `runWithClassLoader` method to support split deployment. So we'd better make the `runWithClassLoader` method more generic, then `LocalContainerRunner` can just use it.

#### Changes
Add new util class `SplitDeploymentUtil ` to define generic `runWithClassLoader` method.

#### Tests
- [x] All unit tests are passed

#### API Changes
None

### Upgrade Instructions
None 

#### Usage Instructions
None

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/util/SplitDeploymentUtil.java
+++ b/samza-core/src/main/java/org/apache/samza/util/SplitDeploymentUtil.java
@@ -82,7 +82,7 @@ public final class SplitDeploymentUtil {
       // wrapping args in object array so that args is passed as a single argument to the method
       runMethod.invoke(null, new Object[]{runMethodArgs});
     } catch (IllegalAccessException | InvocationTargetException e) {
-      throw new SamzaException("Exception while executing runClusterBasedJobCoordinator method", e);
+      throw new SamzaException(String.format("Exception while executing %s method", runMethodName), e);
     }
   }
 }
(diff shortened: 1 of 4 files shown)


samza-core/src/main/java/org/apache/samza/util/SplitDeploymentUtil.java
```diff
@@ -0,0 +1,88 @@
...
+      throw new SamzaException(String.format("Isolation was enabled, but unable to find %s method", runMethodName), e);
+    }
+    // only sets accessible flag for this method instance
+    runMethod.setAccessible(true);
+
+    try {
+      // wrapping args in object array so that args is passed as a single argument to the method
+      runMethod.invoke(null, new Object[]{runMethodArgs});
+    } catch (IllegalAccessException | InvocationTargetException e) {
+      throw new SamzaException("Exception while executing runClusterBasedJobCoordinator method", e);
```
cameronlee314, line 85: Minor: Could you please update the method name in the log message?

alnzng: Thanks for pointing out. Done the fix.

PR #1350, opened 2020-04-23
SAMZA-2498: [Job coordinator isolation] Classloader isolation for SamzaApplication.describe execution in job coordinator

cameronlee314:
Feature: https://cwiki.apache.org/confluence/display/SAMZA/SEP-24%3A+Cluster-based+Job+Coordinator+Dependency+Isolation#SEP-24:Cluster-basedJobCoordinatorDependencyIsolation-HandlingSamzaApplication.describe

Changes: Add a new `samza-framework-descriptors-classes.txt` whitelist which will be the list of descriptors (including table functions) that are considered part of the framework, which apps may use to in their implementation of `SamzaApplication.describe`.

Testing:
1. Deployed `wikipedia-feed` application with isolation enabled and verified (using logs) that the correct classloaders were used to load the descriptors (the app uses a Kafka descriptor and a custom descriptor).
2. Modified `wikipedia-feed` to use a table descriptor and a fake table function which was included as part of the framework descriptors. Verified (using logs) that the correct classloaders loaded the table components in the job coordinator.
Note that further testing is going to be needed when using this new flow on the processing containers. We will need to ensure that the infrastructure classloader can use the descriptors properly when executing processing logic.

API/usage changes:
1. No change for applications.
2. When preparing the tarball for the framework API, an additional `samza-framework-descriptors-classes.txt` needs to be included, and it needs to include the classes (or globs) for the descriptors that are considered part of the framework.

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/classloader/DependencyIsolationUtils.java
+++ b/samza-core/src/main/java/org/apache/samza/classloader/DependencyIsolationUtils.java
@@ -42,9 +42,18 @@ public class DependencyIsolationUtils {
 
   /**
    * Name of the file which contains the class names (or globs) which should be loaded from the framework API
-   * classloader.
+   * classloader by all classloaders.
+   * See {@link IsolatingClassLoaderFactory} for more context about what this is used for.
    */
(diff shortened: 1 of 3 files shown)


cameronlee314:
Not moving forward with SEP-24, closing this branch.

PR #1351, opened 2020-04-24
SAMZA-2515: Kafka consumer synchronized

lakshmi-manasa-g:
**Symptom:** containers dying with ConcurrentModificationException thrown from KafkaConsumer
like `org.apache.samza.SamzaException: java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access...
Caused by: java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access
at org.apache.kafka.clients.consumer.KafkaConsumer.acquire(KafkaConsumer.java:2204)
at org.apache.kafka.clients.consumer.KafkaConsumer.acquireAndEnsureOpen(KafkaConsumer.java:2188) 
at org.apache.kafka.clients.consumer.KafkaConsumer.commitAsync(KafkaConsumer.java:1475)...`

**Cause:**  The root cause is accessing KafkaConsumer inside KafkaConsumerProxy.initializeLags without a synchronized block. 
How it was root caused: From KafkaConsumer.acquire code: KafkaConsumer tracks the current thread accessing the object. If another thread tries to access before the current thread has released then this ConcurrentModificationException is thrown. Checking logs of the failed container, KafkaConsumerProxy thread (say P) had started prior to this exception. Another thread  (say C) on which `KafkaConsumer.commitAsync` was invoked, actually accesses the shared consumer inside a synchronized block (like `synchronized(consumer)`). Thus, it was clear that though exception was thrown in C's code, it was a thread other than C (P in this case) which was holding access but had not entered a sync block prior to that. And of all the methods in P which access the consumer, this is the only method which does not have sync block.

**Changes:** Add a synchronized block around kafkaConsumer access in initializeLags method.

**Tests:** unit test can not simulate this issue


samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaConsumerProxy.java
```diff
@@ -0,0 +1,61 @@
...
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+
+
+public class TestKafkaConsumerProxy {
+
+  @Test
+  public void testSynchronizedAccessOfKafkaConsumer() throws Exception {
```
mynameborat, line 36: I see we don't have tests at all for consumer proxy and thanks for adding one :) However, I have few observations/questions
1. The test still does have some indeterminism and doesn't necessarily tests synchronized access as it claims all the time as it is possible that t1 & t2 run exclusively.
2. Is the intent here to test java `synchronized` or prevent accidental changes to initializeLags making it no longer synchronize on the consumer? 
- If it is the former, it is going to be hard to test consistently and deterministically unless you do something complex gimmicks with latches & `Thread.holdsLock(...)`. IMO, it should be okay to assume Java `synchronized` works as expected
- If it the latter, it is still undoubtedly hard but we can enforce them through other means with useful explicit comments around the code blocks & diligent reviews. 

TLDR; I am fine with not having a test for this scenario instead of having one that is indeterministic in what it is tests. If you prefer to have a test, then it will be nice to have it do what it claims to do.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -170,9 +171,18 @@ Throwable getFailureCause() {
...
     // This is expensive, so only do it once at the beginning. After the first poll, we can rely on metrics for lag.
-    Map<TopicPartition, Long> endOffsets = kafkaConsumer.endOffsets(topicPartitionToSSP.keySet());
+
+    Map<TopicPartition, Long> endOffsets;
+    // Synchronize, in case the consumer is used in some other thread (metadata or something else)
+    synchronized (kafkaConsumer) {
+      endOffsets = kafkaConsumer.endOffsets(topicPartitionToSSP.keySet());
+    }
+    if (endOffsets == null) {
+      throw new SamzaException("Failed to fetch kafka consumer endoffsets for system " + systemName);
```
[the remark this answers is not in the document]
lakshmi-manasa-g: actually, KafkaConsumer.endOffset [does not make a promise](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L2189) that it will be non-null. If i keep digging into implementation details, it looks like it returns an empty hashmap.. but thats an impl detail not an api promise. hence added this.

i agree, for this change (to deal with the sync issue), dealing with this NPE is not needed. but thought it will be a nice to do.

samza-kafka/src/test/java/org/apache/samza/system/kafka/TestKafkaConsumerProxy.java
lakshmi-manasa-g: haha, i debated for a good deal of time about this test and even having a test for this scenario :D 
I agree that this test can not really "test/simulate" the real issue we are solving. I wonder if there can be a test which can really test this scenario. I will remove this test.

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaConsumerProxy.java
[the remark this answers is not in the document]
mynameborat: sounds good 👍

samza-kafka/src/main/java/org/apache/samza/system/kafka/KafkaConsumerProxy.java
```diff
@@ -170,9 +171,18 @@ Throwable getFailureCause() {
     return failureCause;
   }
 
-  private void initializeLags() {
+  @VisibleForTesting
+  void initializeLags() {
```
mynameborat, line 175: revert this to private?

PR #1352, opened 2020-04-24
SAMZA-2516: Migrate BaseKeyValueStorageEngineFactory to be an abstract class instead of trait

cameronlee314:
Issues:
1. If using Scala 2.11, there is a compile error when trying to use a Java class extend a Scala trait with implemented methods. BaseKeyValueStorageEngineFactory can be considered part of the Samza API, but it is a Scala trait with an implemented method, so that restricts extension.
2. It would be good to migrate away from Scala code in general, for consistency with other new Samza code.

Changes:
1. `BaseKeyValueStorageEngineFactory` is now an abstract class written in Java.
2. Added some unit tests for `BaseKeyValueStorageEngineFactory`.
3. The functionality is intended to stay the same.

Tests:
1. Added unit tests
2. Some tests in samza-test use in-memory key-value storage, and those tests pass.
3. Deployed job with "drop.large.messages" enabled and verified that messages get dropped
4. Deployed job with "disallow.large.messages" enabled and verified that exception is thrown
5. Deployed job with changelog enabled and verified that changelog was being written to

API changes:
`BaseKeyValueStorageEngineFactory` is no longer a Scala trait, so that could potentially impact existing inheritors:
1. Any existing Scala classes which uses the `with` syntax to mix in `BaseKeyValueStorageEngineFactory` will need to be updated to use `BaseKeyValueStorageEngineFactory` as an abstract class instead. If an existing Scala class used `extends BaseKeyValueStorageEngineFactory` (such as how `InMemoryKeyValueStorageEngineFactory` and `RocksDbKeyValueStorageEngineFactory` uses `BaseKeyValueStorageEngineFactory`), then it should not need to change.
2. Scala 2.12 compiles traits into interfaces with default methods, so a Java class built against the Scala 2.12 version of `BaseKeyValueStorageEngineFactory` won't be able to use it as an interface. A Java class will need to use `BaseKeyValueStorageEngineFactory` as an abstract class. I suppose `BaseKeyValueStorageEngineFactory` could work as an interface with a default implementation for `getStorageEngine`, but it seems like abstract class fits the pattern better, since `getKVStore` shouldn't be a public method.

Upgrade instructions:
Any custom implementation of `BaseKeyValueStorageEngineFactory` needs to treat it as an abstract class, not a trait/interface. If it is necessary to make changes, only class inheritance structure should need to be changed. The functionality/signature of `getStorageEngine` and the usage of the `getKVStore` abstract method are the same, so those pieces should stay the same.


samza-kv/src/main/java/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.java
```diff
@@ -0,0 +1,221 @@
...
+    } else {
+      KeyValueStore<byte[], byte[]> toBeSerializedStore;
+      if (storageConfig.getDropLargeMessages(storeName)) {
+        toBeSerializedStore = new LargeMessageSafeStore(storeToWrap, storeName, true, maxMessageSize);
+      } else {
+        toBeSerializedStore = storeToWrap;
+      }
+      KeyValueStore<T, U> serializedStore =
+          buildSerializedStore(storeName, registry, toBeSerializedStore, keySerde, msgSerde);
+      return buildMaybeCachedStore(storeName, registry, serializedStore, cacheSize, batchSize);
```
PanTheMan, line 177: Is there a difference here between the order of wrapping each store between disallowing large messages vs. allowing? Ie. in the if statement, you first create the maybeCachedStore first and then the LargeMessageSafeStore, then the SerializedStore vs in the else statement, you first create the LargeMessageSafeStore, the SerializedStore, then the maybeCachedStore?
Also if there isn't a difference in order, couldn't we refactor this section of the code into a function that is passed  the **getDisallowLargeMessages** as a param to return the final store?

samza-kv/src/test/java/org/apache/samza/storage/kv/TestBaseKeyValueStorageEngineFactory.java
```diff
@@ -0,0 +1,273 @@
...
+    assertEquals(expectedLoggedStore, storeProperties.isLoggedStore());
+  }
+
+  /**
+   * @param changelogSSP if non-null, then enables logged store
+   */
+  private StorageEngine callGetStorageEngine(Config config, SystemStreamPartition changelogSSP) {
+    when(this.jobContext.getConfig()).thenReturn(config);
+    return new MockKeyValueStorageEngineFactory(this.rawKeyValueStore).getStorageEngine(STORE_NAME, this.storeDir,
+        this.keySerde, this.msgSerde, this.changelogCollector, this.metricsRegistry, changelogSSP, this.jobContext,
```
PanTheMan, line 270: Is there a need to add a unit test if keySerde or msgSerde is missing?

samza-kv/src/test/java/org/apache/samza/storage/kv/TestBaseKeyValueStorageEngineFactory.java
```diff
@@ -0,0 +1,273 @@
...
+    Config config = new MapConfig(BASE_CONFIG, DISABLE_CACHE);
+    StorageEngine storageEngine = callGetStorageEngine(config, CHANGELOG_SSP);
+    KeyValueStorageEngine<?, ?> keyValueStorageEngine = baseStorageEngineValidation(storageEngine);
+    assertStoreProperties(keyValueStorageEngine.getStoreProperties(), true, true);
+    NullSafeKeyValueStore<?, ?> nullSafeKeyValueStore =
+        assertAndCast(keyValueStorageEngine.getWrapperStore(), NullSafeKeyValueStore.class);
+    SerializedKeyValueStore<?, ?> serializedKeyValueStore =
+        assertAndCast(nullSafeKeyValueStore.getStore(), SerializedKeyValueStore.class);
+    LoggedStore<?, ?> loggedStore = assertAndCast(serializedKeyValueStore.getStore(), LoggedStore.class);
+    // noinspection AssertEqualsBetweenInconvertibleTypes
```
PawasChhokra, line 145: What is this needed for? Can we remove this log line?

samza-kv/src/test/java/org/apache/samza/storage/kv/TestBaseKeyValueStorageEngineFactory.java
```diff
@@ -0,0 +1,273 @@
...
+    assertEquals(this.rawKeyValueStore, loggedStore.getStore());
+  }
+
+  @Test
+  public void testWithLoggedStoreWithCache() {
```
PawasChhokra, line 150: Can we change the name of this function to something like testWithLoggedAndCachedStore()?

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.java
```diff
@@ -0,0 +1,221 @@
...
+      SystemStreamPartition changelogSSP,
+      JobContext jobContext,
+      ContainerContext containerContext,
+      StoreMode storeMode) {
+    Config storageConfigSubset = jobContext.getConfig().subset("stores." + storeName + ".", true);
+    StorageConfig storageConfig = new StorageConfig(jobContext.getConfig());
+    Optional<String> storeFactory = storageConfig.getStorageFactoryClassName(storeName);
+    StoreProperties.StorePropertiesBuilder storePropertiesBuilder = new StoreProperties.StorePropertiesBuilder();
+    if (!storeFactory.isPresent() || StringUtils.isBlank(storeFactory.get())) {
+      throw new SamzaException("Store factory not defined. Cannot proceed with KV store creation!");
```
rmatharu, line 97: Would it help to add the storeName in the error-message, so the user known which store he has not defined the factory for?

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+    if (!storeFactory.get().equals(INMEMORY_KV_STORAGE_ENGINE_FACTORY)) {
+      storePropertiesBuilder.setPersistedToDisk(true);
+    }
+    int batchSize = storageConfigSubset.getInt("write.batch.size", 500);
+    int cacheSize = storageConfigSubset.getInt("object.cache.size", Math.max(batchSize, 1000));
```
rmatharu, line 103: Should we have variables for DEFAULT_BATCH_SIZE and DEFAULT_CACHE_SIZE?

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.java
```diff
@@ -0,0 +1,221 @@
...
+    if (cacheSize > 0 && cacheSize < batchSize) {
+      throw new SamzaException(
+          "A store's cache.size cannot be less than batch.size as batched values reside in cache.");
+    }
+    if (keySerde == null) {
+      throw new SamzaException("Must define a key serde when using key value storage.");
+    }
+    if (msgSerde == null) {
+      throw new SamzaException("Must define a message serde when using key value storage.");
```
rmatharu, line 112: Same as above, might help to have storeName in the error message

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.java
```diff
@@ -0,0 +1,221 @@
...
+    KeyValueStore<K, V> nullSafeStore = new NullSafeKeyValueStore<>(maybeAccessLoggedStore);
+
+    KeyValueStorageEngineMetrics keyValueStorageEngineMetrics = new KeyValueStorageEngineMetrics(storeName, registry);
+    HighResolutionClock clock = buildClock(jobContext.getConfig());
+    return new KeyValueStorageEngine<>(storeName, storeDir, storePropertiesBuilder.build(), nullSafeStore, rawStore,
+        changelogSSP, changelogCollector, keyValueStorageEngineMetrics, batchSize,
+        ScalaJavaUtil.toScalaFunction(clock::nanoTime));
+  }
+
+  private static KeyValueStore<byte[], byte[]> buildMaybeLoggedStore(SystemStreamPartition changelogSSP,
```
rmatharu, line 134: /**
Wraps given store into a LoggedStore if a changeLog for the store is defined, otherwise skips wrapping.
*/

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.java
```diff
@@ -0,0 +1,221 @@
...
+    if (changelogSSP == null) {
+      return storeToWrap;
+    } else {
+      LoggedStoreMetrics loggedStoreMetrics = new LoggedStoreMetrics(storeName, registry);
+      storePropertiesBuilder.setLoggedStore(true);
+      return new LoggedStore<>(storeToWrap, changelogSSP, changelogCollector, loggedStoreMetrics);
+    }
+  }
+
+  private static <T, U> KeyValueStore<T, U> applyLargeMessageHandling(String storeName,
```
rmatharu, line 149: nit: buildStoreWithLargeMessageHandling?

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.java
rmatharu: +1

cameronlee314: Unfortunately, there is a difference. The description of https://github.com/apache/samza/pull/1008 has some more context, although it's not fully consistent with the code since it mentions a non-existent "expect.large.message" config.
I think the idea is that it may be undesirable to cache a large message when it won't be able to be written to the changelog. In order to prevent that, the serialization needs to go before the cached store. However, if it's ok to cache the large message, then that is more performant since the deserialized object can be stored in the cache. So multiple options were provided, but that required these multiple flows.

samza-kv/src/test/java/org/apache/samza/storage/kv/TestBaseKeyValueStorageEngineFactory.java
cameronlee314: This is a comment for suppressing the warning on the next line. It is because the type of the first object is `KeyValueStore<byte[], byte[]>` and the type of the second is `KeyValueStore<?, ?>`. The type generics don't matter in these tests since we are just doing top-level type checks and object equality.

samza-kv/src/main/java/org/apache/samza/storage/kv/BaseKeyValueStorageEngineFactory.java
cameronlee314: I added a couple of javadocs to help clarify.

PR #1353, opened 2020-04-29
SAMZA-2517 : Adding handling and relevant exception message for null in oldest offset from system-admin

rmatharu:
Problem: When a Kafka broker returns null for oldest-offset, because of a OpenJDK bug (https://bugs.openjdk.java.net/browse/JDK-8148463), the collect() operation in TaskSideInputStorageManager can fail with a misleading null-pointer-exception message.
Note however that, Kafka broker cannot return null for oldest-offset. 

Cause: Above.
Fix: The fix is to use the suggested workaround for the bug https://bugs.openjdk.java.net/browse/JDK-8148463
And update the exception and log messages in case the oldest-offset are returned as null from Kafka.
API changes: None
Upgrade Instructions: None

### DIFF
--- a/samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
+++ b/samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
@@ -118,10 +118,10 @@ public class TaskSideInputStorageManager {
     LOG.info("Initializing side input stores.");
 
     Map<SystemStreamPartition, String> fileOffsets = getFileOffsets();
-    LOG.info("File offsets for the task {}: ", taskName, fileOffsets);
+    LOG.info("File offsets for the task {}: {}", taskName, fileOffsets);
 
     Map<SystemStreamPartition, String> oldestOffsets = getOldestOffsets();
-    LOG.info("Oldest offsets for the task {}: ", taskName, fileOffsets);
+    LOG.info("Oldest offsets for the task {}: {}", taskName, oldestOffsets);
 
     startingOffsets = getStartingOffsets(fileOffsets, oldestOffsets);
     LOG.info("Starting offsets for the task {}: {}", taskName, startingOffsets);
@@ -372,15 +372,9 @@ public class TaskSideInputStorageManager {
         // For SSPs belonging to the system stream, use the partition metadata to get the oldest offset
         // if partitionMetadata was not obtained for any SSP, populate oldest-offset as null
         // Because of https://bugs.openjdk.java.net/browse/JDK-8148463 using lambda will NPE when getOldestOffset() is null
-        Map<SystemStreamPartition, String> offsets = new HashMap<>();
         for (SystemStreamPartition ssp : systemStreamToSsp.get(systemStream)) {
-          if (partitionMetadata == null || partitionMetadata.get(ssp.getPartition()) == null) {
-            offsets.put(ssp, null);
-          } else {
-            offsets.put(ssp, partitionMetadata.get(ssp.getPartition()).getOldestOffset());
-          }
+          oldestOffsets.put(ssp, partitionMetadata.get(ssp.getPartition()).getOldestOffset());
         }
-        oldestOffsets.putAll(offsets);
       });
 
     return oldestOffsets;
--- a/samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
+++ b/samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
@@ -346,7 +346,8 @@ public class TaskSideInputStorageManager {
    *   3. Fetches the partition metadata for each system stream and fetch the corresponding partition metadata
    *      and populates the oldest offset for SSPs belonging to the system stream.
    *
-   * @return a {@link Map} of {@link SystemStreamPartition} to their oldest offset.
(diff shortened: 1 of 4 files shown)


samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
```diff
@@ -363,15 +364,22 @@ File getStoreLocation(String storeName) {
...
-              Collectors.toMap(Function.identity(), ssp -> partitionMetadata.get(ssp.getPartition()).getOldestOffset()));
-
+        // if partitionMetadata was not obtained for any SSP, populate oldest-offset as null
+        // Because of https://bugs.openjdk.java.net/browse/JDK-8148463 using lambda will NPE when getOldestOffset() is null
+        Map<SystemStreamPartition, String> offsets = new HashMap<>();
+        for (SystemStreamPartition ssp : systemStreamToSsp.get(systemStream)) {
+          if (partitionMetadata == null || partitionMetadata.get(ssp.getPartition()) == null) {
+            offsets.put(ssp, null);
+          } else {
+            offsets.put(ssp, partitionMetadata.get(ssp.getPartition()).getOldestOffset());
```
bkonold, line 380: This map seems redundant. Why not directly `put` to `oldestOffsets`?

samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
```diff
@@ -363,15 +364,22 @@ File getStoreLocation(String storeName) {
...
         // For SSPs belonging to the system stream, use the partition metadata to get the oldest offset
-        Map<SystemStreamPartition, String> offsets = systemStreamToSsp.get(systemStream).stream()
-          .collect(
-              Collectors.toMap(Function.identity(), ssp -> partitionMetadata.get(ssp.getPartition()).getOldestOffset()));
-
+        // if partitionMetadata was not obtained for any SSP, populate oldest-offset as null
+        // Because of https://bugs.openjdk.java.net/browse/JDK-8148463 using lambda will NPE when getOldestOffset() is null
+        Map<SystemStreamPartition, String> offsets = new HashMap<>();
+        for (SystemStreamPartition ssp : systemStreamToSsp.get(systemStream)) {
+          if (partitionMetadata == null || partitionMetadata.get(ssp.getPartition()) == null) {
```
Sanil15, line 377: SystemStreamMetadata#getSystemStreamPartitionMetadata is also used in CoordinatorStreamStore, ContainerStorageManager and other places, we should also do null handling there right?

Isn't it better to do at StreamMetadataCache level?

rmatharu: the null values are on oldestOffset in the returned metadata-object not the returned objectl itself (returned val is by us and is not-null it appears).

samza-core/src/test/java/org/apache/samza/storage/TestTaskSideInputStorageManager.java
```diff
@@ -182,6 +186,36 @@ public void testGetFileOffsets() {
...
+  public void testGetStartingOffsetsWhenStreamMetadataIsNull() {
+    final String storeName = "test-get-starting-offset-store";
+    final String taskName = "test-get-starting-offset-task";
+
+    Set<SystemStreamPartition> ssps = IntStream.range(1, 6)
+        .mapToObj(idx -> new SystemStreamPartition("test-system", "test-stream", new Partition(idx)))
+        .collect(Collectors.toSet());
+    Map<Partition, SystemStreamMetadata.SystemStreamPartitionMetadata> partitionMetadata = ssps.stream()
+        .collect(Collectors.toMap(SystemStreamPartition::getPartition,
+            x -> new SystemStreamMetadata.SystemStreamPartitionMetadata(null, "1", "2")));
```
bkonold, line 202: What's the purpose for multiple ssps if they are each testing the same condition?

samza-core/src/main/java/org/apache/samza/storage/TaskSideInputStorageManager.java
```diff
@@ -346,7 +346,8 @@ File getStoreLocation(String storeName) {
    *   3. Fetches the partition metadata for each system stream and fetch the corresponding partition metadata
    *      and populates the oldest offset for SSPs belonging to the system stream.
    *
-   * @return a {@link Map} of {@link SystemStreamPartition} to their oldest offset.
+   * @return a {@link Map} of {@link SystemStreamPartition} to their oldest offset. If partitionMetadata could not be
+   * obtained for any {@link SystemStreamPartition} the offset for it is populated as null.
```
bkonold, line 350: thanks for adding docs

bkonold:
@rmatharu the change lgtm once the minor comment on unit test is resolved

prateekm:
Let's update description with https://cwiki.apache.org/confluence/display/SAMZA/SEP-25%3A+PR+Title+And+Description+Guidelines before merging.

samza-core/src/test/java/org/apache/samza/storage/TestTaskSideInputStorageManager.java
rmatharu: the task-sideinputstoragemanager uses maps to index ssps so we should have more than 1, but youre right two are probably enough.

